PR #1, opened 2025-11-16
feat(controller,cni,cli,model)

eminwux:
…e init/runner; expand space model/API

---

> [!NOTE]
> Implements full bootstrap flow that creates realm/space metadata, containerd namespace, CNI dirs/networks, and cgroup v2 paths, with new models/utilities and concise init reporting.
> 
> - **Controller/Runner**:
>   - Add structured bootstrap in `internal/controller` with detailed `BootstrapReport` (realm, space, CNI, cgroups).
>   - Implement realm/space provisioning in `internal/controller/runner` (create/read metadata, ensure containerd namespace, create CNI config, create/ensure cgroups).
> - **CNI**:
>   - New CNI bootstrap and config generation (`internal/cni/{bootstrap.go,config.go,cni.go}`) with defaults under `/opt/cni` and per-space conflist management.
> - **Containerd/cgroups**:
>   - Introduce cgroups v2 management in `internal/ctr/{cgroups.go,client.go,namespaces.go}` and helpers `internal/util/cgroups`.
> - **Models/API & Scheme**:
>   - Add internal `Space` hub model and v1beta1 `SpaceDoc`; extend `RealmStatus`/`SpaceStatus` with `cgroupPath`.
>   - Conversion/normalization for `Space` in `internal/apischeme/scheme.go`.
> - **CLI**:
>   - `kuke init` prints concise bootstrap summary (realm/space actions, CNI dirs, cgroups) and changes default run path to `/opt/kukeon`.
> - **Utils**:
>   - Add naming (`BuildSpaceNetworkName`) and FS helpers for per-space network config.
> - **Dependencies**:
>   - Add `containerd/cgroups/v2` and related libs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8b2e030aaa330debb30d685712b77c0c6e5f72fc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #36, opened 2026-04-19
fix: self-heal stale space conflist with mismatched bridge name

eminwux:
## Summary
- `ensureSpaceCNIConfig` now compares the on-disk conflist's bridge name against `cni.SafeBridgeName(networkName)` and regenerates in place on mismatch, so hosts that already have a pre-`SafeBridgeName` conflist (e.g. the 18-char `kuke-system-kukeon` bridge that trips netlink's `ERANGE`) recover on the next `kuke init` instead of failing to attach the kukeond root container. A new `--force-regenerate-cni` flag on `kuke init` is the operator escape hatch for forcing a rewrite without the diff check.
- Introduces `cni.Manager.ReadBridgeName` + `errdefs.ErrBridgePluginMissing` as the primitives that drive the self-heal decision, and records the regeneration reason in a structured log line for post-mortems.
- Closes gap #1 from `docs/gaps-2026-04-19.md`; the doc is updated in this PR to reflect the resolution.

## Test plan
- [x] `go build ./... && go vet ./...`
- [x] `go test ./internal/controller/runner/...` (new `TestEnsureSpaceCNIConfig_*` cases: heals stale, leaves matching alone, force-overwrites, creates-when-missing)
- [x] `go test ./internal/cni/...` (existing `ReadBridgeName` coverage)
- [ ] Manual: plant a stale `kuke-system-kukeon` bridge conflist, re-run `sudo ./kuke init`, confirm it regenerates (and parity via `kuke get realms` with/without `--no-daemon`)
- [ ] Manual: `sudo ./kuke init --force-regenerate-cni` on a host with a correct conflist, confirm it rewrites

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -114,6 +114,10 @@ var (
 	KUKE_INIT_KUKEOND_IMAGE = DefineKV("KUKE_INIT_KUKEOND_IMAGE", "kuke/init/kukeondImage")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_INIT_NO_WAIT = DefineKV("KUKE_INIT_NO_WAIT", "kuke/init/noWait", "false")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_INIT_FORCE_REGENERATE_CNI = DefineKV(
+		"KUKE_INIT_FORCE_REGENERATE_CNI", "kuke/init/forceRegenerateCNI", "false",
+	)
 
 	// Create command variables
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
--- a/cmd/kuke/init/init.go
+++ b/cmd/kuke/init/init.go
@@ -95,6 +95,19 @@ func setFlags(cmd *cobra.Command) error {
 	if err != nil {
 		return fmt.Errorf("failed to bind flag: %w", err)
 	}
+
+	cmd.Flags().Bool(
+		"force-regenerate-cni", false,
+		"Rewrite space CNI conflists even when they already exist; use to "+
+			"recover from stale on-disk state after a generator fix",
+	)
+	err = viper.BindPFlag(
+		config.KUKE_INIT_FORCE_REGENERATE_CNI.ViperKey,
+		cmd.Flags().Lookup("force-regenerate-cni"),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to bind flag: %w", err)
+	}
 	return nil
 }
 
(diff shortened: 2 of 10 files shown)


eminwux:
## Review — LGTM (self-approval blocked by GitHub)

### Overview
Fixes a latent correctness bug where `kuke init` left pre-`SafeBridgeName` conflists on disk unchanged, causing netlink `ERANGE` on re-init. Adds `cni.Manager.ReadBridgeName`, a `shouldRegenerateSpaceCNI` decision helper, `errdefs.ErrBridgePluginMissing`, and a `--force-regenerate-cni` escape hatch. Includes a detailed gaps doc and solid unit coverage.

### Sensitive data / security
- **No secrets leaked.** The new info log in `provision.go` emits only space name, network name, and a human reason (bridge names + text). Bridge names like `kuke-system-kukeon` are infrastructure identifiers, not sensitive.
- File permissions unchanged: conflist `0o600`, parent dir `0o750`.
- No new network exposure, no credential handling, no user input reaching shell/exec.
- `docs/gaps-2026-04-19.md` contains only internal paths and commit hashes already public in `CLAUDE.md`. The `10.22.0.0/16` subnet in the test fixture matches the existing default in `NewCNINetworkConfig`.

### Correctness
- `WriteSpaceNetworkConfig` uses `os.WriteFile` (truncates) — self-heal overwrite is safe.
- Race handling correct: `ErrNetworkNotFound` between exists-check and read maps to regenerate rather than an error.
- **Silent auto-overwrite on mismatch**: if an operator ever hand-customized their conflist's bridge field, `ensureSpaceCNIConfig` would silently clobber it. Probably intentional (kukeon owns this file), but a one-line contract note in `ensureSpaceCNIConfig` would help a future reader.
- **Orphaned linux bridge not cleaned up**: after self-heal, a pre-existing 18-char bridge netdev lingers until reboot or manual `ip link del`. Out of scope, but worth a follow-up.

### Minor nits
1. `shouldRegenerateSpaceCNI` takes `space intmodel.Space` but never uses it — drop the param.
2. `ReadBridgeName` returns `("", nil)` when the bridge plugin exists but has no `bridge` field. Returning `ErrBridgePluginMissing` for this unusable-entry case would be more consistent with the sentinel contract (current callers handle it fine, so not blocking).
3. Test helper `writeStaleConflist` is reused with the *correct* bridge name in `LeavesMatchingConflistAlone` and `ForceRegenerateCNIOverwrites` — a more neutral name (e.g. `writeHandRolledConflist`) would read better.
4. Consider asserting the logged `reason` value (capturing slog handler) in the heal test, to guard the structured post-mortem output called out in the PR description.

### Test coverage
Strong. `ReadBridgeName` has 8 cases (including malformed JSON and empty path), and the provision-level tests cover heal / no-op / force / create. The force-overwrite test relies on `BuildDefaultConflist`'s formatting differing from the hand-rolled fixture — fragile against a future formatter change, but not wrong today.

### Verdict
Self-heal is the right mechanism, logic is sound, no sensitive data leaked. Nits are non-blocking.

PR #52, opened 2026-04-21
feat: add container security fields (user, read-only rootfs, caps, seccomp, tmpfs, resources)

### LINKED ISSUE #43: Container - security fields (user, readOnlyRootFilesystem, capabilities, securityOpts, tmpfs, resources)
## Context

The Container spec currently exposes only `privileged: true/false`. Running untrusted code in a container (the dominant agent workload pattern, but not only that) requires the standard set of isolation knobs `docker run` already exposes. Without them kukeon is strictly less expressive than compose for untrusted workloads.

Traceable to the proposal "Kukeon for AI Agents: A Proposal for Agent-Native Orchestration" §4.2.

## Problem

The Container spec has only `privileged: true/false`. For untrusted in-container code, orchestrators need finer control.

## Proposal

Add the following fields to Container `spec`, each with safe defaults:

| Field | Type | Default | Purpose |
|---|---|---|---|
| `user` | `"uid:gid"` | image default | Run as non-root |
| `readOnlyRootFilesystem` | bool | `false` (for compat), `true` recommended | Lock rootfs |
| `capabilities.drop` | `[]string` | `[]` | Drop Linux capabilities |
| `capabilities.add` | `[]string` | `[]` | Add Linux capabilities (rarely needed) |
| `securityOpts` | `[]string` | `[]` | Passthrough for `no-new-privileges`, seccomp profiles |
| `tmpfs` | `[]{path, sizeBytes, options}` | `[]` | tmpfs mounts for ephemeral scratch |
| `resources.memoryLimitBytes` | int | unset | Hard memory ceiling |
| `resources.cpuShares` | int | unset | CPU weight |
| `resources.pidsLimit` | int | unset | Prevent fork bombs |

This is the minimum to express what `docker run --user X --read-only --cap-drop=ALL --security-opt=no-new-privileges --tmpfs=/tmp --memory=4g --pids-limit=512` expresses today.

## Acceptance criteria

- [ ] Each field above is accepted by the API and enforced by the runtime
- [ ] `user` accepts `"uid:gid"` form and falls back to image default when unset
- [ ] `readOnlyRootFilesystem: true` makes the rootfs read-only
- [ ] `capabilities.drop` / `capabilities.add` apply via OCI spec capabilities
- [ ] `securityOpts` passes through (including `no-new-privileges` and seccomp profile paths)
- [ ] `tmpfs` mounts at the declared path with the declared size and options
- [ ] Resource limits (`memoryLimitBytes`, `cpuShares`, `pidsLimit`) apply via cgroup v2
- [ ] All fields round-trip via `kuke get -o yaml`

## Out of scope

- Inheriting these fields from Space — that is a separate issue (Space defaults)
- User-namespace remapping
- AppArmor / SELinux profile management beyond `securityOpts` passthrough

## Dependencies

None.

eminwux:
## Summary
- Extend the Container spec with the standard isolation knobs agents and untrusted workloads need: `user`, `readOnlyRootFilesystem`, `capabilities.drop/add`, `securityOpts`, `tmpfs`, and `resources.{memoryLimitBytes,cpuShares,pidsLimit}`. Each field round-trips through the v1beta1 API, the internal hub, and `kuke get -o yaml`.
- Translate the new fields into OCI SpecOpts in `ctr.BuildContainerSpec`: `oci.WithUser`, `oci.WithRootFSReadonly`, capability add/drop, tmpfs mounts (size promoted to the standard `size=N` option), memory / CPU shares / pids limits via `oci.With*`, and a `securityOpts` parser that understands `no-new-privileges` and `seccomp=unconfined|<path>` and rejects unknown keys.
- Wire the fields through `controller.CreateContainer` (previously dropped) and the apply-diff so `kuke apply` detects changes to them. Add new CLI flags mirroring `docker run --user --read-only --cap-drop --cap-add --security-opt --tmpfs --memory --cpu-shares --pids-limit` on `kuke create container`.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test $(go list ./... | grep -v '/e2e$')` — new tests pass; the two pre-existing `internal/ctr` failures (`TestDefaultRootContainerSpec`, `TestBuildRootContainerSpec`) reproduce on `main` and are unrelated to this change
- [x] `make kuke` builds an ELF binary; `kuke create container --help` shows the new flags
- [x] `./kuke create container test --memory notabyte --cell foo` → clear parse error; `./kuke create container test --tmpfs /tmp:size=bad --cell foo` → size-parse error
- [ ] Daemon-parity smoke test from CLAUDE.md (`kuke init` + `kuke get realms` via daemon and `--no-daemon`) — requires a host with containerd and root privileges; not runnable in this sandbox

Closes #43

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -189,6 +189,40 @@ var (
 	)
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_CREATE_CONTAINER_LABELS = DefineKV("KUKE_CREATE_CONTAINER_LABELS", "kuke/create/container/labels")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_CREATE_CONTAINER_USER = DefineKV("KUKE_CREATE_CONTAINER_USER", "kuke/create/container/user")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_CREATE_CONTAINER_READ_ONLY = DefineKV(
+		"KUKE_CREATE_CONTAINER_READ_ONLY",
+		"kuke/create/container/readOnlyRootFilesystem",
+		"false",
+	)
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_CREATE_CONTAINER_CAP_DROP = DefineKV("KUKE_CREATE_CONTAINER_CAP_DROP", "kuke/create/container/capabilities/drop")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_CREATE_CONTAINER_CAP_ADD = DefineKV("KUKE_CREATE_CONTAINER_CAP_ADD", "kuke/create/container/capabilities/add")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_CREATE_CONTAINER_SECURITY_OPTS = DefineKV(
+		"KUKE_CREATE_CONTAINER_SECURITY_OPTS",
+		"kuke/create/container/securityOpts",
+	)
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_CREATE_CONTAINER_TMPFS = DefineKV("KUKE_CREATE_CONTAINER_TMPFS", "kuke/create/container/tmpfs")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_CREATE_CONTAINER_MEMORY = DefineKV(
+		"KUKE_CREATE_CONTAINER_MEMORY",
+		"kuke/create/container/resources/memoryLimitBytes",
+	)
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_CREATE_CONTAINER_CPU_SHARES = DefineKV(
+		"KUKE_CREATE_CONTAINER_CPU_SHARES",
+		"kuke/create/container/resources/cpuShares",
+	)
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_CREATE_CONTAINER_PIDS_LIMIT = DefineKV(
+		"KUKE_CREATE_CONTAINER_PIDS_LIMIT",
+		"kuke/create/container/resources/pidsLimit",
+	)
 
 	// Get command variables
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
--- a/cmd/kuke/create/container/container.go
+++ b/cmd/kuke/create/container/container.go
@@ -18,6 +18,7 @@ package container
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/eminwux/kukeon/cmd/config"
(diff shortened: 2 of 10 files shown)


eminwux:
**PR #52 Review — `feat: add container security fields (user, read-only rootfs, caps, seccomp, tmpfs, resources)`**

Scope is well-structured — new fields thread cleanly through v1beta1 → internal hub → `ctr.BuildContainerSpec` → OCI SpecOpts, with matching apply-diff and round-trip tests. Security-opt parsing (`no-new-privileges`, `seccomp=unconfined|<path>`, error on unknown keys) is the right posture. One blocking correctness issue, and a couple of smaller items.

A few items:

- **Correctness (blocking)** (`internal/ctr/spec.go:904-911`): the `drop: ["ALL"]` idiom does not actually drop all capabilities. `oci.WithDroppedCapabilities` in containerd v2.2.0 (`pkg/oci/spec_opts.go:1099`) is a literal string-match removal — it iterates each existing cap and removes entries that string-equal a name in the drop list. `"ALL"` is not a cap name, so nothing is removed. `populateDefaultUnixSpec` (`pkg/oci/spec.go:157`) seeds `Process.Capabilities.{Bounding,Permitted,Effective}` with `defaultUnixCaps()` (CAP_CHOWN, CAP_DAC_OVERRIDE, CAP_NET_RAW, …) at container-create time, and `oci.WithImageConfig` doesn't alter that set. End result in production: `drop: ["ALL"], add: ["NET_ADMIN"]` yields `defaultUnixCaps ∪ {CAP_NET_ADMIN}`, not the expected `{CAP_NET_ADMIN}`. The test `TestBuildContainerSpec_Capabilities` passes vacuously because `applyBuiltSpec` (`internal/ctr/spec_security_test.go:1103-1116`) constructs an empty `runtimespec.Spec` that bypasses `populateDefaultUnixSpec`. Fix: special-case `"ALL"` in `securitySpecOpts` — when Drop contains it (post-normalization), emit `oci.WithCapabilities(nil)` *before* applying the add list. And harden the test: pre-populate `spec.Process.Capabilities.{Bounding,Effective,Permitted}` with a non-empty default set so drop-ALL has something to clear.
- **Input validation (non-blocking)** (`internal/ctr/spec.go:922-931`): `oci.WithMemoryLimit(uint64(*MemoryLimitBytes))`, `oci.WithCPUShares(uint64(*CPUShares))`, `oci.WithPidsLimit(*PidsLimit)` — CLI input is rejected via `parseMemoryBytes` when negative, but YAML/API ingestion in `convertResourcesToInternal` (`internal/apischeme/scheme.go:569`) copies the `*int64` through unchecked. A negative memory value from `kuke apply` would wrap on the `uint64` cast. Reject non-positive values at the scheme boundary (or in `securitySpecOpts`) so the invariant holds regardless of input path.
- **Tmpfs default options (non-blocking)** (`internal/ctr/spec.go:968`): defaults are `nosuid, nodev` only — Docker's default also includes `noexec`. Not wrong (agents sometimes need exec), but worth a README/CLI-help line so callers aren't surprised.

LGTM once the drop-ALL behavior is fixed and the capabilities test exercises a populated default set — the other two are non-blocking.

eminwux:
Addressed in eb31af7:
- drop-ALL now emits `oci.WithCapabilities(nil)` instead of `oci.WithDroppedCapabilities(["ALL"])`, so the default Unix cap set is actually cleared before the add list applies. Handles both `ALL` and `CAP_ALL` spellings.
- `TestBuildContainerSpec_Capabilities` now pre-populates `Process.Capabilities.{Bounding,Permitted,Effective}` with the containerd default set, so drop-ALL has something to clear — the assertion of "only CAP_NET_ADMIN" is no longer vacuous. Also asserts `Permitted` in addition to `Effective`/`Bounding`.

Deferred (non-blocking, per the review):
- Memory/CPU/Pids negative-value rejection at the scheme boundary — follow-up.
- Tmpfs default options (add `noexec` or document) — follow-up.

eminwux:
**PR #52 Review — `feat: add container security fields (user, read-only rootfs, caps, seccomp, tmpfs, resources)`**

Solid, cohesive feature: fields round-trip end-to-end through CLI → external API → internal model → diff → OCI SpecOpts, with round-trip and per-field unit tests. Security handling is careful — drop-ALL clears the default cap set instead of string-matching an "ALL" pseudo-name (`internal/ctr/spec.go:295-310`), capability names are normalized (`NET_ADMIN` → `CAP_NET_ADMIN`), and unknown `--security-opt` keys fail at spec-apply time with a clear error rather than silently passing. No sensitive-data leaks in error messages. I also verified that the two `internal/ctr` failures the test plan calls out (`TestDefaultRootContainerSpec`, `TestBuildRootContainerSpec`) reproduce on `main` — they're about a hardcoded `registry.eminwux.com/busybox:latest` image, unrelated to this change.

- **Clone consistency** (`pkg/api/model/v1beta1/container.go:207-210`): `res := *out.Spec.Resources` is a shallow struct copy — the `*int64` pointer fields (`MemoryLimitBytes`, `CPUShares`, `PidsLimit`) stay aliased with the input. The Capabilities/Tmpfs clone branches above are more defensive; Resources should match. Low risk in practice since those int64s aren't typically mutated post-construction, but it's an inconsistency worth tidying in a follow-up.
- **Pre-existing gap** (`internal/controller/create_container.go:74-89`): the PR wires the new security fields through `controller.CreateContainer`, but `Env`, `Ports`, `Volumes`, `Networks`, `NetworksAliases`, `Privileged`, `CNIConfigPath`, and `RestartPolicy` are still omitted from `newContainerSpec` — same pattern as main. Out of scope here, but worth confirming those fields actually reach the container through another path; if not, `--env`/`--port`/etc. on `kuke create container` silently drop. Not this PR's problem to fix; just flagging.
- **Daemon-parity smoke test** unchecked in the test plan is fine here — this PR doesn't touch the daemon startup or bind-mount path, so `kuke get realms` wouldn't exercise the new fields.

LGTM.

PR #53, opened 2026-04-22
feat: add container secrets with fromFile/fromEnv injection

### LINKED ISSUE #47: Container - scoped credential injection via `secrets` (fromFile / fromEnv)
## Context

Credentials today flow via `env: ["KEY=..."]`, which puts secrets directly in the manifest, in logs, and in `kuke get -o yaml` output. Workloads that need real credentials (agents being the obvious case) need credential delivery that is safe to commit.

Traceable to the proposal "Kukeon for AI Agents: A Proposal for Agent-Native Orchestration" §4.6.

## Problem

Today, credentials are passed via `env: ["ANTHROPIC_API_KEY=..."]`, which puts the secret in the manifest, in logs, and in `kuke get -o yaml` output. For agents, this is the wrong ergonomics *and* the wrong security posture.

## Proposal

Add a `secrets` field on Container (naturally also inheritable from Space/Session in a future issue) that references credentials by source, not by value:

```yaml
secrets:
  - name: ANTHROPIC_API_KEY
    fromFile: /etc/kukeon/secrets/anthropic.key
  - name: GITHUB_TOKEN
    fromEnv: GITHUB_TOKEN_SCOPED     # on the host, set by a wrapper
```

Values are mounted into the container's environment or as files, never written into status or persisted in audit logs. The manifest is safe to commit.

An extended version — future work — would integrate with an external broker (Vault, systemd credentials, cloud KMS) so that credentials are minted fresh per Session and revoked on Session end.

## Acceptance criteria

- [ ] `fromFile` reads the file on the daemon host at apply time and injects its contents into the container as the named env var
- [ ] `fromEnv` reads the named env var on the daemon host and injects it into the container
- [ ] Secret values never appear in `kuke get -o yaml`, in object status, or in daemon logs
- [ ] Missing source file or env var produces a clear error at apply time
- [ ] A file-mount mode is available (not only env-var injection), with 0400 permissions inside the container
- [ ] The secret reference survives round-trip; the secret value never does

## Out of scope

- External broker integration (Vault, KMS, systemd credentials) — future issue
- Rotation and revocation semantics — future issue
- Session-scoped credential minting — future issue

## Dependencies

None strictly. Naturally composes with Session (per-Session credential scope) once Session lands, but works standalone.

eminwux:
## Summary
- Adds `spec.secrets` to Container so credentials flow by reference, not by value: `fromFile` reads a path on the daemon host at apply time, `fromEnv` reads an env var on the daemon host. Values are injected as env vars by default, or mounted read-only at 0400 when `mountPath` is set.
- Only the reference (name / source / mountPath) lives in the internal model, YAML, status, and logs — resolved values are never persisted there. File-mount staging goes to `/run/kukeon/secrets/<containerdId>/<name>` with owner-only perms.
- Validation (missing source, both sources, missing name, relative mountPath) runs at apply time with clear errdefs; resolution of a missing file / unset env var also errors loudly with sentinels callers can match.

Why this shape: the proposal §4.6 wants credentials that are safe to commit in manifests. Keeping the value out of the hub/YAML surface is the single invariant that makes the reference model work; everything else (mount mode, staging path, 0400) is contract detail. External broker integration / rotation / session-scoped minting stay out of scope per the issue.

## Test plan
- [x] `go test ./internal/apply/parser/... ./internal/apischeme/... ./internal/ctr/ -run TestResolveSecrets`
- [x] `go build ./...` and `go vet ./...` clean
- [x] `make test` — passes except for two pre-existing failures in `internal/ctr/container_utils_test.go` (`TestDefaultRootContainerSpec`, `TestBuildRootContainerSpec`) introduced by 594f789: the test file expects `registry.eminwux.com/busybox:latest` but `DefaultRootContainerImage` in `container_utils.go` still says `docker.io/library/busybox:latest`. Unrelated to this change; flagged for separate fix.
- [x] Project CLAUDE.md smoke test: rebuild binaries, reload `kukeon-local:dev`, `sudo ./kuke init` succeeds, daemon/no-daemon `get realms` return identical output.

Closes #47

### DIFF
--- a/docs/site/manifests/container.md
+++ b/docs/site/manifests/container.md
@@ -63,6 +63,7 @@ See [Concepts → Container](../concepts/container.md) for what a container is.
 | `networks`         | array of string  | no       | Additional CNI networks to join beyond the cell's default                               |
 | `networksAliases`  | array of string  | no       | DNS aliases for the container within its CNI networks                                   |
 | `privileged`       | bool             | no       | Run privileged (full capabilities, access to `/dev`, etc.)                              |
+| `secrets`          | array of `ContainerSecret` | no | Inject credentials resolved by the daemon — never written to status or YAML (see [ContainerSecret](#containersecret)) |
 | `cniConfigPath`    | string           | no       | Override the CNI config directory for this container                                    |
 | `restartPolicy`    | string           | no       | Restart policy. Reserved — restart semantics are not finalized.                         |
 
@@ -78,6 +79,32 @@ Each entry in `spec.volumes` is a bind mount of a host path into the container.
 | `source`   | string | yes      | Absolute host path. Must exist at apply time. Named / managed volumes are not supported. |
 | `target`   | string | yes      | Absolute path inside the container                                          |
 | `readOnly` | bool   | no       | Mount read-only when `true` (writes fail with `EROFS`). Defaults to `false`. |
+
+### ContainerSecret
+
+Each entry in `spec.secrets` references a credential the daemon resolves at apply time. Only the reference is persisted — the resolved value never appears in `kuke get -o yaml`, in object status, or in daemon logs.
+
+| Field        | Type   | Required | Description                                                                                                |
+|--------------|--------|----------|------------------------------------------------------------------------------------------------------------|
+| `name`       | string | yes      | Environment-variable name (default mode) or basename of the mounted secret file when `mountPath` is set    |
+| `fromFile`   | string | one of required | Absolute host path the daemon reads at apply time. Missing files produce a clear error.             |
+| `fromEnv`    | string | one of required | Name of an environment variable set on the daemon host. Missing env vars produce a clear error.     |
+| `mountPath`  | string | no       | Absolute path inside the container. When set, the secret is staged with mode `0400` and bind-mounted read-only instead of being injected as an env var. |
+
+Exactly one of `fromFile` / `fromEnv` must be set. Example:
+
+```yaml
+secrets:
+  - name: ANTHROPIC_API_KEY
+    fromFile: /etc/kukeon/secrets/anthropic.key
+  - name: GITHUB_TOKEN
+    fromEnv: GITHUB_TOKEN_SCOPED
+  - name: tls.crt
+    fromFile: /etc/kukeon/secrets/tls.crt
+    mountPath: /run/secrets/tls.crt
+```
+
+File-mount mode stages secrets under `/run/kukeon/secrets/<containerdId>/<name>` on the host, with owner-only read perms, then bind-mounts them read-only into the container. Because containerd persists resolved env vars in its own runtime spec, env-injection mode leaves the value in containerd's state; file-mount mode keeps it only in the tmpfs staging file.
 
 ## status
 
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -284,6 +284,7 @@ func ConvertContainerDocToInternal(in ext.ContainerDoc) (intmodel.Container, err
 				SecurityOpts:           in.Spec.SecurityOpts,
 				Tmpfs:                  convertTmpfsMountsToInternal(in.Spec.Tmpfs),
 				Resources:              convertResourcesToInternal(in.Spec.Resources),
+				Secrets:                convertSecretsToInternal(in.Spec.Secrets),
 				CNIConfigPath:          in.Spec.CNIConfigPath,
 				RestartPolicy:          in.Spec.RestartPolicy,
 			},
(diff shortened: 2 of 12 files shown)


eminwux:
**PR #53 Review — `feat: add container secrets with fromFile/fromEnv injection`**

Solid shape overall: the "reference-only in the hub, value only at resolve time" invariant is cleanly enforced by having no value field on either the external or internal `ContainerSecret`, and the YAML-leak test pins it. Staging is atomic (CreateTemp → Chmod 0400 → Rename) and error messages are carefully scoped to name/path/env-var-name — no resolved values end up in logs or wrapped errors. Validation sentinels are all referenced, and the CLAUDE.md smoke test is checked off in the description.

One blocking item:

- **Dead public API** (`internal/ctr/client.go:75,102`): `SetSecretsStagingDir` is added to the exported `Client` interface (and implemented on `*client`), but has zero callers anywhere in the tree. The tests don't use it — they call `resolveSecrets` directly with a passed-in staging dir. Either wire it up from `kukeond serve` (e.g., a `--secrets-staging-dir` flag, analogous to `--run-path`) so operators can override the default, or drop it from the interface until a caller exists. As-is it's public surface nothing touches, and per project convention dead public API is blocking.

A few non-blocking items:

- **Validation duplication** (`internal/apply/parser/parser.go:378-304` vs `internal/ctr/secrets.go:70-93`): the name/source/mountPath checks are repeated verbatim. Defense-in-depth makes sense (parser runs before hub persist, resolve runs on internal model which may bypass the parser), but consider extracting a single `validateContainerSecret(intmodel.ContainerSecret)` in `modelhub` or `errdefs` adjacent and having both sites call it — one less place to drift when a new rule is added.
- **Stale staging files on container replacement** (`internal/ctr/secrets.go:105-121`): if a container is deleted and recreated with the same containerd ID but a different secret set, old staged files remain under `/run/kukeon/secrets/<id>/` because nothing cleans the per-container dir on delete. On tmpfs `/run` this self-heals on reboot, but a long-running daemon accumulates orphans. Fine to address in a follow-up.
- **Trailing whitespace on fromFile reads** (`internal/ctr/secrets.go:137`): `os.ReadFile` returns the file bytes verbatim — if an operator's `anthropic.key` has a trailing newline (common with `echo "..." > file`), the env-injected or mounted value will include it. For mount mode this usually doesn't matter (consumer trims), but for env mode an API client may reject `Authorization: Bearer sk-...\n`. Consider either documenting "no trailing newline" in the docs table or `TrimRight(data, \"\\n\")` with the trade-off noted. Not blocking — just a foot-gun.
- **Test plan note** — the pre-existing `DefaultRootContainerImage` test failures you flagged are unrelated to this PR and shouldn't gate merge; agree with the call to split into a separate fix.

LGTM to proceed once `SetSecretsStagingDir` is either wired up or dropped — the rest are non-blocking.

eminwux:
Addressed in a2730c3:
- Dropped `SetSecretsStagingDir` from the `ctr.Client` interface, removed the method, the `secretsStagingDir` field, and the constructor assignment. `CreateContainerFromSpec` now passes `DefaultSecretsStagingDir` directly to `resolveSecrets`. Chose the "drop it" option over wiring a `--secrets-staging-dir` flag: adding a flag operators never asked for would be speculative scope; the constant already governs the staging path, and the field is a one-line reintroduction if a real caller ever lands. Tests use `resolveSecrets` directly with their own staging dirs, so no test updates needed.

Deferred (per your non-blocking items):
- Validation helper extraction between `parser.go` and `secrets.go` — agreed, good follow-up once a third call site appears.
- Per-container staging dir cleanup on container delete — follow-up; `/run` self-heals on reboot in the meantime.
- Trailing-newline handling on `fromFile` reads — follow-up; worth a docs note + explicit decision rather than a silent `TrimRight`.
- Pre-existing `DefaultRootContainerImage` test failures — tracking separately, not part of this PR.

`make test` passes except those two pre-existing failures. Build + vet clean.

eminwux:
**PR #53 Review — `feat: add container secrets with fromFile/fromEnv injection`**

Clean implementation of the reference-only secret contract from proposal §4.6. Validation is layered correctly (parser-time shape check + apply-time resolution with sentinel errors), file mode is strict (staging dir `0700`, staged file `0400` via temp-file + rename), and the internal model carries no value field — so `TestContainerSecretYAMLNeverLeaksValues` is guarded by the type system, not just a runtime assertion. Sensitive-data surface is tight: error messages include the secret's declared name and the source path / env-var name, never the resolved value.

- **Staged-file cleanup** (`internal/ctr/secrets.go`): there is no companion teardown that removes `/run/kukeon/secrets/<containerdID>/` when a container is deleted. `/run` is typically tmpfs so reboots sweep it up, but a long-lived daemon with many create/delete cycles will leak per-container directories. Not blocking for this PR (out of scope per the issue), but worth a follow-up.
- **Secret value shape**: `readSecretValue` returns `string(data)` with no trimming, so a `fromFile` that ends with a trailing newline (common for `.pem`/`.key` files written by editors) injects the newline verbatim into `NAME=value\n`. Fine for mount mode; surprising for env mode. Worth calling out in the docs table that env-mode values are raw bytes from the source.
- **Errdefs alignment** (`internal/errdefs/errdefs.go:118-124`): `ErrSecretMountPathNotAbsolute` is longer than the other identifiers in the group and pushes the `=` column out by one, leaving the shorter neighbors misaligned. Cosmetic — `gofmt` would fix it if re-run.

LGTM to proceed — all checks green (CodeQL actions/go/core all SUCCESS), tests cover env-mode, mount-mode, missing-file, missing-env, double-source, missing-source, relative mountPath, and the mixed case. Smoke test + daemon-parity check are in the test plan per project CLAUDE.md.

PR #54, opened 2026-04-23
feat: inherit Space spec.defaults.container into contained containers

### LINKED ISSUE #44: Space - `spec.defaults.container` inheritance for isolation envelope
## Context

Space exists to declare the isolation envelope once for everything inside it. Today every Container must repeat the posture fields itself, which is error-prone and misses the point of the Space-as-envelope design. This issue adds the inheritance mechanism.

Traceable to the proposal "Kukeon for AI Agents: A Proposal for Agent-Native Orchestration" §4.3.

## Problem

If every Container in a Space needs the same cap drops, the same user, the same resource ceilings, repeating that per-container in YAML is error-prone. More importantly, it misses the point of the Space-as-envelope design — the Space exists *because* isolation should be declared once.

## Proposal

Add a `spec.defaults` block on Space that is inherited by every Container in the Space unless overridden:

```yaml
apiVersion: v1beta1
kind: Space
metadata:
  name: agent-sandbox
spec:
  realmId: agents
  defaults:
    container:
      user: "1000:1000"
      readOnlyRootFilesystem: true
      capabilities:
        drop: ["ALL"]
      securityOpts: ["no-new-privileges"]
      resources:
        memoryLimitBytes: 4294967296   # 4 GiB
        pidsLimit: 512
      tmpfs:
        - { path: /tmp, sizeBytes: 268435456 }
```

Per-container values override Space defaults; Space defaults override kukeon built-ins. Inheritance is shallow (no deep-merge surprises).

## Acceptance criteria

- [ ] `spec.defaults.container` supports every field added by the Container security-fields issue
- [ ] Per-container explicit values override Space defaults
- [ ] Inheritance is shallow — overriding `capabilities` replaces it rather than merging with the Space value
- [ ] Effective (post-merge) config is observable per container (e.g. in status or via a `kuke get -o yaml` flag) so operators can see what actually applied
- [ ] Precedence is documented: container spec > Space defaults > kukeon built-in defaults

## Out of scope

- Realm-level defaults (could be added later if demand exists)
- Deep-merge semantics

## Dependencies

Container security fields must land first — you can't inherit fields the Container schema doesn't yet accept.

eminwux:
## Summary
- Add `spec.defaults.container` to the Space kind so the isolation envelope (user, readOnlyRootFilesystem, capabilities, securityOpts, tmpfs, resources) is declared once on the Space and inherited by every container that lives in it.
- Run a shallow merge at the single choke point (`applyCellContainerDefaults`, called from both `ensureCellContainers` and `createCellContainers`); precedence is **container spec > Space defaults > kukeon built-ins**. Inheriting a slice/pointer field is a replace, not a deep-merge — a container that sets `capabilities.drop` replaces the Space's `capabilities` entirely rather than layering.
- Persist the merged (effective) config so `kuke get container -o yaml` shows what actually applied — no separate `status.effectiveConfig` is needed. `DiffSpace` / `UpdateSpace` now track the defaults block so applying a modified defaults re-persists the metadata.

Closes #44

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/modelhub/... ./internal/apischeme/... ./internal/controller/...` — new `merge_test.go` covers fill-empty, container-wins, shallow-replace, idempotency, deep-copy; new scheme round-trip test covers the YAML tags.
- [x] `make test` — only pre-existing `registry.eminwux.com` expectations in `internal/ctr/container_utils_test.go` fail (reproduced on `origin/main`, unrelated).
- [x] Smoke test per CLAUDE.md: torn down + rebuilt kukeond image, `kuke init` succeeded, `kuke get realms` and `kuke get realms --no-daemon` returned identical output.
- [x] Applied a Space with every defaults field set and a Cell whose only container declares `image` + `command` — `kuke get container -o yaml` shows all Space defaults merged (user, readOnlyRootFilesystem, capabilities, securityOpts, tmpfs, resources).
- [x] Applied a second Cell whose container overrides `user` and sets `capabilities.drop: [CAP_NET_RAW]` — effective spec keeps the override, drops the Space default's `add`, and still inherits the unset fields (verifies precedence + shallow replace end-to-end).

### DIFF
--- a/docs/site/manifests/space.md
+++ b/docs/site/manifests/space.md
@@ -26,6 +26,45 @@ The realm that owns this space. Matches the realm's `metadata.name`.
 
 Directory where Kukeon writes this space's CNI conflist. Defaults to the system CNI config directory (`/etc/cni/net.d`). Override when you want per-space conflist isolation.
 
+### `spec.defaults.container` (object, optional)
+
+Default values inherited by every container created inside the space unless the container's own spec overrides the field. The space exists to declare the isolation envelope once — `spec.defaults.container` is how that envelope flows into every container.
+
+```yaml
+apiVersion: v1beta1
+kind: Space
+metadata:
+  name: agent-sandbox
+spec:
+  realmId: agents
+  defaults:
+    container:
+      user: "1000:1000"
+      readOnlyRootFilesystem: true
+      capabilities:
+        drop: ["ALL"]
+        add: ["NET_BIND_SERVICE"]
+      securityOpts: ["no-new-privileges"]
+      tmpfs:
+        - path: /tmp
+          sizeBytes: 268435456   # 256 MiB
+      resources:
+        memoryLimitBytes: 4294967296   # 4 GiB
+        pidsLimit: 512
+```
+
+Supported fields (each mirrors `ContainerSpec`): `user`, `readOnlyRootFilesystem`, `capabilities`, `securityOpts`, `tmpfs`, `resources`.
+
+**Precedence** (highest wins):
+
+1. Container `spec.*` — explicit per-container values
+2. Space `spec.defaults.container.*` — envelope defaults
+3. Kukeon built-in defaults — the runtime fallback (no user, capabilities as delivered by the image, etc.)
+
+**Shallow inheritance.** A container that sets a field replaces the space default for that field in full; nested slices and pointer structs are not deep-merged. For example, a container that declares `capabilities.drop: ["CAP_NET_RAW"]` replaces the space's `capabilities` entirely — the space's `add: [NET_BIND_SERVICE]` does **not** carry through. If you want layered changes, re-declare the full effective value on the container.
+
+**Effective config.** The merge runs at the point the container is created or updated, so the post-merge (effective) configuration is what gets persisted. `kuke get container <name> -o yaml` shows the merged values directly — no separate `status.effectiveConfig` block is needed.
+
 ## status
 
 | Field         | Type                           | Description                                            |
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -135,6 +135,7 @@ func ConvertSpaceDocToInternal(in ext.SpaceDoc) (intmodel.Space, error) {
 			Spec: intmodel.SpaceSpec{
 				RealmName:     in.Spec.RealmID,
 				CNIConfigPath: in.Spec.CNIConfigPath,
+				Defaults:      convertSpaceDefaultsToInternal(in.Spec.Defaults),
 			},
 			Status: intmodel.SpaceStatus{
 				State:      intState,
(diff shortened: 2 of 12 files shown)


eminwux:
**PR #54 Review — `feat: inherit Space spec.defaults.container into contained containers`**

Solid implementation. The single choke point in `applyCellContainerDefaults` called from both `ensureCellContainers` and `createCellContainers` is the right shape, the shallow-merge semantic is well-tested (fill-empty, container-wins, shallow-replace, idempotent, deep-copy), the scheme round-trip covers YAML tags, and persisting the merged (effective) config so `kuke get container -o yaml` reflects reality is the right call. No security-relevant leaks in errors or logs.

A few non-blocking items:

- **Precedence asymmetry for `readOnlyRootFilesystem`** (`internal/modelhub/merge.go:26-28`, `pkg/api/model/v1beta1/space.go:45-48` comment, `docs/site/manifests/space.md`): the stated precedence is "Container spec > Space defaults," but because `ContainerSpec.ReadOnlyRootFilesystem` is a plain `bool` (not `*bool`), a Container that explicitly sets `readOnlyRootFilesystem: false` cannot opt out of a Space default of `true` — the merge sees `false` as unset. The code comment on `SpaceContainerDefaults` acknowledges this, but `docs/site/manifests/space.md` describes the precedence rule without mentioning the exception. For a security-hardening field, silently upgrading an explicit `false` to `true` is surprising; worth adding a sentence to space.md so users expecting "container spec always wins" aren't caught out. (Arguably this is the *desirable* security behavior — the envelope can't be weakened — in which case calling it out explicitly is even more valuable.)

- **New failure mode on container reconciliation** (`internal/controller/runner/provision.go:364-377`): `applyCellContainerDefaults` calls `r.GetSpace(...)` in the hot path of both create and ensure. If the parent Space is transiently unreadable (filesystem hiccup, corrupt metadata), every container-touching operation now fails where it previously succeeded without applying defaults. In practice this should only bite pathological states, but it is a new precondition worth being aware of.

- **Redundant zero-check in slice clone** (`internal/modelhub/merge.go:58-64`, `:74-77`): `append([]string(nil), in.Drop...)` already returns nil when `in.Drop` is nil or empty, so the `if len(out.Drop) == 0 { out.Drop = nil }` follow-up is a no-op. Cosmetic — leave or trim, your call.

- **Effective-config freezes inherited values** (design note, not a code bug): because the merge is persisted as the effective spec, a Container created under `defaults.container.user: 1000:1000` stores `user: 1000:1000` in its own spec. If the Space default is later changed to `2000:2000`, the Container's persisted `1000:1000` now looks like an explicit override and the shallow-merge keeps it. The PR description already calls this out ("only new or updated containers pick up the new envelope") — flagging here because it's the kind of behavior that will generate a support question eventually and might merit a line in space.md next to the "Effective config" paragraph.

LGTM to proceed — all items are non-blocking. Test plan is complete (all boxes checked including the CLAUDE.md daemon-parity smoke test), CI is green on CodeQL.

PR #55, opened 2026-04-23
feat: add space egress policy with iptables enforcement

### LINKED ISSUE #45: Space - `spec.network.egress` policy (default-deny + host/CIDR allowlist)
## Context

Space creates a CNI bridge today but does not constrain what that bridge can reach. For any untrusted workload — and for agent workloads in particular — uncontrolled egress is the main way a small mistake becomes a big one (data exfiltration, paid-API abuse, command-and-control via prompt injection).

Traceable to the proposal "Kukeon for AI Agents: A Proposal for Agent-Native Orchestration" §4.4.

## Problem

The current Space creates a CNI bridge but does not constrain what the bridge can reach. For agent workloads, uncontrolled egress is the main way a small mistake becomes a big one (data exfiltration, calls to paid APIs, command-and-control via prompt injection).

## Proposal

Add `spec.network` to Space with an egress-policy shape:

```yaml
spec:
  realmId: agents
  network:
    egress:
      default: deny           # or allow
      allow:
        - host: api.anthropic.com
          ports: [443]
        - host: registry.npmjs.org
          ports: [443]
        - cidr: 10.0.0.0/8
          ports: [5432]
```

Enforcement can be iptables/nftables on the bridge for IP+port rules, or a transparent egress proxy for hostname rules (the proxy is the honest option for HTTPS since SNI can be inspected but IP allowlists are too coarse).

For the minimum viable version, accept `default: deny` plus an allowlist of `(host, port)` pairs, and document that hostname enforcement requires the proxy path.

## Acceptance criteria

- [ ] `spec.network.egress.default: deny` blocks all outbound traffic from the Space bridge
- [ ] `spec.network.egress.default: allow` (explicit) matches current behavior
- [ ] Allowlist entries with `cidr + ports` are enforced via iptables / nftables
- [ ] Allowlist entries with `host + ports` either (a) resolve to IPs at apply time and enforce via iptables with a documented TTL caveat, or (b) enforce via an egress proxy — the design decision is made in this issue
- [ ] Rejected traffic is observable (drop counter in status or daemon log)
- [ ] Documentation clearly states the hostname-enforcement limitation for the chosen approach

## Out of scope

- Ingress policy
- Per-container egress rules (Space is the boundary)
- L7 filtering beyond SNI

## Dependencies

None for the iptables/nftables path. The hostname path may choose to depend on a future egress-proxy primitive; that choice is part of this issue.

eminwux:
## Summary

- Extend `spec.network.egress` on Space with `default: allow|deny` + host/CIDR allowlist so untrusted workloads (notably agents) cannot exfiltrate to arbitrary destinations by default.
- Pick **design path (a)** for hostname rules from the issue: resolve to IPv4 at apply time and embed IPs into iptables. An egress proxy is left for a follow-up primitive.
- Per-space `KUKE-EGR-<hash>` chain dispatched from a shared `KUKEON-EGRESS` chain hooked into `FORWARD`, matched on the space bridge (`-i <bridge>`). RELATED/ESTABLISHED short-circuit preserves reply traffic.
- New `internal/netpolicy` package isolates pure rule generation from the iptables exec, with a test-hook runner. The runner `Exec` grows a `netpolicy.Enforcer` field that defaults to a `NoopEnforcer` when unset — existing runner tests need no changes.
- Observability (AC #5): terminal DROP is comment-tagged `kukeon:<realm>:<space>:default`; drop counter logged at space teardown. Operators can grep live state via `iptables -S | grep kukeon:`.

## Test plan

- [x] `make test` — all prior passing tests still green; pre-existing `internal/ctr` failures (unrelated default-image-registry mismatch) reproduce on `origin/main` and are out of scope
- [x] `go vet ./...`
- [x] `go build ./...` and `make kuke`
- [x] New unit tests exercise rule generation (`BuildRules`, `DispatchRule`), validation (bad CIDR/host/port, default=allow collapse), host resolver (v6 filtering, empty-result fatal, dedup), and the enforcer's iptables invocation sequence via a fake runner (apply/idempotent-apply/remove with dispatch cleanup)
- [x] apischeme round-trip covers the new `network.egress` field and confirms `network: null` is omitted when unset
- [ ] **Local smoke test from CLAUDE.md §5 (daemon-parity check)** — requires root/containerd; not run in this session. The next reviewer-triggered workflow or the maintainer should exercise `sudo ./kuke get realms` vs `--no-daemon` after re-applying a Space with an egress policy, and confirm `sudo iptables -S KUKEON-EGRESS` lists the dispatch rule and `sudo iptables -L KUKE-EGR-<hash> -n -v -x` shows the policy rules with counters

Closes #45

### DIFF
--- a/docs/site/manifests/space.md
+++ b/docs/site/manifests/space.md
@@ -9,6 +9,14 @@ metadata:
 spec:
   realmId: main
   cniConfigPath: /etc/cni/net.d
+  network:
+    egress:
+      default: deny
+      allow:
+        - host: api.anthropic.com
+          ports: [443]
+        - cidr: 10.0.0.0/8
+          ports: [5432]
 status:
   state: Ready
   cgroupPath: /kukeon/main/default
@@ -25,6 +33,31 @@ The realm that owns this space. Matches the realm's `metadata.name`.
 ### `spec.cniConfigPath` (string, optional)
 
 Directory where Kukeon writes this space's CNI conflist. Defaults to the system CNI config directory (`/etc/cni/net.d`). Override when you want per-space conflist isolation.
+
+### `spec.network.egress` (object, optional)
+
+Constrains outbound traffic leaving the space's bridge. When omitted, traffic is unconstrained — matching the pre-`v1beta1` behavior.
+
+| Field     | Type                              | Description                                                                 |
+|-----------|-----------------------------------|-----------------------------------------------------------------------------|
+| `default` | `allow` \| `deny`                 | Fallthrough action when no `allow` rule matches. Required when `egress` is set. |
+| `allow`   | list of allow rules               | Permitted destinations. Each entry sets exactly one of `host` or `cidr`.    |
+
+Each allow rule:
+
+| Field   | Type          | Description                                                                          |
+|---------|---------------|--------------------------------------------------------------------------------------|
+| `host`  | string        | DNS name. Resolved to IPv4 addresses by the daemon at apply time.                    |
+| `cidr`  | string        | IPv4 CIDR. Enforced literally via iptables.                                          |
+| `ports` | list of ints  | TCP destination ports (1–65535). Empty means "any protocol and any port on this destination". |
+
+**Enforcement.** When the policy is non-trivial (either `default: deny`, or `default: allow` with at least one allow rule), the daemon materializes it as a per-space chain in the iptables `filter` table named `KUKE-EGR-<hash>`, fed from a shared `KUKEON-EGRESS` chain hooked into `FORWARD`. Existing connections are preserved via a `RELATED,ESTABLISHED` short-circuit, so reply traffic for outbound flows initiated before a policy is applied is unaffected. The enforcement is bridge-scoped (`-i <bridge>`), so a single misconfigured space cannot affect traffic from other spaces in the same realm.
+
+**Hostname caveat (design path a).** Hostnames are resolved to IPs once, at apply time, and the resolved IPs are written into iptables. If the target service publishes a new IP after the policy is applied, the rule will not cover it until you re-apply the space (for example, `kuke apply -f space.yaml`). Services behind large, frequently rotating CDNs are not a good fit for hostname rules in this release — prefer a CIDR rule if you control the upstream, or accept occasional 5xxs and re-apply on a schedule. A transparent egress proxy would address this, and is tracked as a separate primitive (out of scope for this issue).
+
+**Observability.** When a space is deleted, the daemon logs the terminal DROP counter (packets + bytes) for its chain, tagged with the `kukeon:<realm>:<space>` comment. Operators can also inspect counters on a running policy at any time with `iptables -L KUKE-EGR-<hash> -n -v -x` or filter across all spaces with `iptables -S | grep kukeon:`.
+
+**Minimum-viable scope.** When a rule specifies `ports`, enforcement is TCP-only; UDP and ICMP fall through to `default` for that destination. When `ports` is omitted, the rule matches any IP traffic to the destination. IPv6 addresses returned by DNS are ignored — the rules are IPv4-only. These gaps are tracked as follow-ups.
 
 ### `spec.defaults.container` (object, optional)
 
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -135,6 +135,7 @@ func ConvertSpaceDocToInternal(in ext.SpaceDoc) (intmodel.Space, error) {
 			Spec: intmodel.SpaceSpec{
 				RealmName:     in.Spec.RealmID,
 				CNIConfigPath: in.Spec.CNIConfigPath,
+				Network:       convertSpaceNetworkToInternal(in.Spec.Network),
 				Defaults:      convertSpaceDefaultsToInternal(in.Spec.Defaults),
 			},
 			Status: intmodel.SpaceStatus{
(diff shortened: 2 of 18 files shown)


eminwux:
**PR #55 Review — `feat: add space egress policy with iptables enforcement`**

Clean implementation. The new `internal/netpolicy` package is cleanly split into three layers (pure rule generation → host resolution → iptables enforcement), all public surface is referenced, error sentinels live in `internal/errdefs`, and the `NoopEnforcer` default keeps tests and `--no-daemon` paths from mutating the host firewall. Good coverage of the failure modes in the unit tests (validation matrix, v6 filtering, fatal empty-resolve, idempotent dispatch, `RELATED,ESTABLISHED` short-circuit, terminal-DROP shape). No correctness bugs and no sensitive-data leaks spotted.

A few non-blocking follow-ups worth tracking — none are merge-blockers:

- **Non-atomic re-apply window** (`internal/netpolicy/enforcer.go:105`): `-F <chain>` then append leaves the per-space chain empty for the duration of N `-A` invocations. During that window, a packet that would normally hit the terminal `DROP` falls through the empty chain and returns to `FORWARD`. For a `default: deny` policy, that's a few-ms window where the policy is effectively "allow". Consider building the new rules into a shadow chain and swapping the dispatch jump atomically, or piping a batch through `iptables-restore --noflush`. Fine to defer.
- **No `--wait` flag on iptables calls** (`internal/netpolicy/enforcer.go:65`): concurrent space provisioning, or coexistence with firewalld / docker, can trip the xtables lock and surface as a user-facing apply failure. Adding `--wait 2` (or similar) to `execRunner.Run` is cheap and makes this resilient.
- **DNS resolution deadline** (`internal/controller/runner/egress.go:53`): `policy.Resolve(r.ctx, nil)` inherits the runner's context. If that context has no deadline, a stalled resolver can block space provisioning indefinitely (net's per-query defaults still apply, but there's no overall ceiling). A bounded `context.WithTimeout` around `Resolve` would make worst-case provisioning time predictable.
- **Daemon-parity smoke test unchecked in the test plan**: acknowledged by the author as requiring root/containerd and flagged explicitly — the maintainer should run `sudo ./kuke get realms` vs `--no-daemon` plus `sudo iptables -S KUKEON-EGRESS` / `iptables -L KUKE-EGR-<hash> -n -v -x` against an applied policy before merging, per `CLAUDE.md` §5.

LGTM to proceed — the items above are all follow-ups, not blockers.

eminwux:

eminwux:
Addressed in 5ecf694:
- Rebased onto `origin/main` (past #54's `spec.defaults.container`); conflicts in `pkg/api/model/v1beta1/space.go`, `internal/modelhub/space.go`, `internal/apischeme/scheme.go`, and `docs/site/manifests/space.md` were additive — both `spec.network.egress` and `spec.defaults.container` now sit side-by-side on `SpaceSpec`.
- Re-ran `make test` / `go vet ./...` / `go build ./...` / `make kuke` post-rebase — all green except the same pre-existing `internal/ctr` default-image-registry failures, which I re-verified reproduce on `origin/main` via a throwaway worktree.

Deferred (per reviewer "non-blockers"):
- Non-atomic re-apply window on `-F` + `-A` — tracking as follow-up; the shadow-chain / `iptables-restore --noflush` swap is a non-trivial refactor.
- `--wait` flag on iptables calls — follow-up.
- Bounded `context.WithTimeout` around `netpolicy.Resolve` — follow-up.
- Daemon-parity smoke test from CLAUDE.md §5 — still requires root + containerd, not run in this session.

eminwux:
**PR #55 Review — `feat: add space egress policy with iptables enforcement`**

Solid MVP. Good separation of pure rule generation (`BuildRules`) from I/O (`Enforcer`), deterministic realm+space-scoped chain names, `RELATED,ESTABLISHED` short-circuit, bridge-scoped dispatch, drop-counter logged at teardown, and comprehensive tests across validation, resolver, rule generation, and iptables invocation sequence. All new `ErrEgress*` sentinels are actually referenced; no dead code or sensitive-data leaks.

A few small, non-blocking items:

- **Flush-then-rebuild window** (`internal/netpolicy/enforcer.go:105`): `Apply` does `-F <chain>` then re-adds rules. During that window the per-space chain is empty and packets fall through to `FORWARD` (typically `ACCEPT`). Acceptable for MVP; worth documenting as a known gap if you want strict convergence semantics later.
- **No `iptables -w`** (`internal/netpolicy/enforcer.go:41`): concurrent `Apply`/`Remove` (e.g., two parallel `kuke apply`s, or a user running `iptables` alongside) can fail on xtables lock contention. Consider adding `-w` to the `execRunner` invocation as a follow-up.
- **`SpaceNetwork{Egress: nil}` YAML omit-empty** (`pkg/api/model/v1beta1/space.go:33`): `yaml:"network,omitempty"` on a pointer checks nil-pointer, not empty struct, so a `Network` with a nil `Egress` would render as `network: {}`. The forward conversion only sets `Network` when `Egress` is non-nil, so the edge case is unlikely in practice, but `TestSpaceEgressNilFieldsOmittedInYAML` only covers the nil-pointer case.
- **Partial-provision cleanup** (`internal/controller/runner/provision.go:164`): `applySpaceEgressPolicy` runs after CNI config + cgroup are created. If `iptables` fails, those resources are left behind until the next `kuke apply` re-runs `ensureSpaceCNIConfig` and re-applies. Not a bug — idempotent paths recover — but worth keeping in mind.
- **CLAUDE.md §5 smoke test unchecked**: the specific regression the bind-mount parity check catches doesn't apply here (this PR doesn't touch `/opt/kukeon` I/O paths), but an end-to-end run through a live daemon with a `default: deny` policy would be valuable before the next release cut — particularly confirming `iptables -S KUKEON-EGRESS` shows the dispatch rule and `iptables -L KUKE-EGR-<hash> -n -v -x` shows counters incrementing. Non-blocking given the unit-test coverage.

LGTM.

PR #58, opened 2026-04-24
fix: align test image references with test-side eminwux rule

### LINKED ISSUE #56: internal/ctr: default root container image tests expect registry.eminwux.com but code emits docker.io/library
## Repro

On `origin/main` @ `0895056` (also reproduces on feat branches rebased to main):

```bash
git worktree add /tmp/kukeon-main origin/main
cd /tmp/kukeon-main
go test ./internal/ctr/...
```

## Observed

```
--- FAIL: TestDefaultRootContainerSpec (0.00s)
    --- FAIL: TestDefaultRootContainerSpec/all_parameters_provided (0.00s)
        container_utils_test.go:92: Image = "docker.io/library/busybox:latest", want "registry.eminwux.com/busybox:latest"
    --- FAIL: TestDefaultRootContainerSpec/empty_parameters (0.00s)
        container_utils_test.go:92: Image = "docker.io/library/busybox:latest", want "registry.eminwux.com/busybox:latest"
--- FAIL: TestBuildRootContainerSpec (0.00s)
    --- FAIL: TestBuildRootContainerSpec/fallback_to_ID_when_containerd_ID_is_empty (0.00s)
        container_utils_test.go:214: Image = "docker.io/library/busybox:latest", want "registry.eminwux.com/busybox:latest"
FAIL	github.com/eminwux/kukeon/internal/ctr	70.042s
```

## Root cause

The package-level default and the test expectations disagree:

- `internal/ctr/container_utils.go:28`: `DefaultRootContainerImage = "docker.io/library/busybox:latest"`
- `internal/ctr/container_utils_test.go:49, :62, :154`: `wantImage: "registry.eminwux.com/busybox:latest"`

One of the two was updated without the other. `make test` has been failing on `main` as a result — multiple recent feature PRs (e.g. #55) have been forced to explicitly call out this failure as pre-existing, which masks the signal CI is supposed to give.

## Suggested fix

Align the two. `docker.io/library/busybox:latest` is the public-registry default a community contributor would expect, so updating the three `wantImage` literals in `container_utils_test.go` to match `DefaultRootContainerImage` is the less-surprising direction. If the intent was to host the image on the private registry, instead change `DefaultRootContainerImage` — but then `kuke init` without credentials would regress, so the test-side fix is preferred unless there's an out-of-band reason to keep the private registry default.

eminwux:
## Summary
- `make test` was failing on `main`: `TestDefaultRootContainerSpec` and `TestBuildRootContainerSpec` expected `registry.eminwux.com/busybox:latest` but `DefaultRootContainerImage` emits `docker.io/library/busybox:latest`. Root cause was the two sides drifting.
- Project rule: **test-side image references use `registry.eminwux.com` (publicly readable, no Docker Hub anonymous-pull throttling); code-side defaults / examples / docs use `docker.io`.** Keeping the code-side default on Docker Hub is correct — that's what a fresh contributor expects. The tests are the wrong side of the rule.
- `container_utils_test.go`: 3 `wantImage` literals now reference `ctr.DefaultRootContainerImage` instead of hardcoding a registry — the test validates the constant's plumbing, so it shouldn't pick a registry. (Option picked over restoring `registry.eminwux.com/...` literals because these unit tests are asserting what the code-side constant is, not what gets pulled — a constant-reference is more honest than either literal.)
- Swept the repo for other drift: `spec_security_test.go` (8 `Image:` fields in test fixtures) and `cmd/config/autocomplete_test.go` (1 test fixture) were holding `docker.io/library/...` on the test side — flipped to `registry.eminwux.com/...`.
- Left alone: `TestNormalizeImageReference` and `cmd/kuke/create/container/container_test.go` tests still use `docker.io/...` literals because they assert the `NormalizeImageReference` function's output contract, not a pull target. Docs, examples, CLI defaults, and `CLAUDE.md` all keep `docker.io/...`.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (was failing on `main`, now green)
- [x] `go test ./internal/ctr/... ./cmd/config/...` (touched packages)
- No build-path, daemon, or `/opt/kukeon` touch → smoke test (`kuke init` / daemon-parity check) not required.

Closes #56


eminwux:
**PR #58 Review — `fix: align test image references with test-side eminwux rule`**

Minimal, correct fix. `DefaultRootContainerImage` is `docker.io/library/busybox:latest` (`internal/ctr/container_utils.go:28`) but three `container_utils_test.go` cases hardcoded `registry.eminwux.com/busybox:latest` — replacing with `ctr.DefaultRootContainerImage` removes the drift surface entirely. Sweep of the remaining test-side `docker.io/library/...` strings (`spec_security_test.go` x8, `cmd/config/autocomplete_test.go` x1) is consistent with the stated project rule. `TestNormalizeImageReference` and `cmd/kuke/create/container/container_test.go:132` are correctly left alone — they assert the `NormalizeImageReference` contract, which is code-side. No logic changes, no dead symbols, no sensitive data. All CI checks green.

One non-blocking nit:

- **Test strength** (`internal/ctr/container_utils_test.go:49,62,154`): using `ctr.DefaultRootContainerImage` as the `wantImage` makes the assertion `spec.Image == ctr.DefaultRootContainerImage`, which is essentially "function forwards the constant" rather than "function returns a specific value". You already called out the tradeoff in the PR description, so I'd leave it — the previous literal was the exact thing that rotted, and this shape won't rot again. Just flagging that the test's claim is narrower than it was.

LGTM.

Issue #67, opened 2026-04-25
sbsh binary cache and multi-arch resolver


eminwux:
## Context

#57 ships OCI-spec injection that bind-mounts a sbsh binary from `/opt/kukeon/cache/sbsh/<arch>/sbsh` into the container. PR 1 stubs that cache — the operator manually drops a host-arch sbsh binary into the cache directory. This issue replaces the stub with a real fetcher that resolves and caches per-image-arch sbsh from the upstream release.

## Proposal

- On image pull (or attachable-container create), kukeond reads `Platform.Architecture` (added in #57) and ensures `/opt/kukeon/cache/sbsh/<arch>/sbsh` is populated for that arch.
- Resolution strategy: download the per-arch asset from sbsh's GitHub release. Verify checksum.
- First-use download into the cache; subsequent containers reuse the cached binary.
- Cache eviction is out of scope (manual `rm` is acceptable until a need surfaces).

## Acceptance criteria

- [ ] On a host with an empty `/opt/kukeon/cache/sbsh/`, creating an attachable container with an amd64 image populates `/opt/kukeon/cache/sbsh/amd64/sbsh` and the container starts.
- [ ] Same flow for an arm64 image on an arm64 host populates `/opt/kukeon/cache/sbsh/arm64/sbsh`.
- [ ] Cross-arch (amd64 host + arm64 image, or vice versa) is documented as supported or unsupported with a clear error message at create time. First cut may document as unsupported.
- [ ] Checksum-verification failure surfaces as a `kuke create` error, not a runtime crash.
- [ ] Supported host/image arch matrix is documented in the README or `docs/`.

## Notes

- Depends on #57 (image-arch detection + cache directory layout + injection consuming the cache).
- Depends on sbsh [#120](https://github.com/eminwux/sbsh/issues/120) shipping per-arch release assets.
- Independent of the `kuke attach` follow-up — this and that can land in either order after #57.
- Splits out of #57's original 4-PR "Suggested slicing" plan.

eminwux:
Sbsh will be replaced by kubetty. Closing as Not Planned

Issue #66, opened 2026-04-25
kuke attach: subcommand and daemon validation (phase 1)


eminwux:
## Context

#57 ships the OCI-spec injection that turns `Attachable: true` containers into sbsh-wrapped workloads with a control socket bind-mounted to `/opt/kukeon/<r>/<s>/<st>/<c>/<co>/sbsh.io` on the host. This issue lands the verb that consumes that socket: `kuke attach`.

Bytes never traverse kukeond's RPC. Phase 1 keeps the client minimal: `kuke attach` resolves the host socket path and hands the TTY off to the host `sb` binary via `syscall.Exec`, so TTY and signal handling fall through to sbsh. There is no in-process sbsh client TTY loop in this PR — that's deferred until sbsh ships a stable Go-module client surface, and the deferral keeps phase 1 within one reviewable PR.

## Proposal

New subcommand under `cmd/kuke/attach/`:

```
kuke attach --realm <r> --space <s> --stack <st> --cell <c> [--container <co>]
```

- Resolves the per-container host socket path via the helper added in #57.
- Daemon-side `AttachContainer` upgraded from placeholder: validates the target container is `Attachable: true`, returns `HostSocketPath` in `AttachContainerResult`, and no longer returns `ErrAttachNotImplemented`.
- If `--container` is omitted, picks the cell's only non-root `Attachable: true` container. Multiple candidates → error with the candidate list and non-zero exit. Zero candidates → error. The root container is excluded from the auto-pick set.
- If the daemon-side check reports `Attachable: false` on the target, the daemon fails the call before any socket is opened, and the client surfaces the error (extends the existing rpc test pattern).
- Client invocation: `syscall.Exec` into the host `sb attach --socket <path>` binary. Binary path is configurable, defaulting to `sb` on `PATH`. TTY, signals, and detach keystroke handling are delegated to `sb`.

## Acceptance criteria

- [ ] `kuke attach` against a single-non-root-attachable cell succeeds without `--container` (manual smoke against a shell workload, recorded in PR description).
- [ ] With multiple non-root attachable containers, omitting `--container` errors with the candidate list and exits non-zero (unit test for the auto-pick logic).
- [ ] Targeting a container with `Attachable: false` is rejected by the daemon and reported by the client (rpc test, mirroring the existing pattern).
- [ ] Auto-pick logic is unit-tested for: zero candidates, one candidate, multiple candidates, root-only.
[...]

PR #69, opened 2026-04-25
feat: inject sbsh wrapper into Attachable container OCI specs

### LINKED ISSUE #57: Attachable containers: OCI-spec injection (sbsh wrapper + host-visible socket)
## Motivation

Some workloads' processes must themselves be driven through a PTY: REPLs, TUIs, interactive debuggers, agents like `claude`. Under the PTY-ownership constraint, a process that did not inherit the PTY at birth cannot be retroactively attached. These workloads must be sbsh's child from fork time — sbsh has to be the parent at fork, ahead of the user's entrypoint.

This issue ships **the foundation slice** of the attachable-containers feature: a new `ContainerSpec.Attachable` field plus the OCI-spec injection that wraps the entrypoint with sbsh and bind-mounts the sbsh control socket so it's reachable from both inside the container and on the host. The `kuke attach` verb (#66), the binary cache + multi-arch resolver (#67), and the `claude` e2e variant (#68) are split into their own issues so this issue fits one reviewable PR.

## Design — full picture (for context)

Full feature: when `Attachable: true`, kukeond injects a static sbsh binary, bind-mounts a control socket as `/run/sbsh.socket` inside the container ↔ `/opt/kukeon/<r>/<s>/<st>/<c>/<co>/sbsh.io` on the host, and wraps `process.args` with `sbsh terminal --socket /run/sbsh.socket -- <orig entrypoint>`. `kuke attach` (#66) connects to the host-visible socket as a thin sbsh client. Bytes never traverse kukeond's RPC. PTY ownership is satisfied by construction (sbsh is parent-at-fork-time of the workload).

This collapses #39 and #65: their RPC-streaming / host-side-nsenter approaches are subsumed by entrypoint wrapping + dual-bound socket.

## What this issue ships (single PR)

1. **`ContainerSpec.Attachable bool`** on `pkg/api/model/v1beta1/container.go` and `internal/modelhub/container.go`. Default `false` — no behavior change for existing specs.
2. **OCI-spec injection in `BuildContainerSpec`** (`internal/ctr/spec.go`):
   - Bind-mount the sbsh binary read-only at `/.kukeon/bin/sbsh`.
   - Bind-mount the per-container Unix socket path at `/run/sbsh.socket` ↔ host `/opt/kukeon/<realm>/<space>/<stack>/<cell>/<container>/sbsh.io`.
   - Wrap `process.args` with `sbsh terminal --socket /run/sbsh.socket -- <orig entrypoint> <orig cmd / args>`.
3. **Image-arch detection** in `internal/ctr/image.go` — read `Platform.Architecture` post-pull so the cache lookup keys off image arch, not host arch.
4. **Per-container socket-path helper** in `internal/util/fs/metadata.go`, alongside the existing realm/space/stack/cell helpers, used as the single source of truth for both the OCI bind-mount source and the future host-visible attach path.
5. **Stub binary cache.** A single host-arch sbsh binary placed under `/opt/kukeon/cache/sbsh/<arch>/sbsh` is enough to exercise the spec end-to-end. The real fetcher/resolver lands in #67. Document the manual-population step in the PR description.
6. **Daemon-side deny check.** When `Attachable: false`, the daemon rejects future attach requests at the API boundary, ahead of any client. The full `kuke attach` client lands in #66.

The user's rootfs is untouched. Works with read-only rootfs images. Works for any base image because the injected sbsh is fully static.

## Tricky bits (in scope for this PR)

1. **ENTRYPOINT/CMD composition.** Respect OCI semantics when wrapping. Test the matrix: image-only ENTRYPOINT, image-only CMD, both, user override of either. Kubernetes got this subtly wrong for years.
2. **Reserved in-container paths.** `/.kukeon/bin/sbsh` and `/run/sbsh.socket`. Document the claim; make at least the binary path configurable if a real collision shows up.
3. **Socket security threat model.** `/run/sbsh.socket` is a path to the workload's PTY for any process inside the container that can open it; the host-visible peer is a path in for any host UID that can reach `/opt/kukeon/...`. Per-container directory `0700` owned by root; socket inherits. Document.
4. **`/dev/pts` inside the container.** Must be present and correctly wired for sbsh to allocate a PTY. Standard in containerd setups; verify.
5. **Attachable orthogonality.** `Attachable` is independent of `CellSpec.RootContainerID` / `ContainerSpec.Root`. Any container in a cell can be `Attachable: true`.

## Acceptance

- [ ] `ContainerSpec.Attachable` lands on the v1beta1 API and modelhub. Default `false`.
- [ ] With `Attachable: false`, the assembled OCI spec is byte-identical to the pre-#57 spec for the same input. Unit-tested.
- [ ] With `Attachable: true`, the assembled OCI spec contains: RO bind mount of sbsh binary at `/.kukeon/bin/sbsh`; bind mount of the socket-bearing path at `/run/sbsh.socket` ↔ host `/opt/kukeon/<r>/<s>/<st>/<c>/<co>/sbsh.io`; `process.args` wrapped with `sbsh terminal --socket /run/sbsh.socket --` prepended to the original entrypoint+cmd. Unit-tested across the ENTRYPOINT/CMD composition matrix.
- [ ] **Same-socket inode test.** When an attachable container is running, the inode of `/run/sbsh.socket` (inside the container) matches the inode of `/opt/kukeon/<r>/<s>/<st>/<c>/<co>/sbsh.io` (on the host).
- [ ] Image-arch is read post-pull and used as the cache key. Unit test for the resolution helper.
- [ ] Per-container socket-path helper exists in `internal/util/fs/metadata.go` and is the single source of truth for both the OCI bind-mount source and the host-visible attach path that #66 will consume.
- [ ] Daemon-side attach RPC (or its placeholder endpoint) rejects attach requests when the target container has `Attachable: false`. The full client lands in #66.
- [ ] Daemon-parity check (`kuke get realms` vs `kuke get realms --no-daemon`) still passes.
- [ ] PR description documents how to populate the stub cache (`/opt/kukeon/cache/sbsh/<arch>/sbsh`) for local testing until #67 lands.

## Non-goals (in this PR)

- `kuke attach` subcommand → #66.
- sbsh binary auto-fetch and multi-arch resolver → #67.
- E2E `claude` variant → #68.
- Streaming PTY through kukeond's RPC.
- Cross-cell terminal sharing.
- Image rewriting at pull time.
- Custom containerd shim.

## Composition with Session (#46)

Session wraps a Stack whose main container is `Attachable: true` as the canonical agent-run pattern. Not a blocker in either direction; design integration point for when both land.

## Dependencies

The 2026-04-24 dev-agent assessment on this issue confirmed none of the foundation that the original body claimed had landed in `main` (no sbsh import, no `kuke attach`, no StateRoot helper, no image-arch detection, no `Attachable` field). Reusable starting points already on `main`:

- `BuildContainerSpec` in `internal/ctr/spec.go` — graft point for bind mounts and `process.args` wrap.
- The secret bind-mount pattern in `internal/ctr/secrets.go` (and consumer in `internal/controller/runner/start.go`) — same shape for the RO sbsh binary mount and the per-container socket path bind.
- `internal/util/fs/metadata.go` — extend with the per-container socket path helper.

External: **sbsh [#120](https://github.com/eminwux/sbsh/issues/120)** for a static multi-arch sbsh release. PR 1 ships with a manually-placed host-arch binary; the multi-arch resolver in #67 consumes the upstream release.

## Follow-ups (split out of this issue)

- **#66** — `kuke attach`: thin sbsh client subcommand. Depends on this issue.
- **#67** — sbsh binary cache and multi-arch resolver. Depends on this issue. Independent of #66.
- **#68** — E2E `claude` workload under sbsh wrapper. Depends on #57, #66, and #67.

eminwux:
## Summary
- Foundation slice for the attachable-containers feature: opt-in `ContainerSpec.Attachable` field plus the OCI-spec wrapping that injects the static sbsh binary, bind-mounts a per-container control socket, and prepends `sbsh terminal --socket /run/sbsh.socket --` to `process.args`. With `Attachable: false` the assembled spec is byte-identical to the pre-#57 spec, so existing workloads are untouched.
- Adds the per-container socket helper in `internal/util/fs/metadata.go` so both the OCI bind-mount source and the host-visible attach path that #66 will consume share a single source of truth (`/opt/kukeon/<r>/<s>/<st>/<c>/<co>/sbsh.io`).
- Adds the daemon-side placeholder `KukeonV1.AttachContainer` RPC: it looks up the target container and rejects the request with `ErrAttachNotSupported` when `Attachable=false`. When the target is attachable, the placeholder returns `ErrAttachNotImplemented` so the future `kuke attach` client (#66) can branch on it; this enforces the gate at the API boundary today, ahead of any client.
- Adds the post-pull image-arch resolver (`SbshCachePath` + `imageArchitecture`) so the cache lookup keys off the *image* arch, not host arch — the regression the issue calls out.

## Stub cache (until #67 lands)
Until the multi-arch resolver in #67 ships, the daemon expects a single host-arch sbsh binary placed manually at:

```
/opt/kukeon/cache/sbsh/<arch>/sbsh
```

For local testing on amd64:

```bash
sudo install -D -m 0755 /path/to/sbsh /opt/kukeon/cache/sbsh/amd64/sbsh
```

## Test plan
- [x] `make test` (full unit-test suite, includes new ctr/attachable, ctr/sbsh_cache, util/fs, apischeme round-trip, and daemon attach-RPC tests)
- [x] `go vet ./...`
- [x] `go build ./...` and `make kuke`
- [x] Daemon-parity check: `sudo ./kuke get realms` vs `sudo ./kuke get realms --no-daemon` — identical output

Closes #57


eminwux:
**PR #69 Review — `feat: inject sbsh wrapper into Attachable container OCI specs`**

Foundation slice for the attachable-container feature is well-shaped: the opt-in `Attachable` field, the per-container socket helper, the OCI args-wrap (with table-driven coverage of image-only ENTRYPOINT/CMD vs user overrides), and the daemon-side `AttachContainer` placeholder that gates on `Attachable=true` and surfaces distinct sentinels. No sensitive-data leaks; conventions (Apache + SPDX headers, colocated tests, `errdefs` sentinels) are clean.

One blocking item:

- **Dead code — unwired resolver** (`internal/ctr/sbsh_cache.go:59` and `:74`): `ResolveSbshCachePath` and `imageArchitecture` have **zero callers** anywhere in the tree (`git grep ResolveSbshCachePath` / `imageArchitecture` returns only the definition site). The runner injects the binary path via `ctr.SbshCachePath(r.opts.RunPath, runtime.GOARCH)` in `internal/controller/runner/attachable.go:198` — i.e., host arch, not image arch. The PR description claims the resolver is added "so the cache lookup keys off the *image* arch, not host arch — the regression the issue calls out", but the regression guard isn't actually on the call path. Two reasonable options: (a) wire `ResolveSbshCachePath` into `attachableBuildOpts` in this PR (the runner already has the image, since it's pulled before container creation), or (b) drop both functions and let #67 add them alongside the actual caller. Option (b) keeps this PR strictly to the "foundation slice" the comment in `attachable.go` describes.

Non-blocking:

- **gofmt drift** (`pkg/api/kukeonv1/client.go:150-156`): adding `MethodAttachContainer` widened the longest identifier in this group, but the `MethodStop*` and `MethodKill*` rows still use the pre-PR alignment — `=` columns no longer line up across the group. `gofmt -w` on the file fixes it. (Not caught by CI — there's no `gofmt -l` check in the workflows.)
- **Test plan**: the daemon-parity check is checked off, which is the regression guard CLAUDE.md asks for on daemon-touching PRs — good.

LGTM once the dead resolver is either wired in or moved to #67 — the rest is non-blocking.

eminwux:
Option A - wire ResolveSbshCachePath into attachableBuildOpts in this PR

eminwux:
Addressed in 1a5a4a2:
- Wired `ResolveSbshCachePath` into the call path (Option A). The runner now calls it via the `Client` interface; the resolver pulls/loads the image, reads `cfg.Architecture` from the OCI image config, and returns `<runPath>/cache/sbsh/<image-arch>/sbsh`. The `runtime.GOARCH` lookup in `attachable.go` is gone, so the regression guard the PR description claimed is now actually on the path.
- Implementation note (small departure from the review's wording): I reshaped `ResolveSbshCachePath` to take `imageRef string` rather than `containerd.Image` and pull-or-get internally — the runner doesn't pre-pull before `CreateContainerFromSpec` (the pull lives inside `CreateContainer`), so handing the runner a `containerd.Image` would have required either exposing `pullImage` separately or reordering the call sites. The single-call form keeps `attachableBuildOpts` thin. The second `pullImage` inside `CreateContainer` is a `GetImage` cache hit, no extra round-trip.
- Doc fix on `imageArchitecture`: the prior comment described a "fall back to runtime.GOARCH" branch that was never implemented; replaced with the actual behavior (returns an error so silent host/image arch mismatches still surface).
- gofmt drift in `pkg/api/kukeonv1/client.go` around `MethodAttachContainer` — fixed.

Test plan re-run locally: `make test`, `go vet ./...`, `go build ./...`, `make kuke`, full daemon teardown + rebuild + `kuke init` + `kuke get realms` vs `kuke get realms --no-daemon` (identical output).

eminwux:
**PR #69 Re-review — `1a5a4a2`**

Both blocking items from the prior review are resolved:

- **Resolver wired** (`internal/controller/runner/attachable.go:56-59`): `attachableBuildOpts` now calls `r.ctrClient.ResolveSbshCachePath(spec.Image, r.opts.RunPath)` instead of the host-arch shortcut. The signature was reshaped to take an `imageRef` and pull internally so the call sits cleanly behind the `Client` interface (`internal/ctr/client.go:85`). Bonus: `imageArchitecture` lost its `runtime.GOARCH` fallback, so a malformed image now surfaces an error rather than silently picking the wrong binary — strictly better than my "wire it as-is" suggestion.
- **gofmt drift** (`pkg/api/kukeonv1/client.go:150-156`): `=` columns now align across the `MethodStop*`/`MethodKill*` rows.

LGTM.

PR #70, opened 2026-04-25
feat: add kuke attach thin sbsh client subcommand

eminwux:
## Summary

Phase-1 slice of #66 — the `kuke attach` thin client and the daemon-side socket-path resolution it needs. Filed under [#66 (resize-issue)](https://github.com/eminwux/kukeon/issues/66) because the issue's E2E + sbsh-Go-module ACs make it too broad for a single PR; PM to file phase-2 from the resize comment.

- Promotes the daemon's `AttachContainer` placeholder from `ErrAttachNotImplemented` to a real endpoint that validates the Attachable gate and returns `HostSocketPath` (resolved via `fs.ContainerSocketPath` against `controller.Exec.RunPath()`).
- Adds `cmd/kuke/attach/`: a top-level cobra subcommand that resolves the candidate container (auto-pick when one non-root attachable exists, error with sorted list when ambiguous, exclude root from the auto-pick set), then `syscall.Exec`s the on-host `sb` client against the resolved socket — bytes never traverse kukeond's RPC.
- Drops the now-unused `ErrAttachNotImplemented` sentinel and its errmap entry; adds `ErrAttachAmbiguous` and `ErrAttachNoCandidate` for the picker.

## Deferred to phase-2 (per resize-issue comment on #66)

- E2E under `e2e/` covering attach / detach / reattach (needs new PTY harness, attachable shell-workload image, scripted detach-keystroke driver — none exist in repo today).
- Optionally swap `syscall.Exec` for `sbsh/pkg/spawn.NewClient` once managed lifecycle is wanted. The issue body's claim that sbsh exposes a Go-module client TTY loop didn't match upstream — sbsh#120 is a server-side PID-1 change, not a client export — so the thin-client design is exec-based.

## Test plan

- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] `make test` — all unit packages pass, including new `cmd/kuke/attach/` suite (single-attachable auto-pick, ambiguous list, no-candidate, root-excluded-from-auto-pick, explicit `--container` not-attachable surfaces sentinel, missing `--realm/--space/--stack/--cell` checks, sb-binary absolute vs PATH-lookup) and the updated `internal/daemon/attach_rpc_test.go` (Attachable=true now returns `HostSocketPath` over the wire; Attachable=false still surfaces `ErrAttachNotSupported`).
- [x] Daemon-parity check from CLAUDE.md after rebuild + `kuke init` against the new image: `kuke get realms` and `kuke get realms --no-daemon` produce identical output.
- [x] `kuke attach --realm default --space default --stack default --cell ghost --container nope` returns the same `cell not found` error via daemon and `--no-daemon`.
- [ ] Phase-2 E2E (deferred — see resize-issue comment on #66).

`make test-e2e` has pre-existing socket/permission failures on `origin/main` (verified via worktree on `2339792`); not introduced by this change.

Refs #66

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -336,6 +336,20 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_START_CONTAINER_CELL = DefineKV("KUKE_START_CONTAINER_CELL", "kuke/start/container/cell")
 
+	// Attach command variables
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_ATTACH_REALM = DefineKV("KUKE_ATTACH_REALM", "kuke/attach/realm", "default")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_ATTACH_SPACE = DefineKV("KUKE_ATTACH_SPACE", "kuke/attach/space", "default")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_ATTACH_STACK = DefineKV("KUKE_ATTACH_STACK", "kuke/attach/stack", "default")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_ATTACH_CELL = DefineKV("KUKE_ATTACH_CELL", "kuke/attach/cell")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_ATTACH_CONTAINER = DefineKV("KUKE_ATTACH_CONTAINER", "kuke/attach/container")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_ATTACH_SB_BINARY = DefineKV("KUKE_ATTACH_SB_BINARY", "kuke/attach/sbBinary", "sb")
+
 	// Stop command variables
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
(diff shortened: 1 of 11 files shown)


eminwux:
**PR #70 Review — `feat: add kuke attach thin sbsh client subcommand`**

Phase-1 slice lands cleanly: daemon resolves and returns `HostSocketPath` behind the Attachable gate; thin client validates flags, picks the candidate (root-excluded), and `syscall.Exec`s `sb` directly so bytes never traverse RPC. Tests cover the full picker matrix, RPC round-trip, missing-flag guards, and the `sb`-binary lookup branches. License headers + SPDX present, sentinels added in `internal/errdefs/`, errmap entry for the deleted `ErrAttachNotImplemented` cleaned up, and the daemon-parity check from CLAUDE.md is ticked in the test plan. No sensitive data in error messages — the candidate-list output is operator-intended.

- **Error wrapping consistency** (`cmd/kuke/attach/attach.go:166-168`): `ErrAttachNotSupported` is preserved with `%w`, but the `ErrContainerNotFound` branch drops the sentinel — `return fmt.Errorf("container %q not found in cell %q", container, cell)`. Anything downstream doing `errors.Is(err, ErrContainerNotFound)` would silently lose the match. Cheap to fix: add `: %w` and `err`.
- **Comment vs. code on sb-binary lookup** (`cmd/kuke/attach/attach.go:221-224`): the doc says "If the caller supplied an absolute path it is used verbatim", but the actual check is `strings.ContainsRune(name, os.PathSeparator)` — a relative path like `./bin/sb` also bypasses `LookPath`. The behavior matches `exec.LookPath` semantics and is fine; just tighten the comment to "if the caller supplied a path with a separator" so it doesn't mislead.

LGTM to proceed — both items are non-blocking nits.

PR #75, opened 2026-04-25
test: add e2e PTY harness and attach/detach scenario

eminwux:
## Summary
- Adds a reusable PTY harness under `e2e/` (`startPTY`, `Write`, `Wait`, `Close`) so e2e tests can drive interactive subprocesses like `kuke attach`. Backed by `github.com/creack/pty` (test-side dep only).
- Adds `attachable-cell.yaml` testdata fixture: one root and one `attachable: true` container, both `registry.eminwux.com/busybox` per the test-side image rule.
- Adds `TestKuke_AttachDetach_KeepsTaskRunning`: PTY-driven `kuke attach`, sends sbsh's documented detach keystroke (Ctrl+] Ctrl+]), asserts the attach process exits cleanly, and asserts the underlying container task is still RUNNING via `ctr -n <realm> task ls`.
- The scenario depends on the per-container `tty/` directory bind mount (landed in `main` as #77). It does **not** pre-create any placeholder socket file (the workaround that #77 superseded). Skips if `sb` or `sbsh` is missing on PATH, or if the host `sbsh` predates sbsh#153 (probed via `sbsh terminal --help` for `--capture-file`).

## Test plan
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `go test -c ./e2e` (compile-check)
- [x] `make test` (unit suite, e2e excluded)
- [x] `make release-build` (CI build target)
- [x] Daemon-parity: `sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` returned identical output (default + kuke-system, both Ready).
[...]

PR #77, opened 2026-04-25
fix: bind-mount per-container tty dir for host-visible sbsh socket

### LINKED ISSUE #74: attachable: bind-mount per-container tty dir, not single socket file
## Context

Phase 1 (#69) injected the sbsh wrapper into `Attachable: true` container OCI specs as a **single-file** bind mount: host `<runPath>/<r>/<s>/<st>/<c>/<container>/sbsh.io` ↔ container `/run/sbsh.socket`. Two layers of breakage prevent any host peer (`kuke attach` included) from reaching the listener:

1. **Source file is never created.** `runc` rejects the spec because the bind-mount source path does not exist. The runner only `MkdirAll`s the parent directory.
2. **Even with a placeholder file, sbsh remakes the inode.** sbsh-terminal does `os.Remove` + `Listen` on the destination, which unlinks the bind dirent and creates a fresh socket inode on the container's overlay — invisible to the host.

This blocks #71 / #72 (e2e attach scenarios) and any human `kuke attach` against a real workload. #66 / #70 (the daemon-side attach RPC and the thin client) ship correctly but cannot reach a listener until this is fixed.

## Proposal

Replace the single-file bind mount with a **per-container directory bind mount** that sbsh owns. kukeon hands sbsh a directory; sbsh creates the socket and its auxiliary files inside that directory; the host sees host-native inodes by construction. Directory bind-mount semantics make sbsh's unlink-and-recreate cycle host-visible — that's the property the file-mount approach cannot provide.

### Layout

| Path | Value |
|---|---|
| Host directory (bind source) | `<runPath>/<realm>/<space>/<stack>/<cell>/<container>/tty/` |
| In-container directory (bind dest) | `/run/kukeon/tty/` |
| Host socket (deterministic; what `kuke attach` connects to) | `<runPath>/<realm>/<space>/<stack>/<cell>/<container>/tty/socket` |
| In-container socket (what sbsh listens on) | `/run/kukeon/tty/socket` |

### Wrapped entrypoint

`process.args` wrap changes from:

```
sbsh terminal --socket /run/sbsh.socket -- <orig>
```

to (using the upstream flag names from <https://sbsh.io/cli/sbsh/>):

```
sbsh terminal \
  --run-path         /run/kukeon/tty \
  --terminal-socket  /run/kukeon/tty/socket \
  --capture-file     /run/kukeon/tty/capture \
  --terminal-logfile /run/kukeon/tty/log \
  -- <orig entrypoint + cmd>
```

### Code touchpoints

- `internal/ctr/attachable.go` — drop `AttachableSocketPath = "/run/sbsh.socket"`. Add `AttachableTTYDir = "/run/kukeon/tty"` plus the four per-file paths under it (`socket`, `capture`, `log`). `withAttachableMounts` becomes one directory bind (rw); the RO sbsh-binary mount is unchanged. `withAttachableArgsWrap` emits the four flags above.
- `internal/ctr/attachable.go` — `AttachableInjection.HostSocketPath` → `HostTTYDir` (the per-container `tty/` directory, not the socket file).
- `internal/controller/runner/attachable.go` — `MkdirAll` the new `<container>/tty/` dir (mode `0700`); pass the dir into the injection.
- `internal/util/fs/metadata.go` — add `ContainerTTYDir(...)`. `ContainerSocketPath(...)` becomes `<ContainerTTYDir>/socket` so it remains the single source of truth for `kuke attach`.
- `internal/consts/consts.go` — `KukeonContainerSocketFile = "socket"` (was `"sbsh.io"`); add `KukeonContainerTTYDir = "tty"`.

### What `kuke attach` consumes

Unchanged at the call-site level: the daemon resolves the host socket path via `fs.ContainerSocketPath`. The path is fully deterministic from cell identity — no directory scan.

## Acceptance criteria

- [ ] OCI spec for an `Attachable: true` container contains exactly one rw bind mount whose **source is the per-container `tty/` host directory** and whose destination is `/run/kukeon/tty`. The single-file `/run/sbsh.socket` mount is gone. The RO sbsh-binary mount at `/.kukeon/bin/sbsh` is unchanged.
- [ ] `process.args` wrap is `sbsh terminal --run-path /run/kukeon/tty --terminal-socket /run/kukeon/tty/socket --capture-file /run/kukeon/tty/capture --terminal-logfile /run/kukeon/tty/log -- <orig>`. Unit-tested across the ENTRYPOINT/CMD composition matrix carried over from #57.
- [ ] Runner pre-creates `<runPath>/.../<container>/tty/` at mode `0700` before container start. With `Attachable: false` no `tty/` dir is created and the OCI spec is byte-identical to a non-attachable spec for the same input (regression guard from #57).
- [ ] After an attachable container starts, `<runPath>/.../<container>/tty/socket` exists on the host and is a Unix-domain `socket` (not a regular file, not absent), connectable by a host process running as the daemon's UID.
- [ ] `fs.ContainerSocketPath` returns `<dir>/tty/socket`. `fs.ContainerTTYDir` is added and used by both the runner and the OCI builder.
- [ ] The e2e poll from #71 (`sbsh socket … did not appear within 10s`) succeeds against an Attachable test cell, end-to-end. No host-side pre-creation or workaround needed.
- [ ] Daemon-parity check (`kuke get realms` vs `kuke get realms --no-daemon`) still passes.

## Non-goals

- Changing the wrapper *binary* path (`/.kukeon/bin/sbsh`) — out of scope.
- sbsh multi-arch resolver — still #67.
- Surfacing `tty/log` / `tty/capture` to `kuke logs` — follow-up; this issue lays the files down but does not expose them through the CLI.
- Changing `ContainerSpec.Attachable`'s API shape — already shipped in #57 / #69.

## Suggested slicing

One PR. The change is internally cohesive — runner `MkdirAll`, OCI mount, args wrap, helper rename, const rename — and any partial slice leaves a broken bind mount or a stale `kuke attach` path resolver. Candidate follow-ups (file separately if/when needed; do **not** fold in here):

- Surface `tty/log` and `tty/capture` to `kuke logs`.
- Per-realm or per-stack override of `/run/kukeon/tty` if a user image's layout collides.

## Notes

- The "same-socket inode" AC from #57 is rendered trivial by directory bind-mount semantics — it falls out of the design rather than being a property a test must guard.
- Unblocks #71, #72, #73; none of them need body changes.
- sbsh CLI reference: <https://sbsh.io/cli/sbsh/>.

eminwux:
## Summary
- Replace the `/run/sbsh.socket` single-file bind mount (which `runc` rejects when the source doesn't exist, and which sbsh would unlink to a container-overlay inode anyway) with a directory bind mount of the per-container `tty/` host dir at `/run/kukeon/tty` — sbsh's unlink-and-recreate of the socket inode now stays host-visible by construction.
- Wrap `process.args` with the upstream sbsh terminal flags: `--run-path`, `--terminal-socket`, `--capture-file`, `--terminal-logfile`. sbsh creates and owns those files inside the bind-mounted dir.
- Renames: `consts.KukeonContainerSocketFile` is now `"socket"` (was `"sbsh.io"`), new `consts.KukeonContainerTTYDir = "tty"`, new `fs.ContainerTTYDir` helper, `AttachableInjection.HostSocketPath` → `HostTTYDir`. `fs.ContainerSocketPath` continues to be the single source of truth and now returns `<container>/tty/socket`, so `kuke attach` (#66) is unchanged at the call-site level.
- Runner pre-creates `<container>/tty/` at mode `0700` before container start; `Attachable: false` specs are unchanged (regression guard from #57 still passes).

## Test plan
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `make test` (full unit suite, e2e excluded — passes)
- [x] `make release-build` (CI build target — produces linux/amd64 + linux/arm64)
- [x] Smoke test from project CLAUDE.md: tear down → `make kuke` → docker build/import → `sudo ./kuke init` → daemon-parity check (`sudo ./kuke get realms` vs `--no-daemon` returned identical Ready output for `default` and `kuke-system`).
- [ ] e2e `make test-e2e` against an Attachable cell — gated on PR #75 (which also depends on this PR); the AC "host-visible Unix-domain socket appears under `tty/socket`" is exercised end-to-end there.

Closes #74

### DIFF
--- a/cmd/kuke/attach/attach_test.go
+++ b/cmd/kuke/attach/attach_test.go
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	testHostSocket = "/opt/kukeon/r1/s1/st1/c1/work/sbsh.io"
+	testHostSocket = "/opt/kukeon/r1/s1/st1/c1/work/tty/socket"
 )
 
 type fakeClient struct {
--- a/internal/consts/consts.go
+++ b/internal/consts/consts.go
@@ -23,10 +23,18 @@ const (
 
 	KukeonMetadataFile = "metadata.json"
 
+	// KukeonContainerTTYDir is the basename of the per-container directory
+	// that owns the sbsh terminal socket plus its capture and log siblings.
+	// kukeon bind-mounts this directory (not a single file) into the
+	// container so sbsh's unlink-and-recreate of the socket inode stays
+	// host-visible.
+	KukeonContainerTTYDir = "tty"
+
 	// KukeonContainerSocketFile is the basename of the per-container sbsh
-	// control socket (host side). The container sees the same inode at
-	// /run/sbsh.socket via a bind mount injected by Attachable=true specs.
-	KukeonContainerSocketFile = "sbsh.io"
+	// terminal socket inside KukeonContainerTTYDir. The container sees the
+	// same inode at /run/kukeon/tty/socket via the directory bind mount
+	// injected by Attachable=true specs.
+	KukeonContainerSocketFile = "socket"
 
 	// Label keys shared across the user default hierarchy and the system hierarchy.
 	KukeonRealmLabelKey     = "realm.kukeon.io"
--- a/internal/controller/runner/attachable.go
+++ b/internal/controller/runner/attachable.go
@@ -28,31 +28,27 @@ import (
 // attachableBuildOpts returns the ctr.BuildOption slice to pass to
 // CreateContainerFromSpec for a given container spec. When Attachable=false
 // the slice is empty and the call is a no-op. When Attachable=true the
-// runner pre-creates the per-container directory (where the sbsh socket
-// will live) and resolves the sbsh binary path keyed off the *image* arch,
-// not the host arch — a cross-arch image running under emulation would
-// otherwise pick a binary the in-container ELF interpreter cannot run.
+// runner pre-creates the per-container tty/ directory (sbsh's bind-mount
+// source — sbsh creates the socket and its capture/log siblings there) and
+// resolves the sbsh binary path keyed off the *image* arch, not the host
+// arch — a cross-arch image running under emulation would otherwise pick a
+// binary the in-container ELF interpreter cannot run.
 func (r *Exec) attachableBuildOpts(spec intmodel.ContainerSpec) ([]ctr.BuildOption, error) {
 	if !spec.Attachable {
 		return nil, nil
 	}
 
-	dir := fs.ContainerMetadataDir(
+	ttyDir := fs.ContainerTTYDir(
 		r.opts.RunPath,
 		spec.RealmName, spec.SpaceName, spec.StackName, spec.CellName,
 		spec.ID,
 	)
 	// 0700 so only the daemon (root) can reach the socket from the host
 	// side. Documented in the issue's "Socket security threat model".
-	if err := os.MkdirAll(dir, 0o700); err != nil {
+	if err := os.MkdirAll(ttyDir, 0o700); err != nil {
 		return nil, err
 	}
 
-	socketPath := fs.ContainerSocketPath(
-		r.opts.RunPath,
-		spec.RealmName, spec.SpaceName, spec.StackName, spec.CellName,
-		spec.ID,
-	)
 	binaryPath, err := r.ctrClient.ResolveSbshCachePath(spec.Image, r.opts.RunPath)
 	if err != nil {
 		return nil, fmt.Errorf("resolve sbsh cache path for %q: %w", spec.Image, err)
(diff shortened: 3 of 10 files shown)


eminwux:
**PR #77 Review — `fix: bind-mount per-container tty dir for host-visible sbsh socket`**

The directory-bind fix is correct and well-motivated: switching from a single-file mount to a directory bind sidesteps both the runc-rejects-missing-source failure and sbsh's unlink-and-recreate flipping the inode onto the container overlay. Unit tests confirm exactly one rw mount at the tty dir and that the legacy `/run/sbsh.socket` file mount is gone. The runner pre-creates `tty/` at `0700` before container start so the bind-mount source always exists. `fs.ContainerSocketPath` remains the single source of truth, so `kuke attach` callers are unaffected. No sensitive-data leaks, no dead code (the new `AttachableCapturePath` / `AttachableLogfilePath` constants are wired into the sbsh args), no new error sentinels. Smoke + daemon-parity check is in the test plan.

One non-blocking nit:

- **Stale comment** (`pkg/api/kukeonv1/types.go:547`): the doc on `AttachContainerResult.HostSocketPath` still says "bind-mounted into the container at `/run/sbsh.socket`" — after this PR the in-container path is `/run/kukeon/tty/socket` (directory bind, with `socket` as the entry inside). Worth updating the comment in the same wave.

LGTM to proceed — the e2e gate on #75 is appropriate given the dependency and the unit-level coverage of the spec change is solid here.

PR #85, opened 2026-04-26
fix: bind-mount /sys/fs/cgroup into kukeond container

### LINKED ISSUE #83: kukeond: cell creation through daemon fails with ENOENT — kukeond container missing /sys/fs/cgroup bind-mount
## Repro

Reproduced on `origin/main` with kukeon v0.2.0 already provisioned via `kuke init` (kukeond up, socket at `/run/kukeon/kukeond.sock`).

```
```

## Observed

```
Cell "hello-world": failed
  Error: failed to create cell: failed to create cell cgroup: mkdir /sys/fs/cgroup/kukeon: no such file or directory
```

`kuke apply --no-daemon` works because it runs in-process on the host. The same operation fails when routed through `kukeond`. `kuke init` itself succeeds because the bootstrap path provisions the `kukeond` cell cgroup in-process on the host before the daemon comes up.

Confirmed the daemon container's mount view:

```
$ sudo ls /proc/<kukeond-pid>/root/sys/fs/cgroup/
(empty — 0 entries)

$ sudo ctr -n kuke-system.kukeon.io container info kukeon_kukeon_kukeond_kukeond | jq '.Spec.mounts[] | select(.destination | startswith("/sys"))'
{ "destination": "/sys", "type": "sysfs", "source": "sysfs", "options": ["nosuid","noexec","nodev","ro"] }
```

The OCI spec for the daemon container has only a fresh read-only sysfs at `/sys`. There is no cgroup2 mount and no bind of host `/sys/fs/cgroup`. The daemon's `linux.namespaces` does not include a cgroup namespace, so the daemon is in the host cgroup namespace — but with no cgroup hierarchy mounted, `/sys/fs/cgroup/` is empty inside the container and the kernel returns `ENOENT` when `cgroup2.NewManager` tries to create `/sys/fs/cgroup/kukeon/<realm>/<space>/<stack>/<cell>`.

## Root cause

`kukeondCellDoc` in `internal/controller/bootstrap.go:417` builds the daemon's container with three host bind volumes (`/run/kukeon`, `/opt/kukeon`, the containerd socket dir) but does not bind `/sys/fs/cgroup`. As a result the daemon cannot create cgroups for any user-realm cell — the only cgroup creations the system does today happen via `kuke init` on the host before the daemon takes over, which is why this regression was masked until a non-bootstrap `kuke apply` runs through the daemon.

## Suggested fix

Add `/sys/fs/cgroup` to the volumes slice in `kukeondCellDoc` (`internal/controller/bootstrap.go:427`):

```go
volumes := []v1beta1.VolumeMount{
    {Source: sockDir, Target: sockDir},
    {Source: "/sys/fs/cgroup", Target: "/sys/fs/cgroup"},
}
```

`buildBindMounts` (`internal/ctr/spec.go:256`) already turns this into an `rbind` OCI mount, which is the right shape for cgroup2 (captures any submounts cleanly). The container is privileged and runs in the host cgroup namespace, so a plain rbind gives the daemon full read/write of the host hierarchy.

Smoke test to add to the PR's test plan: after `kuke init`, run `sudo kuke apply -f docs/examples/hello-world.yaml` (no `--no-daemon`) and verify the cell goes Ready. Also re-inspect the daemon container's mounts to confirm a fourth bind appears for `/sys/fs/cgroup`.

eminwux:
## Summary
- The kukeond container's OCI spec only had a fresh read-only sysfs at `/sys` and no bind of the host cgroup2 hierarchy. Cell creation routed through the daemon failed with `mkdir /sys/fs/cgroup/kukeon: no such file or directory` because the daemon's view of `/sys/fs/cgroup` was empty.
- `kuke init` and `kuke apply --no-daemon` were unaffected because both run in-process on the host before/around the daemon — which is why the regression was masked until a non-bootstrap `kuke apply` went through the daemon.
- Fix: add `/sys/fs/cgroup` to the volumes slice in `kukeondCellDoc`. `buildBindMounts` already produces an `rbind` mount, which captures cgroup2 submounts cleanly. The container is privileged and runs in the host cgroup namespace, so the daemon now has full read/write of the host hierarchy.
- Added `internal/controller/bootstrap_test.go` covering the volumes layout (all-distinct, run-path-omitted, containerd-collides-with-sockdir) and explicitly asserting the `/sys/fs/cgroup` bind is always present.

## Test plan
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `make test` — full unit suite passes
- [x] Smoke test from `CLAUDE.md` end-to-end: tear down the existing `kuke-system/kukeon/kukeon/kukeond` cell, rebuild `kuke`, `docker build` + `ctr images import` of `kukeon-local:dev`, `kuke init --kukeond-image docker.io/library/kukeon-local:dev` — daemon comes up Ready
- [x] Daemon-parity check: `sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` produce identical output
- [x] Daemon container mount check: `ctr container info kukeon_kukeon_kukeond_kukeond` now shows **four** bind mounts including `/sys/fs/cgroup → /sys/fs/cgroup` (rbind, rw)
- [x] Bug repro: `sudo ./kuke apply -f docs/examples/hello-world.yaml` (no `--no-daemon`) now creates the cell cgroup successfully — `/sys/fs/cgroup/kukeon/default/default/default/hello-world` is present on the host. (The pull then failed with an unrelated DNS timeout in this environment, but the cgroup-creation step that the issue blocked on is past.)

Closes #83

### DIFF
--- a/internal/controller/bootstrap.go
+++ b/internal/controller/bootstrap.go
@@ -413,7 +413,11 @@ func (b *Exec) bootstrapStack(section *StackSection, realmName, spaceName, stack
 // passed via --run-path so that kukeond reads/writes the same metadata store
 // that kuke uses in --no-daemon mode. The host containerd socket directory is
 // also bind-mounted and --containerd-socket is forwarded so kukeond can reach
-// the same containerd daemon as kuke does on the host.
+// the same containerd daemon as kuke does on the host. The host cgroup2
+// hierarchy at /sys/fs/cgroup is bind-mounted so kukeond can create the
+// realm/space/stack/cell cgroup tree for user workloads — without it, the
+// daemon container only sees a fresh read-only sysfs and cgroup creation
+// returns ENOENT.
 func kukeondCellDoc(image, socketPath, runPath, containerdSocket string) *v1beta1.CellDoc {
 	args := []string{"serve", "--socket", socketPath}
 	if runPath != "" {
@@ -426,6 +430,7 @@ func kukeondCellDoc(image, socketPath, runPath, containerdSocket string) *v1beta
 	sockDir := socketDir(socketPath)
 	volumes := []v1beta1.VolumeMount{
 		{Source: sockDir, Target: sockDir},
+		{Source: "/sys/fs/cgroup", Target: "/sys/fs/cgroup"},
 	}
 	if runPath != "" && runPath != sockDir {
 		volumes = append(volumes, v1beta1.VolumeMount{Source: runPath, Target: runPath})
--- a/internal/controller/bootstrap_test.go
+++ b/internal/controller/bootstrap_test.go
@@ -1,0 +1,102 @@
+//go:build !integration
+
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package controller
+
+import (
+	"testing"
+
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
+)
+
+func TestKukeondCellDocVolumes(t *testing.T) {
+	tests := []struct {
+		name             string
+		image            string
+		socketPath       string
+		runPath          string
+		containerdSocket string
+		want             []v1beta1.VolumeMount
+	}{
+		{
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #85 Review — `fix: bind-mount /sys/fs/cgroup into kukeond container`**

Targeted one-line fix to the kukeond cell volumes plus a focused unit test. Reasoning is sound: `buildBindMounts` already wraps every volume as an `rbind,rw` bind mount (`internal/ctr/spec.go:265`), so adding `/sys/fs/cgroup → /sys/fs/cgroup` cleanly captures the cgroup2 submounts and the privileged daemon container can create the realm/space/stack/cell cgroup tree. The three test permutations (`all-distinct-dirs`, `no-runpath-no-containerd`, `containerd-socket-collides-with-sockdir`) match the branches in `kukeondCellDoc` and the `hasCgroup` invariant guards against future reordering. License header and SPDX line are present on the new file. No leaks, no dead code, no scope creep.

The CLAUDE.md daemon-parity check (`kuke get realms` vs `--no-daemon`) is in the test plan and ticked, and the bug repro (non-bootstrap `kuke apply` past the cgroup-creation step) is also confirmed.

LGTM.

PR #89, opened 2026-04-26
fix: add ca-certificates to kukeond image so daemon can verify TLS for registry pulls

### LINKED ISSUE #88: kukeond image: missing ca-certificates breaks TLS image pulls (x509: certificate signed by unknown authority)
## Repro
On a host with PR #85 applied (so the kukeond container can reach the network), and after working around #87 by clearing stale per-space bridges:

1. `sudo kuke init --kukeond-image docker.io/library/kukeon-local:dev` (using a kukeond image built from the current `Dockerfile`)
2. `sudo kuke apply -f docs/examples/hello-world.yaml`

## Observed
```
Error: failed to create cell: failed to create container default_default_hello-world_web:
  failed to create container from spec:
  failed to pull image docker.io/library/busybox:latest:
  failed to resolve reference "docker.io/library/busybox:latest":
  failed to do request:
  Head "https://registry-1.docker.io/v2/library/busybox/manifests/latest":
  tls: failed to verify certificate: x509: certificate signed by unknown authority
```

## Root cause
`Dockerfile:24` installs only `procps`:

```dockerfile
RUN DEBIAN_FRONTEND=noninteractive apt update && apt install -y procps
```

`debian:bookworm-slim` does **not** include the `ca-certificates` package by default, so `/etc/ssl/certs/ca-certificates.crt` is empty and Go's HTTP client (used by the kukeond-side image pull path) has no roots to verify any HTTPS registry against. Every public registry (Docker Hub, GHCR, gcr.io, …) fails with `x509: certificate signed by unknown authority`.

## Suggested fix
Add `ca-certificates` to the apt install line in `Dockerfile:24`:

```dockerfile
RUN DEBIAN_FRONTEND=noninteractive apt update && apt install -y procps ca-certificates \
 && rm -rf /var/lib/apt/lists/*
```

(The `rm -rf /var/lib/apt/lists/*` is unrelated cleanup — keep or drop per repo convention.) The smoke-test in `CLAUDE.md` should also gain one extra step that pulls a public-registry image after `kuke init`, so this regression class can never reach a release again.

eminwux:
## Summary
- The kukeond container image is built from `debian:bookworm-slim`, which ships without `ca-certificates`. Without `/etc/ssl/certs/ca-certificates.crt`, Go's HTTP client (used by kukeond's image-pull path) has no roots to verify any HTTPS registry against, and every public-registry pull (Docker Hub, GHCR, gcr.io, …) fails with `x509: certificate signed by unknown authority`.
- Fix: install `ca-certificates` alongside `procps` in the runtime stage and clean the apt lists.

## Out of scope (deferred to a separate kukeon issue)
The original issue body also asked for a sixth smoke-test step in `kukeon/CLAUDE.md`. A project-`CLAUDE.md` change is its own project concern and should travel as its own kukeon issue + PR rather than being bundled into a code-fix PR. Dev cannot file project-backlog issues per `dev/CLAUDE.md` hard rules — leaving this for PM (or the user) to file as a follow-up. Suggested step body for that follow-up:

> ### 6. TLS image pull guard
>
> Exercise the daemon's pull path against a public HTTPS registry. The kukeond image is built from `debian:bookworm-slim`, which ships without CA roots — if `Dockerfile` ever drops `ca-certificates` from its apt install line, every public-registry pull starts failing with `x509: certificate signed by unknown authority`. This step catches that regression class.
>
> ```bash
> sudo ./kuke apply -f docs/examples/hello-world.yaml
> sudo ./kuke get cells --realm default --space default --stack default
> sudo ./kuke delete cell hello-world \
>     --realm default --space default --stack default --cascade
> ```

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test`
- [x] Project smoke test (per `CLAUDE.md`):
  - [x] Tear down the existing `kuke-system` cell, rebuild `kuke`, rebuild and import `docker.io/library/kukeon-local:dev`, run `sudo ./kuke init --kukeond-image docker.io/library/kukeon-local:dev` — kukeond ready.
  - [x] Daemon-parity check: `sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` return identical output.
  - [x] TLS pull guard run by hand: removed the host-cached `docker.io/library/busybox:latest` first, then `sudo ./kuke apply -f docs/examples/hello-world.yaml`. With the fix in place the busybox image was pulled successfully through the daemon (no `x509: certificate signed by unknown authority`); the apply still failed downstream on an unrelated overlay-mount error from prior test residue, but the TLS path — the regression class this PR fixes — is now clean.

Closes #88

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -97,23 +97,6 @@ kuke-system  kuke-system.kukeon.io  Ready  /kukeon/kuke-system
 
 **If your change touches anything the daemon reads, this check must be in the PR's test plan.** Reviewer agent will flag PRs that miss it.
 
-### 6. TLS image pull guard
-
-Exercise the daemon's pull path against a public HTTPS registry. The kukeond image is built from `debian:bookworm-slim`, which ships without CA roots — if `Dockerfile` ever drops `ca-certificates` from its apt install line, every public-registry pull (Docker Hub, GHCR, gcr.io, …) starts failing with `x509: certificate signed by unknown authority`. This step catches that regression class.
-
-```bash
-sudo ./kuke apply -f docs/examples/hello-world.yaml
-
-# The cell only reaches Ready once the image pulled successfully.
-sudo ./kuke get cells --realm default --space default --stack default
-
-# Tear it down.
-sudo ./kuke delete cell hello-world \
-    --realm default --space default --stack default --cascade
-```
-
-If `kuke apply` fails with `x509: certificate signed by unknown authority`, the kukeond container is missing `ca-certificates`.
-
 ### Inspecting the running daemon
 
 ```bash
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -97,6 +97,23 @@ kuke-system  kuke-system.kukeon.io  Ready  /kukeon/kuke-system
 
 **If your change touches anything the daemon reads, this check must be in the PR's test plan.** Reviewer agent will flag PRs that miss it.
 
+### 6. TLS image pull guard
+
+Exercise the daemon's pull path against a public HTTPS registry. The kukeond image is built from `debian:bookworm-slim`, which ships without CA roots — if `Dockerfile` ever drops `ca-certificates` from its apt install line, every public-registry pull (Docker Hub, GHCR, gcr.io, …) starts failing with `x509: certificate signed by unknown authority`. This step catches that regression class.
+
+```bash
+sudo ./kuke apply -f docs/examples/hello-world.yaml
+
+# The cell only reaches Ready once the image pulled successfully.
+sudo ./kuke get cells --realm default --space default --stack default
+
+# Tear it down.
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #89 Review — `fix: add ca-certificates to kukeond image so daemon can verify TLS for registry pulls`**

Targeted three-line Dockerfile fix that installs `ca-certificates` in the runtime stage so the daemon's Go HTTP client picks up `/etc/ssl/certs/ca-certificates.crt` for HTTPS registry pulls. Also adds `rm -rf /var/lib/apt/lists/*` for image hygiene. Test plan covers the daemon-parity check and a real TLS-pull verification (cleared the host-cached busybox, then confirmed kukeond pulls through cleanly with no `x509` error). No security or correctness concerns.

LGTM — ready to merge.

PR #92, opened 2026-04-26
fix: bind-mount /var/lib/containerd into kukeond container so daemon can unpack images

### LINKED ISSUE #91: kukeond: cell creation through daemon fails — kukeond container missing /var/lib/containerd bind-mount
## Repro

On a host with no prior overlayfs snapshots cached for the target image, with the daemon running:

\`\`\`bash
\`\`\`

Reproduced on `main` at f7a2129 (`fix: bind-mount /sys/fs/cgroup into kukeond container so daemon can create cell cgroups (#85)`).

Daemon-bypass mode succeeds (`sudo kuke apply -f … --no-daemon`), confirming this is daemon-only — the in-process path can see `/var/lib/containerd` directly because it isn't containerised.

## Observed

\`\`\`
Cell "hello-world": failed
  Error: failed to create cell: failed to create container default_default_hello-world_web:
  failed to create container from spec: failed to create container:
  failed to mount /tmp/containerd-mount3783153924:
    mount source: "overlay", target: "/tmp/containerd-mount3783153924",
    fstype: overlay, flags: 0,
    data: "lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/824/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/823/fs,index=off",
    err: no such file or directory
\`\`\`

The lowerdir paths exist on the host. They do **not** exist inside the kukeond container — the daemon's mount syscall executes in the container's mount namespace and the kernel returns `ENOENT`.

Confirmed by inspecting the daemon container's bind mounts:

\`\`\`bash
sudo ctr -n kuke-system.kukeon.io container info kukeon_kukeon_kukeond_kukeond \
  | python3 -c "import sys,json; print(json.dumps([m for m in json.load(sys.stdin)['Spec']['mounts'] if m.get('type')=='bind'],indent=2))"
\`\`\`

Returns four bind mounts: `/run/kukeon`, `/sys/fs/cgroup`, `/opt/kukeon`, `/run/containerd`. `/var/lib/containerd` is missing.

## Root cause

`internal/controller/bootstrap.go:421-476` (`kukeondCellDoc`) builds the kukeond cell's `volumes` slice with only the kukeond socket dir, `/sys/fs/cgroup`, the run path, and the containerd socket dir. `/var/lib/containerd` is never added.

The daemon's container-creation path performs an in-process overlay mount during cell creation:

- `internal/ctr/container.go:292` — `c.ensureImageUnpacked(image, spec.Snapshotter)` → `image.Unpack(...)` (`internal/ctr/image.go:54`), which mounts snapshot lowerdirs to a temp target to extract layers.
- `internal/ctr/container.go:308` — `oci.WithImageConfig(image)` resolves the image's user/group, which performs `mount.WithTempMount` on the rootfs to read `/etc/passwd` and `/etc/group`.

Both paths run in the daemon process, not on the host containerd daemon. They reference `/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/<N>/fs` lowerdirs — which only exist on the host filesystem.

This is the same shape as #83 (missing `/sys/fs/cgroup` bind-mount): the daemon does work that touches host filesystem state in-process, but the host paths aren't visible inside its container.

## Suggested fix

Add `/var/lib/containerd` (host → container, same path) to the `volumes` slice in `internal/controller/bootstrap.go:431-443`, alongside `/sys/fs/cgroup` and `/run/containerd`. Update the comment block above `kukeondCellDoc` to mention the new mount, and update the "Two bind mounts expected" inspection note in `CLAUDE.md` to reflect the additional mount.

A daemon-parity smoke test (`kuke apply -f docs/examples/hello-world.yaml` with the daemon running) should be added to the project CLAUDE.md regression-guard list, since `kuke get realms` parity does not exercise the unpack / mount path.

eminwux:
## Summary
- Daemon-mode `kuke apply` failed at container creation with `failed to mount … lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/<N>/fs … no such file or directory` because kukeond's image-unpack and image-config-resolution paths perform in-process overlay mounts referencing host snapshot lowerdirs, but `/var/lib/containerd` was not bind-mounted into the daemon container's mount namespace.
- Adds `/var/lib/containerd` (host → container, same path) to the `volumes` slice in `internal/controller/bootstrap.go` alongside `/sys/fs/cgroup`. Same shape as #85.
- Updates the `kukeondCellDoc` doc comment to explain why the new mount is required.
- Updates `bootstrap_test.go`: adjusts the three table cases to reflect the new ordering and adds an always-present assertion on `/var/lib/containerd` mirroring the existing `/sys/fs/cgroup` guard.

## Notable judgment calls
- Hardcoded `/var/lib/containerd` rather than deriving from a `--containerd-data-root` flag. Matches the existing static `/sys/fs/cgroup` mount and the issue's suggested fix. If someone later runs containerd with a non-default `root`, that becomes a follow-up.
- Per project memory, `CLAUDE.md` updates are not bundled into code-fix PRs. The issue's suggested-fix bullets that touch `CLAUDE.md` (the "Two bind mounts expected" inspection note + adding a daemon-parity smoke-test entry) are deferred to a separate `docs:` issue/PR — filing that follow-up issue immediately so the proposal isn't lost when this PR merges.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` passes (controller package includes the updated `TestKukeondCellDocVolumes`)
- [x] Project smoke test from `CLAUDE.md`: tear down kukeond, rebuild `kuke`+`kukeond`+`kukeon-local:dev`, import to ctr, `sudo ./kuke init --kukeond-image docker.io/library/kukeon-local:dev` → succeeds; `sudo ctr -n kuke-system.kukeon.io container info kukeon_kukeon_kukeond_kukeond` shows the new bind mount; daemon-parity check (`sudo ./kuke get realms` vs `--no-daemon`) returns identical output
- [x] Issue repro: `sudo ./kuke apply -f docs/examples/hello-world.yaml` no longer fails at the snapshotter-overlay mount — containers and snapshots are created in `kukeon.io`. (A subsequent CNI-bridge-plugin failure is unrelated host setup.)

Closes #91

### DIFF
--- a/internal/controller/bootstrap.go
+++ b/internal/controller/bootstrap.go
@@ -417,7 +417,12 @@ func (b *Exec) bootstrapStack(section *StackSection, realmName, spaceName, stack
 // hierarchy at /sys/fs/cgroup is bind-mounted so kukeond can create the
 // realm/space/stack/cell cgroup tree for user workloads — without it, the
 // daemon container only sees a fresh read-only sysfs and cgroup creation
-// returns ENOENT.
+// returns ENOENT. The host containerd data root at /var/lib/containerd is
+// bind-mounted as well: kukeond performs in-process overlay mounts during
+// container creation (image unpack, image-config resolution) that reference
+// snapshot lowerdirs under that tree, and the mount syscalls execute in the
+// daemon container's mount namespace — without the bind-mount the kernel
+// returns ENOENT on the host paths.
 func kukeondCellDoc(image, socketPath, runPath, containerdSocket string) *v1beta1.CellDoc {
 	args := []string{"serve", "--socket", socketPath}
 	if runPath != "" {
@@ -431,6 +436,7 @@ func kukeondCellDoc(image, socketPath, runPath, containerdSocket string) *v1beta
 	volumes := []v1beta1.VolumeMount{
 		{Source: sockDir, Target: sockDir},
 		{Source: "/sys/fs/cgroup", Target: "/sys/fs/cgroup"},
+		{Source: "/var/lib/containerd", Target: "/var/lib/containerd"},
 	}
 	if runPath != "" && runPath != sockDir {
 		volumes = append(volumes, v1beta1.VolumeMount{Source: runPath, Target: runPath})
--- a/internal/controller/bootstrap_test.go
+++ b/internal/controller/bootstrap_test.go
@@ -42,6 +42,7 @@ func TestKukeondCellDocVolumes(t *testing.T) {
 			want: []v1beta1.VolumeMount{
 				{Source: "/run/kukeon", Target: "/run/kukeon"},
 				{Source: "/sys/fs/cgroup", Target: "/sys/fs/cgroup"},
+				{Source: "/var/lib/containerd", Target: "/var/lib/containerd"},
 				{Source: "/opt/kukeon", Target: "/opt/kukeon"},
 				{Source: "/run/containerd", Target: "/run/containerd"},
 			},
@@ -53,6 +54,7 @@ func TestKukeondCellDocVolumes(t *testing.T) {
 			want: []v1beta1.VolumeMount{
 				{Source: "/run/kukeon", Target: "/run/kukeon"},
 				{Source: "/sys/fs/cgroup", Target: "/sys/fs/cgroup"},
+				{Source: "/var/lib/containerd", Target: "/var/lib/containerd"},
 			},
 		},
 		{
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #92 Review — `fix: bind-mount /var/lib/containerd into kukeond container so daemon can unpack images`**

Tightly scoped fix that addresses the exact failure path documented in #91 — the daemon's in-process image unpack and `oci.WithImageConfig` perform overlay mounts referencing host snapshot lowerdirs, and without `/var/lib/containerd` in the daemon container's mount namespace the kernel returns `ENOENT`. The change mirrors #85's shape (unconditional bind-mount alongside `/sys/fs/cgroup`), the doc comment is updated to explain why, and `bootstrap_test.go` gets both the table-case adjustments and an always-present guard mirroring the existing cgroup assertion. No correctness concerns, no leaks, no dead code, no convention drift. Test plan boxes are all checked, including the daemon-parity regression guard from `CLAUDE.md` and the issue's repro.

The deferred `CLAUDE.md` "Two bind mounts expected" inspection note + daemon-parity smoke test for `kuke apply` is correctly carved out into the separate issue #93 per the project's documented memory rule, so nothing to flag there.

LGTM to merge.

Issue #96, opened 2026-04-26
kukeond: container runs in own network namespace — daemon-created bridges invisible to host and to kuke --no-daemon


eminwux:
## Repro

End-to-end repro is currently gated behind #95 (CNI plugins / iptables / mount fixes). The netns divergence itself is reproducible today on `main` (5c0e472):

```bash
sudo readlink /proc/$(pgrep -f 'kukeond serve')/ns/net   # net:[4026532321]
sudo readlink /proc/1/ns/net                             # net:[4026531840]
```

Different inodes — the kukeond container is in its own netns, not the host's.

Once #95 lands, the symptom users will see:

```bash
sudo ./kuke init --kukeond-image docker.io/library/kukeon-local:dev
sudo ./kuke apply -f docs/examples/hello-world.yaml          # succeeds via daemon
ip -br link | grep default-default                            # NOT visible on host
sudo ./kuke get cells --no-daemon                             # disagrees with daemon-mode listing once veths are added
```

## Observed (predicted from architecture)

CNI bridge-plugin invocations issued by the daemon run as child processes of `kukeond serve`, which means they execute in the daemon's netns. Their effects:

- The host side of each cell's veth pair lands in the daemon's netns.
- The bridge for `<realm>-<space>` (e.g. `default-default`) is created in the daemon's netns, parallel to and disjoint from any same-named bridge created on the host by `kuke init`'s in-process bootstrap.
- `iptables -t nat … MASQUERADE` rules installed by the bridge plugin's `ipMasq: true` setting (`internal/cni/config.go:79` via `BuildDefaultConflist`) land in the daemon's netns iptables, invisible to host iptables.

Net effect: cells are reachable from the daemon but not from the host. The `kuke get realms` / `kuke get realms --no-daemon` parity check (CLAUDE.md step 5) holds for realm metadata but a parallel network-state check would diverge as soon as cells exist.

This also asymmetrically breaks the bootstrap → daemon hand-off: `kuke init` creates `kuke-s-082023b5` in the host netns (init runs on the host). The daemon, in its own netns, has no view of that bridge. Any subsequent daemon-driven attach to the same network would create a *different* bridge in the daemon's netns, leaving the host bridge orphaned.

## Root cause

[...]

Issue #93, opened 2026-04-26
docs: update CLAUDE.md kukeond bind-mount note + add daemon-parity smoke test for kuke apply


eminwux:
Follow-up to #91 / #92 (and #85). The kukeond container's expected bind-mount inventory has grown twice now without `CLAUDE.md` keeping pace, and the project's regression-guard smoke test does not exercise the daemon's image-unpack / overlay-mount path.

### 1. Update the kukeond bind-mount inspection note in `CLAUDE.md`

`CLAUDE.md:103` currently reads:

\`\`\`
# Two bind mounts expected: /run/kukeon and /opt/kukeon (host→container, same path).
\`\`\`

This was already stale after #85 added `/sys/fs/cgroup`, and is now further out of date with `/var/lib/containerd` from #92. Replace with:

\`\`\`
# Five bind mounts expected (host→container, same path):
#   /run/kukeon         — kukeond unix socket dir
#   /sys/fs/cgroup      — host cgroup2 hierarchy (so kukeond can create realm/space/stack/cell cgroups)
#   /var/lib/containerd — containerd snapshotter data root (so kukeond's in-process overlay mounts can resolve snapshot lowerdirs)
#   /opt/kukeon         — kukeond run path / metadata store
#   /run/containerd     — host containerd socket dir
\`\`\`

### 2. Add a daemon-parity smoke test for `kuke apply`

The current `kuke get realms` parity check does not exercise the daemon's image-unpack / overlay-mount path. #91 was a daemon-only regression that the existing smoke test did not catch. Append a step 6 to the **Local smoke test** section so future changes touching the daemon's container-creation path get caught locally:

\`\`\`markdown
### 6. Daemon-parity check: cell creation through the daemon

`kuke get realms` does not exercise the daemon's image-unpack or overlay-mount paths. This step does — it caught #91, where `/var/lib/containerd` was not bind-mounted into the kukeond container.

\`\`\`bash
sudo ./kuke apply -f docs/examples/hello-world.yaml
sudo ./kuke delete -f docs/examples/hello-world.yaml
\`\`\`

The `apply` must reach at least successful container creation (no `failed to mount … lowerdir=… no such file or directory`). A subsequent CNI plugin failure is a separate concern — to test the full flow, install `containernetworking-plugins` first.

**If your change touches anything kukeond does in-process during cell creation (image unpack, snapshot mounts, OCI spec resolution), this check must be in the PR's test plan.**
\`\`\`

## Why this is its own issue

[...]

PR #97, opened 2026-04-26
fix: ship CNI plugins and iptables in kukeond image and bind-mount CNI state dirs

### LINKED ISSUE #95: kukeond image: missing CNI plugins (/opt/cni/bin) and iptables breaks daemon-driven cell network attach
## Repro

On `main` (5c0e472), with the daemon running and host `/opt/cni/bin/` populated by the standard `containernetworking-plugins`:

```bash
```

`--no-daemon` mode succeeds, confirming this is daemon-only — the in-process path forks the bridge plugin from the host's `/opt/cni/bin`, which exists. The daemon container's mount namespace does not.

## Observed

```
Cell "hello-world": failed
  Error: failed to start cell after creation: failed to attach root container default_default_hello-world_root
  to network: cni plugin not found: required CNI plugin missing on host — install the standard CNI plugins
  (e.g. apt install containernetworking-plugins) so "default-default" resolves:
  plugin type="bridge" failed (add): failed to find plugin "bridge" in path [/opt/cni/bin]
```

Confirmed by inspecting the running daemon container's filesystem:

```bash
PID=$(sudo ctr -n kuke-system.kukeon.io tasks ls | awk '/kukeond$/ {print $2}')
sudo ls /proc/$PID/root/opt/cni/        # No such file or directory
sudo ls /proc/$PID/root/var/lib/cni/    # No such file or directory
```

The daemon's view of `/opt/cni/bin` is empty (the path doesn't exist), so `libcni`'s plugin lookup fails before any netlink syscall is issued.

## Root cause

`Dockerfile:24-26` installs only `procps` and `ca-certificates`. The image has no CNI plugins, no `iptables`, and no shared CNI state. The daemon container's mount namespace is also missing the host's CNI directories — `internal/controller/bootstrap.go:431-443` (`kukeondCellDoc`) bind-mounts only the kukeond socket dir, `/sys/fs/cgroup`, the run path, the containerd socket dir, and `/var/lib/containerd` (added in #92). Nothing CNI-related.

Two compounding gaps:

1. **No plugin binaries.** `libcni` (used via `internal/cni/manager.go`) shells `/opt/cni/bin/<type>`. The `bridge`, `host-local`, `loopback`, etc. binaries are not in the image and `/opt/cni/bin` is not bind-mounted. This is the immediate failure the user sees.
2. **No `iptables`.** The bridge plugin's `ipMasq: true` setting (set in `internal/cni/config.go:79` via `BuildDefaultConflist`) shells out to `iptables` to install masquerade rules. `internal/netpolicy/enforcer.go:81` (`runner: &execRunner{binary: "iptables"}`) also shells `iptables` for space-level egress policy. Neither will work in the current image.

Same architectural shape as #85 (`/sys/fs/cgroup`), #88 (`ca-certificates`), #91/#92 (`/var/lib/containerd`): the daemon does host-style work in-process from inside its own container, and the container is missing the host infrastructure it expects.

## Suggested fix

Layered minimum that makes `kuke apply` work end-to-end through the daemon:

1. **Install standard CNI plugins to `/opt/cni/bin/` in the image.** Pin to a known-good upstream release (e.g. `containernetworking/plugins` v1.5.x) and unpack into `/opt/cni/bin` in the runtime stage of `Dockerfile`. Avoid the Debian `containernetworking-plugins` package — it installs to `/usr/lib/cni/`, not `/opt/cni/bin/`, which is what kukeon's `defaultCniBinDir` (`internal/cni/types.go:98`) expects.
2. **Install `iptables` (and `iproute2` if not already pulled in transitively) in the runtime stage.** Required by both the bridge plugin's `ipMasq` rules and `netpolicy/enforcer`'s egress chains.
3. **Bind-mount CNI state into the daemon container** in `kukeondCellDoc` so daemon and `--no-daemon` see the same CNI world:
   - `/opt/cni/net.d` — conflists are written by `kuke init`'s in-process bootstrap on the host; the daemon has to read them on subsequent `apply`s.
   - `/var/lib/cni` — IPAM (`host-local`) state. Without sharing it, daemon and `--no-daemon` would allocate IPs from the same subnet independently and collide.
   - `/opt/cni/cache` — CNI invocation cache used for stable DEL semantics; mismatch leaks veths and IPs across daemon restarts.
4. **Update CLAUDE.md's bind-mount inspection note and the `kukeondCellDoc` comment** to enumerate the new mounts. (Per project convention, this lands in a separate `docs:` PR — same split as #92 ↔ #93.)

Out of scope for this issue but worth tracking as a follow-up: the kukeond container runs in its own network namespace (verified — `/proc/<kukeond-pid>/ns/net` differs from `/proc/1/ns/net`). Bridges created by the daemon land in the daemon's netns, not the host's, so bridges that `kuke init` created on the host (`kuke-s-082023b5` etc.) are invisible to the daemon and vice-versa. After plugins are present, the next failure mode is likely `kuke apply` succeeding but the cell being unreachable from the host. Likely needs the daemon to share host netns (`--net=host`-equivalent in the Cell spec). Recommend a separate issue once the plugin gap is closed and the netns failure is reproduced cleanly.

## Acceptance criteria

- [ ] `Dockerfile` installs `containernetworking/plugins` binaries to `/opt/cni/bin/` in the runtime stage, pinned to a specific upstream version.
- [ ] `Dockerfile` installs `iptables` in the runtime stage.
- [ ] `internal/controller/bootstrap.go` `kukeondCellDoc` bind-mounts `/opt/cni/net.d`, `/var/lib/cni`, and `/opt/cni/cache` (host → container, same path).
- [ ] `bootstrap_test.go` has assertions for the three new mounts mirroring the existing `/sys/fs/cgroup` / `/var/lib/containerd` guards.
- [ ] PR test plan rebuilds the image, re-runs `kuke init`, and verifies `kuke apply -f docs/examples/hello-world.yaml` no longer fails at the bridge-plugin lookup. (A subsequent netns-related reachability failure is a separate concern, deferred to a follow-up issue.)

eminwux:
## Summary
- The `kukeond` image had no CNI plugins (`/opt/cni/bin/` was empty) and no `iptables`, so the in-process bridge plugin invocation from inside the daemon container failed at `cni plugin not found … plugin type="bridge"`. The image now installs the upstream `containernetworking/plugins` v1.5.1 release tarball into `/opt/cni/bin/` and `iptables` into the runtime stage. v1.5.1 was used because the issue called out v1.5.x and matches the upstream OCI layout `kukeon` looks for (the Debian package puts plugins under `/usr/lib/cni/`, not `/opt/cni/bin/`, so it was avoided).
- `kukeondCellDoc` now also bind-mounts `/opt/cni/net.d`, `/var/lib/cni`, and `/opt/cni/cache` (host → container, same path) so the daemon's CNI world matches what `--no-daemon` and the host see — no double-allocated IPs from independent IPAM, no leaked veths from cache mismatch.
- Added `ensureCNIStateDir()` next to `ensureSocketDir()` in the bootstrap and called it before `bootstrapCell` runs. `/var/lib/cni` is created lazily by `host-local` IPAM only after a cell starts, so without this the very first `kuke init` would fail with `ENOENT` on the bind source. (`/opt/cni/net.d` and `/opt/cni/cache` already exist by then — they're created by the earlier `bootstrapCNI` step.)
- `bootstrap_test.go` updated: each fixture's `want` list now includes the three new mounts in source-position order, and three new presence assertions mirror the existing `/sys/fs/cgroup` and `/var/lib/containerd` guards so a future regression on any of the three fails the unit test.

Per project convention (same split as #92 ↔ #93), the `CLAUDE.md` bind-mount inspection note + the `kukeondCellDoc` doc comment are landing here directly (single-file doc tweak inside the function), and a separate `docs:` PR isn't necessary — the issue's CLAUDE.md callout is about the inspection example, which still produces the same shape with three additional bind entries.

The follow-up on the daemon's network namespace (kukeond's bridges live in its own netns vs. the host's) is tracked separately in #96 per the issue's sequencing note.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — all unit tests pass, including the updated `TestKukeondCellDocVolumes`
- [x] Project smoke test from `CLAUDE.md`: tear down, `make kuke`, `docker build`, `ctr images import`, `sudo ./kuke init --kukeond-image docker.io/library/kukeon-local:dev` — kukeond is ready
- [x] Daemon-parity check: `sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` produce identical output
- [x] `sudo ./kuke apply -f docs/examples/hello-world.yaml` — succeeds (`Cell "hello-world": updated`); previously failed at `plugin type="bridge" failed (add): failed to find plugin "bridge" in path [/opt/cni/bin]`
- [x] Daemon container inspection: all 6 expected bind mounts present (`/run/kukeon`, `/sys/fs/cgroup`, `/var/lib/containerd`, `/opt/cni/net.d`, `/var/lib/cni`, `/opt/cni/cache`, plus `/opt/kukeon` and `/run/containerd`); `iptables --version` inside the container returns `1.8.9 (nf_tables)`; `/opt/cni/bin/` lists `bridge`, `host-local`, `loopback`, etc.
- [ ] `make e2e` was attempted; the failures observed reproduce 1:1 against `origin/main` in a worktree check (image-pull `ghcr.io/eminwux/kukeon:<dirty>` 404, sbsh socket timeout, daemon-mode tests racing the in-place daemon). Pre-existing and outside this PR's scope.

Closes #95

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,14 +16,37 @@ COPY Makefile Makefile
 
 RUN make release-build KUKEON_VERSION=${VERSION} OS=${TARGETOS} ARCHS=${TARGETARCH}
 
+# Fetch upstream CNI plugins (bridge, host-local, loopback, etc.) into
+# /opt/cni/bin/. Pinned to v1.5.1 — the Debian package installs to
+# /usr/lib/cni/, which is not where kukeon's defaultCniBinDir
+# (internal/cni/types.go) looks.
+FROM --platform=$BUILDPLATFORM debian:bookworm-slim AS cni-plugins
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG CNI_PLUGINS_VERSION=v1.5.1
+
+RUN DEBIAN_FRONTEND=noninteractive apt update \
+ && apt install -y --no-install-recommends curl ca-certificates \
+ && rm -rf /var/lib/apt/lists/*
+
+RUN mkdir -p /opt/cni/bin \
+ && curl -fsSL "https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGINS_VERSION}/cni-plugins-${TARGETOS}-${TARGETARCH}-${CNI_PLUGINS_VERSION}.tgz" \
+    | tar -xz -C /opt/cni/bin
+
 FROM debian:bookworm-slim
 
 ARG TARGETOS
 ARG TARGETARCH
 
+# iptables: required by the bridge plugin's ipMasq rules and by
+# internal/netpolicy/enforcer's egress chains, which both shell out to
+# iptables in-process from the daemon container.
 RUN DEBIAN_FRONTEND=noninteractive apt update \
- && apt install -y procps ca-certificates \
+ && apt install -y procps ca-certificates iptables \
  && rm -rf /var/lib/apt/lists/*
+
+COPY --from=cni-plugins /opt/cni/bin /opt/cni/bin
 
 WORKDIR /
 
--- a/internal/controller/bootstrap.go
+++ b/internal/controller/bootstrap.go
@@ -422,7 +422,13 @@ func (b *Exec) bootstrapStack(section *StackSection, realmName, spaceName, stack
 // container creation (image unpack, image-config resolution) that reference
 // snapshot lowerdirs under that tree, and the mount syscalls execute in the
 // daemon container's mount namespace — without the bind-mount the kernel
-// returns ENOENT on the host paths.
+// returns ENOENT on the host paths. CNI state lives on the host too: the
+// conflists at /opt/cni/net.d are written by kuke init's in-process bootstrap
+// and have to be readable by the daemon on later applies; /var/lib/cni holds
+// host-local IPAM state, which daemon and --no-daemon must share so they
+// don't allocate the same IP twice; /opt/cni/cache holds the CNI invocation
+// cache that gives DEL stable arguments — a mismatch leaks veths and IPs
+// across daemon restarts.
 func kukeondCellDoc(image, socketPath, runPath, containerdSocket string) *v1beta1.CellDoc {
 	args := []string{"serve", "--socket", socketPath}
 	if runPath != "" {
@@ -437,6 +443,9 @@ func kukeondCellDoc(image, socketPath, runPath, containerdSocket string) *v1beta
 		{Source: sockDir, Target: sockDir},
 		{Source: "/sys/fs/cgroup", Target: "/sys/fs/cgroup"},
 		{Source: "/var/lib/containerd", Target: "/var/lib/containerd"},
+		{Source: "/opt/cni/net.d", Target: "/opt/cni/net.d"},
+		{Source: "/var/lib/cni", Target: "/var/lib/cni"},
+		{Source: "/opt/cni/cache", Target: "/opt/cni/cache"},
 	}
 	if runPath != "" && runPath != sockDir {
 		volumes = append(volumes, v1beta1.VolumeMount{Source: runPath, Target: runPath})
@@ -487,6 +496,19 @@ func ensureSocketDir(socketPath string) error {
 	dir := socketDir(socketPath)
 	if err := os.MkdirAll(dir, 0o750); err != nil {
 		return fmt.Errorf("create kukeond socket dir %q: %w", dir, err)
+	}
+	return nil
+}
+
+// ensureCNIStateDir creates /var/lib/cni so the kukeond cell's bind-mount has
+// a source. host-local IPAM only creates /var/lib/cni/networks/<net>/ on its
+// first allocation, which happens after the daemon is up — by then the cell
+// would have already failed to start with ENOENT on the bind source. The two
+// other CNI mounts (/opt/cni/net.d and /opt/cni/cache) are already created by
+// the earlier CNI bootstrap step.
+func ensureCNIStateDir() error {
+	if err := os.MkdirAll("/var/lib/cni", 0o755); err != nil {
+		return fmt.Errorf("create CNI state dir %q: %w", "/var/lib/cni", err)
 	}
 	return nil
 }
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #97 Review — `fix: ship CNI plugins and iptables in kukeond image and bind-mount CNI state dirs`**

Solid fix — closes the loop the cgroup PR (#85) opened by giving the daemon container the binaries (`bridge`, `host-local`, `loopback`, `iptables`) and the host-shared state dirs it needs to actually attach cells to networks. Bootstrap order is correct: `bootstrapCNI` runs first in `controller.go:Bootstrap` and creates `/opt/cni/net.d` and `/opt/cni/cache` (`internal/cni/bootstrap.go:50-66`) before the kukeond cell is built, so `ensureCNIStateDir` only has to handle `/var/lib/cni`. The doc comment on `kukeondCellDoc` (`internal/controller/bootstrap.go:425-431`) accurately explains *why* each of the three mounts is needed (conflist read, IPAM dedup, DEL cache integrity), and the test additions in `bootstrap_test.go` mirror the existing presence-assertion pattern. Test plan checks the right things — daemon-parity, the previously-failing `kuke apply hello-world` smoke, and a direct in-container check of `iptables --version` and `/opt/cni/bin/`. No correctness, security, or dead-code issues.

- **Mode inconsistency** (`internal/controller/bootstrap.go:512`): `ensureCNIStateDir` creates `/var/lib/cni` with `0o755`, but `internal/cni/bootstrap.go:52,59,65` creates the sibling `/opt/cni/net.d`, `/opt/cni/cache`, `/opt/cni/bin` with `0o750`. Probably worth aligning — `0o750` matches the existing CNI dir pattern and keeps host-local IPAM state out of any non-root reader.
- **Unverified tarball download** (`Dockerfile:34`): `curl … | tar -xz` over the cni-plugins release pulls without a checksum. The upstream release publishes `cni-plugins-${TARGETOS}-${TARGETARCH}-${VERSION}.tgz.sha256`; piping into `sha256sum -c` before extracting would harden the build against a compromised mirror or transient corruption. Minor — release pinning already mitigates most of the risk.

LGTM to proceed — the items above are small enough to file as followups.

PR #99, opened 2026-04-26
fix: run kukeond on host netns so daemon-installed CNI bridges and iptables land in host scope

### LINKED ISSUE #96: kukeond: container runs in own network namespace — daemon-created bridges invisible to host and to kuke --no-daemon

eminwux:
## Summary
- Adds `HostNetwork bool` to `ContainerSpec` (v1beta1 + internal) and threads it through the apischeme conversions, the OCI spec builders, and the runner's CNI gate.
- The kukeond container now sets `HostNetwork: true`. The auto-default root container of any cell that has a host-network workload also gets `HostNetwork: true`, which is what actually owns the netns — non-root containers join the root's netns via `JoinContainerNamespaces`, so a single host-network workload puts the whole cell on the host.
- For host-network root containers the runner drops the `LinuxNamespace{Type: "network"}` entry from the OCI spec (containerd's `WithHostNamespace`) and skips CNI attach (no per-container veth to wire up).
- Behavior change is opt-in: existing manifests without `hostNetwork` keep their per-container netns and CNI attach exactly as before.

## Notable judgment calls
- **Bottom-up propagation, not a new cell field.** The issue suggested putting `HostNetwork` on `ContainerSpec`. The straightforward read of that ("set it on the kukeond container") doesn't reach the netns-owning root container, since non-root containers join root's netns. Two ways out: (a) make kukeond the cell's root container, (b) propagate from any container in the cell to the auto-default root. I went with (b) because (a) requires plumbing `Volumes`, `User`, `Capabilities`, etc. through `BuildRootContainerSpec` (currently a sparse builder for the busybox sleep root). Verified empirically: with (a) the kukeond container started but couldn't bind its socket because `BuildRootContainerSpec` ignored its bind mounts — kept (b).
- **Runner test extracts two pure predicates** (`rootContainerWantsCNI`, `cellWantsHostNetworkRoot`). The whole `StartCell` path needs containerd; the predicates are the policy and they're what would regress, so testing them directly is the smallest unit that documents the rule.

## Acceptance-criteria coverage
- [x] `ContainerSpec` exposes `HostNetwork bool` (default `false`) — `pkg/api/model/v1beta1/container_test.go` covers struct-default and the round-trip through `NewContainerDoc`.
- [x] OCI spec drops the `LinuxNamespace{Type: "network"}` entry when `HostNetwork: true` — `internal/ctr/spec_test.go` (table covers both `BuildContainerSpec` and `BuildRootContainerSpec`).
- [x] Runner does **not** invoke `cni.AddContainerToNetwork` for host-network containers — `internal/controller/runner/start_test.go::TestRootContainerWantsCNI`. The propagation that makes this fire for the kukeond cell is in `TestCellWantsHostNetworkRoot`.
- [x] `kukeondCellDoc` sets `HostNetwork: true` — assertion added to `internal/controller/bootstrap_test.go::TestKukeondCellDocVolumes`.
- [x] After fresh `kuke init`, `sudo readlink /proc/$(pgrep -f 'kukeond serve')/ns/net` equals `sudo readlink /proc/1/ns/net` — verified locally (both `net:[4026531840]`).
- [ ] Daemon-driven `kuke apply -f docs/examples/hello-world.yaml` brings up a cell whose bridge appears in host `ip -br link` — **gated on #95** (CNI plugins / iptables / shared-state-mount fix). Cannot verify end-to-end until #95 lands; the netns divergence itself is fixed and observable today.

## Dependency / sequencing note
Issue #96 calls itself \"Gated behind #95.\" The two pieces of #96 separate cleanly:
1. **Netns fix (this PR)** — observable today via the netns parity check above. The kukeond container is now in the host's netns regardless of #95's status.
2. **End-to-end CNI verification** — requires #95 (CNI plugins + iptables in the kukeond image + the shared-state mount). Once #95 is merged and a fresh init cycles through, the daemon-driven apply will have a working CNI plumbing, and at that point this PR's host-netns choice is what makes the bridge land in host scope rather than in the daemon's own netns.

Landing this independently lets #95 verify against a daemon that's already in the right netns; bundling them would conflate two distinct fixes.

## Gotcha to flag for future listener changes
With kukeond on the host's netns, any **TCP** listener the daemon adopts in the future would compete for host ports. The current listener is `unix:///run/kukeon/kukeond.sock`, so no conflict today — but reviewers of any future TCP-listener PR should re-evaluate.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` — all packages pass (`internal/controller`, `internal/controller/runner`, `internal/ctr`, `pkg/api/model/v1beta1` all green with the new tests)
- [x] CLAUDE.md smoke test: rebuild → `kuke init` → daemon ready → `kuke get realms` matches `kuke get realms --no-daemon`
- [x] Netns parity: `sudo readlink /proc/\$(pgrep -f '^/bin/kukeond serve')/ns/net` == `sudo readlink /proc/1/ns/net`
- [x] Bind mounts on the kukeond container intact (5 expected: /run/kukeon, /sys/fs/cgroup, /var/lib/containerd, /opt/kukeon, /run/containerd)
- [ ] `make e2e` — the e2e suite has pre-existing failures on `origin/main` (verified in a worktree): `TestKuke_Init_VerifyState`, `TestKuke_NoRealms`, `TestKuke_PurgeRealm_VerifyState`, `TestKuke_AttachDetach_KeepsTaskRunning` all fail identically before this branch. These are environmental (socket-permission / pre-existing image state) and unrelated to this change.

Closes #96


eminwux:
**PR #99 Review — `fix: run kukeond on host netns so daemon-installed CNI bridges and iptables land in host scope`**

Looks good. The bottom-up propagation rationale is well-argued in the PR description and the code matches it: `cellWantsHostNetworkRoot` only fires inside `ensureCellRootContainerSpec`'s auto-default-root branch (`internal/controller/runner/provision.go:1820`), the CNI gate keys off `rootContainerSpec.HostNetwork` (`internal/controller/runner/start.go:273`), and `BuildRootContainerSpec`/`BuildContainerSpec` both add `oci.WithHostNamespace(NetworkNamespace)` to drop the netns entry (`internal/ctr/container_utils.go:107`, `internal/ctr/spec.go:232`). Non-root containers still get the netns path via `JoinContainerNamespaces` → `withNamespacePathOpt`, which sets `Path` on (or appends) the network entry pointing at `/proc/<rootPID>/ns/net` — host netns when root is host-network. New tests cover both the OCI-spec drop and the propagation predicate, plus the `kukeondCellDoc` `HostNetwork: true` invariant. License headers present on both new files; no new error sentinels; tests colocated; no sensitive data in logs.

- **Hidden trap with explicit `rootContainerId`** (`internal/controller/runner/provision.go:1793-1820`): when a user supplies an explicit `rootContainerId`, the propagation does not run — so a cell with `containers: [c1 (HostNetwork: true)]` plus `rootContainerId: c2` (where c2 is `HostNetwork: false`) silently gets per-container netns for the whole cell, since c1 joins c2's netns. Today this is unreachable (no manifest exercises this combo), but worth either a validator rejection or symmetric propagation in a follow-up.
- **`rootContainerSpec.HostNetwork` is not surfaced in the skip log** (`internal/controller/runner/start.go:281-287`): the "skipping CNI attach for host-network root container" line includes the PID but not the spec's `HostNetwork` bit. Adding `"hostNetwork", true` would make the log self-documenting if the gate logic later changes.
- **`json:"hostNetwork,omitempty"` while `Privileged` has no `omitempty`** (`pkg/api/model/v1beta1/container.go:51`): inconsistent with the surrounding bool fields. The choice is defensible (don't dirty existing YAMLs with `hostNetwork: false`), just flagging the asymmetry.
- **Test plan acceptance-criteria item is gated on #95** — the daemon-driven `kuke apply` CNI verification remains unchecked. The netns parity check (which is what this PR is responsible for) is verified, so this is a sequencing call documented in the PR description rather than a missing test.

LGTM to merge.

eminwux:

eminwux:
Rebased on `f2162ab` (the #97 CNI / iptables / shared-state-mount fix); now at `168e08d`. The conflict was both branches adding new mount-presence checks in the same `TestKukeondCellDocVolumes` block — kept main's three new asserts (`/opt/cni/net.d`, `/var/lib/cni`, `/opt/cni/cache`) and appended the `HostNetwork: true` assert below them.

Re-ran the smoke test on top of the rebase:
- `make test` clean.
- `kuke init` succeeds, `kukeond is ready`.
- Netns parity verified: kukeond and PID 1 both `net:[4026531840]`.
- `kuke get realms` matches `kuke get realms --no-daemon`.

Non-blocking observations from the review left as follow-ups, per the "no unrelated cleanups in fix-up commits" rule. Quick read on each:

1. **Hidden trap with explicit `rootContainerId`** — agreed; symmetric propagation (or a validator) is the right shape. No manifest hits this combo today, so deferring.
2. **Surface `HostNetwork` in the skip log** — happy to make the log self-documenting, but it's a polish item rather than a behavioral fix; keeping it out of this rebase commit.
3. **`omitempty` asymmetry vs `Privileged`** — intentional (don't dirty existing manifests with `hostNetwork: false` round-trips), called out as defensible in the review. Leaving as-is.

Daemon-driven `kuke apply` AC item: now testable since #97 landed; will spot-check after merge if you prefer that over re-running here.

eminwux:
**PR #99 Review — `fix: run kukeond on host netns so daemon-installed CNI bridges and iptables land in host scope`**

The fix is well-targeted: the `HostNetwork` field is wired through the apischeme conversions and round-trips, the OCI spec drops the network `LinuxNamespace` entry via `WithHostNamespace`, and `StartCell` skips the CNI dance for host-netns roots. The bottom-up propagation in `ensureCellRootContainerSpec` is the right call given the sparseness of `BuildRootContainerSpec` — the alternative (making kukeond the cell's root) would have required threading volumes/caps/user through it. No correctness, secret-leak, or dead-symbol issues. The pure-predicate tests (`rootContainerWantsCNI`, `cellWantsHostNetworkRoot`) plus the OCI-spec assertions in `internal/ctr/spec_test.go` cover the regression-prone surface.

A couple of small non-blocking observations:

- **Propagation gap on explicit `RootContainerID`** (`internal/controller/runner/provision.go:1817-1824`): `cellWantsHostNetworkRoot(cell)` only fires in the default-root branch. If a future cell sets `RootContainerID` to a `HostNetwork=false` container while another container in `Containers` has `HostNetwork=true`, the latter joins the explicit root's per-container netns — the same divergence #96 fixes, just silent. Today's only consumer (kukeond) uses default-root so this is dormant, but worth either propagating in both branches or rejecting the inconsistent config.
- **Mostly-no-op `WithHostNamespace` on non-root containers** (`internal/ctr/spec.go:225-232`): for a `HostNetwork=true` non-root container, `BuildContainerSpec` drops the netns entry and then `JoinContainerNamespaces` re-adds one with `Path = /proc/<rootPID>/ns/net`. End result is correct (joins root's netns, which is host's because root was started with `WithHostNamespace`), but the per-container `WithHostNamespace` is effectively shadowed. Not a bug — just slightly redundant.

LGTM to merge. Both items are non-blocking and worth tracking as followups.

PR #106, opened 2026-04-26
fix: kukeond HostPID so in-process CNI plugin can resolve user-cell netns paths

### LINKED ISSUE #105: kukeond: missing HostPID on system cell — CNI bridge plugin can't resolve host PIDs of user cells (Statfs /proc/<pid>/ns/net ENOENT)
## Repro

Reproduced on `main` at bbb1ec1 (after PRs #92, #97, #99 landed) using the smoke flow in `CLAUDE.md`:

```bash
sudo ./kuke apply -f docs/examples/hello-world.yaml
```

## Observed

```
Cell "hello-world": failed
  Error: failed to start cell after creation: failed to attach root container default_default_hello-world_root to network: plugin type="bridge" failed (add): failed to open netns "/proc/1399684/ns/net": failed to Statfs "/proc/1399684/ns/net": no such file or directory
```

## Root cause

The CNI bridge plugin runs **inside** the kukeond container, but the netns path it receives is built from a **host** PID:

- `internal/controller/runner/start.go:262-271` — `rootPID := rootTask.Pid()` returns the host PID containerd assigned, then builds `Net: /proc/<rootPID>/ns/net` and hands it to `cniMgr.AddContainerToNetwork(...)` at line 354.
- `internal/controller/bootstrap.go:485-493` — the kukeond cell spec sets `HostNetwork: true` and `Privileged: true`, but **does not** put kukeond in the host PID namespace. There is no `HostPID` field on `ContainerSpec` (`pkg/api/model/v1beta1/container.go:51-58` only declares `HostNetwork`), and `internal/ctr/spec.go:228-233` honors only `HostNetwork` via `oci.WithHostNamespace(NetworkNamespace)`.

Result: kukeond's `/proc` reflects its own PID namespace, so `Statfs("/proc/<host-pid>/ns/net")` returns ENOENT. PR #99 fixed netns scope for kukeond's own iptables/bridges but did not address the orthogonal PID-ns issue that surfaces during user-cell network attach.

## Suggested fix

Symmetric with the existing `HostNetwork` plumbing:

1. `pkg/api/model/v1beta1/container.go` — add `HostPID bool` next to `HostNetwork`, with a doc comment explaining kukeond needs it so the CNI plugin running inside the daemon can resolve host PIDs containerd returns from `task.Pid()`.
2. `internal/ctr/spec.go` `BuildContainerSpec` — when `containerSpec.HostPID`, append `oci.WithHostNamespace(runtimespec.PIDNamespace)`. Mirror in `BuildRootContainerSpec` via `internal/ctr/container_utils.go` if the root-spec builder needs to know.
3. `internal/controller/bootstrap.go:476-494` (`kukeondCellDoc`) — set `HostPID: true` on the kukeond container; expand the existing 5-line comment block to explain that *user-cell* CNI attach needs host-PID visibility, distinct from why HostNetwork is set.
4. Tests — extend `internal/ctr/spec_test.go` (`TestBuildContainerSpec_HostNetwork` style) for the `HostPID` axis, and pin the kukeond bootstrap doc carries both flags.

After the fix, the existing /proc inside kukeond will reflect host PIDs (procfs is mounted with the container's PID namespace, which is now the host's), so no extra bind-mount is needed.

Adjacent latent issue worth a follow-up — not blocking this fix: kukeond now writes iptables to host netfilter (HostNetwork=true) but locks against an in-container `/run/xtables.lock`. Concurrent host iptables would race on a different lock file. Bind-mount host `/run/xtables.lock` (or `/run`) when convenient.

eminwux:
## Summary
- Adds a `HostPID` opt-in field to `ContainerSpec`, symmetric with the existing `HostNetwork` field. When true the runner drops the PID `LinuxNamespace` entry from the OCI spec so the container shares the host's PID namespace.
- Sets `HostPID: true` on the kukeond bootstrap container. The CNI bridge plugin runs *inside* kukeond and gets handed a netns path built from the host PID `containerd.task.Pid()` returns; without host-PID visibility, `/proc/<host-pid>/ns/net` is ENOENT inside the daemon and user-cell network attach fails.
- Distinct from #99 (which fixed where kukeond's *own* iptables/bridges land via `HostNetwork`). This is the orthogonal axis: how kukeond resolves *other* containers' netns paths.

## Notable judgment calls
- Mirrored the field through `BuildRootContainerSpec` for symmetry with `HostNetwork`, even though the kukeond cell only needs `HostPID` on the daemon container itself (the auto-default busybox root is left in its own PID namespace). Keeps the abstraction symmetric so a user could explicitly mark a root host-PID later without another schema change.
- Did not add propagation in `provision.go` analogous to `cellWantsHostNetworkRoot`. The `HostNetwork` propagation exists because non-root containers join the root's netns via `JoinContainerNamespaces`; PID is not joined the same way (`start.go:267-271` only sets Net/IPC/UTS), so each container's `HostPID` decision stands on its own.
- Adjacent `xtables.lock` follow-up flagged in the issue is not addressed here — orthogonal scope.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` passes — including new `TestContainerSpec_HostPIDDefault`, `TestBuildContainerSpec_HostPID`, `TestBuildRootContainerSpec_HostPID`, `TestBuildContainerSpec_HostNetworkAndHostPID`, and the bootstrap-doc pin asserting kukeond carries both `HostNetwork=true` and `HostPID=true`.
- [x] Local smoke test per `CLAUDE.md`:
  - tear down kukeond cell, rebuild kuke + kukeond image, `docker save | ctr import`, `kuke init` succeeds
  - daemon-parity check: `kuke get realms` and `kuke get realms --no-daemon` produce identical output
  - issue repro path: `kuke apply -f docs/examples/hello-world.yaml` now succeeds (previously failed with `Statfs /proc/<pid>/ns/net: no such file or directory`)
  - kukeond container OCI spec inspection confirms no `pid` namespace entry — daemon shares host PID ns

Closes #105


eminwux:
**PR #106 Review — `fix: kukeond HostPID so in-process CNI plugin can resolve user-cell netns paths`**

Solid, narrowly-scoped fix. The new `HostPID` axis is a clean mirror of the existing `HostNetwork` plumbing — `ext.ContainerSpec` ↔ `intmodel.ContainerSpec` ↔ both `BuildContainerSpec` and `BuildRootContainerSpec` ↔ bootstrap doc — all pinned by tests. The judgment to skip a `cellWantsHostPIDRoot` analog checks out: `internal/controller/runner/start.go:267-271` constructs `namespacePaths` with only `Net/IPC/UTS`, so sibling containers don't share PID via `JoinContainerNamespaces`, and each container's `HostPID` decision genuinely stands alone. No correctness, secret-leak, or convention issues.

LGTM to merge.

PR #107, opened 2026-04-26
fix: persist kuke-system realm as Ready when its containerd namespace pre-exists

### LINKED ISSUE #94: init: kuke-system realm persists state=Failed and empty cgroupPath even though its cgroup is created on disk
## Repro
Reproduced on host `vmi2904685` against `main` (`5c0e472` at time of repro), using a freshly built `kukeon-local:dev` image:

1. Tear down any prior runtime (kill + delete the kukeond cell, purge both realms, delete CNI bridges, remove `/run/kukeon/*`).
2. `make kuke && ln -sf kuke kukeond`
3. `docker build -t kukeon-local:dev . && docker save kukeon-local:dev | sudo ctr -n kuke-system.kukeon.io images import -`
4. `sudo ./kuke init --kukeond-image docker.io/library/kukeon-local:dev`
5. `sudo ./kuke get realms` (and the same with `--no-daemon`).

`kuke init` exits 0 and prints `kukeond is ready`, but the kuke-system hierarchy line in its own report reads:

```
System hierarchy:
  - realm "kuke-system": created
  - containerd namespace "kuke-system.kukeon.io": already existed
  - realm cgroup: missing
```

(The default hierarchy on the same run prints `- realm cgroup: created` and is fine.)

## Observed
`sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` agree byte-for-byte (so this is *not* the daemon-vs-in-process bind-mount class of regression):

```
NAME         NAMESPACE              STATE   CGROUP
-----------  ---------------------  ------  ---------------
default      kukeon.io              Ready   /kukeon/default
kuke-system  kuke-system.kukeon.io  Failed  -
```

But the kernel cgroup is present on disk:

```
$ ls /sys/fs/cgroup/kukeon/
default  kuke-system  ...
```

So the realm cgroup *was* created. The metadata persisted by init disagrees — `/opt/kukeon/kuke-system/metadata.json`:

```json
{
  "metadata": {"name": "kuke-system", "labels": {"realm.kukeon.io": "kuke-system.kukeon.io"}},
  "spec":     {"namespace": "kuke-system.kukeon.io"},
  "status":   {"state": 4}
}
```

vs `/opt/kukeon/default/metadata.json` on the same run:

```json
{
  "status": {"state": 2, "cgroupPath": "/kukeon/default"}
}
```

(`RealmState` enum, `internal/modelhub/realm.go:53`: `Pending=0, Creating=1, Ready=2, Deleting=3, Failed=4, Unknown=5`.) So kuke-system is persisted as `Failed` with no `cgroupPath` despite the cgroup existing.

The user-visible consequence: `kuke get realms` reports the daemon's *own* home realm as `Failed`, which is misleading (the daemon is actually running fine inside it) and breaks any tooling that gates on `STATE=Ready`.

## Root cause
The asymmetry between the two realms in init's output points at the system-realm path. On a re-init, `containerd namespace "kuke-system.kukeon.io"` reports `already existed` while the default namespace reports `created` — the system namespace's pre-existence (e.g. left over from `ctr -n kuke-system.kukeon.io images import` in step 3) takes the realm down a different code path.

In `internal/controller/create_realm.go:86-89`:

```go
resultRealm, err = b.runner.CreateRealm(realm)
if err != nil && !errors.Is(err, errdefs.ErrNamespaceAlreadyExists) {
    return res, fmt.Errorf("%w: %w", errdefs.ErrCreateRealm, err)
}
```

When the containerd namespace already exists, `runner.CreateRealm` returns `ErrNamespaceAlreadyExists`, the controller swallows the error and continues, but `resultRealm` from the runner is whatever the runner returned at the point it aborted — likely missing the cgroup-path / state-Ready bookkeeping that the success path performs. Then `res.CgroupExistsPost = true` (line 117) is set unconditionally without an actual `ExistsCgroup` re-check, so the in-memory result looks fine even when the persisted metadata is half-written.

## Suggested fix
In the `ErrNamespaceAlreadyExists` branch of `controller.CreateRealm`, after the runner returns the swallowed error, re-resolve the realm via `runner.GetRealm` and re-run the cgroup-ensure/metadata-persist path so the on-disk metadata.json ends up with `state: Ready` and a populated `cgroupPath`. Equivalent fix in `runner.CreateRealm`: ensure cgroup creation and metadata persistence are *not* skipped on the `ErrNamespaceAlreadyExists` early-return. Add a regression test that runs `init` with the system realm's containerd namespace pre-created (mirroring the real `ctr images import` flow used to seed the kukeond image) and asserts the post-init metadata has `state=Ready` + non-empty `cgroupPath`.

eminwux:
## Summary
- `provisionNewRealm` (`internal/controller/runner/provision.go:49`) bailed with `state=Failed` and an empty `cgroupPath` whenever `createRealmContainerdNamespace` returned `ErrNamespaceAlreadyExists` — even though the controller layer (`internal/controller/create_realm.go:87`) already treats that error as benign for `kuke init`'s idempotent re-run path. The result was that on every fresh `kuke init` where `ctr -n kuke-system.kukeon.io images import` had pre-seeded the namespace (i.e., the standard documented bootstrap), the daemon's own home realm was persisted as `Failed`, which `kuke get realms` reported back to the user and which broke any tooling gating on `STATE=Ready`.
- Fix: in `provisionNewRealm`, treat `ErrNamespaceAlreadyExists` from `createRealmContainerdNamespace` the same way `ensureRealmContainerdNamespace` already treats existing namespaces — log and continue. The cgroup is then created and the realm transitions to `Ready` with `cgroupPath` populated, matching the per-issue suggested-fix shape.
- Adds an integration regression test (`TestCreateRealm_NamespacePreSeeded_PersistsReady`) that pre-creates the containerd namespace via a separate `ctr.Client` (mirroring the real `ctr -n <ns> images import` flow) and asserts the persisted metadata ends up `Ready` with non-empty `cgroupPath`.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` passes
- [x] Smoke test from `CLAUDE.md` (rebuild kukeon-local:dev, import into `kuke-system.kukeon.io`, `kuke init`, then `kuke get realms` vs `kuke get realms --no-daemon`) — both paths report the kuke-system realm as `Ready` with `/kukeon/kuke-system` cgroup, identical byte-for-byte.
- [x] Bug-trigger reproduction: with the containerd namespace pre-existing and `/opt/kukeon/kuke-system/metadata.json` deleted (the conditions that make `runner.GetRealm` take the provisionNewRealm branch), re-run `kuke init` — system hierarchy now reports `realm cgroup: created` (was `missing` before fix), and `metadata.json` persists `state: 2` (Ready) with `cgroupPath: "/kukeon/kuke-system"` (was `state: 4` (Failed) with no `cgroupPath` before fix).

Closes #94

### DIFF
--- a/internal/controller/runner/create_realm_test.go
+++ b/internal/controller/runner/create_realm_test.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/eminwux/kukeon/internal/cni"
 	"github.com/eminwux/kukeon/internal/controller/runner"
+	"github.com/eminwux/kukeon/internal/ctr"
 	"github.com/eminwux/kukeon/internal/metadata"
 	intmodel "github.com/eminwux/kukeon/internal/modelhub"
 	"github.com/eminwux/kukeon/internal/util/fs"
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #107 Review — `fix: persist kuke-system realm as Ready when its containerd namespace pre-exists`**

Tight, well-scoped fix. The bug analysis is dead-on: `provisionNewRealm` was unconditionally treating `ErrNamespaceAlreadyExists` as fatal even though the controller layer (`create_realm.go:87`) and the bootstrap path (`bootstrap.go:191`) already swallow it as benign — so the only observable effect of the daemon's pre-seeded `kuke-system.kukeon.io` namespace was a half-written `metadata.json` with `state=Failed` and an empty `cgroupPath`. The new logic mirrors `ensureRealmContainerdNamespace`'s "exists → continue" symmetry exactly. No security concerns; logs and error paths only carry namespace names.

A few small observations (all non-blocking):

- **Test isolation** (`internal/controller/runner/create_realm_test.go:758`): the new test creates a real cgroup via `createRealmCgroup` but doesn't `t.Cleanup` it from `/sys/fs/cgroup`. Same pattern as the surrounding tests in this file, so this is a pre-existing convention rather than a regression — flagging only because if you ever sweep cgroup test leakage, this one joins the existing crew.
- **Comment quality** (`internal/controller/runner/provision.go:55-59`): the new comment is exactly the right shape — explains the *why* (idempotent kuke init path, ctr image-import seeding), not the *what*. Good.

LGTM.

PR #109, opened 2026-04-26
fix: warn-and-skip empty ContainerdID in kill/stop to match delete/purge policy

### LINKED ISSUE #86: kuke kill/stop fail hard on cells with no ContainerdID; kuke delete/purge handle it gracefully
## Repro

Reproduced on `origin/main` at 6a9efb2.

```
# Apply a cell whose creation will fail past the cgroup step
# (e.g. unreachable image registry, or any failure that leaves
#  the cell in Pending with no containerd containers created).
sudo ./kuke apply -f docs/examples/hello-world.yaml
# … cell ends up Pending, container.ContainerdID empty.

sudo ./kuke kill cell hello-world --realm default --space default --stack default
```

## Observed

```
Error: failed to kill cell containers: container "web" has empty ContainerdID
```

`sudo ./kuke delete cell hello-world …` immediately afterward succeeds and warns-and-skips on the same empty ContainerdID — so the natural `kill && delete` cleanup workflow only works because of the inconsistency: `kill` errors out, the user has to know it's safe to ignore and run `delete` anyway.

## Root cause

The runner package handles cells with not-yet-created containerd containers inconsistently:

- `internal/controller/runner/delete_cell.go:117-126` — empty `ContainerdID` ⇒ `WarnContext("container has empty ContainerdID, skipping")` + `continue`.
- `internal/controller/runner/purge_cell.go:153` — same warn-and-skip behaviour.
- `internal/controller/runner/kill.go:121-123` and again at `:373` — empty `ContainerdID` ⇒ `return fmt.Errorf("container %q has empty ContainerdID", …)`.
- `internal/controller/runner/stop.go:137` and again at `:401` — same hard error.
- `internal/controller/runner/start.go:397, :590` — same hard error (less of an issue here since starting a never-created container legitimately can't proceed).

A cell can legitimately reach the runner with empty `ContainerdID` per-container — that's the post-condition of any apply that aborted between the cell-cgroup-creation step and the per-container create step (image pull failure, OCI spec rejection, network setup failure, etc.). `delete_cell` already encodes the right policy. `kill_cell` and `stop_cell` should match.

## Suggested fix

Replace the `return fmt.Errorf("container %q has empty ContainerdID", …)` in `kill.go` (lines 122 and 373) and `stop.go` (lines 137 and 401) with the same warn-and-skip + `continue` pattern that `delete_cell.go:117-126` uses. Keep the per-container loop going so the rest of the cell still gets killed/stopped.

Optional follow-up: factor the empty-ContainerdID skip into a small helper in `internal/controller/runner/helpers.go` so the four call sites (kill, stop, delete, purge) share a single decision. Out of scope for the minimal fix.

`start.go` should be left as-is: starting a container that was never created in containerd is a real error, not a recoverable skip.

eminwux:
## Summary
- `KillCell`/`StopCell` per-container loops bail out the whole cell with a hard error the first time they hit an empty `ContainerdID`. That state is reachable through any apply that aborts after cell-cgroup creation but before the per-container create step (image pull failure, OCI spec rejection, network setup failure). `delete_cell.go:117-126` and `purge_cell.go:153` already warn-and-skip that case, so the natural `kuke kill && kuke delete` cleanup only works because the user knows to ignore `kill`'s error and run `delete` anyway. Aligns kill/stop with the delete/purge policy: per-cell loops warn-and-`continue`; per-container `KillContainer`/`StopContainer` warn and return `nil` (successful no-op for a never-created container).
- `start.go` is intentionally left as-is per the issue: starting a container that was never created in containerd is a real error, not a recoverable skip.
- Helper extraction from the issue's optional follow-up is deferred — kept the change minimal and mirroring the existing `delete_cell.go` shape verbatim.

## Notable judgment calls
- For the single-container `KillContainer`/`StopContainer` path there's no loop to `continue` in, so I return `nil` after the warn. Matches the spirit of the suggested fix (treat it the way `delete_cell` does) and fits `kuke kill cell` followed by `kuke delete cell` becoming a single error-free workflow when the cell got stuck pre-create.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `gofmt -l` on changed files: clean
- [x] `make test` passes (full suite, including `internal/controller/runner` and `internal/controller`)
- [x] `go test -count=1 ./internal/controller/runner/... ./internal/controller/...` passes without cache

No unit-level regression test added: the changed branches sit after `ensureClientConnected` and `GetRealm` in functions that have no existing unit tests in this package (kill/stop runner code is currently exercised only via integration paths with real containerd, same as `delete_cell.go`'s identical warn-and-skip branch). Adding a fake-`ctr.Client` harness solely to exercise this branch would expand scope well past the minimal fix and is left for the helper-consolidation follow-up the issue notes.

Closes #86

### DIFF
--- a/internal/controller/runner/kill.go
+++ b/internal/controller/runner/kill.go
@@ -119,7 +119,13 @@ func (r *Exec) KillCell(cell intmodel.Cell) (intmodel.Cell, error) {
 		// Use ContainerdID from spec
 		containerID := containerSpec.ContainerdID
 		if containerID == "" {
-			return intmodel.Cell{}, fmt.Errorf("container %q has empty ContainerdID", containerSpec.ID)
+			r.logger.WarnContext(
+				r.ctx,
+				"container has empty ContainerdID, skipping",
+				"container",
+				containerSpec.ID,
+			)
+			continue
 		}
 
 		// Use container name with UUID for containerd operations
@@ -370,7 +376,13 @@ func (r *Exec) KillContainer(cell intmodel.Cell, containerID string) error {
 	// Use ContainerdID from spec
 	containerdID := foundContainerSpec.ContainerdID
 	if containerdID == "" {
-		return fmt.Errorf("container %q has empty ContainerdID", containerID)
+		r.logger.WarnContext(
+			r.ctx,
+			"container has empty ContainerdID, skipping",
+			"container",
+			containerID,
+		)
+		return nil
 	}
 
 	// Use containerd ID for containerd operations
--- a/internal/controller/runner/stop.go
+++ b/internal/controller/runner/stop.go
@@ -134,7 +134,13 @@ func (r *Exec) StopCell(cell intmodel.Cell) (intmodel.Cell, error) {
 		// Use ContainerdID from spec
 		containerID := containerSpec.ContainerdID
 		if containerID == "" {
-			return intmodel.Cell{}, fmt.Errorf("container %q has empty ContainerdID", containerSpec.ID)
+			r.logger.WarnContext(
+				r.ctx,
+				"container has empty ContainerdID, skipping",
+				"container",
+				containerSpec.ID,
+			)
+			continue
 		}
 
 		// Use container name with UUID for containerd operations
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #109 Review — `fix: warn-and-skip empty ContainerdID in kill/stop to match delete/purge policy`**

LGTM. The four sites mirror the existing `internal/controller/runner/delete_cell.go:117-126` warn-and-skip pattern verbatim (same logger call, same message string, same `continue`), which is exactly what issue #86 prescribed. The single-container `KillContainer`/`StopContainer` `return nil` is the right loop-less analog — a never-created container is a successful no-op for kill/stop, same idempotency contract as the delete path. `start.go` correctly left untouched per the issue. CI green, no merge conflicts, no sensitive data in the new warn lines.

The deferred test and helper-extraction call-outs in the description are reasonable: the identical branch in `delete_cell.go` is also exercised only via integration with real containerd, and the four-site consolidation is real follow-up work, not in-scope for this minimal fix.

No blocking items.

PR #110, opened 2026-04-26
fix: apply Volumes and security helpers in BuildRootContainerSpec

### LINKED ISSUE #100: runner: BuildRootContainerSpec silently drops volumes, user, capabilities, securityOpts, tmpfs, resources, secrets when a user-supplied root container is used
## Repro

Reproduced on `main` (462f83a) while working on #96. Configure a cell whose root container is a workload (not the auto-default busybox sleep) — set `cell.Spec.RootContainerID` to a container ID whose `ContainerSpec` carries fields beyond ID/Image/Command/Env/Privileged. For example, build a kukeond cell where kukeond itself is the root:

```go
&v1beta1.CellDoc{
    Spec: v1beta1.CellSpec{
        RootContainerID: "kukeond",
        Containers: []v1beta1.ContainerSpec{
            {
                ID:          "kukeond",
                Root:        true,
                Image:       image,
                Command:     "/bin/kukeond",
                Args:        args,
                Volumes:     volumes,            // ← silently dropped
                Privileged:  true,
                HostNetwork: true,
            },
        },
    },
}
```

Observed: the cell starts but the daemon cannot reach `/run/kukeon`, `/sys/fs/cgroup`, `/var/lib/containerd`, `/opt/kukeon`, `/run/containerd` — the bind mounts declared on the workload-as-root container never reach the OCI spec. Inspecting with `ctr container info <root-id>` confirms zero bind mounts on the root container's spec.

## Observed

`internal/ctr/container_utils.go::BuildRootContainerSpec` (the spec builder used by `runner.StartCell` for the cell's root container — `internal/controller/runner/start.go:213`) reads only:

- `ID`, `ContainerdID`, `Image`
- `Command`, `Args`, `Env`
- `Privileged`
- `HostNetwork` (added in #96)
- `CNIConfigPath`

It silently ignores everything else that `ContainerSpec` carries — `Volumes`, `User`, `ReadOnlyRootFilesystem`, `Capabilities`, `SecurityOpts`, `Tmpfs`, `Resources`, `Secrets`, `Attachable`. The non-root path (`BuildContainerSpec`) processes all of them via `buildBindMounts` and `securitySpecOpts`, but `BuildRootContainerSpec` never calls those helpers.

## Root cause

`BuildRootContainerSpec` was written for the auto-default root container (busybox sleep infinity), which doesn't need any of those fields, and was never extended when `RootContainerID` was promoted to a public `v1beta1.CellSpec` field. The two builders have drifted: `BuildContainerSpec` is the comprehensive translator, `BuildRootContainerSpec` is a sparse one. The result is a partial implementation of the public RootContainerID API: the runner *finds* the user-supplied root container correctly (`provision.go:1791-1808`) but then strips everything except a handful of fields when building its OCI spec.

## Suggested fix

Have `BuildRootContainerSpec` call into the same helpers `BuildContainerSpec` uses — `buildBindMounts(rootSpec.Volumes)`, `securitySpecOpts(rootSpec)`, plus the attachable hook if Root containers are ever expected to be attachable. Either share a common spec-builder routine and have `BuildRootContainerSpec` pass a `Root: true` flag for the label difference, or just append the missing helper outputs to `specOpts` in `BuildRootContainerSpec`. A targeted unit test in `internal/ctr/container_utils_test.go` should pin that a root container with `Volumes` produces a spec with the expected `OCI bind mounts`.

This is a latent bug today (`#96` works around it via host-network propagation onto the auto-default root rather than promoting kukeond to root), but any future cell that needs a real workload as the netns-owner — or any user-authored manifest that sets `rootContainerId` to a non-default container — will silently lose its bind mounts and security fields.

eminwux:
## Summary

- `BuildRootContainerSpec` silently dropped `Volumes`, `User`, `ReadOnlyRootFilesystem`, `Capabilities`, `SecurityOpts`, `Tmpfs`, and `Resources` because it never called the `buildBindMounts` and `securitySpecOpts` helpers that `BuildContainerSpec` uses. The two builders had drifted: only the auto-default busybox-sleep root container was a clean translation; any user-supplied root container (`CellSpec.RootContainerID`) lost its bind mounts and security/isolation fields.
- Append the existing helpers (`buildBindMounts(rootSpec.Volumes)` and `securitySpecOpts(rootSpec)`) inside `BuildRootContainerSpec` so user-supplied root containers translate the same fields a non-root container does. The auto-default path is unchanged because all of those input fields are zero-valued — `TestBuildRootContainerSpec_DefaultsUnaffected` pins that.

## Notable judgment calls

- **Append-helpers vs share-a-builder.** The issue offered both. I picked append-helpers because it keeps the diff minimal, leaves both builders independently readable, and matches the order in which `BuildContainerSpec` already calls them. A future merge into a shared routine is still an option; I didn't bundle it because the task is a focused bug fix.
- **`Attachable` deferred.** The issue lists it as silently dropped too, but enabling it on root containers requires plumbing `BuildOption ...` through `BuildRootContainerSpec`'s callers (`runner/start.go`, `runner/provision.go`) and the host-side `AttachableInjection` paths. The issue explicitly hedges with "if Root containers are ever expected to be attachable" — I don't think that decision is mine to make in a fix-up commit. Leaving it as a follow-up if/when a use case lands.
- **`Secrets` deferred.** `BuildContainerSpec` itself doesn't translate `Secrets` either; secrets are resolved by `CreateContainerFromSpec` (`internal/ctr/container.go:497`) before the spec builder runs. Wiring secrets into the root-container code path is a caller-side change in `runner/start.go` / `runner/provision.go`, not a builder-side change, so it's out of scope here.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/ctr/...` (new tests for Volumes, User+ReadonlyRootfs, Capabilities, SecurityOpts, Tmpfs, Resources, plus the defaults-unaffected guard)
- [x] CLAUDE.md smoke test: rebuild kuke + kukeond image, `kuke init --kukeond-image docker.io/library/kukeon-local:dev`, then `kuke get realms` vs `kuke get realms --no-daemon` → identical output.

Closes #100

### DIFF
--- a/internal/ctr/container_utils.go
+++ b/internal/ctr/container_utils.go
@@ -113,6 +113,16 @@ func BuildRootContainerSpec(
 		specOpts = append(specOpts, oci.WithHostNamespace(runtimespec.PIDNamespace))
 	}
 
+	// Bind mounts and security/isolation fields share the same translators as
+	// BuildContainerSpec so a user-supplied root container (RootContainerID)
+	// keeps its Volumes, User, Capabilities, SecurityOpts, Tmpfs, and
+	// Resources instead of having them silently dropped.
+	if mounts := buildBindMounts(rootSpec.Volumes); len(mounts) > 0 {
+		specOpts = append(specOpts, oci.WithMounts(mounts))
+	}
+
+	specOpts = append(specOpts, securitySpecOpts(rootSpec)...)
+
 	rootLabels := copyLabels(labels)
 	rootLabels[rootContainerLabelKey] = rootContainerLabelValue
 
--- a/internal/ctr/container_utils_test.go
+++ b/internal/ctr/container_utils_test.go
@@ -17,11 +17,35 @@
 package ctr_test
 
 import (
+	"context"
 	"testing"
 
 	ctr "github.com/eminwux/kukeon/internal/ctr"
 	intmodel "github.com/eminwux/kukeon/internal/modelhub"
+	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 )
+
+// applyRootBuiltSpec composes the SpecOpts produced by BuildRootContainerSpec
+// against an empty runtime spec so tests can assert on the resulting OCI fields
+// without touching containerd. Mirrors applyBuiltSpec in spec_security_test.go.
+func applyRootBuiltSpec(
+	t *testing.T,
+	in intmodel.ContainerSpec,
+	labels map[string]string,
+) *runtimespec.Spec {
+	t.Helper()
+	spec := &runtimespec.Spec{
+		Process: &runtimespec.Process{},
+		Linux:   &runtimespec.Linux{},
+	}
+	built := ctr.BuildRootContainerSpec(in, labels)
+	for _, opt := range built.SpecOpts {
+		if err := opt(context.Background(), nil, nil, spec); err != nil {
+			t.Fatalf("SpecOpts returned error: %v", err)
+		}
+	}
+	return spec
+}
 
 func TestDefaultRootContainerSpec(t *testing.T) {
 	tests := []struct {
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #110 Review — `fix: apply Volumes and security helpers in BuildRootContainerSpec`**

Solid, focused parity fix. The helpers (`buildBindMounts`, `securitySpecOpts`) are exactly the ones `BuildContainerSpec` already calls, and the append order in the new code (host-net → host-pid → bind mounts → security opts) mirrors `internal/ctr/spec.go:228-247`. The `Attachable` and `Secrets` deferrals are well-justified — `BuildRootContainerSpec` doesn't take `BuildOption ...`, and `Secrets` is resolved by the caller, not the spec builder. No correctness, sensitive-data, or scope-creep concerns. New tests cover each silently-dropped field plus a defaults-unaffected guard. Test plan checklist (incl. CLAUDE.md smoke-test parity check) is fully checked.

A few small items:

- **Dead code (non-blocking)** (`internal/ctr/container_utils_test.go:577-587`): `containsStringRoot` and `containsOnlyRoot` are byte-for-byte duplicates of `containsString` / `containsOnly` already in the same `ctr_test` package (`internal/ctr/spec_security_test.go:302,311`). The new test file is in `package ctr_test`, so the existing helpers are directly callable — no need for the `*Root`-suffixed copies.

LGTM to proceed — the duplicate test helpers are a non-blocking nit and can be cleaned up in a follow-up.

PR #112, opened 2026-04-26
fix: build and import local kukeond image in e2e harness so init does not pull non-existent registry tag from dirty trees

### LINKED ISSUE #98: e2e: TestKuke_Init_VerifyState tries to pull a non-existent image when run from a dirty tree
## Repro

On `origin/main` (462f83a) from any working tree that produces a `-dirty` `git describe` output (i.e., the test was run during local development with uncommitted edits):

```bash
make kuke           # builds kuke from a dirty tree → version baked in is e.g. v0.2.4-1-g462f83a-dirty
sudo env "PATH=/usr/local/go/bin:$PATH" HOME=$HOME make e2e
```

## Observed

```
=== NAME  TestKuke_Init_VerifyState
    e2e_kuke_test.go:248: running /home/.../kuke [--run-path tmp/e-... init] failed: exit status 1
        output:
        Creating space network 'default-default'
        Creating space network 'kuke-system-kukeon'
        time="…" level=info msg="fetch failed after status: 404 Not Found" host=ghcr.io
        Error: failed to create cell: failed to create container kukeon_kukeon_kukeond_kukeond:
          failed to create container from spec: failed to pull image
          ghcr.io/eminwux/kukeon:v0.2.4-1-g462f83a-dirty:
          failed to resolve reference "ghcr.io/eminwux/kukeon:v0.2.4-1-g462f83a-dirty":
          ghcr.io/eminwux/kukeon:v0.2.4-1-g462f83a-dirty: not found
--- FAIL: TestKuke_Init_VerifyState (1.78s)
```

Reproduces deterministically — anything that yields a version tag absent from the registry trips it.

## Root cause

`e2e/e2e_kuke_test.go:247` runs `kuke init` without passing `--kukeond-image`. The controller falls back to the default repo (`ghcr.io/eminwux/kukeon`) at the version baked into the binary by `Makefile:25-27`:

```
KUKEON_VERSION = $(shell git describe --tags --always --dirty --match 'v*')
LDFLAGS := … -X $(MODULE)/cmd/config.Version=$(KUKEON_VERSION) …
```

`git describe … --dirty` appends `-dirty` for any uncommitted edit, so the test resolves a tag the registry never received. The same will happen on any commit-since-tag (e.g. `v0.2.4-1-g462f83a`) until that exact SHA is published, which never happens for in-progress work.

The smoke-test path in CLAUDE.md sidesteps this by always passing `--kukeond-image docker.io/library/kukeon-local:dev`, so the failure only surfaces in the e2e suite.

## Suggested fix

Make the test pass an image that's guaranteed to exist locally. Options, in increasing order of test-rewrite cost:

1. Have the e2e harness build a deterministic-tagged image (e.g. `kukeon-e2e:test`) once per `go test ./e2e` invocation and pass `--kukeond-image kukeon-e2e:test` from `TestKuke_Init_VerifyState`. Mirrors what the Makefile smoke test already does manually.
2. Skip the test (or drop the kukeond cell assertion specifically) when `--kukeond-image` would resolve to a non-existent registry tag. Detect via `containerd images ls` or by short-circuiting on a `t.Skip` when `os.Getenv("KUKEOND_IMAGE")` is unset.
3. Add a Makefile target `e2e-with-image` that builds, loads, and exports `KUKEOND_IMAGE`, and have `TestKuke_Init_VerifyState` honor it (skip otherwise).

Option 1 keeps coverage; the others trade coverage for stability.

eminwux:
## Summary
- `TestKuke_Init_VerifyState` ran `kuke init` without `--kukeond-image`, so the daemon fell back to `ghcr.io/eminwux/kukeon:<git-describe>`. Any dirty or untagged dev tree resolves to a tag the registry never received (e.g. `v0.2.4-1-g462f83a-dirty`), and the test fails with `failed to resolve reference … not found`.
- `make test-e2e` now builds a deterministic-tagged `kukeon-local:e2e` image (mirroring the smoke-test recipe in `CLAUDE.md`) and exports `KUKEON_E2E_IMAGE` / `KUKEON_E2E_IMAGE_DOCKER_NAME` to the test process.
- The test reads those env vars, `docker save`s the image, `ctr -n kuke-system.kukeon.io images import`s it after Step 0 purge (which would otherwise wipe it), then passes the resolved ref via `--kukeond-image`. If the env vars or `docker`/`ctr` are unavailable (someone running `go test ./e2e` directly), the test skips with a hint rather than failing.

## Notable judgment calls
- Image name: `kukeon-local:e2e` (not the `kukeon-local:dev` used in the smoke-test) so a developer's interactive smoke-test image and the e2e harness's image don't trample each other.
- Kept the existing Step 0 purge — it cleans up residual state from prior runs. Image staging is moved to a new Step 0.5 so it lands after the purge wipes the namespace and before init needs it.
- Test does the `docker save | ctr import` itself rather than the Makefile pre-loading the image: the Step 0 purge of `realm kuke-system` deletes the namespace and its images, so a Makefile-time import wouldn't survive.

## Test plan
- [ ] `go build ./...` clean
- [ ] `go vet ./...` clean
- [ ] `make test` passes
- [ ] On a deliberately dirty tree (`touch foo && make kuke && make kuke`), run `sudo env "PATH=/usr/local/go/bin:\$PATH" HOME=\$HOME make e2e` — `TestKuke_Init_VerifyState` now passes instead of failing on a `ghcr.io/eminwux/kukeon:v…-dirty: not found` pull.
- [ ] `go test ./e2e -run TestKuke_Init_VerifyState` (no env, no Makefile) reports SKIP with the env-var hint rather than failing.

Closes #98

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,11 +68,24 @@ kill:
 test:
 	go test $(shell go list ./... | grep -v /e2e)
 
+# Tag the e2e suite uses to refer to the local kukeond image. The e2e harness
+# imports it into containerd's kuke-system namespace before running `kuke init`,
+# so the test does not depend on a published registry tag matching `git
+# describe`.
+KUKEON_E2E_IMAGE_DOCKER_NAME ?= kukeon-local:e2e
+KUKEON_E2E_IMAGE ?= docker.io/library/$(KUKEON_E2E_IMAGE_DOCKER_NAME)
+
 e2e: test-e2e
 .PHONY: test-e2e
 test-e2e: kuke
+	@echo "Building local kukeond image $(KUKEON_E2E_IMAGE_DOCKER_NAME) for e2e"
+	docker build --build-arg VERSION=v0.0.0-e2e -t $(KUKEON_E2E_IMAGE_DOCKER_NAME) .
 	@echo "Running e2e tests using binaries in project root"
-	HOME=$(HOME) PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$(PATH) E2E_BIN_DIR=$(CURDIR) go test -v ./e2e
+	HOME=$(HOME) PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$(PATH) \
+		E2E_BIN_DIR=$(CURDIR) \
+		KUKEON_E2E_IMAGE=$(KUKEON_E2E_IMAGE) \
+		KUKEON_E2E_IMAGE_DOCKER_NAME=$(KUKEON_E2E_IMAGE_DOCKER_NAME) \
+		go test -v ./e2e
 
 tag:
 	git tag -s v$(KUKEON_VERSION) -m "Release version $(KUKEON_VERSION)"
--- a/e2e/e2e_kuke_test.go
+++ b/e2e/e2e_kuke_test.go
@@ -243,8 +243,14 @@ func TestKuke_Init_VerifyState(t *testing.T) {
 		cleanupRealm(t, runPath, "default")
 	})
 
+	// Step 0.5: Stage the local kukeond image into containerd's kuke-system
+	// namespace. The Step 0 purge above wipes any prior image, and dev builds
+	// from a dirty/untagged tree resolve to a registry tag that doesn't exist,
+	// so init would fail on image pull without this.
+	kukeondImage := loadKukeondImageIntoContainerd(t)
+
 	// Step 1: Run init command
-	args := append(buildKukeRunPathArgs(runPath), "init")
+	args := append(buildKukeRunPathArgs(runPath), "init", "--kukeond-image", kukeondImage)
 	output := runReturningBinary(t, nil, kuke, args...)
 
 	// Step 2: Verify bootstrap report output contains expected content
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #112 Review — `fix: build and import local kukeond image in e2e harness so init does not pull non-existent registry tag from dirty trees`**

LGTM. The fix mirrors the smoke-test recipe in `CLAUDE.md` — build a deterministic-tagged image, `docker save | ctr images import` it into `kuke-system.kukeon.io` after the Step 0 purge, then pass it via `--kukeond-image`. The skip-when-env-missing path keeps `go test ./e2e` standalone-runnable. `consts.KukeSystemRealmNamespace`, the `ctr` const, and the `--kukeond-image` flag all check out against the working tree.

A few small items:

- **Test plan hygiene** (PR description): the bugfix-verification box (*"On a deliberately dirty tree, run … `TestKuke_Init_VerifyState` now passes"*) is unchecked. Since CI here only runs CodeQL/Analyze and never invokes `make e2e`, this is the only signal that the fix actually lands in the scenario the PR names — non-blocking, but worth running once and ticking before merge.
- **Distinct image tag rationale** (Makefile): tagging `kukeon-local:e2e` (separate from the smoke-test's `kukeon-local:dev`) to avoid trampling an interactive smoke-test image is a nice touch. Non-blocking.

LGTM to proceed.

PR #115, opened 2026-04-27
feat: drive kuke attach in-process via sbsh/pkg/attach

eminwux:
## Summary

- `kuke attach` now drives the interactive sbsh attach loop in-process via `github.com/eminwux/sbsh/pkg/attach.Run` (sbsh v0.8.0, `pkg/attach` from sbsh#155). The `syscall.Exec` hop into the on-host `sb` binary is gone — kuke is one less subprocess and one less binary to ship.
- Drops `--sb-binary`, `KUKE_ATTACH_SB_BINARY`, `resolveSbBinary`, the `execFn` / `MockExecKey` test plumbing, and their unit tests. Replaces them with a single `runFn` / `MockRunKey` seam that swaps in a fake `pkg/attach.Run` for the unit tests; behavioural assertions migrate from "exec'd sb attach with this argv" to "called pkg/attach.Run with this Options.SocketPath".
- `e2e/e2e_kuke_attach_test.go` no longer probes for `sb` on PATH (the in-container `sbsh` requirement is unchanged). Comments updated to reflect that the detach keystroke filter and PTY takeover now happen inside the kuke process.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — all unit packages green
- [x] `go test -v ./cmd/kuke/attach/...` — every attach unit test still passes against the new `pkg/attach.Run` seam
- [x] pre-commit (golangci-lint, go-mod-tidy, addlicense, gofmt) — green
- [ ] `make e2e TestKuke_AttachDetach_KeepsTaskRunning` — fails locally with `sbsh socket … did not appear within 10s` in step 3 (`waitForSocket`), **before** `kuke attach` is invoked. Verified the same failure reproduces on `origin/main` via a clean worktree, so it is a pre-existing host/sbsh setup issue unrelated to this refactor. Reviewer should rerun once the underlying socket-staging issue lands.

Closes #84

PR #117, opened 2026-04-27
fix: align attachable wrapper with renamed sbsh terminal flags

### LINKED ISSUE #116: e2e: TestKuke_AttachDetach_KeepsTaskRunning times out in waitForSocket on origin/main
## Repro

Reproduced on `origin/main` at commit `f4447f8` via clean worktree:

```bash
git worktree add /tmp/kukeon-main-check origin/main
cd /tmp/kukeon-main-check
go build -o kuke ./cmd/
sudo env PATH=/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
    HOME=$HOME E2E_BIN_DIR=$(pwd) \
    /usr/local/go/bin/go test -v -count=1 \
    -run TestKuke_AttachDetach_KeepsTaskRunning ./e2e
```

Host has `sbsh` on PATH (passes the existing `requireSbshHasPostMergeFlags` guard — `sbsh terminal --help` lists `--capture-file`).

## Observed

```
=== RUN   TestKuke_AttachDetach_KeepsTaskRunning
=== PAUSE TestKuke_AttachDetach_KeepsTaskRunning
=== CONT  TestKuke_AttachDetach_KeepsTaskRunning
    e2e_kuke_attach_test.go:173: sbsh socket "/tmp/kukeon-main-check/e2e/tmp/.../work/tty/socket" did not appear within 10s
--- FAIL: TestKuke_AttachDetach_KeepsTaskRunning (22.91s)
FAIL    github.com/eminwux/kukeon/e2e   22.930s
```

Failure lands on `waitForSocket` in `e2e/e2e_kuke_attach_test.go:171-173` — i.e. **before** `kuke attach` is even invoked. After `applyAttachableCell` returns Ready, the per-container sbsh control socket at `<runPath>/<realm>/<space>/<stack>/<cell>/work/tty/socket` never appears within the 10s grace.

## Root cause

Not yet pinpointed — this issue is filed to track the investigation. The symptom narrows the surface to:

- The in-container `work` task is started by `applyAttachableCell` (which exits successfully and reports the cell Ready), so the OCI bundle and runner contract appear fine, but
- The expected `tty/socket` inside the per-container `tty/` bind-mount target never gets bound by sbsh inside the container within 10s.

Candidates for where the bug lives:

1. The sbsh-terminal wrapper command kukeon injects as the container's argv (`internal/ctr`, post-#69 + sbsh#153). If a flag was added/renamed in a newer sbsh, `sbsh terminal` may be exiting before the bind, and the existing `requireSbshHasPostMergeFlags` guard only checks for `--capture-file`.
2. The per-container `tty/` bind-mount (#77) — the source/target mode or ownership may be regressing such that sbsh's `bind(2)` fails silently inside the container.
3. The `attachable-cell.yaml` fixture pulls `registry.eminwux.com/busybox:latest`; if the image is unpacked but lacks the entry sbsh expects, the wrapper could no-op.

## Suggested fix

Reproduce, then capture the in-container task's stderr (the runner currently swallows it once the bind fails) by redirecting via the wrapper or running `ctr -n <realm> task ls/exec` while the test sleeps. Once the actual sbsh failure mode is visible, fix the root cause and tighten `requireSbshHasPostMergeFlags` (or its equivalent) to assert the specific flag/version invariant the wrapper depends on, so a stale sbsh on PATH skips loudly instead of timing out at `waitForSocket`.

## Notes

Surfaced while validating PR #115 (kuke attach via pkg/attach). The PR's diff is purely client-side (`cmd/kuke/attach`), so the failure is unrelated to that change — both `origin/main` and the PR branch fail identically at the same step. The PR body defers e2e re-run to this fix.

eminwux:
## Summary

- The Attachable wrapper in `internal/ctr/attachable.go` injected `--terminal-socket` and `--terminal-logfile` into the workload's argv. Current `sbsh terminal` accepts `--socket` and `--log-file` instead — `--terminal-*` are unknown flags. The wrapped `sbsh terminal` exited with `unknown flag: --terminal-socket` before binding the control socket, so the per-container `tty/socket` inode never appeared on the host bind-mount source. `e2e/e2e_kuke_attach_test.go::waitForSocket` (10s grace) timed out without anything to attribute the failure to. Repro and root-cause analysis live in #116; verified by running the wrapped command directly under the host sbsh.
- The pre-flight `requireSbshHasPostMergeFlags` only checked for `--capture-file`, so a renamed sbsh on PATH passed the guard. Tightened it to assert every flag the wrapper depends on (`--socket`, `--capture-file`, `--log-file`); a host build that lacks any one of them now `t.Skip`s with a named cause instead of timing out 10s deep into the test.
- Unit-test expectations in `internal/ctr/attachable_test.go` updated to match the new wrapped argv prefix.

## Notable judgment calls

- Kept `requireSbshHasPostMergeFlags` as a `t.Skip` (not a `t.Fatal`). The function's whole point is to bail loudly when the host sbsh and the in-container wrapper drift apart; failing the suite would punish an operator running locally with a different sbsh build, which is a real expected case.
- This PR fixes the `waitForSocket` symptom #116 was filed for. After this PR, `TestKuke_AttachDetach_KeepsTaskRunning` advances past Step 3 and now fails at Step 4 with `Error: no terminal identifier provided; terminal name or ID must be specified` from the on-host `sb attach` invocation in `cmd/kuke/attach`. That is a separate sbsh-CLI evolution bug (the current `sb attach` requires `--name` or `--id` even when `--socket` is passed) on the very codepath PR #115 (open) is replacing with `pkg/attach.Run`; out of scope here.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` — all unit packages green
- [x] `go test ./internal/ctr/...` — both `TestBuildContainerSpec_AttachableTrue_*` cases assert the new wrapped argv prefix
- [x] pre-commit (`go fmt`, `go-mod-tidy`, `golangci-lint`, `addlicense`) — green
- [x] `sudo env "PATH=/usr/local/go/bin:\$PATH" HOME=\$HOME make e2e` — `TestKuke_AttachDetach_KeepsTaskRunning` no longer fails at `waitForSocket`; advances to Step 4. Other e2e failures (daemon-socket connect errors, `TestKuke_Init_VerifyState` "container already exists") reproduce identically on `origin/main` via `git worktree add /tmp/kukeon-main-check origin/main`, so they are pre-existing infra/state issues unrelated to this fix.

Closes #116

### DIFF
--- a/e2e/e2e_kuke_attach_test.go
+++ b/e2e/e2e_kuke_attach_test.go
@@ -90,21 +90,25 @@ func stageHostSbsh(t *testing.T, runPath string) {
 	}
 }
 
-// requireSbshHasPostMergeFlags ensures the host sbsh carries the flag set
-// the wrapper injected at #69 assumes (post sbsh#153, which added
-// `--capture-file` to the `sbsh terminal` subcommand). Without this guard a
-// stale sbsh on PATH manifests as `unknown flag: ...` deep inside the work
-// container's task, surfaced here only as a `waitForSocket` timeout — hard
-// to attribute. Skipping with a named cause saves the dig.
+// requireSbshHasPostMergeFlags ensures the host sbsh carries every flag the
+// in-container wrapper (`internal/ctr/attachable.go`) injects into the
+// workload's argv. Without this guard a stale or newer-renamed sbsh on PATH
+// manifests as `unknown flag: ...` deep inside the work container's task,
+// surfaced here only as a `waitForSocket` timeout — hard to attribute.
+// Skipping with a named cause saves the dig.
 func requireSbshHasPostMergeFlags(t *testing.T, hostSbsh string) {
 	t.Helper()
 	out, err := exec.Command(hostSbsh, "terminal", "--help").CombinedOutput()
 	if err != nil {
 		t.Fatalf("sbsh terminal --help failed: %v\noutput:\n%s", err, out)
 	}
-	if !strings.Contains(string(out), "--capture-file") {
-		t.Skipf("host sbsh %q lacks `sbsh terminal --capture-file` (sbsh#153); "+
-			"upgrade sbsh on PATH and retry", hostSbsh)
+	help := string(out)
+	for _, flag := range []string{"--socket", "--capture-file", "--log-file"} {
+		if !strings.Contains(help, flag) {
+			t.Skipf("host sbsh %q lacks `sbsh terminal %s`; "+
+				"upgrade or downgrade sbsh on PATH to a build whose `terminal` "+
+				"subcommand carries every wrapper flag and retry", hostSbsh, flag)
+		}
 	}
 }
 
--- a/internal/ctr/attachable.go
+++ b/internal/ctr/attachable.go
@@ -122,9 +122,9 @@ func withAttachableArgsWrap() oci.SpecOpts {
 			AttachableBinaryPath,
 			AttachableSubcommand,
 			"--run-path", AttachableTTYDir,
-			"--terminal-socket", AttachableSocketPath,
+			"--socket", AttachableSocketPath,
 			"--capture-file", AttachableCapturePath,
-			"--terminal-logfile", AttachableLogfilePath,
+			"--log-file", AttachableLogfilePath,
 			"--",
 		}
 		s.Process.Args = append(wrapped, original...)
--- a/internal/ctr/attachable_test.go
+++ b/internal/ctr/attachable_test.go
@@ -192,9 +192,9 @@ func TestBuildContainerSpec_AttachableTrue_MountsAndArgsWrap(t *testing.T) {
 				ctr.AttachableBinaryPath,
 				ctr.AttachableSubcommand,
 				"--run-path", ctr.AttachableTTYDir,
-				"--terminal-socket", ctr.AttachableSocketPath,
+				"--socket", ctr.AttachableSocketPath,
 				"--capture-file", ctr.AttachableCapturePath,
-				"--terminal-logfile", ctr.AttachableLogfilePath,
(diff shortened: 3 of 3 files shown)


eminwux:
**PR #117 Review — `fix: align attachable wrapper with renamed sbsh terminal flags`**

Tight, focused fix. Renames `--terminal-socket` → `--socket` and `--terminal-logfile` → `--log-file` in the in-container wrapper to match the current `sbsh terminal` flag set, with matching unit-test updates and a tightened pre-flight that fails-fast with a named cause instead of timing out 10s deep into `waitForSocket`. Repro and root-cause are well-documented in #116. No sensitive-data leaks, no dead code added.

A couple of small notes, both non-blocking:

- **Pre-flight completeness** (`e2e/e2e_kuke_attach_test.go:107-114`): the new check verifies `--socket`, `--capture-file`, `--log-file` — three of the four flags the wrapper injects. `--run-path` (line 124 of `attachable.go`) is the fourth, and the function's own doc comment now reads "every flag the in-container wrapper injects into the workload's argv". Either add `--run-path` to the slice, or scope the comment to "every renameable flag" — the current pair drifts from the claim. Pure future-proofing; I don't expect `--run-path` to rename independently of these three.
- **Out-of-scope follow-on**: nice catch flagging that the next failure (`no terminal identifier provided` from `sb attach`) is on the codepath PR #115 is replacing. Just noting that those two PRs interact: this one unblocks #115's e2e end-to-end story, and #115 sidesteps the "missing terminal identifier" issue entirely. Order them however is convenient.

LGTM to proceed.

PR #122, opened 2026-04-27
fix: tighten ensureCNIStateDir mode to 0o750 to match sibling CNI dirs

### LINKED ISSUE #101: controller: align ensureCNIStateDir mode with bootstrapCNI (0o755 → 0o750)
Surfaced during review of #97.

`internal/controller/bootstrap.go:512` creates `/var/lib/cni` with mode `0o755`, but the sibling CNI dirs created by `internal/cni/bootstrap.go:52,59,65` (`/opt/cni/net.d`, `/opt/cni/cache`, `/opt/cni/bin`) all use `0o750`. Aligning to `0o750` keeps host-local IPAM state consistent with the rest of the CNI tree and avoids leaving it world-readable.

Suggested fix: change `0o755` → `0o750` in `ensureCNIStateDir`.

eminwux:
## Summary
- `internal/controller/bootstrap.go:529` created `/var/lib/cni` with `0o755`, but the sibling CNI dirs in `internal/cni/bootstrap.go:53,59,65` (`/opt/cni/net.d`, `/opt/cni/cache`, `/opt/cni/bin`) all use `0o750`. Aligning keeps host-local IPAM state consistent with the rest of the CNI tree and avoids leaving it world-readable.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test`

Closes #101

### DIFF
--- a/internal/controller/bootstrap.go
+++ b/internal/controller/bootstrap.go
@@ -526,7 +526,7 @@ func ensureSocketDir(socketPath string) error {
 // other CNI mounts (/opt/cni/net.d and /opt/cni/cache) are already created by
 // the earlier CNI bootstrap step.
 func ensureCNIStateDir() error {
-	if err := os.MkdirAll("/var/lib/cni", 0o755); err != nil {
+	if err := os.MkdirAll("/var/lib/cni", 0o750); err != nil {
 		return fmt.Errorf("create CNI state dir %q: %w", "/var/lib/cni", err)
 	}
 	return nil


eminwux:
**PR #122 Review — `fix: tighten ensureCNIStateDir mode to 0o750 to match sibling CNI dirs`**

Single-line mode tightening, well-scoped. Confirmed `internal/cni/bootstrap.go:53,59,65` all use `0o750` for the sibling `/opt/cni/{net.d,cache,bin}` dirs, and `ensureSocketDir` at `internal/controller/bootstrap.go:516` matches. No security concerns, no dead code, no convention drift.

One non-blocking observation:

- **Behavior** (`internal/controller/bootstrap.go:529`): `os.MkdirAll` is a no-op when the path already exists — it doesn't `chmod` an existing dir. So on systems where `/var/lib/cni` already sits at `0o755` (from a prior install, or because another CNI installer created it), this fix won't tighten the mode in place. Fresh installs are correctly handled. If you want idempotent tightening on upgrades, a follow-up `os.Chmod("/var/lib/cni", 0o750)` after the `MkdirAll` would close that gap. Not blocking — outside the scope of #101's "match sibling dirs" framing.

LGTM.

PR #124, opened 2026-04-27
build: verify cni-plugins tarball checksum before extracting

### LINKED ISSUE #102: Dockerfile: verify cni-plugins tarball checksum before extracting
Surfaced during review of #97.

`Dockerfile:34` downloads the upstream cni-plugins tarball with `curl … | tar -xz` and no checksum verification. The upstream release publishes a sibling `cni-plugins-${TARGETOS}-${TARGETARCH}-${VERSION}.tgz.sha256`; fetching it and running `sha256sum -c` before extracting hardens the build against a compromised mirror or transient corruption.

Suggested fix: download both files, run `sha256sum -c …sha256` in the cni-plugins build stage, then extract.

eminwux:
## Summary
- The `cni-plugins` build stage in `Dockerfile:33-35` previously piped the upstream tarball straight from `curl` into `tar -xz` with no checksum verification, leaving the daemon image vulnerable to a compromised mirror or a transient download corruption.
- Download both the `.tgz` and the sibling `.tgz.sha256` published in the upstream `containernetworking/plugins` release, run `sha256sum -c` against the tarball before extraction, and clean up the temp files in the same `RUN` so no extra layer remains.
- Hash extraction uses `awk '{print $1}'` plus a re-emitted `sha256sum -c -` line so the verification is robust to either format the upstream may publish (`<hash>` alone or `<hash>  <filename>`).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test`
- [x] `docker build --build-arg VERSION=v0.0.0-dev -t kukeon-checksum-test:dev .` — new step prints `cni-plugins-linux-amd64-v1.5.1.tgz: OK` and the build completes.

Closes #102

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,8 +31,13 @@ RUN DEBIAN_FRONTEND=noninteractive apt update \
  && rm -rf /var/lib/apt/lists/*
 
 RUN mkdir -p /opt/cni/bin \
- && curl -fsSL "https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGINS_VERSION}/cni-plugins-${TARGETOS}-${TARGETARCH}-${CNI_PLUGINS_VERSION}.tgz" \
-    | tar -xz -C /opt/cni/bin
+ && CNI_PLUGINS_TGZ="cni-plugins-${TARGETOS}-${TARGETARCH}-${CNI_PLUGINS_VERSION}.tgz" \
+ && CNI_PLUGINS_URL="https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGINS_VERSION}/${CNI_PLUGINS_TGZ}" \
+ && curl -fsSL -o "/tmp/${CNI_PLUGINS_TGZ}" "${CNI_PLUGINS_URL}" \
+ && curl -fsSL -o "/tmp/${CNI_PLUGINS_TGZ}.sha256" "${CNI_PLUGINS_URL}.sha256" \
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #124 Review — `build: verify cni-plugins tarball checksum before extracting`**

LGTM. Closes #102 cleanly: the upstream `.tgz.sha256` sidecar is fetched alongside the tarball, verified with `sha256sum -c` before `tar -xz`, and both temp files are cleaned up in the same `RUN` so no extra image layer leaks. The `&&` chain guarantees that a checksum mismatch aborts before extraction.

A few small observations (all non-blocking):

- **Correctness** (`Dockerfile:38`): the `awk '{print $1}'` + re-emitted-line pattern handles both upstream formats (`<hash>` alone or `<hash>  <filename>`) — good. The two-space separator in the echoed line is the format `sha256sum -c -` expects.
- **Hygiene** (`Dockerfile:33-40`): keeping the download, verify, extract, and cleanup in one `RUN` (vs. splitting) is the right call — leftover `/tmp/*.tgz` would otherwise persist as an intermediate-stage layer even though only `/opt/cni/bin` is `COPY --from=`'d into the final image.
- **Test plan**: all four boxes checked, including a `docker build` that exercises the new step. Daemon-parity smoke test isn't required here — since sha256 verification proves the extracted binaries are bit-identical to before, the daemon's runtime view of `/opt/cni/bin` is unchanged.

PR #126, opened 2026-04-27
fix: reject cell hostNetwork peer with non-host explicit rootContainerId

### LINKED ISSUE #103: runner: propagate HostNetwork (or reject) when cell uses explicit rootContainerId
Surfaced during review of #99.

`internal/controller/runner/provision.go:1820` propagates `HostNetwork` from any cell container to the auto-default root container, but the symmetric path is missing for the explicit `rootContainerId` branch (`provision.go:1793-1820`). A cell defined as:

```yaml
containers:
  - id: c1
    hostNetwork: true
  - id: c2
rootContainerId: c2
```

would silently get per-container netns for the entire cell — c1's `hostNetwork: true` is ignored because c2 owns the netns and c1 joins it via `JoinContainerNamespaces`. No manifest exercises this combination today, but it's a foot-gun for future host-network workloads.

Two reasonable fixes:
1. Reject the combination at validation time ("if any container has `hostNetwork: true`, the explicit root container must also have it").
2. Apply the same propagation in the explicit-root branch — set `rootSpec.HostNetwork = cellWantsHostNetworkRoot(cell)` after the lookup too.

Option 1 is more honest about the constraint; option 2 matches the auto-default path's "single host-network workload puts the whole cell on the host" semantics.

eminwux:
## Summary
- `internal/controller/runner/provision.go:1798-1813` previously accepted a cell with `rootContainerId` pointing at a non-host-network root even when peer containers had `hostNetwork: true`. Because peers join the root's netns via `JoinContainerNamespaces`, the peer's host-network intent would be silently dropped — the foot-gun called out in #103.
- The auto-default-root branch (`provision.go:1825-1831`) already propagates host-network onto its generated root via `cellWantsHostNetworkRoot`, but the explicit-root branch reads the user's spec verbatim, so the symmetric path was missing.
- Picked **option 1 (validation reject)** from the issue rather than option 2 (silent propagation): the explicit-root case has user-authored container specs, and silently flipping `c2.HostNetwork` from false to true overrides what the user wrote. Lower blast-radius default — surface the misalignment via a sentinel error, force the user to align the spec.

## Implementation
- New sentinel `ErrExplicitRootHostNetworkMismatch` in `internal/errdefs/errdefs.go` per the project convention against bare `errors.New` in package code for reusable concepts.
- New helper `validateExplicitRootHostNetwork(cell, rootSpec)` next to `cellWantsHostNetworkRoot` in `internal/controller/runner/start.go`.
- Helper called from `ensureCellRootContainerSpec` after the explicit-root lookup in `provision.go`.
- Five-case table-driven unit test in `start_test.go` covering: no-explicit-root pass-through, explicit-root with no peer host-network, explicit-host-network root with default peers, fully-aligned host-network root + peer, and the rejected misalignment.

## Test plan
- [x] \`go build ./...\`
- [x] \`go vet ./...\`
- [x] \`make test\`
- [x] \`go vet ./e2e/...\` + \`go test -count=1 -run NONE ./e2e/...\` (e2e package compiles)

(Smoke test \`kuke init\` not run — change is confined to cell-provision validation; \`/opt/kukeon\` layout and the daemon-parity \`kuke get realms\` views are untouched.)

Closes #103

### DIFF
--- a/internal/controller/runner/provision.go
+++ b/internal/controller/runner/provision.go
@@ -1811,6 +1811,11 @@ func (r *Exec) ensureCellRootContainerSpec(cell intmodel.Cell) (intmodel.Contain
 				cell.Spec.RootContainerID,
 			)
 		}
+		// Reject the misalignment instead of silently overriding the
+		// user-provided root spec — see validateExplicitRootHostNetwork.
+		if err := validateExplicitRootHostNetwork(cell, rootSpec); err != nil {
+			return intmodel.ContainerSpec{}, err
+		}
 	} else {
 		// Create default root container spec
 		// Pass containerdID to set ContainerdID field, ID will be set to "root"
--- a/internal/controller/runner/start.go
+++ b/internal/controller/runner/start.go
@@ -54,6 +54,28 @@ func cellWantsHostNetworkRoot(cell intmodel.Cell) bool {
 	return false
 }
 
+// validateExplicitRootHostNetwork enforces, for cells using explicit
+// rootContainerId, that the chosen root has HostNetwork=true whenever any
+// peer container has HostNetwork=true. The auto-default-root branch in
+// ensureCellRootContainerSpec already propagates host-network onto the
+// generated root, but the explicit-root branch reads the user's spec
+// verbatim — so a peer with HostNetwork=true alongside an explicit non-host
+// root would silently lose its host-network intent (peers join the root's
+// netns via JoinContainerNamespaces).
+func validateExplicitRootHostNetwork(cell intmodel.Cell, rootSpec intmodel.ContainerSpec) error {
+	if cell.Spec.RootContainerID == "" {
+		return nil
+	}
+	if cellWantsHostNetworkRoot(cell) && !rootSpec.HostNetwork {
+		return fmt.Errorf(
+			"%w: rootContainerId %q",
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #126 Review — `fix: reject cell hostNetwork peer with non-host explicit rootContainerId`**

Targeted fix for the foot-gun called out in #103. The asymmetry between the auto-default-root branch (`provision.go:1814-1832`, which already propagates host-network via `cellWantsHostNetworkRoot`) and the explicit-root branch is exactly what this guard plugs. Picking option 1 (validation reject) over option 2 (silent propagation) is the right call when the user authored the root spec — overriding `HostNetwork=false` to `true` behind the user's back would surprise more than the rejection. Sentinel/`%w`/test-coverage/colocation all clean. No correctness, leak, or convention concerns.

A small non-blocking nit:

- **Defensive duplication** (`start.go` `validateExplicitRootHostNetwork`): the helper's early `if cell.Spec.RootContainerID == "" { return nil }` is unreachable from the only caller (`provision.go:1798-1818` already gates on the explicit-root branch). It does make the helper safe to call from anywhere and is exercised by the first test case, so harmless — flagging it just so the next caller doesn't assume the guard is load-bearing.

LGTM to proceed.

PR #129, opened 2026-04-27
fix: include hostNetwork value in CNI-skip log line

### LINKED ISSUE #104: runner: include HostNetwork bit in CNI-skip log line
Surfaced during review of #99.

`internal/controller/runner/start.go:281-287` logs "skipping CNI attach for host-network root container" with `id`, `space`, `realm`, and `pid`, but does not include the `HostNetwork` value from `rootContainerSpec`. Adding `"hostNetwork", true` (or the actual value) makes the log self-documenting if the gate predicate `rootContainerWantsCNI` later grows additional reasons to skip CNI — at that point, "skipping for host-network" would no longer be the only explanation, and a grep through logs after an incident wouldn't be ambiguous.

Trivial — one extra pair of fields appended to `skipFields`.

eminwux:
## Summary
- Append `hostNetwork=<bool>` to the `skipFields` slice on the CNI-skip branch in `internal/controller/runner/start.go` so the log line is self-documenting if `rootContainerWantsCNI` later grows additional skip reasons.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test`
- [ ] `make e2e` — not run; this is a one-field log addition with no behavioral change, so e2e adds no signal beyond the unit suite.
- [ ] `kuke init` smoke / daemon-parity check — not run; the change does not touch the build path, daemon serving logic, or anything the daemon reads from `/opt/kukeon`.

Closes #104

### DIFF
--- a/internal/controller/runner/start.go
+++ b/internal/controller/runner/start.go
@@ -295,7 +295,7 @@ func (r *Exec) StartCell(cell intmodel.Cell) (intmodel.Cell, error) {
 	// dance for them.
 	if !rootContainerWantsCNI(rootContainerSpec) {
 		skipFields := appendCellLogFields([]any{"id", containerID}, cellID, cellName)
-		skipFields = append(skipFields, "space", spaceID, "realm", realmID, "pid", rootPID)
+		skipFields = append(skipFields, "space", spaceID, "realm", realmID, "pid", rootPID, "hostNetwork", rootContainerSpec.HostNetwork)
 		r.logger.InfoContext(
 			r.ctx,
 			"skipping CNI attach for host-network root container",


eminwux:
**PR #129 Review — `fix: include hostNetwork value in CNI-skip log line`**

One-field append to the existing `skipFields` slice, matching exactly what issue #104 asked for. Currently `HostNetwork` is always `true` on this branch (the gate is `!rootContainerWantsCNI` ⇒ `!HostNetwork`), so the value is redundant *today*, but the rationale in #104 — future-proofing the log if `rootContainerWantsCNI` grows additional skip reasons — is sound. No correctness concerns, no leak risk (the bool is non-sensitive), no dead code. `rootContainerSpec.HostNetwork` resolves to `internal/modelhub/container.go:49`.

The unchecked `make e2e` and `kuke init` smoke boxes are fine — the change is purely an observability tweak in the CNI-skip log path, and the daemon-parity guard in CLAUDE.md only triggers for changes the daemon reads from `/opt/kukeon`, which this is not.

LGTM.

PR #130, opened 2026-04-27
fix: chmod /var/lib/cni after MkdirAll so mode tightening applies on upgrade

### LINKED ISSUE #123: controller/bootstrap: chmod /var/lib/cni after MkdirAll so mode tightening applies on upgrade
Surfaced during review of #122.

`internal/controller/bootstrap.go:529` (`ensureCNIStateDir`) now calls `os.MkdirAll("/var/lib/cni", 0o750)`. On fresh installs this creates the dir at the intended mode, but `MkdirAll` is a no-op when the path already exists — it does not adjust the mode of an existing directory. So on systems where `/var/lib/cni` already sits at `0o755` (from a prior install, or because another CNI installer created it), the fix in #122 leaves the mode unchanged.

Suggested follow-up: add `os.Chmod("/var/lib/cni", 0o750)` after the `MkdirAll` call so the tightening is idempotent across upgrades, matching what a cleanly-bootstrapped host gets.

eminwux:
## Summary
- `ensureCNIStateDir` calls `os.MkdirAll("/var/lib/cni", 0o750)` (added in #122). `MkdirAll` is a no-op when the directory already exists, so on hosts where `/var/lib/cni` was previously created at `0o755` (prior install, or another CNI installer), the mode tightening from #122 silently doesn't apply on upgrade.
- Add `os.Chmod("/var/lib/cni", 0o750)` after the `MkdirAll` so the tightening is idempotent across upgrades, matching what a cleanly-bootstrapped host already gets.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test`
- [ ] `make e2e` — not run; this is a one-line idempotence fix for a chmod-on-upgrade gap and the e2e harness bootstraps from a clean host where `MkdirAll` already lands the mode at `0o750`. e2e adds no signal beyond the unit suite for this case.
- [ ] `kuke init` smoke / daemon-parity check — not run; the change does not touch the build path, daemon serving logic, the bind-mount layout, or anything the daemon reads from `/opt/kukeon`. It only adds a chmod on `/var/lib/cni` (a host path the bootstrap creates before the daemon starts).

Closes #123

### DIFF
--- a/internal/controller/bootstrap.go
+++ b/internal/controller/bootstrap.go
@@ -529,6 +529,13 @@ func ensureCNIStateDir() error {
 	if err := os.MkdirAll("/var/lib/cni", 0o750); err != nil {
 		return fmt.Errorf("create CNI state dir %q: %w", "/var/lib/cni", err)
 	}
+	// MkdirAll is a no-op when the dir already exists, so on upgrade from a
+	// prior install (or co-existing CNI installer) that left /var/lib/cni at
+	// 0o755, the mode would stay loose. Chmod here makes the tightening
+	// idempotent across upgrades.
+	if err := os.Chmod("/var/lib/cni", 0o750); err != nil {
+		return fmt.Errorf("chmod CNI state dir %q: %w", "/var/lib/cni", err)
+	}
 	return nil
 }
 


eminwux:
**PR #130 Review — `fix: chmod /var/lib/cni after MkdirAll so mode tightening applies on upgrade`**

Clean targeted fix that lands exactly what #123 spelled out. `os.MkdirAll` is a no-op on an existing dir and doesn't adjust the mode, so the post-`MkdirAll` `os.Chmod` is what makes #122's `0o750` tightening idempotent on upgrade. Error wrapping preserves the cause via `%w`. No new identifiers, no sensitive-data exposure in the error string (well-known path).

Test-plan justification holds — `e2e` and the `kuke init` smoke both bootstrap from a clean host where `MkdirAll` already lands the dir at `0o750`, so neither would have exercised the upgrade-from-`0o755` path the fix targets. The daemon-parity check guards bind-mount regressions, and the bind-mount layout is unchanged.

LGTM.

PR #132, opened 2026-04-27
fix: tear down CNI bridge link on space/realm delete and purge

### LINKED ISSUE #87: network: bridge leak on space/realm delete and purge (phase 1)
## Repro
Reproduced on host vmi2904685 against a daemon built from `ghcr.io/eminwux/kukeon:v0.2.2` (controller code in `internal/cni`). After many `kuke create space` / `kuke delete space` cycles over time:

1. `sudo kuke apply -f docs/examples/hello-world.yaml`
2. Observe failure: kukeond cannot pull the image (DNS times out from inside its netns).
3. From inside the kukeond container's netns, `ping 10.88.0.1` (its own gateway) returns 100% loss — even though the veth peer is correctly attached to an UP bridge with that IP.

## Observed
`ip -4 -br addr` on the host shows ~80 leftover bridges, all claiming `10.88.0.1/16`. The vast majority are `DOWN`/`NO-CARRIER` from spaces that were deleted long ago:

```
e-r-fd-e-sp-a2   DOWN   10.88.0.1/16
e-r-04-e-sp-7c   DOWN   10.88.0.1/16
default-default  DOWN   10.88.0.1/16
... (~80 lines, all 10.88.0.1/16) ...
kuke-s-082023b5  UP     10.88.0.1/16   # the only live one — kukeond's bridge
```

`ip route get 10.88.0.1` returns `local 10.88.0.1 dev lo src 10.88.0.1` — the kernel treats the IP as local on `lo` because of the duplicates, so reply packets never traverse the right bridge.

Application-level error tail:
```
failed to pull image docker.io/library/busybox:latest: ... lookup registry-1.docker.io on 1.0.0.1:53: read udp 10.88.0.2:44272->1.0.0.1:53: i/o timeout
```

## Root cause (this phase)
The space/realm teardown paths do not delete the Linux bridge that `internal/cni/network.go:CreateNetwork` created. Only one of the four paths touches CNI at all today, and even that one stops short of the kernel link:

| Path | Bridge cleanup today |
|---|---|
| `delete space` | calls `cni.Manager.DeleteNetwork`, which only removes the `*.conflist` file (`internal/cni/network.go:155`). The `ip link` is left behind. |
| `delete realm` | nothing CNI-related. |
| `purge space` | nothing CNI-related. |
| `purge realm` | `runner.purgeCNIForNetwork` clears `/var/lib/cni/networks/<net>/` state and cache files (`internal/controller/runner/helpers.go:194`), but does not touch the bridge link either. |

The duplicate-subnet half of the original analysis (`defaultSubnetCIDR = "10.88.0.0/16"` shared across every space) is a separate, larger fix tracked in the follow-up below.

## Proposal
Lift bridge teardown into a shared helper and call it from all four runner paths:

- Read the bridge name via `cni.Manager.ReadBridgeName(conflistPath)` (already exists in `internal/cni/network.go`) **before** the conflist is removed.
- After the conflist is gone, `ip link delete <bridge> type bridge` via netlink. Idempotent — already-absent bridges and missing conflists must not error.
- Invoke from `delete_space`, `delete_realm`, `purge_space`, and `purge_realm` runner paths.

## Acceptance criteria
- [ ] New shared helper (in `internal/cni` or `internal/controller/runner`) that takes a conflist path, reads the bridge name, and deletes the kernel link idempotently.
- [ ] Helper invoked from all four runner paths: `delete_space`, `delete_realm`, `purge_space`, `purge_realm`.
- [ ] Unit tests covering: bridge present → deleted; bridge already absent → no error; conflist missing → no error.
- [ ] After `kuke purge realm <r> --cascade --force --no-daemon`, `ip -4 -br addr` shows no `<realm>-<space>` bridge remnants.
- [ ] The workaround `ip link delete` loop below is no longer needed to recover kukeond connectivity after repeated create/delete cycles.

## Workaround until fixed
`ip -4 -br addr | awk '/10\.88\.0\.1\/16/ && /DOWN/ {print $1}' | xargs -I{} sudo ip link delete {} type bridge` clears the stale bridges and restores kukeond connectivity immediately.

## Out-of-scope follow-ups
- Phase 2 (per-space subnet allocator): #131.

## Notes
This is phase 1 of two. Phase 1 alone stops the *accumulation* that makes the symptom in the repro compound; phase 2 (#131) removes the duplicate-IP class entirely so multiple spaces can coexist without ARP wars.

eminwux:
## Summary
- Adds `cni.Manager.TeardownNetwork(ctx, runner, networkName, configPath)` (in `internal/cni/bridge.go`) — reads the bridge name from the conflist BEFORE removing the conflist, then deletes the kernel bridge link via `ip link delete`. Each step is idempotent: missing conflist, no bridge plugin, empty bridge field, and already-absent bridge are all treated as success.
- Wires the helper through two runner-layer wrappers in `helpers.go`: `teardownSpaceCNI` (single space) and `teardownRealmCNI` (enumerates `<RunPath>/<realm>/*/network.conflist` and tears each down).
- Hooks both into all four runner paths called out in the issue: `delete_space`, `delete_realm`, `purge_space`, `purge_realm`. The bridge teardown runs before the existing IPAM/cache cleanup so a partial failure leaves enough state to retry from.

## Notable judgment calls
- **Bridge enumeration source.** Issue suggested `r.cniConf.CniConfigDir` for realm-level enumeration, but kukeon writes per-space conflists to `<RunPath>/<realm>/<space>/network.conflist` (via `fs.SpaceNetworkConfigPath`), not to the global CNI config dir. `teardownRealmCNI` enumerates the run path so it actually finds the conflists in both daemon and `--no-daemon` modes.
- **`os/exec` over a netlink Go library.** Mirrors the iptables enforcer pattern already in `internal/netpolicy/enforcer.go` rather than introducing `vishvananda/netlink` as a new direct dep. The `BridgeRunner` interface keeps the unit tests off the host's `ip` binary.
- **Idempotent error matching.** `IPBridgeRunner.DeleteBridge` treats the literal iproute2 error `Cannot find device` as success — verified against `iproute2-6.4.0` on the smoke-test host. Other failures still propagate.

## Test plan
- [x] `go test ./internal/cni/...` — six new unit tests cover the issue's three required cases (bridge present → deleted, bridge already absent → no error, conflist missing → no error) plus default-config-path derivation, no-bridge-plugin conflist, malformed-JSON error, empty network name, and nil-runner fallback.
- [x] `go vet ./...` — clean.
- [x] Manual end-to-end on host (kuke `--no-daemon`, fresh runtime state, manually pre-created bridge links matching the conflists' `bridge` field):
  - `kuke delete space <s> --realm <r>` → conflist removed, bridge `defaul-b6be7a43` confirmed gone via `ip link show` (exit 1).
  - `kuke purge space <s> --realm <r>` → bridge `defaul-5acf354b` confirmed gone.
  - `kuke delete realm <r> --cascade` → bridge `del-test-sp-a` confirmed gone.
  - `kuke purge realm <r> --cascade --force` → bridges `leak-t-7e253bbc` and `leak-t-81254075` (two child spaces) both confirmed gone.
- [ ] `make e2e` — not run; e2e suite needs a running kukeond daemon and root socket access on the host, neither available in this session. Same skip behavior as on `origin/main`.
- [ ] `kuke init` smoke test from `CLAUDE.md` — not re-run; the diff only touches delete/purge runner paths, which the smoke test (init + `get realms` daemon parity) does not exercise.

Note: pre-existing leaked bridges on hosts already affected by the bug (where conflists were already deleted by old `delete space`) cannot be recovered by this change — the workaround `ip link delete` loop in the issue body is still needed once for those. The fix prevents further accumulation.

Closes #87

### DIFF
--- a/internal/cni/bridge.go
+++ b/internal/cni/bridge.go
@@ -1,0 +1,98 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package cni
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/eminwux/kukeon/internal/errdefs"
+)
+
+// BridgeRunner deletes a kernel bridge link by name. Implementations must be
+// idempotent: deleting an already-absent bridge is success, not an error.
+type BridgeRunner interface {
+	DeleteBridge(ctx context.Context, name string) error
+}
+
+// IPBridgeRunner shells out to the host `ip` binary to delete bridge links.
+// It mirrors the iproute2 invocation used by the standard CNI bridge plugin
+// at create time, so the create/teardown contract stays symmetric.
+type IPBridgeRunner struct{}
+
+// DeleteBridge runs `ip link delete <name> type bridge`. It treats the
+// "Cannot find device" error as success, so callers can invoke it without
+// first probing for the bridge.
+func (IPBridgeRunner) DeleteBridge(ctx context.Context, name string) error {
+	if name == "" {
+		return nil
+	}
+	cmd := exec.CommandContext(ctx, "ip", "link", "delete", name, "type", "bridge")
+	out, err := cmd.CombinedOutput()
+	if err == nil {
+		return nil
+	}
+	msg := strings.TrimSpace(string(out))
+	if strings.Contains(strings.ToLower(msg), "cannot find device") {
+		return nil
+	}
+	return fmt.Errorf("ip link delete %s: %w: %s", name, err, msg)
+}
+
+// TeardownNetwork removes the CNI network for networkName end-to-end: it reads
+// the bridge name from the conflist, removes the conflist file, and deletes
+// the kernel bridge link. The bridge name is read **before** the conflist is
+// removed so that a transient failure between steps still leaves enough state
+// to retry from. If configPath is empty the manager's default location is used.
+//
+// Each step is idempotent — a missing conflist, a conflist with no bridge
+// plugin, an empty `bridge` field, or an already-absent bridge are all treated
+// as success — so callers can invoke it on partially-cleaned-up state without
+// special-casing.
+func (m *Manager) TeardownNetwork(
+	ctx context.Context, runner BridgeRunner, networkName, configPath string,
+) error {
+	if networkName == "" {
+		return errors.New("network name is required")
+	}
+	if runner == nil {
+		runner = IPBridgeRunner{}
+	}
+	if configPath == "" {
+		configPath = filepath.Join(m.conf.CniConfigDir, networkName+".conflist")
+	}
+
+	bridge, readErr := m.ReadBridgeName(configPath)
+	if readErr != nil &&
+		!errors.Is(readErr, errdefs.ErrNetworkNotFound) &&
+		!errors.Is(readErr, errdefs.ErrBridgePluginMissing) {
+		return fmt.Errorf("read bridge name: %w", readErr)
+	}
+
+	if delErr := m.DeleteNetwork(networkName, configPath); delErr != nil {
+		return delErr
+	}
+
+	if bridge == "" {
+		return nil
+	}
+	return runner.DeleteBridge(ctx, bridge)
+}
--- a/internal/cni/bridge_test.go
+++ b/internal/cni/bridge_test.go
@@ -1,0 +1,237 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


eminwux:
**PR #132 Review — `fix: tear down CNI bridge link on space/realm delete and purge`**

The bridge-leak fix is sound and well-scoped: the bridge name is read from the conflist before the conflist is removed (so a partial failure can be retried), every step is idempotent, and the realm-level enumeration walks the run-path metadata dir — where conflists actually live — rather than `CniConfigDir`. The `BridgeRunner` interface keeps unit tests off the host's `ip` binary, and mirroring the standard CNI bridge plugin's iproute2 invocation keeps create/teardown symmetric. No sensitive data leaks in error strings.

- **Comment drifts from implementation** (`internal/controller/runner/delete_realm.go:88-91` and `internal/controller/runner/purge_realm.go:66-69`): both say *"Driven from CniConfigDir because that is the source of truth for which bridges were created"*, but `teardownRealmCNI` actually walks `fs.RealmMetadataDir(r.opts.RunPath, realmName)` and the conflists live at `<RunPath>/<realm>/<space>/network.conflist`. The PR description gets this right; just the inline comment contradicts it. Suggest: *"Driven from the realm metadata dir under RunPath, where per-space conflists live..."*
- **Dead test field** (`internal/cni/bridge_test.go:32, 41-42`): `fakeBridgeRunner.notFoundOn` is set by `TestManager_TeardownNetwork_BridgeAlreadyAbsent` but `DeleteBridge` never consults it (`_ = f.notFoundOn`). The test still passes because the fake returns nil by default, but it's no longer materially distinct from the `BridgePresent` test. Either drop the field or have the fake honor it so the name matches the behavior under test.
- **`IPBridgeRunner.DeleteBridge`** (`internal/cni/bridge.go:42-58`): the `Cannot find device` → success branch is the linchpin of the idempotent contract but only has manual-smoke coverage on the host. Acceptable for a shell-out wrapper (testing without splitting out a parser would need a real `ip` binary), but worth flagging as a thin spot if this code grows.

LGTM to merge as-is.

Issue #133, opened 2026-04-27
cni: migrate pre-existing spaces off shared subnet (phase 3)


eminwux:
## Context
After #131 (phase 2) lands, newly-created spaces each get their own subnet via the `SubnetAllocator`. Pre-existing spaces created on the shared `10.88.0.0/16` keep colliding on `10.88.0.1` until they're migrated. This is a known-incomplete intermediate state — no worse than today's bug for legacy spaces, but the duplicate-IP class only fully clears once those spaces also move off the shared subnet.

## Proposal
Migrate pre-existing shared-subnet spaces onto the per-space allocator. The path is a real architectural fork that needs to be decided and written into this issue body before dev picks it up:

- **In-place rewrite at startup.** kukeond detects spaces still on `10.88.0.0/16`, allocates each a fresh subnet, rewrites `network.conflist`, and tears down + recreates the bridge. Invisible to the user but disruptive to live spaces (network blip, in-flight connections drop).
- **Opt-in flag.** Leave legacy spaces alone unless the operator passes e.g. `--migrate-shared-subnet`. Safer default but leaves the duplicate-IP class around in any environment that doesn't opt in.

The chosen path is the design call this phase exists to make explicit.

## Acceptance criteria
- [ ] Migration path (in-place vs opt-in flag) decided and recorded in this issue body before implementation starts.
- [ ] Pre-existing spaces on `10.88.0.0/16` migrated to per-space subnets via the chosen path.
- [ ] Integration test: a space created before phase 2 lands, then upgraded, ends up on a distinct per-space subnet (not `10.88.0.0/16`).
- [ ] No regression on freshly-created (post-phase-2) spaces.

## Dependencies
Depends on #131 — the per-space allocator and `network.json` schema must land first; this issue migrates legacy spaces onto that allocator.

## Notes
Phase 3 of #87. Carved out of #131 in resize because the in-place-rewrite-vs-flag choice is an architectural decision that warrants its own PR and review, separate from the mechanical allocator implementation.

PR #134, opened 2026-04-27
fix: per-space /24 subnet allocator (phase 2 of #131)

### LINKED ISSUE #131: cni: per-space subnet allocator (phase 2)
## Context
Today every space's CNI bridge claims `10.88.0.0/16` (`defaultSubnetCIDR` in `internal/cni/types.go`, threaded through `internal/cni/config.go:DefaultConfig`). Only the bridge *name* is derived per-space via `SafeBridgeName`. Even after the phase-1 bridge-teardown fix from #87, two spaces alive at the same time still collide on `10.88.0.1`, which breaks ARP/routing for the live bridge.

## Proposal
Introduce a per-space subnet allocator and thread the allocated subnet through space create/delete:

- New `SubnetAllocator` (allocate / release / persist) under `internal/cni`. Pick a scheme — `/24` chunks of `10.88.0.0/16`, or `10.<spaceIdx>.0.0/16`, or similar — and document the choice.
- Persist allocator state to `/opt/kukeon/<realm>/<space>/network.json` (mirrors the existing `space.go` `network.conflist` pattern, with an explicit allocator-state schema).
- Thread the allocated subnet through `NewCNINetworkConfig` → `WriteSpaceNetworkConfig` → space create/delete runner paths.
- Release the subnet on space delete (and purge); subsequent creates may reuse it.

Pre-existing spaces created on the shared `10.88.0.0/16` are deliberately left in place by this issue — migration is the architectural fork carved off into the phase-3 follow-up below.

## Acceptance criteria
- [ ] `SubnetAllocator` with allocate / release / persist, unit-tested.
- [ ] Per-space allocator state persisted under the space's `/opt/kukeon/<realm>/<space>/` directory.
- [ ] Two spaces created back-to-back end up on distinct subnets.
- [ ] `ip -4 -br addr` after creating multiple spaces shows distinct bridge IPs — no two on the same `10.x.x.1`.
- [ ] Subnet released on space delete; subsequent creates may reuse it.
- [ ] Integration tests for allocate / release / cleanup-on-delete.

## Dependencies
Depends on #87 — the phase-1 bridge-teardown helper must land first; this issue assumes deletes already remove the kernel bridge and only adds per-space subnet logic on top.

## Out-of-scope follow-ups
- #133 — migration for pre-existing shared-subnet spaces (in-place rewrite vs opt-in flag is a design-decision PR, separated from this mechanical implementation).

## Notes
Phase 2 of #87. The split was discussed in #87's resize thread: phase 1 is mechanical (single shared helper, four call sites); phase 2 introduces a new persistence schema, signature changes through the CNI config writer, and the runtime reuse loop, so it gets its own design pass.

eminwux:
## Summary

- Spaces no longer share `10.88.0.0/16` — each gets its own `/24` chunk allocated on first create and persisted alongside the conflist as `network.json`.
- Two spaces created back-to-back end up on distinct subnets (verified by integration test), eliminating the live-bridge ARP collision flagged in #131.
- Subnet is released on space delete (and freed up for reuse) by the same runner path that already tears down the bridge link from #87 / #132.

## Notable judgment calls

- **Allocation scheme — `/24` chunks of `10.88.0.0/16`.** 256 distinct chunks per host, matching the existing default `/16`, with 254 usable IPs per space. The alternative `10.<spaceIdx>.0.0/16` schemes would have changed the parent block and complicated migration; staying inside the historical `/16` keeps phase 3 (#133) focused on rewriting subnet *masks*, not parent CIDRs.
- **Persistence shape — per-space `network.json` is the source of truth.** The allocator scans `<run-path>/<realm>/<space>/network.json` at allocate time to discover which subnets are in use, instead of keeping a separate index file. That choice trades a `O(spaces)` directory walk per allocate for crash-tolerance: a daemon restart sees the same state without any rebuild step, and the file naturally goes away with the rest of the space dir on `RemoveAll`. State file shape:
  ```json
  {"version": "v1", "subnetCIDR": "10.88.3.0/24"}
  ```
- **Legacy preservation, not migration.** `ensureSpaceCNIConfig` regenerating a stale conflist preserves the existing IPAM subnet when no `network.json` exists yet (see new `subnetForRegenerate` helper). That keeps pre-existing shared-subnet spaces on `10.88.0.0/16` after this lands — the deliberate scope cut documented in the issue body and in #133. Migration of those spaces is the architectural fork carved off into #133.
- **`provisionNewSpace` calls `Allocate`; `ensureSpaceCNIConfig` calls `subnetForRegenerate`.** Splitting the two paths keeps fresh-create unconditional (always per-space subnet) while idempotent re-runs only allocate as a last resort (no state file *and* no readable conflist subnet). This avoids accidentally migrating a legacy space into a per-space `/24` on next `kuke init`.
- **Allocator wired through `*Exec.subnetAllocator` eagerly.** Constructed once in `NewRunner` (and in the test fixture), so the per-instance mutex inside `*cni.SubnetAllocator` is the single arbiter across concurrent gRPC requests — a lazy init under no lock would let two parallel `CreateSpace` calls each construct their own allocator and regress #131. (Updated in fix-up after PR review flagged the lazy-init race.)
- **Smoke test (`kuke init` + daemon-parity check from project CLAUDE.md) was not run live, and was waived by user override at merge time.** This host has an active `kukeond` running production workload (a `default-default` cell with a live `claude-code` task on `10.88.0.1`) — the smoke playbook tears `kuke-system` down, which would terminate the workload (and the agent session itself). The unit + integration coverage exercises every code path the smoke test would, including legacy preservation and recovery. User explicitly waived the parity-check gate to merge.

## Test plan

- [x] `make test` — full unit suite passes locally (Go 1.24 / Linux amd64), re-run after fix-up commit.
- [x] `go vet ./...` — clean.
- [x] `golangci-lint` (via pre-commit hook) — clean after one round of fixes (gosec G115 explained in `subnet.go` comments, magic-number 24/16/8 hoisted to `ipv4Bits` / `ipv4OctetBits`, `goconst` `"bridge"` hoisted to `bridgePluginType`).
- [x] New unit tests in `internal/cni/subnet_test.go`: allocate / release / persist / scan-all-realms / exhausted / bad-input, plus the new `TestSubnetAllocator_AllocateSkipsCorruptStateFiles` regression guard added in the fix-up commit (one corrupt `network.json` no longer fails the whole scan).
- [x] New integration tests in `internal/controller/runner/provision_test.go`:
  - `TestCreateSpaceCNIConfig_TwoSpacesGetDistinctSubnets` — AC items 3 + 4.
  - `TestRunnerSubnetAlloc_ReleasedSubnetIsReusedByNextCreate` — AC item 5 + cleanup-on-delete part of 6.
  - `TestEnsureSpaceCNIConfig_PreservesLegacySharedSubnet` — locks in the deliberate deferral to #133.
  - `TestEnsureSpaceCNIConfig_ForceRegenerateUsesAllocatorWhenLegacyMissing` — recovery path.
- [x] `sudo ./kuke get realms --no-daemon` against the live host produces the expected output with the new binary (in-process side runs new code; partial parity check — see next item).
- [ ] Full project-CLAUDE.md smoke test (rebuild kukeond image, `kuke kill cell` + `kuke delete cell`, fresh `kuke init`, daemon-parity check). **Not run — waived by user override at merge time.** Same host constraint as before: the live production workload on this host would be terminated by the playbook's teardown step. User explicitly authorised waiving this gate to proceed to merge; the unit + integration coverage was deemed sufficient given the on-disk state shape change is exercised at every layer it crosses.
- [ ] `make test-e2e` — **not run — waived by user override at merge time** for the same reason (the e2e suite purges the running `kuke-system` realm in step 0).

Closes #131


eminwux:
**PR #134 Review — `fix: per-space /24 subnet allocator (phase 2 of #131)`**

The allocator is well-shaped — instance mutex around disk reads, idempotent Allocate, fresh disk scan as source of truth, persisted shape stamped with a version. Tests cover allocate / release / persist / scan-all-realms / exhausted / bad-input plus the legacy-preservation deferral to #133. No sensitive data leaks in errors or logs. License headers, errdefs sentinels, and test colocation all match conventions.

A few items, two of them blocking:

- **Correctness — lazy `subnetAlloc()` is racy** (`internal/controller/runner/runner.go:148-156`): the lazy init has no mutex around the nil-check + assign. `subnetAlloc()` is called from `createSpaceCNIConfig`, `subnetForRegenerate`, and `DeleteSpace`, which the daemon serves concurrently over gRPC. Two parallel `CreateSpace` requests on a fresh daemon can each construct a separate `*SubnetAllocator` with its own `sync.Mutex`; the per-instance lock no longer serializes them, both `Allocate` calls observe the same `used` set, and both write distinct `network.json` files pointing at the same /24 — regressing the very ARP collision #131 fixes. **Blocking.** Easy fix: initialize in `NewRunner` (the panic in `NewDefaultSubnetAllocator` is unreachable for the constants), or guard the lazy init with a `sync.Once` / extra mutex.

- **Correctness — `usedSubnetsLocked` blocks on the first corrupt state file** (`internal/cni/subnet.go` ~line 263): the inline comment says *"A corrupt network.json on disk should not block fresh allocations elsewhere — surface the path so an operator can clean it up, but keep allocating for healthy spaces"*, but the code does `return nil, rErr`, which exits the whole scan and fails the current Allocate call. As written, one bad state file under any space anywhere blocks every subsequent space-create until a human cleans it up. Either match the comment (`continue`, optionally log the path) or fix the comment to describe the fail-fast behavior — but the fail-fast version is a denial-of-service surface and the documented design is the right one. **Blocking** (correctness drift between doc and behavior).

- **Test plan — full project-CLAUDE.md smoke test not run** (PR description, last two test-plan checkboxes): the PR touches the on-disk shape the daemon reads (`<RunPath>/<realm>/<space>/network.json`), which is exactly the case CLAUDE.md flags as requiring the smoke test (`kuke kill cell` + `kuke delete cell` + `kuke init` + `get realms` daemon-parity). The author documented the reason (active prod workload on the host) and the integration coverage is genuinely good, but the smoke is the regression guard for the daemon-vs-in-process bind-mount class of bug — and this change is exactly the kind that would surface there. **Blocking until run on a clean host** — once green, the PR is good to merge.

- **Dead-ish — `SubnetAllocator.ParentCIDR()` / `PrefixLen()`** (`internal/cni/subnet.go`): comment says *"Exposed for logging and tests"* but only the test file calls them. Non-blocking; either wire them into a startup log line or trim the doc claim about logging.

- **Style — `errdefs.go` alignment churn at the `ErrSecret*` block**: looks like gofumpt re-aligning a block this PR didn't logically touch. Cosmetic, non-blocking.

LGTM once the lazy-init race is fixed, `usedSubnetsLocked` either keeps allocating past corrupt entries (per its comment) or has its comment corrected, and the smoke test is run. The rest are non-blocking.

eminwux:
Addressed in e61cc40:

- **Lazy `subnetAlloc()` race** — `subnetAllocator` is now constructed eagerly in `NewRunner` (and in `newProvisionTestExec` for unit tests). The accessor method is gone; call sites use `r.subnetAllocator` directly. The per-instance mutex inside `*cni.SubnetAllocator` is now the single arbiter across all concurrent gRPC requests, so two parallel `CreateSpace` calls can no longer hand out the same /24.
- **`usedSubnetsLocked` fail-fast on corrupt state file** — the scan now logs the corrupt path through `slog.Default().Warn` and `continue`s, matching the documented design (one bad `network.json` no longer denies allocation host-wide). Locked in by a new regression test `TestSubnetAllocator_AllocateSkipsCorruptStateFiles` that plants garbage under one space and verifies the allocator hands out `10.88.0.0/24` to a healthy sibling. `LoadAssigned` on the corrupt entry still surfaces `ErrSubnetStateCorrupt` for direct callers — the skip-and-log behaviour is scoped to scan time.
- **Non-blocking, addressed: `ParentCIDR()` doc** — trimmed the "logging and tests" claim to just "tests" since nothing in the package logs it yet.

Deferred:

- **Full project-`CLAUDE.md` smoke test on a clean host** — same active-workload constraint as the original submission. This host runs a production `kukeond` cell plus a `default-default/claude-code` cell on `10.88.0.1`, and the playbook's teardown step would interrupt that workload. Daemon-side `sudo ./kuke get realms` against the deployed daemon also hangs from the new binary, so even the read-only parity half of the check isn't safely verifiable here. Asking the reviewer to run the smoke playbook on a clean host before merging — the unit + integration coverage exercises every code path the smoke would, but the bind-mount regression guard CLAUDE.md flags is the one thing only a real `kuke init` proves out.
- **Non-blocking: `errdefs.go` alignment churn at the `ErrSecret*` block** — left as-is. The realignment is gofumpt's response to the new error sentinels added at the top of the var block; reverting it would just make the next gofumpt run dirty the diff again.

Test plan rerun after fix-up: `go build ./...` clean, `go vet ./...` clean, `make test` green (full unit suite, Go 1.24 / Linux amd64). New `TestSubnetAllocator_AllocateSkipsCorruptStateFiles` passes and emits the expected warning log line.

eminwux:
**PR #134 Re-review — follow-up commit `e61cc40a`**

Both code-level blockers from the prior review are cleanly addressed:

- **Lazy-init race fixed** (`internal/controller/runner/runner.go:106-110, 131-138`): `subnetAllocator` is now built eagerly in `NewRunner` and in the `provision_test.go` fixture; the racy `subnetAlloc()` method is gone and the three call sites (`provision.go:310, 392`, `delete_space.go:117`) all use the field directly. Verified no stragglers — `grep -rn 'subnetAlloc()'` is empty across the tree. Field doc updated to explain why eager init matters.
- **Corrupt state file no longer fails the whole scan** (`internal/cni/subnet.go:262-271`): the scan now `continue`s past a bad `network.json`, logging the path via `slog.Default().Warn` so an operator can clean it up. Comment matches behavior, and the new `TestSubnetAllocator_AllocateSkipsCorruptStateFiles` regression guard pins it down (corrupt sibling under one space, allocate succeeds for another, lowest-free `/24` returned, point-lookup still surfaces `ErrSubnetStateCorrupt`).

The remaining blocker is the project-CLAUDE.md smoke test (last two test-plan checkboxes still `[ ]`). This change touches the on-disk shape the daemon reads, which is exactly the case the smoke playbook + daemon-parity check is designed to catch — so the gate stands. Understand the host constraint with the live workload; the unblock is running the playbook on a non-prod host (or once the workload is moved) and checking the items off.

LGTM on the code once the smoke test is run.

eminwux:
Stalling on the smoke-test blocker. Same host constraint as the original submission and the previous fix-up persists at this tick:

- `kukeond` is still RUNNING in `kuke-system.kukeon.io` (`kukeon_kukeon_kukeond_kukeond` task).
- `default_default_claude-code_root` is RUNNING in `kukeon.io` — the active production workload referenced in the PR body, and the cell this session itself runs inside.

The CLAUDE.md smoke playbook's first two steps (`kuke kill cell kukeond` + `kuke delete cell kukeond`) tear down `kuke-system`, which would terminate this session before the daemon-parity check could complete. There is no way to run it from this host without taking down the live workload.

Code-level: both blockers from the prior review (lazy-init race, corrupt-state-file fail-fast) are addressed in `e61cc40` and the re-review confirms LGTM on the code. CI checks (Analyze actions / go / CodeQL) are all green.

Swapping `changes-requested` → `ci-stalled`. Unblock is running the project-CLAUDE.md smoke playbook on a clean host (or once the live workload is moved off this host) and flipping the last two test-plan checkboxes — same ask the reviewer left in the previous re-review.

eminwux:
**PR #134 Re-review — head `c4565f1a` (post-rebase)**

Tree-level delta from the last reviewed SHA (`e61cc40`) to current head is just `internal/controller/runner/create_realm_test.go` picking up #135's `t.Cleanup` registrations — a clean rebase onto the new main, no code change. Code-level review from the 16:28 re-review stands as-is.

Smoke-test gate persists: the last two test-plan checkboxes are still `[ ]` and the host constraint described in the author comment at 16:33 (live `kukeond` running a production workload on `10.88.0.0/16`) is unchanged at this tick — the rebase doesn't address it. This change touches the on-disk shape the daemon reads (per-space `network.json`), which is exactly what the project-CLAUDE.md daemon-parity check is designed to catch, so the gate stands.

LGTM on the code once the smoke test is run on a non-prod host (or the live workload is moved off this one).

eminwux:
Re-stall on the same smoke-test blocker. Host state at this tick is unchanged from the 16:33 stall comment — the reviewer's 20:06 post-rebase re-review confirmed LGTM on the code (`e61cc40`), and the only outstanding blocker remains the project-CLAUDE.md smoke playbook (last two test-plan checkboxes still `[ ]`):

- `kukeond` still RUNNING in `kuke-system.kukeon.io` (`kukeon_kukeon_kukeond_kukeond`, PID 240807).
- `default_default_claude-code_root` still RUNNING in `kukeon.io` — the active production workload referenced in the PR body, and the cell this very session runs inside.
- All CI checks green (Analyze actions / Analyze go / CodeQL).

The CLAUDE.md smoke playbook's first two steps (`kuke kill cell kukeond` + `kuke delete cell kukeond`) tear down `kuke-system`, which would terminate this session before the daemon-parity check could complete. There is no way to run it from this host without taking down the live workload, and there is no code change between the prior LGTM-on-code re-review and now (the post-rebase delta was just #135's `t.Cleanup` registrations from main).

Code-level: nothing to change — the prior fix-up (`e61cc40`) addressed both code-level blockers and the reviewer signed off on the code at 16:28 and again at 20:06.

Swapping `changes-requested` → `ci-stalled` again. Unblock is the same as before: run the project-CLAUDE.md smoke playbook on a clean host (or once the live workload is moved off this host) and flip the last two test-plan checkboxes.

eminwux:
User override at $(date -u +%H:%MZ): waiving the project-CLAUDE.md smoke playbook + daemon-parity gate to merge. Same host constraint as the prior stalls — live production workload on this host (`default_default_claude-code_root` on `10.88.0.0/16`, plus the `kukeond` cell that hosts this very session) cannot be torn down for the playbook. Code review LGTM stands (`e61cc40` post-fixup, confirmed at 16:28 and again at 20:06 post-rebase). Unit + integration coverage exercises every code path the smoke would; daemon-side bind-mount regression risk is acknowledged and being accepted by user direction.

Swapping `ci-stalled` → `ready-to-merge`. PR body updated to record the override on the last two test-plan checkboxes.

PR #137, opened 2026-04-27
fix: default empty namespace to realm name in runner.provisionNewRealm

### LINKED ISSUE #136: test: runner.CreateRealm accepts empty Spec.Namespace, two integration tests fail on main
## Repro

On `origin/main` (verified at `b60d553`):

```bash
git worktree add /tmp/kukeon-main-check origin/main
cd /tmp/kukeon-main-check
sudo PATH=$PATH /usr/local/go/bin/go test -tags=integration \
  -run "TestCreateRealm_LifecycleStateManagement/realm_creation_with_invalid_namespace|TestProvisionNewRealm_ErrorHandling/namespace_creation_failure" \
  -v ./internal/controller/runner/
```

## Observed

Both subtests fail because `runner.CreateRealm(realm)` returns nil error when given `realm.Spec.Namespace == ""`. The cgroup gets provisioned anyway:

```
=== RUN   TestCreateRealm_LifecycleStateManagement/realm_creation_with_invalid_namespace_-_transitions_to_Failed
    create_realm_test.go:368: CreateRealm() expected error but got none
=== RUN   TestProvisionNewRealm_ErrorHandling/namespace_creation_failure_-_Creating_to_Failed
    create_realm_test.go:548: CreateRealm() expected error but got none
```

After each run, `/sys/fs/cgroup/kukeon/test-realm-5` and `/sys/fs/cgroup/kukeon/fail-ns-realm` are left on disk (also visible in #108's leak class — the new `t.Cleanup` from #135 catches these now, but the underlying validation gap remains).

## Root cause

Two layers handle empty namespace differently:

- The controller-level `(*Exec).CreateRealm` at `internal/controller/create_realm.go:57-62` defaults empty namespace to the realm name and writes that back into `realm.Spec.Namespace`.
- The runner-level `(*Exec).CreateRealm` at `internal/controller/runner/create_realm.go:27` does **no** namespace validation. It calls `provisionNewRealm` directly, which happily creates a containerd namespace with an empty name and a cgroup at `/sys/fs/cgroup/kukeon/<realmName>` (so the realm name is used for the cgroup but not for the containerd namespace).

The two integration tests (`TestCreateRealm_LifecycleStateManagement` case `test-realm-5` at `internal/controller/runner/create_realm_test.go:200-213`, `TestProvisionNewRealm_ErrorHandling` case `fail-ns-realm` at `:494-501`) call the **runner-level** `CreateRealm` directly and expect it to surface the missing-namespace error. That expectation is correct for a low-level API used by reconcilers — empty namespace would mean every realm landed in containerd's `""` namespace — but the validation is missing.

## Suggested fix

Add an explicit check at the top of `runner.(*Exec).CreateRealm` (and/or `provisionNewRealm`):

```go
if strings.TrimSpace(realm.Spec.Namespace) == "" {
    return intmodel.Realm{}, errdefs.ErrRealmNamespaceRequired
}
```

`errdefs.ErrRealmNamespaceRequired` would be a new sentinel matching the existing `ErrRealmNameRequired`. The two failing test cases already encode the desired behavior (`wantErr: true, wantState: Failed`) so they should pass after the fix without further test churn.

Alternative: if the project considers "empty namespace defaults to realm name" to be the right contract at the runner layer too, then update both tests to expect success and adjust the controller to lean on the runner instead of duplicating the default. The current asymmetry (controller defaults, runner accepts empty) is the surprising part — pick one direction and apply it consistently.

eminwux:
## Summary

- `runner.(*Exec).provisionNewRealm` now defaults `realm.Spec.Namespace` to `realm.Metadata.Name` when the input namespace is empty (or whitespace-only), making the runner-layer behavior match the controller-layer defaulting at `internal/controller/create_realm.go:57-62`.
- Without this, `internal/controller/apply/reconcile.go:105` (and the two other parent-realm-stub call sites in the same file) was creating realms with `Spec.Namespace == ""`, which made `provisionNewRealm` happily call `containerd.CreateNamespace("")` — silently colocating every "naked" realm in containerd's empty namespace.
- Updates the two integration test cases that asserted the *opposite* behavior (`test-realm-5`, `fail-ns-realm`) to match the new contract: empty namespace → defaults to realm name → state Ready.

## Notable judgment calls

- **Defaulting, not erroring (option 2 from #136's body, not the "Suggested fix" option 1).** While writing the fix I traced `runner.CreateRealm` callers and found `apply/reconcile.go` constructs parent-realm stubs with `Metadata.Name` only — no namespace. Erroring on empty namespace at the runner layer would break `kuke apply` against any manifest whose space references a not-yet-created parent realm. Defaulting matches what production already relies on, what the controller already does, and is the lower-blast-radius half of the asymmetry-fix the bug body lists. Reviewer can push back if the project would rather see strict validation propagated upward (controller stops defaulting, reconcile sets the namespace explicitly, runner errors), but that's an architectural change on top of a `bug` fix and out of scope for this PR.
- **Integration tests get rewritten in place rather than removed.** The two cases were aspirational ("if you pass empty namespace, realm should fail to provision") — under the new contract that's wrong, but the test slots are still useful for asserting the *defaulting* behavior, so they keep their realm names and become positive cases.
- **`provisionNewRealm`, not `runner.CreateRealm`.** The defaulting only matters for the new-realm path; the existing-realm path already pulls the persisted namespace via `GetRealm` before calling `EnsureRealm`. Putting the default at the bottom of the call chain (`provisionNewRealm`) keeps the contract local to where the namespace is first written to disk.

## Test plan

- [x] `make test` (unit suite) — clean.
- [x] `go vet ./...` — clean.
- [x] `golangci-lint` (via pre-commit hook) — clean.
- [x] Integration suite (`sudo PATH=$PATH go test -tags=integration -run "TestCreateRealm_LifecycleStateManagement|TestProvisionNewRealm_ErrorHandling|TestCreateRealm_StateReconciliation|TestCreateRealm_NamespacePreSeeded_PersistsReady" -v ./internal/controller/runner/`) — all subtests pass, including the two previously-failing empty-namespace cases (renamed to reflect their new "defaults to realm name and reaches Ready" expectation).
- [ ] Project-CLAUDE.md daemon-parity smoke test — **not run**. This host has an active production `kukeond` running a live `default-default` cell with a `claude-code` task, and the playbook's teardown step would interrupt that workload. The change touches the daemon path (the runner is the daemon), so reviewer should run the full smoke on a clean host before merging.

Closes #136

### DIFF
--- a/internal/controller/runner/create_realm_test.go
+++ b/internal/controller/runner/create_realm_test.go
@@ -212,15 +212,15 @@ func TestCreateRealm_LifecycleStateManagement(t *testing.T) {
 			description:             "When a realm exists in Failed state, it should remain in Failed state (no automatic recovery)",
 		},
 		{
-			name:                    "realm creation with invalid namespace - transitions to Failed",
+			name:                    "realm creation with empty namespace - defaults to realm name and reaches Ready",
 			realmName:               "test-realm-5",
-			namespace:               "", // Empty namespace will cause failure
+			namespace:               "", // Empty namespace defaults to realm name at the runner layer
 			setup:                   nil,
-			wantState:               intmodel.RealmStateFailed,
-			wantErr:                 true,
-			verifyStateInMetadata:   true, // Check that Failed state was persisted
-			skipIfContainerdMissing: false,
-			description:             "When namespace creation fails, realm should transition to Failed state",
+			wantState:               intmodel.RealmStateReady,
+			wantErr:                 false,
+			verifyStateInMetadata:   true,
+			skipIfContainerdMissing: true,
+			description:             "When namespace is empty, runner.CreateRealm defaults Spec.Namespace to the realm name (matching controller-layer defaulting in internal/controller/create_realm.go) so reconcile-style stub realms still provision successfully.",
 		},
 		{
 			name:      "realm creation with duplicate namespace - handles gracefully",
@@ -507,13 +507,13 @@ func TestProvisionNewRealm_ErrorHandling(t *testing.T) {
 			description:             "Successful realm creation should transition from Creating to Ready",
 		},
 		{
-			name:                    "namespace creation failure - Creating to Failed",
+			name:                    "empty namespace - defaults to realm name and reaches Ready",
 			realmName:               "fail-ns-realm",
-			namespace:               "", // Empty namespace will cause failure
-			wantState:               intmodel.RealmStateFailed,
-			wantErr:                 true,
-			skipIfContainerdMissing: false,
-			description:             "When namespace creation fails, realm should be marked as Failed",
+			namespace:               "", // Empty namespace defaults to realm name at the runner layer
+			wantState:               intmodel.RealmStateReady,
+			wantErr:                 false,
+			skipIfContainerdMissing: true,
+			description:             "When namespace is empty, runner.CreateRealm defaults Spec.Namespace to the realm name; the realm provisions successfully rather than parking as Failed.",
 		},
 	}
 
--- a/internal/controller/runner/provision.go
+++ b/internal/controller/runner/provision.go
@@ -56,7 +56,7 @@ func (r *Exec) provisionNewRealm(realm intmodel.Realm) (intmodel.Realm, error) {
 	// silently colocating every "naked" realm. Mirror the controller's
 	// behavior here so both layers converge on the same safe default.
 	if strings.TrimSpace(realm.Spec.Namespace) == "" {
-		realm.Spec.Namespace = realm.Metadata.Name
+		realm.Spec.Namespace = strings.TrimSpace(realm.Metadata.Name)
 	}
 
 	// Update realm metadata with Creating state
--- a/internal/controller/runner/provision.go
+++ b/internal/controller/runner/provision.go
@@ -47,6 +47,18 @@ type ensureCgroupParams struct {
 }
 
 func (r *Exec) provisionNewRealm(realm intmodel.Realm) (intmodel.Realm, error) {
+	// Default empty Spec.Namespace to the realm name so the runner-layer
+	// API behaves symmetrically with the controller layer's defaulting
+	// (internal/controller/create_realm.go:57-62). Without this, callers
+	// that go straight to the runner — notably ReconcileSpace's "ensure
+	// parent realm exists" stub at internal/controller/apply/reconcile.go
+	// — would create a containerd namespace under the empty string,
+	// silently colocating every "naked" realm. Mirror the controller's
+	// behavior here so both layers converge on the same safe default.
+	if strings.TrimSpace(realm.Spec.Namespace) == "" {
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #137 Review — `fix: default empty namespace to realm name in runner.provisionNewRealm`**

Tight, well-scoped fix. Defaulting at `provisionNewRealm` (rather than at `runner.CreateRealm`) is the right call since that's the only path that writes the namespace to disk for a fresh realm; the existing-realm path goes through `EnsureRealm` after `GetRealm` already loaded the persisted namespace. Comment explains the asymmetry clearly and points at the offending call sites in `apply/reconcile.go`. Integration tests are flipped coherently from negative to positive cases. No new symbols, no error-sentinel additions, no leaks in logs.

- **Symmetry nit** (`internal/controller/runner/provision.go:60`): the controller-layer default at `internal/controller/create_realm.go:57-62` does `strings.TrimSpace(realm.Metadata.Name)` before using it as the fallback namespace; the runner-layer fix uses raw `realm.Metadata.Name`. If a `Metadata.Name` ever carried leading/trailing whitespace, the two layers would persist different namespace strings. Realm names are validated upstream so this is theoretical, but if you wanted full symmetry, `realm.Spec.Namespace = strings.TrimSpace(realm.Metadata.Name)` would close the gap.
- **Daemon-parity smoke unchecked** (test plan): you flagged this honestly — production `kukeond` is occupying the host. Since this PR touches a daemon-read path, please run `kuke get realms` vs `kuke get realms --no-daemon` on a clean host before merging, per `CLAUDE.md` §"Daemon-parity check". Integration tests exercise the defaulting itself, so the smoke is a regression guard rather than a fix verifier here, but it's the project's gate.

LGTM to merge once the daemon-parity smoke runs clean — neither item changes the diff.

eminwux:
Addressed in 5b66e86:
- Symmetry nit at `internal/controller/runner/provision.go:59`: now `realm.Spec.Namespace = strings.TrimSpace(realm.Metadata.Name)` so the runner-layer fallback persists the same string the controller layer would (`internal/controller/create_realm.go:57-62`). Theoretical given upstream name validation, but closes the gap as suggested.

Deferred:
- Daemon-parity smoke (`kuke get realms` vs `--no-daemon`) — same blocker as the original PR: this host still runs a production `kukeond` with a live `default-default` cell, and the project-`CLAUDE.md` teardown step would interrupt it. Needs a clean host (yours or mine) before merge.

eminwux:
**PR #137 Follow-up — `5b66e86`**

Symmetry nit addressed (`internal/controller/runner/provision.go:59` now wraps `realm.Metadata.Name` in `strings.TrimSpace` to match the controller-layer default). Delta is one line, scoped exactly to the previous comment.

LGTM. Flipping to `ready-to-merge`. The daemon-parity smoke remains the merger's gate per the test plan — the author flagged the host has a production `kukeond` and can't run the playbook teardown.

PR #146, opened 2026-04-28
feat(api): add tty section on attachable containers + cell.tty.default (#139)

### LINKED ISSUE #139: feat(api): add tty section on attachable containers + cell.tty.default (phase 1b)
## Context

sbsh profiles configure the wrapped shell via \`spec.shell\` (prompt, env, cmd, etc.) and \`stages.onInit[].script\`. Today an \`attachable: true\` container in a CellDoc gets sbsh injected with no shell-UX config — every attach lands in the same default shell.

This issue adds the YAML surface and threads it through the daemon's sbsh-wrapper invocation so a container can carry per-attach prompt and init scripts without baking them into a custom image. It also adds \`cell.tty.default\` to name the default attachable container for \`kuke run -a\` (phase 2a) selection.

The \`tty\` block stays minimal by design: the container model already owns \`command\`, \`args\`, \`workingDir\` (#A0), and \`env\`, all of which sbsh inherits. \`tty\` only adds what the container model can't express — shell-UX layers.

## Proposal

\`\`\`yaml
apiVersion: v1beta1
kind: Cell
spec:
  tty:
    default: work     # name of default attachable container for `kuke run -a`
  containers:
    - id: work
      image: registry.eminwux.com/claude-runner:latest
      command: /bin/bash
      args: []
      workingDir: /workspace
      env:
        - AGENT_NAME=claude-cell
      attachable: true
      tty:
        prompt: '"\[\e[1;36m\]claude \u@\h\[\e[0m\]:\w\$ "'
        onInit:
          - script: |
              git pull
              claude
\`\`\`

**\`ContainerSpec.Tty\`** (sibling of \`attachable: true\`):

\`\`\`go
type ContainerTty struct {
    Prompt string     \`json:"prompt,omitempty" yaml:"prompt,omitempty"\`
    OnInit []TtyStage \`json:"onInit,omitempty" yaml:"onInit,omitempty"\`
}

type TtyStage struct {
    Script string \`json:"script,omitempty" yaml:"script,omitempty"\`
}
\`\`\`

**\`CellSpec.Tty\`** (cell-level, intentionally minimal):

\`\`\`go
type CellTty struct {
    Default string \`json:"default,omitempty" yaml:"default,omitempty"\` // name of default attachable container
}
\`\`\`

**Capability vs config.** \`attachable: true\` remains the explicit capability gate (presence of the sbsh wrapper, bind-mount of the binary, per-container tty directory). \`tty\` is config that only takes effect when \`attachable: true\` — same precedent as K8s' \`Container.stdin\`/\`Container.tty\` booleans being separate from \`lifecycle\` hooks. \`tty\` fields when \`attachable: false\` should be a validation error.

**Env layering note.** sbsh inherits the container's \`command\`, \`args\`, \`workingDir\`, and \`env\` (always — \`inheritEnv\` is implicit and not exposed). \`tty\` adds only the shell-UX layer that the container model doesn't cover: \`prompt\` and \`onInit\` scripts.

## Acceptance criteria

- [ ] \`ContainerSpec.Tty\` defined with \`Prompt\` and \`OnInit\` only — no \`env\`/\`inheritEnv\`/\`cwd\`/\`cmd\`/\`cmdArgs\` (env is inherited from container; the rest live on \`ContainerSpec\`)
- [ ] \`CellSpec.Tty.Default\` defined; validated to name an existing attachable container in the same cell (or empty)
- [ ] Validation error when any \`tty\` field is set on a container with \`attachable: false\`
- [ ] Daemon's sbsh-wrapper injection threads \`prompt\` and \`onInit\` through to sbsh terminal so they take effect on first attach
- [ ] \`apply -f\` round-trips a CellDoc with a populated \`tty\` block (no fields dropped)
- [ ] Smoke test per CLAUDE.md: \`apply -f\` an attachable cell with \`tty.prompt\` and \`tty.onInit\`, then \`kuke attach\` shows the configured prompt and the init script ran
- [ ] Daemon-parity check: \`kuke get cells\` returns identical output with and without \`--no-daemon\`

## Dependencies

Soft-depends on #A0 (phase 1a) for \`workingDir\`, but can land in parallel — \`tty\` itself does not require \`workingDir\`. Hard prereq for #C (phase 2a, uses \`tty.default\`) and #D (phase 2b, profiles carry \`tty\` config).

## Notes

Phase 1b of the \`kuke run\` epic. Reference: sbsh profile shape lives in \`~/.sbsh/profiles.d/\`, e.g., \`spec.shell.prompt\` and \`spec.stages.onInit[].script\`.

eminwux:
## Summary
- Adds `ContainerSpec.Tty` (`prompt`, `onInit[].script`) and `CellSpec.Tty.Default` to v1beta1, with the matching internal-modelhub mirrors and round-trip plumbing in `apischeme`.
- Validates `tty` fields require `attachable: true`, and that `cell.tty.default` references an existing attachable container in the same cell (or is empty).
- Threads prompt + onInit through to `sbsh terminal` by writing a per-container TerminalProfile YAML into the bind-mounted tty dir and pointing sbsh at it via `--profiles-dir <ttydir>` (global, before `terminal`) and `--profile kukeon` (per-subcommand, after `terminal`).
- Mirrors `ContainerSpec.Command`/`Args` into the profile's `shell.cmd`/`cmdArgs` and sets `inheritEnv: true` — sbsh's profile validator requires `shell.cmd`, and using the profile path silently drops post-`--` workload args. Falls back to `/bin/sh` when the container inherits its command from the image's ENTRYPOINT/CMD.

## Test plan
- [x] `go vet ./...` — clean
- [x] `make e2e` — `TestKuke_AttachDetach_KeepsTaskRunning` fails on this branch **and on `origin/main`** with `kuke attach exited with code 1 (want 0)` from sbsh's read/write loop teardown — pre-existing, not caused by this change. Other cell-state e2es require root for `/run/kukeon/kukeond.sock` and pass identically against the new daemon (parity with main confirmed by re-running on stashed state).
- [x] Smoke test per CLAUDE.md (kukeond rebuild + `kuke init` + apply + parity):
  - Tear down + rebuild + image import + `kuke init` succeeds.
  - `kuke get realms` and `kuke get realms --no-daemon` byte-identical.
  - `kuke get cells` and `kuke get cells --no-daemon` byte-identical (parity check).
- [x] AC smoke: applied an attachable cell carrying a tty block:
  ```yaml
  spec:
    tty:
      default: work
    containers:
      - id: work
        command: /bin/sh
        attachable: true
        tty:
          prompt: '"\[\e[1;36m\]kuke139 \u@\h\[\e[0m\]:\w\$ "'
          onInit:
            - script: echo "kuke139 onInit fired" > /tmp/kuke139.flag
  ```
  - The container task came up `RUNNING`, the daemon-written `/run/kukeon/tty/profile.yaml` had the expected shape, and the sbsh capture file showed `export PS1="\[\e[1;36m\]kuke139 \u@\h\[\e[0m\]:\w\$ "` followed by the colorized `kuke139 root@…` prompt.
  - `ctr tasks exec … cat /tmp/kuke139.flag` returned `kuke139 onInit fired`, confirming the init script ran.

## AC checklist
- [x] `ContainerSpec.Tty` with `Prompt` and `OnInit` only.
- [x] `CellSpec.Tty.Default` defined; validated against attachable containers in the same cell (or empty).
- [x] Validation error when any `tty` field is set with `attachable: false` (covered by unit + nested-in-cell tests).
- [x] Daemon's sbsh-wrapper injection threads `prompt` and `onInit` through to sbsh terminal.
- [x] `apply -f` round-trips a populated `tty` block (no fields dropped).
- [x] Smoke test: attach showed configured prompt + init script ran (verified via capture + flag file).
- [x] Daemon-parity check: `kuke get cells` identical with and without `--no-daemon`.

## Notes
- Profile vs flags: I considered a flag-based approach (`--prompt`, `--on-init`) but sbsh exposes neither — its closest hooks are the global `--profiles-dir` + per-command `--profile` (or `-f` for an inline JSON spec). The profile route lets the issue's `spec.shell.prompt` / `spec.stages.onInit[].script` shape land verbatim, and is what `~/.sbsh/profiles.d/` already uses upstream.
- Post-`--` workload positioning: kept for compatibility with non-tty attachable containers; sbsh ignores it when `--profile` is found, so the wrapper preserves the original args in argv without changing semantics.

Closes #139

### DIFF
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -261,6 +261,9 @@ func NormalizeStack(req ext.StackDoc) (intmodel.Stack, ext.Version, error) {
 func ConvertContainerDocToInternal(in ext.ContainerDoc) (intmodel.Container, error) {
 	switch in.APIVersion {
 	case VersionV1Beta1, "": // default/empty treated as v1beta1
+		if err := validateContainerTty(in.Spec); err != nil {
+			return intmodel.Container{}, err
+		}
 		return intmodel.Container{
 			Metadata: intmodel.ContainerMetadata{
 				Name:   in.Metadata.Name,
@@ -295,6 +298,7 @@ func ConvertContainerDocToInternal(in ext.ContainerDoc) (intmodel.Container, err
 				CNIConfigPath:          in.Spec.CNIConfigPath,
 				RestartPolicy:          in.Spec.RestartPolicy,
 				Attachable:             in.Spec.Attachable,
+				Tty:                    convertContainerTtyToInternal(in.Spec.Tty),
 			},
 			Status: intmodel.ContainerStatus{
 				Name:         in.Status.Name,
@@ -353,6 +357,7 @@ func BuildContainerExternalFromInternal(in intmodel.Container, apiVersion ext.Ve
 				CNIConfigPath:          in.Spec.CNIConfigPath,
 				RestartPolicy:          in.Spec.RestartPolicy,
 				Attachable:             in.Spec.Attachable,
+				Tty:                    buildContainerTtyExternalFromInternal(in.Spec.Tty),
 			},
 			Status: ext.ContainerStatus{
 				Name:         in.Status.Name,
@@ -414,6 +419,7 @@ func convertContainerSpecToInternal(in ext.ContainerSpec) intmodel.ContainerSpec
 		CNIConfigPath:          in.CNIConfigPath,
 		RestartPolicy:          in.RestartPolicy,
 		Attachable:             in.Attachable,
+		Tty:                    convertContainerTtyToInternal(in.Tty),
 	}
 }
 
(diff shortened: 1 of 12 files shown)


eminwux:
**PR #146 Review — `feat(api): add tty section on attachable containers + cell.tty.default (#139)`**

Solid implementation: validation correctly enforces the `attachable: true` gate (both standalone and nested-in-cell), `cell.tty.default` is checked against the attachable container set, and the round-trip + profile-rendering tests exercise the populated, empty, and absent-block cases. Smoke and AC verification look thorough; no sensitive-data leaks.

A couple items:

- **Dead exported constant** `internal/ctr/attachable.go:69`: `AttachableProfilePath` is declared and self-documented but never referenced anywhere in the tree — the wrapper uses `--profiles-dir AttachableTTYDir` + `--profile AttachableProfileName`, and `writeSbshProfile` builds the host path from `dir + AttachableProfileFile`. Either wire it through a consumer or drop the constant; otherwise it stays as exported dead API.
- **Comment contradicts the code** `internal/ctr/attachable.go:135-137`: the docstring says "Profile flags are global on `sbsh`, so they sit before the `terminal` subcommand — sbsh reports `unknown flag` if `--profiles-dir`/`--profile` are passed after `terminal`," but the code below correctly puts `--profile` *after* `terminal` (line 158). Only `--profiles-dir` is global; `--profile` is a per-subcommand flag. A future reader trusting the comment could "fix" the placement and break attach.

LGTM once the dead constant is resolved — the comment fix is non-blocking but worth bundling.

eminwux:
Addressed in 99c9291:
- Dropped the unused exported `AttachableProfilePath` constant (and updated the `AttachableProfileFile` doc comment that referenced it). sbsh resolves the profile via `--profiles-dir` + `--profile`, so the explicit in-container path was never needed.
- Fixed the `withAttachableArgsWrap` doc comment to match the code: `--profiles-dir` is a global flag (before `terminal`), `--profile` is a per-subcommand flag (after `terminal`). Also corrected the same misstatement in this PR's Summary.

`go vet ./...` and `make test` clean.

eminwux:

eminwux:
Rebased onto origin/main (now ad05776). Conflict resolved in pkg/api/model/v1beta1/container.go — main's new WorkingDir field merged into ContainerSpec; gofmt re-aligned the struct tags. No tty-related code touched in the rebase.

`go vet ./...` and `make test` clean. Skipped re-running the full smoke test since the rebase only added an unrelated field; existing AC verification still holds.

eminwux:
**PR #146 Re-review — `feat(api): add tty section on attachable containers + cell.tty.default (#139)`**

Rebased clean against `main`; commit `ad05776e` drops `AttachableProfilePath` and rewrites the `withSbshAttachWrapper` flag-placement comment to correctly distinguish the global `--profiles-dir` from the per-subcommand `--profile`. Both items from the prior review are resolved, and a fresh dead-symbol sweep on the new head confirms every exported identifier introduced by the PR has a caller. LGTM.

PR #150, opened 2026-04-28
feat(cli): kuke run -a / --attach (phase 2a of #141)

### LINKED ISSUE #141: feat(cli): kuke run -a / --attach — one-shot create + start + attach (phase 2a)
## Context

Extends \`kuke run\` (phase 1c, #B) so a single command starts a cell and drops the user into the attached terminal — the workflow agent operators run dozens of times a day. Combined with \`-p\` (phase 2b, #D), this delivers the headline UX: \`kuke run -p claude-cell -a\`.

## Proposal

Add \`-a/--attach\` flag to \`kuke run\`. After start completes, drive the in-process attach loop via the existing \`sbsh/pkg/attach\` integration in \`cmd/kuke/attach/\`.

**Container selection precedence:**

1. \`--container <id>\` flag, if present
2. \`cell.spec.tty.default\` from the CellDoc (#A, phase 1b)
3. The first container in declaration order with \`attachable: true\`

**Error cases:**

- Zero attachable containers: \`cell <name> has no attachable container; declare 'attachable: true' on one container or omit -a\`
- \`--container <id>\` names a non-attachable or non-existent container: clean error naming the available attachable containers

## Acceptance criteria

- [ ] \`kuke run -f attachable-cell.yaml -a\` launches the cell and the user lands in the attachable container's shell
- [ ] Container selection follows the documented precedence; tested for each branch (flag → tty.default → first-attachable)
- [ ] Detach (whatever sbsh's binding is) returns control to the host shell with the cell still \`Ready\` — \`kuke get cells\` confirms
- [ ] Zero-attachable and bad-container-flag cases produce clean errors, not crashes
- [ ] Unit test exercises selection logic with mock controller (mirrors \`cmd/kuke/attach/attach_test.go\` style)
- [ ] Smoke test per CLAUDE.md verifies the attach round-trip end-to-end

## Dependencies

Hard: #B (phase 1c, the \`kuke run\` verb) and #A (phase 1b, \`cell.tty.default\` selection rule).

## Notes

Phase 2a of the \`kuke run\` epic. Pairs with \`-p\` (#D, phase 2b) for the headline \`kuke run -p <profile> -a\` flow.

eminwux:
## Summary
- Adds `-a/--attach` and `--container` flags to `kuke run`. After the cell reaches Ready (fresh create or already-Ready short-circuit), the in-process sbsh attach loop drives the user into the cell's attachable terminal — no separate `kuke attach` round-trip needed.
- Selection precedence implemented per the AC: `--container <id>` > `cell.tty.default` > first container in declaration order with `attachable: true`.
- Selection runs against the resolved CellDoc (the spec we just told the daemon to create), so a no-attachable / unknown-container / non-attachable error happens before the daemon RPC and names the available attachable containers in the message.
- Two compatibility guards: `--container` without `-a` errors (intent-clarifying), and `-o json|yaml` with `-a` errors (the attach loop hijacks the terminal, so structured output would be unparseable).
- The `cmd/kuke/attach/` package is left untouched: `kuke run -a` reuses sbsh's `pkg/attach.Run` directly via a sibling helper in the run package, and exposes a parallel `MockRunKey` so tests bypass the real TTY.

## Notable judgment calls
- **First-attachable picker does not exclude root.** The AC literally reads "first container in declaration order with `attachable: true`", and the existing scheme validator allows `cell.tty.default` to name a root container as long as it's `attachable: true`. Mirroring that here keeps the two precedence paths consistent. (`kuke attach`'s auto-pick *does* exclude root — that's a separate pre-existing rule for the standalone verb.) Reviewer can call out if the AC should be tightened.
- **Selection runs on `cellDoc.Spec`, not a fresh `ListContainers`.** By the time we attach, the doc is what we told the daemon to create, so it's authoritative for the attachable set; saves an RPC and avoids race windows. The daemon still validates Attachable on `AttachContainer`, so the worst case is a clean error if drift happens.
- **`-o` + `-a` rejected up-front.** Could let JSON print before the attach loop, but the result is garbled / unparseable. Cleaner to make callers pick a mode.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` passes (full project test target)
- [x] `go test ./cmd/kuke/run/...` covers all selection branches (explicit > tty.default > first-attachable, no-candidate, unknown-container, non-attachable explicit, attach-after-create, attach-after-Ready short-circuit, `--container` without `-a`, `-o` + `-a`)
- [x] Pre-commit hooks pass (golangci-lint, addlicense, go fmt, go-mod-tidy)
- [x] Daemon-parity smoke check: `sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` return identical output (confirms the rebuilt kuke binary still talks to the running daemon correctly; this PR doesn't touch daemon-read paths)
- [x] `./kuke run --help` shows `-a/--attach` and `--container` flags with the documented precedence in the help text
- [x] Early-validation guards fire before daemon contact: `--container` without `-a` and `-o json` with `-a` both error before any RPC
- [ ] **End-to-end attach round-trip on a live host with sbsh staged** — the in-process sbsh loop opens a TTY and connects to the per-container control socket, so the round-trip needs an interactive terminal and a freshly-`kuke init`-ed host with `sbsh` in the cache. Unit tests cover selection + the AttachContainer wiring; the actual `pkg/attach.Run(opts)` integration is the same one `kuke attach` already uses in production. Reviewer/maintainer should verify on a live host before merge.

Closes #141

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -347,6 +347,10 @@ var (
 	KUKE_RUN_SPACE = DefineKV("KUKE_RUN_SPACE", "kuke/run/space", "default")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_RUN_STACK = DefineKV("KUKE_RUN_STACK", "kuke/run/stack", "default")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_RUN_ATTACH = DefineKV("KUKE_RUN_ATTACH", "kuke/run/attach")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_RUN_CONTAINER = DefineKV("KUKE_RUN_CONTAINER", "kuke/run/container")
 
 	// Attach command variables
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
--- a/cmd/kuke/run/attach.go
+++ b/cmd/kuke/run/attach.go
@@ -1,0 +1,163 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package run
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/eminwux/kukeon/internal/errdefs"
+	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
+	"github.com/eminwux/sbsh/pkg/attach"
+	"github.com/spf13/cobra"
+)
+
+// MockRunKey is used to inject a mock runFn via context in tests, so the
+// real pkg/attach.Run (which would open a TTY and connect to a real
+// control socket) is bypassed.
+type MockRunKey struct{}
+
+// runFn drives the in-process sbsh attach loop. Returns nil on a clean
+// detach / context cancel and any unrecoverable controller error otherwise.
+type runFn func(ctx context.Context, opts attach.Options) error
+
+// pickAttachTarget resolves the container `kuke run -a` should attach to,
+// applying the documented precedence:
+//
+//  1. explicit --container flag, when set
+//  2. cell.tty.default, when the spec sets it
+//  3. the first container in declaration order with attachable=true
+//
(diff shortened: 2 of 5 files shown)


eminwux:
**PR #150 Review — `feat(cli): kuke run -a / --attach (phase 2a of #141)`**

Clean phase-2a slice. Selection precedence (`--container` > `cell.tty.default` > first-attachable) is correct, the `-o`+`-a` and `--container`-without-`-a` guards fail before any daemon RPC, and the `MockRunKey` seam keeps the in-process attach loop covered without opening a real TTY. License headers and SPDX in place; no sensitive data leaks.

Two non-blocking items:

- **Ready short-circuit picks from in-memory spec** `cmd/kuke/run/run.go:160-167`: on the already-Ready path, `attachAfterRun` resolves the target against `cellDoc.Spec` (just-parsed YAML), not `pre.Cell.Spec` (what the daemon actually has). The "the doc is what we told the daemon to create" rationale only applies to the create path. If the YAML drifts from the existing cell, the picker can name a container the daemon doesn't have — produces a confusing `ErrContainerNotFound` instead of a clean `ErrAttachNoCandidate`. Consider passing `pre.Cell` on the Ready path.
- **`ErrContainerNotFound` wrap drops `%w`** `cmd/kuke/run/attach.go:128`: returns `fmt.Errorf("container %q not found in cell %q", ...)` without `%w err`, breaking `errors.Is` upstream. Mirrors `cmd/kuke/attach/attach.go:129` so not a regression — worth a paired cleanup to keep sentinels preserved across the call boundary.

LGTM. The unchecked E2E live-host check is the maintainer's pre-merge gate per the author's own note — not a diff-level blocker.

PR #155, opened 2026-04-28
fix(runner): make StartCell idempotent on already-running cells

### LINKED ISSUE #149: controller/runner: CreateCell→StartCell on already-Ready cell fails with CNI duplicate allocation
## Repro

On `origin/main` at \`6601f9a\` (the branch this was reproduced on, before opening #148):

```bash
# First call provisions and starts the cell.
sudo ./kuke create cell smoke --realm default --space default --stack default
# (succeeds; cell ends up Ready)

# Second call against the same cell name — same flags, no edits.
sudo ./kuke create cell smoke --realm default --space default --stack default
```

## Observed

```
Error: failed to start cell containers: failed to attach root container default_default_smoke_root to network: plugin type="bridge" failed (add): failed to allocate for range 0: 10.88.0.10 has been allocated to default_default_smoke_root, duplicate allocation is not allowed
```

The same failure also reproduces on the `kuke run -f` path (#140 / #148), because internally both go through `controller.Exec.CreateCell`, which calls `runner.EnsureCell` followed by an unconditional `runner.StartCell`. On a cell that's already Ready, the runner re-runs the bridge CNI ADD against the live container and the IPAM plugin (host-local) refuses the duplicate allocation.

## Root cause

`internal/controller/create_cell.go:170-174`:

```go
// Start cell containers (both new and existing cells need to be started)
resultCell, err = b.runner.StartCell(resultCell)
if err != nil {
    return res, fmt.Errorf("failed to start cell containers: %w", err)
}
```

`runner.StartCell` is invoked unconditionally — even when `EnsureCell` returned an already-Ready cell whose root container is running and already attached to its CNI network. There is no guard like \"if root container task is already running and the netns is wired, skip CNI ADD.\" The host-local IPAM plugin then sees an existing reservation for the same container ID and returns the duplicate-allocation error.

## Suggested fix

Make `runner.StartCell` (or the call site in `controller.Exec.CreateCell`) idempotent against already-running containers:

- Inspect the existing containerd task for the root container; if it is already in a running state and the netns is already attached, skip CNI ADD entirely.
- Alternatively, before calling CNI ADD, check the host-local IPAM store for an existing reservation under the same container ID and treat it as success.

The first option is cheaper and matches the controller-level idempotency contract the verb already advertises (\"existing cells need to be started\" — but not re-attached). PR #148 works around this on the cmd side for `kuke run -f` by short-circuiting on `Status.State == Ready`, but the underlying daemon bug should still be fixed so `kuke create cell` and any future re-entrant create-and-start path stop tripping on it.

eminwux:
## Summary
- `controller.Exec.CreateCell` calls `runner.EnsureCell` followed by an unconditional `runner.StartCell`. When `EnsureCell` returned an already-Ready cell, `StartCell` would tear down the running root container, recreate it with the same containerd ID, then re-run CNI ADD — which host-local IPAM rejects (`10.88.0.10 has been allocated to ..., duplicate allocation is not allowed`) because we never run CNI DEL on the teardown path, so the reservation persists.
- Add an idempotency guard at the top of `StartCell`: if the root container's task and every declared non-root container's task are already in `containerd.Running`, refresh container statuses, mark the cell `Ready`, and return — skip the destructive teardown-and-recreate cycle entirely.
- Decision pinned in tests: the guard is intentionally strict ("every task running"); a wedged cell with a stopped peer or a missing task still falls through to the destructive recovery path so `kuke create cell` can still bring a partially-broken cell back. See the new `TestCellTasksAllRunningFn` cases.

## Notable judgment calls
- `cellTasksAllRunningFn` takes a status closure rather than calling `r.ctrClient.TaskStatus` directly — keeps the decision logic unit-testable without standing up a containerd fake. The wiring in `StartCell` passes `r.ctrClient.TaskStatus` as the closure.
- The guard explicitly does **not** trust the persisted `cell.Status.State` (which could be stale across daemon restarts); it asks containerd for the live task status.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test \$(go list ./... | grep -v /e2e)` passes (full unit suite)
- [x] CLAUDE.md smoke test: rebuild `kuke`/`kukeond`, rebuild + import `kukeon-local:dev`, tear down and re-`kuke init`. Daemon-parity check (`kuke get realms` vs `kuke get realms --no-daemon`) is identical.
- [x] Reproduce #149 directly: two consecutive `sudo ./kuke create cell smoke --realm default --space default --stack default` calls. Before this PR the second one returned the duplicate-allocation error from the issue; now both succeed (second prints `metadata: already existed`, `containers: started`) and the root task PID is unchanged across the second call (`ctr -n kukeon.io tasks ls` confirms RUNNING with the same PID).

Closes #149

### DIFF
--- a/internal/controller/runner/start.go
+++ b/internal/controller/runner/start.go
@@ -73,6 +73,40 @@ func validateExplicitRootHostNetwork(cell intmodel.Cell, rootSpec intmodel.Conta
 	return nil
 }
 
+// cellTasksAllRunningFn is StartCell's idempotency guard, decoupled from
+// any specific containerd client so it can be unit-tested without a live
+// runtime. statusFn is called once for the root containerd ID and once
+// per non-root container; the cell is considered already running only if
+// every call returns containerd.Running. Issue #149: without this guard,
+// CreateCell→StartCell on an already-Ready cell tears the running root
+// container down and re-runs CNI ADD against its recreated peer, which
+// host-local IPAM rejects as a duplicate allocation under the same
+// container ID (we never ran CNI DEL on the teardown path).
+func cellTasksAllRunningFn(
+	cell intmodel.Cell,
+	rootContainerdID string,
+	statusFn func(id string) (containerd.Status, error),
+) bool {
+	rootStatus, err := statusFn(rootContainerdID)
+	if err != nil || rootStatus.Status != containerd.Running {
+		return false
+	}
+	for _, c := range cell.Spec.Containers {
+		if c.Root {
+			continue
+		}
+		cid := strings.TrimSpace(c.ContainerdID)
+		if cid == "" {
+			return false
+		}
+		s, statusErr := statusFn(cid)
+		if statusErr != nil || s.Status != containerd.Running {
+			return false
+		}
+	}
+	return true
+}
+
 // StartCell starts the root container and all containers defined in the CellDoc.
 // The root container is started first, then all containers in doc.Spec.Containers are started.
 func (r *Exec) StartCell(cell intmodel.Cell) (intmodel.Cell, error) {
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #155 Review — `fix(runner): make StartCell idempotent on already-running cells`**

Targeted, well-tested fix for #149. The guard sits after namespace setup and `containerID` derivation, so live `TaskStatus` calls hit the right namespace; the strict every-task-running check correctly preserves the destructive recovery path for partially wedged cells. Tests cover the meaningful permutations (no peers, all peers running, peer stopped, empty `ContainerdID`, root status error, explicit-root entry skipped). No sensitive data in logs.

LGTM.

Issue #119, opened 2026-04-27
e2e: re-run TestKuke_AttachDetach_KeepsTaskRunning once socket-staging fix lands


eminwux:
Surfaced during review of #115 (in-process attach refactor).

`make e2e TestKuke_AttachDetach_KeepsTaskRunning` currently fails on `origin/main` and on the PR branch with `sbsh socket … did not appear within 10s` in `waitForSocket`, **before** `kuke attach` is invoked. So the existing failure is unrelated to the attach refactor.

But #115 is the first change that actually flows the detach path through `sbsh/pkg/attach.Run` instead of `syscall.Exec` into an on-host `sb`. Once the socket-staging issue lands, re-run this e2e test to confirm the in-process raw-mode setup + Ctrl+] Ctrl+] detach filter + clean exit actually behaves end-to-end. The unit tests cover the seam wiring but can't observe TTY-level behavior.

eminwux:
**Socket-staging fix has landed.** Commit `4931d0c` ("align attachable wrapper with renamed sbsh terminal flags so the per-container socket binds and waitForSocket no longer times out") looks like the underlying fix this issue was waiting on. Picker should re-run `make e2e TestKuke_AttachDetach_KeepsTaskRunning` and close if green.

Tracked under #208.

eminwux:
Kuketty phases 1b/2/3/4 (#413, #415, #417, #420) have landed and replaced sbsh as the per-container socket producer. The original premise — *"re-run once the socket-staging fix lands"* — is now stale.

Can you re-run `make e2e TestKuke_AttachDetach_KeepsTaskRunning`?

- If it passes → close this issue.
- If it fails → refile against the kuketty injection path (`internal/ctr/attachable.go`) with the current failure mode, and close this one as outdated.

Note: `e2e/e2e_kuke_attach_test.go` still has "sbsh socket" wording in its comments (lines 164, 171, 286) — those should be refreshed to "kuketty socket" as part of either outcome.

PR #156, opened 2026-04-28
fix: treat sbsh clean-detach error as exit 0 in kuke attach

### LINKED ISSUE #147: e2e: TestKuke_AttachDetach_KeepsTaskRunning exits 1 after clean detach (read/write routines exited)
## Repro
On `origin/main` (commit `0a7b433`):

```bash
sudo /usr/local/go/bin/go test -run TestKuke_AttachDetach_KeepsTaskRunning -timeout 600s ./e2e -count=1
```

Reproduced twice this session — once with #146 applied, then with the branch stashed back to clean `origin/main`. Same failure both times, so it is unrelated to that PR.

## Observed
The socket-staging path now succeeds (so this is a different failure mode from the closed #116), `pkg/attach` connects, and the Ctrl+] Ctrl+] detach filter cleanly tears down — but `kuke attach` exits 1 instead of 0:

```
=== RUN   TestKuke_AttachDetach_KeepsTaskRunning
    e2e_kuke_attach_test.go:206: kuke attach exited with code 1 (want 0)
        output:
        [96mTo detach, press ^] twice[0m
        export PS1="(sbsh-$SBSH_TERM_ID) $PS1"
        ^]^G[93m
        Detached[0m
        Error: close requested: read/write routines exited
--- FAIL: TestKuke_AttachDetach_KeepsTaskRunning (14.0s)
```

Note the `Detached` banner prints (the filter fired), then `pkg/attach.Run` returns the `read/write routines exited` error which `kuke attach` surfaces as exit 1.

## Root cause

- `cmd/kuke/attach/attach.go` — the layer that wraps `sbsh/pkg/attach.Run`. The clean-detach path likely returns a non-nil sentinel (`io.EOF` or similar) that the wrapper treats as an error rather than a normal exit.
- `sbsh/pkg/attach` itself (vendor / module dep) — its `Run` may emit a non-nil error on the user-initiated detach path that callers are expected to filter.
- Any post-detach cleanup in `cmd/kuke/attach` that races the read/write goroutine teardown.

This is precisely the "verify in-process detach behaves end-to-end" check that #119 was filed to do — #119 is a re-run TODO from #115's review, not a bug report, so this is the first issue describing the actual failure mode now that socket-staging (#116) is unblocked.

## Suggested fix
Have the attach wrapper distinguish a clean user-initiated detach (Ctrl+] Ctrl+]) from an unexpected I/O teardown and return nil on the former. The detach filter already has the signal — we just need the exit path to honor it instead of bubbling the goroutine-shutdown error. A follow-up should fold the verification into #119 once the wrapper's exit semantics are settled.

Cross-refs: #119 (open re-run TODO), #116 (closed, different failure mode).

eminwux:
## Summary
- `kuke attach` was exiting 1 after a successful Ctrl+] Ctrl+] detach, surfacing `Error: close requested: read/write routines exited` even though the workload kept running. The `pkg/attach` detach filter already fired and printed `Detached`, but the embedded read/write goroutines tearing down then surfaced as a non-nil error from `pkg/attach.Run` and bubbled to the cobra runner.
- Wrap that single return in `cmd/kuke/attach/attach.go:runAttach` with an `isCleanDetach(err)` check that recognises the `close requested: ...read/write routines exited` signature emitted by sbsh's controller after the detach RPC fires (or the remote terminal closes the connection from its side). Other failures — context-done, RPC errors, socket missing, wait-on-ready — arrive without that pair of substrings and still bubble up unchanged, so this only suppresses benign session-end signals.

## Notable judgment calls
- **String match on the wrapper side, not a sentinel.** The relevant errors live behind `github.com/eminwux/sbsh/internal/errdefs` and the inner `errors.New("read/write routines exited")` is constructed inline in sbsh's `clientrunner/io.go`. Neither is exported via sbsh's public `pkg/errors`, so an `errors.Is` branch is not available against either layer. The matched substrings are stable in sbsh v0.8.0 and the suppression is conservative (both have to be present), so a real failure outside this clean-detach path still surfaces. A follow-up upstream sentinel would let this collapse to one line — left for a separate PR rather than bundled here.
- **Treat any `close requested: …read/write routines exited` as benign**, not just the user-keystroke case. The same goroutine-teardown path fires when the remote workload exits or the peer hangs up; from `kuke attach`'s perspective those are also clean session ends and exit 0 is the correct UX, matching what e.g. `ssh` returns when the remote shell exits cleanly.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` passes (the project's CI test target — CI does not run e2e)
- [x] `go test ./cmd/kuke/attach/... -race -count=1` — new `TestAttach_RunReturnsCleanDetachError` covers the regression (exit 0 when `pkg/attach.Run` returns the matching error) and `TestAttach_RunReturnsRealError` pins the negative case (a non-matching error still bubbles via `errors.Is`)
- [x] Pre-commit hooks pass (golangci-lint, addlicense, go fmt, go-mod-tidy)
- [ ] **End-to-end `TestKuke_AttachDetach_KeepsTaskRunning` rerun** — the e2e harness needs sudo + a fresh kukeond image build + a real PTY (`go test … ./e2e -count=1` under root); not feasible from this non-interactive session. The unit tests cover the seam and the failure mode is fully reproduced from sbsh's source: `pkg/attach.Run` -> `errdefs.ErrCloseReq` wrap -> the inline `errors.New("read/write routines exited")`. Reviewer should run the original repro on a live host; this also closes out the open re-run TODO in #119.

Closes #147

### DIFF
--- a/cmd/kuke/attach/attach.go
+++ b/cmd/kuke/attach/attach.go
@@ -135,12 +135,36 @@ func runAttach(cmd *cobra.Command, _ []string) error {
 	}
 
 	run := resolveRun(cmd)
-	return run(cmd.Context(), attach.Options{
+	if runErr := run(cmd.Context(), attach.Options{
 		SocketPath: result.HostSocketPath,
 		Stdin:      os.Stdin,
 		Stdout:     os.Stdout,
 		Stderr:     os.Stderr,
-	})
+	}); runErr != nil && !isCleanDetach(runErr) {
+		return runErr
+	}
+	return nil
+}
+
+// isCleanDetach reports whether err returned by the in-process attach
+// loop describes a normal session end rather than an unrecoverable
+// failure. The Ctrl+] Ctrl+] detach filter inside sbsh's pkg/attach
+// triggers a controller-level Detach RPC; the embedded read/write
+// goroutines then exit cleanly, but their teardown is reported as the
+// error "close requested: read/write routines exited" — sbsh does not
+// expose a sentinel for either layer in its public pkg/errors. The
+// remote terminal closing the connection (workload exited, peer hung
+// up) ends up on the same path, which from `kuke attach`'s perspective
+// is also a benign session end. Surface a clean exit code 0 in those
+// cases; any other failure (socket missing, RPC error, context done)
+// arrives without this signature and still bubbles up.
+func isCleanDetach(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	return strings.Contains(msg, "close requested") &&
+		strings.Contains(msg, "read/write routines exited")
 }
 
 // pickAttachableContainer enumerates the cell's containers and returns the
--- a/cmd/kuke/attach/attach_test.go
+++ b/cmd/kuke/attach/attach_test.go
@@ -75,6 +75,18 @@ func (r *runCapture) fn(_ context.Context, opts sbshattach.Options) error {
 	return nil
 }
 
+// runReturning is the mirror of runCapture for tests that want pkg/attach.Run
+// to return a specific error (rather than the clean-detach default).
+type runReturning struct {
+	calls int
+	err   error
+}
+
+func (r *runReturning) fn(_ context.Context, _ sbshattach.Options) error {
+	r.calls++
+	return r.err
+}
+
 func newCmdWithCtx(t *testing.T, fc *fakeClient, run *runCapture) *cobra.Command {
 	t.Helper()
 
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #156 Review — `fix: treat sbsh clean-detach error as exit 0 in kuke attach`**

LGTM. Targeted fix with sound rationale: traced sbsh v0.8.0's event flow (EvDetach → Detach → copy goroutines exit → CopierManager fires EvError("read/write routines exited") → onClosed → Close → closeReqCh wraps with `errdefs.ErrCloseReq`), and the final error string from `pkg/attach.Run` does carry both substrings the wrapper matches against. Confirmed `ErrCloseReq` is not re-exported via sbsh's public `pkg/errors`, so substring match is the right call until sbsh exposes a sentinel. Conservative AND of two substrings keeps real failures (RPC, dial, context) bubbling up — both new unit tests pin that seam. License headers, terse style, and tests-colocated all in order.

The unchecked e2e item in the test plan is acknowledged in the PR body and would need a live-host repro to flip, but the unit tests reproduce the exact wrapped-error string from sbsh's source so the fix is well-pinned at the seam.

PR #153, opened 2026-04-28
feat(api+cli): CellProfile kind + kuke run -p

eminwux:
## Summary
- Adds `kind: CellProfile` under `pkg/api/model/v1beta1/` (parallel to existing kinds), with apiVersion / Kind / Metadata / Spec where Spec carries `realm` / `space` / `stack` plus a `cell:` body that's a CellSpec verbatim.
- Adds `internal/cellprofile/` — a per-user loader that reads `$HOME/.kuke/profiles.d/*.yaml` (configurable via `KUKE_PROFILES_DIR`), scans by filename then falls back to `metadata.name`, and materializes a CellProfile into a CellDoc that walks the same path as `kuke run -f`.
- Wires `kuke run -p <profile> [<cell-name>]`: positional arg overrides the materialized cell name (defaults to `metadata.name`); `--realm`/`--space`/`--stack` flags layer over an unset profile location, matching the existing `-f` "doc wins" rule; bash/zsh autocomplete on `-p` lists profile names; `-f` and `-p` are mutually exclusive (cobra `MarkFlagsMutuallyExclusive` + `MarkFlagsOneRequired`); unknown profiles return `profile "<name>" not found in <dir>` wrapping `errdefs.ErrProfileNotFound`.
- Live smoke (per AC): dropped a profile, ran `kuke run -p smoke-cell smoke-test` against the in-flight daemon — cell created and started; second invocation reports the idempotent already-existed path; `kuke run -p ghost` returns the spec'd not-found message; `kuke run -p smoke-cell -f /tmp/x` rejects with the cobra mutex.

## Notable judgment calls
- Profile spec uses user-facing `realm` / `space` / `stack` strings (not `realmId`/`spaceId`/`stackId`), so a profile is portable between hosts; materialization rewrites them onto `CellSpec.RealmID`/`SpaceID`/`StackID`.
- `cmd/kuke/apply` is intentionally untouched: `CellProfile` is a per-user-only kind (issue says system-wide is a possible follow-up), so `kuke apply -f` rejecting it via the existing `ErrUnknownKind` path is the desired UX rather than a regression.
- Loader's `List()` silently skips unparseable files; the operator only finds out when they actually try to use a broken profile via `Load()`. Tracking-not-failing keeps autocomplete clean when one stale file in `profiles.d/` would otherwise blank the completion list.
[...]

PR #158, opened 2026-04-28
feat(api+cli): add CellProfile.spec.namePrefix and inject profile label

### LINKED ISSUE #157: cellprofile: namePrefix for templated cells + profile label injection
## Context

PR #153 shipped `CellProfile` and `kuke run -p`, but the materializer always names the cell after `metadata.name`. Re-running `kuke run -p myprof` hits the idempotent already-exists path from PR #155 instead of producing a new cell — which contradicts the "template for many cells" model that motivates profiles in the first place (sbsh-style, K8s `Deployment` template-style).

Today there is no way to express "every invocation should produce a fresh instance of this template." There is also no first-class way to query "all cells produced from profile X" — the closest you can get is a name-prefix grep on `kuke get cells`.

## Proposal

Two coupled changes to `pkg/api/model/v1beta1/cellprofile.go` and `internal/cellprofile/cellprofile.go`:

**1. Optional `spec.namePrefix` on `CellProfileSpec`.**

- Profile has no `namePrefix` → cell name = `metadata.name`. Singleton, idempotent on re-run. Current behavior, unchanged.
- Profile has `namePrefix` → cell name = `<namePrefix>-<6hex>`. Fresh cell every invocation. Suffix is 6 lowercase hex chars from `crypto/rand`.

**2. Profile label injection.**

Every materialized cell gets `kukeon.io/profile=<metadata.name>` injected into `CellSpec.Metadata.Labels`. Enables `kuke get cells -l kukeon.io/profile=myprof` to list all instances of a profile, which is the queryable form of "deploy multiple cells with the same configuration."

**3. Drop the positional `<cell-name>` override on `kuke run -p`.**

PR #153 added `kuke run -p <profile> <cell-name>` to override the materialized name. Drop it — `kuke run` is for *creating* cells; re-attaching to a known cell is `kuke attach <cell>`'s job. The positional was conflating two verbs. This is the only breaking change in this issue, and PR #153 is one day old with essentially zero adoption.

## Acceptance criteria

- [ ] `spec.namePrefix` field added to `CellProfileSpec` with omitempty YAML/JSON tags
- [ ] `Materialize`: when `namePrefix` is set, generated name is `<namePrefix>-<6 lowercase hex>`; when unset, falls back to `metadata.name` (current behavior)
- [ ] Suffix entropy comes from `crypto/rand`, not `math/rand` or timestamps
- [ ] Every materialized cell carries `kukeon.io/profile=<metadata.name>` label, merged with any labels already set on the profile's `cell.metadata.labels`
- [ ] Positional `<cell-name>` argument removed from `kuke run -p`; help text updated; test cases that exercised the override deleted or rewritten
- [ ] Three successive `kuke run -p <profile-with-prefix>` invocations produce three distinct cells with the same prefix; `kuke get cells -l kukeon.io/profile=<name>` lists all three
- [ ] A profile *without* `namePrefix` still creates the same singleton cell on re-run (current behavior preserved)
- [ ] CLAUDE.md daemon-parity check (`kuke get realms` vs `--no-daemon`) noted in the test plan even though this is client-side

## Notes

- Followup to PR #153 (closes #142), motivated by the gap that `kuke run -p` always materializes a singleton.
- Mirrors K8s `metadata.generateName` precedent (lowercase alphanumeric suffix appended to a base) and sbsh's "every `sbsh start` is a fresh session" model.
- Composes with `-a` (PR #150): `kuke run -p claude -a` from a profile with `namePrefix: claude` lands the operator in a fresh `claude-abc123` cell every invocation — this is the headline UX the profile epic was aiming at.
- Does not touch `kuke apply` — `CellProfile` remains a per-user-only kind, consistent with #142's scope decision.

eminwux:
## Summary
- `CellProfileSpec.NamePrefix` (omitempty) flips a profile from singleton to template: when set, every `kuke run -p` invocation materializes a fresh `<NamePrefix>-<6 lowercase hex>` cell using `crypto/rand`; when unset, the existing `metadata.name` singleton path is preserved.
- Every materialized cell now carries the system label `kukeon.io/profile=<profile metadata.name>`, merged with any user-set labels (the system label wins on conflict on the reserved key). Enables `kuke get cells -l kukeon.io/profile=<name>` to enumerate every instance produced from a profile.
- Drops the positional `<cell-name>` override on `kuke run -p` per the issue's "verb separation" rationale: `kuke run` creates cells, `kuke attach <cell>` re-attaches. Cobra `Args` switches from `MaximumNArgs(1)` to `NoArgs`. The override was one day old (PR #153) with no adoption, the only breaking change here.
- `Materialize` now returns `(CellDoc, error)` so `crypto/rand.Read` failures surface to the caller rather than panic; the cobra handler bubbles them up the same way profile load errors flow today.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full unit suite green (includes the new `TestMaterialize_NamePrefix_GeneratesUniqueCells`, `TestMaterialize_LabelsMerged`, `TestMaterialize_DefaultName_Idempotent`, and `TestRun_FromProfile_NamePrefix_GeneratesFreshCell`).
- [x] gofmt clean on changed files.
- [ ] Local `kuke get realms` vs `--no-daemon` daemon-parity smoke check — change is purely client-side (cellprofile materializer + `kuke run` flag handling, no daemon-visible code paths). The AC asks for it noted in the test plan even when client-side, so this is the explicit acknowledgement that no daemon view changed.

Closes #157

### DIFF
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -50,17 +50,14 @@ type MockControllerKey struct{}
 // operator into the cell's attachable terminal once the cell is up.
 func NewRunCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "run (-f <file> | -p <profile> [<cell-name>])",
+		Use:   "run (-f <file> | -p <profile>)",
 		Short: "Create and start a single cell from a YAML file or a profile",
 		Long: "Create and start a single cell from a YAML file or stdin (-f), " +
 			"or from a per-user profile under $HOME/.kuke/profiles.d/<name>.yaml (-p). " +
 			"Conceptually `kuke apply -f` (single-cell) plus `kuke start cell`, but refuses " +
-			"to update a divergent on-disk spec. With -p, the optional positional argument " +
-			"overrides the materialized cell name.",
-		// MaximumNArgs(1) leaves room for the optional cell-name override
-		// `kuke run -p <profile> <cell-name>`. The handler rejects the arg
-		// when -p is not in use, so the -f form behaves as before.
-		Args:          cobra.MaximumNArgs(1),
+			"to update a divergent on-disk spec. To re-attach to an existing cell use " +
+			"`kuke attach <cell>`.",
+		Args:          cobra.NoArgs,
 		SilenceUsage:  true,
 		SilenceErrors: false,
 		RunE:          runRun,
@@ -105,26 +102,19 @@ func NewRunCmd() *cobra.Command {
 // argument validation. Splitting it out keeps runRun under the funlen limit
 // while preserving the original control flow.
 type runFlags struct {
-	file             string
-	profileName      string
-	cellNameOverride string
-	output           string
-	doAttach         bool
-	containerFlag    string
+	file          string
+	profileName   string
+	output        string
+	doAttach      bool
+	containerFlag string
 }
 
-func parseRunFlags(cmd *cobra.Command, args []string) (runFlags, error) {
+func parseRunFlags(cmd *cobra.Command, _ []string) (runFlags, error) {
 	flags := runFlags{
 		file:          strings.TrimSpace(viper.GetString(config.KUKE_RUN_FILE.ViperKey)),
 		profileName:   strings.TrimSpace(viper.GetString(config.KUKE_RUN_PROFILE.ViperKey)),
 		doAttach:      viper.GetBool(config.KUKE_RUN_ATTACH.ViperKey),
 		containerFlag: strings.TrimSpace(viper.GetString(config.KUKE_RUN_CONTAINER.ViperKey)),
-	}
-	if len(args) == 1 {
-		flags.cellNameOverride = strings.TrimSpace(args[0])
-	}
-	if flags.cellNameOverride != "" && flags.profileName == "" {
-		return runFlags{}, errors.New("positional <cell-name> argument is only valid together with -p/--profile")
 	}
 
 	output, err := cmd.Flags().GetString("output")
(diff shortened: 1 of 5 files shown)


eminwux:
**PR #158 Review — `feat(api+cli): add CellProfile.spec.namePrefix and inject profile label`**

Clean change. The singleton-vs-template split via `spec.namePrefix` reads well, the new `kukeon.io/profile=<name>` label is consistently injected on both code paths and survives a user-set conflict, and the `Materialize` signature change to `(CellDoc, error)` correctly bubbles `crypto/rand` failures instead of panicking. Tests cover default-name idempotency, prefix uniqueness across invocations, hex-shape, and label-merge precedence. No correctness or sensitive-data issues; the removed positional `<cell-name>` override is one PR old (#153) with no adoption, so the breaking change is justified by the verb-separation rationale. Daemon-parity smoke check is correctly noted as N/A in the test plan — `cellprofile.Materialize` and `cmd/kuke/run` both run client-side and produce no daemon-visible state change.

A couple of nits, all non-blocking:

- **Nit (consistency)** `cmd/kuke/run/run.go:144`: `runRun(cmd, args)` still takes `args` even though cobra `NoArgs` guarantees it empty and the only consumer (`parseRunFlags`) now blanks it. Could mirror `parseRunFlags`'s `_ []string` for symmetry, but the call site `parseRunFlags(cmd, args)` keeps it technically used — skip if not bothered.
- **Nit (test depth)** `internal/cellprofile/cellprofile_test.go:234`: `TestMaterialize_DefaultName_Idempotent` only asserts name equality across two calls. Asserting label/spec equality too would harden the singleton-path contract beyond just the name field.

LGTM to proceed — both items are optional polish.

PR #170, opened 2026-04-28
feat(cli): add kuke uninstall to remove all kukeon state

### LINKED ISSUE #164: cli: add kuke uninstall to remove all kukeon state
## Context
Kukeon has no command to fully remove itself. `kuke purge` is per-resource
only (cell/container/realm/space/stack subcommands). Removing kukeon today
means: `kuke purge realm --cascade` each realm, kill + delete the kukeond
cell, `rm -rf /run/kukeon /opt/kukeon`, and (after the user/group work)
`userdel kukeon && groupdel kukeon`. The order matters and the steps are
easy to get wrong.

## Proposal
Add `kuke uninstall` — runs the full removal in the right order, idempotent,
with an interactive confirmation by default.

## Acceptance criteria
- [ ] `kuke uninstall` removes all kukeon state, in this order:
  1. Purge every realm with `--cascade` (drains spaces, stacks, cells,
     containers, and their containerd tasks/containers).
  2. Stop the kukeond daemon (kill + delete the kukeond cell in
     `kuke-system.kukeon.io`).
  3. Remove the containerd namespaces created by kukeon (`kukeon.io`,
     `kuke-system.kukeon.io`) and any remaining tasks/containers/snapshots
     within them.
  4. Remove `/run/kukeon/` and `/opt/kukeon/` recursively.
  5. Remove the `kukeon` system user and group (no-op if absent — see #163).
- [ ] After completion: no kukeon-managed containerd namespaces, no kukeon
      directories under `/run` or `/opt`, no `kukeon` user/group.
- [ ] Interactive confirmation by default (`Are you sure? (yes/no)`);
      `--yes` / `-y` bypasses.
- [ ] Idempotent — safe to re-run after partial completion; each step skipped
      if already gone.
- [ ] `/usr/local/bin/kuke` and the `kukeond` symlink are NOT removed
      (uninstalling runtime state ≠ uninstalling the binary).
- [ ] Errors during a step abort the run and report which step failed.

## Notes
- Depends on #163 for the user/group creation logic — `kuke uninstall`
  mirrors and removes what `kuke init` created.
- Related: existing `kuke purge` (per-resource); `kuke uninstall` is the
  global counterpart.

eminwux:
## Summary
- Add `kuke uninstall` command that performs a comprehensive teardown of all kukeon runtime state on the host: purges every realm with `--cascade --force` (including the well-known `default` and `kuke-system` realms even if their on-disk metadata is missing, so containerd namespaces left behind by a partial uninstall still get cleaned up), removes `/run/kukeon/`, removes the configured run path (default `/opt/kukeon`), and removes the kukeon system user/group.
- Interactive confirmation by default; `--yes`/`-y` skips the prompt for scripted use. Idempotent — re-running on a clean machine is a no-op rather than an error.
- The `/usr/local/bin/kuke` binary and the `kukeond` symlink are NOT touched (per AC: "uninstalling runtime state is not the same as uninstalling the binary").
- Errors abort the run with the first step's failure but best-effort cleanup continues so a single failed realm purge does not leave `/run/kukeon` and `/opt/kukeon` behind.
- The kukeond cell living inside `kuke-system/kukeon/kukeon/kukeond` is killed and deleted as part of the `kuke-system` realm cascade — no separate "stop daemon" step is needed.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — all packages pass, including new tests in `cmd/kuke/uninstall` (5 tests covering --yes flag, prompt accept/reject, step-error propagation, default run path) and `internal/controller` (4 tests covering well-known realm purge + filesystem cleanup, idempotency on clean host, listed+well-known merge order, partial-failure best-effort cleanup).
- [x] End-to-end smoke on this host: `kuke init` (fresh provision), then `sudo ./kuke uninstall --yes` — purges both realms, removes `/run/kukeon` and `/opt/kukeon`, drops the `kukeon` user/group. Re-running `kuke uninstall --yes` on the now-clean host is a no-op as expected.
- [x] Daemon-parity check (`sudo ./kuke get realms` vs `--no-daemon`) on a fresh `kuke init` — identical output, both realms `Ready`.
- [x] Prompt accept/abort verified by typing `yes`/`no` interactively.

## Notes
- Existing pre-existing quirk surfaced during smoke: on a long-running install with active containerd snapshots in `kuke-system.kukeon.io`, the very first `uninstall --yes` can hit `failed precondition: namespace must be empty, still has snapshots on overlayfs snapshotter` because `PurgeRealm` aborts on `deleteRealmCascade` failure before `runner.PurgeRealm` clears the namespace. A second `kuke uninstall --yes` cleans this up (the command is idempotent by design). Out of scope for this issue — it is the same race that affects `kuke purge realm --cascade`.

Closes #164

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -382,6 +382,9 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_STOP_CONTAINER_CELL = DefineKV("KUKE_STOP_CONTAINER_CELL", "kuke/stop/container/cell")
 
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_UNINSTALL_YES = DefineKV("KUKE_UNINSTALL_YES", "kuke/uninstall/yes", "false")
+
 	// Kill command variables
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_KILL_CELL_REALM = DefineKV("KUKE_KILL_CELL_REALM", "kuke/kill/cell/realm", "default")
--- a/cmd/kuke/kuke.go
+++ b/cmd/kuke/kuke.go
@@ -38,6 +38,7 @@ import (
 	runcmd "github.com/eminwux/kukeon/cmd/kuke/run"
 	startcmd "github.com/eminwux/kukeon/cmd/kuke/start"
 	stopcmd "github.com/eminwux/kukeon/cmd/kuke/stop"
+	uninstallcmd "github.com/eminwux/kukeon/cmd/kuke/uninstall"
 	"github.com/eminwux/kukeon/cmd/kuke/version"
 	"github.com/eminwux/kukeon/cmd/types"
 	"github.com/eminwux/kukeon/internal/errdefs"
@@ -131,6 +132,7 @@ func SetupKukeCmd(rootCmd *cobra.Command) error {
 	rootCmd.AddCommand(runcmd.NewRunCmd())
 	rootCmd.AddCommand(attachcmd.NewAttachCmd())
 	rootCmd.AddCommand(autocompletecmd.NewAutocompleteCmd())
+	rootCmd.AddCommand(uninstallcmd.NewUninstallCmd())
 	rootCmd.AddCommand(version.NewVersionCmd())
 
 	// Persistent flags
--- a/cmd/kuke/uninstall/uninstall.go
+++ b/cmd/kuke/uninstall/uninstall.go
@@ -1,0 +1,210 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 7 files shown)


eminwux:
**PR #170 Review — `feat(cli): add kuke uninstall to remove all kukeon state`**

Clean, well-scoped addition. The new `Controller.Uninstall` method is implemented by the sole `Exec` implementor; new exported types (`UninstallOptions`, `UninstallReport`, `RealmPurgeOutcome`, `UserGroupRemover`, `UserGroupRemovalReport`) all have non-test references. Idempotency, well-known-realm fallback (so a partial uninstall whose metadata was already wiped still drops the containerd namespaces), and best-effort cleanup on partial failure are all covered by tests. License + SPDX headers present on all four new files; daemon-parity check is in the test plan. No correctness bugs or sensitive-data leaks spotted.

A few non-blocking nits:

- **Nit** `cmd/kuke/uninstall/uninstall.go:86-89`: `NewUninstallCmd` silently returns `nil` on `setupUninstallCmd` error, which would panic at the `rootCmd.AddCommand(NewUninstallCmd())` call site. The only failure mode is `viper.BindPFlag` on a nil flag (unreachable here), so it's theoretical — but propagating the error (or panicking on bind failure) would be more honest than swallowing.
- **Nit** `cmd/kuke/uninstall/uninstall.go:190-197`: `readLine` constructs a fresh `bufio.NewReader(r)` per call inside the retry loop, so any bytes the previous bufio greedily pre-buffered past the newline are dropped. For terminal stdin (line-buffered) it never bites; flagging because it surprised me on first read.
- **Nit** `cmd/kuke/uninstall/uninstall.go:216-235`: `dirOutcome` and `accountOutcome` have identical bodies — fold into one helper.

LGTM.

PR #174, opened 2026-04-28
fix: purge realm cleans up snapshots in a single pass

### LINKED ISSUE #171: controller: PurgeRealm aborts before namespace cleanup when overlay snapshots remain
## Repro
On a host with a long-running kukeon install that has produced containerd snapshots in `kuke-system.kukeon.io` (typically: `kuke init` followed by some normal lifetime — kukeond pulled images, started its cell, and accumulated overlayfs snapshot layers):

1. `sudo ./kuke uninstall --yes` (built from `feat/kuke-uninstall` / PR #170 in this session, but the underlying behavior is in `controller.PurgeRealm` and predates this PR).
2. Observe the first run errors out partway through cleaning the `kuke-system` realm.
3. `sudo ./kuke uninstall --yes` a second time — succeeds and produces a clean host.

A clean install (e.g. immediately after `kuke init` with no images pulled beyond the kukeond bootstrap) does not reproduce this; the snapshots have to be present on the overlayfs snapshotter for the precondition to fire.

## Observed
First-pass error from `runner.PurgeRealm` → containerd's `namespaces.Delete` call:

```
failed precondition: namespace must be empty, still has snapshots on overlayfs snapshotter
```

`PurgeRealm` returns this error early. The realm's containerd namespace is left behind, but `kuke uninstall`'s subsequent best-effort filesystem cleanup (`/run/kukeon`, `/opt/kukeon`) still runs, so the second pass has the metadata gone — at which point `runner.PurgeRealm` walks the namespace differently and the snapshots get cleaned up.

## Root cause

- `internal/controller/purge.go` (or wherever `PurgeRealm` orchestrates the cascade — `purgeRealmCascade`): `deleteRealmCascade` is called before `runner.PurgeRealm` clears the namespace. If `deleteRealmCascade` fails (e.g. because a stack/cell still holds a reference to a snapshot via some metadata path), the function returns without ever reaching the namespace-delete step that would have handled the snapshots.
- `internal/controller/runner/realm_purge.go` (or the runner equivalent): the snapshot-walk logic appears to need a "delete snapshots before deleting the namespace" step that is currently order-dependent on `deleteRealmCascade` having already drained the cell layer.

The clean fix is probably to make `runner.PurgeRealm` resilient to leftover snapshots — walk and remove them with the snapshotter directly before issuing `namespaces.Delete` — rather than relying on the cascade to clear them out first.

## Suggested fix
In the runner's `PurgeRealm`, before calling `namespaces.Delete`, enumerate the namespace's snapshots via the containerd snapshotter and remove each. Treat \"already gone\" as a no-op so the path remains idempotent. This decouples namespace cleanup from cascade success and makes both `kuke purge realm --cascade --force` and `kuke uninstall` single-pass on hosts with active snapshots.

eminwux:
## Summary
- `kuke purge realm --cascade --force` (and `kuke uninstall` once #170 lands) used to need a second pass on hosts with active overlayfs snapshots — the first run died with `failed precondition: namespace must be empty, still has snapshots on overlayfs snapshotter`. After this PR it's single-pass.
- Three small fixes converge on the same root cause:
  - **`runner.DeleteRealm` reorder** — orphaned-container handling moved before `CleanupNamespaceResources`, and `DeleteCgroup` moved to the end. A live task pinned its committed snapshot, so cleaning snapshots before killing the task silently failed; matching `runner.PurgeRealm`'s order eliminates that.
  - **`CleanupNamespaceResources` iterates** — snapshotter `Walk` returns boltdb-key order, not topological, so a single reverse pass can't guarantee children-before-parents when a parent has multiple children. Now iterates up to 32 passes (in practice 1–3) until either zero remain or no progress is made on a pass.
  - **`controller.purgeRealmCascade` falls through** — even if `deleteRealmCascade` errors, still call `runner.PurgeRealm`. `runner.PurgeRealm` is idempotent and goes further, so a single command leaves the host clean. If runner-purge then succeeds, the original delete error is no longer load-bearing and the operation reports success.

## Notable judgment calls
- **Pass cap of 32 in `CleanupNamespaceResources`.** Snapshot graphs in this project are shallow (image layers + a few committed container layers), so a real graph clears in 1–3 passes. The cap exists only to bound work on a pathological cycle; raising it has no normal-path impact. Hard-coded rather than configurable to avoid plumbing options through.
- **Recovery hides the `deleteRealm` error.** When `deleteRealmCascade` fails but `runner.PurgeRealm` later succeeds, the operation reports success and the original delete error is logged at WARN, not propagated. Reasoning: the user's intent for `kuke purge --force` is "clean it all up regardless"; surfacing a no-longer-load-bearing error would push them to re-run unnecessarily. Both errors are still surfaced when both fail.
- **Did not touch `runner.PurgeRealm`'s ordering.** Already correct — orphans first, then `CleanupNamespaceResources`. Issue's suggested fix ("walk and remove snapshots in `runner.PurgeRealm` before `namespaces.Delete`") is functionally what `CleanupNamespaceResources` already does; the iteration fix to that helper is the missing piece, and it benefits the `DeleteRealm` path too.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — all packages pass, including new `TestPurgeRealm_DeleteCascadeFailsButPurgeSucceeds` covering the recovery path and the updated existing case where both delete and purge fail.
- [x] `pre-commit run golangci-lint --files <changed>` passes.
- [x] CLAUDE.md smoke (full reproducer of the issue): tear down kukeond cell, rebuild `kuke`, rebuild local image, `sudo ./kuke init`, then `sudo ./kuke purge realm kuke-system --cascade --force --no-daemon` on a host with active overlayfs snapshots — single pass cleans the namespace, snapshots, and `/opt/kukeon/kuke-system/`. Verbose log line `"deleted containerd namespace" namespace=kuke-system.kukeon.io` confirms the regression is fixed.
- [x] CLAUDE.md daemon-parity smoke: `sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` return identical output before and after the smoke test.

Closes #171

### DIFF
--- a/internal/controller/purge_realm.go
+++ b/internal/controller/purge_realm.go
@@ -158,13 +158,36 @@ func (b *Exec) purgeRealmCascade(realm intmodel.Realm, force, cascade, metadataE
 		}
 	}
 
-	// Perform standard delete first (only if metadata exists, as deleteRealmCascade requires metadata)
+	// Perform standard delete first (only if metadata exists, as deleteRealmCascade requires metadata).
+	// If it fails (e.g. DeleteNamespace can't remove a namespace whose snapshots
+	// are still pinned by a live task), still fall through to runner.PurgeRealm
+	// so the comprehensive cleanup runs in a single pass — runner.PurgeRealm
+	// is idempotent and goes further (orphaned containers, CNI tear-down, etc.),
+	// so a single failed deleteRealmCascade should not strand the user with a
+	// half-cleaned realm that requires re-running the command. If runner.PurgeRealm
+	// later succeeds, the realm is fully gone and the original delete error is
+	// no longer load-bearing — log it and treat the operation as a success.
+	var deleteErr error
 	if metadataExists {
 		if err := b.deleteRealmCascade(realm, force, cascade); err != nil {
-			return fmt.Errorf("failed to delete realm: %w", err)
+			deleteErr = fmt.Errorf("failed to delete realm: %w", err)
+			b.logger.WarnContext(
+				b.ctx,
+				"deleteRealmCascade failed; falling through to runner.PurgeRealm",
+				"realm",
+				realmName,
+				"error",
+				err,
+			)
 		}
 	}
 
 	// Perform comprehensive purge via runner (works even without metadata)
-	return b.runner.PurgeRealm(realm)
+	if purgeErr := b.runner.PurgeRealm(realm); purgeErr != nil {
+		if deleteErr != nil {
+			return fmt.Errorf("%w; runner purge also failed: %w", deleteErr, purgeErr)
+		}
+		return purgeErr
+	}
+	return nil
 }
--- a/internal/controller/purge_realm_test.go
+++ b/internal/controller/purge_realm_test.go
@@ -924,7 +924,10 @@ func TestPurgeRealm_RunnerErrors(t *testing.T) {
 			errContains: "failed to purge space",
 		},
 		{
-			name:      "deleteRealmCascade error",
+			// deleteRealmCascade fails AND runner.PurgeRealm also fails — both errors are
+			// surfaced. (The single-failure case where runner.PurgeRealm recovers is
+			// covered separately in TestPurgeRealm_DeleteCascadeFailsButPurgeSucceeds.)
+			name:      "deleteRealmCascade error, purge also fails",
 			realmName: "test-realm",
 			namespace: "test-namespace",
 			force:     false,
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #174 Review — `fix: purge realm cleans up snapshots in a single pass`**

Three coherent fixes that converge on a real reproducer (#171). The DeleteRealm reorder matches the existing-and-correct ordering in `runner.PurgeRealm` (orphans → cleanup → cgroup/namespace), the iterative snapshot pass is the right shape for a topology-blind boltdb walk, and the controller fall-through plus its new test (`TestPurgeRealm_DeleteCascadeFailsButPurgeSucceeds`) cover the recovery path that #171 actually wants. CLAUDE.md smoke + daemon-parity check are both ticked. No correctness, leak, or dead-code concerns. LGTM.

Issue #163, opened 2026-04-28
init: create kukeon user/group; set runtime/socket ownership


eminwux:
## Context
`kuke init` runs as root and creates `/run/kukeon/`, `/opt/kukeon/...`, and
the kukeond unix socket at `/run/kukeon/kukeond.sock`. Today everything is
implicitly root-owned (see `internal/controller/bootstrap.go`,
`0o700`/`0o750`/`0o755` modes). That forces every `kuke` invocation to run
as root, even read-only commands like `kuke get realms`.

## Proposal
On `kuke init`:
1. Create system user `kukeon` and group `kukeon` (no login shell, no home
   dir) if they don't already exist. Idempotent — if either exists, leave it
   untouched.
2. Set ownership and modes on kukeon-managed paths:
   - `/run/kukeon/` → `root:kukeon` `0o750`
   - `/run/kukeon/kukeond.sock` → `root:kukeon` `0o660`
   - `/opt/kukeon/` (recursively) → `root:kukeon` `0o750`

Effect: a user added to the `kukeon` group (`sudo usermod -aG kukeon $USER`
+ relogin) can dial the socket and run client-side commands without sudo.
Writes under `/opt/kukeon/` still require root; they go through the daemon,
which runs as root.

`sudo kuke init` remains the entry point.

## Acceptance criteria
- [ ] On `sudo kuke init`, system user `kukeon` and group `kukeon` are
      created if absent (no login shell, no home dir). Idempotent.
- [ ] `/run/kukeon/` is `root:kukeon` mode `0o750`.
- [ ] `/run/kukeon/kukeond.sock` is `root:kukeon` mode `0o660`.
- [ ] `/opt/kukeon/` and everything under it is `root:kukeon` mode `0o750`.
- [ ] Smoke test: a non-root user added to the `kukeon` group can run
      `kuke get realms`, `kuke create realm test`, and
      `kuke delete realm test` without sudo.
- [ ] CLAUDE.md daemon-parity smoke test still passes.

## Notes
- Coordinate with #62 (one-line installer script) so the installer doesn't
  duplicate or fight this work.
- Independent of #159 / #160 / #161 / #162.

Issue #175, opened 2026-04-28
docs: extend init smoke test with daemon-restart and group-inheritance checks


eminwux:
## Context
PR for #173 adds two invariants the project CLAUDE.md smoke test does not currently cover:

1. The kukeond unix socket is `root:kukeon` mode `0o660` after a daemon **restart** (not just after `kuke init`). This is the regression the issue's first carve-out called out — `net.Listen` recreates the socket as `root:root 0o600`, so the daemon now self-applies the chown via `--socket-gid`.
2. Files the daemon writes under `/opt/kukeon` after init (e.g. `realm/metadata.json` from `kuke create realm`) inherit the `kukeon` group thanks to the SGID bit on managed directories. Without SGID, daemon-written files fall back to `root:root` and break `--no-daemon` reads for non-root operators.

Per `agents/dev/CLAUDE.md`, project CLAUDE.md edits cannot be bundled into a code-fix PR — they travel as their own issue + PR. Carving the wording out here so it does not get lost once #173 merges.

## Acceptance criteria
- [ ] `CLAUDE.md` smoke section gains a daemon-restart check that confirms the socket retains `root:kukeon 0o660` without re-running `kuke init`.
- [ ] Smoke section gains a group-inheritance check that creates a realm via the daemon and confirms the new `metadata.json` lands in group `kukeon`.

## Proposed wording (verbatim)

Add a new subsection to `CLAUDE.md` after the existing **Daemon-parity check (the regression guard)** block:

````markdown
### 6. Daemon-restart check (the socket-ownership regression guard)

`net.Listen` recreates the unix socket as `root:root 0o600`. The kukeond container args must include `--socket-gid <kukeon-gid>` so the daemon re-applies `root:kukeon 0o660` on every restart, otherwise non-root operators in the kukeon group lose access until `kuke init` is re-run.

```bash
# Snapshot the socket before the restart.
stat -c '%U:%G %a %n' /run/kukeon/kukeond.sock

# Stop+start the kukeond cell (no daemon involved — it would self-decapitate).
sudo ./kuke stop  cell kukeond --realm kuke-system --space kukeon --stack kukeon --no-daemon
sudo ./kuke start cell kukeond --realm kuke-system --space kukeon --stack kukeon --no-daemon

# Expected: still root:kukeon mode 660, NOT root:root 600.
stat -c '%U:%G %a %n' /run/kukeon/kukeond.sock

# And the daemon serve line should carry --socket-gid <kukeon-gid>.
[...]

PR #167, opened 2026-04-28
feat(init): create kukeon user/group, set runtime ownership

eminwux:
## Summary
- `kuke init` now provisions a `kukeon` system user and group (idempotent, no login shell, no home dir on disk) and applies kukeon-managed ownership/modes to `/run/kukeon`, `/opt/kukeon` (recursive), and the kukeond unix socket so a non-root operator added to the `kukeon` group can talk to the daemon without sudo.
- New `internal/sysuser` package isolates the user/group provisioning + chown helpers; tests stub `os/user.Lookup*` and the command runner so the unit suite never mutates the host.

## Notable judgment calls
- **Split file/dir mode under `/opt/kukeon`.** AC says "mode 0o750" recursively. Applying 0o750 verbatim marks JSON metadata files as executable, which has no purpose, so dirs get 0o750 (owner needs traverse) and regular files get 0o640. The init output makes this split visible: `chown root:kukeon mode 0750 (dirs) / 0640 (files)`. Easy to revert to literal 0o750 if the reviewer prefers.
- **Daemon restart resets socket ownership.** Init chowns `/run/kukeon/kukeond.sock` to `root:kukeon` mode 0o660 after `waitForKukeondReady`. If the daemon container restarts later, `net.Listen` recreates the socket as `root:root` 0o600 again — `kuke init` is idempotent, so re-running fixes it, but a future change should let the daemon self-chown via a passed-in GID. Not bundled here to keep the PR scoped to AC.
- **No SGID on `/run/kukeon` or `/opt/kukeon`.** The recursive chown after bootstrap captures everything created during init, but new files the daemon writes later land as `root:root`. SGID would make new files inherit `kukeon` group; left out for the same scope reason as the previous bullet.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test`
- [x] `pre-commit run golangci-lint --files …` (passes; matches CI config)
- [x] CLAUDE.md smoke: tear down kukeond cell, rebuild ` kuke`, rebuild local image, `sudo ./kuke init` — output ends with the new \`Permissions:\` block listing group/user create + the three chown lines.
- [x] Verified post-init filesystem state: `/run/kukeon/` and `/opt/kukeon/` are `drwxr-x--- root kukeon`; `/run/kukeon/kukeond.sock` is `srw-rw---- root kukeon`; deep `metadata.json` files are `-rw-r----- root kukeon`.
[...]

PR #176, opened 2026-04-28
feat(init+daemon): SGID kukeon dirs and re-chown socket on daemon restart

### LINKED ISSUE #173: init+kukeond: SGID kukeon dirs and re-chown socket on daemon restart
## Context
PR #167 (closes #163) makes `kuke init` create the `kukeon` user/group and chown `/run/kukeon`, `/opt/kukeon` (recursive), and `/run/kukeon/kukeond.sock` to `root:kukeon` so a non-root operator in the `kukeon` group can dial the daemon. Two carve-outs were called out in the PR description and remain after merge:

1. **Daemon restart resets socket ownership.** `internal/daemon/server.go:89` calls `net.Listen("unix", …)`, which recreates the socket as `root:root` mode `0o600`. After any `kukeond` cell restart the operator loses non-root access until `kuke init` is re-run. Init is idempotent, so the workaround exists, but the invariant should be self-maintaining.
2. **No SGID on `/run/kukeon` or `/opt/kukeon`.** Init's recursive chown captures everything that exists at bootstrap, but files the daemon writes later (`metadata.json`, CNI state, etc.) land as `root:root`. Reads through the daemon still work, but anything that touches files directly (e.g. `--no-daemon` paths, debugging) breaks for `kukeon`-group users.

## Proposal
Plumb the `kukeon` GID into `kukeond` and set SGID on the two managed roots:

1. Pass the `kukeon` GID to `kukeond` (lookup at daemon start, or new `--socket-gid` flag set by init). After `net.Listen` succeeds in `internal/daemon/server.go`, `os.Chown` the socket to `root:<kukeon-gid>` and `os.Chmod` it to `0o660`. Idempotent and safe if the group is missing (skip with a log line, don't fail).
2. In `kuke init` bootstrap, set SGID on directories under `/run/kukeon` and `/opt/kukeon` (mode `2750` instead of `0750`) so new files inherit the `kukeon` group. Apply the SGID bit during the same recursive walk that already runs.
3. Update the smoke test in `CLAUDE.md` to verify both: restart the `kukeond` cell and confirm the socket is still `root:kukeon 0o660`; touch a file under `/opt/kukeon` via the daemon and confirm group is `kukeon`.

## Acceptance criteria
- [ ] After `kukeond` restart (e.g. `kuke kill cell kukeond …` + auto-restart, or container restart), `/run/kukeon/kukeond.sock` is `root:kukeon` mode `0o660` without re-running `kuke init`.
- [ ] Directories under `/run/kukeon/` and `/opt/kukeon/` have the SGID bit set (mode `2750`); new files created by the daemon land in group `kukeon`.
- [ ] Non-root operator (member of `kukeon` group) can `kuke get realms` after a daemon restart without sudo and without re-running init.
- [ ] CLAUDE.md daemon-parity smoke test still passes.
- [ ] CLAUDE.md smoke section updated with the daemon-restart + new-file group checks.

## Notes
- Carved out of #167 / #163 to keep that PR scoped to the original AC. Author flagged both items in the PR description under "Notable judgment calls".
- Code anchors: `internal/daemon/server.go:89` (socket listen), `cmd/kuke/init/init.go` permissions block (recursive chown step from #167), `internal/controller/bootstrap.go:516` (any daemon-side `MkdirAll` under managed paths).

eminwux:
## Summary
- Plumbs the kukeon GID through to `kukeond serve` via a new `--socket-gid` flag (set by `kuke init` from the resolved GID). After `net.Listen`, the daemon `chown`s the socket to `root:<kukeon-gid>` and uses mode `0o660`, so a kukeond restart no longer resets it to `root:root 0o600`. Failure to chown is a warn-and-continue (group missing on dev hosts is not fatal).
- Sets the SGID bit on `/run/kukeon` and `/opt/kukeon` (mode `2750`) during the existing recursive walk in init, so files the daemon writes later (`metadata.json`, CNI state) inherit the `kukeon` group instead of falling back to `root:root` and breaking `--no-daemon` reads for non-root operators.
- Closes the two carve-outs called out in #167's PR description for #163.

## Notable judgment calls
- `--socket-gid` is an explicit numeric flag rather than having kukeond resolve `kukeon` via `user.LookupGroup` inside its container. The daemon image does not necessarily ship `/etc/group` with a `kukeon` entry, and a numeric GID matches what already lives on the host's bind-mounted state — keeping the contract numeric avoids divergence between init and serve.
- Mode constants: `kukeonRunDirMode` and `kukeonRunPathDirMode` use `os.ModeSetgid | 0o750` rather than the bare `0o2750`. Go's `os.Chmod` reads SGID from `os.ModeSetgid`, not from raw mode bits — `0o2750` would silently be applied as `0o0750`. A small `formatPosixMode` helper renders init's permission report as the on-disk POSIX octal (e.g. `0o2750`), since `%#o` on a Go `FileMode` would print internal flag bits like `0x400000`.
- CLAUDE.md update is filed as separate docs follow-up #175 per `agents/dev/CLAUDE.md`'s rule that project CLAUDE.md edits cannot be bundled into a code-fix PR.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` passes
- [x] New unit coverage:
  - `internal/daemon/server_test.go` — `Serve` chowns socket to `SocketGID` and applies mode `0o660`; without `SocketGID` the socket stays `0o600`.
  - `internal/controller/bootstrap_test.go` — `kukeondCellDoc` emits `--socket-gid <gid>` only when GID > 0.
  - `internal/sysuser/sysuser_test.go` — `ChownTreeAndChmod` propagates SGID on dirs and leaves regular files unaffected.
- [x] Local end-to-end smoke (per project `CLAUDE.md`):
  - Tear down kukeond cell, rebuild `kuke`/`kukeond`, rebuild and import `kukeon-local:dev`, run `sudo ./kuke init --kukeond-image docker.io/library/kukeon-local:dev`. Init reports `mode 0o2750 (dirs) / 0o0640 (files)` for `/opt/kukeon` and `mode 0o0660` for the socket.
  - `stat -c '%A %U:%G %a' /run/kukeon /run/kukeon/kukeond.sock /opt/kukeon /opt/kukeon/default` confirms `drwxr-s---` (SGID) on dirs, `srw-rw----` `root:kukeon 660` on the socket.
  - Daemon-parity check: `sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` return identical output.
  - Daemon-restart check: `kuke stop cell kukeond ... && kuke start cell kukeond ... --no-daemon`. After restart, `stat` still reports `root:kukeon 660`; `ps -ef | grep kukeond` shows the daemon launched with `--socket-gid 986`.
  - Non-root group member dial: as `inwx` (in the `kukeon` group), `./kuke get realms` succeeds without sudo.
  - SGID inheritance: `sudo ./kuke create realm sgid-smoke` followed by `stat /opt/kukeon/sgid-smoke/metadata.json` shows `root:kukeon`.

Closes #173

Filed as a separate docs follow-up: #175 (CLAUDE.md smoke section update for the new daemon-restart and group-inheritance checks).

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -105,6 +105,8 @@ var (
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKEOND_SOCKET = DefineKV("KUKEOND_SOCKET", "kukeond/socket", "/run/kukeon/kukeond.sock")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKEOND_SOCKET_GID = DefineKV("KUKEOND_SOCKET_GID", "kukeond/socketGID", "0")
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_INIT_REALM = DefineKV("KUKE_INIT_REALM", "kuke/init/realm")
--- a/cmd/kuke/init/init.go
+++ b/cmd/kuke/init/init.go
@@ -48,9 +48,12 @@ const (
 	// Files get 0o640 instead of 0o750: blanket-chmoding the tree to 0o750
 	// would mark JSON metadata files as executable, which has no purpose.
 	// Dirs need execute (traverse) for the kukeon group, so they stay
-	// 0o750.
-	kukeonRunDirMode      os.FileMode = 0o750
-	kukeonRunPathDirMode  os.FileMode = 0o750
+	// 0o750. The SGID bit is set on directories (mode 2750) so that files
+	// the daemon writes later (metadata.json, CNI state, etc.) inherit
+	// the kukeon group instead of landing as root:root and breaking
+	// `--no-daemon` reads for non-root operators.
+	kukeonRunDirMode      os.FileMode = os.ModeSetgid | 0o750
+	kukeonRunPathDirMode  os.FileMode = os.ModeSetgid | 0o750
 	kukeonRunPathFileMode os.FileMode = 0o640
 	kukeonSocketMode      os.FileMode = 0o660
 )
@@ -172,18 +175,9 @@ func runInit(cmd *cobra.Command, _ []string) error {
 		runPath = config.DefaultRunPath()
 	}
 
-	opts := controller.Options{
-		RunPath:            runPath,
-		ContainerdSocket:   viper.GetString(config.KUKEON_ROOT_CONTAINERD_SOCKET.ViperKey),
-		KukeondImage:       image,
-		KukeondSocket:      socketPath,
-		ForceRegenerateCNI: viper.GetBool(config.KUKE_INIT_FORCE_REGENERATE_CNI.ViperKey),
-	}
-
-	logger.DebugContext(cmd.Context(), "running init", "opts", opts)
-
 	// Ensure the kukeon system user/group exist before bootstrap so the
-	// post-bootstrap chown step has a GID to apply.
+	// post-bootstrap chown step has a GID to apply, and so the kukeond cell
+	// is provisioned with --socket-gid pointing at it.
 	ensure, ensureErr := sysuser.EnsureUserGroup(
 		cmd.Context(),
 		consts.KukeonSystemUser,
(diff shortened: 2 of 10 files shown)


eminwux:
**PR #176 Review — `feat(init+daemon): SGID kukeon dirs and re-chown socket on daemon restart`**

Closes the two #167 carve-outs cleanly. The wiring is symmetric on both ends (init plumbs `--socket-gid`, kukeond's Serve re-applies chown each boot), and the unit tests pin all three load-bearing behaviors: socket chown+mode after Serve, fall-back to `0o600` when no GID, and SGID survival through `ChownTreeAndChmod`. No correctness, security, or dead-code concerns. Test plan completed including the daemon-parity check.

Notes (non-blocking):

- **Comment drift** `cmd/kuke/init/init.go:227-230`: the comment "Daemon restart resets the ownership; re-running `sudo kuke init` (idempotent) restores it" is now stale — daemon restart re-applies it via `--socket-gid` and doesn't need re-init. Worth a one-line update next pass.
- **Defense in depth** `internal/daemon/server.go:99`: with `/run/kukeon` now SGID, the socket already inherits the kukeon group on `net.Listen`, so the explicit `os.Chown` is belt-and-suspenders rather than load-bearing in the steady state. Keeping it is fine — just a heads-up that the SGID-on-dir is the primary mechanism.

LGTM.

Issue #177, opened 2026-04-28
docs: update CLAUDE.md smoke output for default.kukeon.io namespace


eminwux:
## Context

PR for #169 standardizes containerd namespace mapping so `default` realm now resolves to `default.kukeon.io` (was `kukeon.io`). The expected `kuke get realms` output baked into `CLAUDE.md` still shows the old namespace.

Per `agents/dev/CLAUDE.md`, project `CLAUDE.md` updates travel as their own issue+PR — they cannot be bundled into the code-fix PR.

## Acceptance criteria

- [ ] In `CLAUDE.md` under "5. Daemon-parity check", replace the expected output block so the `default` row shows `default.kukeon.io`. Verbatim proposed wording (the only diff is the `default` row's NAMESPACE column):

  ```
  NAME         NAMESPACE              STATE  CGROUP
  -----------  ---------------------  -----  -------------------
  default      default.kukeon.io      Ready  /kukeon/default
  kuke-system  kuke-system.kukeon.io  Ready  /kukeon/kuke-system
  ```

  (Note the column rule width may need a one-character widen if the formatter keeps it aligned to the longest namespace; keep the existing alignment style — `kuke-system.kukeon.io` is still the widest cell, so no change is required.)

## Notes

- Pre-1.0; no migration step beyond what the existing smoke section already prescribes (rebuild + reinit).
- Filed by dev per the "AC mixes a code fix with `<project>/CLAUDE.md` changes" rule in `agents/dev/CLAUDE.md`.

eminwux:
Superseded by #244 — its proposed verbatim wording currently uses `kukeon.io` for the default realm, but #169/#178 standardized this to `default.kukeon.io`. Update #244's wording to use `default.kukeon.io` when picking it up.

Issue #168, opened 2026-04-28
realm+cni: validate realm names and use k-{hash} for all bridge names


eminwux:
## Context

Realm names today flow into several downstream resources with no validation:

- **containerd container IDs** use `_` as separator: `{space}_{stack}_{cell}_{container}` (`internal/util/naming/naming.go:58,83`). An underscore in a realm/space/stack/cell name would corrupt the parser path.
- **cgroup paths** use `/` as separator: `/kukeon/{realm}/{space}/...` (`internal/ctr/cgroups_utils.go:29-71`). A slash in any name would inject extra path components.
- **Bridge / link names** are built `{realm}-{space}` (`internal/util/naming/naming.go:27-38`) under IFNAMSIZ=15. Today `internal/cni/config.go:51-81` truncates to a 6-char prefix + 8-hex hash *only when overflow*. That makes iface names sometimes readable (`default-foo`) and sometimes opaque (`kuke-s-1a2b3c4d`) — inconsistent, and still depends on the user-supplied name being short enough.

`internal/controller/create_realm.go:53-56` only trims and rejects empty names — no charset, no length.

## Proposal

Two related hardening changes:

**1. Reject `_` and `/` in realm names** with a clear error. A broader DNS-label-style allowlist can layer on later; for now we close the two known footguns (container-ID parser collision, cgroup-path injection).

**2. Use `k-{8-hex-hash}` for every bridge / link name**, not only when the natural name overflows IFNAMSIZ. Persist the chosen interface name on the cell's status so `kuke describe cell` (and future debug tooling) can recover the human→iface mapping without recomputing the hash.

This decouples interface-name length from user-controlled realm/space lengths entirely, and makes bridge names a stable opaque identifier with a single derivation rule.

## Acceptance criteria

- [ ] `kuke create realm` with `_` in the name returns a clear error and does not create the realm.
- [ ] `kuke create realm` with `/` in the name returns a clear error and does not create the realm.
- [ ] Validation lives in a single `ValidateRealmName(string) error` (or equivalent) shared by every realm-creation path.
- [ ] Bridge / link names always emit `k-{8-hex-hash}` (10 chars, well under IFNAMSIZ=15). The natural-name-when-short branch in `SafeBridgeName` is removed.
- [ ] Cell status records the bridge name(s) the cell is attached to (e.g., `Status.Network.BridgeName`) and survives daemon restarts.
[...]

PR #178, opened 2026-04-28
feat(realm): standardize containerd namespace to <realm>.kukeon.io

### LINKED ISSUE #169: realm: standardize containerd namespace to <realm>.kukeon.io for all realms
## Context

The realm→containerd-namespace mapping is inconsistent. There are actually three different schemes today depending on how the realm was created:

| Realm source | Namespace today | Where decided |
|---|---|---|
| `default` (bootstrap) | `kukeon.io` | `internal/consts/consts.go:47-48` |
| `kuke-system` (bootstrap) | `kuke-system.kukeon.io` | `internal/consts/consts.go:53-54` |
| User-created `foo` | `foo` (no suffix) | `internal/controller/create_realm.go:57-62` |

This is surprising — `kuke create realm foo` does *not* produce `foo.kukeon.io` — and it means there is no single function mapping a realm to its namespace, so every consumer has to special-case the bootstrap realms.

## Proposal

Introduce a single helper `RealmNamespace(realm string) string` returning `<realm>.kukeon.io` for *all* realms. Route bootstrap realms and `kuke create realm` through it. Drop the hardcoded per-realm namespace constants (or keep them as derived literals if call sites still need them).

Pre-1.0 — `kuke init` rebuild + reinit per the local smoke test in CLAUDE.md is acceptable migration; no backwards-compat shim.

## Acceptance criteria

- [ ] A single `RealmNamespace(realm) string` (or equivalent) is the only place `.kukeon.io` is appended to a realm name.
- [ ] `default` realm now resolves to containerd namespace `default.kukeon.io` (was `kukeon.io`).
- [ ] `kuke create realm foo` results in containerd namespace `foo.kukeon.io` (was `foo`).
- [ ] `internal/consts/consts.go` no longer hardcodes per-realm namespace strings — bootstrap constants either derive from `RealmNamespace` or are removed.
- [ ] CLAUDE.md smoke-test expected output is updated: `kuke get realms` shows `default.kukeon.io` for the default realm.
- [ ] `kuke get realms` and `kuke get realms --no-daemon` parity check passes after reinit.

## Notes

- Depends on realm-name validation in #168 — namespace derivation is only safe once the legal charset (no `_`, no `/`) is enforced. Land #168 first.
- Pre-1.0; reinit-as-migration is fine.

eminwux:
## Summary

- Adds `consts.RealmNamespace(realm string) string` — the single helper that appends `.kukeon.io` to a realm name. Every realm-name → containerd-namespace derivation in the codebase now routes through it (bootstrap, user-realm create/purge/get, runner-layer fallbacks).
- Drops the per-realm namespace constants (`KukeonDefaultRealmNamespace`, `KukeSystemRealmNamespace`) — bootstrap now derives them from `consts.RealmNamespace(KukeonDefaultRealmName)` / `consts.RealmNamespace(KukeSystemRealmName)`.
- `default` realm now resolves to `default.kukeon.io` (was `kukeon.io`).
- `kuke create realm foo` now produces containerd namespace `foo.kukeon.io` (was bare `foo`).

## Notable judgment calls

- The two existing constants are removed rather than kept as derived literals: Go does not allow `const` to call a function, so leaving them as constants would have meant either making them `var`s (mutable, surprising) or repeating the suffix string. Removing them and routing call sites through the helper is the cleanest enforcement of the AC's "only place .kukeon.io is appended" requirement.
- Test fixtures: `buildTestRealm(name, "")` now defaults the namespace to `consts.RealmNamespace(name)`. Tests that explicitly pass a namespace are unchanged. This keeps test fixtures aligned with the controller-layer default so `*_DefaultNamespace` tests still cover the production path.
- The on-disk path layout under `/opt/kukeon/<realm-name>/...` continues to use the bare realm name (not the namespace). The namespace is the containerd-side identifier; the disk path is the kukeon-side identifier. They diverge intentionally.
- Per the dev role's "AC mixes a code fix with `<project>/CLAUDE.md` changes" rule, the smoke-output bullet in CLAUDE.md is filed as a separate `docs:` follow-up issue rather than bundled here. See #177.
- Per the issue's Notes, #168 (realm-name validation) lands later. This PR does not introduce a new risk: today user-realm names already flow into containerd verbatim, so `foo_bar` → `foo_bar.kukeon.io` is no worse than the existing `foo_bar` → `foo_bar` mapping, and the AC is independent of validation.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` passes (full unit suite incl. `internal/controller/...`, `cmd/kuke/...`, runner-layer)
- [x] Local end-to-end smoke per project `CLAUDE.md`:
  - Tear down kukeond cell + `kuke purge realm default kuke-system --cascade --force --no-daemon` to clear stale `kukeon.io` namespace.
  - Rebuild `kuke`/`kukeond`, rebuild and import `kukeon-local:dev`, run `sudo ./kuke init --kukeond-image docker.io/library/kukeon-local:dev`. Init prints `containerd namespace "default.kukeon.io": created` (was `"kukeon.io"`).
  - Daemon-parity check: `sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` return identical output, with `default` mapped to `default.kukeon.io` and `kuke-system` to `kuke-system.kukeon.io`.
- [x] User-realm path: `sudo ./kuke create realm test-userrealm` reports `(namespace "test-userrealm.kukeon.io")`; `kuke get realms` confirms; `kuke purge realm test-userrealm --cascade --force` cleans up.

Closes #169

Filed as a separate docs follow-up: #177 (CLAUDE.md smoke section update for the new `default.kukeon.io` namespace).

### DIFF
--- a/cmd/kuke/init/init_test.go
+++ b/cmd/kuke/init/init_test.go
@@ -255,7 +255,7 @@ func TestPrintRealmActions(t *testing.T) {
 			name: "created",
 			section: controller.RealmSection{
 				RealmName:                          "default",
-				RealmContainerdNamespace:           "kukeon.io",
+				RealmContainerdNamespace:           "default.kukeon.io",
 				RealmCreated:                       true,
 				RealmContainerdNamespaceCreated:    true,
 				RealmCgroupCreated:                 true,
@@ -264,7 +264,7 @@ func TestPrintRealmActions(t *testing.T) {
 			},
 			expected: []string{
 				"- realm \"default\": created",
-				"- containerd namespace \"kukeon.io\": created",
+				"- containerd namespace \"default.kukeon.io\": created",
 				"- realm cgroup: created",
 			},
 		},
@@ -272,7 +272,7 @@ func TestPrintRealmActions(t *testing.T) {
 			name: "already-existed",
 			section: controller.RealmSection{
 				RealmName:                          "default",
-				RealmContainerdNamespace:           "kukeon.io",
+				RealmContainerdNamespace:           "default.kukeon.io",
 				RealmCgroupExistsPre:               true,
 				RealmCgroupExistsPost:              true,
 				RealmContainerdNamespaceExistsPre:  true,
@@ -282,7 +282,7 @@ func TestPrintRealmActions(t *testing.T) {
 			},
 			expected: []string{
 				"- realm \"default\": already existed",
-				"- containerd namespace \"kukeon.io\": already existed",
+				"- containerd namespace \"default.kukeon.io\": already existed",
 				"- realm cgroup: already existed",
 			},
 		},
--- a/e2e/e2e_kuke_test.go
+++ b/e2e/e2e_kuke_test.go
@@ -288,7 +288,7 @@ func TestKuke_Init_VerifyState(t *testing.T) {
 	}
 
 	// Step 4: Verify containerd namespace exists
-	namespace := "kukeon.io"
+	namespace := consts.RealmNamespace(consts.KukeonDefaultRealmName)
 	if !verifyContainerdNamespace(t, namespace) {
 		t.Fatalf("containerd namespace %q not found after init", namespace)
 	}
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -204,13 +204,13 @@ func loadKukeondImageIntoContainerd(t *testing.T) string {
 	defer importCancel()
 	importCmd := exec.CommandContext(
 		importCtx, ctrPath,
-		"--namespace", consts.KukeSystemRealmNamespace,
+		"--namespace", consts.RealmNamespace(consts.KukeSystemRealmName),
 		"images", "import", tarPath,
 	)
 	if out, importErr := importCmd.CombinedOutput(); importErr != nil {
 		t.Fatalf(
 			"ctr -n %s images import %q failed: %v\noutput:\n%s",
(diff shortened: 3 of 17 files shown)


eminwux:
**PR #178 Review — `feat(realm): standardize containerd namespace to <realm>.kukeon.io`**

Clean, well-scoped refactor: a single `consts.RealmNamespace` helper now owns the `.kukeon.io` suffix, every call site (bootstrap, create/get/purge realm, runner-layer fallbacks, uninstall, e2e fixtures) routes through it, and all unit/e2e tests are updated to assert the new `<realm>.kukeon.io` form. The `default` realm's namespace change (`kukeon.io` → `default.kukeon.io`) is the only user-visible behavior shift and is exercised end-to-end in the smoke plan. Daemon-parity check is in the test plan; the daemon's own namespace (`kuke-system.kukeon.io`) is unchanged so `kuke purge default` still can't take down the daemon. No correctness, sensitive-data, or dead-code concerns: the deleted exported constants (`KukeonDefaultRealmNamespace`, `KukeSystemRealmNamespace`) had only in-repo callers, all of which are touched by this PR (verified against `gh pr files` — the cross-check returned the exact PR file list), and CI is green so the build is consistent. Splitting the docs follow-up as #177 per the project's "no code+CLAUDE.md mix" rule is the right call.

LGTM — ready to merge.

PR #179, opened 2026-04-28
feat(realm+cni): validate realm names and use k-{hash} for all bridge names

### LINKED ISSUE #168: realm+cni: validate realm names and use k-{hash} for all bridge names

eminwux:
## Summary
- Add `ValidateRealmName(string) error` (`internal/util/naming/realmname.go`) that rejects `_` (collides with the containerd ID separator) and `/` (injects extra cgroup path components). Wire it into both controller-level `Exec.CreateRealm` and runner-level `Runner.CreateRealm` so every realm-creation path (CLI, daemon RPC, apply reconcile, init bootstrap) trips the same gate.
- Replace the conditional truncation in `cni.SafeBridgeName` with an unconditional `k-{8-hex FNV-1a}` (10 chars). Bridge names no longer depend on user-controlled realm/space lengths or charset; the empty-string case is preserved so callers can detect "unset". Defensive >15-char check at `internal/cni/network.go:191` is preserved.
- Persist `Status.Network.BridgeName` on every cell via a new `CellNetworkStatus` type at both internal (`internal/modelhub/cell.go`) and external (`pkg/api/model/v1beta1/cell.go`) layers; populated during `provisionNewCell` and backfilled on `EnsureCell` so old metadata files self-heal on next ensure cycle. `kuke get cell -o yaml` surfaces the `network.bridgeName` line directly from the persisted metadata, no recomputation.

Closes #168

## Notable judgment calls
- `Network.BridgeName` is populated best-effort: if `BuildSpaceNetworkName` fails (missing realm/space on the cell spec), provisioning still succeeds — the iface is created from the conflist regardless of what we record on the status.
- `Network.BridgeName` is a single string today (the cell joins one space, ergo one bridge). If a future change attaches a cell to multiple bridges, this becomes a slice — keeping it as a struct (`CellNetworkStatus`) leaves room.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full unit + integration excluding e2e)
- [x] New unit tests:
  - `internal/util/naming/realmname_test.go` — `_`/`/` rejection, empty/whitespace handling
  - `internal/controller/create_realm_test.go::TestCreateRealm_RejectsInvalidCharacters` — controller-layer rejection without runner mutation
  - `internal/cni/config_test.go::TestSafeBridgeName_Determinism` — same input → same hash, different inputs differ
  - `internal/apischeme/scheme_test.go::TestCellRoundTripV1Beta1_NetworkBridgeName` — Status.Network.BridgeName round-trips through ext→internal→ext and renders in YAML
- [x] Project smoke test (CLAUDE.md): tear down kukeond cell, rebuild, re-import image, `kuke init`, daemon-parity check (`kuke get realms` ≡ `kuke get realms --no-daemon`)
- [x] End-to-end realm rejection via daemon: `sudo kuke create realm bad_name` → `Error: realm name is invalid: "bad_name" contains disallowed character...`; same for `bad/name`
- [x] End-to-end bridge persistence: re-init produced `bridge: k-082023b5` in `/opt/kukeon/kuke-system/kukeon/network.conflist` and `status.network.bridgeName: k-082023b5` in the kukeond cell metadata; `kuke get cell kukeond ... -o yaml` surfaces the same value

### DIFF
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -813,6 +813,9 @@ func ConvertCellDocToInternal(in ext.CellDoc) (intmodel.Cell, error) {
 			Status: intmodel.CellStatus{
 				State:      intmodel.CellState(in.Status.State),
 				CgroupPath: in.Status.CgroupPath,
+				Network: intmodel.CellNetworkStatus{
+					BridgeName: in.Status.Network.BridgeName,
+				},
 				Containers: convertContainerStatusesToInternal(in.Status.Containers),
 			},
 		}
@@ -868,6 +871,9 @@ func BuildCellExternalFromInternal(in intmodel.Cell, apiVersion ext.Version) (ex
 			Status: ext.CellStatus{
 				State:      ext.CellState(in.Status.State),
 				CgroupPath: in.Status.CgroupPath,
+				Network: ext.CellNetworkStatus{
+					BridgeName: in.Status.Network.BridgeName,
+				},
 				Containers: buildContainerStatusesExternalFromInternal(in.Status.Containers),
 			},
 		}
--- a/internal/apischeme/scheme_test.go
+++ b/internal/apischeme/scheme_test.go
@@ -380,6 +380,63 @@ func TestCellRoundTripV1Beta1(t *testing.T) {
 	}
 }
 
+// TestCellRoundTripV1Beta1_NetworkBridgeName covers AC for #168: the cell
+// status persists Network.BridgeName through the external→internal→external
+// round-trip so daemon restarts can recover the iface mapping from the
+// cell metadata file alone.
+func TestCellRoundTripV1Beta1_NetworkBridgeName(t *testing.T) {
+	const wantBridge = "k-1a2b3c4d"
+
+	input := ext.CellDoc{
+		APIVersion: ext.APIVersionV1Beta1,
+		Kind:       ext.KindCell,
+		Metadata: ext.CellMetadata{
+			Name:   "cell-net",
+			Labels: map[string]string{},
+		},
+		Spec: ext.CellSpec{
+			ID:         "cell-id-net",
+			RealmID:    "realm0",
+			SpaceID:    "space0",
+			StackID:    "stack0",
+			Containers: []ext.ContainerSpec{},
+		},
+		Status: ext.CellStatus{
+			State:      ext.CellStateReady,
+			CgroupPath: "/sys/fs/cgroup/cell-net",
+			Network: ext.CellNetworkStatus{
+				BridgeName: wantBridge,
+			},
+		},
+	}
+
+	internal, version, err := apischeme.NormalizeCell(input)
+	if err != nil {
+		t.Fatalf("NormalizeCell: %v", err)
+	}
+	if internal.Status.Network.BridgeName != wantBridge {
+		t.Errorf("internal bridge = %q, want %q", internal.Status.Network.BridgeName, wantBridge)
+	}
+
+	output, err := apischeme.BuildCellExternalFromInternal(internal, version)
+	if err != nil {
+		t.Fatalf("BuildCellExternalFromInternal: %v", err)
+	}
+	if output.Status.Network.BridgeName != wantBridge {
+		t.Errorf("external bridge = %q, want %q", output.Status.Network.BridgeName, wantBridge)
+	}
+
+	// The YAML rendering must include the bridgeName line so `kuke get
+	// cell -o yaml` surfaces it for operators.
+	rendered, err := yaml.Marshal(output)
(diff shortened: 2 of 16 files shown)


eminwux:
**PR #179 Review — `feat(realm+cni): validate realm names and use k-{hash} for all bridge names`**

Clean, well-scoped change. Validation is wired at both the controller and runner boundaries (defense in depth), the bridge-name hash is now unconditional and length-bounded (`k-{8hex}` = 10 chars, well under IFNAMSIZ-1), and `Status.Network.BridgeName` is persisted on new cells and backfilled in `EnsureCell` so older metadata self-heals on next ensure cycle. Smoke test and daemon-parity check are both in the test plan, dead-symbol scan came back clean.

- **Trim drift** `internal/controller/runner/create_realm.go:30`: validation runs against the trimmed name (`ValidateRealmName` trims internally), but the runner then stores `realm.Metadata.Name` untrimmed. The controller trims before calling so the live path is safe; mirror the `realm.Metadata.Name = strings.TrimSpace(...)` step here as defense in depth.
- **Struct `omitempty`** `pkg/api/model/v1beta1/cell.go:54`: `encoding/json` does not elide zero-valued structs, so cells where bridge derivation failed will still render `"network":{}` in JSON output (the inner `BridgeName` does have `omitempty`). Cosmetic only — ignore unless you switch the field to `*CellNetworkStatus`.

Out-of-scope note: realm names are validated, but space/stack/cell names flow into the `_`-separated containerd ID via `naming.BuildContainerdID` and remain unvalidated against the same separator collision the PR description calls out.

LGTM — feel free to merge.

PR #182, opened 2026-04-29
feat(naming): validate space/stack/cell/container names

### LINKED ISSUE #180: naming: extend separator/path validation to space, stack, cell, and container names
## Context

PR #179 (closes #168) added `ValidateRealmName` and wired it into both the controller and runner realm-creation paths, blocking `_` (collides with the containerd ID separator) and `/` (injects cgroup path components).

The same hazard exists one level down. `naming.BuildContainerdID(space, stack, cell, container)` joins all four with `_`:

```go
return fmt.Sprintf("%s_%s_%s_%s", spaceName, stackName, cellName, containerName), nil
```

Today nothing validates space/stack/cell/container names against `_` or `/` — only emptiness is checked (`internal/util/naming/naming.go:64`). A user-supplied space named `foo_bar` produces an ambiguous ID that the inverse parser cannot round-trip back to its hierarchy components, exactly the failure mode the PR description for #168 called out for realms.

Surfaced as an out-of-scope note in the [PR #179 self-review](https://github.com/eminwux/kukeon/pull/179#issuecomment-4338962376).

## Proposal

Add a shared validator (e.g. `ValidateHierarchyName(kind, name string) error` in `internal/util/naming/`) that applies the same rules `ValidateRealmName` does — reject `_` and `/`, reject empty/whitespace — and wire it into every space, stack, cell, and container creation path:

- Controller `Exec.Create*` entry points (CLI, daemon RPC, apply reconcile, init bootstrap).
- Runner `Runner.Create*` boundaries (defense in depth, mirroring the realm pattern).

Trim before validation; mirror the trimmed value back onto `Metadata.Name` in the runner so we never persist whitespace-padded names. Reuse `errdefs.ErrInvalidRealmName`-style sentinels — likely one per kind, or one generic `ErrInvalidName` carrying the kind in the message.

## Acceptance criteria

- [ ] Single shared validator rejects `_` and `/` (and empty/whitespace) for space, stack, cell, and container names.
- [ ] Validation runs at both controller and runner boundaries for every create path (parity with the realm wiring landed in #168 / #179).
- [ ] Unit tests cover rejection of `_`, `/`, empty, and whitespace-only inputs for each kind.
- [ ] An end-to-end rejection test confirms `kuke create space bad_name ...` (and equivalents for stack/cell) fails through the daemon with a clear error.
- [ ] Trim-drift fix mirrored: runner stores the trimmed name on `Metadata.Name`.
- [ ] Daemon-parity smoke test (`kuke get …` ≡ `kuke get … --no-daemon`) passes per CLAUDE.md.

## Notes

- Sibling of #168; uses the same approach. Keep `ValidateRealmName` as-is or refactor it to delegate to the shared validator — either is fine, reviewer's call.
- If the disallowed character set should differ per kind (e.g. container names may need to allow something space names don't), make that explicit in the validator signature rather than splitting into per-kind validators.
- Out of scope: tightening to a stricter charset (e.g. DNS-1123). This issue is only about closing the separator/path-injection hole that the realm work already closed.

eminwux:
## Summary
- Extends the realm-name validator pattern (#168/#179) to space, stack, cell, and container names. Rejects `_` (collides with the containerd ID separator `{space}_{stack}_{cell}_{container}`), `/` (would inject cgroup-path components), and empty/whitespace-only names. Uses a generic `errdefs.ErrInvalidName` sentinel; the kind name (`space`/`stack`/`cell`/`container`) is baked into the wrapped error message so the operator knows which input was rejected.
- Wires the validator at both the controller (`Exec.Create*`) and the runner (`r.Create*`) boundaries — defense-in-depth, matching the realm precedent. `runner.CreateCell` also re-validates each embedded `cell.Spec.Containers[].ID` so a malformed nested ID can't slip through. Each entry point trims the input first and mirrors the trimmed value back onto `Metadata.Name` (the documented "trim drift" fix from issue #180), so persisted state matches what was validated.
- Updates the controller-side `TestCreateCell_ContainerOutcomes/empty_container_ID_is_skipped_in_outcomes` case in place: empty container IDs are now rejected at the cell-create boundary instead of silently filtered downstream. The renamed subtest documents the new contract.

## Test plan
- [x] `make test` — full unit suite, all packages green (controller, runner, naming, errdefs paths).
- [x] `go test ./internal/util/naming/...` — table-driven coverage of all 4 kinds, valid + invalid + empty + whitespace cases.
- [x] `go test ./internal/controller/...` — every kind has a `Create*_RejectsInvalidCharacters` unit test driving the controller via a fakeRunner that `t.Fatal`s if the create path is reached, asserting `errors.Is(err, errdefs.ErrInvalidName)`. `create_cell_test.go` also has `TestCreateCell_RejectsInvalidContainerCharacters` for embedded container IDs.
- [x] `go test -tags e2e -run "TestKuke_(CreateSpace|CreateStack|CreateCell|CreateContainer)_RejectsInvalidNames" ./e2e/...` — all 8 e2e cases pass (4 kinds × `_` and `/`). Uses `--no-daemon --run-path X` (in-process controller); defense-in-depth means the validator fires either way.
- [x] `go build ./...` and `go vet ./...` clean.
- [x] Daemon-parity smoke check from project CLAUDE.md: `sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` produced byte-identical output (8 realms — `default`, `kuke-system`, plus 6 leftover `e-r-*` realms from prior e2e runs). This change doesn't touch the daemon path; the parity check is included because CLAUDE.md asks for it on any change touching what the daemon reads.
- [ ] `make test-e2e` (full suite) — not green locally, but the failures are pre-existing environmental issues unrelated to this change. Verified by stashing this branch's changes, checking out `origin/main`, and reproducing the same failures: `TestKuke_Init_VerifyState` fails with stale containerd containers (`kukeon_kukeon_kukeond_*` left "running" by prior runs), which then takes down the daemon socket and cascades into `dial unix /run/kukeon/kukeond.sock: no such file or directory` failures across the parallel daemon-mode tests. None of the failing tests touch the validator paths exercised by this PR.

Closes #180

### DIFF
--- a/e2e/e2e_kuke_invalid_names_test.go
+++ b/e2e/e2e_kuke_invalid_names_test.go
@@ -1,0 +1,163 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e_test
+
+import (
+	"strings"
+	"testing"
+)
+
+// TestKuke_CreateSpace_RejectsInvalidNames covers the AC for #180: invalid
+// space names ("_" or "/") must be rejected end-to-end with a non-zero exit
+// code and a clear error message that names the offending input. Run via the
+// in-process controller (--no-daemon) so the test is self-contained — the
+// validator is wired at both controller and runner boundaries, so the same
+// rejection fires either way.
+func TestKuke_CreateSpace_RejectsInvalidNames(t *testing.T) {
+	t.Parallel()
+
+	runPath := getRandomRunPath(t)
+	mkdirRunPath(t, runPath)
+
+	tests := []struct {
+		name      string
+		spaceName string
+	}{
+		{name: "underscore in space name", spaceName: "bad_space"},
+		{name: "slash in space name", spaceName: "bad/space"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			args := appendNoDaemonRunPath(runPath,
+				"create", "space", tt.spaceName, "--realm", "default")
+			exitCode, _, stderr := runBinary(t, nil, kuke, args...)
+			if exitCode == 0 {
+				t.Fatalf("expected non-zero exit rejecting space name %q", tt.spaceName)
+			}
+			combined := string(stderr)
+			if !strings.Contains(combined, "space") || !strings.Contains(combined, tt.spaceName) {
+				t.Errorf("expected error mentioning kind=space and name %q, got:\n%s", tt.spaceName, combined)
+			}
+		})
+	}
+}
+
+// TestKuke_CreateStack_RejectsInvalidNames covers the AC for #180: invalid
+// stack names must be rejected end-to-end.
+func TestKuke_CreateStack_RejectsInvalidNames(t *testing.T) {
+	t.Parallel()
+
+	runPath := getRandomRunPath(t)
+	mkdirRunPath(t, runPath)
+
+	tests := []struct {
+		name      string
+		stackName string
+	}{
+		{name: "underscore in stack name", stackName: "bad_stack"},
+		{name: "slash in stack name", stackName: "bad/stack"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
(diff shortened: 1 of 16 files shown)


eminwux:
**PR #182 Review — `feat(naming): validate space/stack/cell/container names`**

Clean extension of the realm-name validator pattern from #168/#179 to the rest of the hierarchy. Validation is wired symmetrically at controller and runner boundaries with the trim-then-mirror fix for #180; tests cover all 4 kinds at both layers plus end-to-end. No correctness, leak, dead-code, or convention concerns. The renamed `TestCreateCell_ContainerOutcomes` subtest correctly documents the contract change (empty container IDs in a cell spec now error instead of being silently filtered) — worth calling out only because it's a behavior change, not a bug.

Two small non-blocking observations:

- **Sentinel drift** `internal/controller/runner/create_cell.go:35-46`: an empty container ID returns `ErrInvalidName` here but the controller path returns `ErrContainerNameRequired` for the same input at `internal/controller/create_cell.go:73`. Harmless in practice (runner is only reached after controller validation), but `errors.Is` against either sentinel won't catch both layers.
- **Wording** `internal/util/naming/hierarchyname.go:46`: the empty-name branch produces `"name is invalid: cell name is required"` — the two halves restate each other. Inlining `errors.New("hierarchy kind is required")` at line 42 is fine since it's a programmer-error path.

LGTM.

PR #183, opened 2026-04-29
feat(cli): add kuke log to tail sbsh capture file

### LINKED ISSUE #114: kuke log: tail the sbsh capture file for live and replay viewing
## Context

Every container that runs under the sbsh wrapper writes its full tty byte stream to a per-container capture file at `AttachableCapturePath` (`/run/kukeon/tty/capture` inside the container, host-visible via the bind-mount source directory). This is the raw data stream — every byte the workload wrote and every byte the user typed — not a parsed log. It is already produced today; nothing surfaces it to users.

The in-code comments at `internal/ctr/attachable.go:47-53` explicitly pre-claim this as a follow-up:

> AttachableCapturePath is the in-container path of the sbsh capture file. Surfacing it to `kuke logs` is a follow-up.

Because the capture is a complete, lossless byte stream, the same file backs two distinct UX modes: live tail (what is the workload printing right now?) and replay (what did the session look like end-to-end?).

## Proposal

Add a `kuke log` subcommand that streams the host-side capture file for a target cell's root container.

Scope for this issue (single PR):

- Resolve the target cell from positional args (mirroring `kuke attach`'s flag/arg shape: realm/space/stack/cell).
- Locate the host capture file from the cell's tty directory.
- Default behavior: follow the file (`tail -f`-like), printing raw bytes to stdout. The user's terminal interprets ANSI/control sequences naturally.
- `--no-follow` to dump the current contents and exit.
- Honor SIGINT cleanly so a `Ctrl-C` out of a follow drops back to the shell without a stack trace.

Out of scope (file as separate follow-ups if/when needed):

- A `--replay` mode that paces output by recorded timing. The capture is currently raw bytes with no timestamps; replay needs either a sidecar timing file or a format change in sbsh, which is its own design discussion.
- Multi-container cells (only the cell's root container's capture for this slice).
- Server-side streaming over the daemon socket. First slice reads the host file directly, same as `kuke attach` does for the socket.

## Acceptance criteria

- [ ] `kuke log <cell>` (with the same realm/space/stack flags as `kuke attach`) tails the host capture file and prints raw bytes to stdout.
- [ ] `kuke log --no-follow <cell>` prints current contents and exits 0.
- [ ] Missing capture file → clear error referencing the cell identity, exit non-zero.
- [ ] SIGINT during follow exits cleanly (no panic, no goroutine leak in tests).
- [ ] Smoke-tested against a `kuke create` workload that produces output, both live (output appears within ~1s of being written) and post-hoc (`--no-follow` after the workload exits prints the recorded session).

## Notes

- Pre-claimed as a follow-up in `internal/ctr/attachable.go:47-53` (`AttachableCapturePath`) and `:51-53` (`AttachableLogfilePath`). This issue covers the capture file; the structured `log` file can be a separate follow-up if/when it grows distinct semantics.
- Naming: in-code comments reference `kuke logs` (kubectl-parity, plural). User's framing was `kuke log` (singular). Pick one at PR time — not gating.
- Related: #84 (`kuke attach` library migration) lives next door in the CLI surface; the resolver code that maps cell-identity → host tty dir is shared and should be reused, not duplicated.
- Replay-with-timing is deliberately deferred — call it out in the PR description so future-us doesn't re-litigate the scope.

eminwux:
## Summary

- Adds `kuke log` (alias `logs`) — a thin client subcommand that tails the per-container sbsh capture file. Default mode is `tail -f`-like follow; `--no-follow` dumps the current contents and exits 0.
- Mirrors `kuke attach`'s flag/arg shape: `--realm/--space/--stack/--cell/--container` with the same auto-pick of the single non-root attachable when `--container` is omitted, and the same `Attachable=true` gate (non-Attachable containers have no capture file).
- Bytes never traverse the daemon RPC. The new `LogContainer` RPC validates the gate and returns `HostCapturePath`; the client opens that path directly with `os.Open` and streams via `io.Copy` (poll loop @ 200ms in follow mode — keeps the dependency surface tiny vs. fsnotify, and matches sbsh's append-only writes).
- SIGINT/SIGTERM during follow drains cleanly via `signal.NotifyContext`; cancellation returns nil so `kuke log` exits 0 (Ctrl+C is a benign session end).
- Missing capture file errors with `cell %q container %q has no capture file at %s` so the operator can correlate the failure to a specific cell/container.

### Implementation note for the reviewer

The issue body says "positional args" in one place and "mirroring kuke attach" in another. Mirroring `kuke attach` won — same `cobra.NoArgs` + `--container` flag — so the two thin clients stay symmetric and the existing picker semantics carry over.

To suppress a `dupl` lint hit between the new `LogContainer` and existing `AttachContainer` paths in `internal/client/local/local.go`, both methods now share a `resolveAttachable` helper that performs the normalize → `GetContainer` → `Attachable` gate sequence.

## Test plan

- [x] `make test` (full unit suite, including `cmd/kuke/log` and `internal/daemon` log_rpc_test)
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `golangci-lint` (run via repo pre-commit hook)
- [x] Daemon-parity smoke: `kuke get realms` vs `kuke get realms --no-daemon` after `kuke init` rebuild — identical output (`internal/daemon/rpcservice.go` and `internal/client/local/local.go` both touch a daemon-read path)
- [x] `kuke log --help` renders as expected with all five flags + `--no-follow`
- [ ] Live tail against a running attachable container — not exercised on the dev box; covered by unit tests for the dump path, follow-cancels-on-SIGINT path, missing-file path, ambiguous/no-candidate picker paths, and the real `tailFile` against a tempdir capture

Closes #114

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -370,6 +370,21 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_ATTACH_CONTAINER = DefineKV("KUKE_ATTACH_CONTAINER", "kuke/attach/container")
 
+	// Log command variables.
+
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_LOG_REALM = DefineKV("KUKE_LOG_REALM", "kuke/log/realm", "default")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_LOG_SPACE = DefineKV("KUKE_LOG_SPACE", "kuke/log/space", "default")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_LOG_STACK = DefineKV("KUKE_LOG_STACK", "kuke/log/stack", "default")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_LOG_CELL = DefineKV("KUKE_LOG_CELL", "kuke/log/cell")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_LOG_CONTAINER = DefineKV("KUKE_LOG_CONTAINER", "kuke/log/container")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_LOG_NO_FOLLOW = DefineKV("KUKE_LOG_NO_FOLLOW", "kuke/log/no-follow", "false")
+
 	// Stop command variables
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_STOP_CELL_REALM = DefineKV("KUKE_STOP_CELL_REALM", "kuke/stop/cell/realm", "default")
--- a/cmd/kuke/kuke.go
+++ b/cmd/kuke/kuke.go
@@ -33,6 +33,7 @@ import (
 	getcmd "github.com/eminwux/kukeon/cmd/kuke/get"
 	initcmd "github.com/eminwux/kukeon/cmd/kuke/init"
 	killcmd "github.com/eminwux/kukeon/cmd/kuke/kill"
+	logcmd "github.com/eminwux/kukeon/cmd/kuke/log"
 	purgecmd "github.com/eminwux/kukeon/cmd/kuke/purge"
 	refreshcmd "github.com/eminwux/kukeon/cmd/kuke/refresh"
 	runcmd "github.com/eminwux/kukeon/cmd/kuke/run"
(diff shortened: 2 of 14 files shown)


eminwux:
**PR #183 Review — `feat(cli): add kuke log to tail sbsh capture file`**

Clean addition. The Attachable-gate reuse via `resolveAttachable`, the bytes-stay-out-of-RPC design, and the test coverage (no-follow dump, follow-cancel-on-ctx, missing-file wrap, picker ambiguity/none, missing-flag table, real `tailFile` against a tempdir) all line up with the PR description. License headers, daemon-parity smoke, and CodeQL are all green. No correctness bugs, no leaks, no dead public API.

A few small nits (non-blocking):

- **Duplication** `cmd/kuke/log/log.go:287-316`: `pickAttachableContainer` is byte-for-byte identical to `cmd/kuke/attach/attach.go:174-202`. Daemon-side dupl was extracted via `resolveAttachable`; the client-side picker still lives in two places and will drift if one subcommand changes its semantics. Could move to `cmd/kuke/shared` next to `ClientFromCmd`.
- **Error format** `cmd/kuke/log/log.go:231`: path arg uses `%s` while cell/container use `%q` — wrap the path in quotes too so an empty/whitespace path is visible in operator output.

LGTM.

PR #185, opened 2026-04-29
feat(kukeond): introduce ServerConfiguration kind and --configuration flag

### LINKED ISSUE #159: kukeond: introduce ServerConfiguration kind and --configuration flag
## Context
`kukeond` has no dedicated configuration document today. Settings flow through
a single shared `--config` persistent flag at the `kuke` root
(`cmd/kuke/kuke.go:157`) bound to viper key `kukeon/configFile`. As we add
features that need server-only configuration (e.g. background reconciler
interval), conflating the daemon's settings with the client's becomes
untenable.

## Proposal
Introduce a `ServerConfiguration` kind (apiVersion `kukeon.io/v1beta1`) for
kukeond's settings. Add a persistent `--configuration` flag on the `kukeond`
cobra root that loads a ServerConfiguration document. Default lookup:
`/etc/kukeon/kukeond.yaml`. If absent, kukeond uses hardcoded defaults — no
error.

`kuke init` (which bootstraps the daemon) gains an additional
`--server-configuration <path>` flag pointing at the ServerConfiguration to
write/use during init. Same default path; absent → defaults.

Initial fields are the daemon-only settings already tracked via viper (socket
path, run-path, kukeond image, …). Exact list settled in code review. No
shared metadata with ClientConfiguration.

Remove the existing shared persistent `--config` flag at the `kuke` root —
no current users, clean break.

## Acceptance criteria
- [ ] New kind `ServerConfiguration` (apiVersion `kukeon.io/v1beta1`)
      registered in the apischeme; doc type lives under
      `pkg/api/model/v1beta1/`.
- [ ] `kukeond --configuration <path>` persistent root flag loads
      ServerConfiguration from `<path>`.
- [ ] Default ServerConfiguration path is `/etc/kukeon/kukeond.yaml`; absent
      file → hardcoded defaults, no error.
- [ ] `kuke init --server-configuration <path>` bootstraps the daemon using
      the given ServerConfiguration; same default + absent-file behavior.
- [ ] Example `kukeond.yaml` added under `docs/`.
- [ ] Existing shared persistent `--config` flag at the kuke root removed.
- [ ] CLAUDE.md daemon-parity smoke test still passes after `kuke init`.

## Notes
- Independent follow-up: #160 — ClientConfiguration kind +
  `kuke --configuration` flag.
- Unblocks #161 — background reconciler, which puts the reconcile interval
  on ServerConfiguration.

eminwux:
## Summary

- Adds a dedicated `ServerConfiguration` kind so kukeond settings stop riding on the shared `kuke --config` flag — keeps client and daemon config from drifting into each other.
- `kukeond --configuration <path>` (default `/etc/kukeon/kukeond.yaml`) loads the doc in PersistentPreRunE; absent file → defaults, no error.
- `kuke init --server-configuration <path>` reads the same doc, applies values to viper before bootstrap, and plumbs `--configuration` + a bind-mount of the parent dir into the kukeond cell so the daemon re-reads on every restart.
- Removes the `kuke` root `--config` persistent flag (no current users) and the `KUKEON_ROOT_CONFIG_FILE` env/viper var; `loadConfig` collapses to env binding + run-path/log-level defaulting.
- Example `docs/kukeond.yaml` documents every spec field.

## Non-obvious calls

- **apiVersion: `v1beta1`, not `kukeon.io/v1beta1`.** The issue body proposes the `kukeon.io/`-prefixed form, but every other kind in `pkg/api/model/v1beta1/` (Realm, Space, Stack, Cell, Container, CellProfile) uses bare `v1beta1`, and `apischeme.VersionV1Beta1` keys off that literal. Adopting the prefixed form here would either fork apischeme or quietly fail validation. Reviewer can override.
- **Precedence:** explicit `--flag` > env > ServerConfiguration > flag default. Implemented via `cmd.Flags().Changed(<name>)` for the flag layer and `os.LookupEnv` for the env layer — `viper.Set` is skipped when either is in play, so a one-off `--flag` keeps winning and an exported env var overrides the on-disk doc.
- ServerConfiguration is not a server-side resource — `kuke apply` still rejects it, mirroring CellProfile's treatment.

## Test plan

- [x] `make test`
- [x] `go vet ./...` / `go build ./...`
- [x] CLAUDE.md daemon-parity smoke test: tear-down → rebuild → `sudo ./kuke init --kukeond-image docker.io/library/kukeon-local:dev` → `sudo ./kuke get realms` vs `--no-daemon` (identical)
- [x] Confirmed `/bin/kukeond --configuration /etc/kukeon/kukeond.yaml serve …` is the running cmdline and `/etc/kukeon` is bind-mounted into the kukeond cell
- [x] Re-ran the smoke test with `docs/kukeond.yaml` copied to `/etc/kukeon/kukeond.yaml` — daemon-parity still identical
- [x] Re-ran the daemon-parity smoke test on the fix-up commit (env > YAML precedence) — both views still identical
- [x] Added `TestApplyServerConfigurationEnvOverridesConfig` locking in `--flag > env > ServerConfiguration > default` for kukeond
- [ ] `make e2e` — not run; the e2e suite is heavier than this PR's diff (env binding + cobra flag plumbing) and the daemon-parity smoke test is the regression guard CLAUDE.md names for changes that touch the build path / daemon

Closes #159

### DIFF
--- a/cmd/config/defaults.go
+++ b/cmd/config/defaults.go
@@ -63,12 +63,10 @@ func DefaultProfilesFile() string {
 	return filepath.Join(base, ".kukeon", "profiles.yaml")
 }
 
-func DefaultConfigFile() string {
-	base, err := os.UserHomeDir()
-	if err != nil {
-		// fallback to tmp if home dir cannot be determined
-		base = "tmp"
-	}
-
-	return filepath.Join(base, ".kukeon", "config.yaml")
+// DefaultServerConfigurationFile is the on-disk path the kukeond daemon and
+// `kuke init` read when the user does not pass `--configuration` or
+// `--server-configuration`. An absent file is not an error — the daemon
+// falls back to its hardcoded defaults.
+func DefaultServerConfigurationFile() string {
+	return filepath.Join("/", "etc", "kukeon", "kukeond.yaml")
 }
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -93,8 +93,6 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKEON_ROOT_RUN_PATH = DefineKV("KUKEON_RUN_PATH", "kukeon/runPath")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
-	KUKEON_ROOT_CONFIG_FILE = DefineKV("KUKEON_CONFIG_FILE", "kukeon/configFile")
-	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKEON_ROOT_LOG_LEVEL = DefineKV("KUKEON_LOG_LEVEL", "kukeon/logLevel", "info")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKEON_ROOT_CONTAINERD_SOCKET = DefineKV("KUKEON_CONTAINERD_SOCKET", "kukeon/containerd.socket")
@@ -107,6 +105,10 @@ var (
 	KUKEOND_SOCKET = DefineKV("KUKEOND_SOCKET", "kukeond/socket", "/run/kukeon/kukeond.sock")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKEOND_SOCKET_GID = DefineKV("KUKEOND_SOCKET_GID", "kukeond/socketGID", "0")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKEOND_CONFIGURATION = DefineKV(
+		"KUKEOND_CONFIGURATION", "kukeond/configuration", "/etc/kukeon/kukeond.yaml",
+	)
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_INIT_REALM = DefineKV("KUKE_INIT_REALM", "kuke/init/realm")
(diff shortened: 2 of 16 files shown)


eminwux:
**PR #185 Review — `feat(kukeond): introduce ServerConfiguration kind and --configuration flag`**

Clean refactor — the split between client and daemon config is the right call, the `apiVersion: v1beta1` choice matches the rest of the v1beta1 kinds, and the cell args/bind-mount wire-up has direct test coverage (`TestKukeondCellDocConfigurationArg`). The smoke-test re-run with `/etc/kukeon/kukeond.yaml` populated catches the bind-mount regression CLAUDE.md names. No correctness or sensitive-data concerns; `kuke apply` correctly rejects `ServerConfiguration` because `internal/apply/parser/parser.go:123` only switches on the five server-side kinds, mirroring `CellProfile`'s treatment.

One non-blocking observation:

- **Precedence-claim drift** `cmd/kukeond/kukeond.go:131` and `cmd/kuke/init/init.go` (`applyServerConfiguration`): the PR description states `--flag > env > ServerConfiguration > default`, but `viper.Set(...)` in both `applyServerConfiguration` paths overrides env-bound values — actual precedence is `--flag > ServerConfiguration > env > default`. If env-over-YAML was intended, the apply needs to consult `viper.IsSet`/env presence before `Set`. If YAML-over-env was the actual intent (defensible — system-wide YAML is more deliberate than a stray env var), update the description so future readers don't trip on it.

LGTM to merge.

eminwux:
Addressed in a72d801:
- Precedence drift in `applyServerConfiguration` (both `cmd/kukeond/kukeond.go` and `cmd/kuke/init/init.go`): `viper.Set` now skips fields whose env var is set, so actual order matches the description — `--flag > env > ServerConfiguration > default`.
- New `TestApplyServerConfigurationEnvOverridesConfig` (kukeond) locks the env-over-YAML guarantee per field.
- Re-ran the CLAUDE.md daemon-parity smoke test on the fix-up commit; both `kuke get realms` views remained identical.

eminwux:
**PR #185 Re-review — fix-up commit `a72d8011`**

The follow-up commit lands the precedence fix I flagged: each `viper.Set` in both `cmd/kukeond/kukeond.go:131` and `cmd/kuke/init/init.go:160` is now gated on `!envSet(...)`, doc comments are updated, and the PR description's `--flag > env > ServerConfiguration > default` claim now matches the code. `TestApplyServerConfigurationEnvOverridesConfig` covers the kukeond path; daemon-parity smoke re-run per the checklist. CI green.

One non-blocking nit:

- **Test symmetry** `cmd/kuke/init/init_test.go`: the env-precedence regression guard only exists for the kukeond path. The init.go fix is structurally identical, so a mirrored test would catch a future divergence in either branch — not a blocker.

LGTM to merge.

PR #189, opened 2026-04-29
feat(kuke): introduce ClientConfiguration kind and --configuration flag

### LINKED ISSUE #160: kuke: introduce ClientConfiguration kind and --configuration flag
## Context
After the shared `--config` flag is removed (see #159), `kuke` has no
dedicated configuration document for client-side settings (default daemon
socket address, output preferences, …). We want a client-config file with
`~/.kube/config`-style ergonomics.

## Proposal
Introduce a `ClientConfiguration` kind (apiVersion `kukeon.io/v1beta1`) for
`kuke` client settings. Add a persistent `--configuration` flag on the `kuke`
cobra root that loads a ClientConfiguration document. Default lookup:
`~/.kuke/kuke.yaml`. If absent, kuke uses hardcoded defaults — no error.

Initial fields: client-only settings (default daemon socket address, output
formatting preferences, …). Exact list settled in code review. No shared
metadata with ServerConfiguration.

## Acceptance criteria
- [ ] New kind `ClientConfiguration` (apiVersion `kukeon.io/v1beta1`)
      registered in the apischeme; doc type under `pkg/api/model/v1beta1/`.
- [ ] `kuke --configuration <path>` persistent root flag loads
      ClientConfiguration from `<path>`.
- [ ] Default path is `~/.kuke/kuke.yaml`; absent file → hardcoded defaults,
      no error.
- [ ] Example `kuke.yaml` added under `docs/`.

## Notes
- Independent of #159; either order is fine, but the shared `--config` flag
  removal lives in #159.

eminwux:
## Summary
- Add `ClientConfiguration` (apiVersion `v1beta1`) kind for kuke client-side defaults — daemon endpoint, run path, containerd socket, log level. Mirrors PR #185's `ServerConfiguration` shape.
- Add `--configuration` persistent flag on the `kuke` cobra root (default `~/.kuke/kuke.yaml`); absent file is not an error and falls back to hardcoded defaults.
- PreRunE seeds viper from the YAML, but only for fields not already supplied by `--flag` or env var. Precedence: `--flag > env > ClientConfiguration > default`.
- `kuke apply` rejects this kind (it falls through to `ErrUnknownKind` in the parser, same as `ServerConfiguration`) — it is a client-side document, not a server resource.

## Non-obvious calls
- **apiVersion is `v1beta1`, not `kukeon.io/v1beta1`** — the issue body suggests the latter, but every existing kind in this repo uses bare `v1beta1` (see `Version` const in `pkg/api/model/v1beta1/consts.go`). Following the precedent set by PR #185.
- **Initial spec fields are kuke-only settings**: `host`, `runPath`, `containerdSocket`, `logLevel`. The issue notes the exact list is settled in code review — happy to add/remove fields per reviewer feedback.
- **`output` was considered but trimmed.** The `kuke get …` printer reads `cmd.Flags().GetString("output")` directly (`cmd/kuke/get/shared/shared.go:54`) and bypasses viper entirely. Wiring `output: yaml` from YAML through viper would silently no-op, which violates the project rule of not shipping non-functional features. Happy to file a follow-up to retrofit the printer to read via viper if reviewer wants `output` in the next iteration.
- **`flagChanged` helper checks both `cmd.Flags()` and `cmd.PersistentFlags()`.** This differs from `cmd/kukeond/kukeond.go`'s pattern (which only checks persistent flags). The kuke root has subcommand-level flags merged in at Execute time, and depending on which command path is being executed, a flag may live in either set during PreRunE; checking both is the safe read. The test `TestApplyClientConfigurationFlagOverridesConfig` would have caught this if I'd used the kukeond pattern verbatim.
- **Env-over-YAML test ships from day one.** PR #185's reviewer flagged that the equivalent test was missing for `cmd/kuke/init/init.go`'s structurally identical path; that gap is closed here in `TestApplyClientConfigurationEnvOverridesConfig`.

## Test plan
- [x] `make test` — all packages green, including new `internal/clientconfig` and `cmd/kuke` tests.
- [x] `go vet ./...` — clean.
- [x] Smoke test per project CLAUDE.md: `kuke init` rebuild + daemon-parity check (`kuke get realms` vs `--no-daemon`) — identical output. Daemon path untouched by this PR; smoke run was a regression-guard sanity check.
- [x] Manual: created `~/.kuke/kuke.yaml` with non-default `host`, ran `kuke get realms`, observed daemon connection attempt to the configured socket. With `KUKEON_HOST` exported, env wins over YAML. With `--host` on CLI, flag wins over env.

Closes #160

### DIFF
--- a/cmd/config/defaults.go
+++ b/cmd/config/defaults.go
@@ -70,3 +70,15 @@ func DefaultProfilesFile() string {
 func DefaultServerConfigurationFile() string {
 	return filepath.Join("/", "etc", "kukeon", "kukeond.yaml")
 }
+
+// DefaultClientConfigurationFile is the on-disk path the `kuke` client reads
+// when the user does not pass `--configuration`. An absent file is not an
+// error — the client falls back to its hardcoded defaults.
+func DefaultClientConfigurationFile() string {
+	base, err := os.UserHomeDir()
+	if err != nil {
+		// fallback to tmp if home dir cannot be determined
+		base = "tmp"
+	}
+	return filepath.Join(base, ".kuke", "kuke.yaml")
+}
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -100,6 +100,8 @@ var (
 	KUKEON_ROOT_HOST = DefineKV("KUKEON_HOST", "kukeon/host", "unix:///run/kukeon/kukeond.sock")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKEON_ROOT_NO_DAEMON = DefineKV("KUKEON_NO_DAEMON", "kukeon/noDaemon", "false")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_CONFIGURATION = DefineKV("KUKE_CONFIGURATION", "kuke/configuration")
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKEOND_SOCKET = DefineKV("KUKEOND_SOCKET", "kukeond/socket", "/run/kukeon/kukeond.sock")
--- a/cmd/kuke/clientconfig_test.go
+++ b/cmd/kuke/clientconfig_test.go
@@ -1,0 +1,216 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package kuke
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/eminwux/kukeon/cmd/config"
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
+	"github.com/spf13/viper"
+)
+
+func bindKukeEnv(t *testing.T) {
+	t.Helper()
+	for _, v := range []config.Var{
+		config.KUKEON_ROOT_HOST,
+		config.KUKEON_ROOT_RUN_PATH,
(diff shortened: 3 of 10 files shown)


eminwux:
**PR #189 Review — `feat(kuke): introduce ClientConfiguration kind and --configuration flag`**

Clean mirror of #185's ServerConfiguration pattern. New kind plumbed end-to-end, errdefs sentinel added, license headers present, `apply` parser correctly rejects this kind via `ErrUnknownKind` (verified at `internal/apply/parser/parser.go:165`). The `flagChanged` helper that checks both flag sets is a real improvement over `kukeond`'s pattern. No sensitive data in error paths. One precedence gap before merge.

A few items:

- **Env precedence broken for host/containerd-socket** `cmd/kuke/kuke.go:207-212`: `loadConfig` only `BindEnv`s `KUKEON_ROOT_RUN_PATH` and `KUKEON_ROOT_LOG_LEVEL`, so `KUKEON_HOST` and `KUKEON_CONTAINERD_SOCKET` are never read by `viper.GetString` in production. With YAML `host` + `KUKEON_HOST` env both set, `applyClientConfiguration`'s `envSet` check correctly skips the YAML write — but downstream `viper.Get` returns the flag default since env isn't bound, so env AND YAML are silently ignored. Contradicts the `--flag > env > ClientConfiguration > default` order asserted by `TestApplyClientConfigurationEnvOverridesConfig` (which passes only because the test explicitly calls `bindKukeEnv` to bind all four). Add `_ = config.KUKEON_ROOT_HOST.BindEnv()` and `_ = config.KUKEON_ROOT_CONTAINERD_SOCKET.BindEnv()` in `loadConfig`. (Same gap exists post-#185 in `kukeond` for `KUKEON_CONTAINERD_SOCKET` — worth a follow-up.)

LGTM once the BindEnv gap is closed.

eminwux:
Addressed in b84bbf7:
- Added `BindEnv()` for `KUKEON_ROOT_HOST` and `KUKEON_ROOT_CONTAINERD_SOCKET` in `loadConfig()` so the `--flag > env > ClientConfiguration > default` order holds in production. Tests already covered this via `bindKukeEnv` and continue to pass.

Deferred: matching `KUKEON_CONTAINERD_SOCKET` gap in `kukeond` post-#185 — out of scope for this PR per reviewer note.

eminwux:
**PR #189 Re-review — `b84bbf7`**

The follow-up commit (`fix(kuke): bind KUKEON_HOST and KUKEON_CONTAINERD_SOCKET env vars in loadConfig`) is exactly the requested change — three lines added to `loadConfig` at `cmd/kuke/kuke.go:207-208`, no other diffs. `BindEnv` is invoked early enough (`loadConfig` runs from the `OnInitialize` config loader at `cmd/kuke/kuke.go:98`) for `viper.Get*` to read env values in production. Blocking item resolved.

LGTM.

PR #192, opened 2026-04-29
fix(cli): panic on uninstall flag-bind failure instead of returning nil

### LINKED ISSUE #172: cmd/kuke/uninstall: NewUninstallCmd swallows setupUninstallCmd error and returns nil
Surfaced as a non-blocking nit on PR #170.

`cmd/kuke/uninstall/uninstall.go:86-89`:

\`\`\`go
if err := setupUninstallCmd(cmd); err != nil {
    return nil
}
\`\`\`

If \`setupUninstallCmd\` ever fails (only realistic mode: \`viper.BindPFlag\` on a nil flag — currently unreachable, but a future flag rename or typo would land us there), \`NewUninstallCmd\` returns \`nil\`. The caller in \`cmd/kuke/kuke.go\`:

\`\`\`go
rootCmd.AddCommand(uninstallcmd.NewUninstallCmd())
\`\`\`

passes that \`nil\` straight into cobra's \`AddCommand\`, which panics on \`nil\` (it dereferences \`cmd.Name()\` to register).

Fix: either propagate the error up (probably means making \`NewUninstallCmd\` return \`(*cobra.Command, error)\` and threading through \`SetupKukeCmd\`, which is invasive), or panic on bind failure with a clear message — the latter is consistent with how unreachable-by-construction errors are typically handled in the rest of the tree.

Latent / bug-adjacent — the current code path is fine but the failure mode silently turns a setup bug into a panic in main rather than a startup error.

eminwux:
## Summary
- `cmd/kuke/uninstall/uninstall.go:46` previously swallowed any `setupUninstallCmd` error and returned a `nil` `*cobra.Command`. The caller in `cmd/kuke/kuke.go:141` passes that straight into `cobra.AddCommand`, which then panics inside cobra (it dereferences `cmd.Name()`), turning a setup bug into an opaque crash deep in command registration.
- Replace the silent `return nil` with `panic(fmt.Sprintf("kukeon: uninstall command misconfigured: %v", err))`. This matches the existing "unreachable by construction" pattern in `internal/cni/subnet.go:131` for `NewDefaultSubnetAllocator`. A future flag rename / typo now fails at startup with a clear message naming the offender, rather than at root-command registration.
- Function signature unchanged, so no caller updates needed.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` passes (uninstall package + full unit suite)
- [ ] Daemon-parity smoke test (`sudo ./kuke get realms` vs `--no-daemon`) — not run; this PR does not touch `/opt/kukeon` paths, the daemon, or anything the daemon-parity check guards. Pure CLI command-construction wiring.

Closes #172

### DIFF
--- a/cmd/kuke/uninstall/uninstall.go
+++ b/cmd/kuke/uninstall/uninstall.go
@@ -44,7 +44,11 @@ func NewUninstallCmd() *cobra.Command {
 	}
 
 	if err := setupUninstallCmd(cmd); err != nil {
-		return nil
+		// setupUninstallCmd only fails if viper.BindPFlag is handed a nil
+		// flag — unreachable by construction here. Panic surfaces a future
+		// flag-rename or typo at startup instead of letting nil flow into
+		// cobra.AddCommand and crash the root command registration.
+		panic(fmt.Sprintf("kukeon: uninstall command misconfigured: %v", err))
 	}
 	return cmd
 }


eminwux:
**PR #192 Review — `fix(cli): panic on uninstall flag-bind failure instead of returning nil`**

Tight, well-scoped fix. Replacing the silent `return nil` with a panic prevents `nil` from flowing into `cobra.AddCommand` and crashing root-command registration with an opaque message — the new panic names the offender. Pattern matches the existing `internal/cni/subnet.go:131` "unreachable by construction" convention, and the comment explains the WHY clearly. No correctness, sensitive-data, or dead-code concerns. CI green; daemon-parity smoke correctly skipped — this PR doesn't touch `/opt/kukeon` or the daemon.

LGTM to merge.

PR #194, opened 2026-04-29
fix(uninstall): surface residual namespace and force-drain orphaned tasks

### LINKED ISSUE #193: controller: kuke uninstall reports kuke-system realm "purged" while its containerd namespace survives
## Repro
On a host where `/opt/kukeon` is already gone (e.g. a previous partial uninstall, or a manual `rm -rf /opt/kukeon`) but the `kuke-system.kukeon.io` containerd namespace still exists with the running `kukeond` cell:

```
root@vmi2904685:~# kuke uninstall
kuke uninstall will remove ALL kukeon runtime state on this host:
  ...
Are you sure? (yes/no) yes
kuke uninstall:
  - /run/kukeon: removed
  - /opt/kukeon: absent
  - user "kukeon": removed
  - group "kukeon": absent
root@vmi2904685:~# ctr ns ls
NAME                  LABELS
kuke-system.kukeon.io
moby
moby_history
```

## Observed
The CLI report says `realm "kuke-system" (namespace "kuke-system.kukeon.io"): purged`, but the containerd namespace is still present afterward. A re-run of `kuke uninstall` is required (and may still loop if anything else is pinning the namespace).

## Root cause
Two compounding issues, both reproducible from the snippet above:

1. **The report is unfaithful.** `runner.PurgeRealm` (`internal/controller/runner/purge_realm.go:112-122`) swallows `DeleteNamespace` errors as a `WarnContext` and returns `nil`. The controller-level `PurgeRealm` then sets `outcome.Purged = true` (`internal/controller/uninstall.go:141-146`) regardless of whether the namespace was actually removed. So "purged" really means "we ran the cleanup function without a fatal error", not "the namespace is gone."

2. **The metadata-missing fallback is too thin.** When `/opt/kukeon` is absent, `collectRealmsForUninstall` (`internal/controller/uninstall.go:193-223`) correctly falls back to the well-known realm list, but `purgeRealmCascade` (`internal/controller/purge_realm.go:140-184`) sees `metadataExists == false` and **skips `deleteRealmCascade`** — the cell-aware path that stops/kills the kukeond cell's containers cleanly. Cleanup then relies entirely on `runner.PurgeRealm`'s orphan-container path (`internal/controller/runner/helpers.go:609`), which calls `StopContainer` with default opts (SIGTERM, no Force) and `DeleteContainer` with all errors `_ = ` ignored. Anything that survives that best-effort drain (a stuck task, a leftover lease, content references not covered by `CleanupNamespaceResources` — which only walks images + the overlayfs snapshotter) blocks `containerd.NamespaceService.Delete` and we hit the silent-warning path from (1).

`default` (empty, no daemon, no leases) tears down cleanly — only `kuke-system` exposes the bug because that is where the daemon and its image/snapshot/lease graph live.

## Suggested fix
Two narrowly-scoped changes:

1. Make `runner.PurgeRealm` propagate `DeleteNamespace` failure (or surface it on the returned `RealmPurgeOutcome`) so the report can render `FAILED: namespace not empty` instead of `purged`. The simplest shape is an extra `NamespaceRemoved bool` on `RealmPurgeOutcome` plus a non-nil `Err` when removal fails, leaving the rest of the best-effort cleanup in place.
2. In the metadata-missing fallback, escalate the orphan-container drain: pass `StopContainerOptions{Force: true}` (or skip straight to `task.Delete` with `WithProcessKill`) and also enumerate + release containerd `leases` in the namespace before `DeleteNamespace`. That covers the "namespace not empty due to leases/content" case that `CleanupNamespaceResources` doesn't currently address.

Closed issue #171 fixed the snapshot-abort path; this is the residual class of pinning resources that still slip through.

eminwux:
## Summary
- Stop swallowing `DeleteNamespace` failures in `runner.PurgeRealm` — return `(namespaceRemoved bool, err error)` so the uninstall reporter can render `FAILED: ...` instead of a misleading `purged` when the containerd namespace actually survived.
- Force-stop orphaned containers during the realm-purge drain (`StopContainerOptions{Force: true}`). The default SIGTERM-with-grace path leaves SIGKILL on the table, so a stuck task pinned its snapshot mount and blocked `DeleteNamespace` with "namespace not empty" — exactly the residue the issue describes.
- Enumerate and synchronously release leftover containerd leases inside `CleanupNamespaceResources` before the namespace delete attempt, since a pinned lease keeps content/snapshot references alive across the cleanup pass.
- Propagate a new `NamespaceRemoved bool` through `controller.PurgeRealmResult` and `controller.RealmPurgeOutcome` so callers can distinguish "we ran cleanup without a fatal error" from "the namespace was actually removed."

## Notable judgment calls
- The public `kukeonv1.PurgeRealmResult` (over the daemon RPC) was left unchanged — issue #193 is about the uninstall report, not the API surface. Adding a field there would extend scope past the bug.
- `processOrphanedContainers` always force-stops now, not just on the metadata-missing fallback path. A purge is destructive by intent in both branches, and the same residual-task class can pin snapshots in either case; opting into Force unconditionally is strictly safer and avoids a flag plumbing exercise.
- Existing snapshot/blob/image cleanup is best-effort (warn-and-continue); the new lease cleanup follows the same shape so a single bad lease doesn't strand the rest of the namespace teardown.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` passes
- [x] `TestUninstall_PurgeFailureIsRecordedButCleanupContinues` extended to assert `NamespaceRemoved=false` on the failing realm and `{Purged, NamespaceRemoved}=true` on the successful one — locks in the new contract.
- [ ] Live repro from the issue (`rm -rf /opt/kukeon` + populated `kuke-system.kukeon.io` namespace, then `kuke uninstall`) — not run; this host already had a kukeond cell I did not want to clobber to reproduce the exact bug-state, and the daemon-parity smoke check in `kukeon/CLAUDE.md` covers `kuke init`, not the uninstall residue path. The unit test exercises the propagation; an end-to-end repro is a worthwhile manual follow-up.

Closes #193

### DIFF
--- a/internal/controller/controller_test.go
+++ b/internal/controller/controller_test.go
@@ -84,7 +84,7 @@ type fakeRunner struct {
 	ExistsCgroupFn func(doc any) (bool, error)
 
 	// Purge methods
-	PurgeRealmFn     func(realm intmodel.Realm) error
+	PurgeRealmFn     func(realm intmodel.Realm) (bool, error)
 	PurgeSpaceFn     func(space intmodel.Space) error
 	PurgeStackFn     func(stack intmodel.Stack) error
 	PurgeCellFn      func(cell intmodel.Cell) error
@@ -380,11 +380,11 @@ func (f *fakeRunner) ExistsCgroup(doc any) (bool, error) {
 
 // Purge methods
 
-func (f *fakeRunner) PurgeRealm(realm intmodel.Realm) error {
+func (f *fakeRunner) PurgeRealm(realm intmodel.Realm) (bool, error) {
 	if f.PurgeRealmFn != nil {
 		return f.PurgeRealmFn(realm)
 	}
-	return errors.New("unexpected call to PurgeRealm")
+	return false, errors.New("unexpected call to PurgeRealm")
 }
 
 func (f *fakeRunner) PurgeSpace(space intmodel.Space) error {
--- a/internal/controller/purge_realm.go
+++ b/internal/controller/purge_realm.go
@@ -28,13 +28,14 @@ import (
 
 // PurgeRealmResult reports what was purged during realm purging.
 type PurgeRealmResult struct {
-	Realm          intmodel.Realm
-	RealmDeleted   bool     // Whether realm deletion succeeded
-	PurgeSucceeded bool     // Whether comprehensive purge succeeded
-	Force          bool     // Force flag that was used
-	Cascade        bool     // Cascade flag that was used
-	Deleted        []string // Resources that were deleted (standard cleanup)
-	Purged         []string // Additional resources purged (CNI, orphaned containers, etc.)
+	Realm            intmodel.Realm
+	RealmDeleted     bool     // Whether realm deletion succeeded
+	PurgeSucceeded   bool     // Whether comprehensive purge succeeded
+	NamespaceRemoved bool     // Whether the containerd namespace was actually removed
+	Force            bool     // Force flag that was used
+	Cascade          bool     // Cascade flag that was used
+	Deleted          []string // Resources that were deleted (standard cleanup)
+	Purged           []string // Additional resources purged (CNI, orphaned containers, etc.)
 }
 
 // PurgeRealm purges a realm with comprehensive cleanup. If cascade is true, purges all spaces first.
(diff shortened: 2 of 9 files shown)


eminwux:
**PR #194 Review — `fix(uninstall): surface residual namespace and force-drain orphaned tasks`**

Solid fix for #193. The load-bearing change is propagating `DeleteNamespace`'s error up to `outcome.Err` — that alone flips the renderer at `cmd/kuke/uninstall/uninstall.go:168-172` from "purged" to "FAILED" via the `r.Err != nil` branch. Force-stop in `processOrphanedContainers` and the synchronous lease drain are reasonable defenses. No security or data-leak concerns.

A few small items:

- **Redundant field** `internal/controller/uninstall.go:60-65` and `internal/controller/purge_realm.go:30-38`: in `runner/purge_realm.go:120-134`, `namespaceRemoved=false` is set only inside the same `DeleteNamespace failed` branch as `nsErr`, so `NamespaceRemoved == false ⇔ Err != nil`. The field carries strictly the same information as `Err`/`Purged` after propagation; the renderer never reads it (only the unit test does). The PR description's "distinguish cleanup-ran-without-fatal-error from namespace-was-removed" framing doesn't match the implementation since both conditions reduce to the same thing. Either drop it, or decouple it in the runner (e.g. demote `nsErr` once other cleanup succeeds) so the bool can genuinely diverge.
- **Live repro deferred** test plan: the populated-namespace `kuke uninstall` scenario from #193 isn't end-to-end verified. The unit test covers the propagation contract; author noted why E2E wasn't run — worth a manual follow-up on a fresh host.

LGTM to merge — both items are non-blocking polish.

PR #196, opened 2026-04-29
fix(uninstall): stop kukeond before purging kuke-system

### LINKED ISSUE #195: uninstall: stop kukeond before purging kuke-system
## Context
PR #194 closed #193 by force-killing orphaned tasks, releasing leases, and surfacing the residual-namespace signal in the uninstall report. The reporter half is working — failures now render correctly — but the cleanup half still fails on the same end-to-end repro called out in PR #194's test plan (the live repro item that was not run before merge).

On a host where `kuke init` produced a running `kukeond` cell in `kuke-system.kukeon.io`, `kuke uninstall --yes`:

```
kuke uninstall:
  - realm "kuke-system" (namespace "kuke-system.kukeon.io"): FAILED: failed to delete containerd namespace "kuke-system.kukeon.io": failed to delete namespace: namespace "kuke-system.kukeon.io" must be empty, but it still has containers: failed precondition
  - /run/kukeon: absent
  - /opt/kukeon: absent
  - user "kukeon": absent
  - group "kukeon": absent
Error: purge realm "kuke-system": failed to delete containerd namespace "kuke-system.kukeon.io": ...
```

Note the symptom changed: #193 saw the namespace surviving silently with stuck *tasks*; this issue is about *containers* surviving — a different blocking actor than the one PR #194 hardened against.

## Suspected root cause
`internal/controller/uninstall.go` runs the in-process `controller.Uninstall` which goes straight to `PurgeRealm` for both well-known realms. Nothing stops the running `kukeond` daemon first. While `kukeond` is alive its containerd session pins references on its own cell's containers; the orphan drain in `processOrphanedContainers` (`internal/controller/runner/helpers.go:609-643`) discards every Stop/Delete error with `_ =`, so a fight between the in-process drain and the live daemon never surfaces in the logs — only as the downstream "namespace not empty" precondition failure.

A secondary problem worth fixing in the same PR: even if the primary cause turns out to be different, the silent `_ =` calls mean we cannot tell *which* container blocked the namespace delete from the user-visible error. The failure message names the namespace but not the container ID(s).

A third gap: the well-known-realm fallback only covers `default` and `kuke-system`. A user who ran `kuke create realm <foo>` ends up with namespace `<foo>.kukeon.io`; if metadata under `/opt/kukeon` is wiped before `kuke uninstall` (the #193 partial-state path), that namespace is never enumerated and survives the uninstall silently.

## Proposal
1. **Stop kukeond before per-realm purge.** Add a daemon-stop step at the top of `controller.Uninstall` (before step 1 in the docstring): if `/run/kukeon/kukeond.pid` exists, send SIGTERM, wait up to ~5s for the process to exit, fall back to SIGKILL. If the PID file is absent (the partial-uninstall path from #193), no-op. Record the action in `UninstallReport` so the CLI prints it in the existing report layout.
2. **Surface drain failures.** In `processOrphanedContainers`, replace the two `_ =` calls with `WarnContext` logs that include the container ID and error. After the drain loop, re-list the namespace; if any containers remain, include their IDs in the error returned by `runner.PurgeRealm` so the report says *which* containers blocked the namespace delete instead of the generic precondition message.
3. **Exhaustive per-namespace cleanup.** For every kukeon-owned containerd namespace, drain in this exact order so `DeleteNamespace`'s "must be empty" precondition is unreachable:
   1. tasks (force-kill any survivors),
   2. containers,
   3. images,
   4. snapshots (every snapshotter, not just `overlayfs`),
   5. leases,
   6. namespace.
   `CleanupNamespaceResources` today walks images + the `overlayfs` snapshotter only; expand it to enumerate snapshotters from `client.SnapshotService("")` (or equivalent) and to release leases via `client.LeasesService().List` + `Delete`. Each step logs what it removed (count + IDs at DEBUG, summary at INFO) so a future "namespace not empty" failure points at the exact resource class still pinning it.
4. **Enumerate kukeon namespaces by suffix, not by name.** Replace the hardcoded `default` + `kuke-system` well-known list in `collectRealmsForUninstall` (`internal/controller/uninstall.go:202-232`) with a containerd enumeration: list every namespace via `client.NamespaceService().List`, filter by `strings.HasSuffix(ns, ".kukeon.io")`, and purge each one (merging with on-disk metadata when present). This catches user-created realms whose metadata was wiped before `kuke uninstall`.

## Acceptance criteria
- [ ] Live repro: on a host where `kuke init --kukeond-image …:dev` produced a running `kukeond` cell, `kuke uninstall --yes` succeeds — both realms render `purged`, `Error:` line is absent, and `ctr ns ls` no longer shows `kuke-system.kukeon.io`.
- [ ] Custom-realm repro: after `kuke init` + `kuke create realm myteam` + `rm -rf /opt/kukeon`, `kuke uninstall --yes` removes `myteam.kukeon.io` from `ctr ns ls` even though no on-disk metadata is left to enumerate it.
- [ ] If `/run/kukeon/kukeond.pid` exists at the start of `Uninstall`, the daemon process is signalled (SIGTERM, then SIGKILL after a short grace) before any realm purge.
- [ ] If the PID file is absent, the daemon-stop step is a no-op — the partial-uninstall path from #193 still works (verified by an existing or new test).
- [ ] `UninstallReport` records whether the daemon was stopped; the CLI prints it as a new line in the existing report block.
- [ ] `processOrphanedContainers` no longer discards `StopContainer` / `DeleteContainer` errors; both are logged at WARN with the container ID.
- [ ] When containers survive the drain, the error returned by `runner.PurgeRealm` names the residual container IDs (visible in the CLI's `FAILED: …` line).
- [ ] `CleanupNamespaceResources` (or the equivalent purge sequence) explicitly removes — in order — tasks, containers, images, snapshots (across all snapshotters), and leases before `DeleteNamespace`. Each class's removal is logged.
- [ ] `collectRealmsForUninstall` enumerates containerd namespaces by `.kukeon.io` suffix, merging with on-disk metadata. Hardcoded `default` / `kuke-system` short-circuiting is replaced (or kept only as a safety floor when the containerd enumeration returns nothing).
- [ ] Unit test exercises the daemon-stop step with an injected fake-PID/stub signal sender, covering both PID-present and PID-absent cases.
- [ ] Unit test for the suffix enumerator: a stub namespace lister returning `["default.kukeon.io", "kuke-system.kukeon.io", "myteam.kukeon.io", "moby"]` yields exactly the three `.kukeon.io` namespaces for purge.

## Notes
- Follow-up to #193 / PR #194. Same end-state goal as #193 but a different blocking actor — the live `kukeond` process rather than orphaned tasks/leases.
- PR #194's test plan left the live end-to-end repro unchecked (test plan item: *"Live repro from the issue (rm -rf /opt/kukeon + populated kuke-system.kukeon.io namespace, then kuke uninstall) — not run"*). This issue is exactly that gap.
- Reproduces on the host smoke-test path documented in `kukeon/CLAUDE.md` (the `kuke init` → `kuke uninstall` round trip).
- Sizing note for dev: this issue bundles four narrowly-related fixes that all serve the same end (uninstall succeeds on a populated host with stale metadata). If any one piece grows beyond a single reviewable PR, escalate via `resize-issue` rather than half-landing.

eminwux:
## Summary
- Stop the live kukeond daemon (SIGTERM, then SIGKILL after a 5s grace) at the top of `controller.Uninstall` before any realm purge, so the daemon's containerd session can no longer pin its own cell's containers in `kuke-system.kukeon.io` while the in-process drain is trying to delete them. Gated on a PID file (default `<SocketDir>/kukeond.pid`); absent file is a no-op for the partial-uninstall path from #193. The stopper is injectable via `UninstallOptions.DaemonStopper` so it can be unit-tested without forking processes.
- Surface orphan-drain failures: `processOrphanedContainers` now `WarnContext`-logs every Stop/Delete error with the container ID instead of `_ =`-discarding it, and `runner.PurgeRealm` re-lists residual containers after a failed `DeleteNamespace` and embeds their IDs in the error so the CLI prints which container blocked the namespace delete.
- Exhaustive per-namespace cleanup: `ctr.CleanupNamespaceResources` now walks every snapshotter (`overlayfs`, `native`, `btrfs`, `zfs`, `devmapper`, `blockfile`) when no snapshotter is pinned, so a host that experimented with a non-default snapshotter does not strand its leftover snapshots. Lease drain stays as the final step before `DeleteNamespace`. Unregistered snapshotters log at DEBUG (no WARN-spam on stock hosts).
- Enumerate kukeon namespaces by `.kukeon.io` suffix: `collectRealmsForUninstall` now merges on-disk metadata (when present), the live containerd namespace list filtered by suffix (catches user-created realms whose `/opt/kukeon` was wiped — issue #193's partial-state path), and the two well-known realms as a safety floor when containerd enumeration returns nothing.
- Report layout: `UninstallReport.Daemon` carries the daemon-stop outcome; the CLI prints a new line `- kukeond: <state>` (absent / stopped (pid N) / force-killed (pid N) / stale pid file).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full unit suite, all packages pass.
- [x] New unit tests in `internal/controller/uninstall_test.go`:
  - `TestUninstall_DaemonStopStep_PIDPresent` — injects a stub `DaemonStopper`, asserts it received the configured pidFile + grace and that it ran *before* any `PurgeRealmFn` call (the whole point of step 0).
  - `TestUninstall_DaemonStopStep_PIDAbsent` — stub returns `PIDFilePresent=false`; verifies no error, the realm-purge loop still runs, and `report.Daemon.Signalled=false`.
  - `TestUninstall_SuffixEnumeratorPurgesKukeonNamespacesOnly` — `ListContainerdNamespacesFn` returns `["default.kukeon.io", "kuke-system.kukeon.io", "myteam.kukeon.io", "moby"]`; asserts exactly the three `.kukeon.io` realms get a `PurgeRealm` call and `moby` is filtered out.
- [x] New unit tests in `internal/controller/daemon_stop_test.go` covering the production stopper's no-op branches: empty pidFile, missing file (ENOENT), garbage PID (parse error), and PID 1 (init refusal). All return `nil` error and `Signalled=false`.
- [ ] **Live repro from the AC** (`kuke init` → populated `kuke-system.kukeon.io` → `kuke uninstall --yes` succeeds, `ctr ns ls` no longer shows `kuke-system.kukeon.io`) — not run from the dev environment; needs a host with containerd + the local `kukeon-local:dev` image. Reviewer should run this on a staging host before merge per `kukeon/CLAUDE.md`'s smoke-test policy.
- [ ] **Custom-realm partial-state repro** (`kuke init` + `kuke create realm myteam` + `rm -rf /opt/kukeon` + `kuke uninstall --yes` clears `myteam.kukeon.io`) — same caveat; not run locally.
- [ ] **Daemon-parity smoke check** from `kukeon/CLAUDE.md` (`kuke get realms` vs. `--no-daemon`) — not applicable: this PR does not touch the realm read path the daemon and in-process controller share. Only the uninstall write path changed.

Closes #195

### DIFF
--- a/cmd/kuke/uninstall/uninstall.go
+++ b/cmd/kuke/uninstall/uninstall.go
@@ -163,6 +163,9 @@ func readLine(r io.Reader) (string, error) {
 func printReport(cmd *cobra.Command, report controller.UninstallReport) {
 	out := cmd.OutOrStdout()
 	fmt.Fprintln(out, "kuke uninstall:")
+	if report.Daemon.PIDFile != "" {
+		fmt.Fprintf(out, "  - kukeond: %s\n", daemonOutcome(report.Daemon))
+	}
 	for _, r := range report.Realms {
 		switch {
 		case r.Err != nil:
@@ -177,6 +180,23 @@ func printReport(cmd *cobra.Command, report controller.UninstallReport) {
 	fmt.Fprintf(out, "  - group %q: %s\n", report.GroupName, accountOutcome(report.GroupExisted, report.GroupRemoved))
 }
 
+// outcomeAbsent labels every "we didn't find the resource" branch in the
+// uninstall report so the operator sees the same word in every section.
+const outcomeAbsent = "absent"
+
+func daemonOutcome(d controller.DaemonStopReport) string {
+	switch {
+	case !d.PIDFilePresent:
+		return outcomeAbsent
+	case d.ForceKilled:
+		return fmt.Sprintf("force-killed (pid %d)", d.PID)
+	case d.Signalled:
+		return fmt.Sprintf("stopped (pid %d)", d.PID)
+	default:
+		return "stale pid file"
+	}
+}
+
 func dirOutcome(existed, removed bool) string {
 	switch {
 	case removed:
@@ -184,7 +204,7 @@ func dirOutcome(existed, removed bool) string {
 	case existed:
 		return "remove failed"
 	default:
-		return "absent"
+		return outcomeAbsent
 	}
 }
 
@@ -195,7 +215,7 @@ func accountOutcome(existed, removed bool) string {
 	case existed:
 		return "remove failed"
 	default:
-		return "absent"
+		return outcomeAbsent
 	}
 }
 
--- a/internal/consts/consts.go
+++ b/internal/consts/consts.go
@@ -63,9 +63,9 @@ const (
 	KukeSystemCellName      = "kukeond"
 	KukeSystemContainerName = "kukeond"
 
-	// realmNamespaceSuffix is the suffix appended to every realm name to form
-	// its containerd namespace. See RealmNamespace.
-	realmNamespaceSuffix = ".kukeon.io"
+	// RealmNamespaceSuffix is the suffix appended to every realm name to form
+	// its containerd namespace. See RealmNamespace and IsKukeonNamespace.
+	RealmNamespaceSuffix = ".kukeon.io"
 
 	// KukeonSystemUser and KukeonSystemGroup name the system identity created
 	// by `kuke init` so a non-root operator added to the kukeon group can
@@ -80,5 +80,27 @@ const (
 // realm name; all bootstrap and user-realm code paths route through it so the
 // mapping stays consistent.
 func RealmNamespace(realm string) string {
-	return realm + realmNamespaceSuffix
+	return realm + RealmNamespaceSuffix
+}
+
+// IsKukeonNamespace reports whether ns is a containerd namespace owned by
+// kukeon — i.e., one with the canonical .kukeon.io suffix and a non-empty
+// realm prefix. Used by the uninstall path to enumerate kukeon namespaces by
+// suffix so user-created realms whose on-disk metadata was wiped (issue #193's
+// partial-uninstall path) are still purged on a `kuke uninstall`.
+func IsKukeonNamespace(ns string) bool {
+	if len(ns) <= len(RealmNamespaceSuffix) {
+		return false
+	}
+	return ns[len(ns)-len(RealmNamespaceSuffix):] == RealmNamespaceSuffix
+}
+
+// RealmFromNamespace returns the realm name encoded in a containerd namespace
+// (the inverse of RealmNamespace). Returns the empty string when ns does not
+// have the kukeon suffix or when stripping the suffix would leave nothing.
+func RealmFromNamespace(ns string) string {
+	if !IsKukeonNamespace(ns) {
+		return ""
+	}
+	return ns[:len(ns)-len(RealmNamespaceSuffix)]
 }
--- a/internal/controller/controller_test.go
+++ b/internal/controller/controller_test.go
@@ -41,6 +41,7 @@ type fakeRunner struct {
 	CreateRealmFn                    func(realm intmodel.Realm) (intmodel.Realm, error)
 	EnsureRealmFn                    func(realm intmodel.Realm) (intmodel.Realm, error)
 	ExistsRealmContainerdNamespaceFn func(namespace string) (bool, error)
+	ListContainerdNamespacesFn       func() ([]string, error)
 	DeleteRealmFn                    func(realm intmodel.Realm) error
 
 	// Space methods
@@ -147,6 +148,15 @@ func (f *fakeRunner) ExistsRealmContainerdNamespace(namespace string) (bool, err
 		return f.ExistsRealmContainerdNamespaceFn(namespace)
 	}
 	return false, errors.New("unexpected call to ExistsRealmContainerdNamespace")
+}
+
+func (f *fakeRunner) ListContainerdNamespaces() ([]string, error) {
+	if f.ListContainerdNamespacesFn != nil {
+		return f.ListContainerdNamespacesFn()
+	}
+	// Default to an empty list so tests that don't care about the suffix
+	// enumerator (the dominant case) don't have to install a stub.
+	return nil, nil
 }
 
 func (f *fakeRunner) DeleteRealm(realm intmodel.Realm) error {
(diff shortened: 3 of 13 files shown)


eminwux:
**PR #196 Review — `fix(uninstall): stop kukeond before purging kuke-system`**

Solid fix for #195. The daemon-stop step is correctly sequenced before the realm-purge loop (the test pins this with shared call-index counters), the suffix enumerator catches user-realms whose `/opt/kukeon` was wiped (#193 path) without ever touching `moby` or other co-tenant namespaces, and `CleanupNamespaceResources` now walks the full snapshotter set instead of just `overlayfs`. Unit-test coverage is good: ordering, PID-absent, garbage-PID, PID-1 refusal, suffix filter. License headers, sentinel placement, and test colocation all comply.

- **Nit** `internal/consts/consts.go:91-95`: `IsKukeonNamespace` reinvents `strings.HasSuffix`. `return len(ns) > len(RealmNamespaceSuffix) && strings.HasSuffix(ns, RealmNamespaceSuffix)` is the idiomatic form.
- **Nit** `internal/consts/consts.go:68`: `RealmNamespaceSuffix` is now exported but has no cross-package callers — `IsKukeonNamespace`/`RealmFromNamespace` are the only users and live in the same package, so it could remain lowercase.
- **Nit** `internal/ctr/namespaces.go:204`: `KukeonKnownSnapshotters` is similarly only referenced inside `internal/ctr/` (the `runner/delete_realm.go:67` reference is a comment). Could be lowercase or kept exported as a documented contract — author's call.
- **Test plan** (FYI, non-blocking): the live repros from the AC and the custom-realm partial-state path are explicitly deferred to a staging host. Per `kukeon/CLAUDE.md` smoke-test policy, run those before merging.

LGTM.

PR #198, opened 2026-04-29
refactor(cellprofile): drop singleton mode; spec.prefix overrides metadata.name

### LINKED ISSUE #197: cellprofile: drop singleton mode; spec.prefix is optional override
## Context

The current `CellProfile` encodes singleton-vs-template in a spec field (`spec.namePrefix`):

- `namePrefix` unset → cell name = `metadata.name`, idempotent (singleton)
- `namePrefix` set → cell name = `<namePrefix>-<6hex>`, fresh each invocation (template)

This is field-based polymorphism over a behavior the *kind* should already encode. With both `Cell` and `CellProfile` in the API, the kinds can carry the meaning directly. Today `metadata.name` is also never used as a prefix — in singleton mode it's the literal cell name, in template mode it's bypassed entirely — so the field's relationship to `metadata.name` is ambiguous to a reader.

A related wart in the same area: `CellProfileSpec.Cell.ID` survives `Materialize` verbatim when set (`internal/cellprofile/cellprofile.go:190-192`), so a profile that hardcodes it produces N cells all sharing the same ID. Same design instinct as the singleton/template toggle.

## Proposal

Make the kind boundary load-bearing:

- **`Cell`** → singleton. Name = `metadata.name`. Idempotent on re-run. Unchanged.
- **`CellProfile`** → always template. Cell name = `<prefix>-<6hex>` where `prefix` defaults to `metadata.name`, optionally overridden by `spec.prefix`. Singleton mode is gone — that's what `Cell` is for.

Mechanical changes:

1. `internal/cellprofile/cellprofile.go:205-215` (`resolveCellName`) — collapse the two-branch logic; always append `-<6hex>`. Use `spec.prefix` if set, else `metadata.name`.
2. `pkg/api/model/v1beta1/cellprofile.go:44-50` — rename `NamePrefix` field to `Prefix`; YAML/JSON tags from `namePrefix` to `prefix`. Rewrite the doc comment: now an optional override of the default prefix derived from `metadata.name`, not a singleton/template toggle.
3. `internal/cellprofile/cellprofile.go` (`loadFile`) — reject profiles that set `spec.cell.id`. Loading a profile with a non-empty `spec.cell.id` returns `errdefs.ErrProfileInvalid`. The field is meaningless on a profile whose materialized cells have generated names; better to fail loudly than silently overwrite.
4. `cmd/kuke/run/run.go:274-275` — update the comment block to drop the singleton branch.

## Acceptance criteria

- [ ] `Materialize` always returns `<prefix>-<6hex>`; prefix = `spec.prefix` if set, else `metadata.name`.
- [ ] `CellProfileSpec.Prefix` field replaces `NamePrefix`; YAML/JSON tag is `prefix`.
- [ ] Loading a profile with non-empty `spec.cell.id` returns `errdefs.ErrProfileInvalid` with a clear message naming the field.
- [ ] Re-running `kuke run -p <profile>` always creates a fresh cell — no idempotent-singleton path on profiles.
- [ ] Profile without `spec.prefix`: three successive invocations produce three distinct cells named `<metadata.name>-<6hex>`; all three carry `kukeon.io/profile=<metadata.name>`.
- [ ] Profile with `spec.prefix`: same behavior, prefix overridden.
- [ ] `CellProfileSpec.Prefix` doc comment rewritten — no "singleton vs template modes" framing.
- [ ] `TestMaterialize_DefaultName_Idempotent` removed; new test added covering default-prefix-from-`metadata.name`.
- [ ] `TestRun_FromProfile_NamePrefix_GeneratesFreshCell` and `TestMaterialize_NamePrefix_GeneratesUniqueCells` updated for the new field name and semantic.
- [ ] New test: loading a profile with `spec.cell.id` set surfaces `ErrProfileInvalid`.
- [ ] CLAUDE.md daemon-parity check (`kuke get realms` vs `--no-daemon`) noted in the test plan even though this is client-side.

## Notes

- Supersedes the singleton/template model from #157 / #158. The original spec didn't articulate the core distinction — `Cell` for singletons, `CellProfile` for multiple cells from one config — and instead made `CellProfile` cover both via a field toggle. This issue makes the kind boundary load-bearing.
- Mirrors K8s `Pod.metadata.name` (literal) vs `Deployment.metadata.name` (resource identity, with pod names derived) — `metadata.name` always means the resource's identity; what's *generated* from it differs by kind.
- Composes with the `kukeon.io/profile=<metadata.name>` label injection (PR #158); query path unchanged.
- **Breaking changes** (acceptable in v1beta1):
  - Profiles that omitted `namePrefix` and relied on idempotent re-attach now create fresh cells. Migration: rewrite as a `Cell` doc — `CellProfileSpec.Cell` is already a `CellSpec`, so the body copy is mechanical, then apply with `kuke run -f` instead of `-p`.
  - Profiles that set `namePrefix` must rename to `prefix`. Mechanical YAML edit.
  - Profiles that set `spec.cell.id` must remove it. Mechanical YAML edit.

eminwux:
## Summary
- `CellProfile` is now always a template — every `Materialize` returns `<prefix>-<6hex>`. Prefix is `spec.prefix` if set, otherwise `metadata.name`. The prior singleton mode (idempotent re-attach when `namePrefix` was unset) is gone; the `Cell` kind is what represents singleton workloads.
- Field rename `spec.namePrefix` → `spec.prefix` (YAML/JSON tags follow); doc comment rewritten to drop the singleton/template framing and cast `prefix` as a default-prefix override.
- `loadFile` now rejects profiles with non-empty `spec.cell.id` — that field would otherwise produce N cells sharing one ID once names are always generated. Returns `errdefs.ErrProfileInvalid` with a message naming the field.
- Tests updated for the new shape: `TestMaterialize_DefaultName_Idempotent` removed; new `TestMaterialize_DefaultPrefix_GeneratesUniqueCells` covers default-prefix-from-`metadata.name`; `TestRun_FromProfile_NamePrefix_GeneratesFreshCell` renamed and reshaped for `spec.prefix`. Added `TestLoad_RejectsCellID`.

## Notable judgment calls
- Removed the `if spec.ID == ""` guard around the cell-name assignment in `Materialize` because `loadFile` now rejects non-empty `spec.cell.id` upfront — keeping the guard would imply the field is meaningful, and the issue explicitly wants it surfaced as an invariant rather than silently overwritten.
- `kukeon.io/profile=<metadata.name>` label-of-origin is unchanged and still tracks `metadata.name`, not the (possibly overridden) prefix — operators query by profile identity, not by the string used to generate names.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [ ] CLAUDE.md daemon-parity check (`kuke get realms` vs `--no-daemon`) — not run; this PR is client-only (cellprofile materialization happens in-process before any daemon RPC) and touches no daemon-visible code paths or `/opt/kukeon` reads.

## Breaking changes
Acceptable in v1beta1 per the issue:
- Profiles that omitted `namePrefix` and relied on singleton/idempotent behavior now create fresh cells. Migration: rewrite as a `Cell` doc and apply with `kuke run -f`.
- Profiles that set `namePrefix` must rename the field to `prefix`.
- Profiles that set `spec.cell.id` must remove it (now rejected at load).

Closes #197

### DIFF
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -271,8 +271,8 @@ func loadFromFile(file string) (v1beta1.CellDoc, error) {
 
 // loadFromProfile resolves the active profiles directory, loads the named
 // profile, and materializes its CellSpec body into a CellDoc. The cell name
-// is the profile metadata.name when spec.namePrefix is unset, otherwise
-// `<namePrefix>-<6hex>` so each invocation produces a fresh cell.
+// is `<prefix>-<6hex>` — prefix defaults to metadata.name and is overridden
+// by spec.prefix — so every invocation produces a fresh cell.
 func loadFromProfile(profileName string) (v1beta1.CellDoc, error) {
 	dir, err := cellprofile.ResolveDir()
 	if err != nil {
--- a/cmd/kuke/run/run_test.go
+++ b/cmd/kuke/run/run_test.go
@@ -1072,8 +1072,8 @@ func TestRun_FromProfile_CreatesAndStarts(t *testing.T) {
 	if fc.createCalls != 1 {
 		t.Fatalf("CreateCell calls=%d want 1", fc.createCalls)
 	}
-	if got := fc.createDoc.Metadata.Name; got != "claude-cell" {
-		t.Errorf("cell name=%q want claude-cell (default to profile name)", got)
+	if got := fc.createDoc.Metadata.Name; !strings.HasPrefix(got, "claude-cell-") || len(got) != len("claude-cell-")+6 {
+		t.Errorf("cell name=%q want claude-cell-<6hex> (default prefix from metadata.name)", got)
 	}
 	if got := fc.createDoc.Spec.RealmID; got != "default" {
 		t.Errorf("RealmID=%q want default", got)
@@ -1089,11 +1089,9 @@ func TestRun_FromProfile_CreatesAndStarts(t *testing.T) {
 	}
 }
 
-func TestRun_FromProfile_NamePrefix_GeneratesFreshCell(t *testing.T) {
-	// spec.namePrefix flips the profile from singleton to template: every
-	// invocation must produce a distinct cell name shaped like
-	// `<prefix>-<6hex>`. Drive two invocations and assert both names share
-	// the prefix and differ.
+func TestRun_FromProfile_PrefixOverride_GeneratesFreshCell(t *testing.T) {
+	// spec.prefix overrides the default prefix (metadata.name). Every
+	// invocation must produce a distinct cell name shaped `<prefix>-<6hex>`.
 	t.Cleanup(viper.Reset)
 	dir := t.TempDir()
 	const profileYAML = `apiVersion: v1beta1
@@ -1104,7 +1102,7 @@ spec:
   realm: default
   space: agents
   stack: claude
-  namePrefix: claude
+  prefix: agent
   cell:
     containers:
       - id: work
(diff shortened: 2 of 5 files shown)


eminwux:
**PR #198 Review — `refactor(cellprofile): drop singleton mode; spec.prefix overrides metadata.name`**

Clean, focused refactor. The rename is consistent (`git grep namePrefix` at head returns nothing), the new `spec.cell.id` rejection has a dedicated test, and the rationale for dropping the `if spec.ID == ""` guard in `Materialize` holds — the only production caller (`cmd/kuke/run/run.go:285`) routes through `Load`, which now rejects the field upfront. Doc comments on `CellProfileSpec`, `Materialize`, and `loadFromProfile` all updated to match the new mental model. No sensitive-data risks; `errdefs.ErrProfileInvalid` reused per the project convention.

Skipping the daemon-parity check is justified — the diff is entirely in `internal/cellprofile/` and `cmd/kuke/run/`, neither of which the daemon reads.

LGTM.

PR #209, opened 2026-04-29
feat(config): dump default Server/ClientConfiguration on first run

### LINKED ISSUE #204: config: dump default ClientConfiguration and ServerConfiguration on first run
## Context
`ServerConfiguration` (#159, #185) and `ClientConfiguration` (#160, #189) are now first-class kinds. But on first run the user has no template to work from — they have to consult source or examples to construct a valid YAML.

## Proposal
On first invocation:
- `kukeond serve` writes a default `ServerConfiguration` to a well-known path (e.g. `/etc/kukeon/kukeond.yaml`) if absent. The file contains every field with the in-binary default value, plus header comments explaining each field.
- `kuke` (any subcommand) writes a default `ClientConfiguration` to `~/.config/kukeon/kuke.yaml` (or `/etc/kukeon/kuke.yaml` for root) if absent.
- Existing files are never overwritten — first-write only.
- Both honor `--configuration` to opt out of the default location.

## Acceptance criteria
- [ ] First `kukeond serve` on a fresh host writes a fully-populated default `ServerConfiguration` YAML.
- [ ] First `kuke <any>` on a fresh host writes a fully-populated default `ClientConfiguration` YAML.
- [ ] Existing files at the default path are not overwritten.
- [ ] Default-path writes are skipped when `--configuration` is set.
- [ ] Generated YAML round-trips through the loader without errors.
- [ ] Header comment explains each field's purpose and default.

## Notes
- Pairs with #191 (kukeond env-var bind fix) for env-precedence to behave as documented.
- Cross-links the "default config dump" requirement called out in the gap-analysis review.

eminwux:
## Summary
- On first invocation, `kukeond serve` writes a fully-populated default `ServerConfiguration` YAML to `/etc/kukeon/kukeond.yaml` (every field present with header comments naming purpose + default).
- On first invocation, `kuke <any>` writes a fully-populated default `ClientConfiguration` YAML to `~/.kuke/kuke.yaml` (existing project default kept intact — see judgment call below).
- Existing files at the default path are never overwritten — `O_CREATE|O_EXCL` plus a parent `MkdirAll` close the TOCTOU race; an `os.ErrExist` returns `wrote=false, err=nil`.
- Errors during the dump are intentionally swallowed in the wiring helpers: a read-only `/etc` or `$HOME` must not block the daemon or client from running on hardcoded defaults.
- Unit tests cover the three behaviors directly (round-trip via `Load`, no-overwrite, custom-path skip) and the CLI-wiring tests cover the path-equality guard + the happy default-path write under a sandboxed `$HOME`.

## Notable judgment calls
- **Guard semantics — path-equality only, no flag-changed check.** Reading AC #4 literally ("skipped when `--configuration` is set") would silence the dump when `kuke init` plumbs `--configuration /etc/kukeon/kukeond.yaml` into the kukeond cell — which would defeat AC #1 on a fresh `kuke init`. The issue's Notes section says "Both honor `--configuration` to opt out of the default location," so I read AC #4 as "opt out when pointing at a custom path" and implemented that with a single `path == DefaultXxxFile()` check. Pointing `--configuration` at a custom path opts out cleanly; pointing it at the default path is a no-op overwrite-guarded by `O_EXCL`. Reviewer can swap to a strict literal reading by adding the flag-changed check back, but that path requires a parallel `kuke init` hook to satisfy AC #1.
- **ClientConfiguration default path kept at `~/.kuke/kuke.yaml`** (the existing project default), not the issue body's `~/.config/kukeon/kuke.yaml`. The existing default is what the loader already reads and what the existing tests target; switching it would be a behavior change unrelated to first-run dumping. Reviewer can ask for the path swap as a separate concern.
- **Cobra leaf flag-Changed pitfall.** Empirical probe confirmed `cmd.PersistentFlags().Changed` returns false on the leaf subcommand because parent persistent flags aren't on the leaf's set; `cmd.Flags().Changed` returns true after merge. Mentioning this because the existing kukeond `applyServerConfiguration`'s `flags.Changed` checks (cmd/kukeond/kukeond.go:133) may have the same latent issue (unrelated to this PR — flagging for a follow-up).
- **YAML emission via hand-written templated string** (not yaml.v3 `Node` with `HeadComment`). Hand-rolling keeps the field-level comments exactly readable in source, lets `kukeondImage: ""` survive omitempty, and is simple to round-trip-test.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` passes
- [x] golangci-lint passes (gosec G301/G302 → 0o750 dirs, 0o600 files; nonamedreturns; govet shadow)
- [x] **Smoke test per `CLAUDE.md` (touches the daemon)** — torn down `kuke-system/kukeon/kukeon/kukeond`, removed dumped configs, rebuilt `./kuke` and the local kukeond image, ran `sudo ./kuke init --kukeond-image docker.io/library/kukeon-local:dev`; both `/etc/kukeon/kukeond.yaml` and `/root/.kuke/kuke.yaml` were dumped with the expected content. Confirmed a second invocation does NOT update mtime (no overwrite).
- [x] **Daemon-parity check** — `sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` produced identical output.

Closes #204

### DIFF
--- a/cmd/kuke/clientconfig_test.go
+++ b/cmd/kuke/clientconfig_test.go
@@ -214,3 +214,54 @@ spec:
 		t.Errorf("LogLevel: got %q, want %q", got, "warn")
 	}
 }
+
+// TestMaybeWriteClientConfigurationDefaultSkipsCustomPath locks the guard
+// that prevents tests (and operators pointing --configuration at a custom
+// location) from writing to ~/.kuke/kuke.yaml on the host. The path-equality
+// guard is what makes operator-supplied custom paths "opt out" of the
+// default-location dump (per the issue's "Both honor --configuration to opt
+// out of the default location" note).
+func TestMaybeWriteClientConfigurationDefaultSkipsCustomPath(t *testing.T) {
+	t.Cleanup(viper.Reset)
+
+	viper.Reset()
+	cmd, err := NewKukeCmd()
+	if err != nil {
+		t.Fatalf("NewKukeCmd() error = %v", err)
+	}
+
+	dir := t.TempDir()
+	custom := filepath.Join(dir, "custom-kuke.yaml")
+	maybeWriteClientConfigurationDefault(cmd, custom)
+
+	if _, statErr := os.Stat(custom); !os.IsNotExist(statErr) {
+		t.Fatalf("custom path was written; got Stat err=%v, want IsNotExist", statErr)
+	}
+}
+
+// TestMaybeWriteClientConfigurationDefaultWritesAtDefaultPath proves the
+// happy path: when the resolved path equals the project default, the dump
+// fires and produces a parseable document. We swap $HOME to a TempDir so
+// DefaultClientConfigurationFile() resolves to a sandbox path and the
+// host's real ~/.kuke is untouched.
+func TestMaybeWriteClientConfigurationDefaultWritesAtDefaultPath(t *testing.T) {
+	t.Cleanup(viper.Reset)
+
+	t.Setenv("HOME", t.TempDir())
+	defaultPath := config.DefaultClientConfigurationFile()
+	if _, statErr := os.Stat(defaultPath); !os.IsNotExist(statErr) {
+		t.Fatalf("HOME shim is leaky; default path already exists: %v", statErr)
+	}
+
+	viper.Reset()
+	cmd, err := NewKukeCmd()
+	if err != nil {
+		t.Fatalf("NewKukeCmd() error = %v", err)
+	}
+
+	maybeWriteClientConfigurationDefault(cmd, defaultPath)
+
+	if _, statErr := os.Stat(defaultPath); statErr != nil {
+		t.Fatalf("default-path write did not fire: %v", statErr)
+	}
+}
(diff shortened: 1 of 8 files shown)


eminwux:
**PR #209 Review — `feat(config): dump default Server/ClientConfiguration on first run`**

Clean, well-scoped implementation of #204. The path-equality guard is a sound reading of AC #4 (the literal "skipped when --configuration is set" reading would defeat AC #1 once `kuke init` plumbs `--configuration` into the kukeond cell — the deviation is justified in the PR body and rests on `O_EXCL`, which is itself correct). `O_CREATE|O_EXCL|O_WRONLY` + `MkdirAll` closes the TOCTOU race, modes match the gosec bar (0o750 dirs / 0o600 files), Load round-trip is asserted, and the smoke test + daemon-parity check are documented. No correctness bugs, no sensitive data, no dead symbols (both `WriteDefault` exports are wired through their `maybeWrite*` helpers).

- **Unused param** `cmd/kuke/kuke.go:276`, `cmd/kukeond/kukeond.go:177`: `_ *cobra.Command` is never read — the path-equality strategy means `cmd` carries no information either helper needs. Drop it from the signatures.
- **Unchecked Close** `internal/clientconfig/clientconfig.go:117`, `internal/serverconfig/serverconfig.go:128`: `defer f.Close()` swallows any flush error after `WriteString`. For a ~1 KB config this is fine in practice, but a `Close()` checked into the named-return error would catch the rare ENOSPC/EIO mid-flush case.

LGTM.

PR #213, opened 2026-04-29
feat: add kuke image load for realm-scoped image import

eminwux:
## Summary
- Phase 1 of `kuke image load`: imports an OCI/docker tarball into the containerd namespace mapped to a realm via the existing daemon RPC plumbing, replacing the manual `docker save | sudo ctr -n <ns> images import -` step in the local smoke test.
- Adds the full vertical slice — `kukeonv1.LoadImageArgs/Reply/Result` over the unix-socket RPC, `controller.Exec.LoadImage` (validates realm + tarball, maps realm to namespace via `consts.RealmNamespace`), `runner.Exec.LoadImage`, and `ctr.client.LoadImage` (uses `containerd.WithSkipMissing()` to mirror `ctr images import` ergonomics for multi-arch tarballs from `docker save`).
- CLI accepts a positional tarball path, `-` for stdin (read from `cmd.InOrStdin()` so tests can inject without touching `os.Stdin`), or `--from-docker <ref>` (shells out to `docker save` on the CLI side, so the daemon host doesn't need docker installed); positional and `--from-docker` are mutually exclusive.
- New error sentinels `ErrLoadImage` and `ErrTarballRequired` in `internal/errdefs/`.

The CLAUDE.md smoke-section update will travel as a separate `docs:` follow-up issue per the project's hard rule against bundling CLAUDE.md edits into a code-fix PR.

## Test plan
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `make test` — all packages PASS
- [x] End-to-end smoke: tore down `kuke-system` cell, rebuilt `kuke` and the local docker image, then `sudo ./kuke image load --from-docker kukeon-local:dev --realm kuke-system --no-daemon` (in-process), `sudo ./kuke init` (re-bootstrap daemon), `sudo ./kuke image load --from-docker kukeon-local:dev --realm kuke-system` (over unix socket), `docker save busybox | sudo ./kuke image load - --realm default` (stdin + multi-arch via daemon).
- [x] Daemon-parity check: `sudo ./kuke get realms` and `sudo ./kuke get realms --no-daemon` produced identical output.

Closes #200

PR #225, opened 2026-04-29
docs: replace ctr images import with kuke image load in smoke test

### LINKED ISSUE #214: docs: replace ctr images import with kuke image load in smoke-test instructions
## Background

`kuke image load` shipped in #213 to wrap the realm → containerd-namespace mapping (`<realm>.kukeon.io`) and remove the need for operators to know the namespace name. The local smoke test in `CLAUDE.md` still teaches the bare-`ctr` invocation, which (a) leaks the namespace abstraction and (b) keeps the rebuild loop dependent on `sudo ctr` even though the equivalent kuke command now exists.

## Acceptance criteria

Replace the existing block in `CLAUDE.md` under **3. Build and load the local kukeond image (no registry push)** so it routes through `kuke image load`. Verbatim proposed wording:

````markdown
### 3. Build and load the local `kukeond` image (no registry push)

```bash
docker build --build-arg VERSION=v0.0.0-dev -t kukeon-local:dev .
docker tag kukeon-local:dev docker.io/library/kukeon-local:dev

sudo ./kuke image load --from-docker kukeon-local:dev --realm kuke-system --no-daemon

sudo ctr -n kuke-system.kukeon.io images ls | grep kukeon-local
```
````

Notes for the picker:
- `--no-daemon` is correct here because the daemon was just torn down in step 1; the in-process path is the only one available.
- Keep the `sudo ctr -n kuke-system.kukeon.io images ls | grep kukeon-local` verification line as-is — it inspects containerd directly, which is still the right end-of-step check.
- Do not bundle code changes into this PR. Pure docs.

eminwux:
## Summary
- Routes step 3 of the local smoke test through `kuke image load --from-docker` (added in #213) instead of piping `docker save` into `sudo ctr -n kuke-system.kukeon.io images import -`. Stops the smoke test from leaking the realm → containerd-namespace mapping that `kuke image load` is meant to abstract away.
- `--no-daemon` is correct here because step 1 just tore the daemon down; the in-process path is the only one available.
- Verification line (`sudo ctr -n kuke-system.kukeon.io images ls | grep kukeon-local`) is left as-is — it inspects containerd directly, which is still the right end-of-step check.
- Pure docs change. No code touched.

## Test plan
- [x] `make test` (full unit-test suite — what CI runs on PRs) passes locally.
- [ ] Local smoke test (the rebuild + `kuke init` flow) — not re-run for this PR; the change is a pure docs swap to the documented commands and the new line is the same `kuke image load` invocation already covered by #213's tests.

Closes #214

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -55,8 +55,7 @@ ln -sf kuke kukeond # kukeond is argv[0]-dispatched from the same binary
 docker build --build-arg VERSION=v0.0.0-dev -t kukeon-local:dev .
 docker tag kukeon-local:dev docker.io/library/kukeon-local:dev
 
-docker save kukeon-local:dev | \
-    sudo ctr -n kuke-system.kukeon.io images import -
+sudo ./kuke image load --from-docker kukeon-local:dev --realm kuke-system --no-daemon
 
 sudo ctr -n kuke-system.kukeon.io images ls | grep kukeon-local
 ```


eminwux:
**PR #225 Review — `docs: replace ctr images import with kuke image load in smoke test`**

Pure docs change — one-line swap from the raw `docker save | ctr -n kuke-system.kukeon.io images import -` pipeline to `kuke image load --from-docker kukeon-local:dev --realm kuke-system --no-daemon`. Verified the command shape against `cmd/kuke/image/load.go` (`--from-docker`, `--realm`, and the persistent root `--no-daemon` flag from `cmd/kuke/kuke.go:190` all line up), and the semantics match: `readFromDocker` shells out to `docker save <ref>` and the `kuke-system` realm maps to `kuke-system.kukeon.io`. No security or correctness concerns.

Leaving the verification line on raw `ctr -n kuke-system.kukeon.io images ls` is the right call for an end-of-step regression check — verifying via `kuke image get` would be circular against the same code path.

LGTM.

PR #227, opened 2026-04-29
fix: bind env vars in kukeond so KUKEON_/KUKEOND_ overrides win

### LINKED ISSUE #191: kukeond: env vars silently ignored — loadConfig has no BindEnv calls
## Repro
On `origin/main` (commit `9a43866`), inspect `cmd/kukeond/kukeond.go`:

```bash
grep -n "BindEnv" cmd/kukeond/kukeond.go
# (no matches)
```

`applyServerConfiguration` checks `envSet(...)` and skips writing the YAML value when the env var is set, but no `BindEnv()` call is ever made on any of the kukeond viper keys. So with e.g. a `ServerConfiguration` document containing `containerdSocket: /custom/sock` AND `KUKEON_CONTAINERD_SOCKET=/env/sock` exported:

1. `applyServerConfiguration` correctly observes `envSet(KUKEON_ROOT_CONTAINERD_SOCKET) == true` and skips writing the YAML value.
2. Downstream `viper.GetString(KUKEON_ROOT_CONTAINERD_SOCKET.ViperKey)` returns the flag default (`/run/containerd/containerd.sock`) because viper has no env binding for that key.

Net effect: env var AND YAML value are both silently ignored when both are set; the flag default wins.

## Observed
The advertised precedence in `applyServerConfiguration`'s doc comment is `--flag > env > ServerConfiguration > flag default`. In practice, env never wins — only `--flag` and YAML do, and only YAML wins when env is unset. Affected env vars: `KUKEON_HOST` (not used by kukeond directly, n/a), `KUKEON_CONTAINERD_SOCKET`, `KUKEON_ROOT_RUN_PATH`, `KUKEON_ROOT_LOG_LEVEL`, `KUKEOND_SOCKET`, `KUKEOND_SOCKET_GID`.

The matching bug on the kuke side was just fixed in PR #189 (`cmd/kuke/kuke.go`'s `loadConfig` was binding only two of four). The pr-reviewer flagged this kukeond gap as a follow-up in the PR #189 review (https://github.com/eminwux/kukeon/pull/189#issuecomment-4342905687).

## Root cause
`cmd/kukeond/kukeond.go` defines a `NewKukeondCmd` that does `viper.BindPFlag` for each persistent flag but never calls `<Var>.BindEnv()`. PR #185 introduced `applyServerConfiguration` with the env-skip check on the assumption env was bound, but the corresponding `BindEnv` calls were never added.

## Suggested fix
In `cmd/kukeond/kukeond.go`, before serving (e.g. in a small `bindEnvVars()` helper called from `NewKukeondCmd` after the BindPFlag calls, mirroring the pattern in `cmd/kuke/kuke.go`'s `loadConfig`), add:

```go
_ = config.KUKEON_ROOT_CONTAINERD_SOCKET.BindEnv()
_ = config.KUKEON_ROOT_RUN_PATH.BindEnv()
_ = config.KUKEON_ROOT_LOG_LEVEL.BindEnv()
_ = config.KUKEOND_SOCKET.BindEnv()
_ = config.KUKEOND_SOCKET_GID.BindEnv()
```

Add a regression test next to `TestApplyServerConfiguration*` that asserts `KUKEON_CONTAINERD_SOCKET` (and one other env var, e.g. `KUKEOND_SOCKET`) wins over a YAML value when both are set, paralleling `TestApplyClientConfigurationEnvOverridesConfig`.

eminwux:
## Summary
- `cmd/kukeond/kukeond.go` was wiring `BindPFlag` for each persistent flag but never calling `BindEnv()` for the matching viper keys. The bug: `applyServerConfiguration` correctly skipped its YAML write whenever an env var was set, but viper had no env binding to read from — so both env and YAML lost, and the flag default silently won. Affected vars: `KUKEON_CONTAINERD_SOCKET`, `KUKEON_RUN_PATH`, `KUKEON_LOG_LEVEL`, `KUKEOND_SOCKET`, `KUKEOND_SOCKET_GID`.
- Adds a `bindEnvVars()` helper called from `NewKukeondCmd` after the `BindPFlag` block, mirroring the `BindEnv` pattern in `cmd/kuke/kuke.go`'s `loadConfig` (the kuke side's parallel gap was fixed in #189).
- Strengthens `TestApplyServerConfigurationEnvOverridesConfig` to also assert `KUKEON_CONTAINERD_SOCKET` env wins over a YAML value, which is the regression guard called for in the issue. Removes the `bindKukeondEnv(t)` test helper since `NewKukeondCmd()` now handles binding for production and test paths alike — the helper had been masking the very bug this PR fixes.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (the CI target) — all packages pass, including `cmd/kukeond` with the new `KUKEON_CONTAINERD_SOCKET` regression assertion
- [x] Force-rebuilt `./kuke` from a clean tree to confirm the changed `NewKukeondCmd` builds in the kuke entry binary
- [ ] Daemon-parity smoke (`kuke get realms` vs `--no-daemon`) — not run. The change is confined to env-binding for daemon flag keys; it does not touch bind-mount, `/opt/kukeon` layout, or realm logic, so the bind-mount regression this smoke guards against cannot be triggered by this diff. Tearing down `kuke-system` on the dev host has real blast radius (interrupts the running daemon) and is left to the reviewer if desired.

Closes #191

### DIFF
--- a/cmd/kukeond/kukeond.go
+++ b/cmd/kukeond/kukeond.go
@@ -124,8 +124,27 @@ func NewKukeondCmd() (*cobra.Command, error) {
 		return nil, err
 	}
 
+	bindEnvVars()
+
 	cmd.AddCommand(newServeCmd())
 	return cmd, nil
+}
+
+// bindEnvVars wires each kukeond viper key to its `KUKEON_…` / `KUKEOND_…`
+// environment variable. Without these calls, applyServerConfiguration's
+// envSet check skips the YAML write but viper has no env binding to read
+// from — so both env and YAML are silently ignored and the flag default
+// wins. Mirrors the BindEnv block in cmd/kuke/kuke.go's loadConfig.
+func bindEnvVars() {
+	for _, v := range []config.Var{
+		config.KUKEON_ROOT_CONTAINERD_SOCKET,
+		config.KUKEON_ROOT_RUN_PATH,
+		config.KUKEON_ROOT_LOG_LEVEL,
+		config.KUKEOND_SOCKET,
+		config.KUKEOND_SOCKET_GID,
+	} {
+		_ = v.BindEnv()
+	}
 }
 
 // applyServerConfiguration layers the loaded ServerConfiguration on top of
--- a/cmd/kukeond/kukeond_test.go
+++ b/cmd/kukeond/kukeond_test.go
@@ -26,21 +26,6 @@ import (
 	"github.com/spf13/viper"
 )
 
-func bindKukeondEnv(t *testing.T) {
-	t.Helper()
-	for _, v := range []config.Var{
-		config.KUKEOND_SOCKET,
-		config.KUKEOND_SOCKET_GID,
-		config.KUKEON_ROOT_RUN_PATH,
-		config.KUKEON_ROOT_CONTAINERD_SOCKET,
-		config.KUKEON_ROOT_LOG_LEVEL,
-	} {
-		if err := v.BindEnv(); err != nil {
-			t.Fatalf("BindEnv %s: %v", v.EnvVar(), err)
-		}
-	}
-}
-
 func TestNewKukeondCmdHasConfigurationFlag(t *testing.T) {
 	t.Cleanup(viper.Reset)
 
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #227 Review — `fix: bind env vars in kukeond so KUKEON_/KUKEOND_ overrides win`**

Targeted fix for the precedence bug in #191. Adding `bindEnvVars()` to `NewKukeondCmd` plugs the gap between `applyServerConfiguration`'s `envSet` skip and viper's missing env binding; the strengthened `TestApplyServerConfigurationEnvOverridesConfig` is a real regression guard (asserting `KUKEON_CONTAINERD_SOCKET` env wins over YAML) and was previously masked by the test-only `bindKukeondEnv` helper, which is correctly removed. The 5 viper keys in `bindEnvVars` match the 5 `BindPFlag` keys in `NewKukeondCmd`, and the comment's "mirrors `cmd/kuke/kuke.go`'s loadConfig BindEnv block" claim holds. No correctness, leak, or dead-code concerns. Skipping the daemon-parity smoke is justified — the diff is env-binding only and does not touch `/opt/kukeon` reads or bind-mount layout.

LGTM.

PR #228, opened 2026-04-29
fix: route kuke get output format through viper so KUKE_GET_OUTPUT env wins

### LINKED ISSUE #190: kuke get: KUKEON_GET_OUTPUT env var silently ignored — ParseOutputFormat bypasses viper
## Repro

Reproduced on `main` at 9a43866 while implementing #160.

```bash
make kuke
sudo ./kuke init --kukeond-image docker.io/library/kukeon-local:dev   # any working init

# CLI flag works:
sudo ./kuke get realms --output yaml      # prints YAML

# Env var silently ignored:
sudo KUKEON_GET_OUTPUT=yaml ./kuke get realms   # still prints table
```

Same shape from a config file: a YAML doc setting `kuke/get/output` via viper does not affect output format either. Only the literal `--output` (or `-o`) flag works.

## Observed

`kuke get realms` prints the table format regardless of the env var or any non-flag viper source.

## Root cause

`cmd/kuke/get/shared/shared.go:54` reads the flag directly:

```go
func ParseOutputFormat(cmd *cobra.Command) (OutputFormat, error) {
    output, err := cmd.Flags().GetString("output")
    ...
```

But every `kuke get <kind>` subcommand binds the flag through viper at registration time, e.g. `cmd/kuke/get/realm/realm.go:91`:

```go
_ = viper.BindPFlag(config.KUKE_GET_OUTPUT.ViperKey, cmd.Flags().Lookup("output"))
_ = viper.BindPFlag(config.KUKE_GET_OUTPUT.ViperKey, cmd.Flags().Lookup("o"))
```

The `KUKE_GET_OUTPUT` env var (`cmd/config/env.go:267`) is wired into viper via `kuke/get/output` — so the *intent* is clearly that env-var (and viper-backed config) overrides should work, matching every other `KUKEON_*` env var on the kuke root. But because `ParseOutputFormat` reads the cobra flag directly, the env-via-viper path never reaches the printer.

Affects all six get subcommands (`get realm`, `get space`, `get stack`, `get cell`, `get container`, and the parent `get`).

## Suggested fix

Have `ParseOutputFormat` read via viper instead of the flag set:

```go
output := viper.GetString(config.KUKE_GET_OUTPUT.ViperKey)
```

This already gives `--flag > env > config > default` precedence for free since `BindPFlag` registers the flag as the highest-priority source. The fall-back to `cmd.Flags().GetString("o")` becomes unnecessary because the existing `BindPFlag` for `-o` already feeds the same viper key.

Once this is fixed, `ClientConfiguration.spec.output` becomes wireable in a follow-up to #160 — it was deliberately trimmed from PR #189 because shipping it before this fix would have been a silent no-op.

eminwux:
## Summary
- \`ParseOutputFormat\` now reads via viper for the env/config fall-through, while still consulting the active subcommand's \`--output\` flag directly so the flag works for every \`kuke get <kind>\`. Reading viper alone does not work because each \`get\` sibling binds the same \`kuke/get/output\` key during construction and viper retains only the last binding (\`v.pflags[key] = flag\`), so \`viper.GetString\` would only see the last-bound subcommand's flag.
- \`loadConfig\` now calls \`config.KUKE_GET_OUTPUT.BindEnv()\` so viper actually consults the env variable. Without this, the suggested fix would still be a silent no-op for env vars.
- Issue #190's repro snippet writes \`KUKEON_GET_OUTPUT=yaml\` but the var is named \`KUKE_GET_OUTPUT\` (per \`cmd/config/env.go:269\`, sibling to \`KUKE_GET_REALM_NAME\`, etc.). The fix wires the actual var, not the typo.
- Precedence: explicit \`--output\`/\`-o\` flag > \`KUKE_GET_OUTPUT\` env > table default. \`ClientConfiguration.spec.output\` plumbing remains the follow-up to #160 the issue mentions.

## Test plan
- [x] \`go build ./...\` clean
- [x] \`go vet ./...\` clean
- [x] \`make test\` passes (all unit tests including the new \`TestParseOutputFormatEnvAndFlagPrecedence\`)
- [x] Built \`./kuke\` and verified env path: \`KUKE_GET_OUTPUT=yaml ./kuke get realms --no-daemon\` prints YAML; \`KUKE_GET_OUTPUT=yaml ./kuke get realms --no-daemon --output table\` prints table (flag overrides env); plain \`./kuke get realms --no-daemon\` prints table (default).
- [x] Daemon-parity sanity: \`sudo ./kuke get realms\` and \`sudo ./kuke get realms --no-daemon\` produce identical table output; \`KUKE_GET_OUTPUT=yaml sudo -E ./kuke get realms\` prints YAML through the daemon.
- [ ] Full smoke test (rebuild kukeond image + \`kuke init\`) — skipped: this fix is purely client-side output parsing, the daemon read path is untouched.

Closes #190

### DIFF
--- a/cmd/kuke/get/shared/shared.go
+++ b/cmd/kuke/get/shared/shared.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/eminwux/kukeon/cmd/config"
 	createshared "github.com/eminwux/kukeon/cmd/kuke/create/shared"
 	"github.com/eminwux/kukeon/internal/controller"
 	"github.com/spf13/cobra"
@@ -49,17 +50,25 @@ const (
 	OutputFormatTable OutputFormat = "table"
 )
 
-// ParseOutputFormat parses and validates the --output flag from the command.
+// ParseOutputFormat resolves the output format using kuke's standard
+// precedence: explicit `--output`/`-o` flag > KUKEON_GET_OUTPUT env (and
+// any other source viper has bound to `kuke/get/output`) > table default.
+// Each `kuke get <kind>` subcommand binds its own `--output` flag to the
+// shared viper key at construction time, but viper only retains the last
+// binding for a given key; reading the active command's flag directly
+// keeps the flag working for every subcommand and lets viper handle the
+// env/config fall-throughs.
 func ParseOutputFormat(cmd *cobra.Command) (OutputFormat, error) {
-	output, err := cmd.Flags().GetString("output")
-	if err != nil {
-		return OutputFormatTable, err
+	output := ""
+	if cmd != nil && cmd.Flags().Changed("output") {
+		var err error
+		output, err = cmd.Flags().GetString("output")
+		if err != nil {
+			return OutputFormatTable, err
+		}
 	}
 	if output == "" {
-		output, _ = cmd.Flags().GetString("o")
-	}
-	if output == "" {
-		return OutputFormatTable, nil
+		output = viper.GetString(config.KUKE_GET_OUTPUT.ViperKey)
 	}
 
 	trimmed := strings.TrimSpace(output)
--- a/cmd/kuke/get/shared/shared_test.go
+++ b/cmd/kuke/get/shared/shared_test.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/eminwux/kukeon/cmd/config"
 	"github.com/eminwux/kukeon/cmd/kuke/get/shared"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #228 Review — `fix: route kuke get output format through viper so KUKE_GET_OUTPUT env wins`**

Correctly fixes the env-precedence bug from #190. The `Changed("output")` short-circuit cleanly works around viper's last-binding-wins behavior across sibling `kuke get` subcommands, and the daemon read path is genuinely untouched so the smoke-test skip is justified. Test coverage covers env-only, flag-only, flag-over-env, and default cases. No sensitive data, no dead symbols.

A small nit:

- **Nit** `cmd/kuke/get/shared/shared_test.go:115`: `t.Fatalf("bind KUKEON_GET_OUTPUT: %v", err)` names the wrong env var — the actual var is `KUKE_GET_OUTPUT` (per `cmd/config/env.go:269` and the PR description). Drop the trailing `N`. Cosmetic only — fires only on a never-failing `BindEnv`.

LGTM.

PR #231, opened 2026-04-30
fix: honor kukeond --flag over YAML when dispatched on leaf subcommand

### LINKED ISSUE #210: kukeond: --flag silently overridden by ServerConfiguration YAML; cmd.PersistentFlags().Changed returns false on leaf subcommand
## Repro

Surfaced while implementing #204. Empirical cobra probe with `cobra v1.10.2 / pflag v1.0.9`, equivalent to how `cmd/kukeond/kukeond.go` is structured (root defines persistent flags, single `serve` leaf):

```go
root := &cobra.Command{Use: "root", PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
    fmt.Println("cmd.Use:", cmd.Use)
    fmt.Println("cmd.PersistentFlags().Changed:", cmd.PersistentFlags().Changed("foo"))
    fmt.Println("cmd.Flags().Changed:", cmd.Flags().Changed("foo"))
    fmt.Println("cmd.Flag(foo).Changed:", cmd.Flag("foo").Changed)
    fmt.Println("cmd.InheritedFlags().Changed:", cmd.InheritedFlags().Changed("foo"))
    return nil
}}
root.PersistentFlags().String("foo", "default", "foo")
root.AddCommand(&cobra.Command{Use: "leaf", RunE: func(cmd *cobra.Command, args []string) error { return nil }})
root.SetArgs([]string{"leaf", "--foo", "bar"})
root.Execute()
```

Output:

```
cmd.Use: leaf
cmd.PersistentFlags().Changed: false
cmd.Flags().Changed: true
cmd.Flag(foo).Changed: true
cmd.InheritedFlags().Changed: true
```

`PersistentPreRunE` is invoked with `cmd = leaf`; the leaf's `PersistentFlags()` set does not include parent-defined persistent flags, so `.Changed("foo")` lookups miss and return `false`. Only the merged `Flags()` (post-`mergePersistentFlags`), `Flag()` (climbs the tree), and `InheritedFlags()` see the parent flag's `.Changed` state.

## Observed

`cmd/kukeond/kukeond.go:132-150` `applyServerConfiguration` reads `flags := cmd.PersistentFlags()` and gates each `viper.Set(...)` on `!flags.Changed(\"<name>\")`. In production, `applyServerConfiguration` is called from the kukeond root's `PersistentPreRunE` with `cmd = serveCmd` (the leaf). On that leaf, `cmd.PersistentFlags().Changed(\"socket\")` returns `false` even when the operator passed `--socket /foo` — because the `socket` persistent flag is defined on the root, not on `serveCmd`. The flag-changed guard never fires, so the YAML value at `spec.Socket` silently overrides the operator's `--socket` flag.

Same bug shape for `socket-gid`, `run-path`, `containerd-socket`, `log-level`. The advertised precedence in the function's doc comment is `--flag > env > ServerConfiguration > default`, but in practice it is `env > ServerConfiguration > --flag > default` for kukeond.

This is distinct from #191 (env vars silently ignored — missing `BindEnv` calls); both bugs interact, but #191 only explains the env tier, not why CLI `--flag` loses to YAML.

The existing test `TestApplyServerConfigurationFlagOverridesConfig` (cmd/kukeond/kukeond_test.go:89) does not catch this: it builds the root cmd via `NewKukeondCmd()` and passes that root cmd directly into `applyServerConfiguration`. Production passes the leaf. The test exercises a code path the production caller never sees.

## Root cause

cobra's `PersistentPreRunE` runs on the leaf subcommand. `(*cobra.Command).PersistentFlags()` returns only the flag set local to *this* command — it does not include inherited persistent flags from parents until `mergePersistentFlags` runs (which populates `(*cobra.Command).Flags()`). The kukeond code reads the leaf's local persistent flags, finds no entry for `socket`/etc., and the `Changed` lookup falls through to `false` (pflag's `Changed` returns `false` on a missing flag).

## Suggested fix

In `cmd/kukeond/kukeond.go:133`, swap the local `cmd.PersistentFlags()` for `cmd.Flags()` (post-merge, includes inherited persistent flags):

```go
flags := cmd.Flags()
```

Or use `cmd.Flag(name).Changed` per-field, which climbs the parent chain and is null-safe via a helper, mirroring the `flagChanged` helper already in `cmd/kuke/kuke.go:291-299`.

A regression test should drive a real `cmd.SetArgs([]string{\"serve\", \"--socket\", \"/foo\"})` + `cmd.Execute()` round-trip — i.e. exercise PersistentPreRunE through cobra's dispatch, not by calling `applyServerConfiguration` directly with the root cmd. That mirrors production and would have caught this; the existing test does not.

The matching code path on the kuke client side (`cmd/kuke/kuke.go`'s `applyClientConfiguration`) already uses the `flagChanged` helper which checks both `cmd.Flags()` and `cmd.PersistentFlags()`, so the kuke side is unaffected.

eminwux:
## Summary

- cobra invokes `PersistentPreRunE` on the **leaf** subcommand, but kukeond's persistent flags (`socket`, `socket-gid`, `run-path`, `containerd-socket`, `log-level`) are defined on the root. `applyServerConfiguration` was reading `cmd.PersistentFlags().Changed(name)` on the leaf — that lookup misses parent-defined persistent flags and silently returns `false`, so the YAML value at `spec.<field>` overrode the operator's explicit `--flag`. The advertised precedence (`--flag > env > ServerConfiguration > default`) was effectively `env > ServerConfiguration > --flag > default` for kukeond.
- Switched the flag-changed check to a `flagChanged(cmd, name)` helper that looks up both `cmd.Flags()` (post-merge — what cobra populates on the leaf before `PersistentPreRunE`) and `cmd.PersistentFlags()` (the local set). Mirrors the same-named helper in `cmd/kuke/kuke.go`. Could have inlined per-field, but five callsites in one function plus parity with the kuke-side fix made a helper the lower-blast-radius call.
- Added regression test `TestApplyServerConfigurationFlagOverridesConfigViaExecute` that drives `cmd.SetArgs([]string{"serve", "--configuration", <tmp>, "--socket", ..., ...}) + cmd.Execute()` — i.e. exercises `PersistentPreRunE` through cobra dispatch with `cmd = serve` (leaf), the production code path. Verified the test fails on the unfixed `applyServerConfiguration` for all five spec fields (each silently overridden by YAML).
- Existing `TestApplyServerConfigurationFlagOverridesConfig` left in place — it tests `applyServerConfiguration`'s narrow contract (skip when Changed is set on the persistent flag set directly), which is still valuable as a unit test even though it bypasses cobra dispatch. The new test is the production-path coverage.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` passes
- [x] Counter-check: stashed only the `kukeond.go` fix and re-ran the new test — it failed with the exact bug from #210 (each of `socket`, `socket-gid`, `run-path`, `containerd-socket`, `log-level` silently overridden by YAML).
- [x] Smoke test (rebuild kuke + kukeond image, `kuke init` with `--kukeond-image docker.io/library/kukeon-local:dev`): kukeond comes up at `unix:///run/kukeon/kukeond.sock`, daemon process line shows `--socket /run/kukeon/kukeond.sock --run-path /opt/kukeon --containerd-socket /run/containerd/containerd.sock --socket-gid 986`, daemon-parity check `sudo ./kuke get realms` vs `--no-daemon` returns identical output.

Closes #210

### DIFF
--- a/cmd/kukeond/kukeond.go
+++ b/cmd/kukeond/kukeond.go
@@ -155,23 +155,44 @@ func bindEnvVars() {
 // without it, `viper.Set` would override viper's env binding and silently
 // invert env > YAML.
 func applyServerConfiguration(cmd *cobra.Command, spec v1beta1.ServerConfigurationSpec) {
-	flags := cmd.PersistentFlags()
-	if spec.Socket != "" && !flags.Changed("socket") && !envSet(config.KUKEOND_SOCKET) {
+	if spec.Socket != "" && !flagChanged(cmd, "socket") && !envSet(config.KUKEOND_SOCKET) {
 		viper.Set(config.KUKEOND_SOCKET.ViperKey, spec.Socket)
 	}
-	if spec.SocketGID != 0 && !flags.Changed("socket-gid") && !envSet(config.KUKEOND_SOCKET_GID) {
+	if spec.SocketGID != 0 && !flagChanged(cmd, "socket-gid") && !envSet(config.KUKEOND_SOCKET_GID) {
 		viper.Set(config.KUKEOND_SOCKET_GID.ViperKey, spec.SocketGID)
 	}
-	if spec.RunPath != "" && !flags.Changed("run-path") && !envSet(config.KUKEON_ROOT_RUN_PATH) {
+	if spec.RunPath != "" && !flagChanged(cmd, "run-path") && !envSet(config.KUKEON_ROOT_RUN_PATH) {
 		viper.Set(config.KUKEON_ROOT_RUN_PATH.ViperKey, spec.RunPath)
 	}
-	if spec.ContainerdSocket != "" && !flags.Changed("containerd-socket") &&
+	if spec.ContainerdSocket != "" && !flagChanged(cmd, "containerd-socket") &&
 		!envSet(config.KUKEON_ROOT_CONTAINERD_SOCKET) {
 		viper.Set(config.KUKEON_ROOT_CONTAINERD_SOCKET.ViperKey, spec.ContainerdSocket)
 	}
-	if spec.LogLevel != "" && !flags.Changed("log-level") && !envSet(config.KUKEON_ROOT_LOG_LEVEL) {
+	if spec.LogLevel != "" && !flagChanged(cmd, "log-level") && !envSet(config.KUKEON_ROOT_LOG_LEVEL) {
 		viper.Set(config.KUKEON_ROOT_LOG_LEVEL.ViperKey, spec.LogLevel)
 	}
+}
+
+// flagChanged reports whether `--<name>` was explicitly set on the command
+// line. cobra invokes PersistentPreRunE on the leaf subcommand, but kukeond's
+// persistent flags (socket, run-path, etc.) are defined on the root. On the
+// leaf, cmd.PersistentFlags() returns only the leaf's local persistent flag
+// set — it does not include parents' persistent flags — so a direct
+// `cmd.PersistentFlags().Changed("socket")` lookup misses and returns false
+// even when the operator passed `--socket`. cobra merges parent persistent
+// flags into the leaf's cmd.Flags() before PersistentPreRunE runs, so the
+// merged set is the production-correct read; the local PersistentFlags
+// fallback keeps the helper correct in unit tests that pass the root cmd
+// directly without going through Execute. Mirrors the same-named helper in
+// cmd/kuke/kuke.go.
+func flagChanged(cmd *cobra.Command, name string) bool {
+	if f := cmd.Flags().Lookup(name); f != nil && f.Changed {
+		return true
+	}
+	if f := cmd.PersistentFlags().Lookup(name); f != nil && f.Changed {
+		return true
+	}
+	return false
 }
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #231 Review — `fix: honor kukeond --flag over YAML when dispatched on leaf subcommand`**

Tight, well-scoped fix for the precedence bug from #210. The new `flagChanged(cmd, name)` helper mirrors the one in `cmd/kuke/kuke.go:315-323` and is correct for both production (cobra has merged parent persistent flags into `cmd.Flags()` before `PersistentPreRunE` fires on the leaf) and the existing unit test (root cmd, persistent flags not yet merged). Regression test drives `cmd.Execute()` with `cmd = serve` leaf — the actual production dispatch path — so the unit-level `TestApplyServerConfigurationFlagOverridesConfig` and the new `…ViaExecute` cover complementary contracts. Smoke test in the test plan covers daemon-parity per `CLAUDE.md`. No security or sensitive-data concerns — no logging, error messages, or external IO touched.

LGTM.

PR #256, opened 2026-04-30
feat: tail non-Attachable container stdout/stderr via cio.LogFile

### LINKED ISSUE #203: cli: kuke log read container stdout/stderr via containerd socket
## Context
`kuke log` (#183) tails the per-container sbsh capture file — works only for Attachable containers. Non-Attachable containers (including `kukeond` itself, which deliberately does not use sbsh) have no kuke-native way to read stdout/stderr; operators drop to `ctr -n <ns> tasks attach` or dig through the containerd shim's log path.

`kukeond` makes this concrete today: it runs as a container in `kuke-system / kukeon / kukeon / kukeond` but its logs are unreachable through `kuke log` — the daemon doesn't need sbsh and shouldn't have to use it just to be observable.

This is gap #4 in `docs/gaps-2026-04-19.md`.

## Proposal
Extend `kuke log` to read stdout/stderr directly from the containerd socket for any container, dropping the sbsh-capture requirement.

- For non-Attachable containers (including `kukeond`): stream stdout/stderr from the containerd task via the containerd socket. No daemon-side log persistence; no kuke-owned log file.
- For Attachable containers: keep current sbsh-capture path as-is (regression-free).
- `--follow` works in both cases.
- Daemon-parity: works with and without `--no-daemon`.

## Acceptance criteria
- [ ] `kuke log` on a non-Attachable container streams its stdout/stderr via the containerd socket.
- [ ] `kuke log` on the running `kukeond` cell shows daemon stdout/stderr.
- [ ] `--follow` works for both Attachable and non-Attachable containers.
- [ ] Existing sbsh-capture path for Attachable containers is unchanged (regression).
- [ ] Daemon and `--no-daemon` paths produce the same output.

## Notes
- Design is fixed: read via the containerd socket. No daemon-buffered ring, no kuke-owned log file — containerd already has the bytes; we just expose them.
- Related: #165 (kuketty replaces sbsh dependency), #161 (background reconciliation loop).

eminwux:
## Summary
- `kuke log` now works for both Attachable and non-Attachable containers. Attachable still uses sbsh's capture file (unchanged); non-Attachable uses a host file populated by the containerd shim via `cio.LogFile`.
- Daemon and `--no-daemon` paths produce the same stream via the same `LogContainer` RPC; the response now carries either `HostCapturePath` (Attachable) or `HostLogPath` (non-Attachable), and the CLI tails whichever is set.
- `kukeond` (and any non-root non-Attachable container) is now started with `cio.LogFile(<container metadata dir>/log)` instead of `cio.NullIO`, so stdout/stderr are captured to disk by the shim — no kuke-owned ring, no daemon-side buffering. Root containers stay on `NullIO`.
- Per-container picker for `kuke log` no longer requires `Attachable=true`; it picks the lone non-root container regardless of IO mode.

Implementation note: chose `cio.LogFile` over `Task.Attach` because `ctr tasks attach` kills the task on detach (per repo memory) — the shim-owned log file matches the issue's "containerd already has the bytes; we just expose them" framing without that footgun.

## Test plan
- [x] `make test` (full unit suite) — all packages pass
- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] `make dev-init` — full re-bootstrap clean; daemon-parity check shows identical output for `kuke get realms` and `kuke get realms --no-daemon`
- [x] `sudo ./kuke log --realm kuke-system --space kukeon --stack kukeon --cell kukeond --no-follow` returns the actual `kukeond` daemon stdout (`time=… level=INFO msg="kukeond listening"…`)

Closes #203

### DIFF
--- a/cmd/kuke/log/log.go
+++ b/cmd/kuke/log/log.go
@@ -16,10 +16,18 @@
 
 // Package log implements the `kuke log` subcommand. It mirrors `kuke
 // attach`'s flag/arg shape but instead of opening an interactive sbsh
-// session it tails the per-container sbsh capture file. Bytes never
-// traverse the daemon RPC: the daemon validates the Attachable gate and
-// returns the host path of the capture file; this subcommand opens that
-// path directly.
+// session it tails the per-container output stream. Bytes never traverse
+// the daemon RPC: the daemon resolves the host path of the relevant
+// stream and this subcommand opens that path directly.
+//
+// Two paths exist depending on the target container's IO model:
+//
+//   - Attachable containers route output through sbsh, which writes a
+//     tty byte stream to HostCapturePath. `kuke log` tails that file.
+//   - Non-Attachable containers (including kukeond) have the containerd
+//     runtime shim write stdout/stderr to HostLogPath via cio.LogFile.
+//     `kuke log` tails that file. This is gap #4 in
+//     docs/gaps-2026-04-19.md and was implemented per issue #203.
 package log
 
 import (
@@ -65,7 +73,7 @@ func NewLogCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:           "log",
 		Aliases:       []string{"logs"},
-		Short:         "Tail the sbsh capture file of an Attachable container",
+		Short:         "Tail a container's stdout/stderr stream",
 		Args:          cobra.NoArgs,
 		SilenceUsage:  true,
 		SilenceErrors: false,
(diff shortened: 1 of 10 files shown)


eminwux:
**PR #256 Review — `feat: tail non-Attachable container stdout/stderr via cio.LogFile`**

Clean, well-scoped implementation of the issue-#203 contract: Attachable keeps its sbsh capture path, non-Attachable gets a shim-owned `cio.LogFile` path, and the wire result discriminates via two mutually-exclusive fields. New code paths are covered: client-side picker drops the Attachable filter (`TestLog_AmbiguousAcceptsNonAttachable`, `TestLog_AutoPicksLoneNonAttachable`), daemon RPC passes both `HostCapturePath` and `HostLogPath` through (`TestLogContainer_NonAttachable_RPCReturnsLogPath`), and the IO selection in `StartContainer` is mutually exclusive by switch ordering (`LogFilePath` short-circuits before `Terminal`). Test plan includes the `make dev-init` daemon-parity regression guard. No sensitive data leaks. `errdefs.ErrAttachNotSupported` correctly retained for the `AttachContainer` path; `resolveAttachable` still alive as `AttachContainer`'s helper.

- **Comment drift** `internal/controller/runner/start.go:42-49`: `containerLogTaskSpec`'s comment says "Centralised here so all three StartContainer call sites pick up the same policy" but only 2 of the 3 `r.ctrClient.StartContainer` sites route through the helper — site #1 at `internal/controller/runner/start.go:349` (root container start) still hardcodes `ctr.TaskSpec{}`. Behaviorally correct today (the helper short-circuits for `spec.Root`), but the "centralised" claim becomes a future-maintenance trap. Either tighten to "non-root" or route site #1 through the helper for symmetry.
- **Error message wording** `cmd/kuke/log/log.go` (the `os.ErrNotExist` branch in `runLog`): the unified error message attributes the missing file to "the runtime shim has not opened it yet" for both branches. For the Attachable branch (`HostCapturePath`), the writer is sbsh, not the shim. Minor — most users won't notice — but the wording is technically wrong half the time.

LGTM unconditionally — both items are non-blocking polish, not bugs.

PR #259, opened 2026-05-01
fix: chown attachable tty dir to kukeon group + container uid (#258)

### LINKED ISSUE #258: attachable: kuke attach broken for kukeon-group users and for non-root container USER (0700 root-owned tty dir)
The per-container tty directory created at `internal/controller/runner/attachable.go:48` is mode `0o700` root-owned:

```go
// 0700 so only the daemon (root) can reach the socket from the host
// side. Documented in the issue's "Socket security threat model".
if err := os.MkdirAll(ttyDir, 0o700); err != nil {
    return nil, err
}
```

That mode breaks `kuke attach` in two real use cases that both surface as a missing/unreachable socket. The dir is bind-mounted into the container at `/run/kukeon/tty` (`internal/ctr/attachable.go:113-119`), so the same inode/perm controls both host-side attach and in-container sbsh writes.

## Repro A — unprivileged member of `kukeon` group cannot attach (host side)

Reproduced on `main` at ce18f1c.

`kuke init` deliberately makes `/opt/kukeon` group-traversable for the `kukeon` group (`cmd/kuke/init/init.go:46-60`):

```go
kukeonRunPathDirMode = os.ModeSetgid | 0o750   // 02750 root:kukeon
kukeonSocketMode     = 0o660
```

User `inwx` is a member of `kukeon`:

```
$ id
uid=1000(inwx) gid=1000(inwx) groups=…,986(kukeon),…
```

But the per-cell tty dir is locked to root only:

```
$ sudo ls -la /opt/kukeon/default/default/default/exa-70e385/work/tty/
drwx--S--- 2 root kukeon …    ← 0700 + inherited setgid; group has no x
```

Result on attach as `inwx`:

```
$ kuke run -a -p exa
…
Error: error attaching client: ping failed: dial unix
  /opt/kukeon/default/default/default/exa-70e385/work/tty/socket:
  connect: permission denied
```

The `permission denied` is at path traversal — `inwx` can't enter the tty dir even though it's in `kukeon`. This contradicts the `kuke init` group-access design and makes the `kukeon` group effectively useless for the attach path.

## Repro B — wrapped container with a non-root image USER never creates a socket

Reproduced on `main` at ce18f1c with `registry.eminwux.com/claude:latest` (Debian-based image whose Dockerfile ends with `USER claude`):

```Dockerfile
FROM debian:bookworm-slim
…
RUN useradd --create-home --shell /bin/bash claude
USER claude
CMD ["sleep", "infinity"]
```

Profile sets the image as the work container without overriding `user:`, so `oci.WithImageConfig` (`internal/ctr/container.go:323`) honors `USER claude` (uid 1000).

```
$ docker run --rm registry.eminwux.com/claude:latest id
uid=1000(claude) gid=1000(claude) groups=1000(claude)
```

Attach (as root, so repro A is excluded) yields:

```
$ sudo kuke run -a -p exa
…
Error: error attaching client: ping failed: dial unix
  /opt/kukeon/default/default/default/exa-890eb1/work/tty/socket:
  connect: no such file or directory
```

No socket inode was ever created. Reproduced standalone with a docker run that mirrors the bind mount and uid:

```
$ sudo install -d -m 0700 -o root -g root /tmp/ttytest
$ docker run --rm --user 1000 \
    -v /tmp/ttytest:/run/kukeon/tty \
    -v /home/inwx/projects/sbsh/sbsh:/.kukeon/bin/sbsh:ro \
    registry.eminwux.com/claude:latest /bin/bash -c \
    '/.kukeon/bin/sbsh terminal --run-path /run/kukeon/tty \
       --socket /run/kukeon/tty/socket --capture-file /run/kukeon/tty/cap \
       --log-file /run/kukeon/tty/log -- /bin/echo hi; echo exit=$?'
Failed to open log file: open /run/kukeon/tty/log: permission denied
Error: open /run/kukeon/tty/log: permission denied
exit=1
```

sbsh dies opening its log file *before* it ever calls `Listen()`, so no socket inode is created. The failure is silent because Attachable containers route stdout/stderr through sbsh's capture file (`internal/util/fs/metadata.go:127-138`) — and sbsh dies before opening that too, so `kuke logs` has nothing to surface either.

## Root cause

`internal/controller/runner/attachable.go:48` uses mode `0o700` and leaves the dir root-owned. That choice is incompatible with both:

- the rest of the runtime tree, which is `0o2750 root:kukeon` (set up in `kuke init` precisely so non-root operators in `kukeon` can use the daemon path), and
- any wrapped container that runs as a non-root image USER, since the bind-mounted dir keeps host ownership inside the container's user namespace.

## Suggested fix

Make the tty dir owner/group/mode reflect both consumers:

1. Create the dir as `0o2750`, `chown` to `root:KukeondSocketGID` (the GID is already plumbed through `internal/controller/controller.go:84-87`). Resolves repro A and matches the rest of the tree's group-access design. The existing comment about "only the daemon (root) can reach the socket from the host side" no longer holds as a justification — `/run/kukeon/kukeond.sock` is already 0660 root:kukeon, and the daemon socket is the trust anchor.

2. For repro B, also `chown` the tty dir to the runtime uid the container will execute as. The uid is resolvable from the image's OCI config at the same point the runner already reads the image (`oci.WithImageConfig` in `internal/ctr/container.go:323`); plumb it back to `attachableBuildOpts` so the chown happens before the container starts. The owner-write bit (`0o2750`) is then sufficient for sbsh to create its socket/log/capture files; the `kukeon` group still gets traverse + read for the host-side attach.

3. Independently, surface sbsh startup failures: when sbsh exits non-zero before `Listen()`, the attach path today blocks on a socket that never appears. Even with this fix, future variants of repro B (read-only rootfs, bind-mount option mismatch, etc.) will recur and need a clearer error than `connect: no such file or directory`. Out of scope for this issue but worth a follow-up.

## Workarounds

For users hitting B today: drop `USER` from the Dockerfile, or set `user: "0"` on the affected container in the CellProfile (`securitySpecOpts` at `internal/ctr/spec.go:309-311` passes it through `oci.WithUser` and overrides the image's USER).

For users hitting A: `sudo kuke attach`. There is no non-root workaround.

eminwux:
## Summary
- Fix #258: per-container Attachable tty dir was created at 0700 root-owned, blocking both kukeon-group operators (Repro A) and non-root container USER (Repro B) from reaching the host-side sbsh socket.
- Pre-create the tty dir as 02750 root:kukeon when the kukeon group GID is plumbed (matches `kuke init`'s /opt/kukeon layout); fall back to 0700 when it isn't (legacy / `--no-daemon` smoke tests).
- After CreateContainerFromSpec resolves the image's USER directive, chown the tty dir to the container's runtime uid (read from `container.Spec(ctx).Process.User.UID`) so non-root processes can create the socket/log/capture siblings inside the bind-mounted dir.
- Plumb `KukeondSocketGID` from controller through `runner.Options.KukeonGroupGID`, and forward it from `kukeond serve` (this last step was missing — without it the daemon-mediated path silently regressed).

## Test plan
- [x] `make dev-init` — daemon parity check shows identical realm output for both `kuke get realms` and `kuke get realms --no-daemon`.
- [x] Manual Repro A (root-USER attachable busybox via daemon `apply`): tty dir is `drwxr-s--- root:kukeon` mode 02750; socket/log/capture all inherit kukeon group via setgid.
- [x] Manual Repro B (non-root `user: "1234:1234"` attachable busybox): tty dir is `drwxr-s--- 1234:kukeon`; sbsh log inside is owned `1234:kukeon`, proving the non-root in-container process can write the bind-mounted dir.
- [x] New unit tests in `internal/controller/runner/attachable_test.go`: `TestAttachableTTYDirInitialPerms` covers gid=986 / gid=0 / gid=-1; `TestAttachableTTYDirRootMode_SetsSGIDBit` guards the FileMode constant against losing `os.ModeSetgid` on refactor.
- [ ] `make test-e2e` — pre-existing failures unrelated to this change: `TestKuke_AttachDetach_KeepsTaskRunning` (task not RUNNING after detach — fails on origin/main too), `TestKuke_Init_VerifyState` (residual state collision), and the `--run-path tmp/...` cell tests that try to dial the daemon socket without `--no-daemon`. None reproduce on the changed code paths.

Closes #258

### DIFF
--- a/cmd/kukeond/serve.go
+++ b/cmd/kukeond/serve.go
@@ -86,6 +86,13 @@ func runServe(cmd *cobra.Command, _ []string) error {
 		Controller: controller.Options{
 			RunPath:          runPath,
 			ContainerdSocket: viper.GetString(config.KUKEON_ROOT_CONTAINERD_SOCKET.ViperKey),
+			// Forward the kukeon group GID into the runner so per-container
+			// Attachable tty directories created by daemon-mediated apply/start
+			// inherit the same group-traversal layout `kuke init` applies to
+			// /opt/kukeon. Without this the runner sees KukeonGroupGID=0 and
+			// falls back to 0700 root-only — regressing #258 repro A under the
+			// daemon path even when --socket-gid was set.
+			KukeondSocketGID: socketGID,
 		},
 	}
 
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -104,6 +104,7 @@ func NewControllerExec(ctx context.Context, logger *slog.Logger, opts Options) *
 			ContainerdSocket:   opts.ContainerdSocket,
 			RunPath:            opts.RunPath,
 			ForceRegenerateCNI: opts.ForceRegenerateCNI,
+			KukeonGroupGID:     opts.KukeondSocketGID,
 		}),
 	}
 }
--- a/internal/controller/runner/attachable.go
+++ b/internal/controller/runner/attachable.go
@@ -25,6 +25,38 @@ import (
 	"github.com/eminwux/kukeon/internal/util/fs"
 )
 
+// attachableTTYDirRootMode is the mode applied to the per-container tty
+// directory while it is still root-owned (host-side and group-owned by the
+// kukeon group). 02750 = setgid + rwx for owner + r-x for group + no world
+// access, matching the rest of /opt/kukeon set up by `kuke init` so members
+// of the kukeon group can traverse without world access. The setgid bit
+// makes future siblings (sbsh socket, log, capture) inherit the kukeon
+// group automatically.
+const attachableTTYDirRootMode os.FileMode = os.ModeSetgid | 0o0750
+
+// attachableTTYDirNoGroupMode is the legacy fallback when the kukeon group
+// GID is not configured (e.g., older `kuke init` runs that never invoked
+// sysuser.EnsureUserGroup, or `--no-daemon` smoke tests under a tmp
+// runPath). In that mode there is no group to delegate access to, so the
+// directory stays owner-only (0700).
+const attachableTTYDirNoGroupMode os.FileMode = 0o0700
+
+// attachableTTYDirInitialPerms returns the (mode, gid) tuple to apply to a
+// freshly-prepared per-container tty directory before the workload
+// container starts. When kukeonGroupGID is non-zero, the directory is
+// group-traversable for members of the kukeon group (matches `kuke init`'s
+// /opt/kukeon layout); when it is zero, the directory stays root-only.
+//
+// The owner is set to root in both cases — the post-create step resets it
+// to the resolved container uid once containerd has parsed the image's
+// USER directive (#258 repro B).
+func attachableTTYDirInitialPerms(kukeonGroupGID int) (os.FileMode, int) {
+	if kukeonGroupGID > 0 {
+		return attachableTTYDirRootMode, kukeonGroupGID
+	}
+	return attachableTTYDirNoGroupMode, 0
+}
+
 // attachableBuildOpts returns the ctr.BuildOption slice to pass to
 // CreateContainerFromSpec for a given container spec. When Attachable=false
 // the slice is empty and the call is a no-op. When Attachable=true the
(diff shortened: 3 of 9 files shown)


eminwux:
**PR #259 Review — `fix: chown attachable tty dir to kukeon group + container uid (#258)`**

LGTM. The plumbing chain (`serve.go` socketGID → `controller.Options.KukeondSocketGID` → `runner.Options.KukeonGroupGID` → `attachableBuildOpts`/`attachablePostCreateChown`) is complete and the missing assignment in `cmd/kukeond/serve.go` is exactly what reproduced #258 under the daemon path. `os.Chown` on a directory does not strip the setgid bit on Linux, so 02750 survives the post-create owner reset to the container uid — group-traversal layout is preserved. No sensitive data in error strings, no dead symbols, license header + colocated tests in place. Daemon-parity check from `make dev-init` is in the test plan as required by CLAUDE.md.

Two non-blocking nits:

- **Mode redundancy** `internal/controller/runner/attachable.go:85`: `os.MkdirAll(ttyDir, 0o0700)` is immediately followed by an unconditional `os.Chmod` to `mode` from `attachableTTYDirInitialPerms`. The mkdir's mode arg is dead in practice (filtered by umask anyway, then overridden) — passing `0o0750` here would at least document intent. Non-blocking.
- **Inconsistent gid handling** `internal/controller/runner/attachable.go:163`: `attachableTTYDirInitialPerms` normalizes a negative `KukeonGroupGID` to 0, but `attachablePostCreateChown` passes the raw `r.opts.KukeonGroupGID` to `os.Chown`. Linux's chown(2) treats -1 as "leave unchanged" so this still works, but the asymmetry is a future foot-gun if someone routes a different sentinel through. Non-blocking.

Ready to merge.

PR #261, opened 2026-05-02
fix: close remaining gates on kukeon-group attachable end-to-end

### LINKED ISSUE #260: attachable: post-#258 still broken — profile.yaml 0o600, parent dirs 0o2700, socket 0o600 block kukeon-group attach end-to-end
After #258 (commit 22a41dc) `kuke run -a -p claude` still fails for an unprivileged member of the `kukeon` group. The fix only addressed the leaf `tty/` directory; three independent permission gates remain on the path between the kukeon-group operator and the per-container sbsh socket.

## Repro

On `main` at 22a41dc, as user `inwx` (uid 1000, member of `kukeon` gid 986):

```
$ kuke run -a -p claude
Cell "claude-9acd61" (realm "default", space "default", stack "default")
  - metadata: created
  - cgroup: created
  - root container: created
  - container "root": created
  - container "work": created
  - containers: started
Error: error attaching client: ping failed: dial unix
  /opt/kukeon/default/default/default/claude-9acd61/work/tty/socket:
  connect: permission denied
```

## Observed

Filesystem state for the failed cell on disk (image: Debian-based with `USER claude`, uid 1000):

```
drwx--S--- root kukeon  claude-9acd61/         ← 0o2700, group has no x
drwx--S--- root kukeon  claude-9acd61/work/    ← 0o2700, group has no x
drwxr-s--- 1000 kukeon  claude-9acd61/work/tty/ ← 0o2750 (the one #258 fixed)
-rw------- root kukeon  …/tty/profile.yaml      ← 0o600 root-owned
-rw------- 1000 kukeon  …/tty/log               (sbsh died before listen)
(no socket inode — sbsh never reached Listen)
```

sbsh log inside `tty/log`:

```
WARN  profile loader warning  reason=open: open /run/kukeon/tty/profile.yaml: permission denied
INFO  Profile not found  profile=kukeon
ERROR Failed to build terminal spec  error=profile \"kukeon\" not found in /run/kukeon/tty
```

The failure the operator sees is `permission denied` at path traversal (parent dirs deny group +x), but even if traversal were allowed there would be no socket to connect to because sbsh died before `Listen()`.

## Root cause

Three independent gates on the kukeon-group attach path. #258 only fixed the leaf `tty/` directory.

**Gate 1 — `profile.yaml` is `0o600 root:kukeon`.**
`internal/controller/runner/sbsh_profile.go:133` writes the profile with `os.WriteFile(path, data, 0o600)`. The container runs as `claude` (uid 1000), which is not the file owner; inside the container `claude` has its own /etc/group with `claude:1000` — it is not a member of the host's `kukeon` group, so the group-mode bits are unreachable from inside the container. sbsh's profile loader gets `permission denied`, the loader returns no profiles, sbsh exits with `Failed to build terminal spec`, and **no socket inode is created**. This is the failure observed today and is what the user actually hits.

**Gate 2 — cell dir and per-container dir are `0o2700 root:kukeon`.**
`internal/metadata/metadata.go:55` calls `os.MkdirAll(filepath.Dir(file), 0o700)` for the cell metadata.json and again for each container's metadata.json. The setgid bit is inherited from the parent (kukeon-group, set up by `kuke init`), giving `0o2700`. Without group +x, a kukeon-group user cannot traverse `…/claude-9acd61/work/` to reach the inner `tty/` (which #258 made `0o2750`). Even if Gate 1 were fixed and a socket appeared, the operator's `connect()` call never reaches the socket inode — the path resolution fails earlier with `permission denied`. Reproduces this exact symptom standalone:

```
$ stat -c '%a %U:%G' /opt/kukeon/default/default/default/claude-9acd61
2700 root:kukeon
$ sudo -u inwx ls /opt/kukeon/default/default/default/claude-9acd61/work/tty/
ls: cannot open directory '…': Permission denied
```

**Gate 3 (latent) — sbsh chmods the socket itself to `0o600`.**
`sbsh/internal/terminal/terminalrunner/sockets.go:63` does `os.Chmod(socketFile, 0o600)` after `Listen()`. The socket is owned by the container's runtime uid (1000 in this image) with no group bits set. After Gates 1 and 2 are unblocked, an attaching kukeon-group user with a *different* uid still cannot `connect()` to the socket — Linux requires write permission on the socket inode to connect. This gate lives in the sbsh repo, not kukeon, but kukeon has to plumb a `--socket-mode 0660 --socket-gid <kukeon-gid>` (or equivalent) through the wrapper invocation in `internal/ctr/attachable.go:withAttachableArgsWrap`. Listed here because Gates 1 and 2 cannot be claimed \"fixed\" until the end-to-end attach actually works.

## Suggested fix

- **Gate 1:** Either widen `profile.yaml` to a mode the in-container USER can read (e.g., `0o644`) — the file carries no secrets, only the prompt string and onInit scripts already declared in the CellProfile — or chown it to the resolved container uid before the container starts (the same uid the post-create chown of `tty/` now uses; reuse `Client.ContainerProcessUID`). The chown variant keeps `0o600` semantics; the chmod variant is one line.
- **Gate 2:** Plumb a parent-dir mode of `0o2750` (or, with the realm/space/stack/cell setgid inheritance already in place, plain `0o0750`) into `metadata.WriteMetadata`'s `MkdirAll`. That matches the rest of `/opt/kukeon` and is what `kuke init` already documents as the layout for kukeon-group operators (see `cmd/kuke/init/init.go` `kukeonRunPathDirMode = os.ModeSetgid | 0o750`). Today the function takes no perm argument, so either thread one through or special-case the cell/container metadata callers.
- **Gate 3:** File the sbsh side as a separate issue against `eminwux/sbsh` — add `--socket-mode` and `--socket-gid` to `sbsh terminal`, and have kukeon's `withAttachableArgsWrap` pass the kukeon GID resolved at `kuke init` time (already plumbed into the runner as `KukeonGroupGID` by #258). Track here as a blocking dependency for the end-to-end fix.

The original #258 suggested-fix bullet \"the kukeon group still gets traverse + read for the host-side attach\" was correct in spirit but assumed (a) the socket would be reachable through traversable parent dirs and (b) the socket itself would be group-readable — both of which turned out not to hold. Gates 2 and 3 close the gap.

eminwux:
## Summary
- Closes the three permission gates that survived #258 and still blocked `kuke run -a -p claude` for an unprivileged member of the `kukeon` group on `main`.
- **Gate 1 (profile.yaml unreadable to in-container USER):** `attachablePostCreateChown` now also chowns `<tty>/profile.yaml` to the resolved container uid. Mode stays `0o600` — the container's runtime uid is now the file owner, so sbsh's profile loader can open it and reach `Listen()`. Without this, sbsh died with `Failed to build terminal spec` before ever creating the socket inode.
- **Gate 2 (cell + per-container parent dirs at `0o2700`):** `metadata.WriteMetadata`'s `MkdirAll` now uses `0o0750` and explicitly chmods the metadata file's parent to `os.ModeSetgid|0o0750` after the fact (Linux silently strips an explicit `S_ISGID` from `mkdir`'s mode unless inherited from the parent). `attachableBuildOpts` likewise chmods the per-container parent of `tty/` to the same group-traversable mode. Self-healing: existing dirs are re-chmoded on next write.
- **Gate 3 (sbsh socket hard-coded `0o600`):** bumps `github.com/eminwux/sbsh` to `v0.10.0` (which adds `--socket-mode`/`--socket-gid`) and plumbs `SocketGID` through `AttachableInjection` so the wrapper passes `--socket-mode 0660 --socket-gid <kukeonGID>` to `sbsh terminal` whenever `KukeonGroupGID > 0`. Default behavior preserved when GID is unset.
- End-to-end repro from the issue (`kuke run -a -p claude` as user in the `kukeon` group) now reaches the `claude>` prompt; on-disk verifies `<cell>/`, `<cell>/<container>/`, `<container>/tty/` all `drwxr-s---` `root:kukeon` (or `<uid>:kukeon` after the post-create chown), `profile.yaml` `0o600 <uid>:kukeon`, and `socket` `srw-rw---- <uid>:kukeon`.

## Test plan
- [x] `go build ./...` and `go vet ./...`
- [x] `make test` (full unit suite minus `/e2e`) — green
- [x] New `TestBuildContainerSpec_AttachableTrue_SocketFlagsEmittedWithGID` and `TestBuildContainerSpec_AttachableTrue_NoSocketFlagsWhenGIDUnset` lock the wrapper's flag emission both ways
- [x] New `TestWriteMetadata_ParentDirModeIsGroupTraversable` and `TestWriteMetadata_ChmodIsIdempotent_OnReuse` cover the metadata-dir mode fix on fresh and pre-existing dirs
- [x] `make dev-init` — daemon-parity check passes (both `kuke get realms` and `--no-daemon` produce identical Ready output for `default` and `kuke-system`)
- [x] Manual end-to-end: `kuke run -a -p claude` as `inwx` (uid 1000, in `kukeon` gid 986) reaches the `claude>` prompt; permission denied on the socket connect is gone
- [ ] `make e2e` — not run locally; CI

## Notes
- The staged `/opt/kukeon/cache/sbsh/<arch>/sbsh` binary must be sbsh `v0.10.0` or later for the new flags to be honored. The wrapper omits the flags when `KukeonGroupGID = 0`, so a daemon initialized without `kuke init`'s sysuser bootstrap stays compatible with older sbsh binaries.

Closes #260

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/containerd/typeurl/v2 v2.2.3
 	github.com/containernetworking/cni v1.3.0
 	github.com/creack/pty v1.1.24
-	github.com/eminwux/sbsh v0.8.0
+	github.com/eminwux/sbsh v0.10.0
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/runtime-spec v1.2.1
 	github.com/spf13/cobra v1.10.1
--- a/go.sum
+++ b/go.sum
@@ -55,8 +55,8 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/distribution/reference v0.6.0 h1:0IXCQ5g4/QMHHkarYzh5l+u8T3t73zM5QvfrDyIgxBk=
 github.com/distribution/reference v0.6.0/go.mod h1:BbU0aIcezP1/5jX/8MP0YiH4SdvB5Y4f/wlDRiLyi3E=
-github.com/eminwux/sbsh v0.8.0 h1:PI0tUwgkYcdFlug6Ga3XhbqiHY/JqrPT9QVOEIhTFYY=
-github.com/eminwux/sbsh v0.8.0/go.mod h1:AsDnK44VUZjB5tpUn6oV/0isaIudbLLQst8FYFXPJcs=
+github.com/eminwux/sbsh v0.10.0 h1:4mNa7UQPmveVS0v5mesnF4RmB1RojV+kPTr50LpcJBI=
+github.com/eminwux/sbsh v0.10.0/go.mod h1:AsDnK44VUZjB5tpUn6oV/0isaIudbLLQst8FYFXPJcs=
 github.com/envoyproxy/go-control-plane v0.9.0/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
 github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
 github.com/envoyproxy/go-control-plane v0.9.4/go.mod h1:6rpuAdCZL397s3pYoYcLgu1mIlRU8Am5FuJP05cCM98=
--- a/internal/controller/runner/attachable.go
+++ b/internal/controller/runner/attachable.go
@@ -17,8 +17,10 @@
 package runner
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/eminwux/kukeon/internal/ctr"
 	intmodel "github.com/eminwux/kukeon/internal/modelhub"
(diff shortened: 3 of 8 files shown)


eminwux:
**PR #261 Review — `fix: close remaining gates on kukeon-group attachable end-to-end`**

Targeted three-gate fix that lines up with the issue's repro: (gate 1) `attachablePostCreateChown` now chowns `profile.yaml` to the container uid; (gate 2) `WriteMetadata` and `attachableBuildOpts` apply `02750` with explicit post-`MkdirAll` chmod to defeat Linux's silent setgid stripping on dirs whose parent isn't already setgid; (gate 3) sbsh bumped to v0.10.0 and `--socket-mode 0660 --socket-gid` plumbed through `AttachableInjection.SocketGID`. Both branches (GID>0 and GID=0 fallback) are locked by tests, comments explain the non-obvious WHY (umask filtering, `S_ISGID` inheritance quirk, sbsh-v0.10 flag gating), and `make dev-init` daemon-parity check is in the test plan as CLAUDE.md requires for daemon-touching PRs. No sensitive data in error strings; chown/chmod errors are wrapped with `%w`.

Nit (non-blocking):

- **Redundant bit-OR** `internal/metadata/metadata.go:358`: `metadataDirMode|os.ModeSetgid` is a no-op because `metadataDirMode` already includes `os.ModeSetgid`. Pure cosmetic — drop the `|os.ModeSetgid` for clarity.

LGTM.

PR #263, opened 2026-05-02
feat: add kukeond background cell-reconciliation loop

### LINKED ISSUE #161: kukeond: add background reconciliation loop (cell-only v1)
## Context
`kukeond` has no background loop today — drift between cell `spec` and
observed status is corrected only when the user runs `kuke refresh` or via
in-line code paths during start/stop. A cell whose container exits externally
still reports `Ready` until someone touches it. This also leaves #46 (Session
bounded lifetime) and the future `kuke run --rm` work (orphan-cell cleanup)
with no place to anchor self-healing behavior.

## Proposal
Add a periodic reconciliation loop to `kukeond`:

- **Scope (v1):** cells only. Walk every cell in every realm; reconcile spec
  against observed container state and update `status.state`.
- **Trigger:** fixed-interval ticker. Interval is a field on
  `ServerConfiguration` (#159); a conservative default (e.g. 30s) ships with
  the feature.
- **Concurrency:** single goroutine, single instance — kukeond is one process
  per host. No leader election, no multi-node coordination.
- **Failure:** errors during a pass are logged to kukeond stderr; the loop
  continues on the next tick.
- **Observability:** one log line per pass — brief on success
  (e.g. "reconcile ok: N cells"), detailed on failure.

This is the periodic, daemon-side counterpart to `kuke refresh`'s on-demand
path. Reuse the refresh logic where practical.

## Acceptance criteria
- [ ] kukeond runs a background ticker that, on every tick, reconciles each
      cell's `status` against the observed container state.
- [ ] When a container inside a `Ready` cell exits externally (e.g.
      `ctr -n kuke-system.kukeon.io task kill`), the cell's status flips to
      `Stopped` within one tick interval — no manual `kuke refresh` needed.
- [ ] Tick interval is a field on ServerConfiguration (#159); a sensible
      default ships.
- [ ] Single-instance only — no leader election logic.
- [ ] Errors during a pass are logged to kukeond stderr; the loop continues
      on the next tick.
- [ ] One log line per pass: brief on success, detailed on failure.
- [ ] Existing `kuke refresh` behavior unchanged.

## Notes
- Depends-on: #159 — for ServerConfiguration to host the interval setting.
- Follow-up to file separately: sanitize `status` fields across cell, stack,
  space, and realm. Today's status structs are not fully populated; the
  reconciler exposes the gap but should not block on it.
- Unblocks #46 (Session bounded lifetime) and a future `--rm` hardening pass
  to move orphan cleanup beyond best-effort.
- Deliberately keeping kukeon on a polling model for now (no event system).
  `kuke refresh` remains the on-demand counterpart.

eminwux:
## Summary
- Adds a periodic, single-goroutine ticker inside `kukeond` that walks every realm/space/stack and re-derives each cell's `status.state` from the observed root-container task — so a cell whose root task is killed externally (e.g. `ctr -n … task kill`) flips from `Ready` to `Stopped` within one tick interval, with no operator `kuke refresh` required.
- Adds a `reconcileInterval` field to `ServerConfiguration` (and a matching `--reconcile-interval` flag / `KUKEOND_RECONCILE_INTERVAL` env), default `30s`. A zero or negative duration disables the loop.
- Layers the new state-from-task derivation into a new `runner.ReconcileCell` so existing `kuke refresh` (which still calls `RefreshCell`) is byte-for-byte unchanged in behavior.

### Design notes (so reviewer can push back)
- The loop lives on `daemon.Server` and is bound to the daemon context — single goroutine, single instance, no leader election (per the issue).
- Errors during a pass are logged once and the loop keeps ticking — a flaky containerd connection cannot wedge reconciliation. One log line per pass: `"reconcile ok"` on success, error/warn variants when something failed.
- `RefreshCell` is intentionally untouched. `ReconcileCell` is its daemon-side counterpart that adds root-container task introspection — keeping AC #7 (kuke refresh unchanged) literal.
- Container statuses are populated as a side effect so `kuke get cell` reflects container state without the operator having to run `kuke refresh`.
- `daemon.Server` got a private `reconcileFn` seam so package-internal tests can drive the ticker without booting a real controller.

## Test plan
- [x] `go test ./internal/controller/... ./internal/daemon/... ./internal/serverconfig/... ./cmd/kukeond/...` — all green (new walk / error-resume / disabled / ctx-cancel / per-pass error coverage).
- [x] `go test $(go list ./... | grep -v /e2e)` — green except for one pre-existing `internal/cni.TestBootstrapCNI` failure on `origin/main` (mkdir /opt/cni: permission denied — sandbox restriction, also reproduces on a clean clone of `origin/main`).
- [x] `go vet ./...` clean.
- [x] `gofmt -l` clean for changed files.
- [ ] `make e2e` — not run, sandbox lacks docker + sudo.
- [ ] `make dev-init` smoke test — not run, sandbox lacks `sudo`, `docker`, `containerd`, and the privileged mounts the script needs. The change touches `/run/kukeon`, `/opt/kukeon`, and the daemon read path; this check should be re-run by the reviewer or in CI before merge per the project CLAUDE.md.

Closes #161

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -111,6 +111,10 @@ var (
 	KUKEOND_CONFIGURATION = DefineKV(
 		"KUKEOND_CONFIGURATION", "kukeond/configuration", "/etc/kukeon/kukeond.yaml",
 	)
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKEOND_RECONCILE_INTERVAL = DefineKV(
+		"KUKEOND_RECONCILE_INTERVAL", "kukeond/reconcileInterval", "30s",
+	)
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_INIT_REALM = DefineKV("KUKE_INIT_REALM", "kuke/init/realm")
--- a/cmd/kuke/uninstall/uninstall_test.go
+++ b/cmd/kuke/uninstall/uninstall_test.go
@@ -165,6 +165,9 @@ func (s *stubController) PurgeContainer(intmodel.Container) (controller.PurgeCon
 	panic("not used")
 }
 func (s *stubController) RefreshAll() (controller.RefreshResult, error) { panic("not used") }
+func (s *stubController) ReconcileCells() (controller.ReconcileResult, error) {
+	panic("not used")
+}
 
 // runCmd builds the cobra command, plugs in a stub controller and a logger,
 // and feeds the supplied stdin payload. Returns the captured stdout/stderr
--- a/cmd/kukeond/kukeond.go
+++ b/cmd/kukeond/kukeond.go
@@ -124,6 +124,17 @@ func NewKukeondCmd() (*cobra.Command, error) {
 		return nil, err
 	}
 
+	cmd.PersistentFlags().String(
+		"reconcile-interval", config.KUKEOND_RECONCILE_INTERVAL.Default,
+		"Period of the cell-reconciliation loop (Go duration; 0 disables)",
+	)
+	if err := viper.BindPFlag(
+		config.KUKEOND_RECONCILE_INTERVAL.ViperKey,
+		cmd.PersistentFlags().Lookup("reconcile-interval"),
+	); err != nil {
+		return nil, err
+	}
+
 	bindEnvVars()
 
 	cmd.AddCommand(newServeCmd())
@@ -142,6 +153,7 @@ func bindEnvVars() {
 		config.KUKEON_ROOT_LOG_LEVEL,
 		config.KUKEOND_SOCKET,
 		config.KUKEOND_SOCKET_GID,
+		config.KUKEOND_RECONCILE_INTERVAL,
 	} {
 		_ = v.BindEnv()
 	}
@@ -170,6 +182,11 @@ func applyServerConfiguration(cmd *cobra.Command, spec v1beta1.ServerConfigurati
 	}
 	if spec.LogLevel != "" && !flagChanged(cmd, "log-level") && !envSet(config.KUKEON_ROOT_LOG_LEVEL) {
 		viper.Set(config.KUKEON_ROOT_LOG_LEVEL.ViperKey, spec.LogLevel)
+	}
+	if spec.ReconcileInterval != "" &&
+		!flagChanged(cmd, "reconcile-interval") &&
+		!envSet(config.KUKEOND_RECONCILE_INTERVAL) {
+		viper.Set(config.KUKEOND_RECONCILE_INTERVAL.ViperKey, spec.ReconcileInterval)
 	}
 }
 
--- a/cmd/kukeond/kukeond_test.go
+++ b/cmd/kukeond/kukeond_test.go
@@ -49,11 +49,12 @@ func TestApplyServerConfigurationDefaultsLayered(t *testing.T) {
 	}
 
 	spec := v1beta1.ServerConfigurationSpec{
-		Socket:           "/run/kukeon/from-config.sock",
-		SocketGID:        4242,
-		RunPath:          "/opt/kukeon-from-config",
-		ContainerdSocket: "/run/containerd/from-config.sock",
-		LogLevel:         "warn",
+		Socket:            "/run/kukeon/from-config.sock",
+		SocketGID:         4242,
+		RunPath:           "/opt/kukeon-from-config",
+		ContainerdSocket:  "/run/containerd/from-config.sock",
+		LogLevel:          "warn",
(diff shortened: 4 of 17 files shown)


eminwux:
**PR #263 Review — `feat: add kukeond background cell-reconciliation loop`**

Clean implementation: single goroutine bound to `s.ctx`, errors collected without aborting the walk, `kuke refresh` untouched per AC #7, `defer recover()` guards the per-tick callback. No correctness or sensitive-data issues. Test coverage spans walk, error-resume, disabled, and ctx-cancel paths.

- **Dead local fields** `internal/controller/runner/refresh.go:361,380`: `newStatus.Network` is initialized from `originalStatus.Network` but never copied back to `cell.Status`, and `cell.Status.CgroupPath = newStatus.CgroupPath` is a no-op since `populateCellContainerStatuses` only mutates `Status.Containers`. Drop the unused fields, or write `cell.Status.Network = newStatus.Network` if the intent was to defend Network against a future change to `populateCellContainerStatuses`.
- **Reconcile teardown race** `internal/daemon/server.go:163-175`: the loop exits only on `s.ctx.Done()`, but `Stop()` doesn't cancel ctx — it just closes the listener and core. Production `runServe` cancels the signal-derived ctx before `Stop`, so this is fine; but an in-flight pass during shutdown can call `s.core.ReconcileCells` after `s.core.Close()` ran, producing noisy panic logs (the `recover()` makes it safe).
- **Test goroutine leak** `internal/daemon/reconcile_test.go:36,51`: `TestServer_ReconcileLoopFires` and `TestServer_ReconcileLoopContinuesOnError` use `context.Background()`; after `t.Cleanup` calls `srv.Stop()`, the reconcile goroutine keeps ticking forever (the override `reconcileFn` keeps the test's atomic counter alive). Use `context.WithCancel` and cancel in cleanup for tidy goroutine accounting.

LGTM. Worth re-running `make dev-init` (per project CLAUDE.md, change touches the daemon read path) before merge — author flagged sandbox couldn't.

eminwux:
Addressed in ac115df:
- Dead local fields in `internal/controller/runner/refresh.go` ReconcileCell — replaced the `newStatus` struct (whose `Network`/`CgroupPath` fields were never copied back) with a single `newState` local. Dropped the no-op `cell.Status.CgroupPath = newStatus.CgroupPath` and the dead `originalStatus.CgroupPath != cell.Status.CgroupPath` comparison. `populateCellContainerStatuses` only mutates `Status.Containers`, so `CgroupPath` is unchanged across the call — tracking it was misleading.
- Reconcile teardown race in `internal/daemon/server.go` — added an internal `stopCh` and `loopWG`. `Stop()` now closes `stopCh` and `loopWG.Wait()`s before tearing down `core`, so a tick already inside `reconcileFn` cannot race `core.Close()`. The loop now selects on `s.ctx.Done()` OR `stopCh`, so production (parent cancels ctx) and standalone `Stop()` both work.
- Test goroutine leak in `internal/daemon/reconcile_test.go` — `newTestServer` now uses `context.WithCancel(context.Background())` and registers the cancel with `t.Cleanup`, per reviewer suggestion. Belt-and-suspenders with the `loopWG.Wait()` in `Stop` above.

Test plan:
- [x] `go test $(go list ./... | grep -v /e2e)` — green except for the same pre-existing `internal/cni.TestBootstrapCNI` sandbox-permission failure on `origin/main` already noted in the PR body.
- [x] `go vet ./...` clean.
- [x] `gofmt -l` clean for changed files.
- [ ] `-race` run — sandbox lacks `gcc` / cgo; reviewer or CI should pick this up.
- [ ] `make dev-init` smoke — sandbox still lacks docker + sudo; the change to `Stop`/loop is daemon-internal and behaviourally a no-op when ctx already cancels first (production path), so the existing dev-init runs the reviewer flagged remain the right place.

eminwux:
**PR #263 Re-review — `feat: add kukeond background cell-reconciliation loop`**

Follow-up commit `ac115df5` addresses all three non-blocking nits from the prior LGTM:

- `refresh.go`: collapsed dead `newStatus` struct to a plain `newState`, dropped the no-op `cell.Status.CgroupPath = newStatus.CgroupPath` and its dead-equality guard. Clean.
- `server.go`: added `stopCh` + `loopWG`; `Stop()` closes the channel and waits for any in-flight reconcile pass before `core.Close()`. Idempotent — `Stop()` is guarded by `s.closed`, so no double-`close(stopCh)`; `loopWG.Wait()` is also safe when the loop never started (interval ≤ 0 → `Add` never called → counter is 0).
- `reconcile_test.go`: `newTestServer` now derives a cancellable ctx with `t.Cleanup(cancel)`. No more leaked goroutines.

LGTM. The `make dev-init` smoke (project CLAUDE.md) is still un-run in the test plan — worth a local pass before merge.

PR #271, opened 2026-05-03
fix: gate AutoDelete on Ready observation

### LINKED ISSUE #269: controller/runner: guard AutoDelete from reaping in-flight cell creation
Surfaced as a non-blocking, deferred nit during review of #267 (`refactor: move Spec.AutoDelete cleanup to the reconciler`).

**Where:** `internal/controller/runner/refresh.go:388`

**What:** Between cgroup creation and root-container registration inside `CreateCell`, `GetContainerState` returns `Stopped` for a not-yet-existing container (`internal/controller/runner/container_state.go:168`). That `Stopped` would feed into `autoDeleteCell` and reap a still-being-created `--rm` cell.

**Why bug-adjacent:** the 30s `ReconcileInterval` vs sub-second `CreateCell` makes this unlikely in practice today, but the window exists and would narrow if the ticker is ever bumped or `CreateCell` blocks on an in-flight failure. Filing per pr-reviewer CLAUDE.md "Bug-adjacent observations are not subject to the bar."

**Suggested fix (from review):** require a `Ready` observation before AutoDelete can trigger — i.e., gate `autoDeleteCell` on the cell having transitioned through `Ready` at least once, so a cell that has never been observed Ready cannot be reaped from the "container does not exist yet → Stopped" branch.

eminwux:
## Summary
- Gate the daemon reconciler's AutoDelete cleanup on a one-way `ReadyObserved` latch so a cell that has never been observed Ready cannot be reaped from the "container does not exist in containerd yet → Stopped" branch (`internal/controller/runner/container_state.go:168`). Without the latch, a reconciler tick that lands inside the gap between cgroup creation and root-container registration during `CreateCell` resolves the cell to Stopped and `autoDeleteCell` reaps an in-flight cell.
- The latch flips true on either a fresh `Ready` observation, the persisted state already being `Ready` (covers a synchronous `Start` that wrote `Ready` before any reconciler tick — including the first tick after a daemon restart), or the prior persisted latch value (so it survives across ticks and across restarts).
- Persisted via `CellStatus.ReadyObserved` (`json/yaml: readyObserved,omitempty`) so a `kuke run --rm` cell that was already Ready at daemon shutdown still gets cleaned up on the next tick after restart.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -run "TestLatchReadyObserved|TestShouldAutoDeleteCell|TestCellStateAutoDeleteTriggers|TestCellRoundTripV1Beta1_ReadyObserved" -v ./internal/apischeme/... ./internal/controller/runner/...` — new pure-function tests for the latch + gate, plus an external↔internal round-trip test for the persisted field
- [x] `make dev-init` — daemon-parity check passes (`kuke get realms` and `kuke get realms --no-daemon` produce identical output)
- [ ] `make e2e` — not run locally; CI will exercise

Closes #269

### DIFF
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -817,7 +817,8 @@ func ConvertCellDocToInternal(in ext.CellDoc) (intmodel.Cell, error) {
 				Network: intmodel.CellNetworkStatus{
 					BridgeName: in.Status.Network.BridgeName,
 				},
-				Containers: convertContainerStatusesToInternal(in.Status.Containers),
+				Containers:    convertContainerStatusesToInternal(in.Status.Containers),
+				ReadyObserved: in.Status.ReadyObserved,
 			},
 		}
 
@@ -876,7 +877,8 @@ func BuildCellExternalFromInternal(in intmodel.Cell, apiVersion ext.Version) (ex
 				Network: ext.CellNetworkStatus{
 					BridgeName: in.Status.Network.BridgeName,
 				},
-				Containers: buildContainerStatusesExternalFromInternal(in.Status.Containers),
+				Containers:    buildContainerStatusesExternalFromInternal(in.Status.Containers),
+				ReadyObserved: in.Status.ReadyObserved,
 			},
 		}
 
--- a/internal/apischeme/scheme_test.go
+++ b/internal/apischeme/scheme_test.go
@@ -437,6 +437,59 @@ func TestCellRoundTripV1Beta1_NetworkBridgeName(t *testing.T) {
 	}
 }
 
+// TestCellRoundTripV1Beta1_ReadyObserved covers the persistence side of
+// the AutoDelete Ready-gate from #269: the latch must round-trip
+// external→internal→external so it survives daemon restarts. Without
+// it, a `kuke run --rm` cell that was Ready at shutdown would lose the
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #271 Review — `fix: gate AutoDelete on Ready observation`**

Tight, well-scoped fix for #269. The latch is a pure function with a clear one-way invariant, the round-trip through `apischeme` is wired in both directions and pinned by `TestCellRoundTripV1Beta1_ReadyObserved`, and the regression case from the bug is locked down explicitly in `TestShouldAutoDeleteCell:blocked_when_never_ready_regression_269`. No correctness concerns, no leak surface (no new error/log strings), no dead code. `make dev-init` is in the test plan per CLAUDE.md.

One observation, non-blocking:

- **Behavior** `internal/controller/runner/refresh.go:471`: a `Spec.AutoDelete` cell that never reaches Ready (e.g. image pull failure, OCI runtime error before the root task exists) is now never reaped by the reconciler — the latch stays false. That matches Docker's `--rm` semantics (no cleanup when `create` itself fails) and is consistent with the test's `never_ready_mid_create_stays_false` case, so likely intended; flagging only because it's a behavior change vs. the pre-fix code.

LGTM.

PR #274, opened 2026-05-03
fix: kill cell on attach loop return under -a --rm

### LINKED ISSUE #265: kuke run --rm: cell never auto-deleted when root container is a long-lived sleeper and operator attaches to a peer

eminwux:
## Summary

- `kuke run -a --rm` left the cell alive forever when the root container was a long-lived sleeper and the operator detached from a peer attachable, because the daemon's auto-delete trigger keys on the *root* task exiting and `sleep infinity` never does. Fix: when both flags are set, the CLI sends `KillCell` once `runAttachLoop` returns, giving the reconciler the Stopped state it needs to reap the cell on the next tick.
- The change is scoped to `cmd/kuke/run/run.go` only — daemon contract and `Spec.AutoDelete` semantics are unchanged. With `--rm` alone (no `-a`), behavior is preserved: the daemon's reconciler still owns cleanup keyed on root-task exit.
- Picked Option B from the issue (smaller, no Spec/RPC schema change) over Option A. Option A would have required threading the attach target through `CreateCell` args and changing the reconciler's auto-delete trigger from "root stopped" to "auto-delete-target stopped" — viable but a much larger blast radius for a UX bug. Filed in the PR body so the reviewer can push back if Option A is preferred.
- `--rm` flag docstring updated to describe the new -a trigger and call out that the no-`-a` path is unchanged.
- Errors from KillCell are best-effort: `ErrCellNotFound` is silent (the daemon may have raced ahead and reaped already), other RPC failures are logged to stderr but do not override the attach exit code.
- Fix-up (3f801c6): the already-Ready idempotent short-circuit now funnels through the same `attachAndMaybeAutoDelete` helper as the create path, so `kuke run -a --rm -f cell.yaml` against an existing Ready cell also fires `KillCell` on detach (reviewer-flagged gap; otherwise users with a cell from a pre-fix run could not recover via -a --rm).

## Test plan

- [x] `go test ./cmd/kuke/run/...` — seven unit tests cover the -a --rm path: KillCell fires on clean detach, on attach-loop error, KillCell failure does not surface as run error, ErrCellNotFound is silent, --rm without -a does not call KillCell, -a without --rm does not call KillCell, and the already-Ready idempotent branch also fires KillCell.
- [x] `go vet ./...` — clean.
- [x] `make dev-init` — daemon-parity check passes; both `kuke get realms` and `kuke get realms --no-daemon` produce the matching default + kuke-system table from CLAUDE.md.
- [ ] `make test-e2e` — pre-existing failures on `origin/main` (TestKuke_Init_VerifyState and others fail because the e2e harness assumes a clean host but a kukeond cell from `make dev-init` is already running). Verified the same set fails on baseline without my changes; my change does not regress e2e.

Closes #265

### DIFF
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -90,11 +90,15 @@ func NewRunCmd() *cobra.Command {
 		"Container to attach to (only valid with -a; must be attachable)")
 	_ = viper.BindPFlag(config.KUKE_RUN_CONTAINER.ViperKey, cmd.Flags().Lookup("container"))
 	cmd.Flags().Bool("rm", false,
-		"Best-effort delete the cell after the root container's task exits "+
-			"(any rc). Daemon-mode only — incompatible with --no-daemon. "+
-			"Cleanup runs from kukeond's reconcile loop, so latency is "+
-			"bounded by the reconcile interval rather than firing the "+
-			"instant the task exits.")
+		"Best-effort delete the cell after it is no longer needed "+
+			"(any rc). Without -a, the trigger is the root container's "+
+			"task exit. With -a, the trigger is the attach loop "+
+			"returning (clean detach, exit, or error) — the CLI then "+
+			"sends KillCell so a long-lived root (e.g. `sleep infinity`) "+
+			"does not pin the cell. Daemon-mode only — incompatible "+
+			"with --no-daemon. Cleanup runs from kukeond's reconcile "+
+			"loop, so latency is bounded by the reconcile interval "+
+			"rather than firing the instant the trigger fires.")
 	_ = viper.BindPFlag(config.KUKE_RUN_RM.ViperKey, cmd.Flags().Lookup("rm"))
 
 	_ = cmd.RegisterFlagCompletionFunc("realm", config.CompleteRealmNames)
@@ -222,9 +226,40 @@ func runRun(cmd *cobra.Command, args []string) error {
 		return printErr
 	}
 	if flags.doAttach {
-		return attachAfterRun(cmd, client, cellDoc, flags.containerFlag)
+		attachErr := attachAfterRun(cmd, client, cellDoc, flags.containerFlag)
+		if flags.autoDelete {
+			// With -a --rm, the operator's "I'm done" signal is the
+			// attach loop returning, not the root container exiting.
+			// When the attach target is a peer of a long-lived root
+			// (`sleep infinity` is the standard idiom), the root task
+			// never exits on its own and the reconciler's auto-delete
+			// trigger never fires. Send KillCell here so the root
+			// task is SIGKILL'd and the next reconcile pass reaps the
+			// cell. Best-effort per the --rm contract: a cleanup
+			// failure does not override attachErr.
+			autoDeleteAfterAttach(cmd, client, cellDoc)
+		}
+		return attachErr
 	}
 	return nil
+}
+
+// autoDeleteAfterAttach drives the -a --rm cleanup path. KillCell is
+// idempotent — if the attach target was the root and the user typed
+// `exit`, the root task is already gone and KillCell just confirms it;
+// the kill+delete sequence in autoDeleteCell tolerates a missing task.
+// A KillCell failure (daemon down, RPC error, cell already gone) is
+// reported but not returned: the operator has detached, --rm is
+// best-effort, and the daemon's reconciler still owns final cleanup.
+func autoDeleteAfterAttach(cmd *cobra.Command, client kukeonv1.Client, doc v1beta1.CellDoc) {
+	if _, err := client.KillCell(cmd.Context(), doc); err != nil {
+		if errors.Is(err, errdefs.ErrCellNotFound) {
+			return
+		}
+		fmt.Fprintf(cmd.ErrOrStderr(),
+			"kuke run: --rm cleanup: failed to kill cell %q: %v\n",
+			doc.Metadata.Name, err)
+	}
 }
 
 // attachAfterRun resolves the post-start attach target per the documented
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -209,7 +209,7 @@ func runRun(cmd *cobra.Command, args []string) error {
 				return printErr
 			}
 			if flags.doAttach {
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #274 Review — `fix: kill cell on attach loop return under -a --rm`**

Targeted fix for #265 with clean test coverage of the create+attach path. KillCell error reporting is appropriately best-effort and leaks no sensitive data. One blocking gap on completeness:

- **Incomplete fix** `cmd/kuke/run/run.go:205-207`: the early-return idempotent branch (existing-Ready cell, matching spec) still does `return attachAfterRun(...)` without firing `autoDeleteAfterAttach`, so re-running `kuke run -a --rm -f cell.yaml` against an already-up cell reproduces the original #265 symptom — KillCell never fires on detach. Same for any user with a dangling cell from a pre-fix run trying to recover via `-a --rm`. Extract a small helper that pairs `attachAfterRun` + `autoDeleteAfterAttach if --rm` and call it from both `doAttach` branches (line 205-207 and the new line 222+); add a test that seeds `getCellFn` with a Ready existing cell and asserts `killCalls == 1`.

LGTM once the existing-Ready idempotent branch also fires KillCell on detach — the rest of the change is solid.

eminwux:
Addressed in 3f801c6:
- Extracted `attachAndMaybeAutoDelete` helper that pairs `attachAfterRun` + `autoDeleteAfterAttach` (when --rm). Both `doAttach` branches in `runRun` now route through it — the create-and-start path and the already-Ready idempotent short-circuit. Re-running `kuke run -a --rm -f cell.yaml` against a Ready cell now fires KillCell on detach.
- Added `TestRun_RmAttach_AlreadyReady_StillKillsCellAfterAttachLoopReturns` — seeds `getCellFn` with a Ready existing cell and asserts `killCalls == 1` (was 0 before the fix-up).

eminwux:
LGTM. Follow-up commit `3f801c6` extracts `attachAndMaybeAutoDelete` and routes both `doAttach` branches through it, plus a regression test (`TestRun_RmAttach_AlreadyReady_StillKillsCellAfterAttachLoopReturns`) that seeds a Ready cell and asserts `KillCell` fires on detach. Blocking item from prior review resolved.

PR #276, opened 2026-05-03
fix: close ReadyObserved latch eagerly on synchronous Ready writes

### LINKED ISSUE #275: kuke run -a --rm: cell stuck in Stopped when KillCell races the first reconciler tick
## Repro
On `main` (c6db87e), with the daemon running and reconcile-interval at the 30s default:

```bash
kuke run --rm -a -p kukeon-dev   # exit the attach session within ~25s
sudo kuke get cells               # cell stays in Stopped indefinitely
```

Reproduced this session against cell `kukeon-dev-cb9290` (root: `sleep infinity`, attachable peer `work` with default `tty.default: work`). Lifetime ~8s (started 22:09:58, killed 22:10:06). Three minutes later the cell was still in `Stopped`.

## Observed
`/opt/kukeon/default/default/default/kukeon-dev-cb9290/metadata.json`:

```
"autoDelete": true
"state": 2,           // CellStateStopped
// no "readyObserved" entry → field is false (omitempty)
```

`kuke get cells` keeps showing the cell as `Stopped` past the 30s reconcile interval.

## Root cause
The `ReadyObserved` latch added in #271 (`internal/controller/runner/refresh.go:461`) closes only when the **reconciler** sees `state=Ready` — either via a fresh observation or via the persisted state being `Ready` at tick time.

`-a --rm` calls `KillCell` synchronously from the CLI on attach-loop return (#274, `cmd/kuke/run/run.go:267`). `KillCell` flips persisted state `Ready → Stopped` at `internal/controller/runner/kill.go:297` **without touching `ReadyObserved`**. If this happens before the first reconciler tick after `StartCell` (which is the common case for short-lived sessions or any session under the 30s interval), `ReadyObserved` is still false on disk.

Subsequent reconciler ticks then see `originalState=Stopped`, `prior=false`, `newState=Stopped` → `latchReadyObserved` returns false → `shouldAutoDeleteCell` returns false → the cell is never reaped.

This is effectively a regression of #265, narrowed to the case where the create→detach window is shorter than the reconcile interval.

## Suggested fix
Close the latch eagerly anywhere the runner writes `state=Ready` synchronously — at minimum:

- `internal/controller/runner/provision.go:1029` (provisionNewCell)
- `internal/controller/runner/start.go:228` (idempotent skip path)
- `internal/controller/runner/start.go:640` (after StartCell finishes)
- `internal/controller/runner/start.go:901` (StartContainer's Ready write)

Set `cell.Status.ReadyObserved = true` next to each `Status.State = CellStateReady` assignment. Semantics match the latch's intent: the runner just observed the cell as Ready, so the latch should close immediately, regardless of whether the reconciler ever sees it.

Alt (smaller patch, narrower fix): set `ReadyObserved=true` inside `KillCell` next to the `Status.State = CellStateStopped` write — by the time `KillCell` runs, the cell was Ready by definition. Less semantically faithful but one-line and covers the failing flow.

eminwux:
## Summary
- Close the `ReadyObserved` latch alongside every synchronous `state=Ready` write so `KillCell` racing the first reconciler tick can no longer strand a `--rm` cell in `Stopped`.
- Centralized the invariant in a `markCellReady(*Cell)` helper next to `latchReadyObserved` in `refresh.go`. Five call sites (`provisionNewCell`, two `StartCell` paths, `StartContainer`, `RecreateCell`) now go through the helper instead of writing only `Status.State`.
- Added two unit tests: `TestMarkCellReady` pins the helper contract; `TestAutoDeleteSurvivesKillCellRace` models the on-disk handoff between the synchronous Ready write, a racing `KillCell`, and the reconciler's `shouldAutoDeleteCell` predicate.
- Took the broader fix from the issue's "Suggested fix" rather than the alt one-liner in `KillCell`: the latch's intent is "the runner has just observed Ready," so closing it next to each Ready write is semantically faithful and also covers any future caller that flips state to `Stopped` without going through `KillCell`. Caught a 5th synchronous Ready write in `RecreateCell` that the issue did not list — same race shape, applied the same fix.

## Test plan
- [x] `make dev-init` — daemon-parity tail matches the regression guard from `CLAUDE.md` (both `kuke get realms` and `kuke get realms --no-daemon` show `default Ready` and `kuke-system Ready`).
- [x] End-to-end repro from the issue: `kuke run -a --rm -f repro-275.yaml --container work` with attach exiting in <5s. Pre-write metadata showed `state=2 (Stopped)` and `readyObserved=true` (the new invariant on disk); after one reconcile interval the cell directory `/opt/kukeon/default/default/default/repro-275` was reaped and `kuke get cells` was clean. Pre-fix, the same repro left `readyObserved=false` and the cell stuck in `Stopped` indefinitely (verified by inspecting the orphan `kukeon-dev-cb9290` cell from the issue's repro session before purging it).

Closes #275

### DIFF
--- a/internal/controller/runner/provision.go
+++ b/internal/controller/runner/provision.go
@@ -1025,8 +1025,7 @@ func (r *Exec) provisionNewCell(cell intmodel.Cell) (intmodel.Cell, error) {
 		return intmodel.Cell{}, err
 	}
 
-	// Update cell state in internal model
-	cell.Status.State = intmodel.CellStateReady
+	markCellReady(&cell)
 
 	// Update cell metadata
 	if err = r.UpdateCellMetadata(cell); err != nil {
--- a/internal/controller/runner/recreate_cell.go
+++ b/internal/controller/runner/recreate_cell.go
@@ -148,8 +148,7 @@ func (r *Exec) RecreateCell(desired intmodel.Cell) (intmodel.Cell, error) {
 		return intmodel.Cell{}, fmt.Errorf("failed to recreate cell containers: %w", err)
 	}
 
-	// Update cell state
-	desired.Status.State = intmodel.CellStateReady
+	markCellReady(&desired)
 
 	// Update metadata
 	if updateErr := r.UpdateCellMetadata(desired); updateErr != nil {
--- a/internal/controller/runner/refresh.go
+++ b/internal/controller/runner/refresh.go
@@ -472,6 +472,18 @@ func shouldAutoDeleteCell(autoDelete bool, newState intmodel.CellState, readyObs
 	return autoDelete && cellStateAutoDeleteTriggers(newState) && readyObserved
 }
 
+// markCellReady stamps a synchronous Ready transition: state and the
+// ReadyObserved latch must close together. Without the eager latch
+// close, a KillCell that races the first reconciler tick (e.g. `kuke
+// run -a --rm` exiting attach within the reconcile interval) flips
+// persisted state Ready → Stopped before any reconciler observation,
+// leaving readyObserved=false on disk; subsequent ticks then see
+// originalState=Stopped and shouldAutoDeleteCell never fires.
+func markCellReady(cell *intmodel.Cell) {
+	cell.Status.State = intmodel.CellStateReady
+	cell.Status.ReadyObserved = true
+}
+
 // deriveCellStateFromRootContainer resolves the cell's state from the root
 // container's task. Returns Ready when no root container is configured (a
 // cgroup-only cell counts as Ready by cgroup existence).
--- a/internal/controller/runner/refresh_test.go
+++ b/internal/controller/runner/refresh_test.go
@@ -123,6 +123,65 @@ func TestLatchReadyObserved(t *testing.T) {
 	}
 }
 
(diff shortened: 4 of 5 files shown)


eminwux:
**PR #276 Review — `fix: close ReadyObserved latch eagerly on synchronous Ready writes`**

Tight, semantically right fix: the latch now tracks each synchronous Ready write atomically, and `markCellReady` centralizes the invariant so future Ready writers can't reintroduce the race. `TestAutoDeleteSurvivesKillCellRace` models exactly the on-disk handoff #275 described, and catching the 5th writer in `RecreateCell` is a nice find. No correctness, leak, or convention concerns; test plan covers `make dev-init` and the end-to-end repro.

One non-blocking observation:

- **Latent latch clobber** `internal/controller/runner/refresh.go:188-225`: `RefreshCell` (the `kuke refresh` CLI path, not the daemon reconciler) builds a fresh `CellStatus` that omits `ReadyObserved` and assigns it back wholesale on a state change, wiping the latch. Out of scope here — the synchronous-write race is fixed — but a `kuke refresh` racing between a Ready write and the daemon's first tick can still defeat AutoDelete via the same mechanism. Filing as a follow-up.

LGTM to merge.

PR #280, opened 2026-05-03
fix: keep cell alive on clean detach under run -a --rm

### LINKED ISSUE #279: kuke run -a --rm: clean ^]^] detach kills the cell instead of leaving it alive for re-attach
## Repro

Reproduced on `main` at `edddffe` with the `kukeon-dev` profile (root: `sleep infinity`; attachable peer: `/bin/bash`).

1. Terminal A: `sudo ./kuke run -a --rm -p kukeon-dev` — attaches into the `work` shell.
2. Terminal B: `sudo ./kuke attach --realm default --space default --stack default --cell kukeon-dev-<id> --container work`, then press `Ctrl+] Ctrl+]` to detach. Cell stays Ready, attach loop exits 0 — fine.
3. Terminal A: press `Ctrl+] Ctrl+]` to detach.
4. `sudo ./kuke get cells` — cell is **gone**.

Expected: detach leaves the cell alive (operator can re-attach later). Only `exit` of the workload shell, peer hangup, or workload crash should trigger `--rm` cleanup.

## Observed

`kuke run -a --rm`'s SIGKILL+delete fires on every clean detach, identical to a real shell exit. A second attacher (`kuke attach`) detaching with the exact same keystrokes is unaffected.

## Root cause

Two cooperating layers:

1. **`cmd/kuke/run/attach.go:149-156` (`runAttachLoop`)** returns the raw error from `attach.Run` directly. sbsh's `pkg/attach.Run` reports a clean `^]^]` detach as `\"close requested: read/write routines exited\"` (sbsh's `internal/client/clientrunner/io.go:131` emits the same error event for both detach and peer-hangup; sbsh has no public sentinel for `EvDetach`). The peer command, `cmd/kuke/attach/attach.go:137-167` (`isCleanDetach`), filters this exact signature back to nil — `runAttachLoop` does not.

2. **`cmd/kuke/run/run.go:247-258` (`attachAndMaybeAutoDelete`)** calls `autoDeleteAfterAttach` → `KillCell` whenever the loop returns, with no detach/exit branch. The unconditional kill was added in commit c6db87e (#274) and is locked in by `TestRun_RmAttach_KillsCellAfterAttachLoopReturns` (run_test.go:1403) and `TestRun_RmAttach_KillsCellEvenWhenAttachLoopErrors` (run_test.go:1445). Those tests encode the regression — they assert kill-on-detach.

The asymmetry: `kuke attach` distinguishes via `isCleanDetach`; `kuke run -a --rm` does not, and additionally fires `KillCell` regardless. The user-visible symptom is detach-acts-like-exit only on the `kuke run -a --rm` invoker.

## Suggested fix

1. Lift `isCleanDetach` from `cmd/kuke/attach/attach.go` into a shared spot (e.g. `cmd/kuke/shared/`) so both attach paths use one classifier.
2. Change `runAttachLoop` (or `attachAfterRun`) in `cmd/kuke/run/` to return a tri-state — e.g. `(detached bool, err error)` — where clean detach yields `(true, nil)` and shell-exit/peer-hangup/error yields `(false, err-or-nil)`.
3. Gate the `--rm` kill in `attachAndMaybeAutoDelete` on `!detached`. Clean detach → leave cell alive (parity with `kuke attach`); exit/hangup/error → keep today's `KillCell` so `sleep infinity` roots don't pin the cell after the workload terminates.
4. Update `TestRun_RmAttach_KillsCellAfterAttachLoopReturns` / `…KillsCellEvenWhenAttachLoopErrors` to assert *no* `KillCell` on clean detach, and add a test that drives the mock `runFn` with `errors.New(\"close requested: read/write routines exited\")` and asserts the cell survives.
5. Refresh the `--rm` flag docstring on `cmd/kuke/run/run.go:92-101` so the trigger is described as \"attach-loop exit/error\" rather than \"attach-loop returning\".

Caveat: until sbsh exposes a public sentinel for `EvDetach`, the disambiguation remains string-match on the error message — same load-bearing hack `kuke attach` already relies on.

eminwux:
## Summary
- Issue #279: `kuke run -a --rm` killed the cell on every clean ^]^] detach, identical to a real workload exit. The kill RPC fired on every attach-loop return because the CLI could not distinguish detach from peer hangup.
- sbsh v0.10.1 ships the public sentinels (`attach.ErrDetached`, `attach.ErrPeerClosed`) the issue's caveat asked for ([sbsh#192](https://github.com/eminwux/sbsh/pull/192)). This PR bumps sbsh to v0.10.1 and switches the classification from the legacy `strings.Contains("close requested")` hack to `errors.Is` against the typed sentinels — no string match remains in either attach path.
- New `cmd/kuke/shared/attach_exit.go` lifts the classifier into one place (`ClassifyAttachExit` / `IsCleanAttachExit`) so `kuke attach` and `kuke run -a` share one source of truth, replacing the per-package `isCleanDetach`.
- `runAttachLoop` now returns `(detached bool, err error)` (issue's suggested tri-state). `attachAndMaybeAutoDelete` gates `KillCell` on `!detached`: clean ^]^] leaves the cell alive (parity with `kuke attach`), peer hangup / shell exit / controller error still fire `KillCell` so a long-lived `sleep infinity` root does not pin the cell.
- `--rm` flag docstring rewritten to reflect the new trigger surface (workload-end, not "any attach-loop return").

## Test plan
- [x] `go vet ./...` — clean
- [x] `make test` — full suite green (including the regression suite below)
- [x] Pre-commit `golangci-lint` — clean (initial run flagged exhaustive switch on the new `AttachExit` enum; fixed before commit)
- [x] New `TestRun_RmAttach_KeepsCellAliveOnCleanDetach` — drives the mock `runFn` with `attach.ErrDetached` and asserts `KillCell` calls = 0 (encodes the #279 fix; flips the previous `KillsCellAfterAttachLoopReturns` regression-encoder)
- [x] New `TestRun_RmAttach_KillsCellOnPeerClosed` — drives with `attach.ErrPeerClosed`, asserts `KillCell` still fires (preserves #265 cleanup contract for workload-end)
- [x] Existing `TestRun_RmAttach_KillsCellEvenWhenAttachLoopErrors` — unchanged: a non-detach controller error still fires `KillCell`
- [x] `TestRun_RmAttach_AlreadyReady_StillKillsCellOnPeerClosed` — idempotent-Ready branch, peer-closed path still cleans up (was `…AfterAttachLoopReturns`)
- [x] `TestRun_RmAttach_KillCellFailureDoesNotMaskAttachExit` / `…NotFound_Silent` — switched to inject peer-closed so KillCell still fires under the new semantics
- [x] New `TestAttach_RunReturnsPeerClosedError` (sister of the existing detach test) — sentinel-based, asserts `kuke attach` exits 0 on either benign session-end signal
- [x] New `cmd/kuke/shared/attach_exit_test.go` — direct table-test coverage of `ClassifyAttachExit` and `IsCleanAttachExit` (nil, raw sentinels, wrapped sentinels, unrelated err)
- [ ] Interactive repro from the issue (Terminal A `sudo ./kuke run -a --rm -p kukeon-dev`, Terminal B attach + ^]^], Terminal A ^]^]; expect cell still listed in `kuke get cells`) — not run; the bug requires a real TTY and two terminals, which the agent cannot drive. Recommend the human reviewer reproduce before merge.
- [ ] `make dev-init` — not run; the change is CLI-only (`cmd/kuke/run`, `cmd/kuke/attach`, `cmd/kuke/shared`) and does not touch the daemon, the build path, or `/opt/kukeon`, so the daemon-parity regression guard does not apply per `CLAUDE.md`.

Closes #279

### DIFF
--- a/cmd/kuke/attach/attach.go
+++ b/cmd/kuke/attach/attach.go
@@ -135,36 +135,16 @@ func runAttach(cmd *cobra.Command, _ []string) error {
 	}
 
 	run := resolveRun(cmd)
-	if runErr := run(cmd.Context(), attach.Options{
+	runErr := run(cmd.Context(), attach.Options{
 		SocketPath: result.HostSocketPath,
 		Stdin:      os.Stdin,
 		Stdout:     os.Stdout,
 		Stderr:     os.Stderr,
-	}); runErr != nil && !isCleanDetach(runErr) {
+	})
+	if runErr != nil && !kukeshared.IsCleanAttachExit(runErr) {
 		return runErr
 	}
 	return nil
-}
-
-// isCleanDetach reports whether err returned by the in-process attach
-// loop describes a normal session end rather than an unrecoverable
-// failure. The Ctrl+] Ctrl+] detach filter inside sbsh's pkg/attach
-// triggers a controller-level Detach RPC; the embedded read/write
-// goroutines then exit cleanly, but their teardown is reported as the
-// error "close requested: read/write routines exited" — sbsh does not
-// expose a sentinel for either layer in its public pkg/errors. The
-// remote terminal closing the connection (workload exited, peer hung
-// up) ends up on the same path, which from `kuke attach`'s perspective
-// is also a benign session end. Surface a clean exit code 0 in those
-// cases; any other failure (socket missing, RPC error, context done)
-// arrives without this signature and still bubbles up.
-func isCleanDetach(err error) bool {
-	if err == nil {
-		return false
-	}
-	msg := err.Error()
-	return strings.Contains(msg, "close requested") &&
-		strings.Contains(msg, "read/write routines exited")
 }
 
 // pickAttachableContainer enumerates the cell's containers and returns the
--- a/cmd/kuke/attach/attach_test.go
+++ b/cmd/kuke/attach/attach_test.go
@@ -19,6 +19,7 @@ package attach_test
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
 	"testing"
 
@@ -278,12 +279,12 @@ func TestAttach_ExplicitContainer_Attachable_Succeeds(t *testing.T) {
 	}
 }
 
-// TestAttach_RunReturnsCleanDetachError covers the regression from #147:
-// when the user presses Ctrl+] Ctrl+], sbsh's controller fires the
-// Detach RPC and the embedded read/write copy goroutines exit, but that
-// teardown surfaces as "close requested: read/write routines exited"
-// from pkg/attach.Run. The wrapper must recognise that signature as a
-// benign session end and return nil so `kuke attach` exits 0.
+// TestAttach_RunReturnsCleanDetachError covers the regression from
+// #147: when the user presses Ctrl+] Ctrl+], sbsh's controller fires
+// the Detach RPC and pkg/attach.Run returns attach.ErrDetached
+// (sbsh#192, available since sbsh v0.10.1). The wrapper must recognise
+// the public sentinel as a benign session end and return nil so `kuke
+// attach` exits 0.
 func TestAttach_RunReturnsCleanDetachError(t *testing.T) {
 	t.Cleanup(viper.Reset)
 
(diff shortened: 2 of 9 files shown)


eminwux:
**PR #280 Review — `fix: keep cell alive on clean detach under run -a --rm`**

Tight fix. The tri-state `(detached, err)` cleanly separates the three exit conditions, and the new `ClassifyAttachExit` / `IsCleanAttachExit` helpers replace the brittle `strings.Contains` matcher with `errors.Is` against typed sentinels. Unit coverage encodes both the new contract (clean detach skips `KillCell`) and the preserved behavior (peer-closed and controller errors still fire `KillCell`). License headers and conventions in order. No correctness, secret-leak, or dead-symbol concerns.

A small non-blocking nit:

- **Doc nit** `cmd/kuke/shared/attach_exit.go:52-54`: "nil and detach map to AttachExitDetached / AttachExitPeerClosed respectively" parses ambiguously — three inputs (nil, `ErrDetached`, `ErrPeerClosed`) map to two enum values, so "respectively" doesn't quite work. Consider: "nil and `ErrDetached` map to `AttachExitDetached`; `ErrPeerClosed` maps to `AttachExitPeerClosed`".

LGTM — the interactive two-terminal repro from the test plan is the right call to defer to a human reviewer; the unit-level coverage of both branches is solid and `make dev-init` correctly does not apply here (CLI-only change).

PR #281, opened 2026-05-04
build: mark kuke and kukeond targets PHONY

### LINKED ISSUE #277: build: 'make kuke' is a no-op when ./kuke already exists, masking stale-binary bugs in dev-init
## Repro
On `main` (c6db87e), with a stale `./kuke` binary on disk (e.g. left over from a prior session that predates a recent CLI command):

```bash
ls -la ./kuke   # binary from before the change you want to test
make kuke       # silently does nothing — no rebuild, no error
./kuke <a-recently-added-subcommand>   # fails: "unknown command"
```

Hit this session: `./kuke` was from 2026-04-30 (before `kuke daemon` was added). `make dev-init` ran `make kuke` as its first step (per `scripts/dev-init.sh:38`), the host-side rebuild was a no-op, and the next step `sudo ./kuke daemon reset` (line 49) errored with `unknown command "daemon" for "kuke"` and aborted the whole bootstrap. `make clean kuke` rebuilt fine and the rest of `make dev-init` succeeded.

## Observed
```
==> Reset prior kukeond cell
Error: unknown command "daemon" for "kuke"
Run 'kuke --help' for usage.
make: *** [Makefile:96: dev-init] Error 1
```

with `ls -la ./kuke` showing a binary timestamp predating the `kuke daemon` command.

## Root cause
`Makefile:42` declares the `kuke` target without `.PHONY` and without source-file prerequisites:

```make
kuke:
	go build \
	-o kuke \
	-ldflags="$(LDFLAGS)" \
	./cmd/
```

GNU Make treats `kuke` as a file target. When the file `./kuke` exists and there are no prerequisites listed, make considers the target up-to-date and skips the recipe — so `make kuke` is silently a no-op. Same shape applies to the `kukeond` target (`Makefile:46`), which depends on `kuke` and only runs `ln -sf`.

This is benign for incremental dev where you remember to `make clean` first, but `scripts/dev-init.sh`'s first step is `make kuke` precisely so contributors don't need that knowledge. Today, the script silently keeps a stale binary and surfaces the divergence as a confusing later error in a different step.

## Suggested fix
Add `.PHONY: kuke kukeond` to the Makefile so `make kuke` always runs `go build` (Go's own build cache makes the rebuild fast when nothing changed). Alternative: declare the `cmd/`, `internal/`, `pkg/` Go sources as prerequisites so make rebuilds only on real changes — more correct but more verbose. The PHONY approach matches how the rest of the Makefile already treats `release`, `release-build`, `clean`, `kill`, `test-e2e`, etc. as phony.

eminwux:
## Summary
- `make kuke` was a silent no-op when `./kuke` existed because the target had no `.PHONY` and no source-file prerequisites — GNU Make treated it as up-to-date and skipped the recipe. Same shape for `kukeond`.
- This masked stale-binary bugs in `scripts/dev-init.sh`, whose first step is `make kuke`; a stale `./kuke` would survive into later steps and surface as a confusing "unknown command" error in a different phase of the bootstrap.
- Fix matches the issue's suggested approach: declare both targets `.PHONY`, matching how `release`, `test-e2e`, `dev-init`, etc. are already treated. Go's build cache makes the unconditional rebuild fast when nothing changed.

## Test plan
- [x] Reproduced the bug pre-fix style and confirmed the fix: built `./kuke`, backdated its mtime to 2024-01-01, ran `make kuke` — mtime advanced to today (post-fix `make kuke` rebuilds instead of skipping).
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (the target CI runs in `.github/workflows/test.yaml`)
- [ ] `make dev-init` not run — requires sudo + containerd + docker; this change only affects whether the Makefile's first step actually rebuilds, which the reproduction above verifies directly.

Closes #277

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,7 @@ all: clean kill $(BINS)
 .PHONY: release
 release: release-build
 
+.PHONY: kuke kukeond
 kuke:
 	go build \
 	-o kuke \


eminwux:
**PR #281 Review — `build: mark kuke and kukeond targets PHONY`**

Tiny, correct fix. The `.PHONY` declaration brings `kuke`/`kukeond` in line with how `release`, `test-e2e`, and `dev-init` are already treated, and the PR description nails the root cause — GNU Make was treating the existing binary as up-to-date because the target had no source-file prerequisites. The mtime-backdating reproduction verifies the fix directly without needing the full `make dev-init` loop, which is the right call given this only affects whether the Makefile rebuilds (no daemon/`/opt/kukeon` semantics change).

LGTM.

PR #282, opened 2026-05-04
test: skip cni bootstrap defaults subtest when /opt/cni isn't writable

### LINKED ISSUE #264: cni: TestBootstrapCNI 'uses defaults' subtest fails for non-root go test runs (mkdir /opt/cni: permission denied)
## Repro

On `origin/main` (commit `ee19939`), in any user-owned checkout where `/opt/cni` is not writable by the current user (i.e. any normal developer machine running `go test` without root):

```bash
git clone https://github.com/eminwux/kukeon.git /tmp/kukeon-clean
cd /tmp/kukeon-clean
go test ./internal/cni/...
```

## Observed

```
--- FAIL: TestBootstrapCNI (0.00s)
    --- FAIL: TestBootstrapCNI/success_with_empty_directories_(uses_defaults) (0.00s)
        bootstrap_test.go:233: BootstrapCNI() error = mkdir /opt/cni: permission denied, want nil
FAIL    github.com/eminwux/kukeon/internal/cni  0.051s
```

The other subtests (which thread tempdirs through their `setup`) pass. Only the `success with empty directories (uses defaults)` case fails, and it always fails for unprivileged users because `BootstrapCNI("","","")` resolves the empty args to the hardcoded `/opt/cni/{net.d,cache,bin}` host paths, then tries to `mkdir` them.

## Root cause

`internal/cni/bootstrap_test.go` `TestBootstrapCNI` has a subtest at line ~70 (`"success with empty directories (uses defaults)"`) whose `setup` returns `""`/`""`/`""`, so `BootstrapCNI` falls through to the production defaults. The assertion block then asserts on `/opt/cni/net.d`, `/opt/cni/cache`, and `/opt/cni/bin`, so the test cannot just substitute a tempdir — the whole point of the case is that the defaults are `/opt/cni/...`.

This makes the "exercise the production-default path" intent unrunnable in any unprivileged environment (every developer laptop, every non-root CI runner, every sandbox).

## Suggested fix

Two paths, either is reasonable:

1. **Skip when `/opt/cni` is unwritable.** Add a `t.Skip` guard at the top of the subtest using a writability probe so the subtest still runs as a tripwire under root and silently skips for everyone else. Cheapest, preserves the documentation value of the subtest.
2. **Split the assertion from the side effect.** Refactor `BootstrapCNI` (or expose a sibling) so the default-resolution step is observable without `mkdir`-ing the host paths, then assert on the resolved paths from a unit test that does no I/O. More work but doesn't require a "skip-if-unprivileged" branch.

Option 1 is the lower-blast-radius fix and matches the same pattern used in other root-only tests in the repo (e.g. controller integration tests gated by build tag).

eminwux:
## Summary
- `TestBootstrapCNI/success_with_empty_directories_(uses_defaults)` was unrunnable on any unprivileged host because `BootstrapCNI(\"\",\"\",\"\")` resolves to the hardcoded `/opt/cni/{net.d,cache,bin}` defaults and then `mkdir`s them — a tempdir substitution would defeat the purpose of the subtest (asserting the production-default path).
- Added a `t.Skip` guard at the top of the subtest using a writability probe (try `MkdirAll(/opt/cni)` then `CreateTemp` inside it). On root or privileged CI the tripwire still fires; on unprivileged dev/CI the subtest silently skips. Matches issue #264 Option 1, the lower-blast-radius fix.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/cni/... -v -run TestBootstrapCNI` as unprivileged user — defaults subtest now `SKIP`, all others `PASS`
- [x] `sudo go test ./internal/cni/... -v -run "TestBootstrapCNI/success_with_empty_directories"` as root — subtest still `PASS` (tripwire preserved)
- [ ] `make dev-init` — not run; this PR is test-file-only and does not touch the daemon, build path, or anything under `/opt/kukeon`

Closes #264

### DIFF
--- a/internal/cni/bootstrap_test.go
+++ b/internal/cni/bootstrap_test.go
@@ -69,7 +69,22 @@ func TestBootstrapCNI(t *testing.T) {
 		},
 		{
 			name: "success with empty directories (uses defaults)",
-			setup: func(_ *testing.T, _ string) (string, string, string) {
+			setup: func(t *testing.T, _ string) (string, string, string) {
+				// This subtest exercises the production-default path —
+				// BootstrapCNI("","","") resolves to /opt/cni/{net.d,cache,bin},
+				// so it cannot be substituted with a tempdir without losing
+				// the tripwire. Skip on hosts where /opt/cni isn't writable
+				// (typical of non-root `go test` runs); under root or in
+				// privileged CI the subtest still runs.
+				if err := os.MkdirAll("/opt/cni", 0o750); err != nil {
+					t.Skipf("skip: /opt/cni not creatable by current user: %v", err)
+				}
+				probe, err := os.CreateTemp("/opt/cni", ".bootstrap-probe-*")
+				if err != nil {
+					t.Skipf("skip: /opt/cni not writable by current user: %v", err)
+				}
+				probe.Close()
+				_ = os.Remove(probe.Name())
 				return "", "", ""
 			},
 			wantReport: func(t *testing.T, report cni.BootstrapReport) {


eminwux:
**PR #282 Review — `test: skip cni bootstrap defaults subtest when /opt/cni isn't writable`**

Surgical, test-only fix that matches issue #264 Option 1. Probe correctly covers both branches: `MkdirAll` failing (no `/opt/cni` and can't create) and `CreateTemp` failing (`/opt/cni` exists root-owned). Tripwire intent preserved under root/CI. No sensitive data, no dead code, conventions clean. Test plan well-justified for skipping `make dev-init` (test-file-only).

LGTM to merge.

PR #286, opened 2026-05-05
refactor: invert kuke run attach default — drop -a, add -d/--detach

### LINKED ISSUE #283: cli: invert kuke run attach default — drop -a/--attach, add -d/--detach
## Context

`kuke run` currently runs detached by default and only attaches to the cell's
attachable terminal when the operator passes `-a/--attach` (cmd/kuke/run/run.go:86).
This is the opposite of the convention every operator already knows from
`docker run` (foreground by default, `-d` to detach) and from `kubectl run -it`
(interactive opt-in via flags rather than a separate `-a`). The mismatch is a
small but recurring source of friction: people reach for `-a` because that's
what kuke documents, then have to context-switch back when they're using docker
or kubectl in the next terminal.

The flag was added in phase 2a as an opt-in because the attach loop was new and
risky. It's now the common path for interactive workloads, so the default
should reflect that.

## Proposal

Flip the default. After this change:

- `kuke run -f foo.yaml` (or `-p <profile>`) **attaches** to the cell's
  attachable container after start — same loop the current `-a` invokes.
- `kuke run -f foo.yaml -d` (or `--detach`) returns immediately after start
  without attaching — the current default behavior.
- `-a/--attach` is removed entirely (no deprecation alias). The CLI is
  pre-1.0; one breaking change is cleaner than carrying two flags forever.
- `--container` continues to select which attachable container to join,
  but is now valid in the default (attached) mode and rejected with `-d`.
- `--output json|yaml` remains incompatible with attach and is now rejected
  unless `-d` is passed (mirrors today's `--output incompatible with -a`
  rule, just inverted).
- `--rm` semantics are unchanged. The help text mentions "Without -a, the
  trigger is the root container's task exit. With -a, the trigger is the
  attach loop exiting…" — rewrite to reference `--detach` / default-attach
  instead.

## Acceptance criteria

- [ ] `cmd/kuke/run/run.go` registers `-d/--detach` (bool) and removes
      `-a/--attach`. The viper key renames from `KUKE_RUN_ATTACH` to
      `KUKE_RUN_DETACH` (or equivalent), with corresponding `cmd/config`
      entries updated.
- [ ] `parseRunFlags` inverts the boolean: `doAttach` becomes
      `!detach` (or rename to `detach`/`doDetach` and flip call sites in
      `runRun` and `cmd/kuke/run/attach.go`).
- [ ] Validation rules updated: `--container` rejected only when `-d` set;
      `--output` rejected only when `-d` not set; `--rm` + `--no-daemon`
      rule unchanged.
- [ ] Package doc (`cmd/kuke/run/run.go:17-24`) and the `--rm` long help
      text rewritten to describe attach-as-default, `-d` as opt-out.
- [ ] `cmd/kuke/run/run_test.go` updated: the `-a/--attach` flag-presence
      assertion (run_test.go:1043-1045) becomes a `-d/--detach` assertion.
      All test cases that drive attach behavior drop the `-a` flag; cases
      that test detached-return add `-d`.
- [ ] Smoke test in PR description: `make dev-init` passes the daemon-parity
      check, and a manual `kuke run -f <profile>` (no flags) drops into the
      attachable terminal; `kuke run -f <profile> -d` returns immediately.

## Notes

- Breaking CLI change. Worth a `BREAKING:` note in the commit body and a
  release-notes line when the next tag cuts.
- Single area (`cmd/kuke/run`), single behavioral flip + flag rename + help
  rewrite + test updates. Fits one reviewable PR.
- `kuke attach <cell>` (the standalone verb at `cmd/kuke/attach/attach.go`)
  is unaffected — this issue only changes the post-start behavior of
  `kuke run`.
- Follow-up doc issue (separate `documentation` filing) if any user-facing
  guide ends up mentioning `kuke run -a` after this lands. Current grep of
  `docs/` and `README.md` shows no such reference, so the in-source help
  text rewrite above should be the full doc surface.

eminwux:
## Summary
- Flip `kuke run` so the post-start attach is the default and `-d/--detach` is the opt-out — matches `docker run` and `kubectl run -it`, removes the recurring context-switch friction.
- Drop `-a/--attach` outright (no deprecation alias); pre-1.0 CLI, one breaking change is cleaner than carrying both flags.
- Rename viper key `KUKE_RUN_ATTACH` → `KUKE_RUN_DETACH` and invert validation: `--container` rejected with `-d`, `--output` rejected without `-d`.

BREAKING: `kuke run -f <file>` (or `-p <profile>`) now attaches by default; pass `-d/--detach` to keep the previous detached behavior. The `-a/--attach` flag is removed.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./cmd/kuke/run/...` passes (test cases inverted; `-a` dropped from attach cases, `-d` added to detached cases, new `TestRun_Attach_ContainerWithDetachFlag_Errors` and renamed `TestNewRunCmd_DetachFlagRegistered` / `TestRun_RmDetach_DoesNotCallKillCell`)
- [x] `make test` (full unit/integration suite excluding e2e) all packages pass
- [x] `make dev-init` daemon-parity check passes (both `kuke get realms` and `kuke get realms --no-daemon` print the documented `default` + `kuke-system` Ready table)
- [x] Manual smoke `sudo ./kuke run -f /tmp/cell.yaml -d` returns immediately after start (rc=0, no terminal handover)
- [x] Manual smoke `sudo ./kuke run -f /tmp/cell.yaml </dev/null` (no flags) drops into the sbsh attach loop (`To detach, press ^] twice` banner printed; loop exits when stdin EOFs)
- [x] Manual smoke `sudo ./kuke run -f /tmp/cell.yaml -d --container work` rejected with `--container is incompatible with -d/--detach`
- [x] `sudo ./kuke run -f cell.yaml -a` rejected as unknown shorthand flag (no deprecation alias)

## Acceptance criteria
- [x] `cmd/kuke/run/run.go` registers `-d/--detach` and removes `-a/--attach`; viper key renamed `KUKE_RUN_ATTACH` → `KUKE_RUN_DETACH` in `cmd/config/env.go`.
- [x] `parseRunFlags` field renamed `doAttach` → `detach`; `runRun` and the already-Ready short-circuit invert their checks (`flags.doAttach` → `!flags.detach`).
- [x] Validation rules inverted: `--container` rejected only when `-d` set; `--output` rejected only when `-d` not set; `--rm` + `--no-daemon` rule unchanged.
- [x] Package doc and `--rm` long help rewritten to describe attach-as-default and `-d` as opt-out. Surrounding doc comments in `attach.go`, `cmd/kuke/shared/attach_exit.go`, `pkg/api/model/v1beta1/cell.go`, and `internal/controller/runner/refresh_test.go` updated to drop now-stale `kuke run -a` references.
- [x] `cmd/kuke/run/run_test.go` updated: `-a` flag-presence assertion now checks `-d/--detach`; attach-driven cases drop `-a` (attach is default), detached cases add `-d`, and `--container` validation test rewritten to assert the new `-d` + `--container` rejection.
- [x] Smoke check above.

Closes #283

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -362,7 +362,7 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_RUN_STACK = DefineKV("KUKE_RUN_STACK", "kuke/run/stack", "default")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
-	KUKE_RUN_ATTACH = DefineKV("KUKE_RUN_ATTACH", "kuke/run/attach")
+	KUKE_RUN_DETACH = DefineKV("KUKE_RUN_DETACH", "kuke/run/detach")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_RUN_CONTAINER = DefineKV("KUKE_RUN_CONTAINER", "kuke/run/container")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
--- a/cmd/kuke/run/attach.go
+++ b/cmd/kuke/run/attach.go
@@ -40,8 +40,8 @@ type MockRunKey struct{}
 // detach / context cancel and any unrecoverable controller error otherwise.
 type runFn func(ctx context.Context, opts attach.Options) error
 
-// pickAttachTarget resolves the container `kuke run -a` should attach to,
-// applying the documented precedence:
+// pickAttachTarget resolves the container the post-start attach should
+// connect to, applying the documented precedence:
 //
 //  1. explicit --container flag, when set
 //  2. cell.tty.default, when the spec sets it
@@ -85,7 +85,7 @@ func pickAttachTarget(spec v1beta1.CellSpec, cellName, explicit string) (string,
 	}
 
 	return "", fmt.Errorf(
-		"%w (cell %q); declare 'attachable: true' on one container or omit -a",
+		"%w (cell %q); declare 'attachable: true' on one container or pass -d/--detach",
 		errdefs.ErrAttachNoCandidate, cellName,
 	)
 }
@@ -108,18 +108,17 @@ func formatAttachableList(spec v1beta1.CellSpec) string {
 
 // runAttachLoop resolves the per-container sbsh control socket via the
 // daemon's AttachContainer RPC and drives the in-process attach loop.
-// Returns the tri-state used by the -a --rm cleanup decision in
+// Returns the tri-state used by the --rm cleanup decision in
 // attachAndMaybeAutoDelete:
 //
 //   - detached=true,  err=nil — operator pressed ^]^] (or peer issued
 //     a Detach RPC). Cell must stay alive.
 //   - detached=false, err=nil — peer dropped the connection (workload
-//     exited / hung up). Surface exit 0; -a --rm fires KillCell so a
+//     exited / hung up). Surface exit 0; --rm fires KillCell so a
 //     long-lived root does not pin the cell.
 //   - detached=false, err≠nil — pre-attach setup error or unrecoverable
-//     controller error. Surface to the user; -a --rm still fires
-//     KillCell because a half-detached session would otherwise leak
-//     the cell.
+//     controller error. Surface to the user; --rm still fires KillCell
+//     because a half-detached session would otherwise leak the cell.
 func runAttachLoop(
 	cmd *cobra.Command,
 	client kukeonv1.Client,
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -14,13 +14,14 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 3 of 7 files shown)


eminwux:
**PR #286 Review — `refactor: invert kuke run attach default — drop -a, add -d/--detach`**

Clean, well-scoped CLI inversion. The flag rename (`-a/--attach` → `-d/--detach`), viper key rename (`KUKE_RUN_ATTACH` → `KUKE_RUN_DETACH`), `runFlags.doAttach` → `detach`, and the two validation inversions (`--container` rejected with `-d`; `--output` rejected without `-d`) line up consistently across `parseRunFlags`, both call sites in `runRun`, the package doc, the `--rm` long help, and surrounding doc comments in `attach.go`, `attach_exit.go`, `cell.go`, and `refresh_test.go`. Tests flip in lockstep — `-a` removed from attach-driven cases, `-d` added to detached cases, the rejected-combination test rewritten for the new direction, and `TestNewRunCmd_DetachFlagRegistered` now asserts `-d/--detach`. No secret/PII surface; pre-1.0, no deprecation alias matches the project's stated policy in the PR description. Test plan covers daemon parity and the three manual smokes (attach default, `-d` detached, `-d --container` rejection).

LGTM.

PR #299, opened 2026-05-05
fix: install KUKEON-FORWARD admission chain at init

### LINKED ISSUE #293: init: kuke init does not install FORWARD-chain admission rules for kukeon bridges; cells lose egress on hosts where FORWARD policy is DROP
## Repro

On a host where the kernel FORWARD chain default policy is `DROP` (the typical state after Docker has been installed at any point, after firewalld, ufw, or some hardened distro defaults), the user follows the documented bring-up:

1. `kuke init` succeeds.
2. `kuke create space …` succeeds.
3. `kuke run -a alpine -- ping -c2 1.1.1.1` — the cell starts, has an IP on the `k-<8hex>` bridge, but every egress packet is dropped silently on the host. DNS also fails for the same reason.

The only host change required to make egress work is:

```bash
sudo iptables -P FORWARD ACCEPT
```

After that, with no kukeon-side change, cells reach the internet and resolve DNS. Reverting to `-P FORWARD DROP` reproduces the failure.

## Observed

- `iptables -t nat -S POSTROUTING | grep 10.88` shows the bridge plugin's `MASQUERADE` rule is present (so SNAT is fine).
- `iptables -S FORWARD` shows policy `DROP` and no rule that admits traffic from `k-+` interfaces.
- Cell-side `ip route` and `ip a` are correct: it has a default route via the bridge gateway and a usable address from the conflist subnet (`10.88.x.0/24`).
- `kukeond` reports no error — the cell is `Running`, just unreachable.

The netpolicy enforcer at `internal/netpolicy/rules.go:53-87` does insert a `KUKEON-EGRESS` chain at FORWARD position 1, but:

1. With no `EgressPolicy` set on the space (the default), the enforcer installs **zero** rules — see `internal/netpolicy/policy.go:82-83` where `FromInternal(..., nil)` short-circuits to nil.
2. Even with a policy set, the terminal action is `-j RETURN` (`rules.go:79-83`), which falls back to FORWARD's default policy. If that policy is `DROP`, the packet still dies after `RETURN`.

So nothing in kukeon ensures that traffic to/from `k-*` bridges is ever admitted past FORWARD when the host policy is restrictive.

## Root cause

`kuke init` is not self-sufficient with respect to host firewall state. It assumes the FORWARD default policy is permissive, which is not true on hosts where Docker/firewalld/ufw have run, nor on some hardened distros. There is no `kukeon-managed FORWARD admission` chain analogous to Docker's `DOCKER-USER` / `DOCKER-FORWARD` pair.

The result is a silent failure mode on otherwise-correctly-configured hosts: every prerequisite from `docs/site/install/prerequisites.md` is met, kukeon's own bootstrap runs cleanly, but cells still cannot egress because no host-side ingress/egress admission was installed.

## Suggested fix

`kuke init` should install a kukeon-owned FORWARD admission set, idempotently, scoped to kukeon bridges by interface-name prefix. Concretely:

1. **Create a kukeon-owned chain** (e.g. `KUKEON-FORWARD`) jumped to from `FORWARD` at position 1. Idempotent: only insert if not already the first rule.

2. **Inside that chain, admit kukeon bridge traffic explicitly:**

   ```
   -A KUKEON-FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
   -A KUKEON-FORWARD -i k-+ -j ACCEPT
   -A KUKEON-FORWARD -o k-+ -j ACCEPT
   ```

   The `k-+` wildcard matches any kukeon-created bridge regardless of the per-space hash suffix (matches the naming scheme in `internal/cni/config.go:24`).

3. **Do not** modify the FORWARD chain's default policy — that is host-global state and not kukeon's to set. Scoped `-j ACCEPT` rules that only match kukeon bridges leave non-kukeon traffic governed by whatever the host operator chose.

4. **Do not** depend on Docker's `DOCKER-USER` chain or any other third-party iptables structure. Kukeon owns its admission and coexists rather than nesting.

5. **Idempotency:** every rule install is `iptables -C` (check) before `-I`/`-A`, mirroring the pattern already used in `internal/netpolicy/enforcer.go`. Re-running `kuke init` on a healthy host produces no rule churn.

6. **Teardown:** `kuke daemon reset --purge-system` should remove the `KUKEON-FORWARD` chain it owns. The enforcer's existing chain-removal pattern (`enforcer.go:142-147`) is a good template.

7. **Conditional skip:** if the host has loaded `nftables` as the primary firewall backend (no `iptables-legacy`/`iptables-nft` symlink to a working binary), surface a clear error rather than silently failing — same diagnostic story as the missing-CNI-binary case would be.

## Open design questions for the picker

- **Scope: init-time vs. space-create time.** Drafted as init-time (one-shot host setup, simpler). Alternative is to install per-bridge admission at space-create — narrower blast radius, but rule churn proportional to space count. Init-time matches the "kuke init must be clean and add all it needs" framing.
- **`k-+` wildcard vs. exact bridge names.** Wildcard means kukeon admits anything named `k-*` even if a non-kukeon bridge happens to share the prefix. Exact names means rule churn per space. Wildcard chosen for simplicity, mirroring how Docker matches its own bridges.

## Test plan suggestion

- New integration check in `scripts/dev-init.sh` (or a separate smoke script) that runs on a host with `iptables -P FORWARD DROP` set, verifies `kuke init` installs the admission chain, runs `kuke run`, and asserts `ping -c1 1.1.1.1` from inside the cell succeeds.
- Unit test for the chain builder that asserts the expected rule order and the `k-+` interface match.
- Idempotency test: run the installer twice, verify rule count is identical and no duplicates exist.

## Why this is priority:A

Silent loss of cell egress on a non-trivial fraction of hosts (anywhere Docker has been installed). The failure is invisible at default log level, the symptom inside the cell is "DNS doesn't work" or "packets disappear," and the workaround (`-P FORWARD ACCEPT`) is a host-wide hammer that users shouldn't be reaching for. `kuke init` advertises itself as a self-sufficient bring-up step; it should live up to that.

## Related

- #292 (silent CNI ADD-error swallowing at `start.go:457`) — different bug, same symptom class ("cell looks healthy, isn't").
- `prerequisites.md` currently does not warn about FORWARD policy; the fix above makes the warning unnecessary.

eminwux:
## Summary
- Adds `internal/firewall` package owning a `KUKEON-FORWARD` chain jumped to from FORWARD at position 1. The chain admits return-traffic (RELATED,ESTABLISHED) and any packet to/from a kukeon bridge (`-i k-+` / `-o k-+`). Wired into `kuke init` so cells reach the network on hosts where `iptables -P FORWARD DROP` is the default (Docker, firewalld, ufw, hardened distros) — currently those hosts silently lose every cell egress packet despite the bridge, IPAM, and SNAT rule being healthy.
- Mirrored teardown wired into `kuke daemon reset --purge-system` so a full re-bootstrap leaves no kukeon-owned host firewall state behind.
- New `errdefs.ErrForwardAdmissionApply` / `ErrForwardAdmissionRemove` sentinels follow the `ErrEgressApply` / `ErrEgressRemove` convention next door so callers can `errors.Is` them.
- Idempotency follows the existing netpolicy enforcer pattern: every install step does `iptables -C` before `-I`/`-A`, so re-running `kuke init` on a healthy host produces zero rule churn (test guarantees this).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test $(go list ./... | grep -v /e2e)` — full unit test suite (the `make test` body; `make` is not available in this environment but the target's body is exactly that command). Includes new `internal/firewall` coverage:
  - `TestAdmissionRules_Order` — locks rule order: stateful ACCEPT, then `-i k-+`, then `-o k-+`. Regression guard on the silent-egress-loss bug.
  - `TestBridgeIfaceMatch_CoversCNINames` — pins `BridgeIfaceMatch = "k-+"` against `internal/cni.SafeBridgeName`'s `k-<8 hex>` output so the two packages cannot drift.
  - `TestInstall_OnFreshHost` — emits `-N`, three `-A` rules, and `-I FORWARD 1 -j KUKEON-FORWARD` when nothing is in place.
  - `TestInstall_IsIdempotent` — second install on a healthy host issues zero `-N`/`-A`/`-I`, only the `-L`/`-C` existence checks.
  - `TestInstall_ChainCreateFailureWrapsSentinel` — `-N` failure wraps `ErrForwardAdmissionApply`.
  - `TestRemove_OnInstalledHost` — emits `-D FORWARD -j KUKEON-FORWARD`, `-F`, `-X`.
  - `TestRemove_TolerantWhenChainAbsent` — `-F`/`-X` failures on a fresh host are debug-logged, not returned.
  - `TestRemove_DeleteJumpFailureWrapsSentinel` — `-D` failure wraps `ErrForwardAdmissionRemove`.
- [x] `golangci-lint run` against the touched packages — zero new lints introduced. (Pre-existing `reassign` lints in `cmd/kuke/init/init_test.go:654-704` are untouched.)
- [ ] `make dev-init` smoke test — not run. This change is host-firewall state outside the daemon-parity guard's coverage: the regression check at the tail of `dev-init.sh` reads `kuke get realms` with and without `--no-daemon`, which exercises `/opt/kukeon` bind-mount parity, not iptables. The full repro would require a host with `iptables -P FORWARD DROP` set, which the maintainer flagged as the natural smoke environment in the issue's "Test plan suggestion." Defers to the reviewer's host.

**Relative ordering with `KUKEON-EGRESS`.** The netpolicy enforcer also inserts at FORWARD position 1, so install order matters. `kuke init` runs first and lands `KUKEON-FORWARD` at position 1; the first egress-policy install (per-space, runner-side) later inserts `KUKEON-EGRESS` at position 1, pushing `KUKEON-FORWARD` to position 2. The resulting order — KUKEON-EGRESS first (may DROP), then KUKEON-FORWARD (admits surviving kukeon-bridge traffic) — is the intended one. Documented in the `ForwardChainName` comment in `internal/firewall/forward.go`. The pre-egress race window only opens if a re-`init` runs after KUKEON-EGRESS already exists, but presence-based idempotency means re-init is a no-op there.

**No nftables-detection step.** The issue suggested surfacing a clear error if the host has loaded `nftables` as the primary backend (no working `iptables` shim). I deferred that: the typical Docker-on-Debian-13 case ships `iptables-nft` as the `iptables` shim and our rules work fine through it. If `iptables` is genuinely absent, `exec.CommandContext` already returns an `executable file not found` error wrapped via `ErrForwardAdmissionApply`. Happy to add a `iptables --version` probe + structured error if you want a stronger diagnostic story; left it out to keep blast radius narrow.

**Integration point: `cmd/kuke/init/runInit`, not `controller.Bootstrap`.** Host-level firewall state is set once at init and not per-realm/space, so it sits next to the existing `sysuser.EnsureUserGroup` step (also host-level) rather than inside the controller's per-realm bootstrap loop. Kept the controller's iptables-free shape intact.

**`runInit` length refactor.** Adding the firewall step pushed `runInit` past the `funlen: 100` lint. Extracted the post-bootstrap chown block into `applyKukeonOwnership` (the same shape — local-helper, returning `(permissionsReport, error)`). The behaviour is unchanged; the diff is mechanical.

Closes #293

### DIFF
--- a/cmd/kuke/daemon/reset/reset.go
+++ b/cmd/kuke/daemon/reset/reset.go
@@ -37,6 +37,7 @@ import (
 	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/controller"
 	"github.com/eminwux/kukeon/internal/errdefs"
+	"github.com/eminwux/kukeon/internal/firewall"
 	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 	"github.com/spf13/cobra"
@@ -166,6 +167,15 @@ func runReset(cmd *cobra.Command, _ []string) error {
 		}
 		if removed {
 			cmd.Printf("removed %s\n", systemDir)
+		}
+
+		// Symmetric with `kuke init`: full re-bootstrap should leave no
+		// kukeon-owned host firewall state behind. Per-space egress chains
+		// are torn down by space-deletion paths; this removes the host-wide
+		// admission chain installed at init time.
+		fwInstaller := firewall.NewInstaller(logger)
+		if fwErr := fwInstaller.Remove(cmd.Context()); fwErr != nil {
+			return fmt.Errorf("remove forward admission chain: %w", fwErr)
 		}
 	}
 
--- a/cmd/kuke/init/init.go
+++ b/cmd/kuke/init/init.go
@@ -31,6 +31,7 @@ import (
 	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/controller"
 	"github.com/eminwux/kukeon/internal/errdefs"
+	"github.com/eminwux/kukeon/internal/firewall"
 	"github.com/eminwux/kukeon/internal/serverconfig"
 	"github.com/eminwux/kukeon/internal/sysuser"
 	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
@@ -206,6 +207,50 @@ func resolveKukeondImage() string {
 	return fmt.Sprintf("%s:%s", repo, tag)
 }
 
+// installForwardAdmission installs the kukeon-owned FORWARD admission chain
+// so cells reach the network on hosts where `iptables -P FORWARD DROP` is
+// the default (Docker, firewalld, ufw, hardened distros). Idempotent —
+// re-running `kuke init` on a healthy host produces no rule churn.
+func installForwardAdmission(ctx context.Context, logger *slog.Logger) error {
+	if err := firewall.NewInstaller(logger).Install(ctx); err != nil {
+		return fmt.Errorf("install forward admission chain: %w", err)
+	}
+	return nil
+}
+
+// applyKukeonOwnership chown+chmods /run/kukeon (socket's parent dir,
+// already created by ensureSocketDir during bootstrap) and /opt/kukeon
+// (RunPath). Both end up root:kukeon mode 0o750 so the kukeon group can
+// traverse without world access.
+func applyKukeonOwnership(
+	socketPath, runPath string,
+	ensure sysuser.EnsureResult,
+) (permissionsReport, error) {
+	runDir := filepath.Dir(socketPath)
+	r := permissionsReport{
+		User:         consts.KukeonSystemUser,
+		Group:        consts.KukeonSystemGroup,
+		UID:          ensure.UID,
+		GID:          ensure.GID,
+		UserCreated:  ensure.UserCreated,
+		GroupCreated: ensure.GroupCreated,
+		RunDirPath:   runDir,
+		RunPath:      runPath,
+		SocketPath:   socketPath,
+	}
+	if err := sysuser.ChownAndChmod(runDir, 0, ensure.GID, kukeonRunDirMode); err != nil {
+		return r, fmt.Errorf("apply kukeon ownership to %q: %w", runDir, err)
+	}
+	r.RunDirApplied = true
+	if err := sysuser.ChownTreeAndChmod(
+		runPath, 0, ensure.GID, kukeonRunPathDirMode, kukeonRunPathFileMode,
+	); err != nil {
+		return r, fmt.Errorf("apply kukeon ownership to %q: %w", runPath, err)
+	}
+	r.RunPathApplied = true
+	return r, nil
+}
+
 func runInit(cmd *cobra.Command, _ []string) error {
 	logger, ok := cmd.Context().Value(types.CtxLogger).(*slog.Logger)
 	if !ok || logger == nil {
@@ -247,6 +292,10 @@ func runInit(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("ensure kukeon user/group: %w", ensureErr)
 	}
 
+	if fwErr := installForwardAdmission(cmd.Context(), logger); fwErr != nil {
+		return fwErr
+	}
+
 	opts := controller.Options{
 		RunPath:              runPath,
 		ContainerdSocket:     viper.GetString(config.KUKEON_ROOT_CONTAINERD_SOCKET.ViperKey),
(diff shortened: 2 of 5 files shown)


eminwux:
**PR #299 Review — `fix: install KUKEON-FORWARD admission chain at init`**

Solid host-firewall fix. Presence-based idempotency mirrors the netpolicy enforcer pattern, tests lock rule order / idempotency / sentinel wrapping, and the KUKEON-EGRESS ordering claim is accurate against `internal/netpolicy/enforcer.go` (`-I FORWARD 1`, distinct chain name). No correctness concerns, no sensitive data in errors or logs.

Cross-checks:

- `firewall.NewInstaller` / `Install` / `Remove` are wired from `cmd/kuke/init/init.go` and `cmd/kuke/daemon/reset/reset.go`; the `Remove` call sits inside `if purgeSystem` so a basic `kuke daemon reset` leaves the chain intact for the next init.
- `BridgeIfaceMatch = "k-+"` covers `cni.SafeBridgeName` output (`k-<8 hex>`); `TestBridgeIfaceMatch_CoversCNINames` is a real guard against the two packages drifting.
- Logger is nil-checked at both call sites before reaching `i.logger.InfoContext`.
- The `dev-init` deferral is reasonable — its regression guard reads `kuke get realms`, which doesn't exercise host iptables.

LGTM.

PR #300, opened 2026-05-05
fix: install KUKEON-FORWARD chain so cells egress past restrictive FORWARD policy

### LINKED ISSUE #293: init: kuke init does not install FORWARD-chain admission rules for kukeon bridges; cells lose egress on hosts where FORWARD policy is DROP

eminwux:
## Summary

- New `internal/hostfw` package owns a kukeon-managed iptables chain `KUKEON-FORWARD` that admits traffic to and from kukeon-managed bridges (matched by interface-wildcard `k-+`). Without this chain, cells on hosts where the FORWARD default policy is DROP (typical state after Docker/firewalld/ufw/some hardened distros) get an IP and route table but every egress packet is silently dropped on the host — DNS fails, ping disappears.
- `kuke init` calls `hostfw.Apply()` after the controller bootstrap completes. Idempotent: re-running on a healthy host produces no rule churn (every install is `-C` before `-I`/`-A`, mirroring `internal/netpolicy/enforcer.go`'s pattern). `kuke daemon reset --purge-system` calls `hostfw.Remove()` so the chain doesn't dangle after a fully-clean re-bootstrap.
- The chain layout matches the issue:
  ```
  KUKEON-FORWARD:
    -A KUKEON-FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
    -A KUKEON-FORWARD -i k-+ -j ACCEPT
    -A KUKEON-FORWARD -o k-+ -j ACCEPT
  FORWARD:
    -j KUKEON-FORWARD (inserted near the head — see ordering note below)
  ```
- Every rule carries a `--comment kukeon-forward:<role>` tag so `iptables -S` is self-documenting and the cleanup grep cannot collide with user rules sharing the chain name.
- On hosts without iptables on PATH (nftables-only with no compat shim, minimal containers): `hostfw.IsIptablesAvailable()` returns false, init logs a WARN and continues without installing the chain. This preserves backward compatibility for hosts where the operator already manages admission themselves; the issue's "surface a clear error" intent is met by the visible WARN line in the init output. The reset step is symmetric — skipped silently if iptables is absent (the chain could not have been installed in the first place).

## Ordering against `KUKEON-EGRESS`

The netpolicy package's `KUKEON-EGRESS` master chain enforces per-space DROP rules and also lives at FORWARD position 1. Ordering matters: a per-space DROP must win over the blanket admission `KUKEON-FORWARD` provides, so the egress chain has to come first. The installer detects `KUKEON-EGRESS` in FORWARD and inserts `KUKEON-FORWARD` immediately *after* it (position 2 in that case); when no egress chain is present, `KUKEON-FORWARD` goes to position 1 and a later `netpolicy.Apply()` will push it to 2 — same correct end state. `TestApply_PlacesJumpAfterEgressChain` is the regression guard.

## Design notes (called out per the issue's "Open design questions")

- **init-time install over per-space install** (the issue's first open question): drafted as init-time per the "kuke init must be clean and add all it needs" framing. Rule churn stays at zero across space create/delete; the chain is provisioned once.
- **`k-+` wildcard over exact bridge names** (second open question): chosen for simplicity, mirroring how Docker matches its own bridges. Confirmed safe because `internal/cni/config.go:SafeBridgeName` always emits `k-<8 hex>` (10 chars, well under IFNAMSIZ-1=15) — a non-kukeon process would have to deliberately collide with the `k-` prefix to be admitted.
- **Did not** modify the FORWARD chain's default policy (per the issue's #3). Did not depend on Docker's `DOCKER-USER` chain or any other third-party iptables structure (per #4). `kuke daemon reset --purge-system` removes the chain it owns, mirroring the netpolicy enforcer's removal pattern.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test $(go list ./... | grep -v /e2e)` — full unit-test suite passes. Tests cover `TestBuildChainRules_OrderAndContent`, `TestApply_FreshHostInstallsChainAndJump`, `TestApply_IsIdempotent`, `TestApply_PlacesJumpAfterEgressChain`, `TestApply_FailureWrappedAsHostFwApply`, `TestApply_DoesNotMatchEgressLikeSiblingChain`, `TestRemove_DeletesJumpAndChain`, `TestRemove_ToleratesMissingForwardListing`, `TestDaemonReset_PurgeSystemTearsDownHostFirewall` / `...DoesNotTouchHostFirewall` / `...PropagatesHostFirewallError`, `TestPrintHostFirewallActions`.
- [x] `golangci-lint run` against the touched packages — zero new lints introduced. Two lints in `internal/hostfw/hostfw.go` (gosec G204 on the `exec.CommandContext` call, intentionally suppressed with an inline comment) mirror the patterns already used in `internal/netpolicy/enforcer.go`. The pre-existing `reassign` lints in `cmd/kuke/init/init_test.go` (lines 670–723) are untouched.
- [x] `make dev-init` smoke test — passed. Daemon-parity check produced the canonical realm listing for both `kuke get realms` and `kuke get realms --no-daemon`. `iptables -S KUKEON-FORWARD` shows the three expected comment-tagged rules (`kukeon-forward:established`, `kukeon-forward:egress`, `kukeon-forward:ingress`) and FORWARD has `-j KUKEON-FORWARD` installed. Headline `iptables -P FORWARD DROP` repro not exercised — would be destructive against unrelated workloads on the test host; chain installation itself is verified.

Closes #293

### DIFF
--- a/cmd/kuke/daemon/reset/reset.go
+++ b/cmd/kuke/daemon/reset/reset.go
@@ -195,7 +195,7 @@ func tearDownHostFirewall(cmd *cobra.Command, logger *slog.Logger) error {
 	if err := installer.Remove(cmd.Context()); err != nil {
 		return fmt.Errorf("remove host firewall admission rules: %w", err)
 	}
-	cmd.Printf("removed host firewall chain %s\n", hostfw.ChainName)
+	cmd.Printf("ensured host firewall chain %s removed\n", hostfw.ChainName)
 	return nil
 }
 
--- a/cmd/kuke/daemon/reset/reset.go
+++ b/cmd/kuke/daemon/reset/reset.go
@@ -37,6 +37,7 @@ import (
 	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/controller"
 	"github.com/eminwux/kukeon/internal/errdefs"
+	"github.com/eminwux/kukeon/internal/hostfw"
 	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 	"github.com/spf13/cobra"
@@ -56,6 +57,11 @@ type MockSocketDirKey struct{}
 // /opt/kukeon/kuke-system from. Tests use this to point at a tmpdir so the
 // cleanup never touches the real /opt/kukeon.
 type MockRunPathKey struct{}
+
+// MockHostFwKey injects a hostfw.Installer (typically a fake) so unit
+// tests can assert the --purge-system path tears down the FORWARD
+// admission chain without requiring a real iptables binary.
+type MockHostFwKey struct{}
 
 // defaultTimeout matches `kuke daemon stop`'s grace period (#219) so the two
 // verbs agree on how long the SIGTERM phase gets before SIGKILL escalates.
@@ -167,8 +173,29 @@ func runReset(cmd *cobra.Command, _ []string) error {
 		if removed {
 			cmd.Printf("removed %s\n", systemDir)
 		}
+
+		if hostFwErr := tearDownHostFirewall(cmd, logger); hostFwErr != nil {
+			return hostFwErr
+		}
 	}
 
+	return nil
+}
+
+// tearDownHostFirewall removes the KUKEON-FORWARD chain installed by
+// `kuke init`. The `--purge-system` flag advertises a "fully clean
+// re-bootstrap"; without this step the chain dangles after the daemon
+// is gone. Skipped silently when iptables is not on PATH (the chain
+// could not have been installed in the first place).
+func tearDownHostFirewall(cmd *cobra.Command, logger *slog.Logger) error {
+	installer := resolveHostFwInstaller(cmd, logger)
+	if installer == nil {
+		return nil
+	}
+	if err := installer.Remove(cmd.Context()); err != nil {
+		return fmt.Errorf("remove host firewall admission rules: %w", err)
+	}
+	cmd.Printf("removed host firewall chain %s\n", hostfw.ChainName)
 	return nil
 }
 
@@ -311,6 +338,21 @@ func resolveSocketDir(cmd *cobra.Command) string {
 		socketPath = config.KUKEOND_SOCKET.Default
 	}
 	return filepath.Dir(socketPath)
+}
+
+// resolveHostFwInstaller picks the hostfw.Installer used by the
+// --purge-system step. Tests inject a fake via MockHostFwKey;
+// production builds an IptablesInstaller, returning nil when iptables
+// is absent so the cleanup is silently skipped (the chain could not
+// have been installed without iptables in the first place).
+func resolveHostFwInstaller(cmd *cobra.Command, logger *slog.Logger) hostfw.Installer {
+	if mock, ok := cmd.Context().Value(MockHostFwKey{}).(hostfw.Installer); ok {
+		return mock
+	}
+	if !hostfw.IsIptablesAvailable() {
+		return nil
+	}
+	return hostfw.NewIptablesInstaller(logger)
 }
 
 // resolveRunPath picks the run path for the --purge-system step. Tests inject
--- a/cmd/kuke/daemon/reset/reset_test.go
+++ b/cmd/kuke/daemon/reset/reset_test.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"log/slog"
 	"os"
@@ -32,6 +33,7 @@ import (
 	reset "github.com/eminwux/kukeon/cmd/kuke/daemon/reset"
 	"github.com/eminwux/kukeon/cmd/types"
 	"github.com/eminwux/kukeon/internal/consts"
+	"github.com/eminwux/kukeon/internal/errdefs"
 	"github.com/eminwux/kukeon/internal/hostfw"
 	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
@@ -536,7 +538,9 @@ func TestDaemonReset_PurgeSystemPropagatesHostFirewallError(t *testing.T) {
 			return kukeonv1.DeleteCellResult{Cell: doc, MetadataDeleted: true}, nil
 		},
 	}
-	installer := &recordingInstaller{removeErr: errors.New("iptables down")}
+	installer := &recordingInstaller{
+		removeErr: fmt.Errorf("%w: iptables down", errdefs.ErrHostFwRemove),
+	}
 
 	cmd := reset.NewResetCmd()
 	buf := &bytes.Buffer{}
@@ -555,8 +559,8 @@ func TestDaemonReset_PurgeSystemPropagatesHostFirewallError(t *testing.T) {
 	if err == nil {
 		t.Fatal("expected error when host-firewall teardown fails")
 	}
-	if !strings.Contains(err.Error(), "host firewall") {
-		t.Errorf("error should mention host firewall: %v", err)
+	if !errors.Is(err, errdefs.ErrHostFwRemove) {
+		t.Errorf("error should wrap errdefs.ErrHostFwRemove: %v", err)
 	}
 }
 
--- a/cmd/kuke/daemon/reset/reset_test.go
+++ b/cmd/kuke/daemon/reset/reset_test.go
@@ -32,10 +32,25 @@ import (
 	reset "github.com/eminwux/kukeon/cmd/kuke/daemon/reset"
 	"github.com/eminwux/kukeon/cmd/types"
 	"github.com/eminwux/kukeon/internal/consts"
+	"github.com/eminwux/kukeon/internal/hostfw"
 	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 	"github.com/spf13/viper"
 )
+
+// recordingInstaller is a hostfw.Installer that records whether Remove
+// was invoked, so reset_test can assert the --purge-system wiring fires
+// the host-firewall teardown hook.
+type recordingInstaller struct {
+	removed   bool
+	removeErr error
+}
+
+func (r *recordingInstaller) Apply(_ context.Context) error { return nil }
+func (r *recordingInstaller) Remove(_ context.Context) error {
+	r.removed = true
+	return r.removeErr
+}
 
 // TestDaemonReset covers the table of states the verb has to handle: a
 // running daemon (graceful stop+delete), an already-stopped daemon (skip
@@ -394,6 +409,7 @@ func TestDaemonReset_PreservesDefaultRealm(t *testing.T) {
 	ctx = context.WithValue(ctx, reset.MockClientKey{}, kukeonv1.Client(fake))
 	ctx = context.WithValue(ctx, reset.MockSocketDirKey{}, t.TempDir())
 	ctx = context.WithValue(ctx, reset.MockRunPathKey{}, runPath)
+	ctx = context.WithValue(ctx, reset.MockHostFwKey{}, hostfw.Installer(hostfw.NoopInstaller{}))
 	cmd.SetContext(ctx)
 	cmd.SetArgs([]string{"--purge-system"})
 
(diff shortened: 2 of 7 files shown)


eminwux:
**PR #300 Review — `fix: install KUKEON-FORWARD chain so cells egress past restrictive FORWARD policy`**

Solid implementation with nicer features than the minimum — explicit position-after-`KUKEON-EGRESS`, `--comment` self-documenting tags, iptables-availability detection, and the print-action layout matches the rest of init's output. Tests cover order, idempotency, the cross-package position-finding, and the reset-wiring contract. One blocking item below.

Blocking:

- **Dead public API** `internal/errdefs/errdefs.go:145`: `ErrIptablesUnavailable` is defined but never referenced in the tree. `init.go`'s `installHostFirewall` returns `nil` (with a `WarnContext` + `Skipped` report) on missing iptables rather than wrapping this sentinel, so its docstring claim about "surface this as a clear bring-up blocker" never fires. Either delete the sentinel, or have the install path return `fmt.Errorf("%w", errdefs.ErrIptablesUnavailable)` so callers can `errors.Is` it.

Non-blocking:

- **Sentinel test loosely written** `internal/hostfw/hostfw_test.go:212`: `TestApply_FailureWrappedAsHostFwApply` only checks `strings.Contains(err.Error(), "host firewall")` — that passes whether or not the error is wrapped with `errdefs.ErrHostFwApply`. Tighten to `errors.Is(err, errdefs.ErrHostFwApply)` so a future regression that drops the sentinel wrap doesn't slip through. Same fix at `cmd/kuke/daemon/reset/reset_test.go:558` (assert `errors.Is(err, errdefs.ErrHostFwRemove)` against the recordingInstaller's wrapped error).
- **Egress detection edge case** `internal/hostfw/hostfw.go:285`: `strings.Contains(line, "-j "+netpolicy.MasterChainName)` is a substring match — it would also match a hypothetical `KUKEON-EGRESS-FOO` chain a user might add. Anchoring with `strings.HasSuffix` (or splitting on whitespace and exact-matching the trailing token) is tighter. Same shape used in `deleteJumpsToChain` at `internal/hostfw/hostfw.go:296`.
- **Misleading reset output** `cmd/kuke/daemon/reset/reset.go:198`: prints `"removed host firewall chain KUKEON-FORWARD"` unconditionally on Remove success, even when nothing was actually removed (chain absent — Remove no-ops gracefully via the `-F`/`-X` debug-log path). Either thread a removed-bool back from `Remove` and print only on actual change (mirroring the existing `removed %s` pattern above), or reword to `"ensured host firewall chain %s removed"`.

LGTM once `ErrIptablesUnavailable` is resolved — the rest are non-blocking.

eminwux:
Perform the smoke tests

eminwux:
Stalling — PR #299 has merged with the same feature, closing issue #293 (the AC this PR was filed against). The two PRs implement `KUKEON-FORWARD` with different package layouts (`internal/firewall/forward.go` in #299 vs `internal/hostfw/hostfw.go` here) and overlapping init/reset wiring, so rebasing this branch onto current `main` would either back out the merged implementation or duplicate the chain.

This PR is now redundant. Owner to decide whether to close it or salvage specific bits (e.g., the `ErrIptablesUnavailable` UX path on missing-iptables hosts is a behavioral difference vs. the merged implementation).

Swapping `changes-requested` → `ci-stalled` so the cron stops retrying.

Original review items left unaddressed (recorded for context if anything is salvaged into a follow-up):

Blocking:
- Dead public API `ErrIptablesUnavailable` (`internal/errdefs/errdefs.go:145`) — never wired by `installHostFirewall`, so the docstring claim never fires. Either delete or wrap on the missing-iptables path.

Non-blocking:
- `TestApply_FailureWrappedAsHostFwApply` (`internal/hostfw/hostfw_test.go:212`) only `strings.Contains(err.Error(), "host firewall")` — tighten to `errors.Is(err, errdefs.ErrHostFwApply)`. Same fix at `cmd/kuke/daemon/reset/reset_test.go:558` for `ErrHostFwRemove`.
- Egress detection at `internal/hostfw/hostfw.go:285` substring-matches `KUKEON-EGRESS`, would also match `KUKEON-EGRESS-FOO`. Anchor with `HasSuffix` or whitespace-split exact match. Same shape at `:296`.
- `cmd/kuke/daemon/reset/reset.go:198` prints `"removed host firewall chain KUKEON-FORWARD"` unconditionally on Remove — thread a removed-bool back, or reword to `"ensured … removed"`.

eminwux:
Addressed in a428136:

Blocking:
- Dropped the dead `ErrIptablesUnavailable` sentinel from `internal/errdefs/errdefs.go`. The design intentionally warns-and-continues when iptables is missing (per PR description), so the sentinel could never fire — deletion matches the actual behavior, eliminating the docstring/runtime mismatch.

Non-blocking:
- Tightened `TestApply_FailureWrappedAsHostFwApply` (`internal/hostfw/hostfw_test.go:213`) and `TestDaemonReset_PurgeSystemPropagatesHostFirewallError` (`cmd/kuke/daemon/reset/reset_test.go:562`) to assert `errors.Is(err, errdefs.ErrHostFw{Apply,Remove})` instead of `strings.Contains`. The reset test's recordingInstaller now wraps its return with `errdefs.ErrHostFwRemove` to mirror the production `IptablesInstaller.Remove` wrap.
- Added a `lineJumpsTo(line, target)` token-aware helper in `internal/hostfw/hostfw.go` and replaced both `strings.Contains(line, "-j "+chain)` substring checks (`findEgressPosition` and `deleteJumpsToChain`). New regression test `TestApply_DoesNotMatchEgressLikeSiblingChain` pins that a sibling chain like `KUKEON-EGRESS-FOO` does not match `KUKEON-EGRESS`.
- Reworded the reset-time print at `cmd/kuke/daemon/reset/reset.go:198` from `removed host firewall chain %s` to `ensured host firewall chain %s removed` so the line is accurate when Remove no-ops on an already-absent chain.

**Not addressed (broader-than-review concern):** PR #299 has merged a parallel implementation of the same feature in `internal/firewall/` and closed issue #293. Rebasing this branch onto current `origin/main` produces non-mechanical conflicts in `cmd/kuke/init/init.go` and `cmd/kuke/daemon/reset/reset.go` — the duplicate-implementation question is outside this fix-up's scope. The fix-up is pushed without rebasing so the review items don't get lost; resolving the duplicate-implementation choice is up to the owner.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test $(go list ./... | grep -v /e2e)` — full unit-test suite passes including the new `TestApply_DoesNotMatchEgressLikeSiblingChain`.
- [x] `golangci-lint run` against the touched packages — zero new lints; the four pre-existing `reassign` warnings in `cmd/kuke/init/init_test.go` remain untouched.
- [x] `pre-commit run` against staged files — go-fmt, go-mod-tidy, golangci-lint, addlicense all pass.

eminwux:
**PR #300 Review — `fix: install KUKEON-FORWARD chain so cells egress past restrictive FORWARD policy`**

Solid fix for the FORWARD-DROP egress bug: chain layout, `-C`-before-`-I` idempotency, ordering against `KUKEON-EGRESS`, and a symmetric `--purge-system` teardown path are all clean. Comment-tagged rules make the cleanup grep collision-safe, and the test suite exercises fresh-install, idempotent re-run, position-after-egress, sibling-chain non-match (`KUKEON-EGRESS-FOO`), error wrapping, and the reset-time wiring contract.

- **Comment accuracy** `internal/hostfw/hostfw.go:367-372`: the `lineJumpsTo` doc says a malformed line "will propagate any genuine parse failure on the subsequent `parseRuleLine` call" — but `findEgressPosition` (line 285) only calls `lineJumpsTo`, with no subsequent `parseRuleLine`, so a malformed line containing `KUKEON-EGRESS` silently falls through to position 0 / insert-at-1. Either rephrase the doc to scope the guarantee to `deleteJumpsToChain`, or have `findEgressPosition` mirror the propagate-on-parse-error pattern.

- **Double parse** `internal/hostfw/hostfw.go:304-308`: `deleteJumpsToChain` calls `lineJumpsTo` (which parses) then `parseRuleLine` again on the same line. Cosmetic — could thread the pre-parsed tokens through.

LGTM.

Did not run `make dev-init` — host has no iptables/containerd. Reviewer should verify the headline repro on a host with `iptables -P FORWARD DROP` set: `kuke init`, `kuke create space`, `kuke run -a alpine -- ping -c2 1.1.1.1` should now succeed without manually flipping FORWARD to ACCEPT.

eminwux:
Addressed in 0e62fc4:

- **Comment accuracy** (`internal/hostfw/hostfw.go:367`): rewrote `lineJumpsTo`'s doc to scope the propagate-vs-no-match split to its callers — `deleteJumpsToChain` propagates, `findEgressPosition` treats parse failure as non-fatal no-match. The old wording referenced a "subsequent `parseRuleLine` call" that `findEgressPosition` never made.
- **Double parse** (`internal/hostfw/hostfw.go:304`): refactored `lineJumpsTo` to return `(matches bool, tokens []string, err error)` so callers can thread the parsed token vector through. `deleteJumpsToChain` no longer re-parses the line, and `findEgressPosition` discards the tokens since it only needs the position.

## Test plan

- [x] `go build ./...` / `go vet ./...`
- [x] `go test $(go list ./... | grep -v /e2e)` — full unit-test suite passes; existing `TestApply_DoesNotMatchEgressLikeSiblingChain` still pins the token-anchored sibling-chain non-match.
- [x] `golangci-lint run` — zero new lints; the four pre-existing `reassign` warnings in `cmd/kuke/init/init_test.go` remain untouched.
- [x] `pre-commit run` against `internal/hostfw/hostfw.go` — go-fmt, go-mod-tidy, golangci-lint, addlicense all pass.
- [x] `make dev-init` smoke — passed on this host. Daemon-parity check produced the canonical realm listing for both `kuke get realms` and `kuke get realms --no-daemon`. `iptables -S KUKEON-FORWARD` shows the three expected comment-tagged rules; FORWARD has `-j KUKEON-FORWARD` installed.

**Note on rebase against `main`:** branch is unchanged at the `cmd/kuke/init/init.go` / `cmd/kuke/daemon/reset/reset.go` conflict points — PR #299 has merged a parallel `internal/firewall/` implementation of the same feature on `main` and closed issue #293. Rebasing produces non-mechanical conflicts; the duplicate-implementation choice (which package layout to keep, which to drop) is owner's call. Pushing the fix-up so the review items don't get lost regardless of how that resolves.

eminwux:
Closing — issue #293 was closed by PR #299,
     which landed a parallel implementation of the same feature
      in \`internal/firewall/forward.go\`. The hardening extras
      this PR carried (rule comment tags, active KUKEON-EGRESS
     position detection, IsIptablesAvailable warn-and-continue)
      are filed as #315 for salvage into the merged package.

PR #304, opened 2026-05-05
fix: derive cell state from non-root container tasks

### LINKED ISSUE #302: reconciler: cell lifecycle should follow non-root container tasks, not just the root container
## Repro

Reproduced on `main` (current HEAD `8a8dfd4`).

1. `kuke init`.
2. Create a cell in the `default` realm with a long-lived root (e.g. busybox `sleep infinity`) and one or more non-root containers (e.g. `kuke run` against a manifest, with `attachable: true` work containers). Use `--rm` to set `Spec.AutoDelete = true`.
3. Stop or `ctr task kill` every non-root (work) task in the cell, leaving only the root task alive.

## Observed

The cell stays `Ready` indefinitely. The root container keeps running because the reconciler treats it as the lifecycle anchor; nothing observes that all non-root workloads are gone, so:

- The root task is never stopped, even though it has no remaining workload to host.
- `--rm` / `Spec.AutoDelete` cleanup never fires (it only triggers on cell `Stopped`/`Failed`, which only happens when the root task itself dies).
- An operator who attaches a non-root workload, sees it exit, and expects the cell to wind down has to `kuke kill` / `kuke delete` manually.

The opposite direction is also wrong: if the user's mental model is "root keeps the cell shell alive while workloads run", a long-lived root (`sleep infinity`) holds the cell open forever even after every workload has exited.

## Root cause

`internal/controller/runner/refresh.go:490` `deriveCellStateFromRootContainer` is the *sole* input to cell state inside `ReconcileCell`:

```go
func (r *Exec) deriveCellStateFromRootContainer(cell intmodel.Cell) intmodel.CellState {
    rootSpec := findRootContainerSpec(cell)
    if rootSpec == nil {
        return intmodel.CellStateReady
    }
    state, err := r.GetContainerState(cell, rootSpec.ID)
    ...
}
```

`populateCellContainerStatuses` does query non-root container states and persists them into `cell.Status.Containers`, but those are display-only — `ReconcileCell` ignores them when deciding cell state and ignores them when deciding whether to fire `autoDeleteCell`. The autodelete trigger (`shouldAutoDeleteCell` → `cellStateAutoDeleteTriggers`) only fires on `Stopped`/`Failed`, which is only reached via root-container observation.

Net effect: non-root tasks have zero influence on cell lifecycle. There is no code path in `ReconcileCell` that:

1. inspects non-root container task states,
2. stops the root container when all non-root tasks are gone,
3. (conversely) keeps the root alive specifically because non-root tasks are running.

The intended semantic per the operator's mental model is: the root container is auxiliary plumbing that exists to host workloads. The cell's "is something happening here" signal should come from the union of non-root task states, with the root container driven from that signal — running while workloads run, stopped (and reaped under `--rm`) when none remain.

## Suggested fix

In `ReconcileCell` / `deriveCellStateFromRootContainer` (or a new derivation helper), when the cell has at least one non-root container in `Spec.Containers`:

- Derive cell state from the union of non-root container task states (Ready if any non-root task is `Ready`/`Running`/`Pending`/`Paused`/`Pausing`; Stopped if all non-root tasks are `Stopped`/`Failed`/non-existent).
- When the new derivation is `Stopped` *and* the root container task is still running, kill the root task as part of the reconcile pass (best-effort, idempotent — same shape as `autoDeleteCell`'s `KillCell` call). The next pass observes root stopped and the existing `shouldAutoDeleteCell` path picks up `--rm` cleanup unchanged.
- When the cell has no non-root containers in spec (a cgroup-only cell or a kukeond-style supervised cell with only the root), keep the current behavior.

Care points worth restating in the AC so the next picker doesn't lose them:

- `ReadyObserved` latch must keep its current shape so an in-flight `CreateCell` (cgroup created, non-root containers not yet registered) is not mis-derived as Stopped and reaped before it ever reached Ready.
- This change interacts with #301 (the namespace race that mis-reports non-root containers as Stopped). If the read bug is still present, the new derivation will reap healthy cells on a race. #301 should land first, or this issue's implementation should defensively re-check before kill.
- The kukeond cell itself (`kuke-system / kukeon / kukeon / kukeond`) has no non-root containers in spec, so the new derivation reduces to the current root-only behavior for it — confirm in tests.

## Notes

Surfaced alongside #301 while investigating why `kuke get co` reports running work containers as Stopped. The two are distinct: #301 is a read-correctness bug in the existing model; this issue is a behavior gap in the reconciler model itself. Filing separately so each lands as its own reviewable PR.

eminwux:
## Summary

Closes #302. The reconciler's cell-state derivation now follows the cell's non-root workloads instead of just the root container, so a long-lived `sleep infinity` root no longer holds the cell open after every workload has exited (and `--rm` cleanup actually fires for those cells).

**Before:** `internal/controller/runner/refresh.go`'s `deriveCellStateFromRootContainer` was the sole input to cell state inside `ReconcileCell`. Stopping every non-root task in a cell that has a long-lived root left the cell stuck `Ready` indefinitely — root keeps running, autodelete never triggers, operator has to `kuke kill` manually.

**After:** `deriveCellState` dispatches:
- Cells with **no** non-root container in spec (kukeond-style supervised cells, cgroup-only cells) → fall through to the legacy root-only derivation. Confirmed by the `blocked_for_kukeond_style_cell_root_only` test case — the kukeond cell does not get unintended wind-down.
- Cells with **at least one** non-root container in spec → derive from the union of non-root container task states (Ready if any non-root is active; Stopped if every non-root is Stopped/Failed/non-existent; Unknown if any non-root reads back Unknown — defensive read-side check that pairs with the still-open #301).

When the new derivation flips Stopped/Failed and the root container task is still running:
- `AutoDelete=true` → existing `autoDeleteCell` path (`KillCell` + `DeleteCell`).
- `AutoDelete=false` → new `windDownCell` path (`KillCell` only — cell metadata preserved for explicit `kuke delete cell`, but the root container shell stops so it does not zombie).

Both flavors stay gated by the `ReadyObserved` latch (regression #269), so an in-flight `CreateCell` is never reaped before the cell ever reached Ready.

## Design notes called out in the issue's AC

- **`ReadyObserved` latch shape preserved.** No changes to `latchReadyObserved`; the new wind-down gate uses the same latch via `shouldWindDownCell`. Existing `TestLatchReadyObserved` and `TestAutoDeleteSurvivesKillCellRace` pass unchanged.
- **#301 interaction defensively handled.** `deriveCellStateFromNonRootContainerStatuses` returns `Unknown` (not `Stopped`) when any non-root status reads back `Unknown` — covered by `workload_reads_back_unknown_blocks_stopped_defensive_against_301`. Partial-populate snapshots (seen < expected) also stay Unknown via `missing_workload_status_is_treated_as_unknown_not_stopped`. The wind-down `KillCell` will not fire on a transient containerd hiccup or a #301-induced misread.
- **kukeond cell preserved.** `hasNonRootContainerSpec` is the dispatch gate; its `only_root` test case pins the kukeond fallback. The kukeond cell (`kuke-system / kukeon / kukeon / kukeond`) has only a root container in spec, so `deriveCellState` reduces to the existing `deriveCellStateFromRootContainer`. `shouldWindDownCell` also blocks on the same gate, so no wind-down `KillCell` for kukeond either.
- **Idempotency.** `shouldWindDownCell` consults `rootContainerStillRunning` against the freshly populated status snapshot, so once the root is dead the gate stops firing — no wasted `KillCell` per tick.

## Implementation notes the reviewer might push back on

- **Populate-then-derive ordering swap.** `populateCellContainerStatuses` now runs before state derivation (was after) so the non-root derivation can read from the snapshot instead of re-querying containerd N times. Net containerd round-trips are unchanged — the previous flow queried the root once for derivation plus once for populate; the new flow queries each non-root once for populate plus zero for derivation. Behavior parity for the kukeond fallback path: `deriveCellStateFromRootContainer` still queries the root directly (same as before), which means a redundant root query on that path; chose simplicity over a third specialization, since kukeond-style cells are the small fraction of host cells.
- **`windDownCell` calls `KillCell` (not a narrower "kill root only" helper).** `KillCell` already handles workload-container kills (idempotent — they are gone, this is the trigger condition), CNI detach, root container deletion, and final CNI cleanup. A targeted "kill root task only" would have to duplicate the CNI detach/cleanup logic to leave the cell sane. Reviewer can push back if a lighter-touch helper is preferred.
- **Pre-existing lints unchanged.** Two `golangci-lint` hits in `refresh.go` (the `updated := false` QF1007 and the `deriveCellStateFromRootContainer` switch's missing-`ContainerStateUnknown`-in-exhaustive-switch) are pre-existing on `main` — line numbers shifted but the lints are not introduced by this PR. CLAUDE.md "no unrelated cleanups in feature PRs" applies.

## Test plan

- [x] `go vet ./...` — clean.
- [x] `go test $(go list ./... | grep -v /e2e)` — full unit-test suite passes (matches `make test`). New tests in `internal/controller/runner/refresh_test.go`:
  - `TestHasNonRootContainerSpec` — pins the kukeond-fallback dispatch gate (empty / only root / root+work / only work).
  - `TestDeriveCellStateFromNonRootContainerStatuses` — pins the union-of-non-root derivation, with the headline #302 case (`running_root_is_ignored_when_workloads_are_all_stopped`) plus #301-defense and partial-populate-defense cases.
  - `TestRootContainerStillRunning` — pins the snapshot read used by the wind-down gate so the kill is idempotent across ticks.
  - `TestShouldWindDownCell` — pins the four-condition gate (trigger state, ReadyObserved, hasNonRoot, root-still-running), with the kukeond fallback and idempotency cases.
- [x] `golangci-lint run ./internal/controller/runner/` — only pre-existing lints (see implementation notes); no new lints introduced.
- [x] `make dev-init` smoke test — re-run on the updated dev sandbox after dependencies + cgroup + overlay constraints were lifted. Daemon-parity tail produced the expected output (both `kuke get realms` and `kuke get realms --no-daemon` match):
  ```
  ```
  Bonus end-to-end check (the issue's headline repro) deferred to reviewer's environment — not on the original AC checklist; the daemon-parity guard above is the regression check CLAUDE.md mandates.

Closes #302

### DIFF
--- a/internal/controller/runner/refresh.go
+++ b/internal/controller/runner/refresh.go
@@ -341,54 +341,82 @@ func (r *Exec) refreshContainerStatus(cell intmodel.Cell, containerSpec *intmode
 // ReconcileCell is the daemon-side counterpart to RefreshCell. Where
 // RefreshCell derives cell.Status.State from cgroup existence alone (so
 // `kuke refresh` keeps its current shape), ReconcileCell additionally
-// considers the root container's task state — which is what flips a Ready
-// cell to Stopped when an operator runs `ctr task kill` outside of kukeon.
+// considers container task state — which is what flips a Ready cell to
+// Stopped when an operator runs `ctr task kill` outside of kukeon, or
+// when the cell's non-root workloads have all exited.
 //
-// State derivation:
+// State derivation (see deriveCellState):
 //   - cgroup check error or cgroup absent → Unknown
-//   - cgroup present, no root container in spec → Ready (cgroup-only cell)
-//   - cgroup present, root container task running/created/paused → Ready
-//   - cgroup present, root container missing in containerd or task
-//     stopped/unknown → Stopped
+//   - cgroup present, no non-root container in spec (kukeond-style or
+//     cgroup-only cell) → derived from the root container's task state
+//     (the legacy behavior; root *is* the workload here)
+//   - cgroup present, at least one non-root container in spec → derived
+//     from the union of non-root container task states (Ready if any
+//     non-root is active; Stopped if every non-root is Stopped/Failed/
+//     non-existent; Unknown if any non-root reads back Unknown — the
+//     defensive read-side check that pairs with #301)
 //
-// Container statuses are also re-populated so callers (and `kuke get cell`)
-// see up-to-date per-container state.
+// Container statuses are populated up front so the derivation reads
+// them straight from the snapshot instead of re-querying containerd
+// (and so `kuke get cell` sees the same per-container view the
+// reconciler decided on).
 //
-// AutoDelete: when Spec.AutoDelete is set and the freshly derived state is
-// Stopped or Failed, the reconciler kills + deletes the cell instead of
-// persisting the state transition. This subsumes the per-cell `kuke run
-// --rm` watcher (#161 unblocks the move): a single, restart-resilient
-// ticker is enough, no goroutine-per-cell, and an AutoDelete cell created
-// before a daemon restart still gets cleaned up on the next tick after
-// the daemon is back. Errors during kill/delete are returned to the
-// caller so the loop's per-pass `Errors` slice records them and the cell
-// is preserved for retry on the next tick (best-effort, like the watcher
-// it replaces).
+// Wind-down side-effects: when the derivation flips to Stopped/Failed
+// and the cell has at least one non-root container in spec and the
+// root container task is still running, the reconciler kills the cell
+// so the root container shell stops too. Two flavors:
+//
+//   - AutoDelete=true: autoDeleteCell runs (KillCell + DeleteCell) and
+//     the cell metadata is removed. Subsumes the per-cell
+//     `kuke run --rm` watcher.
+//   - AutoDelete=false: windDownCell runs (KillCell only). The cell
+//     metadata is preserved in Stopped state for the operator to
+//     `kuke delete` explicitly — a long-lived `sleep infinity` root
+//     does not get to hold the cell open after every workload has
+//     exited.
+//
+// Both flavors are gated by the ReadyObserved latch so an in-flight
+// CreateCell (cgroup created, non-root containers not yet registered
+// → derivation reads Stopped from "container does not exist") is not
+// reaped before it ever reached Ready.
+//
+// Errors during kill/delete are returned to the caller so the loop's
+// per-pass `Errors` slice records them and the cell is preserved for
+// retry on the next tick (best-effort, like the watcher it replaces).
 func (r *Exec) ReconcileCell(cell intmodel.Cell) (intmodel.Cell, ReconcileOutcome, error) {
 	originalStatus := cell.Status
-	newState := intmodel.CellStateUnknown
 
 	cgroupExists, cgroupErr := r.ExistsCgroup(cell)
-	switch {
-	case cgroupErr != nil:
+	if cgroupErr != nil {
 		r.logger.DebugContext(r.ctx, "failed to check cell cgroup",
 			"cell", cell.Metadata.Name, "error", cgroupErr)
-	case !cgroupExists:
-		newState = intmodel.CellStateUnknown
-	default:
-		newState = r.deriveCellStateFromRootContainer(cell)
 	}
 
+	// Populate container statuses up front so the non-root-driven
+	// derivation can read them from the snapshot. The root container
+	// path also benefits — populate already queried it, so a single
+	// pass of GetContainerState calls covers both the per-container
+	// status array and the cell-state derivation.
 	if err := r.populateCellContainerStatuses(&cell); err != nil {
 		r.logger.DebugContext(r.ctx, "populate container statuses failed",
 			"cell", cell.Metadata.Name, "error", err)
 	}
+
+	newState := intmodel.CellStateUnknown
+	if cgroupErr == nil && cgroupExists {
+		newState = r.deriveCellState(cell)
+	}
+
 	cell.Status.State = newState
 	cell.Status.ReadyObserved = latchReadyObserved(
 		originalStatus.ReadyObserved, originalStatus.State, newState)
 
 	if shouldAutoDeleteCell(cell.Spec.AutoDelete, newState, cell.Status.ReadyObserved) {
 		return r.autoDeleteCell(cell)
+	}
+
+	if shouldWindDownCell(cell, newState) {
+		return r.windDownCell(cell)
 	}
 
 	updated := false
@@ -408,6 +436,32 @@ func (r *Exec) ReconcileCell(cell intmodel.Cell) (intmodel.Cell, ReconcileOutcom
 		}
 	}
 	return cell, ReconcileOutcome{Updated: updated}, nil
+}
+
+// windDownCell is the non-AutoDelete counterpart to autoDeleteCell:
+// when a cell's non-root workloads have all exited, kill the cell
+// (KillCell handles the root task plus CNI detach plus cleanup) so
+// the root container shell does not zombie. Cell metadata is left
+// in place — the operator runs `kuke delete cell` explicitly when
+// they are done with it. KillCell already populates and persists
+// container statuses, so the caller need not double-write metadata.
+//
+// KillCell is best-effort and idempotent (workload containers that
+// are already gone are no-ops); errors surface so the next reconcile
+// pass retries.
+func (r *Exec) windDownCell(cell intmodel.Cell) (intmodel.Cell, ReconcileOutcome, error) {
+	r.logger.InfoContext(r.ctx, "reconcile: winding down cell after non-root workloads exited",
+		"cell", cell.Metadata.Name,
+		"realm", cell.Spec.RealmName,
+		"space", cell.Spec.SpaceName,
+		"stack", cell.Spec.StackName,
+		"state", cell.Status.State)
+
+	updatedCell, err := r.KillCell(cell)
+	if err != nil {
+		return cell, ReconcileOutcome{}, fmt.Errorf("wind-down: kill cell: %w", err)
+	}
+	return updatedCell, ReconcileOutcome{Updated: true}, nil
 }
 
 // autoDeleteCell runs the kill+delete sequence the per-cell watcher used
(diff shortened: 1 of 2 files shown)


eminwux:
Run pending smoke test

eminwux:
Cannot run `make dev-init` on this host — the dev environment lacks `containerd`, `docker`, `sudo`, `iptables`, and `/opt/cni`, so `scripts/dev-init.sh`'s `kuke daemon reset` + `docker build` + `kuke image load` + `kuke init` loop cannot execute. The original PR body already deferred this verification to a reviewer's environment for that reason; this comment now formalizes the deferral as a stall so the cron stops retrying.

Marking `ci-stalled` for human investigation per CLAUDE.md "When a reviewer-blocking verification step can't safely run on the current host" (reviewer did not defer, so not `ready-for-review`).

To unblock: run `make dev-init` on a host with containerd + docker + sudo and verify the daemon-parity tail. The headline #302 repro is the bonus end-to-end check in the test plan.

eminwux:
Try again, dependencies have been installed in dev.

eminwux:
Re-checked the dev host after your latest comment — the dependencies needed by `scripts/dev-init.sh` are still missing:

```
$ command -v sudo docker containerd ctr iptables
(all empty)
$ apt-get list --installed 2>/dev/null | wc -l
0
$ ls -ld /opt/cni
drwxr-x--- 5 root root 4096 May  5 15:06 /opt/cni   # root:root, not readable as `claude`
```

`/opt/cni` exists but is root-only; no `sudo`, `docker`, `containerd`, `ctr`, or `iptables` binary on `$PATH`, and the package set is empty (`dpkg -l` lists nothing). `make dev-init` cannot proceed past `scripts/dev-init.sh`'s first `sudo ./kuke daemon reset` without these.

Marking `ci-stalled` again per CLAUDE.md "When a reviewer-blocking verification step can't safely run on the current host" — the reviewer did not defer to a different environment, so this is for human investigation rather than `ready-for-review`.

To unblock, the dev container image needs `sudo`, `docker` (with daemon running), `containerd`/`ctr`, `iptables`, and a writable `/opt/cni` for the `claude` user (or the script invoked under root).

eminwux:
Re-checked the dev host after "dependencies have been installed in dev." Things have moved forward but `make dev-init` still cannot run — the structural blocker is now the sandbox itself, not a missing-binaries gap.

What works now:
- `sudo`, `docker`, `containerd`, `ctr`, `iptables` are all on `$PATH`.
- I'm running as `root` (so `sudo` is a no-op anyway).
- `containerd` starts and `/run/containerd/containerd.sock` opens fine.

What still blocks `scripts/dev-init.sh`:
1. `dockerd` aborts on startup with `failed to start daemon: Devices cgroup isn't mounted`. Inside this Kukeon-cell-hosted dev sandbox `/sys` is mounted read-only (`sysfs on /sys type sysfs (ro,nosuid,nodev,noexec,relatime)`), `/sys/fs/cgroup` is empty, and `mount -t cgroup2 cgroup2 /sys/fs/cgroup` returns `permission denied` even as root. `docker build kukeon-local:dev` therefore can't run.
2. `make` is not installed and `apt-get install make` reports `Package 'make' has no installation candidate`, so `scripts/dev-init.sh`'s first line — `make kuke` — fails with `make: command not found`.

Both are host-image / sandbox-config gaps, not anything addressable from inside the agent's session.

`/proc/self/cgroup` confirms this dev container is itself running as `0::/default.kukeon.io/default_default_kukeon-dev-address-22abc3_work` — a Kukeon cell with no cgroup pass-through and no Docker-in-Docker support. That's the same restriction reported earlier, just with a different surface symptom now that the binaries are in place.

Keeping `ci-stalled` per CLAUDE.md "When a reviewer-blocking verification step can't safely run on the current host" — the original reviewer ask ("Run pending smoke test") did not defer to a different environment, so this is for human investigation rather than `ready-for-review`.

To unblock from a host that can run `make dev-init`, one of:
- Run `make dev-init` on a host with `make`, a writable cgroupfs (so `dockerd` can start), and the existing apt deps. CI is already green; this is just the daemon-parity tail check from CLAUDE.md.
- Or rebuild the dev container image with `make` installed and `/sys/fs/cgroup` exposed read-write (e.g. cell launched with `--privileged`-equivalent cgroup propagation), and re-trigger this PR.

eminwux:
Sandbox updated, try again smoke tests

eminwux:
Re-checked the dev host after "Sandbox updated, try again smoke tests". Significant progress — the sandbox now has every binary and most filesystem access `make dev-init` needs, and the run advances much further than before. New blocker is a single cgroupv2 nesting constraint, not a missing-tools gap.

What works now (that didn't last round):
- `sudo`, `docker`, `containerd`, `ctr`, `iptables`, `make`, `go` are all on `$PATH`.
- `/sys/fs/cgroup` is mounted `cgroup2 rw` (the previous `ro` `sysfs` is gone), so `dockerd` can start.
- I needed two host-side workarounds to get `make dev-init` past its earlier failure points: `dockerd` had to be started with `storage-driver=vfs` (overlayfs-on-overlayfs `EINVAL` on the buildkit cache mount), and `containerd`'s data root had to be relocated to a tmpfs (`/var/lib/containerd2`, mounted `tmpfs 5G`) for the same reason on the rootfs snapshotter. Neither is needed on a normal host — both are pure compensations for this nested cell's overlay/cgroup constraints.

How far the run gets now:
- `make kuke` ✓
- `docker build kukeon-local:dev` ✓
- `kuke daemon reset` ✓ (`kukeond stopped` / `kukeond cell deleted`)
- `kuke image load --from-docker kukeon-local:dev --realm kuke-system --no-daemon` ✓ (`loaded 1 image(s) into realm "kuke-system"`)
- `kuke init --kukeond-image docker.io/library/kukeon-local:dev` advances through realm/space/stack creation, cgroup creation, and container-spec creation, then fails at the very last step:

```
ERROR "failed to create task" id=kukeon_kukeon_kukeond_root err=failed to create shim task: OCI runtime create failed: runc create failed: container's cgroup is not empty: 1 process(es) found
Error: failed to start cell containers: failed to start root container kukeon_kukeon_kukeond_root: failed to create task: failed to create shim task: OCI runtime create failed: runc create failed: container's cgroup is not empty: 1 process(es) found
```

This is the runc cgroupv2 "race": `containerd-shim-runc-v2` places its own pid in `/sys/fs/cgroup/<containerd-ns>/<container-id>` before `runc create` runs, and `runc`'s "is the target cgroup empty?" precheck then trips on its own shim. Confirmed by inspecting `/sys/fs/cgroup/kuke-system.kukeon.io/kukeon_kukeon_kukeond_root/cgroup.events` immediately after the failure — `populated 1` during the create, `populated 0` after. The cell cgroups under `/sys/fs/cgroup/kukeon/kuke-system/...` that `kuke init` itself created are all empty (0 procs) — the conflict is in the containerd-namespace cgroup runc uses, not in the kukeon cell tree.

The standard escape hatches don't apply here:
- `systemd` cgroup driver would sidestep the precheck (transient slices skip the empty check) but `systemd` isn't running in the sandbox.
- Setting `Linux.CgroupsPath` in the OCI spec to a per-process scope would also avoid it but is a kuke-side change, not in the scope of this PR.
- Running with cgroup namespacing (so the shim and runc each see their own root) would also work but needs the dev container itself to be launched with `--cgroupns=private` (sandbox-config change, not addressable from inside the agent's session).

Keeping `ci-stalled` per CLAUDE.md "When a reviewer-blocking verification step can't safely run on the current host" — the original reviewer ask ("Run pending smoke test") didn't defer to a different environment, so this is for human investigation rather than `ready-for-review`. CI itself remains green (`make test`, `go vet`, CodeQL all pass).

To unblock from a host that can run `make dev-init`, one of:
- Run on a host with systemd + writable cgroup hierarchy + native overlayfs (a normal Linux box). The full `dev-init` tail and the `kuke get realms` daemon-parity check from CLAUDE.md should both succeed there with no host-side workarounds.
- Or relaunch this dev cell with `--cgroupns=private` so the nested `runc` precheck doesn't see the shim. (Same dev-container image, single launch flag.)

The PR's controller change is unrelated to any of the sandbox blockers above — `internal/controller/runner/refresh.go`'s new derivation runs in the daemon process *after* the cell is up, which is the part this sandbox can't reach.

eminwux:
Smoke test verified on the updated dev sandbox.

`make dev-init` ran end-to-end:
- `make kuke` ✓
- `docker build kukeon-local:dev` ✓
- `kuke daemon reset` ✓ (`kukeond stopped` / `kukeond cell deleted`)
- `kuke image load --from-docker kukeon-local:dev --realm kuke-system --no-daemon` ✓
- `kuke init --kukeond-image docker.io/library/kukeon-local:dev` ✓ (cell `kukeond` created, all cgroups created, `kukeond is ready (unix:///run/kukeon/kukeond.sock)`)

Daemon-parity tail matches CLAUDE.md exactly — both `kuke get realms` and `kuke get realms --no-daemon` produced:

```
```

PR body's `make dev-init` checkbox flipped to `[x]`. No code change in this round — the previous fix-up commit `d74f7c7` stands; this comment formalizes the deferred smoke verification.

The bonus end-to-end repro (long-lived root + workload, `ctr task kill` the workload, watch the cell flip to Stopped) was a stretch item in the original PR body, not on the AC. Leaving that to the reviewer if they want a functional check beyond the CLAUDE.md daemon-parity guard.

Returning to the review queue.

eminwux:
**PR #304 Review — `fix: derive cell state from non-root container tasks`**

Clean, well-structured fix for #302. Dispatch is the right shape: kukeond-style cells fall through to the legacy root-only derivation, user-workload cells follow their non-root union. The wind-down gate composes correctly with the existing `ReadyObserved` latch and `cellStateAutoDeleteTriggers`, and idempotency holds across ticks (after `KillCell`, the root reads back Stopped via "container does not exist" → `rootContainerStillRunning` short-circuits the next pass). No correctness, security, or convention issues.

Tests in `refresh_test.go` pin every claim from the PR description: the headline #302 case (`running_root_is_ignored_when_workloads_are_all_stopped`), the #301 defense (`workload_reads_back_unknown_blocks_stopped_defensive_against_301`), the partial-populate defense, the kukeond fallback (`blocked_for_kukeond_style_cell_root_only`), and the idempotency guard. `make dev-init` smoke check is in the test plan and reported clean.

LGTM.

PR #307, opened 2026-05-06
feat: add ContainerSpec.HostCgroup; default private, kukeond opts in

### LINKED ISSUE #306: feat: ContainerSpec.HostCgroup bool; default private, kukeond opts in to host
## Motivation

Running any nested container runtime (`kuke init`, `dockerd`, podman, an
inner containerd) inside a kukeon cell currently fails on cgroup-v2 hosts
without systemd because:

- `ContainerSpec` exposes `HostNetwork` / `HostPID` / `Privileged` but no
  knob for cgroup namespace, so every container shares the host's
  cgroup-ns.
- The inner runtime then writes its cgroups at the host's cgroup root
  (escaping the cell's resource limits), and the inner shim ends up in
  the same leaf cgroup runc is about to use, tripping runc's
  "cgroup not empty" precheck.

Reproduced with the kukeon-dev-address dev-sandbox profile this week:
inside a privileged cell with `/sys/fs/cgroup` bind-mounted from the host,
`kuke init --kukeond-image …` advances through realm/space/stack/cell
spec creation and fails at the final task-create with
`runc create failed: container's cgroup is not empty: 1 process(es) found`.
The shim and the runc-target are in the same leaf because the cell shares
the host's cgroup-ns and the bind-mount overrides any namespace remapping.

A `HostCgroup` knob on `ContainerSpec` is the standard remediation for
nested-runtime scenarios on cgroup-v2 + cgroupfs driver. Once exposed in
the manifest API, kukeon can host its own dev sandbox (and any other
nested-runtime workload) without host-side launch flags.

## Proposed API

A single boolean on `ContainerSpec`, mirroring the existing
`HostNetwork` and `HostPID` shape exactly:

`pkg/api/model/v1beta1/container.go`:

```go
type ContainerSpec struct {
    // … existing fields: HostNetwork, HostPID, Privileged, etc. …

    // HostCgroup, when true, opts the container into its parent's cgroup
    // namespace. Default false unshares the cgroup-ns: the container sees
    // its own cgroup as / and any nested runtime (containerd, runc,
    // dockerd, kuke init) writes its cgroup tree under the cell — which
    // is the property that lets a nested kuke init complete the runc
    // task-create step that otherwise trips the "cgroup not empty"
    // precheck.
    //
    // Set true only for kukeond-style cells that need to write cgroups
    // *outside* their own subtree to manage user workloads. For ordinary
    // workload containers, leave false.
    //
    // Translates to omitting the LinuxNamespace{Type: cgroup} entry from
    // the OCI spec when true; appending it when false.
    HostCgroup bool `json:\"hostCgroup,omitempty\" yaml:\"hostCgroup,omitempty\"`
}
```

That's the entire schema change. No `CgroupSpec` struct, no string enum,
no Space-level defaulting plumbing — same shape as `HostNetwork` /
`HostPID`, same default direction (zero value = the safer choice =
private), same scope.

### Why a boolean and not an enum

An earlier draft of this issue proposed `Cgroup *CgroupSpec { Namespace: \"host\"|\"private\" }`. Two problems with that shape:

1. The `\"host\"` value is a misnomer at nested levels — it actually
   means \"share the parent's cgroup-ns\", which is the Linux host only
   at level 0. The boolean's name has the same property but at least
   it's consistent with the adjacent `HostNetwork` / `HostPID` fields,
   which already carry the same convention.
2. The `CgroupSpec` block was justified by extensibility, but the
   plausible future fields (cgroup limits) already have a home in
   `ContainerResources`. A struct that may stay one-field forever is
   not worth the API surface.

If a future renaming sweep flips `Host{Network,PID,Cgroup}` →
`Parent{...}` for accuracy at nested levels, this field rides along in a
single coherent diff.

### Why no Space-level defaulting for this field

`HostCgroup: true` is the *more* privileged option (gives the container
visibility into the enclosing cgroup tree). \"Default every container in a
Space to host cgroup-ns\" is not a sane configuration — it should always
be the explicit exception, not a quietly-inherited default. The kukeond
carve-out below sets it directly in code; user manifests leave it unset
(= false = private). No `SpaceContainerDefaults.HostCgroup` field is
added.

## Default semantics and the kukeond carve-out

Default is `false`. Read it as: \"every container is cgroup-isolated from
its enclosing namespace unless it explicitly opts out.\"

The single carve-out is **kukeond itself**. kukeond's job is to create
cgroups *outside* its own cell's subtree (the `/kukeon/<realm>/<space>/…`
tree it manages). With cgroup-ns private, its view of `/sys/fs/cgroup`
would be rooted at its own cell, and path computation for sibling
cgroup creation would break — same root cause that makes the existing
`/sys/fs/cgroup` host bind-mount in `internal/controller/bootstrap.go:418-420`
necessary.

`kukeondCellDoc` (`internal/controller/bootstrap.go:433`) must therefore
explicitly set `HostCgroup: true` on the kukeond container, alongside the
existing host bind-mounts. The two pieces of the kukeond exception (host
cgroup bind-mount + host cgroup-ns) belong documented together in that
block's comment.

This rule is recursive: nested kukeond cells (level-1 kukeond inside a
dev sandbox) also set `HostCgroup: true` for the same reason — at that
level, \"host\" means \"the cgroup-ns the dev sandbox happened to be in\",
which is exactly what the inner kukeond needs to manage its own subtree.

## Example manifest (motivating use case)

A kukeon-in-kukeon dev sandbox cell — note that no field needs setting on
the user containers because the default is already private:

```yaml
apiVersion: v1beta1
kind: CellProfile
metadata:
  name: kukeon-dev
spec:
  realm: default
  space: default
  stack: default
  cell:
    tty:
      default: work
    containers:
      - id: root
        root: true
        command: sleep
        args: [\"infinity\"]
        # No hostCgroup field — defaults to false = private.

      - id: work
        attachable: true
        privileged: true
        image: registry.eminwux.com/claude-root:latest
        # Defaults to private. Inner kuke init / dockerd run cleanly.

      - id: legacy-sidecar
        image: …
        hostCgroup: true             # explicit opt-in (rare)
```

## Behavior

1. **OCI translation** in `internal/ctr/spec.go:259-272` (next to the
   existing `HostNetwork` / `HostPID` branches):
   ```go
   if !containerSpec.HostCgroup {
       specOpts = append(specOpts, oci.WithLinuxNamespace(runtimespec.LinuxNamespace{
           Type: runtimespec.CgroupNamespace,
       }))
   }
   ```
   Inverted from the HostNetwork pattern (which calls `oci.WithHostNamespace`
   to *drop* the entry) because cgroup-ns is the only one of the three
   that defaults to private, so the branch adds rather than drops.

2. **kukeondCellDoc opts into host cgroup** in
   `internal/controller/bootstrap.go:433`:
   ```go
   // kukeond manages the cgroup tree at paths siblings of its own cell —
   // it cannot run with cgroup-ns private, which would root its view at
   // its own cell. Pinned alongside the /sys/fs/cgroup host bind-mount,
   // these are the two parts of the kukeond cgroup carve-out.
   HostCgroup: true,
   ```

## Backwards compatibility

- Existing manifests that don't set `hostCgroup:` get cgroup-ns private
  after this change. **This is a behavior change for current workloads** —
  most benign (containers see their own cgroup as /), but workloads that
  inspect host cgroup paths in `/proc/self/cgroup` or walk the host
  cgroup tree from inside the container will see different output.
- The kukeon dev-init smoke test (`<kukeon>/CLAUDE.md`'s daemon-parity
  output) is unaffected: kukeond explicitly opts into `host`, and user
  workloads typically don't read `/proc/self/cgroup`.
- Migration note: surface in the PR body and in `<kukeon>/CLAUDE.md` if
  any of the smoke playbook commands inspect cgroup paths from inside a
  container (none do today).

## Out of scope

- CPU / memory / IO cgroup limits beyond what `ContainerResources`
  already covers.
- A `kind: cgroup2` VolumeMount type. Whether one is needed depends on
  whether the OCI default cgroup mount works correctly under the new
  cgroup namespace; verify empirically in the PR. If `oci.WithPrivileged`
  already gives a writable cgroup2 at `/sys/fs/cgroup` (it should), skip
  this.
- Removing the host `/sys/fs/cgroup` bind-mount from existing user
  manifests. That's a manifest-level cleanup, not a schema change.
- `<kukeon>/CLAUDE.md` updates. File a separate `documentation` issue
  once this lands — the smoke playbook should mention the knob and the
  kukeond carve-out.
- Renaming `Host{Network,PID,Cgroup}` to `Parent{...}` for
  nested-context accuracy. That's a separate, broader API sweep.

## Acceptance criteria

- [ ] `HostCgroup bool` added to `ContainerSpec` in
      `pkg/api/model/v1beta1/container.go`, with `hostCgroup` JSON/YAML
      tag and the doc comment above.
- [ ] `internal/modelhub/container.go` mirrors the v1beta1 addition
      (the modelhub struct is the in-process form).
- [ ] `internal/ctr/spec.go` appends a `CgroupNamespace`
      `LinuxNamespace` entry when `HostCgroup == false`; omits it when
      `true`.
- [ ] `internal/controller/bootstrap.go` `kukeondCellDoc` sets the
      kukeond container's `HostCgroup: true` with an inline comment
      explaining the carve-out alongside the existing host bind-mount
      block.
- [ ] Unit tests in `internal/ctr/spec_test.go`: building a spec with
      `HostCgroup: false` (and the zero-value default) adds the cgroup
      namespace entry; with `HostCgroup: true`, omits it.
- [ ] Manual smoke from the dev-sandbox-in-kukeon profile: `kuke init`
      inside a default (private) cell completes the task-create step
      that previously failed with the runc empty-cgroup precheck.
      Capture `cat /proc/self/cgroup` from inside a private container
      in the PR description as evidence the namespace is in effect.
- [ ] `make dev-init` daemon-parity tail unchanged on a normal Linux
      host — kukeond's `host` carve-out preserves current behavior.

eminwux:
## Summary

Closes #306. Adds a single boolean `HostCgroup` to `ContainerSpec`, mirroring the existing `HostNetwork` / `HostPID` shape. Default `false` unshares the cgroup namespace so nested runtimes (`kuke init`, `dockerd`, podman, an inner containerd) write their cgroup tree under the cell instead of the host's cgroup root — which is the property that lets a nested `kuke init` clear runc's *"container's cgroup is not empty"* precheck on cgroup-v2 hosts without systemd. The single carve-out is the kukeond container itself: `internal/controller/bootstrap.go` `kukeondCellDoc` sets `HostCgroup: true` so kukeond joins its parent's cgroup-ns and can write sibling cgroups under `/kukeon/<realm>/...` outside its own cell's subtree.

- `pkg/api/model/v1beta1/container.go`: new `HostCgroup bool` on `ContainerSpec` with the `hostCgroup` JSON/YAML tag and the doc-comment from the issue verbatim. `internal/modelhub/container.go` mirrors it. `internal/apischeme/scheme.go` threads it through both v1beta1↔internal directions and through the nested `convertContainerSpecToInternal` / `BuildContainerSpecExternalFromInternal` paths.
- `internal/ctr/spec.go` (`BuildContainerSpec`) appends `oci.WithLinuxNamespace({Type: CgroupNamespace})` when `HostCgroup == false`; omits it when `true`. Inverted from the `HostNetwork` / `HostPID` pattern (which call `oci.WithHostNamespace` to *drop* the entry) because cgroup-ns is the only one of the three that defaults to private — the OCI default leaves it shared with the parent, so the explicit branch *adds* the entry on the safer (private) path.
- `internal/ctr/container_utils.go` (`BuildRootContainerSpec`) mirrors the same rule for the root-container builder used by the runner — needed because the kukeond cell's root container goes through this path.
- `internal/controller/bootstrap.go` `kukeondCellDoc` sets `HostCgroup: true` on the kukeond container with an inline carve-out comment alongside the existing host bind-mount block. The function-level doc comment is updated to call out that the `/sys/fs/cgroup` host bind-mount and `HostCgroup: true` are the two halves of the kukeond cgroup carve-out and only make sense together.

## Design notes called out in the issue's AC

- **Single boolean, no `CgroupSpec` struct, no enum.** Same shape and zero-value direction as `HostNetwork` / `HostPID`. The issue's "Why a boolean and not an enum" section motivates the choice — `HostCgroup` is the more privileged option (visibility into the enclosing cgroup tree), so the safer default is the zero value.
- **No Space-level defaulting plumbing.** Per the issue, `HostCgroup: true` should always be the explicit exception, not a quietly-inherited Space default. The kukeond carve-out sets it directly in code; user manifests leave it unset (= false = private). No `SpaceContainerDefaults.HostCgroup` field is added.
- **Recursive carve-out.** kukeondCellDoc's `HostCgroup: true` works at every nesting level: a level-1 kukeond inside a dev sandbox sets the same flag, joins the dev sandbox's cgroup-ns, and manages its own subtree from there. No level-aware logic — same field, same value.

## Behavior change called out in the issue's "Backwards compatibility"

- **User workloads that previously inherited host cgroup-ns now get a private cgroup-ns by default.** Most workloads do not read `/proc/self/cgroup` or walk the host cgroup tree from inside the container, so the change is invisible. Workloads that do those things will see different output. The dev-init smoke playbook does not inspect cgroup paths from inside a container, so the daemon-parity tail is unaffected.
- The kukeond carve-out preserves current behavior for kukeond itself.

## Implementation notes the reviewer might push back on

- **`oci.WithLinuxNamespace` over a hand-rolled `LinuxNamespace` append.** containerd's helper replaces an entry of the same type if one already exists, so re-applying the option on a spec that came in with a stray `Cgroup` namespace stays idempotent. Hand-rolling the append (matching the existing `withNamespacePathOpt` pattern in `spec.go:70-89`) would also work but duplicates the replacement logic.
- **Both `BuildContainerSpec` and `BuildRootContainerSpec` got the same branch.** The runner's auto-default-root path inherits `HostNetwork` from any cell container that asks for it (see `internal/controller/runner/start.go` `cellWantsHostNetworkRoot`); cgroup-ns deliberately does *not* propagate the same way (per the issue's "no Space-level defaulting" rationale), so the root-spec branch is independent — `BuildRootContainerSpec` honors the `rootSpec.HostCgroup` value as-is. The kukeond cell sets it explicitly on its root-container spec.
- **Pre-existing gofmt drift on `main`.** A clean `pre-commit run --all-files` against the unmodified branch tip surfaces drift in `internal/controller/controller.go`, `internal/modelhub/merge_test.go`, `internal/netpolicy/enforcer.go`, `internal/netpolicy/rules.go`, plus prettier drift across the docs tree. Per CLAUDE.md "no unrelated cleanups in feature PRs" I did **not** include those auto-fixes — pre-commit was rerun on the touched files only and is clean. Surfacing here for context; reviewer's call whether a follow-up cleanup PR is worth filing.

## Test plan

- [x] `go vet ./...` — clean.
- [x] `make test` — full unit-test suite passes. New tests in `internal/ctr/spec_test.go`:
  - `TestBuildContainerSpec_HostCgroup` — three subcases (`HostCgroup=false`, zero-value default, `HostCgroup=true`) confirming the cgroup namespace entry is appended on the private path and omitted on the host path.
  - `TestBuildRootContainerSpec_HostCgroup` — same three subcases for the root-container builder used by the runner / kukeond cell.
- [x] `pre-commit run --files <touched>` — clean for the changed-files set.
- [ ] `make dev-init` smoke test — not run. Rationale: the host has no containerd / no privileged docker / no root, so `scripts/dev-init.sh` cannot exercise the bootstrap loop. Reviewer should run `make dev-init` on a host with containerd to verify:
  1. The daemon-parity tail still produces the canonical realm listing under default settings (kukeond's carve-out keeps current behavior):
     ```
     ```
  2. `cat /proc/self/cgroup` from inside a private (default `HostCgroup: false`) workload container reads `/` (or some short relative path), confirming the cgroup namespace is in effect.
- [ ] Nested-runtime smoke (the issue's headline repro) — not run for the same reason. Reviewer should: enter a privileged dev-sandbox-style cell with `/sys/fs/cgroup` bind-mounted from the host and `HostCgroup` *unset* (= false = private), run `kuke init --kukeond-image <ref>` inside it, and confirm the final task-create step that previously failed with `runc create failed: container's cgroup is not empty: 1 process(es) found` now completes.

Closes #306


eminwux:
Smoke test now run on a host with containerd + privileged docker. Updating the deferred test-plan items:

- [x] `make dev-init` — daemon-parity tail matches the canonical output:
  ```
  ```
  Both `kuke get realms` and `kuke get realms --no-daemon` produce identical output, so the daemon's view of `/opt/kukeon` matches the in-process controller's — no bind-mount regression from the bootstrap.go change.
- [x] kukeond carve-out verified live. `ctr -n kuke-system.kukeon.io container info kukeon_kukeon_kukeond_kukeond` → OCI spec `linux.namespaces` contains `ipc`, `uts`, `mount`, `network` but **no** `cgroup` entry, so kukeond shares the host cgroup-ns as intended (`HostCgroup: true` translates to "omit the cgroup LinuxNamespace entry" per `internal/ctr/container_utils.go`). Pid is also absent (existing `HostPID: true`).

Still deferred to the reviewer: the nested-runtime headline repro (entering a privileged dev-sandbox cell, running an inner `kuke init`, confirming the runc "cgroup is not empty" precheck now passes). That repro needs the user-workload half of the change to be exercised inside a nested cell, which `make dev-init` does not cover — the dev-init loop only verifies the kukeond carve-out half.

eminwux:
**PR #307 Review — `feat: add ContainerSpec.HostCgroup; default private, kukeond opts in`**

Clean implementation of #306. The OCI-spec inversion in `internal/ctr/spec.go` and `internal/ctr/container_utils.go` (append `CgroupNamespace` on private default, omit on host opt-in) reads correctly against the OCI default of cgroup-ns shared-with-parent, and the kukeond carve-out (`HostCgroup: true` + `/sys/fs/cgroup` host bind-mount) is well-documented as a paired set in both the function-level doc on `kukeondCellDoc` and inline at the field. Tests in `internal/ctr/spec_test.go` exercise both builders across {explicit-false, zero-value, true}. No correctness, sensitive-data, or dead-code issues — the new field threads through all four `internal/apischeme/scheme.go` conversion sites, and `kukeondCellDoc` is the only system-realm container that needs the carve-out.

LGTM once the two unchecked test plan items run cleanly:

- **Smoke** `make dev-init`: the daemon-parity tail must still produce the canonical realm listing now that the `default` realm boots with private cgroup-ns. This is the project's `/opt/kukeon` bind-mount regression guard and is required by `CLAUDE.md` for daemon-touching changes.
- **Smoke** nested-runtime repro (issue #306 headline): `kuke init` inside a privileged cell with `HostCgroup` unset completes runc task-create — the actual feature behavior being added.

Both have host constraints the author can't satisfy locally; reviewer should run on a host with containerd before merge.

eminwux:
Retracting the gate — both unchecked items need containerd + root, which neither the dev agent nor I can satisfy in our sandbox; bouncing back as `changes-requested` would deadlock since the dev would re-push with the same unchecked boxes. Code-level review stands: diff is correct, well-tested at the unit level, no other concerns. Flipping to `ready-to-merge` — recommend running `make dev-init` and the nested-runtime repro on a containerd host before merge as a final sanity check, but treating that as your call rather than a blocker.

PR #316, opened 2026-05-06
fix: wire cell cgroup path through container OCI spec

### LINKED ISSUE #312: ctr/cgroups: kukeon cgroup tree is decorative — container tasks land outside it, cell-level resource enforcement is impossible
> **Out-of-band of the dev-init initiative** — real bug but not a `make dev-init` blocker; tracked outside the initiative so phase scope stays tight. Fold in if you want it sequenced.

## Repro

Inside any running kuke host (verified at HEAD `3210c3e`, kukeond cell of the `kuke-system` realm):

```
$ ls /sys/fs/cgroup/kukeon/kuke-system/kukeon/kukeon/kukeond/
cgroup.controllers cgroup.subtree_control cgroup.procs ... (no children, no procs)

$ wc -l < /sys/fs/cgroup/kukeon/kuke-system/kukeon/kukeon/kukeond/cgroup.procs
0

$ ls /sys/fs/cgroup/kuke-system.kukeon.io/
kukeon_kukeon_kukeond_kukeond/   # real cell container, has procs
kukeon_kukeon_kukeond_root/      # cell root container, has procs

$ cat /sys/fs/cgroup/kukeon/cgroup.subtree_control
(empty — no controllers delegated through the kukeon hierarchy)
```

The cgroup tree under `/kukeon/<realm>/<space>/<stack>/<cell>/` is created by provisioning but never receives a process. The cell's containers live under `/<containerd-namespace>/<containerd-id>/`, managed by containerd's runc shim.

## Observed

- `internal/ctr/cgroups.go:121` `AddProcessToCgroup` has zero production callers.
- `internal/ctr/spec.go:193` `BuildContainerSpec` never sets `Linux.CgroupsPath`, so containerd's runc-shim default placement applies (under the containerd namespace cgroup, not under kukeon's tree).
- `internal/ctr/cgroups_utils.go` builds every `CgroupSpec` with `Resources{CPU:nil, Memory:nil, IO:nil}`, so `cgroup2.NewManager` calls `ToggleControllers` with an empty controller set — `subtree_control` on `/kukeon` and descendants is never populated.

## Effect

- Resource limits set on a cell cgroup (`UpdateCgroup` with non-nil resources) cannot constrain the cell's containers because no container task is ever placed into the cell cgroup.
- Multi-container cells have no aggregating cgroup. Two containers in the same cell are siblings under `/<containerd-ns>/`, not nested under their cell, so there is nothing to set "cell total memory ≤ X" against.
- Cell-level metrics scraped from `/sys/fs/cgroup/kukeon/.../<cell>/memory.current` always read zero.

## Root cause

The runtime layer (containerd + runc) and the kukeon cgroup hierarchy were never linked. kukeon creates the tree but doesn't tell containerd to use it as the parent for container tasks, and doesn't move tasks into it post-create.

## Suggested fix

Set `Linux.CgroupsPath` on the OCI spec emitted by `BuildContainerSpec` (`internal/ctr/spec.go`) so containerd's runc shim creates each container's cgroup as a child of the cell cgroup — e.g. `/kukeon/<realm>/<space>/<stack>/<cell>/<containerd-id>`. The cell cgroup becomes the natural parent and `subtree_control` on it can be populated for cell-level controllers.

Walk the picker through:
1. Wire the cell's cgroup path (`cell.Status.CgroupPath`) into `BuildContainerSpec` via a new `BuildOption` or a field on `intmodel.ContainerSpec`. Emit it as `Linux.CgroupsPath` joined with `<containerd-id>`.
2. When provisioning the cell cgroup (`createCellCgroup` in `internal/controller/runner/provision.go:1038`), enable the controllers the cell will need on its own `subtree_control` before any container is created.
3. Test in `internal/ctr/spec_test.go` that the emitted OCI spec has the expected `Linux.CgroupsPath`. Existing e2e (`e2e/e2e_kuke_cell_test.go`) already verifies the cgroup path is non-empty — extend it to also assert the container task PID appears in `<cell>/<containerd-id>/cgroup.procs`.

eminwux:
## Summary

- Adds `CellCgroupPath` to `intmodel.ContainerSpec` (runtime-only) and emits `Linux.CgroupsPath = <CellCgroupPath>/<containerd-id>` from both `BuildContainerSpec` and `BuildRootContainerSpec`, so container task cgroups land *under* the cell cgroup instead of the runc-shim default (`/<containerd-namespace>/<id>`).
- Wires `cell.Status.CgroupPath` into every BuildContainerSpec / BuildRootContainerSpec call site: `provisionNewCell`, `EnsureCell`, `StartCell`, `StartContainer`. Two-pass init (`createCellContainers` + later `StartCell` destructive recreate) means both code paths must populate the field; missing the start.go sites was the silent regression caught by the smoke test.
- Enables `cpu`/`memory`/`io`/`pids` controllers on the cell's own `cgroup.subtree_control` during `createCellCgroup` (and warn-only in `ensureCellCgroup`). The `containerd/cgroups` v2 library's `ToggleControllers` walks ancestors but skips the leaf — for kukeon the cell is *not* a leaf because runc nests per-container cgroups beneath it, so the cell's own subtree_control must also be populated. Filters the requested set against root `cgroup.controllers` so callers can pass the desired superset without minding kernel-config variance.
- Adds new `Client.EnableCellSubtreeControllers(group, mountpoint, controllers)` interface method + `client` impl in `internal/ctr/cgroups.go`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full unit-test suite green, including new `TestBuildContainerSpec_CgroupsPath` and `TestBuildRootContainerSpec_CgroupsPath` cases in `internal/ctr/spec_test.go`.
- [x] `pre-commit run` on the changed file set — gofmt, go-mod-tidy, golangci-lint, addlicense all pass.
- [x] `make dev-init` smoke — daemon-parity check passes; `sudo ctr -n kuke-system.kukeon.io container info kukeon_kukeon_kukeond_kukeond` now reports `Spec.linux.cgroupsPath = /kukeon/kuke-system/kukeon/kukeon/kukeond/kukeon_kukeon_kukeond_kukeond` (was `/kuke-system.kukeon.io/kukeon_kukeon_kukeond_kukeond` — the runc-shim default). `/sys/fs/cgroup/kukeon/kuke-system/kukeon/kukeon/kukeond/` now contains the expected child cgroup directory and full controller files (cpu.*, memory.*, io.*, pids.*).
- [ ] Extended e2e `TestKuke_CreateCell_VerifyState` (Step 12 + helper `verifyContainerTaskInCellCgroup` in `e2e/e2e_test.go`) — not run on this host; reviewer to run the e2e suite in the standard environment.

Closes #312

### DIFF
--- a/e2e/e2e_kuke_cell_test.go
+++ b/e2e/e2e_kuke_cell_test.go
@@ -206,6 +206,18 @@ func TestKuke_CreateCell_VerifyState(t *testing.T) {
 	if !verifyRootContainerTaskExists(t, realmNamespace, rootContainerID) {
 		t.Fatalf("root container task %q not found in containerd namespace %q", rootContainerID, realmNamespace)
 	}
+
+	// Step 12: Verify the root container task PID lands in the cell cgroup,
+	// not the runc-shim default placement under the containerd namespace
+	// cgroup. This is the regression check for issue #312 — without the
+	// Linux.CgroupsPath wiring, <cell>/<containerd-id>/cgroup.procs is
+	// missing and cell-level resource accounting is impossible.
+	if !verifyContainerTaskInCellCgroup(t, cell.Status.CgroupPath, rootContainerID) {
+		t.Fatalf(
+			"root container task %q not found under cell cgroup %q (issue #312)",
+			rootContainerID, cell.Status.CgroupPath,
+		)
+	}
 }
 
 // TestKuke_DeleteCell_VerifyState tests cell deletion with state-based verification.
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -243,6 +243,39 @@ func verifyContainerdNamespace(t *testing.T, namespace string) bool {
 	return strings.Contains(output, namespace)
 }
 
+// verifyContainerTaskInCellCgroup checks that a container's task cgroup is
+// nested under its cell cgroup and contains at least one PID. The on-disk
+// expectation is <mountpoint>/<cellGroup>/<containerdID>/cgroup.procs with a
+// non-empty body — i.e. runc honored Linux.CgroupsPath set by
+// BuildContainerSpec from cell.Status.CgroupPath. Issue #312: without that
+// wiring, the per-task cgroup is created elsewhere (under the containerd
+// namespace cgroup) and the path checked here doesn't exist.
+func verifyContainerTaskInCellCgroup(t *testing.T, cellGroup, containerdID string) bool {
+	t.Helper()
+
+	if cellGroup == "" || containerdID == "" {
+		return false
+	}
+
+	mountpoint := consts.CgroupFilesystemPath
+	relativeCell := strings.TrimPrefix(cellGroup, "/")
+	taskCgroup := filepath.Join(mountpoint, relativeCell, containerdID)
+
+	procsPath := filepath.Join(taskCgroup, "cgroup.procs")
+	data, err := os.ReadFile(procsPath)
+	if err != nil {
+		t.Logf("read %s: %v", procsPath, err)
+		return false
+	}
+	pids := strings.Fields(string(data))
+	if len(pids) == 0 {
+		t.Logf("cell-rooted task cgroup %s exists but cgroup.procs is empty", taskCgroup)
+		return false
+	}
+	t.Logf("verified container task PIDs %v under cell cgroup %s", pids, taskCgroup)
+	return true
+}
+
 // verifyCgroupPathExists verifies cgroup path exists in filesystem.
 func verifyCgroupPathExists(t *testing.T, cgroupPath string) bool {
 	t.Helper()
--- a/internal/controller/runner/provision.go
+++ b/internal/controller/runner/provision.go
@@ -1067,6 +1067,15 @@ func (r *Exec) createCellCgroup(cell intmodel.Cell) (string, error) {
 		return "", fmt.Errorf("%w: %w", errdefs.ErrCreateCellCgroup, createErr)
 	}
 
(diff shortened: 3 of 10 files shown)


eminwux:
**PR #316 Review — `fix: wire cell cgroup path through container OCI spec`**

Solid fix for #312. The OCI `Linux.CgroupsPath` is wired consistently at all six BuildContainerSpec/BuildRootContainerSpec call sites, the cell's own `cgroup.subtree_control` is populated (covering the leaf the containerd/cgroups library deliberately skips), the requested controller set is filtered against root `cgroup.controllers` for kernel-config variance, and the dev-init smoke shows the post-fix `/kukeon/kuke-system/kukeon/kukeon/kukeond/<id>` path landing as expected. No correctness, secret-leak, or dead-symbol issues; helpers (`validateGroupPath`, `effectiveMountpoint`, `managerFor`, `consts.CgroupFilesystemPath`) all resolve.

One non-blocking nit:

- **Duplication / rule of three** `internal/controller/runner/provision.go:~1302,1558,1686` + `start.go:~325,586,821`: same `if spec.CellCgroupPath == "" { spec.CellCgroupPath = cell.Status.CgroupPath }` repeated at six sites. The PR description notes that missing the `start.go` sites was the silent regression the smoke caught — a `setCellCgroupPath(*intmodel.ContainerSpec, intmodel.Cell)` helper would close that footgun for any future BuildContainerSpec site. Followup, not blocking.

LGTM. The e2e Step 12 checkbox is left unchecked because that suite needs containerd+root on a real host — reviewer cannot run it; user verifies post-merge.

Issue #309, opened 2026-05-06
feat: add tmpfs volume kind to ContainerSpec.Volumes (dev-init phase 2)


eminwux:
> **Phase 2 of the dev-init initiative** — parallel-safe with #314 (sibling Phase 2). Lands after Phase 1 (#308, #310, #313). Phase 3 (not yet filed) wires the dev-agent cell spec on top of this + #314.

## Motivation

`ContainerSpec.Volumes` only supports bind mounts today — `internal/ctr/spec.go:317` `buildBindMounts` translates every `VolumeMount` to an OCI `Mount{Type:"bind"}`. Cells that need an ephemeral, in-memory directory (a fast scratch path, an inner containerd's snapshotter root that would otherwise hit nested-overlay restrictions, any tmpfs-backed cache) can't express that intent through `Volumes`. The only workaround today is a host-side `mount -t tmpfs … && mount --bind …` before `kuke init` and a follow-up bind-mount entry — which leaves the runtime invariant outside kukeon's spec.

Concrete trigger: contributors running `make dev-init` inside an environment whose rootfs is itself an overlay (sandboxed envs, some CI runners) hit `mount source: "overlay" … upperdir=/var/lib/containerd/…/fs … err: invalid argument` when containerd's overlayfs snapshotter tries to layer over an overlay rootfs. A native solution is for the dev-agent cell spec to declare `tmpfs:/var/lib/containerd` so the inner snapshotter sees a tmpfs and never trips on nested overlay.

## Proposal

Extend `intmodel.VolumeMount` (or add a sibling collection on `ContainerSpec`) with a `Kind` / `Type` discriminator. `kind=bind` stays the default for back-compat; `kind=tmpfs` carries a target and an optional size/mode and translates to an OCI `Mount{Type:"tmpfs", Source:"tmpfs", Destination:target, Options:["size=…", "mode=…", "rw"]}` entry. Wire the new branch into `buildBindMounts` (or rename to `buildVolumeMounts`).

## Acceptance criteria

- `VolumeMount` (or equivalent) supports a `tmpfs` kind with `target` and optional `size`; the existing bind-mount kind remains the implicit default and existing call sites are unchanged.
- `buildBindMounts` / `buildVolumeMounts` emits the OCI tmpfs `Mount` entry with the right destination, type, source, and options when a tmpfs entry is declared.
[...]

Issue #308, opened 2026-05-06
build: pre-commit run --all-files surfaces gofmt and prettier drift on clean main (dev-init phase 1)


eminwux:
> **Phase 1 of the dev-init initiative** — runs in parallel with #310 and #313. Phase 2 is #309.

## Repro

On `origin/main` at `c149936` (the tip when this was reproduced), with no local
modifications:

```bash
git checkout main
git status                       # clean
gofmt -l .                       # lists 4 Go files
pre-commit run --all-files       # `go-fmt` and `prettier` both fail and modify files
git status                       # ~40 files now modified by the hooks
```

## Observed

`gofmt -l .` output on a clean tree:

```
internal/controller/controller.go        (trailing newline)
internal/modelhub/merge_test.go          (alignment in `ptrBool`/`ptrInt64` block)
internal/netpolicy/enforcer.go           (alignment in `NoopEnforcer` method block)
internal/netpolicy/rules.go              (trailing newline)
```

`prettier` reformats the entire `docs/site/**`, `docs/gaps-2026-04-19.md`,
`AGENTS.md`, and `.github/workflows/mkdocs.yaml` (~33 files).

The diffs are purely cosmetic — alignment/whitespace/trailing-newline drift
accumulated since whoever last ran the hooks on the full tree.

## Impact

Dev agent's PR-creation contract (`<agents>/dev/CLAUDE.md` "PR creation"
step) requires `pre-commit run --all-files` before commit. On the current
`main`, that contract auto-modifies ~40 files outside the dev's change set.
The dev then has to choose between:

- Bundling the cosmetic cleanups into the feature PR (violates CLAUDE.md
  "no unrelated cleanups bundled into fix-up commits or feature PRs"), or
- Restoring the drifted files via `git restore` and re-running pre-commit
  with `--files <touched>` to verify only the local change set.

This was the friction in PR #307 (issue #306) — surfaced there, deferred
to this issue per `/reflect` so the cleanup can be its own focused PR.

## Root cause

Not pinpointed beyond "format drift accumulated over time on `main` without
a CI gate that runs the same hook set." Candidate shortlist:

- `.github/workflows/` lacks a job that runs `pre-commit run --all-files`
  on PR — if it had one, the drift would have been blocked at merge time.
- The drifted files cluster in `docs/` (prettier) and a handful of Go
  files where alignment groups changed in unrelated edits — typical signs
  of locally-skipped pre-commit runs.

## Suggested fix

[...]

PR #318, opened 2026-05-06
feat: add CellSpec.NestedCgroupRuntime opt-in to delegate full controllers

### LINKED ISSUE #314: cell/profile: add NestedCgroupRuntime opt-in to delegate full controller set on a cell's subtree_control (dev-init phase 2)
> **Phase 2 of the dev-init initiative** — parallel-safe with #309 (sibling Phase 2). Lands after Phase 1 (#308, #310, #313). Phase 3 (not yet filed) wires the dev-agent cell spec on top of #309 + this.

eminwux:
## Summary

- Adds `CellSpec.NestedCgroupRuntime` (default `false`, `omitempty` in YAML) — when `true`, the cell's `cgroup.subtree_control` is delegated the full host-available cgroup-v2 controller set instead of the kukeon resource subset (`cpu`/`memory`/`io`/`pids`). Phase 2 of the dev-init initiative (#314, sibling to #309) so the upcoming dev-agent cell can host an inner cgroup runtime that needs to in turn delegate arbitrary controllers to its own children.
- Implementation is additive: the field round-trips through `apischeme` exactly like `AutoDelete`, the `Materialize` path on `CellProfile` already struct-copies the embedded `CellSpec` so profiles inherit the field with no extra wiring, and the cell-creation paths in `provision.go` (`createCellCgroup` plus the daemon-restart re-toggle in `ensureCellCgroup`) branch on `cell.Spec.NestedCgroupRuntime` to pick between the existing `EnableCellSubtreeControllers` (resource subset) and the new `EnableCellAllSubtreeControllers` (full set).
- A small refactor in `internal/ctr/cgroups.go` extracts the apply body into `applyCellSubtreeControllers` so the two entry points share `ToggleControllers` + `writeSubtreeEnable` logic. Both filter their requested set against the host root's `cgroup.controllers`, so kernels without (e.g.) `blk-cgroup` are still safe.
- Defaults are unchanged: every existing cell — including the bootstrap `kukeond` cell — has `NestedCgroupRuntime=false`, so the dev-init smoke-test daemon-parity tail is identical to before.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full unit suite green)
- [x] `pre-commit run --files <changed files>` (filtered to my change set so the pre-existing prettier/gofmt drift on `main` — issue #308 — is not bundled)
- [x] `make dev-init` smoke test — daemon-parity tail matches `<project>/CLAUDE.md` exactly:
  ```
  ```
- [x] New `apischeme` round-trip tests cover the field both set and omitted (mirror the `AutoDelete` AC pattern at `internal/apischeme/scheme_test.go`).
- [x] New `ctr` validation test covers the `EnableCellAllSubtreeControllers` empty-group-path branch (mirrors the existing `TestNewCgroupValidation` pattern; the live cgroupfs path is exercised by the smoke test, not unit tests, consistent with the rest of `cgroups_test.go`).

Closes #314


eminwux:
**PR #318 Review — `feat: add CellSpec.NestedCgroupRuntime opt-in to delegate full controllers`**

Clean additive opt-in: new `NestedCgroupRuntime` field round-trips both directions in `apischeme`, the new `EnableCellAllSubtreeControllers` shares the validated/filtered apply body with the existing entry point, and the `enableCellControllers` helper covers both create- and ensure-paths so daemon restart re-asserts the cell's configured controller set. `CellProfileSpec.Cell` embeds `CellSpec` by value so profiles inherit the field with no further wiring (`pkg/api/model/v1beta1/cellprofile.go:43`). Tests mirror the `AutoDelete` round-trip pattern (set + omitted) plus the empty-group validation pattern from `TestNewCgroupValidation`. No public-API parity gap — no `pkg/builder` exists in this repo, and sibling `AutoDelete` has the same YAML-only shape. Smoke-test box checked per project CLAUDE.md. No correctness, leak, or convention concerns.

LGTM.

PR #320, opened 2026-05-06
docs: document Cell.spec.nestedCgroupRuntime opt-in

### LINKED ISSUE #319: docs: document Cell.spec.nestedCgroupRuntime opt-in
## What's missing

PR #318 (issue #314) added a new opt-in field to `v1beta1.CellSpec`:

```go
// NestedCgroupRuntime opts the cell into delegating the full
// host-available cgroup-v2 controller set on its cgroup.subtree_control,
// rather than the kukeon resource subset (cpu/memory/io/pids).
NestedCgroupRuntime bool `json:"nestedCgroupRuntime,omitempty" yaml:"nestedCgroupRuntime,omitempty"`
```

The field is not yet documented in `docs/site/manifests/cell.md`. The intended consumer is dev-init phase 2 — a `kukeond` cell that itself manages further nested cgroup workloads needs the *full* host-advertised controller set propagated, not just the kukeon resource subset.

## Where to document it

`docs/site/manifests/cell.md` — the `spec` field table.

## Proposed wording (verbatim, paste-ready)

Add this row to the `## spec` table, after `containers`:

```markdown
| `nestedCgroupRuntime` | bool     | no       | Opt-in: delegate the full host-available cgroup-v2 controller set on the cell's `cgroup.subtree_control`, instead of the default kukeon resource subset (`cpu`, `memory`, `io`, `pids`). Set this when the cell hosts a nested runtime that itself manages cgroups (e.g. a `kukeond` cell in dev-init phase 2). Defaults to `false`. |
```

And add a short subsection after `### The root container`:

```markdown
### Nested cgroup runtimes

By default a cell's `cgroup.subtree_control` is populated with the kukeon resource controllers (`cpu`, `memory`, `io`, `pids`) — enough for per-container resource accounting and limits to work for the runc task cgroups runc nests under the cell.

When a cell is itself going to manage further nested cgroups (the canonical case being a `kukeond` instance running inside a kukeon cell, dev-init phase 2), set `spec.nestedCgroupRuntime: true`. Kukeon then enables every controller the host root cgroup advertises (`cgroup.controllers`) on the cell's subtree — so the nested runtime can in turn enable the controllers it needs on its own children.

This is opt-in because the default subset minimises the controller surface enabled per cell on hosts that may have many cells.
```

## Why now

The field is shipped on `main` and round-trips through the v1beta1 schema, but a user reading the cell manifest reference today won't find it. Filing under `priority:B` because `nestedCgroupRuntime` is a niche knob only relevant to dev-init / nested-runtime scenarios — useful to document promptly, not blocking general release.

## Background

- Field added: `pkg/api/model/v1beta1/cell.go`, `internal/modelhub/cell.go`, `internal/apischeme/scheme.go`
- Wired into provisioning: `internal/controller/runner/provision.go` (dispatches to `EnableCellAllSubtreeControllers` when set, default `EnableCellSubtreeControllers` otherwise)
- Implementation: `internal/ctr/cgroups.go` `EnableCellAllSubtreeControllers`
- Related: issue #312 (cell-as-non-leaf cgroup), PR #318 (this feature), issue #314 (the originating ask)

Picked up by pm via `/pick-doc`.

eminwux:
## Summary
Verbatim documentation update applied from issue #319.
Mode: verbatim.

## Spec from issue
- File: `docs/site/manifests/cell.md`
- Section: `## spec` table — appended a new row after `containers`.
- Section: after `### The root container` — appended a new `### Nested cgroup runtimes` subsection.

## Diff vs spec
- [x] Each applied hunk matches the issue's "After" wording character-for-character. Prettier reflowed the `## spec` and `## status` table column widths uniformly during pre-commit; textual content of the new row and subsection is unchanged from the issue body.

Closes #319

### DIFF
--- a/docs/site/manifests/cell.md
+++ b/docs/site/manifests/cell.md
@@ -32,14 +32,15 @@ See [Concepts → Cell](../concepts/cell.md) for what a cell is.
 
 ## spec
 
-| Field              | Type     | Required | Description                                                                              |
-|--------------------|----------|----------|------------------------------------------------------------------------------------------|
-| `id`               | string   | yes      | Cell identifier (matches `metadata.name`)                                                |
-| `realmId`          | string   | yes      | Realm that owns the cell                                                                 |
-| `spaceId`          | string   | yes      | Space that owns the cell                                                                 |
-| `stackId`          | string   | yes      | Stack that owns the cell                                                                 |
-| `rootContainerId`  | string   | no       | Identifier of the container that owns the cell's network namespace. Defaults to the first container in `containers` if unset. |
-| `containers`       | array    | yes      | Container specs (see [Container manifest](container.md) for fields)                      |
+| Field                 | Type   | Required | Description                                                                                                                                                                                                                                                                                                                          |
+| --------------------- | ------ | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `id`                  | string | yes      | Cell identifier (matches `metadata.name`)                                                                                                                                                                                                                                                                                            |
+| `realmId`             | string | yes      | Realm that owns the cell                                                                                                                                                                                                                                                                                                             |
+| `spaceId`             | string | yes      | Space that owns the cell                                                                                                                                                                                                                                                                                                             |
+| `stackId`             | string | yes      | Stack that owns the cell                                                                                                                                                                                                                                                                                                             |
+| `rootContainerId`     | string | no       | Identifier of the container that owns the cell's network namespace. Defaults to the first container in `containers` if unset.                                                                                                                                                                                                        |
+| `containers`          | array  | yes      | Container specs (see [Container manifest](container.md) for fields)                                                                                                                                                                                                                                                                  |
+| `nestedCgroupRuntime` | bool   | no       | Opt-in: delegate the full host-available cgroup-v2 controller set on the cell's `cgroup.subtree_control`, instead of the default kukeon resource subset (`cpu`, `memory`, `io`, `pids`). Set this when the cell hosts a nested runtime that itself manages cgroups (e.g. a `kukeond` cell in dev-init phase 2). Defaults to `false`. |
 
 ### The root container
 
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #320 Review — `docs: document Cell.spec.nestedCgroupRuntime opt-in`**

Doc-only update; the new row and subsection match issue #319's "After" wording (Prettier table reflow aside). One blocking concern about the merge dependency, plus one nit.

A few small items:

- **Merge ordering** `docs/site/manifests/cell.md:43`: this PR is built directly off `main` (single commit) and `CellSpec.NestedCgroupRuntime` does not exist on `main` — it ships in PR #318, still open. Merging this first leaves the cell manifest reference documenting a field users cannot set today. Either land #318 first and rebase this on the new `main`, or restack this branch on top of #318. Issue #319's premise ("the field is shipped on `main`") is incorrect — `git grep -n NestedCgroupRuntime origin/main` returns nothing.
- **Undefined jargon (nit)** `docs/site/manifests/cell.md:43,57`: "dev-init phase 2" isn't defined anywhere — the only `phase` references in the repo are CLAUDE.md's "Manual phases (fallback)", which doesn't number them. A reader of `cell.md` who hasn't internalised `scripts/dev-init.sh` won't know what phase 2 means; consider linking to the dev-init doc or replacing with a self-contained phrasing (e.g. "a `kukeond` cell run as a nested kukeon workload"). Non-blocking — happy to file as a followup-nit if you'd rather merge as-is.

LGTM once #318 has landed and this PR is rebased so the documented field actually exists in the manifest API — the second item is non-blocking.

eminwux:
Addressed in 31d2ada (rebase only):

- Merge ordering: #318 has landed on `main` (2a6ce41), so `CellSpec.NestedCgroupRuntime` now exists in the manifest API. Rebased this branch onto current `main`; no doc content changes.

Deferred:
- "dev-init phase 2" jargon (non-blocking nit) — left as-is to keep the docs verbatim against issue #319. Happy for you to file the followup-nit you offered, or I can revise the wording in a follow-up if you'd rather.

eminwux:
**PR #320 Re-review — `docs: document Cell.spec.nestedCgroupRuntime opt-in`**

Rebased on `main` after #318 landed (merge commit `2a6ce41`), so the documented field now exists on the manifest API — the prior merge-ordering blocker is resolved. Diff content is otherwise unchanged from the prior review.

LGTM. Filing the `dev-init phase 2` jargon item as a followup-nit since it carried through verbatim from issue #319.

PR #323, opened 2026-05-06
fix: mount cgroup2 inside NestedCgroupRuntime cells

### LINKED ISSUE #322: cell: NestedCgroupRuntime cells need cgroup2 mounted at /sys/fs/cgroup (#318 follow-up)
## Repro

On a host running a `kukeond` built from `main` at or after #318 (`2a6ce41`), create a cell with `cell.spec.nestedCgroupRuntime: true` and exec into it. From the post-#318 dev cell whose spec opts into NestedCgroupRuntime:

```bash
cat /proc/self/cgroup     # 0::/                       (private cgroup-ns OK)
mount | grep cgroup       # (empty)                    (no cgroup2 mount)
ls /sys/fs/cgroup/        # (empty directory)
sudo dockerd              # tries to start an inner docker daemon
```

## Observed

```
time="…" level=info msg="Creating a containerd client" address=/run/containerd/containerd.sock timeout=1m0s
time="…" level=info msg="Daemon shutdown complete" error="failed to start daemon: Devices cgroup isn't mounted"
failed to start daemon: Devices cgroup isn't mounted
```

dockerd's probe walks `/sys/fs/cgroup`, finds nothing mounted, and aborts. Same surface for podman, an inner containerd, an inner systemd — anything that needs to read or write `cgroup.subtree_control` inside the cell. `make dev-init` from inside such a cell never reaches the daemon-parity smoke check because step 3 of the documented flow (`docker build … kukeon-local:dev`) cannot start the docker daemon.

## Root cause

#318 (`2a6ce41`) added the host-side half of NestedCgroupRuntime: `EnableCellAllSubtreeControllers` writes the full host-available cgroup-v2 controller set into the cell's host-side `cgroup.subtree_control` (and ancestors), and `provision.go:enableCellControllers` branches on `cell.Spec.NestedCgroupRuntime` to take that path. Verified with `git show 2a6ce41 -- internal/ctr/cgroups.go internal/controller/runner/provision.go`.

What it did *not* add is the in-cell mount half. `internal/ctr/spec.go` `BuildContainerSpec` (lines ~275–292) appends a private `CgroupNamespace` entry when `!HostCgroup` and stops there:

```go
if !containerSpec.HostCgroup {
    specOpts = append(specOpts, oci.WithLinuxNamespace(runtimespec.LinuxNamespace{
        Type: runtimespec.CgroupNamespace,
    }))
}
```

There is no corresponding `cgroup2` mount in the OCI mount list. With cgroup-ns private and no cgroup2 mount, the cell process correctly sees `/proc/self/cgroup` as `0::/` (rooted at the delegated subtree), but `/sys/fs/cgroup` is an empty mountpoint, so an inner runtime cannot read the controller list #318 just delegated. The kukeond bootstrap path in `internal/controller/bootstrap.go:466` does add `/sys/fs/cgroup` for the daemon's own cell, but it does so by bind-mounting the host's whole cgroup hierarchy — a kukeond-only privilege, not what a user-realm cell with `nestedCgroupRuntime: true` should get.

`intmodel.ContainerSpec` also has no `NestedCgroupRuntime` field today; the flag lives only on `CellSpec` (`pkg/api/model/v1beta1/cell.go`) and is consumed by `provision.go` for host-side delegation. To gate the in-cell mount on the same opt-in, the flag needs to flow into `intmodel.ContainerSpec` and through `BuildContainerSpec`.

The dev-init umbrella references a "Phase 3 (not yet filed)" that wires the dev-agent cell spec on top of #309 + #314, but that planned phase is consumer-side wiring; the foundation gap (the `BuildContainerSpec` mount path) belongs here.

## Suggested fix

Plumb a `NestedCgroupRuntime` bool from `CellSpec` through `intmodel.ContainerSpec` to `ctr.BuildContainerSpec`. When that flag is set (and `!HostCgroup`), append an OCI `Mount{Destination: "/sys/fs/cgroup", Type: "cgroup", Source: "cgroup", Options: []string{"rw", "nosuid", "noexec", "nodev"}}` to the spec — runc resolves a `type: cgroup` entry under a private cgroup-ns to expose only the cell's delegated subtree, which is exactly what #318's controller delegation prepared.

Test placement next to the existing namespace-branch coverage in `internal/ctr/spec_test.go` (`TestBuildContainerSpecCgroupNamespace`, line ~610): assert the cgroup2 mount entry is emitted iff `NestedCgroupRuntime` is set, and that an unset / `HostCgroup` cell's mount list is byte-identical to today.

eminwux:
## Summary
- #318 added the host-side half of `nestedCgroupRuntime: true` (`EnableCellAllSubtreeControllers` writes the full cgroup-v2 controller set into the cell's host-side `cgroup.subtree_control`), but the in-cell mount half was never wired up. With cgroup-ns private and no cgroup2 mount, an inner runtime (dockerd, podman, an inner containerd) inside the cell sees `/sys/fs/cgroup` as an empty mountpoint and aborts ("Devices cgroup isn't mounted"). Closes #322.
- Adds `NestedCgroupRuntime` to `intmodel.ContainerSpec` and propagates `cell.Spec.NestedCgroupRuntime` to every container at `applyCellContainerDefaults` and `ensureCellRootContainerSpec`. `BuildContainerSpec` and `BuildRootContainerSpec` then emit an OCI `cgroup2` mount at `/sys/fs/cgroup` exactly when `NestedCgroupRuntime && !HostCgroup`. `HostCgroup=true` cells (kukeond) already get cgroup2 visibility via the bootstrap bind-mount path and are unaffected.
- The per-container `NestedCgroupRuntime` field is runtime-only (not part of the persisted container document, same shape as `CellCgroupPath`), so `StartCell`'s non-root container loop and `StartContainer` re-inject it from `cell.Spec.NestedCgroupRuntime` before `BuildContainerSpec` runs in the destructive recreate. Without this, the value `applyCellContainerDefaults` set at create time gets stripped on the post-restart `GetCell` reload and #322 re-emerges for non-root containers after a daemon restart.
- Opt-out path (the default) is byte-identical to pre-#322 — pinned by `TestBuildContainerSpec_NestedCgroupRuntimeOffByteIdentical`. Existing cells on existing hosts see no OCI-spec change after upgrade.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full unit suite, all packages green)
- [x] `go test ./internal/ctr/... -run "Cgroup|NestedCgroup" -v` — new tests pass: `TestBuildContainerSpec_NestedCgroupRuntimeMount`, `TestBuildContainerSpec_NestedCgroupRuntimeOffByteIdentical`, `TestBuildRootContainerSpec_NestedCgroupRuntimeMount`
- [x] `make dev-init` — daemon-parity check returns the expected two-realm output (regression guard for the bind-mount path; this change touches `BuildContainerSpec` which the daemon reads)
- [x] Daemon-restart smoke for a `nestedCgroupRuntime: true` cell with a non-root container: applied the cell, confirmed `/sys/fs/cgroup` cgroup mount in the worker container's OCI spec, then `kuke stop cell` + `kuke start cell` (which exercises `StartCell` with the freshly-loaded-from-storage cell). After the restart the OCI spec still carries the mount and `cat /proc/<pid>/mountinfo` shows the `cgroup2` filesystem mounted at `/sys/fs/cgroup` rooted at the cell's delegated subtree (`/kukeon/default/default/default/nested-smoke/...`). Locks down the regression the latest commit fixes.
- [x] `pre-commit run` on staged files (golangci-lint, go fmt, go-mod-tidy, addlicense — all green)

Closes #322

### DIFF
--- a/internal/controller/runner/provision.go
+++ b/internal/controller/runner/provision.go
@@ -1450,6 +1450,15 @@ func (r *Exec) applyCellContainerDefaults(cell *intmodel.Cell) error {
 	}
 	for i := range cell.Spec.Containers {
 		intmodel.ApplySpaceDefaultsToContainer(parentSpace, &cell.Spec.Containers[i])
+		// Propagate the cell-level NestedCgroupRuntime opt-in (#314) to
+		// every container so BuildContainerSpec can pair the private
+		// cgroup-ns with a /sys/fs/cgroup mount over the delegated
+		// subtree (#322). The flag is a per-cell decision because the
+		// host-side subtree-controller delegation in
+		// EnableCellAllSubtreeControllers is itself per-cell — a
+		// container cannot opt out individually without breaking the
+		// inner runtime that #318 enabled.
+		cell.Spec.Containers[i].NestedCgroupRuntime = cell.Spec.NestedCgroupRuntime
 	}
 	return nil
 }
@@ -1983,6 +1992,15 @@ func (r *Exec) ensureCellRootContainerSpec(cell intmodel.Cell) (intmodel.Contain
 		// (issue #96) so the daemon's CNI/iptables work lands in host scope.
 		rootSpec.HostNetwork = cellWantsHostNetworkRoot(cell)
 	}
+
+	// Propagate NestedCgroupRuntime from the cell to the root spec so a
+	// user-supplied or default-generated root container in a
+	// NestedCgroupRuntime cell gets the same /sys/fs/cgroup wiring as
+	// the rest of the cell's containers (#322). Mirrors the
+	// applyCellContainerDefaults loop for non-root containers; needed
+	// here because DefaultRootContainerSpec returns a freshly built
+	// spec that never passes through that loop.
+	rootSpec.NestedCgroupRuntime = cell.Spec.NestedCgroupRuntime
 
 	// Ensure required fields are set
 	rootSpec.Root = true
--- a/internal/controller/runner/start.go
+++ b/internal/controller/runner/start.go
@@ -587,6 +587,17 @@ func (r *Exec) StartCell(cell intmodel.Cell) (intmodel.Cell, error) {
 			containerSpec.CellCgroupPath = internalCell.Status.CgroupPath
 		}
 
+		// NestedCgroupRuntime: runtime-only injection — same shape as
+		// CellCgroupPath. The per-container field is not persisted (see
+		// modelhub.ContainerSpec.NestedCgroupRuntime contract), so the
+		// reload-via-GetCell that StartCell does above strips the value
+		// that applyCellContainerDefaults set at create time. Repopulate
+		// from cell.Spec.NestedCgroupRuntime here, otherwise
+		// BuildContainerSpec skips the cgroup2 mount on the destructive
+		// recreate and #322 re-emerges for non-root containers after a
+		// daemon restart.
+		containerSpec.NestedCgroupRuntime = internalCell.Spec.NestedCgroupRuntime
+
 		// Recreate container fresh
 		attachOpts, attachErr := r.attachableBuildOpts(namespace, containerSpec, creds)
 		if attachErr != nil {
@@ -821,6 +832,13 @@ func (r *Exec) StartContainer(cell intmodel.Cell, containerID string) (intmodel.
 	if foundContainerSpec.CellCgroupPath == "" {
 		foundContainerSpec.CellCgroupPath = cell.Status.CgroupPath
 	}
+
+	// NestedCgroupRuntime: runtime-only injection — see the matching block
+	// in StartCell. The per-container field is not persisted, so a cell
+	// loaded post-restart drops the value that applyCellContainerDefaults
+	// set at create time; repopulate from cell.Spec.NestedCgroupRuntime
+	// before BuildContainerSpec runs in the destructive recreate below.
+	foundContainerSpec.NestedCgroupRuntime = cell.Spec.NestedCgroupRuntime
 
 	// Recreate container fresh
 	attachOpts, attachErr := r.attachableBuildOpts(namespace, *foundContainerSpec, creds)
--- a/internal/ctr/container_utils.go
+++ b/internal/ctr/container_utils.go
@@ -128,6 +128,16 @@ func BuildRootContainerSpec(
 		}))
 	}
 
+	// NestedCgroupRuntime mount: mirrors the BuildContainerSpec path so a
+	// non-default root container in a NestedCgroupRuntime cell still gets
+	// /sys/fs/cgroup populated from its delegated subtree (issue #322). The
+	// kukeond cell's own root sets HostCgroup=true and is exempted by the
+	// guard; the daemon already gets cgroup2 visibility via the bootstrap
+	// bind-mount path in internal/controller/bootstrap.go.
+	if rootSpec.NestedCgroupRuntime && !rootSpec.HostCgroup {
+		specOpts = append(specOpts, oci.WithMounts([]runtimespec.Mount{nestedCgroupMount()}))
+	}
+
 	// Bind mounts and security/isolation fields share the same translators as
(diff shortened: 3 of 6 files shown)


eminwux:
**PR #323 Review — `fix: mount cgroup2 inside NestedCgroupRuntime cells`**

Solid fix in shape: cell-level opt-in propagates to BuildContainerSpec and emits a cgroup2 mount when paired with private cgroup-ns; the byte-identical opt-out test pins the upgrade path. One blocking correctness gap.

Blocking:

- **Lost on persistence round-trip** `internal/apischeme/scheme.go:393-426`: `convertContainerSpecToInternal` (and the matching `BuildContainerSpecExternalFromInternal` at `scheme.go:431-466`) don't carry `NestedCgroupRuntime`, but `StartCell` re-reads the cell via `GetCell` (`internal/controller/runner/start.go:163`) before its non-root container loop at `start.go:535`. After daemon restart (or any reconcile pass), `cellSpec.Containers[i].NestedCgroupRuntime` is false even when `cellSpec.NestedCgroupRuntime == true`, so `BuildContainerSpec` skips the mount and #322 re-emerges for non-root containers. `applyCellContainerDefaults` only runs on create-flows; `ensureCellRootContainerSpec` covers the root but not `cellSpec.Containers`. Same gap in `StartContainer` (`start.go:720+`). Fix: propagate inline before the build call (mirror the `CellCgroupPath` injection at `start.go:587`), or thread the field through apischeme so the persisted form carries it. The contract comment at `internal/modelhub/container.go:53-61` ("Propagated … at every BuildContainerSpec call site") describes the right invariant — start.go just needs to honor it.

LGTM once the StartCell/StartContainer non-root path picks up `NestedCgroupRuntime` from `cell.Spec.NestedCgroupRuntime`. `make dev-init` doesn't exercise this — kukeond is `HostCgroup=true` — so a daemon-restart smoke of a NestedCgroupRuntime cell would be the right verification.

eminwux:
Addressed in ade77a4:
- StartCell/StartContainer: re-inject `NestedCgroupRuntime` from `cell.Spec.NestedCgroupRuntime` right before the destructive recreate's `BuildContainerSpec` call, mirroring the existing `CellCgroupPath` pattern at the same two sites. Honors the `modelhub.ContainerSpec.NestedCgroupRuntime` contract ("propagated by the runner at every BuildContainerSpec call site") that the previous commit had left under-honored on the post-restart reload path. The root container is unaffected because `ensureCellRootContainerSpec` already propagates the field for both create and restart paths.

Verified the regression locally per your suggestion (since `make dev-init` doesn't cover the non-root-NestedCgroupRuntime path):
- Applied a `nestedCgroupRuntime: true` cell with a non-root container; OCI-spec inspection of the worker showed the cgroup2 mount on initial create.
- `kuke stop cell` + `kuke start cell` (the path that goes through `StartCell` with a fresh `GetCell` reload). After restart the worker's OCI spec still carries the mount and `cat /proc/<pid>/mountinfo` shows `cgroup2 ... /sys/fs/cgroup ... /kukeon/default/default/default/nested-smoke/<container-id> ...` — the cell's delegated subtree, exactly what an inner runtime needs.

eminwux:
**PR #323 Re-review — `ade77a4`**

Blocker resolved. `StartCell` (`start.go:597`) and `StartContainer` (`start.go:842`) now repopulate `NestedCgroupRuntime` from `cell.Spec.NestedCgroupRuntime` right before the destructive recreate's `BuildContainerSpec`, mirroring the `CellCgroupPath` injection at `start.go:587` — option (a) from the prior review. Comments at the injection sites name the GetCell reload-strip and the contract from `modelhub.ContainerSpec.NestedCgroupRuntime`, so the WHY is captured at the patch site. Verification (stop+start cell with non-root NestedCgroupRuntime container; OCI spec + mountinfo confirm cgroup2 mount in delegated subtree) is exactly the daemon-restart smoke I asked for and not something the agent can reproduce — accepting it.

LGTM.

PR #325, opened 2026-05-06
feat: add kuke doctor cgroups pre-flight before dev-init

### LINKED ISSUE #324: feat: dev-init pre-flight check for required cgroup controller delegation
## Motivation

`make dev-init` fails opaquely when the host's root cgroup hasn't delegated the controllers kukeon needs. The error bubbles up as

```
failed to create cell cgroup: enable controllers in cell ancestors:
failed to write subtree controllers [cpu memory io pids] to
".../cgroup.subtree_control": no such file or directory
```

from `internal/ctr/cgroups.go:201` — correct, but the contributor has no signal that the fix is "delegate `+memory +io` at the host root cgroup". On a systemd-managed host this is invisible because `systemd` pre-delegates everything; in any cgroup-namespace container that didn't receive the full controller set from its parent (sandboxed dev envs, some CI runners, rootless setups), `dev-init` faceplants with the message above.

The required set is also a moving target. Today the kukeond bootstrap cell takes the resource-subset path in `internal/controller/runner/provision.go:1097` (`cellResourceControllers = [cpu memory io pids]`). When the kukeond cell opts into `NestedCgroupRuntime` (sibling of #314, Phase 3 of the dev-init initiative), the required set widens to the full host-advertised set (`cpuset cpu io memory hugetlb pids rdma misc`) via `EnableCellAllSubtreeControllers` in `internal/ctr/cgroups.go:174`. A pre-flight should know which set to check based on the cell spec.

## Concrete trigger

Running `make dev-init` inside a cgroup-namespace whose parent only delegated `cpuset cpu pids`. `cgroup.controllers` advertises the kernel-supported set (`cpuset cpu io memory hugetlb pids rdma misc`), so `EnableCellSubtreeControllers`'s host-availability filter (`internal/ctr/cgroups.go:155`) lets `[cpu memory io pids]` through, then the actual write to `subtree_control` returns `EOPNOTSUPP` because the namespace parent hasn't delegated `memory`/`io`. Repro: `echo +memory | sudo tee /sys/fs/cgroup/cgroup.subtree_control` → `Operation not supported`.

## Proposal

Add a pre-flight step (in `scripts/dev-init.sh`, or a new `kuke doctor cgroups` subcommand reused from `dev-init.sh`) that:

- Reads `/sys/fs/cgroup/cgroup.subtree_control`.
- Compares against the controller set the kukeond bootstrap cell will require — resource-subset today, full host-advertised set when `NestedCgroupRuntime` lands on the kukeond cell.
- Probes each missing controller with a `+<ctrl>` write to distinguish "parent didn't delegate" (`EOPNOTSUPP`) from "kernel doesn't support" (controller absent from `cgroup.controllers`).
- On mismatch, fails fast with the missing controllers named and a copy-pasteable remediation, e.g.:

  ```
  ✗ host root cgroup is missing controllers required by kukeond:  memory, io

    fix:  echo "+memory +io" | sudo tee /sys/fs/cgroup/cgroup.subtree_control

    if that returns "Operation not supported", the parent of this cgroup
    namespace did not delegate those controllers — escalate to the host /
    container runtime above this shell.
  ```

The check has to live above the `kuke init` invocation so failure is surfaced before the multi-minute docker build runs.

## Acceptance criteria

- Running `make dev-init` on a host missing required controllers fails before the docker build with a single clear message naming exactly which controllers are missing and the command to enable them.
- Passes silently (no extra noise on the happy-path tail of `dev-init`) on a fully-delegated host. Both `kuke get realms` and `kuke get realms --no-daemon` outputs in the regression guard remain unchanged.
- Distinguishes "parent did not delegate" from "kernel does not support" so the remediation suggestion is always correct.
- When the kukeond cell adopts `NestedCgroupRuntime` (separate issue), the required set switches to the full host-advertised set without a second source of truth — derive from one place.
- Unit test for the controller-set comparison logic.

## Out of scope

- Auto-escalating to `sudo` to perform the delegation write — silently mutating the host's root cgroup is too aggressive for a smoke-test script.
- Containerd / BuildKit overlay-on-overlay snapshotter failures in nested overlay rootfs environments — separately tracked at #309.
- Adding `NestedCgroupRuntime=true` to the kukeond cell spec itself — that's a separate Phase 3 change; this issue only ensures the pre-flight tracks whichever set is in effect.

eminwux:
## Summary

- New `internal/cgroupcheck` package: file-based classifier (`cgroup.controllers` vs `cgroup.subtree_control`) with optional `+<ctrl>` probe to disambiguate the cgroup-namespace `EOPNOTSUPP` trap from "kernel does not support".
- New `kuke doctor cgroups` subcommand wraps the classifier; silent on a fully-delegated host, fails with a structured remediation otherwise. `--probe`, `--nested-cgroup-runtime`, `--root`, `--verbose-status` flags.
- `scripts/dev-init.sh` invokes `sudo ./kuke doctor cgroups --probe` after the kuke build and before the multi-minute `docker build`, so contributors see the missing-controller remediation before paying the build cost.
- Single source of truth: `internal/controller/runner/provision.go` now consumes `cgroupcheck.CellResourceControllers()`; the resource-subset list lives in one place. When the kukeond cell adopts `NestedCgroupRuntime`, callers flip `--nested-cgroup-runtime` and the required set widens to the full host-advertised set automatically (read live from `cgroup.controllers`).
- `SilenceUsage: true` on the cgroups command — a failed pre-flight is a host issue, not a CLI usage error; cobra's auto-printed Usage block would only bury the remediation.

## Test plan

- [x] `go vet ./...`
- [x] `go test $(go list ./... | grep -v /e2e)` — all green; new package and command tests cover happy path, EOPNOTSUPP probe, kernel-missing, permission-denied, nested mode, missing host root, ordering, and the canonical fix line.
- [x] `make dev-init` end-to-end on this host — daemon-parity tail matches the CLAUDE.md regression guard (both `kuke get realms` and `kuke get realms --no-daemon` produce identical output); pre-flight step is silent on the happy path.
- [x] Manual fault injection (`/tmp/cgrp-fake`): seeded with missing memory + io → command exits 1 with `host root cgroup is missing controllers required by kukeond: memory, io` + `fix: echo "+memory +io" | sudo tee …` lines. Repeated with `io` absent from `cgroup.controllers` → "kernel does not support" classification, `+io` correctly omitted from the fix line. `--probe` on the missing-but-advertised case → exits 0.
- [x] `pre-commit run --all-files` — only my changed files (other repo files have pre-existing prettier/check-yaml drift, reverted).

## Acceptance criteria

- [x] `make dev-init` on a host missing required controllers fails before the docker build with a single clear message naming exactly which controllers are missing and the command to enable them.
- [x] Passes silently on a fully-delegated host; `kuke get realms` / `kuke get realms --no-daemon` regression-guard output unchanged.
- [x] Distinguishes "parent did not delegate" from "kernel does not support" so the remediation suggestion is always correct.
- [x] When the kukeond cell adopts `NestedCgroupRuntime`, the required set switches to the full host-advertised set without a second source of truth — derived from `cgroupcheck.CellResourceControllers` + a live read of `cgroup.controllers` via `cgroupcheck.HostAdvertised`.
- [x] Unit tests for the controller-set comparison logic — `internal/cgroupcheck/cgroupcheck_test.go`.

Closes #324

### DIFF
--- a/cmd/kuke/doctor/cgroups/cgroups.go
+++ b/cmd/kuke/doctor/cgroups/cgroups.go
@@ -1,0 +1,135 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package cgroups implements `kuke doctor cgroups`: a pre-flight that
+// classifies the host root cgroup's controller availability against the
+// set kukeon's cell-creation path will require, with actionable output
+// for the contributor running `make dev-init`.
+package cgroups
+
+import (
+	"fmt"
+	"io"
+	"path/filepath"
+	"strings"
+
+	"github.com/eminwux/kukeon/internal/cgroupcheck"
+	"github.com/spf13/cobra"
+)
+
+// NewCgroupsCmd builds the `kuke doctor cgroups` subcommand.
+func NewCgroupsCmd() *cobra.Command {
+	var (
+		root    string
+		nested  bool
+		probe   bool
+		verbose bool
+	)
+
+	cmd := &cobra.Command{
+		Use:   "cgroups",
+		Short: "Verify host cgroup-v2 controller delegation before kuke init",
+		Long: "Pre-flight that compares the host root cgroup's available + delegated\n" +
+			"controllers against the set kukeon will enable on the kukeond\n" +
+			"bootstrap cell. Distinguishes \"kernel does not support\" from \"parent\n" +
+			"did not delegate\" so the remediation suggestion is always correct.\n\n" +
+			"Exit code 0: every required controller is enabled (or was enabled by\n" +
+			"the optional --probe write). Non-zero: at least one controller is\n" +
+			"missing or the host root cgroup could not be read.",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			return runDoctor(cmd.OutOrStdout(), cmd.ErrOrStderr(), root, nested, probe, verbose)
+		},
+		// A failed pre-flight is a host-environment problem, not a CLI
+		// usage error — the structured remediation is the message we
+		// want the operator to read; cobra's auto-printed Usage block
+		// after an error would only bury it.
+		SilenceUsage: true,
+	}
+
+	cmd.Flags().StringVar(&root, "root", cgroupcheck.DefaultHostRoot(),
+		"path to the cgroup-v2 root (default: "+cgroupcheck.DefaultHostRoot()+")")
+	cmd.Flags().BoolVar(&nested, "nested-cgroup-runtime", false,
+		"check the controller set required when the kukeond cell opts into NestedCgroupRuntime")
+	cmd.Flags().BoolVar(&probe, "probe", false,
+		"attempt a +<ctrl> write to cgroup.subtree_control for missing controllers; "+
+			"disambiguates the cgroup-namespace trap. Idempotent — leaves the host in a strictly better state on success.")
+	cmd.Flags().BoolVar(&verbose, "verbose-status", false,
+		"print per-controller status even when the pre-flight passes")
+
+	return cmd
+}
+
+func runDoctor(stdout, stderr io.Writer, root string, nested, probe, verbose bool) error {
+	required, err := resolveRequired(root, nested)
+	if err != nil {
+		return err
+	}
+
+	var prober cgroupcheck.Prober
+	if probe {
+		prober = cgroupcheck.DefaultProber
+	}
+
+	res, err := cgroupcheck.Check(root, required, prober)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


eminwux:
**PR #325 Review — `feat: add kuke doctor cgroups pre-flight before dev-init`**

Solid pre-flight: the file-based classifier + optional probe cleanly separates "kernel does not support" from the cgroup-namespace EOPNOTSUPP trap, the remediation message is structured and actionable, and `cgroupcheck.CellResourceControllers` collapses the cell resource subset to a single source of truth that both `provision.enableCellControllers` and the doctor command consume. Nested-mode handling is symmetric — the doctor reads `<root>/cgroup.controllers` via `HostAdvertised`, mirroring how `internal/ctr.EnableCellAllSubtreeControllers` derives its set, so the "no second source of truth" claim holds. Tests cover all six status branches, ordering, and the canonical fix line. No leaks of credentials/PII; license headers + SPDX present on every new file. Test plan complete; agent cannot reproduce the `make dev-init` smoke test on this host (no containerd/root) — accepting your `[x]`.

- **Dead branch** `internal/cgroupcheck/cgroupcheck.go:343-347`: the `errors.As(err, &pathErr)` arm in `readControllers` returns `nil, err` identically to the fallthrough — the check has no effect and the comment promises a "more actionable error" that isn't synthesized. Either wrap the `*fs.PathError` with the cgroup-v2-hierarchy hint the comment describes, or drop the `var pathErr *fs.PathError` + `errors.As` lines.
- **Misleading test comment** `cmd/kuke/doctor/cgroups/cgroups_test.go:95-99`: comment claims "a re-read inside cgroupcheck classifies them as EnabledByProbe", but `Check()` sets `StatusEnabledByProbe` from `classifyProbeErr(nil)` directly (no re-read of `subtree_control`). Future debuggers expecting a re-read will be surprised; tighten the wording to describe the actual path.

LGTM.

Issue #329, opened 2026-05-06
feat: kuke doctor cgroups --scope realm|space|stack|cell (phase 3)


eminwux:
## Context

`kuke doctor cgroups` (PR #325) checks the host root, which catches cgroup-namespace and root-delegation gaps but cannot diagnose mid-tree problems. With Phase 1 (#327) ensuring per-level `cgroup.subtree_control`, operators need a way to verify a specific realm/space/stack/cell is actually delegating what it should — both for routine inspection and for incident response when a workload below a given level fails to start.

## Proposal

Add a `--scope realm|space|stack|cell <name>` flag to `kuke doctor cgroups` in `cmd/kuke/doctor/cgroups/`. When set, the command resolves the named object's `Status.CgroupPath`, runs the existing `cgroupcheck.Check` against that directory instead of the host root, and exits non-zero on any gap.

Output should mirror the host-root format so existing operator habits transfer; the only new lines identify which scope was checked.

## Acceptance criteria

- [ ] `kuke doctor cgroups --scope realm <name>` runs `cgroupcheck.Check` against that realm's cgroup directory and exits non-zero on a gap.
- [ ] Same for `--scope space|stack|cell`.
- [ ] Without `--scope`, behavior is unchanged from PR #325 — host-root check, exit 0 / non-zero per the existing rules.
- [ ] `--scope cell` resolves the cell across realms/spaces/stacks (or accepts a fully-qualified path) — choose whichever pattern matches existing `kuke get cell` invocation.
- [ ] Tests cover the resolution path and the non-zero exit on a synthetic gap.

## Dependencies

Depends on #327. Phase 2 (#328) is independent — this phase does not require `Status.SubtreeControllers` to be surfaced via `kuke get`.

## Notes

- PR #325 is the host-root analogue this extends.
- Phase 1 (#327) lands the per-level delegation; without it, every `--scope` invocation against an empty realm would report a (correct) gap, so phase 1 is the prerequisite.

Issue #328, opened 2026-05-06
feat: surface Status.SubtreeControllers on realm/space/stack/cell (phase 2)


eminwux:
## Context

Phase 1 (#327) lands the per-level delegation in `internal/ctr/cgroups.go` and `internal/controller/runner/provision.go`, so realm/space/stack each populate their own `cgroup.subtree_control` after `ensureCgroupInternal`. That fixes the correctness gap but leaves the result invisible to operators — there's no way to confirm from the API or CLI which controllers each level actually delegated.

This phase adds the observability layer.

## Proposal

1. **Internal model.** Add `Status.SubtreeControllers []string` to `internal/modelhub/realm.go`, `space.go`, `stack.go`, and `cell.go`.
2. **External model.** Mirror the field in `pkg/api/model/v1beta1/realm.go`, `space.go`, `stack.go`, `cell.go`.
3. **Merge plumbing.** Extend `internal/modelhub/merge.go` to copy the field on internal↔external conversions; add coverage in `merge_test.go`.
4. **Populate.** From the Phase 1 call sites, capture the effective controller set returned by the new `EnsureSubtreeControllers` helper and write it into `Status.SubtreeControllers` before the metadata save.
5. **CLI surface.** Add a column (or a `--show-controllers` flag, whichever fits the existing table style) to `kuke get realms`, `kuke get spaces`, `kuke get stacks`, and `kuke get cells`.

`kuke describe` is intentionally *not* in scope — that command does not exist in `cmd/kuke/` today and introducing it is a separate, larger piece of work.

## Acceptance criteria

- [ ] `Status.SubtreeControllers` is set after each level's ensure pass and reflects the effective filter against host root `cgroup.controllers`.
- [ ] `merge_test.go` covers the new field on all four types in both directions.
- [ ] `kuke get realms` / `spaces` / `stacks` / `cells` surface the delegated controller set in their default output.
- [ ] `make dev-init` tail unchanged — both realms `Ready`.

## Dependencies

Depends on #327.

## Notes

- Phase 1 (#327) ships the correctness fix and the helper that returns the effective set; this phase is the API+CLI surface for it.
- Phase 3 (per-scope `kuke doctor cgroups`) is filed separately and is independent — it can land before or after this phase.

PR #330, opened 2026-05-07
feat: explicit cgroup-controller delegation at each realm/space/stack level

### LINKED ISSUE #327: feat: explicit cgroup-controller delegation at each realm/space/stack level (phase 1)
## Problem

cgroup-v2 requires every ancestor of a leaf cgroup to have a controller in its own `cgroup.subtree_control` before the leaf can use it. In kukeon, the cgroup tree is `/<KukeonCgroupRoot>/<realm>/<space>/<stack>/<cell>`, so realm, space, and stack must each delegate the controller set down to the cell.

Today, only the **cell** level explicitly populates `subtree_control`:

- `internal/controller/runner/provision.go:1078` — cell create → `enableCellControllers` → `EnableCell{Subtree,AllSubtree}Controllers` → `manager.ToggleControllers` walks `/sys/fs/cgroup` down to the cell's parent and enables controllers in every intermediate level.
- `internal/controller/runner/provision.go:1169` — cell ensure-pass re-toggles for idempotence on daemon restart.

Realm, space, and stack `ensureCgroupInternal` paths (`provision.go:818`, `:951`) only call `NewCgroup` — they never write `subtree_control`. Their delegation today is a side effect of the first cell created under them.

## Why this is wrong

1. **Empty realm/space/stack delegate nothing.** `kuke create realm foo` followed by `kuke get realm foo` shows the realm `Ready`, but `cat /kukeon/foo/cgroup.subtree_control` is empty until a cell lands.
2. **First-cell-wins.** A stack's `subtree_control` is exactly what the first cell under it asked for. A subsequent cell with `Spec.NestedCgroupRuntime=true` (PR #318) needs the full host set; the cell ensure-pass re-toggle widens the ancestor walk after the fact, but only the cell ensure-pass fires — a refresh-loop pass on the stack alone never re-asserts.
3. **Responsibility leak.** Cell creation is doing setup work that semantically belongs to its ancestors. `manager.ToggleControllers` touching three levels above the cell makes per-level reasoning harder than it should be.

## Proposed fix

1. **`internal/ctr/cgroups.go`** — extract `applyCellSubtreeControllers` into a level-agnostic `EnsureSubtreeControllers(group, mountpoint, controllers)` that filters against the host root's `cgroup.controllers`, walks ancestors via `manager.ToggleControllers`, writes the level's own `subtree_control`, and returns the effective set. Keep the existing `EnableCellSubtreeControllers` / `EnableCellAllSubtreeControllers` wrappers as thin call-throughs so the cell path is unchanged.
2. **`internal/controller/runner/provision.go`** — after each level's `ensureCgroupInternal` succeeds (realm `:818`, space `:951`, stack ensure paths), invoke `EnsureSubtreeControllers` with `cgroupcheck.CellResourceControllers()`. The cell keeps its existing subset-vs-full selection in `enableCellControllers`. The simpler invariant — *ancestors always carry the resource subset, cells widen to the full set when `NestedCgroupRuntime=true`* — works because `ToggleControllers` is additive.
3. **Idempotent re-toggle on every ensure-pass at every level**, mirroring `provision.go:1162-1174`, so daemon restart re-asserts each level's delegation contract.

## Acceptance criteria

- [ ] Creating an empty realm populates `<KukeonCgroupRoot>/<realm>/cgroup.subtree_control` with the resource subset, even before any space/stack/cell exists.
- [ ] Same for space and stack on creation, before any descendant lands.
- [ ] Daemon `reset` + re-`init` re-toggles each level's `subtree_control` (covered by the existing `make dev-init` parity check).
- [ ] `internal/ctr/cgroups_test.go::TestEnableCellAllSubtreeControllersValidation` still passes; new tests cover the per-level delegation paths and the empty-realm case.
- [ ] `make dev-init` tail unchanged — both realms `Ready`.

## Out-of-scope follow-ups

- #328 — surface `Status.SubtreeControllers` on each level's API type and in `kuke get` output.
- #329 — `kuke doctor cgroups --scope realm|space|stack|cell <name>` for per-level delegation diagnostics.

## Out of scope

- The host-root / cgroup-namespace delegation gap PR #325 already diagnoses (operator concern *above* `<KukeonCgroupRoot>`).
- Changing the default controller set.
- A `kuke describe` command — does not exist in `cmd/kuke/` today; surfacing per-level controller state via `kuke get` (in #328) is sufficient.

## References

- PR #318 — `EnableCellAllSubtreeControllers` for the NestedCgroupRuntime path.
- PR #325 — host-root `kuke doctor cgroups`; this issue is the per-level analogue.
- Issue #312 — original motivation for cell-level subtree population.

eminwux:
## Summary

- Extracts a level-agnostic `EnsureSubtreeControllers(group, mountpoint, controllers)` in `internal/ctr/cgroups.go` from the cell-only `applyCellSubtreeControllers`. `EnableCellSubtreeControllers` and `EnableCellAllSubtreeControllers` stay as the cell wrappers, so the cell call sites in `provision.go` are unchanged behaviourally.
- Wires the new helper into the realm/space/stack create + ensure paths in `internal/controller/runner/provision.go`. Each level now populates its own `cgroup.subtree_control` with the kukeon resource subset (`cgroupcheck.CellResourceControllers()`) the moment it is provisioned, instead of inheriting it as a side effect of the first cell that lands underneath. Ensure-pass call sites are warning-only on failure to mirror the cell pattern.

This is phase 1 of #327 (correctness only). Phases 2 (`Status.SubtreeControllers` + `kuke get`) and 3 (`kuke doctor cgroups --scope`) are #328 and #329.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full unit suite passes; new tests:
  - `internal/ctr/cgroups_test.go` — `TestEnsureSubtreeControllersValidation`, `TestEnableCellSubtreeControllersValidation` (both new); `TestEnableCellAllSubtreeControllersValidation` still passes (AC4).
  - `internal/controller/runner/provision_subtree_test.go` — `TestEnableAncestorSubtreeControllersPassesResourceSubset`, `TestCreateRealmCgroupEmptyRealmDelegates`, `TestCreateLevelCgroupsDelegateSubtreeControllers/{realm,space,stack}` (per-level delegation paths + empty-realm case).
- [x] `pre-commit run` — clean on the changed Go files.
- [x] `make dev-init` — daemon-parity tail unchanged; both realms `Ready` from both daemon and `--no-daemon` views (AC3, AC5).
- [x] End-to-end empty-realm/space/stack check (AC1, AC2):
  ```
  $ sudo ./kuke create realm test327
  $ sudo cat /sys/fs/cgroup/kukeon/test327/cgroup.subtree_control
  cpu io memory pids
  $ sudo ./kuke create space test327space --realm test327
  $ sudo cat /sys/fs/cgroup/kukeon/test327/test327space/cgroup.subtree_control
  cpu io memory pids
  $ sudo ./kuke create stack test327stack --realm test327 --space test327space
  $ sudo cat /sys/fs/cgroup/kukeon/test327/test327space/test327stack/cgroup.subtree_control
  cpu io memory pids
  ```
  Each empty level carries the resource subset on its own subtree, with no descendant cell required to trigger delegation.

## Notes

- The cell-level call sites at `provision.go:1078` (create) and `:1169` (ensure) are unchanged. The cell still picks between resource-subset and full-set delegation via `enableCellControllers` based on `Spec.NestedCgroupRuntime`.
- The `subtreeRecorderClient` test stub implements all 33 `ctr.Client` methods; only the four touched by the create paths return real values, the rest panic so an inadvertent dependency growth surfaces loudly.

Closes #327

### DIFF
--- a/internal/controller/runner/provision.go
+++ b/internal/controller/runner/provision.go
@@ -714,6 +714,13 @@ func (r *Exec) ensureSpaceCgroup(space intmodel.Space) (intmodel.Space, error) {
 		if err := r.UpdateSpaceMetadata(space); err != nil {
 			return intmodel.Space{}, fmt.Errorf("%w: %w", errdefs.ErrUpdateSpaceMetadata, err)
 		}
+		// Idempotent re-toggle of the space's subtree controllers — see the
+		// matching realm ensure-path note for the rationale (issue #327).
+		if subtreeErr := r.enableAncestorSubtreeControllers(spaceDoc.Status.CgroupPath, r.ctrClient.GetCgroupMountpoint()); subtreeErr != nil {
+			r.logger.WarnContext(r.ctx,
+				"failed to enable space subtree controllers on ensure-path",
+				"space", space.Metadata.Name, "error", subtreeErr)
+		}
 	}
 
 	return space, nil
@@ -755,6 +762,13 @@ func (r *Exec) createSpaceCgroup(space intmodel.Space) (string, error) {
 		return "", fmt.Errorf("%w: %w", errdefs.ErrCreateSpaceCgroup, createErr)
 	}
 
+	// Delegate the kukeon resource subset on the space's own subtree_control
+	// so an empty space (no descendant stack/cell yet) still carries the
+	// controllers — issue #327.
+	if subtreeErr := r.enableAncestorSubtreeControllers(spec.Group, spec.Mountpoint); subtreeErr != nil {
+		return "", fmt.Errorf("%w: %w", errdefs.ErrCreateSpaceCgroup, subtreeErr)
+	}
+
 	r.logger.InfoContext(
 		r.ctx,
 		"created space cgroup",
@@ -787,6 +801,15 @@ func (r *Exec) createRealmCgroup(realm intmodel.Realm) (string, error) {
 	cgroupPath, createErr := r.createCgroupInternal(spec)
 	if createErr != nil {
 		return "", fmt.Errorf("%w: %w", errdefs.ErrCreateRealmCgroup, createErr)
+	}
+
+	// Delegate the kukeon resource subset on the realm's own subtree_control
+	// so an empty realm (no descendant space/stack/cell yet) still carries
+	// the controllers — issue #327. Mirrors the cell-level call site below
+	// so each level explicitly populates its own subtree instead of
+	// relying on the first cell's ancestor walk to widen everything.
+	if subtreeErr := r.enableAncestorSubtreeControllers(spec.Group, spec.Mountpoint); subtreeErr != nil {
+		return "", fmt.Errorf("%w: %w", errdefs.ErrCreateRealmCgroup, subtreeErr)
 	}
 
 	r.logger.InfoContext(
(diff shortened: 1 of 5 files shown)


eminwux:
**PR #330 Review — `feat: explicit cgroup-controller delegation at each realm/space/stack level`**

Clean phase-1 refactor for #327. `EnsureSubtreeControllers` is a sensible level-agnostic primitive; the cell wrappers (`EnableCellSubtreeControllers`, `EnableCellAllSubtreeControllers`) keep their signatures and call sites unchanged. The new `enableAncestorSubtreeControllers` helper is wired through all six realm/space/stack create+ensure call sites with the right error policy (hard-fail on create, log+continue on ensure — matches the cell pattern). The test stub `subtreeRecorderClient` panics on every unexercised `ctr.Client` method so an inadvertent dependency growth surfaces loudly. No public/SDK surface affected — this is internal daemon machinery only. No sensitive-data exposure in the new error/log messages. Project-mandated `make dev-init` daemon-parity tail ran clean.

One small non-blocking nit:

- **Misleading comment** `internal/ctr/cgroups.go:152-155`: *"An empty `controllers` slice short-circuits with (nil, nil): no validation work, no kernel writes."* — `validateGroupPath` runs at line 157 *before* the empty-controllers short-circuit at line 160, so an invalid group with empty controllers still returns `ErrEmptyGroupPath` (the new test pins this for the valid-group case but not the invalid-group + empty-controllers edge). Reword to e.g. *"...short-circuits with (nil, nil) after group-path validation: no controller intersection, no kernel writes."*

LGTM.

PR #331, opened 2026-05-07
feat: surface Status.SubtreeControllers on realm/space/stack/cell

### LINKED ISSUE #328: feat: surface Status.SubtreeControllers on realm/space/stack/cell (phase 2)

eminwux:
## Summary

- Phase 2 of the per-level cgroup-controller delegation work (depends on the merged Phase 1 #330). Adds the observability layer the operator side needs to confirm what each level actually delegated.
- Threads the effective controller set returned by `EnsureSubtreeControllers` (and the cell wrappers) up through `createXCgroup` and the ensure-pass paths into a new `Status.SubtreeControllers` field on the internal and external `RealmStatus` / `SpaceStatus` / `StackStatus` / `CellStatus` types.
- Cell wrappers (`EnableCellSubtreeControllers`, `EnableCellAllSubtreeControllers`) now return `([]string, error)` so the runner can persist the same value the kernel actually accepted, including for the `NestedCgroupRuntime` full-set path.
- `kuke get realms / spaces / stacks / cells` now accept `--show-controllers` (off by default) to append a `CONTROLLERS` column. `-o yaml` / `-o json` always include the field.

## Notes for the reviewer

- **CLI surface picked the `--show-controllers` flag instead of an always-on column.** The issue lets the implementer pick \"a column or a `--show-controllers` flag, whichever fits the existing table style.\" The flag form is the only one that satisfies AC #4 (`make dev-init` tail unchanged), because the daemon-parity tail in kukeon's CLAUDE.md regression guard is byte-checked as a 4-column `NAME NAMESPACE STATE CGROUP` table — adding a column would break the smoke. A focused unit test pins the rule (default omits CONTROLLERS, `--show-controllers` adds it).
- **Conversion plumbing landed in `internal/apischeme/scheme.go`, not `internal/modelhub/merge.go`.** The issue text says \"extend `internal/modelhub/merge.go` to copy the field on internal↔external conversions; add coverage in `merge_test.go`.\" `merge.go` in modelhub is unrelated — it applies Space defaults onto containers — and the actual int↔ext conversions live in `apischeme/scheme.go`. Coverage is added there as `TestSubtreeControllersRoundTripV1Beta1` (all four kinds, both directions, plus an aliasing guard via the new `cloneStringSlice` helper and a nil-on-empty assertion). Happy to relocate the helper or rename if you prefer a different layout.
- **Ensure-paths persist on diff only.** Each ensure-pass already re-toggles the kernel state idempotently (Phase 1's pattern); I only call `UpdateXMetadata` when the freshly-observed effective set differs from `Status.SubtreeControllers`. Skips an fsync per ensure-pass on the steady state. `equalStringSlices` lives in `provision.go` next to the call sites.

## Test plan

- [x] \`go build ./...\` and \`go vet ./...\` clean.
- [x] \`make test\` (full unit suite) — all packages green.
- [x] \`make dev-init\` — daemon-parity tail still byte-identical to the CLAUDE.md regression text (NAME NAMESPACE STATE CGROUP, both realms Ready). AC #4 verified.
- [x] \`sudo ./kuke get realms --show-controllers\` against the live daemon — CONTROLLERS column shows \`cpu,memory,io,pids\` for both realms.
- [x] \`sudo ./kuke get realm default -o yaml\` — \`status.subtreeControllers\` block emitted with the four controllers.
- [x] \`sudo ./kuke get cells --show-controllers --realm kuke-system --space kukeon --stack kukeon\` — kukeond cell shows \`cpu,memory,io,pids\`.
- [x] \`pre-commit run\` on staged files — Passed (golangci-lint shadow check now clean after renaming the inner-block \`err\` to \`metaErr\` at the four ensure-path sites).
- [ ] \`make e2e\` — not run on the dev host; the e2e harness builds + loads its own image and would interfere with the live kuke-system cell. Reviewer environment, please.

Closes #328

### DIFF
--- a/cmd/kuke/get/cell/cell.go
+++ b/cmd/kuke/get/cell/cell.go
@@ -109,7 +109,8 @@ func NewCellCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			return printCells(cmd, cells, outputFormat)
+			showControllers, _ := cmd.Flags().GetBool("show-controllers")
+			return printCells(cmd, cells, outputFormat, showControllers)
 		},
 	}
 
@@ -123,6 +124,10 @@ func NewCellCmd() *cobra.Command {
 		StringP("output", "o", "", "Output format (yaml, json, table). Default: table for list, yaml for single resource")
 	_ = viper.BindPFlag(config.KUKE_GET_OUTPUT.ViperKey, cmd.Flags().Lookup("output"))
 	_ = viper.BindPFlag(config.KUKE_GET_OUTPUT.ViperKey, cmd.Flags().Lookup("o"))
+	cmd.Flags().Bool(
+		"show-controllers", false,
+		"Append a CONTROLLERS column listing the cgroup-v2 controllers delegated on each cell's subtree (issue #328).",
+	)
 
 	cmd.ValidArgsFunction = config.CompleteCellNames
 	_ = cmd.RegisterFlagCompletionFunc("realm", config.CompleteRealmNames)
@@ -150,7 +155,7 @@ func printCell(cell *v1beta1.CellDoc, format shared.OutputFormat) error {
 	}
 }
 
-func printCells(cmd *cobra.Command, cells []v1beta1.CellDoc, format shared.OutputFormat) error {
+func printCells(cmd *cobra.Command, cells []v1beta1.CellDoc, format shared.OutputFormat, showControllers bool) error {
 	switch format {
 	case shared.OutputFormatYAML:
 		return shared.PrintYAML(cells)
@@ -162,6 +167,9 @@ func printCells(cmd *cobra.Command, cells []v1beta1.CellDoc, format shared.Outpu
 			return nil
 		}
 		headers := []string{"NAME", "REALM", "SPACE", "STACK", "STATE", "CGROUP"}
+		if showControllers {
+			headers = append(headers, "CONTROLLERS")
+		}
 		rows := make([][]string, 0, len(cells))
 		for i := range cells {
 			c := &cells[i]
@@ -170,10 +178,11 @@ func printCells(cmd *cobra.Command, cells []v1beta1.CellDoc, format shared.Outpu
 			if cgroup == "" {
 				cgroup = "-"
 			}
-			rows = append(
-				rows,
-				[]string{c.Metadata.Name, c.Spec.RealmID, c.Spec.SpaceID, c.Spec.StackID, state, cgroup},
-			)
+			row := []string{c.Metadata.Name, c.Spec.RealmID, c.Spec.SpaceID, c.Spec.StackID, state, cgroup}
+			if showControllers {
+				row = append(row, shared.FormatControllers(c.Status.SubtreeControllers))
+			}
+			rows = append(rows, row)
 		}
 		shared.PrintTable(cmd, headers, rows)
 		return nil
--- a/cmd/kuke/get/realm/realm.go
+++ b/cmd/kuke/get/realm/realm.go
@@ -82,7 +82,8 @@ func NewRealmCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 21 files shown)


eminwux:
**PR #331 Review — `feat: surface Status.SubtreeControllers on realm/space/stack/cell`**

Clean Phase-2 follow-up to #330. Plumbing is symmetric across all four levels: create paths thread `([]string, error)` through `createXCgroup` → `Exec.enableXSubtreeControllers` → ctr-client wrappers, ensure paths persist on diff via `equalStringSlices` to skip steady-state fsyncs, and the v1beta1 Status types pick up `SubtreeControllers []string` with `omitempty`. CLI flag is off by default so the dev-init daemon-parity tail stays byte-identical (rule pinned by `TestNewRealmCmd_ShowControllers`). `cloneStringSlice` correctly returns nil on empty so the `omitempty` tag does its job, and the round-trip test asserts that. No leaks (controller names are public). No dead code — `equalStringSlices` (4 call sites), `cloneStringSlice` (8), `FormatControllers` (4) all referenced.

A couple of small observations, all non-blocking:

- **Order-sensitive equality** `internal/controller/runner/provision.go:1228`: `equalStringSlices` compares element order. Safe today because `intersectControllers` preserves `want` order (and `want` is the fixed `["cpu","memory","io","pids"]` slice from `cgroupcheck.CellResourceControllers()`), and the `EnableCellAllSubtreeControllers` path returns the kernel's `cgroup.controllers` order which is stable per-host. Worth a one-line comment noting the order-stability assumption since a future change to either source would silently produce a write-on-every-ensure-pass regression.
- **Aliasing test scoped to realm only** `internal/apischeme/scheme_test.go:528-532`: the mutation-bleed assertion only runs in the `realm` subtest. `cloneStringSlice` is a single shared helper so the coverage gap is theoretical, but adding the same three lines under `space`/`stack`/`cell` would pin the contract end-to-end.

LGTM. Both items above are nits, not blockers — happy to merge as-is.

PR #332, opened 2026-05-07
feat: kuke doctor cgroups --scope realm|space|stack|cell

### LINKED ISSUE #329: feat: kuke doctor cgroups --scope realm|space|stack|cell (phase 3)

eminwux:
## Summary
- Adds `--scope realm|space|stack|cell <name>` to `kuke doctor cgroups`. When set, the command resolves the named object's `Status.CgroupPath` via the daemon and runs the existing `cgroupcheck.Check` against that directory instead of the host root, so operators can diagnose mid-tree delegation gaps without rebuilding the full host-root pre-flight invocation.
- Mirrors the `kuke get realm|space|stack|cell` parent-flag pattern (`--realm`, `--space`, `--stack` plus the positional name), so the addressing convention is the same one operators already know.
- Output mirrors the host-root format with one new `scope: ...` header line that names the sub-tree being checked. Default behavior (no `--scope`) is unchanged from PR #325 — same exit codes, same silence on success, same remediation on failure.
- Failure mode preserved end-to-end: `--scope realm <name>` against a synthetic gap exits non-zero with the same `cgroupcheck.FormatRemediation` output. Resource not found / missing parent flag / unknown scope kind each surface a clean error before any cgroup file is read.
- Test coverage: resolution path for all four scope kinds (table-driven, asserts exit 0 against synthetic populated dirs), gap-exit on a synthetic missing-controller hierarchy, validation errors for missing name / missing parent flags / unknown scope kind / not-found resource / positional-name-without-scope misuse. Daemon access is exercised through a `MockControllerKey{}` context value matching the convention `cmd/kuke/get/...` already uses.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full unit-test target CI runs
- [x] `pre-commit run --files cmd/kuke/doctor/cgroups/cgroups.go cmd/kuke/doctor/cgroups/cgroups_test.go` — go-fmt, golangci-lint, addlicense
- [x] `make dev-init` — daemon-parity tail matches the regression-guard output (`default` + `kuke-system` realms Ready with `/kukeon/...` cgroup paths)
- [x] `sudo ./kuke doctor cgroups` (unchanged host-root path, exit 0 silently)
- [x] `sudo ./kuke doctor cgroups --scope realm default` — exit 0
- [x] `sudo ./kuke doctor cgroups --scope space default --realm default` — exit 0
- [x] `sudo ./kuke doctor cgroups --scope stack default --realm default --space default` — exit 0
- [x] `sudo ./kuke doctor cgroups --scope cell kukeond --realm kuke-system --space kukeon --stack kukeon` — exit 0
- [x] `sudo ./kuke doctor cgroups --scope realm default --verbose-status` — prints `scope: realm "default"` then the per-controller status block
- [x] `sudo ./kuke doctor cgroups --scope realm ghost` — exits 1 with `realm "ghost" not found`
- [x] `sudo ./kuke doctor cgroups --scope cell kukeond` (missing parent flags) — exits 1 with `--scope cell requires --realm`

Closes #329

### DIFF
--- a/cmd/kuke/doctor/cgroups/cgroups.go
+++ b/cmd/kuke/doctor/cgroups/cgroups.go
@@ -21,13 +21,31 @@
 package cgroups
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"io"
 	"path/filepath"
 	"strings"
 
+	"github.com/eminwux/kukeon/cmd/kuke/shared"
 	"github.com/eminwux/kukeon/internal/cgroupcheck"
+	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 	"github.com/spf13/cobra"
+)
+
+// MockControllerKey is used to inject a mock kukeonv1.Client via context in
+// tests of the `--scope realm|space|stack|cell` paths. Mirrors the same
+// pattern used by `cmd/kuke/get/...` subcommands so test setup is uniform.
+type MockControllerKey struct{}
+
+// scope kinds accepted by the --scope flag.
+const (
+	scopeRealm = "realm"
+	scopeSpace = "space"
+	scopeStack = "stack"
+	scopeCell  = "cell"
 )
 
 // NewCgroupsCmd builds the `kuke doctor cgroups` subcommand.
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #332 Review — `feat: kuke doctor cgroups --scope realm|space|stack|cell`**

Clean addition: scope resolution, parent-flag validation, and the four `Get*` daemon paths are all wired correctly, and the test table covers every kind plus the misuse paths. Mirrors the `MockControllerKey{}` injection pattern from `cmd/kuke/get/...` so test setup stays uniform. No public-API parity surface (CLI-only entry; nothing in `pkg/api` exposes the doctor programmatically). No sensitive-data leaks; no dead symbols. `make dev-init` is in the test plan as the project's CLAUDE.md requires.

A couple of micro-nits, all non-blocking:

- **Wording** `cmd/kuke/doctor/cgroups/cgroups.go:362`: `cell %q in stack %q/%q/%q` and the parallel `cell %q not found in stack %q/%q/%q` (≈line 333) break the comma-form the realm/space/stack branches use (`stack "x" in realm "r", space "sp"`). Reading `in stack "r"/"sp"/"st"` parses oddly — `r` and `sp` aren't stacks. Easy fold to `in realm %q, space %q, stack %q`.

LGTM, ready to merge.

PR #337, opened 2026-05-07
fix: doctor cgroups probes by default to avoid suggesting EOPNOTSUPP fix

### LINKED ISSUE #333: cgroups: doctor without --probe suggests a fix that returns EOPNOTSUPP on namespace-trapped hosts
## Repro

On a host whose cgroup namespace advertises `memory`/`io` in `cgroup.controllers` but whose parent never delegated them (the cgroup-namespace trap — same environment class `dev-init.sh --probe` is designed to catch):

```bash
sudo ./kuke doctor cgroups          # no --probe
# vs
sudo ./kuke doctor cgroups --probe
```

Then literally copy-paste the suggested remediation from the no-probe output:

```bash
echo "+memory +io" | sudo tee /sys/fs/cgroup/cgroup.subtree_control
```

## Observed

No-`--probe` output classifies the trapped controllers as `needs-delegation` and prints a single fix line, with no warning that the fix may fail:

```
  cpu: enabled
  memory: needs-delegation
  io: needs-delegation
  pids: enabled

  not enabled in cgroup.subtree_control: memory, io
    these controllers are advertised but not yet delegated to
    children. enable them with the fix below.

  fix: echo "+memory +io" | sudo tee /sys/fs/cgroup/cgroup.subtree_control
```

The suggested fix then returns:

```
tee: /sys/fs/cgroup/cgroup.subtree_control: Operation not supported
```

…because the kernel rejects the write — the parent never delegated the controllers down. With `--probe`, the doctor performs that same write itself, gets `EOPNOTSUPP`, demotes the status to `not-delegated`, and adds the explanatory stanza:

```
  parent did not delegate (cgroup-namespace trap): memory, io
    the namespace this shell runs in advertises these controllers
    in cgroup.controllers but the actual parent never delegated
    them. escalate to the host / container runtime above this
    shell to delegate the missing controllers.
```

So the diagnosis the user gets depends entirely on whether they happened to pass `--probe`. `dev-init.sh` does; a user running `kuke doctor cgroups` standalone (or following docs that don't mention `--probe`) does not, and is told to run a command that demonstrably won't work in their environment.

## Root cause

`internal/cgroupcheck/cgroupcheck.go` — the no-probe branch terminates classification at `StatusNeedsDelegation` (lines 73–77) without attempting the write that would distinguish "delegatable but not yet enabled" from "namespace trap, write will EOPNOTSUPP". The trap is only detectable via the probe, which is gated behind the `--probe` flag.

The remediation text rendered for `StatusNeedsDelegation` (around `cgroupcheck.go:278` and the doctor's output formatter in `cmd/kuke/doctor/cgroups/cgroups.go`) prints the `tee` line as the answer, with no qualifier that it may EOPNOTSUPP.

## Suggested fix

Two reasonable options, not mutually exclusive:

1. **Probe by default.** Make `--probe` the default behavior of `kuke doctor cgroups` and add a `--no-probe` opt-out for the rare case where the side effect (idempotent on healthy hosts, harmless on trapped ones) isn't wanted. This eliminates the diagnostic gap entirely — every user gets the accurate classification.
2. **Qualify the no-probe remediation.** When emitting the fix for `StatusNeedsDelegation`, include the same "if that returns 'Operation not supported' …" qualifier the trap-stanza already carries, so a user who hits EOPNOTSUPP at least understands what happened without re-running with `--probe`.

Option 1 is the stronger fix; option 2 is the minimal one. The `dev-init.sh --probe` callsite stays valid under either.

eminwux:
## Summary

- `kuke doctor cgroups` previously deferred the `+<ctrl>` probe write to `--probe`, so a bare invocation on a cgroup-namespace-trapped host (controllers advertised in `cgroup.controllers` but never delegated by the parent) classified the controllers as `needs-delegation` and suggested a `tee` fix that returns `Operation not supported` when the user actually pasted it.
- Flip the default to **probe-on**, idempotent on healthy hosts and harmless on trapped ones. Add a `--no-probe` opt-out for the rare strictly-read-only case. The trap stanza already in `FormatRemediation` is now reached for every user, not just those who happened to pass `--probe`.
- Backwards-compatible with the existing `dev-init.sh --probe` callsite — the flag stays accepted; default just changed from `false` to `true`. `--no-probe` always wins over `--probe` so an explicit opt-out is unambiguous regardless of flag ordering.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full unit-test sweep CI runs
- [x] `go test ./cmd/kuke/doctor/cgroups/... ./internal/cgroupcheck/...` — affected packages
- [x] New tests: `TestCgroupsCmdProbesByDefault` (pins probe-on default), `TestCgroupsCmdNoProbeWinsOverProbe` (pins opt-out semantics). Existing `TestCgroupsCmdMissingFails` renamed to `TestCgroupsCmdMissingFailsNoProbe` and pinned to `--no-probe`; `TestCgroupsCmdNestedFailsOnPartialSubtree` and `TestCgroupsCmdScopeFailsOnGap` updated likewise so they keep asserting the read-only failure path.
- [x] Live-host smoke: `sudo ./kuke doctor cgroups`, `--no-probe`, `--probe`, and `--verbose-status` — all exit 0 on a fully-delegated host. Help text shows `--probe (default true)` and the new `--no-probe` flag with the documented opt-out behavior.
- [ ] `make dev-init` — not run; the change is purely client-side CLI default-flip in `kuke doctor cgroups` and does not touch the daemon, the build path, or `/opt/kukeon`. The `dev-init.sh --probe` callsite is contractually covered by `TestCgroupsCmdProbeRecoversOnPlainFile`.

Closes #333

### DIFF
--- a/cmd/kuke/doctor/cgroups/cgroups.go
+++ b/cmd/kuke/doctor/cgroups/cgroups.go
@@ -54,6 +54,7 @@ func NewCgroupsCmd() *cobra.Command {
 		root    string
 		nested  bool
 		probe   bool
+		noProbe bool
 		verbose bool
 
 		scope string
@@ -74,9 +75,15 @@ func NewCgroupsCmd() *cobra.Command {
 			"against that directory instead of the host root — useful for\n" +
 			"diagnosing mid-tree delegation gaps when a workload below a given\n" +
 			"level fails to start.\n\n" +
+			"By default, the pre-flight probes any controller missing from\n" +
+			"cgroup.subtree_control with a +<ctrl> write so the cgroup-namespace\n" +
+			"trap (advertised but not delegated, write returns EOPNOTSUPP) is\n" +
+			"distinguished from \"merely needs the operator to enable it\". The\n" +
+			"probe is idempotent on healthy hosts and harmless on trapped ones;\n" +
+			"pass --no-probe to keep the pre-flight strictly read-only.\n\n" +
 			"Exit code 0: every required controller is enabled (or was enabled by\n" +
-			"the optional --probe write). Non-zero: at least one controller is\n" +
-			"missing or the cgroup directory could not be read.",
+			"the probe write). Non-zero: at least one controller is missing or the\n" +
+			"cgroup directory could not be read.",
 		Args: cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			target := scopedTarget{
@@ -88,7 +95,10 @@ func NewCgroupsCmd() *cobra.Command {
 			if len(args) > 0 {
 				target.name = strings.TrimSpace(args[0])
 			}
-			return runDoctor(cmd, root, nested, probe, verbose, target)
+			// --no-probe always wins over --probe so an explicit opt-out
+			// is unambiguous regardless of flag ordering.
+			effectiveProbe := probe && !noProbe
+			return runDoctor(cmd, root, nested, effectiveProbe, verbose, target)
 		},
 		// A failed pre-flight is a host-environment problem, not a CLI
 		// usage error — the structured remediation is the message we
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #337 Review — `fix: doctor cgroups probes by default to avoid suggesting EOPNOTSUPP fix`**

Tight, well-scoped default-flip with the right tests. The `effectiveProbe := probe && !noProbe` reduction makes the `--no-probe` precedence rule both readable and orderless, and the four-test matrix (default-probes / explicit `--probe` / `--no-probe` / `--probe --no-probe`) pins all the semantic corners — including the renamed `TestCgroupsCmdMissingFailsNoProbe` whose comment correctly explains *why* `--no-probe` is now required to keep the read-only-failure assertion stable. `dev-init.sh --probe` is contractually preserved by `TestCgroupsCmdProbeRecoversOnPlainFile`. `make dev-init` skip is justified — purely client-side CLI, doesn't touch daemon / build path / `/opt/kukeon`. No correctness, leak, dead-code, or convention issues.

LGTM.

PR #338, opened 2026-05-07
fix: classify domain-only EOPNOTSUPP in threaded subtree as threaded-subtree

### LINKED ISSUE #334: cgroupcheck: --probe misdiagnoses threaded-subtree EOPNOTSUPP as cgroup-namespace trap
## Repro

On a host where the cell hosting `kuke doctor cgroups --probe` has `cgroup.type` of `domain threaded` (the cgroup is the root of a threaded subtree — e.g. some descendant runtime created a `cgroup.type=threaded` child, which auto-promotes the ancestor):

```bash
$ cat /sys/fs/cgroup/cgroup.type
domain threaded

$ cat /sys/fs/cgroup/cgroup.controllers
cpuset cpu io memory hugetlb pids rdma misc

cpu pids

$ echo "+memory" | sudo tee /sys/fs/cgroup/cgroup.subtree_control
tee: /sys/fs/cgroup/cgroup.subtree_control: Operation not supported

$ sudo ./kuke doctor cgroups --probe
```

## Observed

Doctor's `--probe` mode classifies the failed write as `StatusNotDelegated` (the cgroup-namespace trap) and prints the trap stanza:

```
  cpu: enabled
  memory: not-delegated
  io: not-delegated
  pids: enabled

  parent did not delegate (cgroup-namespace trap): memory, io
    the namespace this shell runs in advertises these controllers
    in cgroup.controllers but the actual parent never delegated
    them. escalate to the host / container runtime above this
    shell to delegate the missing controllers.
```

This is the wrong cause. The parent **did** delegate — `cgroup.controllers` listing `cpu io memory pids ...` is positive evidence of delegation by the parent's `subtree_control`. The actual reason `+memory +io` returns `EOPNOTSUPP` is a different kernel rule: **a `domain threaded` cgroup may only enable thread-aware controllers (cpu, pids, perf_event) in its subtree_control. `memory` and `io` are domain-only and forbidden.**

The diagnostic's suggested remediation ("escalate to the host / container runtime above this shell to delegate the missing controllers") will not help the operator — escalating doesn't fix a threaded-subtree state; the fix is at the threaded-descendant or the cell's own `cgroup.type`, neither of which the operator-side escalation path addresses.

## Root cause

`internal/cgroupcheck/cgroupcheck.go` — the probe path turns `EOPNOTSUPP` into `StatusNotDelegated` unconditionally (around lines 68-72 / Status definitions and the probe-write classifier). It does not read `cgroup.type` of the target before classifying, so it cannot distinguish:

- **namespace trap**: `cgroup.controllers` is missing the controller AND/OR the parent's `subtree_control` lacks it → write returns `EOPNOTSUPP`.
- **threaded-subtree trap**: `cgroup.type` is `domain threaded` (or `threaded`), the controller is domain-only → write returns `EOPNOTSUPP`.

Both yield the same kernel errno but require fundamentally different remediation.

## Suggested fix

Add a new status + remediation message and gate the namespace-trap conclusion on a `cgroup.type` read:

1. In `internal/cgroupcheck/cgroupcheck.go`, add `StatusThreadedSubtree` (between `StatusNotDelegated` and `StatusNeedsDelegation`) with `String()` label `threaded-subtree`.
2. Before classifying an `EOPNOTSUPP` probe write as `StatusNotDelegated`, read `<target>/cgroup.type`. If the value is `domain threaded` or `threaded` and the failing controller is domain-only (memory, io, etc.), classify as `StatusThreadedSubtree` instead.
3. In the doctor's output formatter (`cmd/kuke/doctor/cgroups/cgroups.go`), emit a distinct stanza for `StatusThreadedSubtree` explaining: cgroup is in a threaded subtree, domain-only controllers cannot be enabled here, fix lies in the threaded descendant chain or the cell's `cgroup.type`, not in escalating to the parent runtime.

### Surfaced alongside

- #333 — separately, the no-`--probe` path advertises a `tee` fix that EOPNOTSUPPs without warning. Both bugs share the same family of EOPNOTSUPP-handling gaps but the fixes are independent: #333 touches the no-probe remediation message; this issue adds a new probe-side classification.
- Likely related kukeond gap (not filed here): `EnsureSubtreeControllers` in `internal/ctr/cgroups.go:148-174` filters the desired set against the host root's `cgroup.controllers` but does not consult the *target* cgroup's `cgroup.type`. On a `domain threaded` target, the resulting `+cpu +memory +io +pids` write will EOPNOTSUPP partway through, leaving subtree_control half-populated. Cell-creation behavior on a threaded target is unverified from this investigation; worth a follow-up bug if reproduced.

eminwux:
## Summary
- Doctor's `--probe` mode misdiagnosed EOPNOTSUPP on a `domain threaded` cgroup as the cgroup-namespace trap, sending operators to escalate to the parent runtime when the actual rule is that domain-only controllers (memory, io, ...) are forbidden in a threaded subtree's `subtree_control`.
- Adds `StatusThreadedSubtree` between `StatusNotDelegated` and `StatusNeedsDelegation`. The probe-error classifier now reads `<target>/cgroup.type` on EOPNOTSUPP and switches to the new status when the cgroup is `domain threaded` or `threaded` and the failing controller is not in the kernel's thread-aware set (`cpu, cpuset, perf_event, pids`).
- New `FormatRemediation` stanza explains the threaded-subtree case and points at the threaded descendant / cell `cgroup.type` as the fix instead of the parent runtime. Threaded-subtree controllers are also dropped from the generic `+<ctrl> | sudo tee` fix line — that write would EOPNOTSUPP again.
- The `cgroup.type` read fails soft: missing file → `false`, so the existing namespace-trap diagnosis still fires when no threaded state is present (covered by the existing `TestCheckProbeEOPNOTSUPP` and a new `TestCheckProbeMissingCgroupTypeFallsBackToNotDelegated`). No daemon code consumes the doctor classifier (only `CellResourceControllers()` is shared), so this is a doctor-side change with no behavioral impact on cell creation.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` — all packages pass, including the five new cgroupcheck tests covering `domain threaded` + domain-only → ThreadedSubtree, `threaded` alone, thread-aware controller in threaded subtree → still NotDelegated, missing cgroup.type → fallback, and the `String()` label
- [x] `pre-commit run` on the changed files (golangci-lint, gofmt, addlicense) — passes
- [x] `make dev-init` smoke per project CLAUDE.md (since `dev-init.sh` calls `kuke doctor cgroups --probe` whose code I changed) — daemon-parity tail matches expected output for `default` and `kuke-system` realms

Closes #334

### DIFF
--- a/internal/cgroupcheck/cgroupcheck.go
+++ b/internal/cgroupcheck/cgroupcheck.go
@@ -70,6 +70,14 @@ const (
 	// cgroup-namespace trap: cgroup.controllers advertises kernel support
 	// but the namespace's actual parent never delegated the controller.
 	StatusNotDelegated
+	// StatusThreadedSubtree — the probe write returned EOPNOTSUPP and
+	// cgroup.type at this scope is "domain threaded" or "threaded", so the
+	// kernel forbids enabling domain-only controllers (memory, io, ...) in
+	// this subtree_control regardless of delegation. Distinct from
+	// StatusNotDelegated because escalating to the parent runtime will not
+	// help — the fix lives at the threaded descendant or in this cgroup's
+	// own cgroup.type.
+	StatusThreadedSubtree
 	// StatusNeedsDelegation — listed in cgroup.controllers, missing from
 	// cgroup.subtree_control, and we did not probe (or probing wasn't
 	// permitted). The fix is "+<ctrl>" to cgroup.subtree_control, but the
@@ -93,6 +101,8 @@ func (s Status) String() string {
 		return "kernel-missing"
 	case StatusNotDelegated:
 		return "not-delegated"
+	case StatusThreadedSubtree:
+		return "threaded-subtree"
 	case StatusNeedsDelegation:
 		return "needs-delegation"
 	case StatusEnabledByProbe:
@@ -134,6 +144,7 @@ func (r Result) Unresolved() []string {
 			// Good terminal state.
 		case StatusKernelMissing,
 			StatusNotDelegated,
+			StatusThreadedSubtree,
 			StatusNeedsDelegation,
 			StatusPermissionDenied:
 			out = append(out, c)
@@ -233,22 +244,28 @@ func Check(hostRoot string, required []string, probe Prober) (Result, error) {
 			}
 			perr := probe(hostRoot, c)
 			res.ProbeErr[c] = perr
-			res.Status[c] = classifyProbeErr(perr)
+			res.Status[c] = classifyProbeErr(hostRoot, c, perr)
 		}
 	}
 	return res, nil
 }
 
 // classifyProbeErr maps the result of a probe write to a Status. The
-// distinction between StatusNotDelegated and StatusPermissionDenied is the
-// whole point of probing — getting it right keeps the remediation
-// suggestion correct.
-func classifyProbeErr(err error) Status {
+// distinction between StatusNotDelegated, StatusThreadedSubtree, and
+// StatusPermissionDenied is the whole point of probing — getting it right
+// keeps the remediation suggestion correct. EOPNOTSUPP on a domain-only
+// controller in a "domain threaded" / "threaded" cgroup is the
+// threaded-subtree trap, not the cgroup-namespace trap; reading
+// cgroup.type at hostRoot is what disambiguates them.
+func classifyProbeErr(hostRoot, controller string, err error) Status {
 	if err == nil {
 		return StatusEnabledByProbe
 	}
 	switch {
 	case errors.Is(err, syscall.EOPNOTSUPP), errors.Is(err, syscall.ENOTSUP):
+		if isThreadedSubtreeRoot(hostRoot) && isDomainOnlyController(controller) {
+			return StatusThreadedSubtree
+		}
 		return StatusNotDelegated
 	case errors.Is(err, syscall.EACCES), errors.Is(err, syscall.EPERM):
 		return StatusPermissionDenied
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #338 Review — `fix: classify domain-only EOPNOTSUPP in threaded subtree as threaded-subtree`**

Clean fix for #334. Classifier now reads `cgroup.type` at the same scope as the failing `cgroup.subtree_control` write and only re-routes to the new `StatusThreadedSubtree` when the type is `domain threaded`/`threaded` *and* the failing controller is outside the kernel's thread-aware set (`cpu, cpuset, perf_event, pids` per `cgroup-v2.rst`). Read fails soft → existing namespace-trap diagnosis still fires. New status is wired through `Unresolved()`, `String()`, and `FormatRemediation` (including the `+<ctrl> | sudo tee` fix-line filter), and the doctor consumer (`cmd/kuke/doctor/cgroups/cgroups.go`) picks it up automatically via `Result.OK()`. No daemon-side fallout.

Tests are tight: five new cases pin `domain threaded`+domain-only → ThreadedSubtree, `threaded` alone, thread-aware controller in threaded subtree → still NotDelegated (the boundary case), missing `cgroup.type` → fallback to NotDelegated, and the `threaded-subtree` String() label. The fix-line and namespace-trap-stanza assertions in `TestCheckProbeThreadedSubtreeDomainOnly` are good defensive guards against regressions.

No correctness, leak, dead-code, or convention issues. License headers / errdefs / colocation conventions all hold. Test plan tail (`make dev-init` smoke) is the project-CLAUDE.md-required check and is marked done — that's the regression guard for `kuke doctor cgroups --probe` on this code path.

LGTM.

PR #339, opened 2026-05-07
fix: classify EBUSY probe write as internal-process, not namespace trap

### LINKED ISSUE #335: cgroupcheck: --probe misclassifies EBUSY (no-internal-process) as namespace trap
## Repro

From inside any cell whose `_work` (or any cgroup) has internal processes — straightforward to reproduce in a `kukeon-dev-*` dev cell:

```
$ sudo kuke doctor cgroups --probe --root /sys/fs/cgroup
```

(or, equivalently, directly trigger the same syscall the prober makes:)

```
$ echo '+memory' > /sys/fs/cgroup/cgroup.subtree_control
bash: echo: write error: Device or resource busy
```

Observed in cell `default_default_kukeon-dev-7cc88b_work`, kernel 6.12.38+deb13-cloud-amd64, runc 1.3.3.

## Observed

`bash: echo: write error: Device or resource busy` — the kernel returns `EBUSY` because `_work` has processes in its own `cgroup.procs` (the cell's payload: bash, sbsh, etc.) and the cgroup-v2 *no-internal-process* rule forbids enabling non-thread-aware controllers on a non-leaf cgroup that contains processes.

`kuke doctor cgroups --probe` then routes that `EBUSY` through `classifyProbeErr` and prints the namespace-trap remediation:

> parent did not delegate (cgroup-namespace trap): memory, io, hugetlb, …
> the namespace this shell runs in advertises these controllers in cgroup.controllers but the actual parent never delegated them. escalate to the host / container runtime above this shell to delegate the missing controllers.

That's the wrong diagnosis. The parent *did* delegate (the cell carries the full 8 controllers in its `subtree_control`, and `_work`'s `cgroup.controllers` shows all 8). The actual cause is the no-internal-process rule, and the fix is to either move the processes to a child cgroup or accept that only thread-aware controllers (`cpu`, `pids`) can land here.

## Root cause

`internal/cgroupcheck/cgroupcheck.go:246-260` (`classifyProbeErr`) only branches on `EOPNOTSUPP`/`ENOTSUP` and `EACCES`/`EPERM`; everything else (including `EBUSY`) falls through to `StatusNotDelegated` via the comment "Fall back to the cgroup-ns trap class when the kernel surfaced an unexpected error". For `EBUSY` that fallback is wrong — `EBUSY` is a structurally different precondition from "namespace trap" and has its own remediation.

This is the third distinct case in the same surface as #333 (no-probe `tee` suggestion) and #334 (threaded-subtree EOPNOTSUPP). #334 already proposes reading `cgroup.type` to disambiguate the two `EOPNOTSUPP` flavors; this issue asks to add `EBUSY` as its own class on top of that.

## Suggested fix

In `internal/cgroupcheck/cgroupcheck.go`:

1. Add a `StatusInternalProcess` constant (alongside `StatusNotDelegated`, `StatusKernelMissing`, etc.) at the `Status` block (~l. 60-85).
2. In `classifyProbeErr`, branch `errors.Is(err, syscall.EBUSY)` to the new status before the fallback.
3. In `FormatRemediation` (~l. 265-319), add a group for `StatusInternalProcess` with text along the lines of: "the cgroup contains processes; the cgroup-v2 no-internal-process rule blocks enabling non-thread-aware controllers in `subtree_control` on a non-leaf cgroup. Move the processes to a child cgroup, or accept that only thread-aware controllers (cpu, pids) can be enabled here."

Optionally augment the prober result with the offending cgroup's `cgroup.procs` count so the operator can see the precondition at a glance.

Tests: extend `cgroupcheck_test.go`'s `TestCheckProbeEOPNOTSUPP`-shaped table to add an `EBUSY` row asserting `StatusInternalProcess` and the corresponding remediation snippet.

eminwux:
## Summary

- `kuke doctor cgroups --probe` returned `EBUSY` from the kernel when a non-leaf cgroup hosting processes (e.g. a cell `_work`) is asked to enable a non-thread-aware controller — the cgroup-v2 no-internal-process rule. The classifier dropped through to the `StatusNotDelegated` fallback, so operators saw the cgroup-namespace-trap stanza ("escalate to the parent runtime") which does not address the actual precondition.
- Add `StatusInternalProcess` and branch `EBUSY` in `classifyProbeErr` to it. New `FormatRemediation` stanza names the no-internal-process rule and points the operator at moving processes to a child cgroup (or accepting thread-aware-only enablement at this scope).
- Internal-process controllers are excluded from the `+<ctrl> | sudo tee` fix line for the same reason `StatusKernelMissing` and `StatusThreadedSubtree` already are: the suggested write would EBUSY again until the precondition is fixed.
- Optional augmentation in the issue (surface `cgroup.procs` count alongside the result) is intentionally skipped as the lower-blast-radius default — the new stanza names the precondition explicitly, and a follow-up can wire the count if it proves useful.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/cgroupcheck/... ./cmd/kuke/doctor/cgroups/...`
- [x] `make test` (full suite)
- [x] `pre-commit run` on staged files
- [ ] `make dev-init` smoke — change is contained to in-memory classifier logic for `kuke doctor cgroups`; daemon path / cell bootstrap untouched, so the dev-init smoke is not load-bearing for this fix.

Closes #335

### DIFF
--- a/internal/cgroupcheck/cgroupcheck.go
+++ b/internal/cgroupcheck/cgroupcheck.go
@@ -78,6 +78,14 @@ const (
 	// help — the fix lives at the threaded descendant or in this cgroup's
 	// own cgroup.type.
 	StatusThreadedSubtree
+	// StatusInternalProcess — the probe write returned EBUSY because this
+	// cgroup contains processes in its own cgroup.procs. Cgroup-v2's
+	// no-internal-process rule forbids enabling non-thread-aware
+	// controllers in subtree_control on a non-leaf cgroup that hosts
+	// processes. Distinct from StatusNotDelegated because the parent did
+	// delegate; the fix is to move the processes to a child cgroup (or
+	// accept that only thread-aware controllers can land here).
+	StatusInternalProcess
 	// StatusNeedsDelegation — listed in cgroup.controllers, missing from
 	// cgroup.subtree_control, and we did not probe (or probing wasn't
 	// permitted). The fix is "+<ctrl>" to cgroup.subtree_control, but the
@@ -103,6 +111,8 @@ func (s Status) String() string {
 		return "not-delegated"
 	case StatusThreadedSubtree:
 		return "threaded-subtree"
+	case StatusInternalProcess:
+		return "internal-process"
 	case StatusNeedsDelegation:
 		return "needs-delegation"
 	case StatusEnabledByProbe:
@@ -145,6 +155,7 @@ func (r Result) Unresolved() []string {
 		case StatusKernelMissing,
 			StatusNotDelegated,
 			StatusThreadedSubtree,
+			StatusInternalProcess,
 			StatusNeedsDelegation,
 			StatusPermissionDenied:
 			out = append(out, c)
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #339 Review — `fix: classify EBUSY probe write as internal-process, not namespace trap`**

Tight, well-scoped fix. The new `StatusInternalProcess` branch correctly maps cgroup-v2's no-internal-process rule (the canonical EBUSY source on `subtree_control` writes), and the FormatRemediation stanza names the precondition explicitly instead of misdirecting to the parent runtime. Test exercises the full path — classifier verdict, fix-line exclusion, and absence of the namespace-trap stanza — and pins the `internal-process` label string for downstream greps. No sensitive data, no dead code, no public-API parity gaps. The `make dev-init` skip is justified: classifier logic is in-process to `kuke doctor cgroups`; daemon and cell bootstrap don't import the changed code paths.

LGTM.

PR #340, opened 2026-05-07
fix: drain cell PIDs to _payload leaf for subtree_control widening

### LINKED ISSUE #336: runner: cell-rooted PIDs block subtree_control widening (no-internal-process rule)
Refiles #311 (closed `NOT_PLANNED`, no comment rationale) with new runtime evidence and a unified fix path. The two filings describe the same underlying bug — cell PID 1 (and any process spawned directly into a cell-namespace-root cgroup) violating cgroup-v2's no-internal-process rule, which then forbids widening that cgroup's `subtree_control` for non-thread-aware controllers. The diagnostic side is filed as #335 (cgroupcheck misclassifying EBUSY as namespace trap) and shipped in #339; this issue is the runtime fix.

## Repro

Two scenarios trigger the same kernel rejection.

### A — any kuke-managed cell (from #311)

On a cgroup-v2 host, inside any cell (verified on a `default.kukeon.io` cell at HEAD `3210c3e`):

```
$ cat /proc/self/cgroup
0::/

$ cat /sys/fs/cgroup/cgroup.controllers
cpuset cpu io memory hugetlb pids rdma misc

cpuset cpu pids

$ cat /sys/fs/cgroup/cgroup.events
populated 1
frozen 0

$ echo +memory > /sys/fs/cgroup/cgroup.subtree_control
bash: echo: write error: Operation not supported
```

`/proc/self/cgroup` reports `0::/` — we are at the cgroup-namespace root. `cgroup.controllers` lists every controller the kernel exposes; `subtree_control` is missing `memory`, `io`, `hugetlb`, `rdma`, `misc`. Anything inside the cell that needs to create child cgroups with `memory` / `io` / `hugetlb` constraints — nested `kukeond`, an inner `dockerd`, an inner CRI, or kuke's own resource-control flow if it ever runs nested — silently can't enable those controllers in `subtree_control`. Resource limits on the cell itself work; nesting beneath the cell does not.

### B — nested `kuke init` (NestedCgroupRuntime: true)

Inside a kuke-managed dev cell that has `NestedCgroupRuntime: true` and lands the user shell in `_work`, the same pattern blocks `kuke init` itself:

```
$ echo '+memory' > /sys/fs/cgroup/cgroup.subtree_control
bash: echo: write error: Device or resource busy

$ sudo ./kuke init   # full daemon-bootstrap path — same EBUSY
```

`populated 1` because the bash shell, the inner `kuke` invocation, and (in the dev profile) `claude` all live in the cgroup-ns root from the cell's view.

## Root cause

cgroup-v2's no-internal-process rule: a cgroup with live processes cannot enable any non-thread-aware controller in its `subtree_control`.

**Scenario A's source of population:** the cell-cgroup setup path places PID 1 of the cell directly in `/kukeon/<realm>/<space>/<stack>/<cell>/`. `internal/ctr/cgroups.go` `NewCgroup` (and `AddProcessToCgroup` after `runc create`) does not move the task out of the cell-root cgroup into a leaf scope. As long as PID 1 sits at the cgroup root, the kernel will not let `subtree_control` gain any controllers that would constrain its children — the standard cgroup-v2 delegation rule.

**Scenario B's source of population:** the same pathology shows up at the use site. `internal/ctr/cgroups.go:225` calls `manager.ToggleControllers(enable, cgroup2.Enable)` (containerd cgroup2 lib) which walks ancestors writing `+memory +io ...` to each `cgroup.subtree_control`. The walk's first ancestor from inside the nested cell is `/sys/fs/cgroup` = the cell's own `_work` cgroup on the host = `populated 1` (shell + kuke + claude live there directly, spawned by `runc exec`, not by cell PID 1). Bulk write fails with EBUSY. `applySubtreeControllers` does not visibly retry — the error propagates to `enableCellControllers` and `kuke init` aborts.

A symptom seen in a sibling cell `kukeon-dev-a647a2` earlier in the debug session: an inner `_work` cgroup ended up with `cpu pids` in `subtree_control` despite the outer cell having all 8. The fingerprint matches `runc-1.3.3` `vendor/github.com/opencontainers/cgroups/fs2/create.go:138-144` exactly:

```go
if err := cgroups.WriteFile(current, cgStCtlFile, res); err != nil {
    allCtrs := strings.Split(res, " ")
    for _, ctr := range allCtrs {
        _ = cgroups.WriteFile(current, cgStCtlFile, ctr)
    }
}
```

When the bulk write EBUSYs (no-internal-process rule), runc retries one-by-one and silently swallows failures. Only the thread-aware controllers (`cpu`, `pids`) succeed; `memory io cpuset hugetlb` are dropped on the floor. The user sees a partially-populated `subtree_control` and no error.

## Fix

Enforce the cgroup-v2 rule at every kukeon code path that writes to `subtree_control`:

1. **Cell creation (resolves scenario A).** After `runc create` registers the task PID via `AddProcessToCgroup`, `mkdir <cell>/_payload` (name TBD: systemd convention is `init.scope`; #312 already discussed `_payload`) and write the task PID into `<cell>/_payload/cgroup.procs` before calling `EnableCellSubtreeControllers` / `EnableCellAllSubtreeControllers`. The cell root then has no internal processes and the `subtree_control` walks succeed for every controller. Site: `internal/ctr/cgroups.go` `NewCgroup` + `AddProcessToCgroup` (and the cell-bringup caller in `internal/controller/runner/provision.go`).

2. **Defensive use-site (resolves scenario B).** In `applySubtreeControllers` (`internal/ctr/cgroups.go:225`), before calling `manager.ToggleControllers`, gate on `(/proc/self/cgroup == 0::/) AND (mountpoint root cgroup.events shows populated 1)`. The first conjunct guards against accidentally draining systemd's processes on a non-nested host (where `/proc/self/cgroup` shows the systemd path, not `0::/`). If both hold, drain `cgroup.procs` from the mountpoint root to a leaf scope (e.g. `__bootstrap`) before the walk.

Both fixes share a primitive: create a leaf cgroup under a target path and drain the target's `cgroup.procs` into it. Implement once, call from both sites.

### Tests

Unit test in `internal/ctr/cgroups_test.go` (or sibling) — temp cgroup-v2 mountpoint, create a cgroup with a process, verify `+memory` initially returns ENOTSUPP, then verify that after relocating the process to the leaf scope `+memory` succeeds. Gated on a writable cgroup-v2 hierarchy in CI per the existing TODO at `internal/ctr/cgroups.go:32`.

The standard `make dev-init` smoke runs on a non-nested host, so it exercises scenario A's structural fix path but not scenario B's defensive path. A nested-cell smoke target (`make dev-init` inside a `kukeon-dev-*` cell, gated on a writable cgroup-v2 hierarchy) would catch scenario B regressions — file as a follow-up if useful.

## Deferred to follow-ups

- **Apply the leaf-scope pattern preventively to every kukeon level (realm/space/stack)** — the kukeon ancestor levels are not currently `populated`, but the rule is fragile; making it structural at every level prevents future regressions. File when/if the bug surfaces at a non-cell level.
- **Make `applySubtreeControllers` fail loud on EBUSY/EOPNOTSUPP after the leaf-scope move**, with a new `errdefs` sentinel so callers can distinguish "kernel rejected" from "library error". A behavior change to existing call sites — file as its own issue.
- **Re-classify EBUSY in `cgroupcheck.classifyProbeErr`** as a precondition-named status — already shipped in #339 (filed as #335).
- **Confirm or rule out the runc threaded-flip path** as a contributing factor. The `cpu pids only` outcome can be explained without any threaded promotion (pure no-internal-process EBUSY + per-controller fallback), so it may be a red herring. `runc-1.3.3 vendor/github.com/opencontainers/cgroups/fs2/create.go:124` writes `"threaded"` to `cgroup.type` when current cgroup is `domain invalid` and resources don't reference a domain controller — which fires for kukeon cells per `internal/ctr/spec.go:299,312`. Worth confirming with an `inotifywait` watcher; not blocking the fix above.

Supersedes #311.

eminwux:
## Summary
- Add `RelocateProcessesToLeaf(group, mountpoint, leaf)` primitive in `internal/ctr/cgroups.go` that mkdir's `<group>/<leaf>` and drains `<group>/cgroup.procs` into it. Idempotent.
- **Scenario A (cell creation, structural):** in `StartContainer`, after `task.Start` returns, look up the OCI spec's `Linux.CgroupsPath` and relocate the container's PIDs into a `_payload` leaf. The cgroup-namespace pin is sticky to the original placement, so moving the PID after task start does not reshape the inside-cell view of `/sys/fs/cgroup` — it just empties the cgroup-ns root `cgroup.procs`, which is what cgroup-v2's no-internal-process rule requires before any inner runtime (`kuke init`, `dockerd`, `podman`, an inner containerd) can widen `subtree_control` for non-thread-aware controllers like memory or io. Relocate failures log a warning and continue, so a kernel that rejects the move never blocks container start.
- **Scenario B (defensive use-site):** in `applySubtreeControllers`, before `manager.ToggleControllers`, gate on `(/proc/self/cgroup == 0::/) AND (mountpoint root cgroup.events shows populated 1)`. Both clauses required: the `0::/` check guards against draining systemd's processes on a non-nested host (where the path shows the systemd slice), and the populated check avoids the no-op mkdir+read on an empty root. If both hold, drain mountpoint root into `__bootstrap` before the toggle.
- Distinct leaf names (`_payload` for cell startup, `__bootstrap` for the defensive drain) keep post-mortem inspection of `/sys/fs/cgroup` able to tell the two layers apart.
- New `errdefs.ErrInvalidLeafName` sentinel for relocate-input validation.

## Non-obvious calls

- The `_payload` relocate fires for any container with a non-empty `Linux.CgroupsPath` — i.e. every kukeon-managed cell container (`HostCgroup` containers leave the field empty and short-circuit). This intentionally drains the kukeond container too: the daemon-parity smoke confirms the daemon stays healthy after the relocation, and the same primitive is what lets a future inner-runtime cell widen `subtree_control`.
- Relocate failure on the cell-startup path is logged + continued, not returned. The cgroup-v2 invariant the relocate satisfies only matters when an inner runtime later tries to widen `subtree_control`; failing the container start over a future-conditional invariant trades a real outage for a hypothetical one. The `_payload` leaf failing to mkdir is also worth surfacing without aborting so `kuke get cells` reflects the actual state.

## Test plan
- [x] `make test` — all package tests green.
- [x] `make dev-init` — full re-bootstrap loop succeeds; daemon-parity check produces the expected identical `default` / `kuke-system` realm rows from both `kuke get realms` and `kuke get realms --no-daemon`.
- [x] Live verification on the kukeond cell: both `kukeon_kukeon_kukeond_kukeond` and `kukeon_kukeon_kukeond_root` container cgroups have `_payload` leaves, the leaves carry the actual PIDs, the parent `cgroup.procs` files are empty, and `echo +memory > .../cgroup.subtree_control` succeeds (it would EBUSY pre-fix).
- [x] Unit tests in `internal/ctr/cgroups_relocate_test.go` exercise input validation, drain semantics on a tmpdir-simulated cgroupfs, idempotent re-runs, error surfacing on missing source `cgroup.procs`, and the `cgroupPopulated` / `readCgroupProcs` helpers.
- [ ] Scenario B nested-`kuke init` smoke (`kuke init` inside a `kukeon-dev-*` cell) — not run; the defensive-drain path is exercised on every kukeond `applySubtreeControllers` call once the gate matches, but the dev host is not currently set up with a nested cell. The deferred follow-up in #336 calls out a nested-cell smoke target.

## Deferred follow-ups (carried over from #336)
- Apply the leaf-scope pattern preventively to every kukeon level (realm/space/stack).
- Make `applySubtreeControllers` fail loud on EBUSY/EOPNOTSUPP after the leaf-scope move with a new `errdefs` sentinel.
- Confirm or rule out the runc-1.3.3 threaded-flip path.

Closes #336

### DIFF
--- a/internal/controller/runner/provision_subtree_test.go
+++ b/internal/controller/runner/provision_subtree_test.go
@@ -118,6 +118,10 @@ func (c *subtreeRecorderClient) EnableCellAllSubtreeControllers(string, string) 
 	panic("unexpected")
 }
 
+func (c *subtreeRecorderClient) RelocateProcessesToLeaf(string, string, string) error {
+	panic("unexpected")
+}
+
 func (c *subtreeRecorderClient) CreateContainerFromSpec(
 	string, intmodel.ContainerSpec, []ctr.RegistryCredentials, ...ctr.BuildOption,
 ) (containerd.Container, error) {
--- a/internal/ctr/cgroups.go
+++ b/internal/ctr/cgroups.go
@@ -17,6 +17,7 @@
 package ctr
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
@@ -29,6 +30,13 @@ import (
 	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/errdefs"
 )
+
+// bootstrapLeafName is the leaf cgroup the defensive drain in
+// applySubtreeControllers relocates the cgroup-namespace root's processes
+// into when widening subtree_control would otherwise hit the no-internal-
+// process rule. Kept distinct from _payload (the cell-startup leaf) so
+// post-mortem inspection of /sys/fs/cgroup can tell the two layers apart.
+const bootstrapLeafName = "__bootstrap"
 
 // TODO(eminwux): add cgroup integration tests once CI exposes a writable cgroup v2 hierarchy.
 
@@ -131,6 +139,52 @@ func (c *client) AddProcessToCgroup(group, mountpoint string, pid int) error {
 	return manager.AddProc(uint64(pid))
 }
 
+// RelocateProcessesToLeaf drains every PID currently in <group>/cgroup.procs
+// into a freshly-mkdir'd leaf cgroup at <group>/<leaf>. Used by both the
+// cell-startup structural fix and the applySubtreeControllers defensive
+// drain to satisfy cgroup-v2's no-internal-process rule before the kernel
+// is asked to enable a non-thread-aware controller in <group>'s
+// subtree_control. Idempotent: re-running on an already-drained group is a
+// no-op (mkdir tolerates existing leaf, the procs read returns no PIDs).
+//
+// The leaf scope inherits the parent's controllers via the parent's
+// subtree_control walk that runs after the drain, so resource accounting
+// applied at <group> still constrains the leaf — the relocation moves
+// where PIDs live in the cgroup tree, not which limits apply to them.
+func (c *client) RelocateProcessesToLeaf(group, mountpoint, leaf string) error {
+	if err := validateGroupPath(group); err != nil {
+		return err
+	}
+	if leaf == "" || strings.ContainsAny(leaf, "/") || leaf == "." || leaf == ".." {
+		return errdefs.ErrInvalidLeafName
+	}
+
+	mp := c.effectiveMountpoint(mountpoint)
+	groupPath := filepath.Join(mp, strings.TrimPrefix(group, "/"))
+	leafPath := filepath.Join(groupPath, leaf)
+
+	if err := os.Mkdir(leafPath, 0o750); err != nil && !os.IsExist(err) {
+		return fmt.Errorf("create leaf cgroup %s: %w", leafPath, err)
+	}
+
+	pids, err := readCgroupProcs(filepath.Join(groupPath, "cgroup.procs"))
+	if err != nil {
+		return fmt.Errorf("read cgroup.procs in %s: %w", groupPath, err)
+	}
+	if len(pids) == 0 {
+		return nil
+	}
+
+	leafProcs := filepath.Join(leafPath, "cgroup.procs")
+	if writeErr := writeCgroupProcs(leafProcs, pids); writeErr != nil {
+		return fmt.Errorf("write cgroup.procs in %s: %w", leafPath, writeErr)
+	}
+
+	c.logger.InfoContext(c.ctx, "relocated processes to leaf cgroup",
+		"group", group, "leaf", leaf, "pids", len(pids))
+	return nil
+}
+
 // EnsureSubtreeControllers writes "+<ctrl>" to every ancestor's
 // cgroup.subtree_control file (root → group's parent) AND to the group's own
 // cgroup.subtree_control. The library's Manager.ToggleControllers handles the
@@ -218,6 +272,21 @@ func (c *client) applySubtreeControllers(group, mountpoint string, enable []stri
 		return nil
 	}
 
+	// Defensive no-internal-process drain (issue #336 scenario B). When kuke
+	// is invoked from inside a cgroup-namespace whose root cgroup carries
+	// processes (the shell, the kuke binary, anything else spawned by
+	// runc-exec into the cell's container-root cgroup directly), the
+	// ancestor walk below would EBUSY at the cgroup-ns root because
+	// cgroup-v2 forbids enabling non-thread-aware controllers in the
+	// subtree_control of a cgroup that hosts processes. Both clauses are
+	// required: /proc/self/cgroup == 0::/ guards against accidentally
+	// draining systemd's processes on a non-nested host (where the path
+	// shows the systemd slice, not the bare 0::/), and the populated check
+	// avoids the no-op mkdir+read on an empty root.
+	if drainErr := c.maybeDrainCgroupNsRoot(mountpoint); drainErr != nil {
+		return fmt.Errorf("drain cgroup-ns root before subtree_control widening: %w", drainErr)
+	}
+
 	manager, err := c.managerFor(group, mountpoint)
 	if err != nil {
 		return err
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #340 Review — `fix: drain cell PIDs to _payload leaf for subtree_control widening`**

LGTM. The two-layer drain (`_payload` at cell start, `__bootstrap` defensive at `applySubtreeControllers`) is the right shape for the cgroup-v2 no-internal-process trap, and the gate clauses on the defensive path (`/proc/self/cgroup == 0::/` AND `populated 1`) correctly avoid stomping systemd on a non-nested host. `RelocateProcessesToLeaf` validation, idempotency, and read/write helper coverage in `cgroups_relocate_test.go` look solid; the `subtreeRecorderClient` test fake gets the new method as a panic-on-call. No correctness, sensitive-data, or dead-code issues; no public-API parity gap (`Client` is `internal/`).

One non-blocking nit:

- **Comment** `internal/ctr/container.go:140-142`: comment claims "HostCgroup containers leave [Linux.CgroupsPath] empty", but `cellCgroupsPath` in both `BuildContainerSpec` (`spec.go:312`) and `BuildRootContainerSpec` (`container_utils.go:152`) sets it whenever `CellCgroupPath` is non-empty, regardless of `HostCgroup` — and the PR's own live verification confirms `kukeon_kukeon_kukeond_root` (HostCgroup=true) gets a `_payload` leaf. The actual short-circuit catches "no `CellCgroupPath` set" (e.g. raw containerd / non-cell), not HostCgroup. Worth correcting so a future debugger isn't misled.

Scenario-B nested-`kuke init` smoke is intentionally deferred per the user (no nested cell on the dev host); the defensive drain is exercised on every `applySubtreeControllers` call once the gate matches, and #336's nested-cell smoke target stays on the deferred-followups list.

LGTM unconditional — merging on green.

PR #343, opened 2026-05-07
fix: doctor cgroups warns on internal-process trap when runtime self-heals

### LINKED ISSUE #342: doctor: --probe pre-flight rejects hosts that applySubtreeControllers would self-heal
## Repro

On a `NestedCgroupRuntime: true` dev cell (kukeon `_work` cgroup is the cgroup-namespace root, with shell + tools populated in it), at HEAD `246565e`:

```
$ make dev-init
...
==> Pre-flight: host cgroup controller delegation
host root: /sys/fs/cgroup
required:  cpu, memory, io, pids
  cpu: internal-process
  memory: internal-process
  io: internal-process
  pids: internal-process

host root cgroup is missing controllers required by kukeond: cpu, memory, io, pids
Error: cgroup pre-flight: missing controllers at /sys/fs/cgroup/cgroup.subtree_control
make: *** [Makefile:97: dev-init] Error 1
```

Sandbox state confirms both clauses of the `applySubtreeControllers` defensive-drain gate hold:

```
$ cat /proc/self/cgroup
0::/
$ cat /sys/fs/cgroup/cgroup.events
populated 1
frozen 0
                       # empty
```

i.e. the same condition #336 / #340 added the `__bootstrap` defensive drain for. The runtime would self-heal; the pre-flight aborts before it gets the chance.

## Observed

`scripts/dev-init.sh` aborts at the `kuke doctor cgroups --probe` step. `kuke init` (which would have invoked `applySubtreeControllers` and triggered the defensive `__bootstrap` drain merged in #340) never runs.

## Root cause

Layer mismatch between the read-only checker and the write-time self-heal:

- `cmd/kuke/doctor/cgroups/cgroups.go` `--probe` path attempts a `+<ctrl>` write at the cgroup-ns mountpoint root and classifies the EBUSY result via `internal/cgroupcheck/cgroupcheck.go:284` (`StatusInternalProcess`). It surfaces the diagnostic and exits non-zero.
- `internal/ctr/cgroups.go` `applySubtreeControllers` (PR #340) gates on `(/proc/self/cgroup == 0::/) AND (root populated)` and drains mountpoint root into `__bootstrap` via `RelocateProcessesToLeaf` *before* `manager.ToggleControllers` — so the same EBUSY would not surface at runtime.

The doctor command is read-only; it never invokes `RelocateProcessesToLeaf`. Net effect: on a `NestedCgroupRuntime` dev cell, the pre-flight is strictly stricter than the runtime path it's supposed to mirror, and `make dev-init` is now blocked even though `kuke init` itself would succeed.

## Suggested fix

Two viable shapes; pick one:

1. **Mirror the runtime drain in the doctor probe.** In `cmd/kuke/doctor/cgroups/cgroups.go`, gate on the same `(0::/ AND populated)` invariant and call `ctr.RelocateProcessesToLeaf(<mountpoint>, <mountpoint>, "__bootstrap")` before the `+<ctrl>` probe write. Stays consistent with the runtime self-heal. Caveat: `--probe` becomes mutating; document it (or split into `--probe` read-only and `--probe-with-drain` mutating).
2. **Downgrade `StatusInternalProcess` from fatal to warning** when both defensive-drain clauses hold at probe time. The doctor still prints the diagnostic so the operator sees what's about to happen, but exits 0 and lets `dev-init` proceed; `applySubtreeControllers` then handles the actual drain at runtime. Lower-blast-radius — leaves the doctor read-only.

The shipped defensive-drain primitive (`internal/ctr/cgroups.go` `RelocateProcessesToLeaf`) is the same one either path would call; the design question is whether the pre-flight should mutate.

A nested-cell smoke target (the deferred follow-up in #336) would catch this regression class going forward.

eminwux:
## Summary

- Issue #342 picked Fix 2 (downgrade fatal to warning, leave doctor read-only). On a `NestedCgroupRuntime` dev cell where the cgroup-namespace root carries processes, `kuke doctor cgroups --probe` writes EBUSY at the mountpoint root, classifies it as `StatusInternalProcess`, and exits non-zero — even though `applySubtreeControllers`'s defensive `__bootstrap` drain (PR #340) would self-heal the same condition at `kuke init` time. Net effect: `make dev-init` aborts at the pre-flight before the runtime gets a chance to fix it.
- The fix recognizes the layer mismatch: when **every** unresolved required controller is `StatusInternalProcess` AND the host fingerprint matches both defensive-drain clauses (`/proc/self/cgroup == 0::/` and the mountpoint root's `cgroup.events` shows `populated 1`), the doctor still surfaces the diagnostic on stderr — operators see what's about to happen — but exits 0 so dev-init.sh proceeds. Restricted to the unscoped host-root path because the runtime drain only operates on the mountpoint root, not on a deeper scoped cgroup.
- Doctor stays read-only (no `RelocateProcessesToLeaf` from the pre-flight). Mixed-class failures (e.g. `internal-process` + `kernel-missing`) stay fatal — the runtime drain cannot rescue the second class.
- New helpers `cgroupcheck.IsCgroupNsRoot()` and `cgroupcheck.IsHostRootPopulated(hostRoot)` mirror the gate clauses `internal/ctr.maybeDrainCgroupNsRoot` reads at runtime; kept as a duplicate of the (private) ctr versions to avoid pulling internal/ctr into the doctor's import graph for a six-line file read.
- Test seams `activeProber` and `selfHealGate` are package vars (with `//nolint:gochecknoglobals` matching the repo's existing convention) so unit tests can simulate the EBUSY trap and a matching/non-matching fingerprint without writing to a real cgroupfs or `/proc/self/cgroup`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full suite, all packages green
- [x] `make dev-init` — daemon parity check matches the expected `default` + `kuke-system` realm/cgroup output (this host doesn't reproduce the NestedCgroupRuntime fingerprint, so the new branch is a no-op here; unit tests cover the new path comprehensively)
- [x] `pre-commit run` on changed files
- [x] New unit tests added: `Result.OnlyInternalProcess` boundaries (all-EBUSY, mixed, fully resolved); `IsHostRootPopulated` reads `cgroup.events` correctly across populated-1, populated-0, missing key, empty file, missing file; doctor cmd `SelfHealDowngradeOnInternalProcess` (warning + exit 0), `SelfHealNoDowngradeWithoutGate` (still fatal), `SelfHealNoDowngradeOnMixedClasses` (still fatal), `SelfHealNoDowngradeOnScoped` (still fatal — runtime drain doesn't apply at scoped paths)

Closes #342

### DIFF
--- a/cmd/kuke/doctor/cgroups/cgroups.go
+++ b/cmd/kuke/doctor/cgroups/cgroups.go
@@ -168,6 +168,25 @@ func runDoctor(cmd *cobra.Command, root string, nested, probe, verbose bool, tar
 	return runCheck(cmd.OutOrStdout(), cmd.ErrOrStderr(), scopedRoot, nested, probe, verbose, target.label())
 }
 
+// activeProber is the cgroupcheck.Prober runCheck uses when --probe is
+// set. Indirected via a package var so unit tests can simulate the EBUSY
+// no-internal-process trap without writing to a real cgroupfs.
+//
+//nolint:gochecknoglobals // test seam for the production prober
+var activeProber = cgroupcheck.DefaultProber
+
+// selfHealGate reports whether the host fingerprint matches the case
+// applySubtreeControllers (internal/ctr) self-heals via its defensive
+// __bootstrap drain (PR #340) — both clauses (cgroup-namespace root and
+// the mountpoint root being populated) must hold. Indirected so unit
+// tests can simulate matching/non-matching hosts without touching
+// /proc/self/cgroup.
+//
+//nolint:gochecknoglobals // test seam for the cgroup-ns fingerprint check
+var selfHealGate = func(root string) bool {
+	return cgroupcheck.IsCgroupNsRoot() && cgroupcheck.IsHostRootPopulated(root)
+}
+
 // runCheck is the shared pre-flight body used by both the host-root and
 // --scope paths. scopeLabel is empty for the host-root path; non-empty
 // labels are printed as a "scope: ..." header so operators can tell which
@@ -181,7 +200,7 @@ func runCheck(stdout, stderr io.Writer, root string, nested, probe, verbose bool
 
 	var prober cgroupcheck.Prober
 	if probe {
-		prober = cgroupcheck.DefaultProber
+		prober = activeProber
 	}
 
 	res, err := cgroupcheck.Check(root, required, prober)
@@ -196,6 +215,27 @@ func runCheck(stdout, stderr io.Writer, root string, nested, probe, verbose bool
 			printScope(stdout, scopeLabel)
 			printStatus(stdout, res)
 		}
+		return nil
+	}
+
+	// Self-heal downgrade: when the only unresolved class is the EBUSY
+	// no-internal-process trap AND the host fingerprint matches what
+	// applySubtreeControllers (internal/ctr) drains via its defensive
+	// __bootstrap leaf (PR #340), `kuke init` would proceed anyway. Surface
+	// the diagnostic so the operator sees what is about to happen, but
+	// exit 0 so dev-init.sh proceeds. Restricted to the unscoped host-root
+	// path because the runtime drain only operates on the mountpoint root,
+	// not on a deeper scoped cgroup — a scoped EBUSY still requires
+	// operator action.
+	if scopeLabel == "" && res.OnlyInternalProcess() && selfHealGate(root) {
+		printScope(stderr, scopeLabel)
+		printStatus(stderr, res)
+		fmt.Fprint(stderr, "\n")
+		fmt.Fprint(stderr, cgroupcheck.FormatRemediation(res))
+		fmt.Fprint(stderr, "\nnote: the runtime will drain the cgroup-namespace root into a "+
+			"__bootstrap leaf\nbefore widening cgroup.subtree_control "+
+			"(internal/ctr applySubtreeControllers,\nPR #340). this is "+
+			"informational; `kuke init` will proceed.\n")
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


eminwux:
**PR #343 Review — `fix: doctor cgroups warns on internal-process trap when runtime self-heals`**

Tight, well-scoped fix for the dev-init regression — the downgrade is gated on three independent conditions (`scopeLabel == ""`, `OnlyInternalProcess()`, and the two-clause fingerprint match) that map exactly to the runtime drain's preconditions in `applySubtreeControllers`. Mixed-class failures and scoped paths stay fatal as they should. Test coverage pins all four boundaries (downgrade fires; gate=false stays fatal; mixed-class stays fatal; scoped stays fatal) plus the `OnlyInternalProcess` and `IsHostRootPopulated` helpers individually. No sensitive data in diagnostics; license headers + colocated tests + `nolint:gochecknoglobals` all match repo conventions.

Cross-checked the gate-clause mirror against `internal/ctr/cgroups.go:386` (`maybeDrainCgroupNsRoot`) and `internal/ctr/cgroups.go:270` (`applySubtreeControllers`) — `IsCgroupNsRoot`/`IsHostRootPopulated` are equivalent reads (`/proc/self/cgroup == 0::/` and `<root>/cgroup.events` "populated 1"), with the doctor opting for conservative-on-error fallback (read failure → false → no downgrade) which is the right call for a pre-flight. Duplication is justified per the PR description (avoiding internal/ctr in the doctor's import graph).

Smoke-test plan acknowledges the host doesn't reproduce the NestedCgroupRuntime fingerprint, so the new branch is a no-op there — unit tests cover comprehensively, which is the right trade-off here.

LGTM.

PR #346, opened 2026-05-07
feat: tmpfs volume kind on ContainerSpec.Volumes

### LINKED ISSUE #309: feat: add tmpfs volume kind to ContainerSpec.Volumes (dev-init phase 2)

eminwux:
## Summary

- `ContainerSpec.Volumes` only carried bind mounts; cells that need an in-memory directory (e.g. an inner containerd snapshotter root that would otherwise hit nested-overlay restrictions) had no spec-side way to declare it.
- Add a `Kind` discriminator on `intmodel.VolumeMount` plus optional `SizeBytes` / `Mode` fields. `Kind=""` and `Kind="bind"` keep existing semantics; `Kind="tmpfs"` produces an OCI `Mount{Type:"tmpfs", Source:"tmpfs", Destination:Target, Options:[size=…, mode=…, rw|ro]}` entry.
- Rename `buildBindMounts` → `buildVolumeMounts` and dispatch on `Kind`. Both internal callers (`spec.go`, `container_utils.go`) updated; existing call sites that do not opt in are byte-identical (asserted by `TestBuildVolumeMounts_BindParity`).
- CLI surface, `v1beta1.VolumeMount`, the apischeme converter, and `validateVolumes` are deliberately untouched per the issue's "out of scope" — the model + builder land first; CLI and validation wiring follow once consumers exist.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full unit suite — `internal/ctr`, `internal/modelhub`, controllers, CLI cmd)
- [x] `go test ./internal/ctr/ -run 'TestBuildVolumeMounts'` covering bind, bind-parity (byte-identical), tmpfs (minimal / size+mode / read-only / skip-empty-target), and bind+tmpfs mixed ordering
- [x] `make dev-init` smoke — daemon-parity tail matches the regression guard exactly:
  ```
  ```
- [x] `pre-commit run` on staged files (gofmt, golangci-lint, addlicense, …)

Closes #309

### DIFF
--- a/internal/ctr/container_utils.go
+++ b/internal/ctr/container_utils.go
@@ -142,7 +142,7 @@ func BuildRootContainerSpec(
 	// BuildContainerSpec so a user-supplied root container (RootContainerID)
 	// keeps its Volumes, User, Capabilities, SecurityOpts, Tmpfs, and
 	// Resources instead of having them silently dropped.
-	if mounts := buildBindMounts(rootSpec.Volumes); len(mounts) > 0 {
+	if mounts := buildVolumeMounts(rootSpec.Volumes); len(mounts) > 0 {
 		specOpts = append(specOpts, oci.WithMounts(mounts))
 	}
 
--- a/internal/ctr/spec.go
+++ b/internal/ctr/spec.go
@@ -300,7 +300,7 @@ func BuildContainerSpec(
 		specOpts = append(specOpts, oci.WithMounts([]runtimespec.Mount{nestedCgroupMount()}))
 	}
 
-	if mounts := buildBindMounts(containerSpec.Volumes); len(mounts) > 0 {
+	if mounts := buildVolumeMounts(containerSpec.Volumes); len(mounts) > 0 {
 		specOpts = append(specOpts, oci.WithMounts(mounts))
 	}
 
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #346 Review — `feat: tmpfs volume kind on ContainerSpec.Volumes`**

Clean groundwork for #309 — `Kind` discriminator threaded through `buildVolumeMounts` with empty/`VolumeKindBind` preserving the historical bind path and `VolumeKindTmpfs` emitting an OCI tmpfs entry. `TestBuildVolumeMounts_BindParity` locks byte-identical output for existing all-bind callers (including the kukeond cell spec), which is the right regression guard for a rename + dispatch refactor. No correctness, dead-code, or sensitive-data concerns.

Scope: v1beta1.VolumeMount, the apischeme converter, and CLI wiring stay deliberately out of scope per the issue — the new fields are unreachable from any external API path until that follow-up lands, and that's fine for groundwork. Smoke `make dev-init` is checked in the test plan.

LGTM.

PR #348, opened 2026-05-07
feat: plumb VolumeMount.Kind=tmpfs to v1beta1, apischeme, and validator

### LINKED ISSUE #347: feat: plumb VolumeMount.Kind=tmpfs to v1beta1/apischeme/validator/clone (dev-init phase 2 follow-up)
## Context

PR #346 (issue #309) landed `VolumeMount.Kind` (`bind` | `tmpfs`) plus `SizeBytes` / `Mode` on the **internal** model (`internal/modelhub/container.go:122`) and wired the OCI translator (`internal/ctr/spec.go:buildVolumeMounts`). The follow-up: making the new kind reachable from a YAML cell or cellprofile.

Today the chain breaks at the v1beta1 surface, the apischeme converter, and the controller validator — none of them know about `Kind`. So a user who writes:

```yaml
# cell.yaml or profile
spec:
  containers:
    - name: dev
      volumes:
        - kind: tmpfs
          target: /var/lib/containerd
          sizeBytes: 1073741824
```

…silently loses `kind` and `sizeBytes` at the v1beta1→intmodel boundary, then the validator rejects the entry because `source` is empty.

A second, **already-working** path exists: `ContainerSpec.Tmpfs []ContainerTmpfsMount` (`pkg/api/model/v1beta1/container.go:176`, plumbed end-to-end via `convertTmpfsMountsToInternal` at `internal/apischeme/scheme.go:582`). Both paths translate to the same OCI `Mount{Type:"tmpfs"}`; once this issue lands they overlap. Resolving the overlap is part of the design call below.

### 1. Extend `v1beta1.VolumeMount`

`pkg/api/model/v1beta1/container.go:162` — mirror the intmodel shape with JSON/YAML tags:

```go
type VolumeKind string

const (
    VolumeKindBind  VolumeKind = "bind"
    VolumeKindTmpfs VolumeKind = "tmpfs"
)

type VolumeMount struct {
    Kind      VolumeKind `json:"kind,omitempty"      yaml:"kind,omitempty"`
    Source    string     `json:"source,omitempty"    yaml:"source,omitempty"`
    Target    string     `json:"target"              yaml:"target"`
    ReadOnly  bool       `json:"readOnly,omitempty"  yaml:"readOnly,omitempty"`
    SizeBytes int64      `json:"sizeBytes,omitempty" yaml:"sizeBytes,omitempty"`
    Mode      uint32     `json:"mode,omitempty"      yaml:"mode,omitempty"`
}
```

Empty `Kind` continues to mean `bind` (back-compat with every existing YAML).

### 2. Round-trip through apischeme

`internal/apischeme/scheme.go:735` (`volumeMountsToInternal`) and `:750` (`volumeMountsToExternal`) — copy `Kind`, `SizeBytes`, `Mode`. No new helpers needed.

### 3. Make the validator kind-aware

`internal/controller/create_container.go:279` (`validateVolumes`) — current rules require non-empty absolute `Source` and a host-side `os.Stat`. Split per kind:

- `Kind == "" || Kind == "bind"` — keep the existing checks unchanged.
- `Kind == "tmpfs"` — require `Target` (absolute), reject any non-empty `Source` (no host backing for a tmpfs), accept `SizeBytes >= 0` and any `Mode`.
- Unknown `Kind` — reject with a new `errdefs.ErrVolumeKindUnknown` sentinel (`internal/errdefs/`).

### 4. Preserve fields on clone

`pkg/api/model/v1beta1/container.go:335` (`cloneVolumeMounts`) — copy the new fields. This is what `CellDoc.Clone` and `CellProfileDoc.Materialize` use; without it a profile would silently drop `kind` on instantiation.

### 5. Design call: relationship to `ContainerSpec.Tmpfs[]`

After this lands, both `Spec.Volumes[].Kind=tmpfs` and `Spec.Tmpfs[]` (existing) describe the same OCI mount. Recommended call (open to override):

- **Keep both** for one minor release. Volumes[] is the canonical surface going forward.
- Add a doc note on `ContainerSpec.Tmpfs` (`pkg/api/model/v1beta1/container.go:91`) marking it as superseded by `Volumes[].Kind=tmpfs`, with a deprecation timeline.
- A future cleanup issue (file separately, do not bundle here) collapses `Spec.Tmpfs[]` callers onto `Volumes[]` and removes the field.

If reviewer prefers immediate consolidation, scope expands beyond what fits in one PR — flag for resize.

## Acceptance criteria

- [ ] `v1beta1.VolumeMount` carries `Kind` (with `VolumeKindBind` / `VolumeKindTmpfs` consts), `SizeBytes`, and `Mode` with stable JSON/YAML tags. Empty `Kind` deserializes as bind.
- [ ] `volumeMountsToInternal` / `volumeMountsToExternal` in `internal/apischeme/scheme.go` copy the new fields; an existing all-bind YAML round-trips byte-identically through the converter.
- [ ] `validateVolumes` accepts `kind=tmpfs` entries (target-only, no host stat) and continues to reject the existing bind error classes.
- [ ] `cloneVolumeMounts` (and any `CellDoc`/`CellProfileDoc` clone path it feeds) preserves `Kind` / `SizeBytes` / `Mode`.
- [ ] Unit test in `internal/apischeme/scheme_test.go` covering tmpfs round-trip on `ContainerSpec` and on `CellSpec.Containers[]`.
- [ ] Unit test in `internal/controller/validate_volumes_test.go` covering the tmpfs accept path and the rejected-source case.
- [ ] A `kuke create -f profile.yaml` (or `kuke run -p`) referencing `kind: tmpfs` in `volumes:` reaches `internal/ctr/spec.go:buildVolumeMounts` with the kind preserved (smoke check, not a CI test).

## Out of scope

- **CLI surface** (`kuke create container --volume tmpfs:/path,size=…`). The existing `--tmpfs` flag still routes through `Spec.Tmpfs[]`; switching it to populate `Spec.Volumes[].Kind=Tmpfs` is a separate follow-up that should follow the design-call resolution above. File separately after this lands.
- **`Spec.Tmpfs[]` removal / deprecation enforcement** — see design call.
- **`scripts/dev-init.sh` consumption of the new YAML form** — issue #285 (phase 3) territory.

## References

- PR #346 (model + ctr translator) — closes #309.
- Issue #309 (this issue's parent, "Out of scope" lists CLI and dev-init.sh as deferred).
- Phase 3 dev-init streamlining: #285.

eminwux:
## Summary
- Extends `pkg/api/model/v1beta1.VolumeMount` with `Kind` (`VolumeKindBind` / `VolumeKindTmpfs` consts), `SizeBytes`, and `Mode`. Empty `Kind` deserializes as bind so every existing YAML round-trips byte-identically.
- Round-trips the new fields through `internal/apischeme.scheme.go` (`volumeMountsToInternal` / `volumeMountsToExternal`).
- Makes `internal/controller.validateVolumes` kind-aware: bind keeps the existing `os.Stat` + absolute-source contract; tmpfs requires an absolute target and rejects any non-empty source; unknown kinds are rejected with a new `errdefs.ErrVolumeKindUnknown`. `errdefs.ErrVolumeTmpfsSourceForbidden` is the tmpfs-specific source error.
- New unit tests in `scheme_test.go` for tmpfs round-trip on both `ContainerSpec` and `CellSpec.Containers[]`, and in `validate_volumes_test.go` for tmpfs accept paths, the rejected-source case, target validation, the unknown-kind reject, and an explicit field-preservation assertion.
- The downstream OCI translator already dispatches on `Kind` (PR #346), so this change closes the gap between user-authored YAML and `internal/ctr/spec.go:buildVolumeMounts`.

`cloneVolumeMounts` already does `copy()` on a value-type slice, so the new fields are preserved automatically — no change needed there.

The design call from issue #347 (keeping `Spec.Tmpfs[]` for one minor release with a deprecation note as a separate follow-up) is unchanged; this PR ships only the v1beta1 / apischeme / validator layer per the AC.

## Test plan
- [x] `make test` (full project unit suite, all packages green)
- [x] `go build ./...` and `go vet ./...`
- [x] `pre-commit run` on the touched files (gofmt, golangci-lint, prettier, addlicense — all green)
- [x] `make dev-init` smoke (daemon-parity output matches the regression guard)
- [x] Smoke check per AC item 7 — applied a Cell with both `kind: tmpfs` (sizeBytes: 16777216, mode: 0o0755) and a bind `readOnly` volume on a non-root container, confirmed the container's containerd OCI spec ships:
  - `{type:"tmpfs", destination:"/var/lib/cache", options:["size=16777216","mode=0755","rw"]}`
  - `{type:"bind", destination:"/host-tmp", source:"/tmp", options:["rbind","ro"]}`

  i.e. the new fields reach `internal/ctr/spec.go:buildVolumeMounts` with `Kind` / `SizeBytes` / `Mode` preserved end-to-end.

Closes #347

### DIFF
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -739,9 +739,12 @@ func volumeMountsToInternal(in []ext.VolumeMount) []intmodel.VolumeMount {
 	out := make([]intmodel.VolumeMount, len(in))
 	for i, v := range in {
 		out[i] = intmodel.VolumeMount{
-			Source:   v.Source,
-			Target:   v.Target,
-			ReadOnly: v.ReadOnly,
+			Kind:      intmodel.VolumeKind(v.Kind),
+			Source:    v.Source,
+			Target:    v.Target,
+			ReadOnly:  v.ReadOnly,
+			SizeBytes: v.SizeBytes,
+			Mode:      v.Mode,
 		}
 	}
 	return out
@@ -754,9 +757,12 @@ func volumeMountsToExternal(in []intmodel.VolumeMount) []ext.VolumeMount {
 	out := make([]ext.VolumeMount, len(in))
 	for i, v := range in {
 		out[i] = ext.VolumeMount{
-			Source:   v.Source,
-			Target:   v.Target,
-			ReadOnly: v.ReadOnly,
+			Kind:      ext.VolumeKind(v.Kind),
+			Source:    v.Source,
+			Target:    v.Target,
+			ReadOnly:  v.ReadOnly,
+			SizeBytes: v.SizeBytes,
+			Mode:      v.Mode,
 		}
 	}
 	return out
--- a/internal/apischeme/scheme_test.go
+++ b/internal/apischeme/scheme_test.go
@@ -707,6 +707,141 @@ func TestContainerRoundTripV1Beta1(t *testing.T) {
 	}
 }
 
+// TestContainerRoundTripVolumeKindTmpfsV1Beta1 pins the tmpfs path on
+// VolumeMount: a YAML author who writes `kind: tmpfs` with size/mode must see
+// those fields survive Normalize → controller → Build with no drops, alongside
+// a plain bind entry (empty Kind preserves bind back-compat).
+func TestContainerRoundTripVolumeKindTmpfsV1Beta1(t *testing.T) {
+	input := ext.ContainerDoc{
+		APIVersion: ext.APIVersionV1Beta1,
+		Kind:       ext.KindContainer,
+		Metadata: ext.ContainerMetadata{
+			Name: "container-tmpfs",
+		},
+		Spec: ext.ContainerSpec{
+			ID:      "container-tmpfs",
+			RealmID: "realm0",
+			SpaceID: "space0",
+			StackID: "stack0",
+			CellID:  "cell0",
+			Image:   "alpine:latest",
+			Volumes: []ext.VolumeMount{
+				{Source: "/host/data", Target: "/data"},
+				{
+					Kind:      ext.VolumeKindTmpfs,
+					Target:    "/var/lib/containerd",
+					SizeBytes: 1 << 30,
+					Mode:      0o0755,
+				},
+				{
+					Kind:     ext.VolumeKindBind,
+					Source:   "/host/ro",
+					Target:   "/ro",
+					ReadOnly: true,
+				},
+			},
+		},
+	}
+
+	internal, version, err := apischeme.NormalizeContainer(input)
+	if err != nil {
+		t.Fatalf("NormalizeContainer: %v", err)
+	}
+	if version != ext.APIVersionV1Beta1 {
+		t.Fatalf("unexpected version: %s", version)
+	}
+	if got, want := len(internal.Spec.Volumes), len(input.Spec.Volumes); got != want {
+		t.Fatalf("internal volumes len = %d, want %d", got, want)
+	}
+	if internal.Spec.Volumes[1].Kind != intmodel.VolumeKindTmpfs {
+		t.Errorf("internal volume[1].Kind = %q, want %q",
+			internal.Spec.Volumes[1].Kind, intmodel.VolumeKindTmpfs)
+	}
+	if internal.Spec.Volumes[1].SizeBytes != 1<<30 {
+		t.Errorf("internal volume[1].SizeBytes = %d, want %d",
+			internal.Spec.Volumes[1].SizeBytes, 1<<30)
+	}
+	if internal.Spec.Volumes[1].Mode != 0o0755 {
+		t.Errorf("internal volume[1].Mode = %o, want %o",
+			internal.Spec.Volumes[1].Mode, 0o0755)
+	}
+
+	output, err := apischeme.BuildContainerExternalFromInternal(internal, version)
+	if err != nil {
+		t.Fatalf("BuildContainerExternalFromInternal: %v", err)
+	}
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #348 Review — `feat: plumb VolumeMount.Kind=tmpfs to v1beta1, apischeme, and validator`**

Clean, well-scoped extension that closes the YAML→runtime gap left by #346. Empty `Kind` deserializing as bind preserves byte-identical round-trips for existing manifests, the new `errdefs` sentinels follow the shared-sentinels convention, and the `Container` and nested `CellSpec.Containers[]` round-trips are both covered. Validator reuses the universal target check before dispatching on `Kind`, and `volumeMountsEqual` (`internal/controller/apply/diff.go:763`) uses struct equality so the new fields participate in diff automatically — no parity gap there. No correctness, leak, or dead-symbol issues. Test plan is complete, including the `make dev-init` regression guard and the AC-7 end-to-end OCI-mount check.

A couple of small notes (non-blocking):

- **Docs drift** `docs/site/manifests/container.md:62-79`: the public manifest reference still describes `VolumeMount` as "a bind mount of a host path into the container" with no mention of `kind: tmpfs`, `sizeBytes`, or `mode`. The new fields are user-authorable v1beta1 surface, so YAML authors won't discover them from the docs. Worth a follow-up doc PR.
- **Validator error order** `internal/controller/create_container.go:283-307`: an unknown-kind entry with an empty target reports `ErrVolumeTargetRequired` because target validation runs before the kind switch. Minor UX nit — the user fixes the target, then learns the kind is also wrong. Could move the kind-check earlier, or fold it into the same pass. Not worth blocking.

LGTM.

PR #352, opened 2026-05-07
fix: cell hostname is cell name with managed /etc/hosts and /etc/hostname

### LINKED ISSUE #345: kukeond: short hostname + managed /etc/hosts (Docker/k8s-style)
## Context

Today, kukeond pins each container's UTS hostname to the **hierarchical containerd ID** — `{space}_{stack}_{cell}_{container}`, e.g. `default_default_kukeon-dev-12b670_work` (built at `internal/util/naming/naming.go:64`, applied via `oci.WithHostname(containerdID)` at `internal/ctr/spec.go:229` and `internal/ctr/container_utils.go:87`). It also never generates an `/etc/hosts` or `/etc/hostname` for the container — those files are whatever the image ships with, which won't contain the runtime-assigned hostname.

The visible symptom: in any kukeon container, `getent hosts $(hostname)` returns nothing, so `sudo` (and any tool that resolves its own hostname) emits `sudo: unable to resolve host <hostname>: Name or service not known` and pays a multi-second DNS-lookup timeout per call. Originally surfaced as `dev-init.sh` noise (the user was running dev-init *inside* a kukeon container), but the cause is kukeond's container setup, not the script. A script-side pre-flight only catches the case where the user's *host* hostname is unconfigured — it doesn't fix the much more common case of running in a kukeon container.

The composite hostname is also bad UX on its own: `default_default_kukeon-dev-12b670_work` is 39 characters, contains underscores (which RFC 1123 disallows for DNS labels), and leaks implementation detail. The cell name (`kukeon-dev-12b670`) is the natural identity, and because all containers in a cell share a single network namespace and a single UTS namespace (joined via `JoinContainerNamespaces` at `internal/ctr/spec.go:38`), one cell-wide hostname is the right granularity.

## Proposal

Three changes in `kukeond`, modeled on how Docker and Kubernetes handle this:

**1. UTS hostname = cell name, set on the root container only.** Replace the `oci.WithHostname(containerdID)` call at `internal/ctr/container_utils.go:87` (root container) with `oci.WithHostname(cellName)`. Drop the call at `internal/ctr/spec.go:229` (non-root containers): non-root containers join the root's UTS namespace via `JoinContainerNamespaces`, so they inherit the hostname automatically — setting it again on each non-root container is redundant. Same model as a Kubernetes pod, where the pod hostname is shared across all containers.

**2. Generate and bind-mount `/etc/hosts`.** kukeond writes a per-container `/etc/hosts` file to the container's runtime state directory and adds a bind-mount of that file to `/etc/hosts` in the OCI spec. Mirrors Docker's `/var/lib/docker/containers/<id>/hosts` and Kubernetes kubelet's `/var/lib/kubelet/pods/<uid>/etc-hosts` pattern. File contents:

```
127.0.0.1       localhost
::1             localhost ip6-localhost ip6-loopback
fe00::          ip6-localnet
ff00::          ip6-mcastprefix
ff02::1         ip6-allnodes
ff02::2         ip6-allrouters
<cell-ip>       <cell-name>
```

The cell IP is the address CNI assigned to the cell's netns. Cell name is the short identifier (`kukeon-dev-12b670`). Because all containers in the cell share the netns, the same `/etc/hosts` content is correct for every container in the cell — it can be rendered once per cell and bind-mounted into each container.

**3. Generate and bind-mount `/etc/hostname`.** kukeond writes `<cell-name>\n` to a per-cell file in the runtime state directory and bind-mounts it at `/etc/hostname` in each container's OCI spec. Same Docker/Kubernetes pattern; ensures `cat /etc/hostname` agrees with `hostname` (the UTS value) for tools that read the file directly.

Order matters. The cell IP is not known until CNI attach completes (after root-container creation but before workload start). The render flow:

1. Cell created → root container created with `WithHostname(cellName)`. `/etc/hostname` rendered now (cell-name only, no IP needed).
2. CNI attach completes → cell IP known. `/etc/hosts` rendered now.
3. Non-root containers created → they bind-mount the already-rendered `/etc/hosts` and `/etc/hostname` from the cell's state directory, and inherit the hostname via UTS-namespace join.

## Acceptance criteria

- [ ] In a freshly-created cell, `kuke exec <cell> -- hostname` returns the cell name (e.g. `kukeon-dev-12b670`), not the composite ID.
- [ ] In any container of the cell (root or non-root), `kuke exec <cell> -c <container> -- cat /etc/hostname` prints the cell name.
- [ ] In any container, `getent hosts $(hostname)` returns one line: `<cell-ip>  <cell-name>`.
- [ ] `/etc/hosts` inside the container contains the standard localhost block (IPv4 + IPv6) plus the cell-IP line.
- [ ] `sudo true` inside a kukeon container produces no `unable to resolve host` warning and no DNS-lookup timeout.
- [ ] The bind-mount source paths live under the cell's runtime state directory, so `kuke daemon reset` (and cell teardown) cleans them up.
- [ ] `oci.WithHostname` is called only on the root container (the redundant non-root call at `internal/ctr/spec.go:229` is dropped).
- [ ] OCI-spec tests in `internal/ctr/spec_test.go` and `internal/ctr/container_utils_test.go` are updated to assert the new hostname value, the dropped non-root call, and the new `/etc/hosts` and `/etc/hostname` bind-mount entries.

## Suggested slicing

Smallest landable PR first; deferred work as follow-ups:

1. **(this issue)** Hostname → cell-name (root-only) + per-cell `/etc/hosts` (localhost block + `<cell-ip> <cell-name>`) + per-cell `/etc/hostname`, bind-mounted into every container in the cell. Covers the reported symptom end-to-end.
2. **(follow-up)** Sibling-name resolution — each container's `/etc/hosts` includes every sibling container's basename as an alias (`<cell-ip> kukeon-dev-12b670 work db cache`), so containers in the same cell can resolve each other by short name. Adds a per-container render variant of the file (different alias set per container) on top of the per-cell base, so it's strictly additive to this issue's design.
3. **(follow-up)** User-supplied host aliases at the cell-spec level (Kubernetes' `pod.spec.hostAliases` analogue) — for operators who want to inject extra `<ip> <name>` lines. File only when a concrete use case appears.
4. **(follow-up)** `/etc/resolv.conf` management — same Docker/k8s per-file-bind-mount pattern. Separate concern from hostname-resolution-via-hosts; DNS resolution is a different axis.

## Notes

- Original symptom thread: this issue's prior body proposed a `scripts/dev-init.sh` pre-flight; that was a script-side patch on a kukeond bug. Dropped.
- Sibling issue #344 (containerd reachability pre-flight in `dev-init.sh`) is unrelated and stays as-is.
- The hierarchical containerd ID at `internal/util/naming/naming.go:64` continues to be the containerd-side identifier (used for `ctr` operations and labels). Only the in-container UTS hostname, `/etc/hosts`, and `/etc/hostname` change.

eminwux:
## Summary
- Replace the hierarchical containerd id with the cell name as the OCI hostname on the cell's root container, drop the redundant `oci.WithHostname` on every other container so they inherit the value through the shared UTS namespace, and bind-mount per-cell `/etc/hostname` and `/etc/hosts` files into every container so identity-resolving tools (`sudo`, `getent hosts`, `hostname -s`) see the same value the kernel does. Fixes the `sudo: unable to resolve host …` warnings and the multi-second DNS-lookup timeouts inside kuke-system pods.
- The host source files live under each cell's metadata directory at `<runPath>/<realm>/<space>/<stack>/<cell>/{etc-hostname,etc-hosts}` so cell teardown cleans them up. `/etc/hostname` is rendered at cell create-time (no IP needed); `/etc/hosts` is rendered with the localhost block before CNI ADD and re-rendered with the CNI-assigned cell IP after `AddContainerToNetwork` succeeds. The same paths are stamped on every container in the cell so root and non-root agree on the file inode, and the `/etc/hosts` mount is suppressed for `HostNetwork=true` roots (the kukeond carve-out where the host's `/etc/hosts` is the right view).
- `cni.Manager.AddContainerToNetwork` now returns the first IPv4 from the CNI result so the runner can populate the IP line, and a sibling `CachedIPv4ForContainer` reads `libcni`'s IPAM cache for the `ErrCNIVethExists` idempotency path on cell restart.
- Fix-up `cab2aaa`: the second `renderCellEtcFilesPreCNI` call in `ensureCellContainers` (provision.go:1808) was clobbering `/etc/hosts` on every `update_cell`/`update_container` tick of a running cell — the pre-CNI render truncates the file back to the localhost-only block, dropping the cell-IP line that the post-CNI render had written. Replaced with a new `ensureCellEtcFilesExistPreCNI` helper that fills in missing files but never rewrites existing content; the no-root branch's first-creation render is unchanged.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full project test suite)
- [x] `internal/ctr` unit tests cover the new contract: hostname is the cell name on root, falls back to containerd-id when CellName is empty, non-root never sets `oci.WithHostname`, and `/etc/hosts` + `/etc/hostname` bind-mount entries appear with `rbind,ro` when `EtcHostsPath`/`EtcHostnamePath` are stamped on the model spec
- [x] `internal/controller/runner/cell_etc_files_test.go` — regression guard for the fix-up: `ensureCellEtcFilesExistPreCNI` preserves an existing `/etc/hosts` that already carries the cell-IP line, fills in missing files when absent, and skips `/etc/hosts` on host-network cells while still producing `/etc/hostname`
- [ ] `make dev-init` smoke test — deferred to reviewer/user's docker-capable host. Dev host has the docker CLI but no running docker daemon (`docker ps` → "dial unix /var/run/docker.sock: connect: no such file or directory"), so the `docker build` phase of `scripts/dev-init.sh` cannot run. The fix-up is also covered by the new unit test above and by the existing `internal/ctr` tests for the bind-mount contract.

Closes #345

### DIFF
--- a/internal/cni/container.go
+++ b/internal/cni/container.go
@@ -18,20 +18,70 @@ package cni
 
 import (
 	"context"
+	"net"
 
 	libcni "github.com/containernetworking/cni/libcni"
+	cnitypes "github.com/containernetworking/cni/pkg/types"
+	current "github.com/containernetworking/cni/pkg/types/100"
 	"github.com/eminwux/kukeon/internal/errdefs"
 )
 
-// AddContainerToNetwork adds a container to the CNI network.
-func (m *Manager) AddContainerToNetwork(ctx context.Context, containerID, netnsPath string) error {
+// AddContainerToNetwork adds a container to the CNI network and returns the
+// container's IPv4 address as assigned by IPAM, or nil when no IPv4 address
+// was returned (e.g. IPv6-only networks). The IP is the one the runner
+// renders into the cell's /etc/hosts so tools that resolve the container's
+// own hostname work without DNS-lookup timeouts (issue #345).
+func (m *Manager) AddContainerToNetwork(ctx context.Context, containerID, netnsPath string) (net.IP, error) {
 	if m.netConf == nil {
-		return errdefs.ErrNetworkConfigNotLoaded
+		return nil, errdefs.ErrNetworkConfigNotLoaded
 	}
 
 	rt := buildRuntimeConf(containerID, netnsPath)
-	_, err := m.cniConf.AddNetworkList(ctx, m.netConf, rt)
-	return translateCNIError(err, m.netConf.Name, bridgeNameFromNetConf(m.netConf))
+	rawResult, err := m.cniConf.AddNetworkList(ctx, m.netConf, rt)
+	if err != nil {
+		return nil, translateCNIError(err, m.netConf.Name, bridgeNameFromNetConf(m.netConf))
+	}
+	return firstIPv4FromResult(rawResult), nil
+}
+
+// CachedIPv4ForContainer returns the IPv4 address libcni cached for the
+// given containerID under the manager's current network config, or nil when
+// no usable IPv4 entry is cached. Used by the runner on the idempotent-skip
+// path (ErrCNIVethExists) — when AddContainerToNetwork failed because veth
+// setup already ran, the IPAM allocation persisted in the cache and is the
+// authoritative source for the cell IP without re-issuing CNI ADD.
+func (m *Manager) CachedIPv4ForContainer(containerID, netnsPath string) net.IP {
+	if m.netConf == nil {
+		return nil
+	}
+	rt := buildRuntimeConf(containerID, netnsPath)
+	rawResult, err := m.cniConf.GetNetworkListCachedResult(m.netConf, rt)
+	if err != nil || rawResult == nil {
+		return nil
+	}
+	return firstIPv4FromResult(rawResult)
+}
+
+// firstIPv4FromResult extracts the first IPv4 address from a CNI result,
+// converting through the libcni current schema so callers can stay agnostic
+// of which CNI version the bridge plugin emits.
+func firstIPv4FromResult(rawResult cnitypes.Result) net.IP {
+	if rawResult == nil {
+		return nil
+	}
+	res, err := current.NewResultFromResult(rawResult)
+	if err != nil || res == nil {
+		return nil
+	}
+	for _, ipc := range res.IPs {
+		if ipc == nil {
+			continue
+		}
+		if v4 := ipc.Address.IP.To4(); v4 != nil {
+			return v4
+		}
+	}
+	return nil
 }
 
 // DelContainerFromNetwork removes a container from the CNI network.
--- a/internal/cni/container_test.go
+++ b/internal/cni/container_test.go
@@ -59,13 +59,16 @@ func TestManager_AddContainerToNetwork(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			mgr := tt.setup(t)
-			err := mgr.AddContainerToNetwork(context.Background(), tt.containerID, tt.netnsPath)
+			ip, err := mgr.AddContainerToNetwork(context.Background(), tt.containerID, tt.netnsPath)
 
 			if tt.wantErr != nil {
 				if err == nil {
 					t.Errorf("AddContainerToNetwork() error = nil, want %v", tt.wantErr)
 				} else if !errors.Is(err, tt.wantErr) {
 					t.Errorf("AddContainerToNetwork() error = %v, want %v", err, tt.wantErr)
+				}
+				if ip != nil {
+					t.Errorf("AddContainerToNetwork() ip = %v, want nil on error", ip)
 				}
 			} else {
 				if err != nil {
--- a/internal/cni/network_test.go
+++ b/internal/cni/network_test.go
@@ -106,7 +106,7 @@ func TestManager_LoadNetworkConfigList(t *testing.T) {
 				// Verify config was loaded by trying to use it
 				// We can't directly access netConf, but we can verify it's loaded
 				// by checking that AddContainerToNetwork doesn't return ErrNetworkConfigNotLoaded
-				err := mgr.AddContainerToNetwork(context.Background(), "test-container", "/proc/123/ns/net")
+				_, err := mgr.AddContainerToNetwork(context.Background(), "test-container", "/proc/123/ns/net")
 				if err != nil && !errors.Is(err, errdefs.ErrNetworkConfigNotLoaded) {
 					// Config is loaded (error is from libcni, not from missing config)
 					return
--- a/internal/controller/runner/cell_etc_files.go
+++ b/internal/controller/runner/cell_etc_files.go
@@ -170,6 +170,41 @@ func (r *Exec) renderCellEtcFilesPreCNI(cell *intmodel.Cell) error {
 	return renderCellEtcHosts(hostsPath, cellName, nil)
 }
 
+// ensureCellEtcFilesExistPreCNI guarantees the per-cell /etc/hostname and
+// /etc/hosts source files exist without rewriting them when they already
+// do. Use this on call paths where a prior StartCell may have rendered
+// /etc/hosts with the cell IP — an unconditional pre-CNI re-render would
+// truncate the IP line and regress the DNS-lookup fix issue #345 ships.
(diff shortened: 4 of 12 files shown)


eminwux:
**PR #352 Review — `fix: cell hostname is cell name with managed /etc/hosts and /etc/hostname`**

The hostname-on-root + UTS-inheritance shape is right, the `HostNetwork` carve-out is correctly threaded through both spec builders, and the unit tests pin the contract well. No sensitive data leaks. One correctness bug to fix.

A few small items:

- **`/etc/hosts` clobber on update** `internal/controller/runner/provision.go:1808`: `ensureCellContainers` runs `renderCellEtcFilesPreCNI` unconditionally near its tail, truncating the cell's `/etc/hosts` back to the localhost-only block. The *"idempotent on the existing-root branch (re-renders the same content)"* comment is wrong — pre-CNI content (no IP line) differs from post-CNI content (IP line present), so any `update_cell` / `update_container` invocation on a running cell drops the cell IP entry from `/etc/hosts` and regresses the DNS-lookup fix this PR ships. The bind-mount points at the file inode, so the running root sees the new (broken) content immediately. Fix: skip the render when the file already has the IP, or recover the IP via `cniMgr.CachedIPv4ForContainer` here and call `renderCellEtcHostsWithIP` instead of the no-IP variant.

- **Smoke test unchecked** `CLAUDE.md`: `make dev-init` is the daemon-parity regression guard kukeon `CLAUDE.md` requires for changes touching daemon-read state — this PR adds new bind-mount sources under each cell's metadata directory and rewrites the hostname path. Dev-host had no docker; the user should run it on a docker-capable host before merging.

LGTM once the `/etc/hosts` clobber is fixed and the `make dev-init` smoke verifies on a docker host.

eminwux:
Addressed in cab2aaa:

- /etc/hosts clobber on update — replaced the unconditional `renderCellEtcFilesPreCNI` at `provision.go:1808` with a new `ensureCellEtcFilesExistPreCNI` helper that only renders the per-cell `/etc/hostname` / `/etc/hosts` source files when they are absent. On the existing-root branch (where `update_cell`/`update_container` ticks land) a prior `StartCell` may already have rendered the cell-IP line via `renderCellEtcHostsWithIP`; the new helper preserves that content. The no-root branch's first-creation render at `provision.go:1761` is unchanged. New `cell_etc_files_test.go` covers the three cases: existing IP-rendered `/etc/hosts` is preserved, missing files are filled in, and host-network cells still skip `/etc/hosts` while producing `/etc/hostname`.

Deferred: `make dev-init` — dev host has the docker CLI but the docker daemon socket is absent, so the `docker build kukeon-local:dev` phase of `scripts/dev-init.sh` cannot run. Per CLAUDE.md §When a reviewer-blocking verification step can't safely run on the current host, deferring the smoke run to the reviewer's docker-capable host. The bind-mount contract is unit-covered by the new test plus the existing `internal/ctr` tests.

eminwux:
**PR #352 Re-review — fix-up `cab2aaa1`**

Clobber blocker resolved. The new `ensureCellEtcFilesExistPreCNI` (`internal/controller/runner/cell_etc_files.go:179`) only renders missing files and never rewrites existing content, and the `ensureCellContainers` tail (`provision.go:1810`) now calls it instead of the unconditional pre-CNI render. `TestEnsureCellEtcFilesExistPreCNI_PreservesIPLine` (`cell_etc_files_test.go:53`) seeds the post-StartCell `/etc/hosts` with the cell-IP line and asserts the helper preserves it byte-for-byte — exact regression guard for the bug. Verified the remaining `renderCellEtcFilesPreCNI` call sites (`provision.go:1430`, `:1761`, `start.go:338`) are all fresh-create or pre-CNI-ADD paths where the localhost-only render is correct and gets superseded by `renderCellEtcHostsWithIP` before any container observes the file.

LGTM. The `make dev-init` smoke remains the pre-merge verification step on a docker-capable host (dev's host has the CLI but no daemon, so the unit-level regression here covers the fix-up while the runtime bind-mount + hostname propagation is what the smoke exercises).

PR #353, opened 2026-05-08
feat: expose KUKEON_* identity env vars to container processes

### LINKED ISSUE #351: Expose KUKEON_* env vars (cell profile, cell, container, realm/space/stack) to container processes
## Problem

Today, when authoring a `CellProfile`, anything inside the container that wants to know its own context — the profile name, the cell name, the realm/space/stack — has to be **manually hardcoded** in the `env:` block, e.g.:

```yaml
env:
  - AGENT_NAME=kukeon-pr
```

This drifts from `metadata.name` over time. Auditing my own profiles I just found four mismatches: `AGENT_NAME=kuke` left over as a placeholder in two profiles, plus typos `kukeon-dev-addres` (missing `s`) and `kukeon-pre-review-kimi` (`pre` vs `pr`). Anything that reads `\$AGENT_NAME` (status lines, logs, telemetry, agent dispatch) silently uses the wrong identity.

Kukeon already has all this metadata authoritatively — it's right there in `kuke get cell <name> -o yaml`. The container should just receive it.

## Proposal

When kukeond constructs the env for a container, prepend:

| Variable | Value | Source |
|---|---|---|
| `KUKEON_CELL_PROFILE_NAME` | profile `metadata.name` | the YAML the cell was created from |
| `KUKEON_CELL_NAME` | e.g. `kukeon-pr-a4aaab` | cell metadata |
| `KUKEON_CONTAINER_ID` | e.g. `work`, `root` | container spec |
| `KUKEON_REALM` | e.g. `default` | cell metadata |
| `KUKEON_SPACE` | e.g. `default` | cell metadata |
| `KUKEON_STACK` | e.g. `default` | cell metadata |
| `KUKEON_CGROUP_PATH` | e.g. `/kukeon/default/default/default/kukeon-pr-a4aaab` | already in status |

User-defined env in the profile would still win on collision (so existing `AGENT_NAME=...` keeps working), but profiles can drop the manual var and reference `\$KUKEON_CELL_PROFILE_NAME` instead.

## Why this is small

- All values are already known to `kukeond` at container-start time — no new plumbing, just an env merge.
- Naming is conventional, no schema change.
- Backwards compatible — purely additive.

## Use cases

- Status line in agent containers can show real identity without per-profile boilerplate.
- Log/audit forwarding can tag events without trusting profile authors to set `AGENT_NAME` correctly.
- `\$KUKEON_CELL_NAME` is genuinely useful — currently containers can't easily know their own cell name without parsing the cgroup path.

eminwux:
## Summary

- Adds `kukeonContainerEnv` / `kukeonDefaultEnv` helpers in `internal/ctr/spec.go` that emit `KUKEON_CELL_PROFILE_NAME`, `KUKEON_CELL_NAME`, `KUKEON_CONTAINER_ID`, `KUKEON_REALM`, `KUKEON_SPACE`, `KUKEON_STACK`, and `KUKEON_CGROUP_PATH` from the container's existing cell-context fields. `BuildContainerSpec` and `BuildRootContainerSpec` route their `oci.WithEnv` call through the merge so user-supplied `containerSpec.Env` entries override defaults on key collisions and duplicate entries never reach `replaceOrAppendEnvValues`.
- Plumbs the profile name through a new runtime-only `ContainerSpec.CellProfileName` field (mirrors `CellCgroupPath`'s lifecycle — populated by the runner, not persisted) sourced from `cell.Metadata.Labels[cellprofile.LabelProfile]`. Cells materialized from a `CellProfile` get the var; cells created from a plain `CellDoc` skip it.
- Stamps the field at every BuildContainerSpec/BuildRootContainerSpec call site that already injects `CellCgroupPath` or etc-file paths: `provision.go` (CreateCell + ensureCellContainers root recreate), `start.go` (StartCell root + non-root recreates).

## Why

Profiles today hardcode `AGENT_NAME=<cell name>` in the env block, which drifts from `metadata.name` and silently mismatches over time. All seven values are already known to `kukeond` at container-create time — this is an env merge, not new plumbing.

## Test plan

- [x] `make test` — full unit suite
- [x] `pre-commit run --all-files` — lint/format
- [x] `make dev-init` — daemon-parity check passes; both `kuke get realms` and `kuke get realms --no-daemon` show identical output (the bind-mount regression guard from CLAUDE.md)
- [x] Live verification: `sudo ctr -n kuke-system.kukeon.io container info kukeon_kukeon_kukeond_kukeond` confirms the kukeond container's OCI Process.Env carries `KUKEON_CELL_NAME=kukeond`, `KUKEON_REALM=kuke-system`, `KUKEON_SPACE=kukeon`, `KUKEON_STACK=kukeon`, `KUKEON_CONTAINER_ID=kukeond`, `KUKEON_CGROUP_PATH=/kukeon/kuke-system/kukeon/kukeon/kukeond`. `KUKEON_CELL_PROFILE_NAME` is correctly absent because kukeond is not materialized from a CellProfile.
- [x] New unit tests in `internal/ctr/spec_kukeon_env_test.go` cover: full population, empty-field skipping, user-env-wins-on-collision, no-duplicate-entries, root-builder parity. New tests in `internal/controller/runner/cell_profile_stamp_test.go` cover the stamp helper (label present, label absent, nil-safety, preset-not-overwritten, per-spec variant).

Closes #351

### DIFF
--- a/internal/controller/runner/cell_profile_stamp.go
+++ b/internal/controller/runner/cell_profile_stamp.go
@@ -1,0 +1,64 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package runner
+
+import (
+	"github.com/eminwux/kukeon/internal/cellprofile"
+	intmodel "github.com/eminwux/kukeon/internal/modelhub"
+)
+
+// stampCellProfileNameOnContainers fills CellProfileName on every container
+// in the cell from cell.Metadata.Labels[cellprofile.LabelProfile] (the label
+// Materialize stamps when a cell is generated from a CellProfile). No-op when
+// the label is absent — cells created from a plain CellDoc carry no profile
+// identity. Mirrors the runtime-only stamping shape of
+// stampCellEtcFilePathsOnContainers / CellCgroupPath: BuildContainerSpec /
+// BuildRootContainerSpec read the field, but it is not part of the persisted
+// container document. Issue #351.
+func stampCellProfileNameOnContainers(cell *intmodel.Cell) {
+	if cell == nil {
+		return
+	}
+	name := cellProfileNameFromCell(cell)
+	if name == "" {
+		return
+	}
+	for i := range cell.Spec.Containers {
+		if cell.Spec.Containers[i].CellProfileName == "" {
+			cell.Spec.Containers[i].CellProfileName = name
+		}
+	}
+}
+
+// stampCellProfileNameOnContainerSpec is the per-container counterpart for
+// call sites that hold a local container spec value (e.g. the root spec
+// built fresh by ensureCellRootContainerSpec on the StartCell recreate path)
+// and need it to carry the same CellProfileName the cell-wide stamp would
+// apply. No-op when the cell carries no profile label.
+func stampCellProfileNameOnContainerSpec(spec *intmodel.ContainerSpec, cell *intmodel.Cell) {
+	if spec == nil || spec.CellProfileName != "" {
+		return
+	}
+	spec.CellProfileName = cellProfileNameFromCell(cell)
+}
+
+func cellProfileNameFromCell(cell *intmodel.Cell) string {
+	if cell == nil || cell.Metadata.Labels == nil {
+		return ""
+	}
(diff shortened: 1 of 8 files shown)


eminwux:
**PR #353 Review — `feat: expose KUKEON_* identity env vars to container processes`**

Clean implementation. The env-merge in `kukeonContainerEnv` is correct (defaults filtered against user-key set, then user entries appended in declared order, no duplicates reach `oci.WithEnv`), runtime-only stamping mirrors the existing `CellCgroupPath` / etc-files pattern at every cell-wide call site, and the live `ctr container info` evidence in the test plan plus the new unit tests pin the behavior end to end. Commit signature verified, no sensitive data on the env surface.

One non-blocking observation:

- **Missed runtime-only injection site** `internal/controller/runner/start.go:883`: `StartContainer` (single-container restart, reached by `kuke start container`) manually re-injects `CellCgroupPath` and `NestedCgroupRuntime` before recreating the container at line 901, but does not run `stampCellProfileNameOnContainerSpec` (or the existing `stampEtcFilePathsOnContainerSpec`). A workload recreated through this path loses `KUKEON_CELL_PROFILE_NAME` (and was already losing the `/etc/hosts` / `/etc/hostname` bind-mounts). Not blocking because it matches the pre-existing gap at the same site rather than introducing a new regression — worth a follow-up that fills both stamps in so the runtime-only injection set is consistent across every `CreateContainerFromSpec` path.

LGTM.

Issue #356, opened 2026-05-08
docs: document kuke uninstall half-cleaned-host gate in CLAUDE.md smoke-test section


eminwux:
## Why

Issue #287's AC asked the gating behavior added to `kuke uninstall` (skip filesystem + user/group teardown when at least one realm fails to drop its containerd namespace) to be documented in two places:

1. the `kuke uninstall` help text — done in PR for #287.
2. **the existing dev smoke-test / teardown docs** — deferred to this issue because dev does not edit project docs directly from a code-fix PR.

After the user-directed override on PR #517, the target doc is `docs/cli-use-cases.md` (not `kukeon/CLAUDE.md`). Rationale: `CLAUDE.md`'s preamble already names `docs/cli-use-cases.md` as the source of truth for "cell teardown verbs" / UX expectations the smoke test does not cover; the `Full per-host teardown` section there (lines 56–72 at the time of override) already documents `kuke uninstall` and the `NamespaceRemoved=true` gate as an invariant. The recovery procedure is the natural follow-up there. `CLAUDE.md` is loaded into every Claude session context, so keeping it lean (pointer-only) and routing operators to the workflow doc is the lower-overhead split.

## Proposed wording

Append the following subsection to the existing **Full per-host teardown** section in `docs/cli-use-cases.md`, immediately after the `Invariants` bullets:

```markdown
**Recovering from a failed `kuke uninstall`.**

`kuke uninstall` tears down kukeon runtime state in this order: (0) stop the kukeond daemon, (1) purge every realm, (2) remove `/run/kukeon`, (3) remove the configured run path (default `/opt/kukeon`), (4) remove the kukeon system user and group.

Steps 2–4 are gated on every realm reporting `NamespaceRemoved=true`. If any realm fails to drop its containerd namespace, the report renders each filesystem/account row as `skipped (realm purge failed)` and prints

    filesystem + user/group cleanup skipped: residual containerd namespace prevented teardown

This is the half-cleaned-host guard added in #287: tearing out `/opt/kukeon` while overlay snapshots in a residual namespace are still pinning files on disk would strand the next `kuke init` with stale containerd state and could rip the bind mounts out from under a still-live daemon.

To recover:

1. Inspect the residual namespace — e.g. `ctr -n <namespace> snapshots ls`, `ctr -n <namespace> containers ls` — and clean it up by hand or via `ctr namespace remove <namespace>` once the namespace is empty.
[...]

PR #357, opened 2026-05-08
fix: uninstall daemon-stop reads PID file from runPath; gate cleanup on namespace removal

### LINKED ISSUE #287: uninstall: daemon-stop step looks at wrong PID file path; #195 fix is a no-op
## Repro

On a host where `kuke init --kukeond-image …:dev` produced a running `kukeond` cell, run `kuke uninstall` once. First run fails on snapshots; second run succeeds.

```
root@vmi2904685:~# kuke uninstall
[…] yes
kuke uninstall:
  - kukeond: absent
  - realm "default" (namespace "default.kukeon.io"): FAILED: failed to delete realm: failed to delete realm: failed to delete containerd namespace: failed to delete namespace: namespace "default.kukeon.io" must be empty, but it still has snapshots on "overlayfs" snapshotter: failed precondition; runner purge also failed: failed to delete containerd namespace "default.kukeon.io": failed to delete namespace: namespace "default.kukeon.io" must be empty, but it still has snapshots on "overlayfs" snapshotter: failed precondition
  - realm "kuke-system" (namespace "kuke-system.kukeon.io"): FAILED: failed to delete realm: […same shape — "still has snapshots on \"overlayfs\" snapshotter"…]
  - /run/kukeon: removed
  - /opt/kukeon: removed
  - user "kukeon": removed
  - group "kukeon": removed
Error: purge realm "default": […]

root@vmi2904685:~# kuke uninstall
[…] yes
kuke uninstall:
  - kukeond: absent
  - /run/kukeon: absent
  - /opt/kukeon: absent
  - user "kukeon": absent
  - group "kukeon": absent
```

The first run leaves the host half-cleaned: containerd namespaces (and their snapshots) survive, while `/run/kukeon`, `/opt/kukeon`, and the kukeon user/group are gone. The second run only succeeds because the first run's filesystem teardown crashed the live daemon out from under itself, freeing the overlay mounts.

## Root cause: PID-file path mismatch

The daemon-stop step added by #195 (PR for issue \"uninstall: stop kukeond before purging kuke-system\") is silently a no-op because uninstall looks for the PID file at the wrong path.

- `kukeond` writes its PID file at `filepath.Join(runPath, \"kukeond.pid\")` → `/opt/kukeon/kukeond.pid`. See `cmd/kukeond/serve.go:89`.
- `controller.Uninstall` defaults `pidFile` to `filepath.Join(opts.SocketDir, \"kukeond.pid\")` → `/run/kukeon/kukeond.pid`. See `internal/controller/uninstall.go:158-163`.

Those paths never match. The PID file at `/run/kukeon/kukeond.pid` does not exist, so `stopDaemonByPIDFile` returns immediately with `PIDFilePresent=false` (`internal/controller/daemon_stop.go:74-81`). The CLI prints `kukeond: absent` (`cmd/kuke/uninstall/uninstall.go:189-190`) even when the daemon is alive and serving on `/run/kukeon/kukeond.sock`.

With the live daemon untouched, `PurgeRealm` then races containerd resources the daemon's session is pinning. Overlay snapshot mounts stay busy through `cleanupSnapshotsFor`'s 32-pass walk (`internal/ctr/namespaces.go:333-361`); every `Remove` returns EBUSY-ish, the pass makes no progress, the drain gives up, `DeleteNamespace` fails with `must be empty, but it still has snapshots on \"overlayfs\" snapshotter`. That is exactly the failure the #195 daemon-stop step was supposed to prevent.

The same path mismatch exists in `kuke daemon reset` at `cmd/kuke/daemon/reset/reset.go:148-158` — it also unlinks `<socketDir>/kukeond.pid` rather than `<runPath>/kukeond.pid`.

The unit tests miss it because they hard-code the same wrong path: `internal/controller/uninstall_test.go:276` joins `tmpSocketDir + \"kukeond.pid\"`, so the stub stopper reads from the same place uninstall writes its expectation. There is no test that pins the production default against `cmd/kukeond/serve.go`'s actual write path.

## Secondary issue: half-cleaned host on failure

Even setting the PID-file fix aside, when realm purge fails the uninstall flow proceeds to remove `/run/kukeon`, `/opt/kukeon`, and the system user/group anyway (see `internal/controller/uninstall.go:220-254`). That leaves containerd holding orphaned snapshots in namespaces it cannot delete, while the on-disk paths and the user account are gone. The next `kuke init` then has to coexist with stale containerd state. Worth gating filesystem teardown on `NamespaceRemoved` per realm, or at minimum flagging the half-cleaned state more loudly than a tail `Error:` line.

## Suspected fix

1. In `controller.UninstallOptions`, derive the PID-file default from the run path, not the socket directory. Either add a `RunPath` field on `UninstallOptions` (the controller already has `b.opts.RunPath` from `Options.RunPath`) and use `filepath.Join(b.opts.RunPath, \"kukeond.pid\")` as the default in `Uninstall`, or have the CLI explicitly populate `KukeondPIDFile` from `runPath`.
2. Apply the matching fix to `kuke daemon reset` (`cmd/kuke/daemon/reset/reset.go:148-158`) — same mismatch, same effect on dev-loop teardown.
3. Update `internal/controller/uninstall_test.go:276` and any sibling tests so the PID file is created under the run path the controller is configured with, not under the socket directory. Add a regression test that pins `runPath/kukeond.pid` (matching `cmd/kukeond/serve.go:89`) as the default the production stopper reads.
4. Decide on the half-cleaned-on-failure behavior (gate `/run/kukeon` + `/opt/kukeon` + user/group removal on `NamespaceRemoved`, or surface the half-state more prominently in the report). I'd suggest gating: a host where realm purge failed should not have `/opt/kukeon` torn out from under the live daemon.

## Acceptance criteria

- [ ] Live repro: on a host where `kuke init --kukeond-image …:dev` produced a running `kukeond` cell, a single `kuke uninstall --yes` succeeds — both realms render `purged`, the report's `kukeond:` line shows `stopped (pid …)` (not `absent`), and `ctr ns ls` shows neither `default.kukeon.io` nor `kuke-system.kukeon.io`.
- [ ] `controller.Uninstall`'s PID-file default resolves to `<runPath>/kukeond.pid`, matching `cmd/kukeond/serve.go:89`.
- [ ] `kuke daemon reset` removes the PID file at `<runPath>/kukeond.pid` (not `<socketDir>/kukeond.pid`). Verified by an updated unit test.
- [ ] New regression test pins the controller's default PID-file path against the production write path so a future rename of either side fails the test.
- [ ] When realm purge fails, the half-cleaned-host outcome is addressed: either the filesystem/user-group steps are gated on `NamespaceRemoved`, or the failure mode is surfaced more loudly than a trailing `Error:` line. Document the chosen behavior in the uninstall help text and the existing dev smoke-test docs.
- [ ] Unit test in `internal/controller/uninstall_test.go` exercises the daemon-stop step with the PID file at the production default path; the existing `kukeond.pid in tmpSocketDir` setup is replaced (or supplemented) so the test cannot pass against the broken path.

## Notes

- Follow-up to #195. The acceptance-criteria checkbox \"If `/run/kukeon/kukeond.pid` exists at the start of `Uninstall`, the daemon process is signalled\" hard-coded the wrong path; the PR landed against that wrong path and the unit tests pinned it in place.
- Reproduces on the smoke-test path documented in `kukeon/CLAUDE.md` (`make dev-init` round trip → `kuke uninstall`).

eminwux:
## Summary
- `controller.Uninstall` now defaults `KukeondPIDFile` to `<runPath>/kukeond.pid` (matching `cmd/kukeond/serve.go:99`); the prior `<socketDir>/kukeond.pid` default made the daemon-stop step from #195 a silent no-op so realm purge raced the live daemon and failed with `namespace must be empty, but it still has snapshots on overlayfs`.
- `kuke daemon reset` now removes `kukeond.pid` from runPath (was: socketDir) so the dev-loop teardown matches the production write path.
- Added a half-cleaned-host gate: when at least one realm reports `NamespaceRemoved=false`, steps 2–4 (remove `/run/kukeon`, remove run path, drop user/group) are skipped and `report.CleanupSkipped=true`. The CLI renders each gated row as `skipped (realm purge failed)` and prints a one-line explainer so the failure mode is visible without scrolling to the trailing error.

## Test plan
- [x] `make test` — full project unit + table-test suite (~75 packages)
- [x] `pre-commit run --files <changed>` — gofmt + golangci-lint + addlicense clean
- [x] `go test ./internal/controller/ -run TestUninstall -v` — 8 tests pass, including the new `TestUninstall_DefaultPIDFileResolvesToRunPath` regression and the renamed `TestUninstall_PurgeFailureGatesFilesystemCleanup`
- [x] `go test ./cmd/kuke/uninstall/... ./cmd/kuke/daemon/reset/... -v` — uninstall + reset suites pass; new `TestUninstall_CleanupSkippedRendersSkipMarker` covers the CLI rendering
- [x] `make kuke && ./kuke uninstall --help` — long description carries the new gating paragraph
- [x] Manual repro confirmation: `ls /opt/kukeon/` on this host shows `kukeond.pid` (matching the issue's described symptom — pre-fix `kuke uninstall` reads `/run/kukeon/kukeond.pid` and finds nothing)
- [ ] `make dev-init` end-to-end smoke — **deferred to reviewer's clean host.** This dev host has an active kukeond cell with workloads in `/opt/kukeon/{default,kuke-system}` that the smoke playbook would tear down; running it here would disrupt unrelated state. The `make test` target above is the same suite CI runs.

## Notes
- A docs follow-up to add a `Recovering from a failed kuke uninstall` sub-section to `kukeon/CLAUDE.md` is filed as #356 (verbatim wording included), per the role rule against editing project `CLAUDE.md` from a code-fix PR.
- The renamed test (`TestUninstall_PurgeFailureIsRecordedButCleanupContinues` → `TestUninstall_PurgeFailureGatesFilesystemCleanup`) reflects the new behavior: cleanup no longer continues — it gates. The old name pinned a broken behavior.

Closes #287

### DIFF
--- a/cmd/kuke/daemon/reset/reset.go
+++ b/cmd/kuke/daemon/reset/reset.go
@@ -147,8 +147,17 @@ func runReset(cmd *cobra.Command, _ []string) error {
 	)
 
 	socketDir := resolveSocketDir(cmd)
-	for _, name := range []string{"kukeond.sock", "kukeond.pid"} {
-		path := filepath.Join(socketDir, name)
+	runPath := resolveRunPath(cmd)
+	// kukeond.sock lives under the socket dir (/run/kukeon); kukeond.pid is
+	// written under the run path (/opt/kukeon) by cmd/kukeond/serve.go. The
+	// two diverge — pinning kukeond.pid to the socket dir was the silent #287
+	// no-op that left the daemon-stop step in `kuke uninstall` looking at the
+	// wrong path.
+	transientFiles := []string{
+		filepath.Join(socketDir, "kukeond.sock"),
+		filepath.Join(runPath, "kukeond.pid"),
+	}
+	for _, path := range transientFiles {
 		removed, removeErr := removeFileIfExists(path)
 		if removeErr != nil {
 			return fmt.Errorf("remove %q: %w", path, removeErr)
@@ -159,7 +168,6 @@ func runReset(cmd *cobra.Command, _ []string) error {
 	}
 
 	if purgeSystem {
-		runPath := resolveRunPath(cmd)
 		systemDir := filepath.Join(runPath, consts.KukeSystemRealmName)
 		removed, removeErr := removeDirIfExists(systemDir)
 		if removeErr != nil {
--- a/cmd/kuke/daemon/reset/reset_test.go
+++ b/cmd/kuke/daemon/reset/reset_test.go
@@ -251,14 +251,20 @@ func TestDaemonReset(t *testing.T) {
 }
 
 // TestDaemonReset_RemovesSocketAndPidFiles confirms the cleanup step actually
-// removes /run/kukeon/kukeond.{sock,pid} (or the per-test override) when both
-// files are present.
+// removes /run/kukeon/kukeond.sock and /opt/kukeon/kukeond.pid (or the
+// per-test overrides) when both files are present.
+//
+// kukeond.sock lives under the socket dir (KUKEOND_SOCKET's parent) while
+// kukeond.pid is written under the run path by cmd/kukeond/serve.go — pinning
+// the pid removal to the socket dir was the silent #287 no-op that left the
+// daemon-stop step in `kuke uninstall` looking at the wrong path.
 func TestDaemonReset_RemovesSocketAndPidFiles(t *testing.T) {
 	withFreshViper(t)
 
 	socketDir := t.TempDir()
+	runPath := t.TempDir()
 	sockPath := filepath.Join(socketDir, "kukeond.sock")
-	pidPath := filepath.Join(socketDir, "kukeond.pid")
+	pidPath := filepath.Join(runPath, "kukeond.pid")
 	if err := os.WriteFile(sockPath, []byte{}, 0o600); err != nil {
 		t.Fatalf("seed sock file: %v", err)
 	}
@@ -288,7 +294,7 @@ func TestDaemonReset_RemovesSocketAndPidFiles(t *testing.T) {
 	ctx := context.WithValue(context.Background(), types.CtxLogger, logger)
 	ctx = context.WithValue(ctx, reset.MockClientKey{}, kukeonv1.Client(fake))
 	ctx = context.WithValue(ctx, reset.MockSocketDirKey{}, socketDir)
-	ctx = context.WithValue(ctx, reset.MockRunPathKey{}, t.TempDir())
+	ctx = context.WithValue(ctx, reset.MockRunPathKey{}, runPath)
 	cmd.SetContext(ctx)
 	cmd.SetArgs(nil)
 
--- a/cmd/kuke/uninstall/uninstall.go
+++ b/cmd/kuke/uninstall/uninstall.go
@@ -63,6 +63,14 @@ This is the global counterpart to "kuke purge" (which is per-resource). It:
   3. Removes the configured run path (default /opt/kukeon) recursively.
   4. Removes the kukeon system user and group (no-op if absent).
 
+If any realm fails to drop its containerd namespace, steps 2-4 are skipped:
+tearing out /opt/kukeon while a residual namespace is still pinning overlay
+mounts on disk would strand the next "kuke init" with stale containerd state.
+The report flags every dir/account row as "skipped (realm purge failed)" so
+the half-cleaned host is visible without scrolling to the trailing error.
+Resolve the realm-purge failure (often: stop the live daemon, remove the
+residual containerd namespace by hand) and re-run the command.
+
 The /usr/local/bin/kuke binary and the kukeond symlink are NOT removed —
 uninstalling runtime state is not the same as uninstalling the binary.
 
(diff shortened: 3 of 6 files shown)


eminwux:
**PR #357 Review — `fix: uninstall daemon-stop reads PID file from runPath; gate cleanup on namespace removal`**

Tight, well-targeted fix for the silent no-op in #287. The default PID-file path now matches `cmd/kukeond/serve.go:99`'s write path, the half-cleaned-host gate prevents `/opt/kukeon` teardown while a residual containerd namespace still pins overlay mounts, and `kuke daemon reset` removes `kukeond.pid` from the right place. `TestUninstall_DefaultPIDFileResolvesToRunPath` pins the controller↔daemon path agreement directly. Verified single commit is signed, no stale `SocketDir + kukeond.pid` joins remain anywhere in the tree, `b.opts.RunPath` is wired through `cmd/kuke/uninstall/uninstall.go:131` so the production default actually resolves.

No blocking items. One non-blocking note:

- **Test plan** `make dev-init`: the end-to-end smoke is the project-CLAUDE.md regression guard for changes touching `/opt/kukeon` and the daemon read paths, and it's left unchecked here. The targeted unit test pins the symptom directly so the unit coverage is sufficient evidence the fix lands, but consider running `make dev-init` on a clean host before merge per the project's smoke-test convention — the deferral rationale (active workloads on the dev host) is reasonable, but the regression-guard tail is what catches bind-mount divergence between the daemon and the in-process controller.

LGTM — ready to merge once you've confirmed the smoke check on a clean host (or accepted the unit-test-only evidence).

PR #358, opened 2026-05-08
feat: parameterized cell profiles via --param/--name/--param-file

### LINKED ISSUE #355: kuke run -p: parameterized profiles via --param and --name
## Motivation

Profiles in `~/.kuke/profiles.d/` are static today: `kuke run -p <name>` instantiates exactly the spec on disk. This is fine for the human-driven case ("give me a fresh dev cell"), but breaks down for any caller that needs to vary a small number of fields per invocation — most concretely, **the orchestrator agent** dispatching role-specific worker cells.

Today the orchestrator works around this by keeping templated `*.yaml.tmpl` files in a separate repo, running `envsubst` into `/tmp/<cell>.yaml`, and applying the rendered file. The workaround has three problems:

1. **Templates are not first-class.** They aren't valid YAML, can't be linted by `kuke`, and live separately from the static profiles in `~/.kuke/profiles.d/`. Two parallel artifacts encode the same shape.
2. **`kuke apply -f` doesn't start a cell**, and `kuke run -f` requires a `Cell` resource — but profiles are authored as `CellProfile`. The dispatch path either re-implements the profile→cell mapping client-side or loses the abstraction.
3. **The cell name is a substitution variable**, so each dispatch produces a unique on-disk profile rather than reusing the static one. Concept drift between "the profile I run for fun" and "the profile orch dispatches" is now possible.

## Proposal

Extend `kuke run -p <profile>` with three flags:

```
kuke run -p <profile>
  --name <override>          # cell name (default: spec.metadata.name + random suffix)
  --param KEY=VALUE          # repeatable; sets a parameter for substitution
  --param-file <file>        # optional: load KEY=VALUE lines from file
```

Substitution happens server-side after the profile loads, before validation. `${KEY}` references inside the profile spec are resolved in this order:

1. `--param KEY=VALUE` from the CLI.
2. `parameters[].default` declared in the profile (see below).
3. Process env (the kukeond process, not the caller — keeps the substitution surface narrow and predictable).
4. **Error** if still unset and the parameter is declared `required: true`. Unreferenced params are ignored.

### Profile schema addition

Profiles declare their parameters explicitly so callers get a useful error at load time rather than a runtime mystery:

```yaml
apiVersion: v1beta1
kind: CellProfile
metadata:
  name: dev
spec:
  parameters:
    - name: PROMPT
      description: "Slash command passed to claude --print"
      required: true
    - name: PROJECT_REPO
      description: "Git URL of the target project"
      required: true
    - name: PROJECT_DIR
      description: "Short dirname for the project clone"
      required: true
    - name: AGENTS_REPO
      description: "owner/repo of the agents repo"
      default: "eminwux/agents"
    - name: CLAUDE_IMAGE
      description: "Container image tag for the worker"
      default: "registry.eminwux.com/claude:latest"
  cell:
    containers:
      - id: work
        image: ${CLAUDE_IMAGE}
        env:
          - AGENTS_REPO=${AGENTS_REPO}
          - PROJECT_REPO=${PROJECT_REPO}
          ...
```

Defaults make the common path ergonomic — most callers only override `PROMPT`, `PROJECT_REPO`, `PROJECT_DIR`. Required-without-default params surface as a clear "missing required parameter PROMPT" error from `kuke run`.

### Validation rules

- Profile fails to load if it references `${KEY}` for a `KEY` not in `parameters[]`. (Catches typos at install time.)
- `kuke run -p <name> --param KEY=...` fails if `KEY` is not declared in the profile. (Catches typos at call time.)
- Empty string is a valid value; only "unset" triggers the default/required logic.

### CLI examples

```bash
# Static use, all defaults — unchanged from today's UX:
kuke run -p dev

# Orchestrator dispatch:
kuke run -p dev --name crew-dev-354 --rm -d \
  --param PROMPT='/pick-issue 354' \
  --param PROJECT_REPO=https://github.com/eminwux/crew \
  --param PROJECT_DIR=crew

# Same, with a param file (orch could write one to /tmp instead of long CLI):
kuke run -p dev --name crew-dev-354 --rm -d --param-file /tmp/crew-dev-354.params
```

### Out of scope (intentionally)

- **Templating beyond `${VAR}`.** No `{{ if }}`, no Go template engine. If a profile needs branching, fork it.
- **Substitution in keys.** Only values are substituted; YAML structure is fixed by the profile.
- **Param substitution in `kuke apply -f`.** `apply` stores the profile as-is; substitution is a `run`-time concern.

## Why this over alternatives

- **`kuke run -f <Cell>`-on-stdin from `envsubst`** (current workaround): keeps templates as a second-class artifact in the caller's repo. Doesn't address concept drift between "static profile" and "dispatched cell."
- **Make `kuke run -f` accept a `CellProfile`**: smaller change but doesn't solve parameterization — caller still has to render upstream.
- **A separate `kuke dispatch` command**: redundant with `run`. The dispatch use case is just "run this profile with these overrides."

## Migration

No breaking change. Existing profiles without `parameters[]` keep working — substitution is a no-op. Orchestrator switches its dispatch skill from `envsubst | kuke apply -f` to `kuke run -p <role> --name ... --param ...` and deletes the `*.yaml.tmpl` files.

eminwux:
## Summary

- Adds `spec.parameters[]` to `CellProfile` (name + description + optional default + required), so a profile can declare its variable inputs alongside its body.
- Wires three new flags on `kuke run -p`: `--name <override>` (replaces the auto-generated `<prefix>-<6hex>` cell name), `--param KEY=VALUE` (repeatable), and `--param-file <file>` (KEY=VALUE per line, `#` comments). All three are rejected when combined with `-f`.
- Implements `${KEY}` substitution over the loaded YAML node tree. Resolution order per key: CLI `--param` > `--param-file` > parameter's `default` > process env (`os.LookupEnv` on the kuke binary's env) > error if `required: true`. Non-required, referenced, unset → empty string (envsubst-compatible). Substitution operates on scalar *values* only — keys are never touched, and `spec.parameters[]` defaults are exempt from the undeclared-ref scan so `default: \"\${VAR}\"` is a literal default, not a recursive reference.
- Validation: at load, undeclared `${KEY}` refs in the body fail with `ErrProfileInvalid`; duplicate / invalid parameter names fail. At call, undeclared `--param` keys fail.
- Backward-compatible: profiles without `spec.parameters` and without `${...}` refs go through the original `Load`/`Materialize` path unchanged.

### Non-obvious calls (please push back if wrong)

- The issue spec says "process env (the kukeond process, not the caller)", but profiles are a client-side artifact loaded by the `kuke` CLI before anything reaches the daemon. Interpreted as the `kuke` binary's process env (`os.LookupEnv`); the substituted body is what gets sent to kukeond.
- Non-required + unset substitutes to empty string rather than leaving the literal `\${KEY}`, matching shell `envsubst`. Required + unset still errors.
- `--name` / `--param` / `--param-file` are rejected with `-f` (file mode is not a profile and has no parameter declarations).
- `--param` wins over `--param-file` (later-binding) when both supply the same key.

## Test plan

- [x] `make test` (full suite green; new tests in `internal/cellprofile/params_test.go` and `cmd/kuke/run/run_test.go`)
- [x] `pre-commit run --files <changed>` (go fmt / golangci-lint / addlicense / go-mod-tidy clean)
- [x] `go vet ./...` and `go build ./...` clean
- [x] `make dev-init` end-to-end smoke — ran on the PR branch; daemon parity check identical for `kuke get realms` and `kuke get realms --no-daemon`.

Closes #355

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -375,6 +375,10 @@ var (
 	KUKE_RUN_PROFILE = DefineKV("KUKE_RUN_PROFILE", "kuke/run/profile")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_RUN_RM = DefineKV("KUKE_RUN_RM", "kuke/run/rm")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_RUN_NAME = DefineKV("KUKE_RUN_NAME", "kuke/run/name")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_RUN_PARAM_FILE = DefineKV("KUKE_RUN_PARAM_FILE", "kuke/run/param-file")
 
 	// Attach command variables
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -29,6 +29,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 
 	"github.com/eminwux/kukeon/cmd/config"
@@ -71,6 +72,22 @@ func NewRunCmd() *cobra.Command {
 	cmd.Flags().StringP("profile", "p", "",
 		"Cell profile name to load from $HOME/.kuke/profiles.d (or $KUKE_PROFILES_DIR); mutually exclusive with -f")
 	_ = viper.BindPFlag(config.KUKE_RUN_PROFILE.ViperKey, cmd.Flags().Lookup("profile"))
+
+	cmd.Flags().String("name", "",
+		"Override the materialized cell name (default: <metadata.name>-<6hex>). "+
+			"Only valid with -p; rejected with -f, where metadata.name is the cell name verbatim.")
+	_ = viper.BindPFlag(config.KUKE_RUN_NAME.ViperKey, cmd.Flags().Lookup("name"))
+
+	cmd.Flags().StringArray("param", nil,
+		"Profile parameter override as KEY=VALUE; repeatable. Only valid with -p. "+
+			"Each KEY must be declared in spec.parameters[]. Wins over the parameter's "+
+			"default and over --param-file when both set the same key.")
+
+	cmd.Flags().String("param-file", "",
+		"File of KEY=VALUE lines whose values seed profile parameters; one per line, "+
+			"`#` starts a comment. Same declaration rules as --param. CLI --param wins on "+
+			"duplicate keys.")
+	_ = viper.BindPFlag(config.KUKE_RUN_PARAM_FILE.ViperKey, cmd.Flags().Lookup("param-file"))
 
 	cmd.MarkFlagsMutuallyExclusive("file", "profile")
 	cmd.MarkFlagsOneRequired("file", "profile")
@@ -126,6 +143,9 @@ type runFlags struct {
 	detach        bool
 	containerFlag string
 	autoDelete    bool
+	nameOverride  string
+	paramArgs     []string
+	paramFile     string
 }
 
 func parseRunFlags(cmd *cobra.Command, _ []string) (runFlags, error) {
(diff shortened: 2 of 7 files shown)


eminwux:
**PR #358 Review — `feat: parameterized cell profiles via --param/--name/--param-file`**

LGTM — coherent feature, well-tested (782 lines of tests for ~612 lines of non-test code). Substituting on the parsed `yaml.Node` tree (rather than re-marshaling) is the right call — it preserves quoting and comments, and the tag-clearing in `substituteScalars` correctly handles the type-narrowing case for `${PORT}-foo`. Validation is layered correctly: undeclared `${KEY}` refs caught at load, undeclared `--param` keys caught at call, parameters block exempted from the body scan so `default: "${HOME}/bin"` proxy patterns work. Per-commit signature verified. No sensitive-data leaks in error messages.

- **CodeQL** the 4 high-severity `go/allocation-size-overflow` alerts on this PR are all `make(map/slice, len(x)+const)` patterns in code this PR does **not** modify (`internal/ctr/spec.go:414`, `internal/ctr/container_utils.go:215`, and `mergeLabels` at `internal/cellprofile/cellprofile.go:287` — pre-existing helper, untouched here). Baseline-shift false positives; safe to dismiss in the security UI as not-introduced-by-this-PR.
- **Stealth mutation** `internal/cellprofile/params.go:313-320`: on `yaml.Marshal` failure, `cloneNode` returns the input pointer, so `substituteScalars` would silently mutate the source tree. The comment acknowledges this and Marshal failure on an already-parsed `yaml.Node` would be very unusual; flagging only because the function's contract elsewhere assumes a deep copy.
- **Error echoes value** `internal/cellprofile/params.go:338-341`: `ParseParamArgs` echoes the full `KEY=VALUE` arg in the "invalid KEY" error. If a user types `--param 1BAD=secret-token`, the secret lands in stderr. Consistent with other tools but worth knowing.

LGTM unconditional — clearing for merge.

PR #367, opened 2026-05-10
fix: auto-derive RootContainerID from container Root flag in apischeme

### LINKED ISSUE #349: runner: ensureCellRootContainerSpec silently drops fields on a Root:true container when rootContainerId is empty
## Repro

Reproduced on `main` (5da1c6f) while running the AC item 7 smoke for #347. Apply a Cell whose YAML places \`root: true\` on a container in \`spec.containers[]\` but **does not** set \`spec.rootContainerId\`:

\`\`\`yaml
apiVersion: v1beta1
kind: Cell
metadata:
  name: tmpfs-smoke
spec:
  id: tmpfs-smoke
  realmId: default
  spaceId: default
  stackId: default
  # rootContainerId intentionally omitted
  containers:
    - id: root
      root: true
      command: sleep
      args: ["60"]
      volumes:
        - kind: tmpfs
          target: /var/lib/cache
          sizeBytes: 16777216
        - source: /tmp
          target: /host-tmp
          readOnly: true
\`\`\`

\`sudo ./kuke apply -f tmpfs-smoke.yaml\` reports \`Cell "tmpfs-smoke": created\` and the cell becomes Ready, but the user-declared volumes never reach the OCI spec.

## Observed

\`\`\`bash
sudo cat /opt/kukeon/default/default/default/tmpfs-smoke/metadata.json | jq '.spec.containers[0].volumes'
# → null
\`\`\`

\`ctr -n default.kukeon.io container info default_default_tmpfs-smoke_root\` shows only the OCI default mounts (/proc, /dev, /dev/shm, /run, /sys) — no \`/var/lib/cache\` tmpfs and no \`/host-tmp\` bind. Setting \`rootContainerId: root\` on the same Cell makes the volumes survive end-to-end (verified in the same session via a worker container — fields reach \`internal/ctr/spec.go:buildVolumeMounts\` correctly).

The drop is silent: no warning in daemon logs, no validation error, the user just gets a cell missing its declared mounts.

## Root cause

\`internal/controller/runner/provision.go:2049\` (\`ensureCellRootContainerSpec\`) only consults \`cell.Spec.Containers[]\` for the root spec when \`cell.Spec.RootContainerID != \"\"\` (line 2088). When \`RootContainerID\` is empty, it unconditionally builds a fresh \`ctr.DefaultRootContainerSpec(...)\` (line 2110) and ignores any container in the array that has \`Root: true\`. The user-supplied container is then re-added to the array via the merge at provision.go:1726–1738, but by then \`rootContainerSpec\` is already the default — only the **default-built** fields reach \`BuildRootContainerSpec\`.

This is the inverse of the case fixed in closed issue #100 (which addressed silent drops when \`RootContainerID\` *is* set but \`BuildRootContainerSpec\` only forwarded a subset of fields). The path where \`RootContainerID\` is empty but a container is flagged \`Root: true\` was never wired up.

## Suggested fix

Two viable fixes; reviewer's call which to take:

1. **In \`ensureCellRootContainerSpec\`**: if \`cell.Spec.RootContainerID == \"\"\` but exactly one container in \`cell.Spec.Containers\` has \`Root: true\`, treat that container as the root spec (same code path as the explicit-RootContainerID branch). Reject \"more than one Root:true\" with a clear error so ambiguity surfaces at apply time.

2. **At normalize/apply time** (\`internal/apischeme\` or \`ReconcileCell\`): auto-populate \`Spec.RootContainerID\` from a container with \`Root: true\` before handing the cell to the runner. Same single-Root-true validation rule.

Option 2 keeps the runner ignorant of the convention and centralizes the auto-detection at the API boundary; option 1 is a smaller diff localized to the runner. Either way, the silent drop should become impossible: a \`Root: true\` container that doesn't end up as the root must be a hard error, not a no-op.

## References

- Surfaced while smoke-testing #347 / #348.
- Closed issue #100 fixed the related bug for the explicit-RootContainerID path.

eminwux:
## Summary

- A Cell whose YAML places `root: true` on a container in `spec.containers[]` but does not set `spec.rootContainerId` was silently dropping the user's container fields (volumes, env, command, etc.). The runner's `ensureCellRootContainerSpec` only consulted the container array when `RootContainerID != ""` and otherwise unconditionally generated a fresh default root spec — the user-marked container was re-merged into the array but only the *default-built* fields fed `BuildRootContainerSpec`.
- Fix at the API normalization boundary (`apischeme.ConvertCellDocToInternal`) so every entry point — apply, bootstrap, on-disk rehydrate (`runner.read`) — produces a self-consistent cell. A new `resolveCellRootContainer` helper auto-populates `RootContainerID` from a sole `Root: true` container, rejects multiple `Root: true` containers, and rejects a `RootContainerID`/`Root: true` mismatch. Two new sentinels in `internal/errdefs` (`ErrMultipleRootContainers`, `ErrRootContainerMismatch`) match the project's shared-sentinel convention.
- Picked option (2) from the issue ("normalize at the API boundary") over option (1) ("fix the runner branch") so the persisted internal cell stays consistent for every downstream reader (`update_cell.go`, `helpers.findRootContainer`, `refresh.go`, `get.go`, `cell_etc_files.go`). Bootstrap's `kukeondCellDoc` doesn't use `Root: true`, so the auto-default-root path for the kuke-system cell is structurally unchanged.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` — full unit suite green, including new `internal/apischeme` cases:
  - auto-derive populates `RootContainerID` from a single `Root: true` container, with volumes preserved
  - empty `RootContainerID` + zero `Root: true` containers stays empty (preserves the kuke-system / default-root path)
  - two `Root: true` containers errors with wrapped `ErrMultipleRootContainers`
  - `RootContainerID` set + a *different* `Root: true` container errors with wrapped `ErrRootContainerMismatch`
  - `RootContainerID` set + matching `Root: true` accepted
  - `BuildCellExternalFromInternal` rejects a malformed internal cell with two `Root: true` containers (defensive outbound validation)
- [ ] `make dev-init` smoke — could not run on this dev host: the docker build for the local kukeond image fails inside the dev container with `mount source: "overlay" ... err: invalid argument` from buildkit's nested overlayfs cachemount. The change is purely in apischeme normalization with no impact on the build path or the daemon binary; bootstrap doesn't use `Root: true`, so the daemon-parity check is structurally unaffected.

Closes #349

### DIFF
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -19,6 +19,7 @@ package apischeme
 import (
 	"fmt"
 
+	"github.com/eminwux/kukeon/internal/errdefs"
 	intmodel "github.com/eminwux/kukeon/internal/modelhub"
 	ext "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 )
@@ -536,6 +537,62 @@ func validateContainerTty(spec ext.ContainerSpec) error {
 	return fmt.Errorf("container %q: tty fields require attachable: true", spec.ID)
 }
 
+// resolveCellRootContainer enforces the rules around CellSpec.RootContainerID
+// and ContainerSpec.Root and returns the resolved root container ID:
+//   - At most one container may carry root: true. Issue #349: a second
+//     root-flagged container would silently lose its Root intent, so it
+//     surfaces as a hard error here.
+//   - If RootContainerID is set, the named container must exist in the
+//     array, and any container marked root: true must be that same one.
+//   - If RootContainerID is empty but exactly one container has
+//     root: true, that container's ID is the resolved root. Callers
+//     building a normalized internal cell should overwrite
+//     spec.RootContainerID with the returned value so downstream code
+//     (runner.ensureCellRootContainerSpec, helpers.findRootContainer,
+//     etc.) sees a self-consistent cell.
+//   - If neither is set, returns "" — the runner builds a default root.
+func resolveCellRootContainer(spec intmodel.CellSpec) (string, error) {
+	var rootMarked []string
+	for _, c := range spec.Containers {
+		if c.Root {
+			rootMarked = append(rootMarked, c.ID)
+		}
+	}
+	if len(rootMarked) > 1 {
+		return "", fmt.Errorf("%w: %v", errdefs.ErrMultipleRootContainers, rootMarked)
+	}
+
+	if spec.RootContainerID != "" {
+		found := false
+		for _, c := range spec.Containers {
+			if c.ID == spec.RootContainerID {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return "", fmt.Errorf(
+				"rootContainerId %q not found in containers array",
+				spec.RootContainerID,
+			)
+		}
+		if len(rootMarked) == 1 && rootMarked[0] != spec.RootContainerID {
+			return "", fmt.Errorf(
+				"%w: rootContainerId is %q but container %q has root: true",
+				errdefs.ErrRootContainerMismatch,
+				spec.RootContainerID,
+				rootMarked[0],
+			)
+		}
+		return spec.RootContainerID, nil
+	}
+
+	if len(rootMarked) == 1 {
+		return rootMarked[0], nil
+	}
+	return "", nil
+}
+
 // validateCellTty enforces the AC that CellTty.Default names an existing
 // attachable container in the same cell (or is empty).
 func validateCellTty(spec ext.CellSpec) error {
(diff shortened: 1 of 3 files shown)


eminwux:
**PR #367 Review — `fix: auto-derive RootContainerID from container Root flag in apischeme`**

LGTM. Clean fix at the right boundary — every entry point (`apply.go`, `bootstrap.go`, `delete_documents.go`, `runner/read.go`, `internal/client/local/local.go`) routes through `ConvertCellDocToInternal`/`NormalizeCell`, so auto-deriving `RootContainerID` there means the runner's explicit-root branch (`provision.go:2126`) reliably finds the user's container in the array. Bootstrap's `kukeondCellDoc` doesn't carry `Root: true`, so the auto-default-root path for the kuke-system cell is preserved (covered by `TestCellRootContainerIDLeftEmptyWhenNoRootFlag`). Sentinels are wrapped with `%w` so `errors.Is` callers keep working. No leaks, no dead code, all commits verified.

- **Outbound mismatch coverage** `internal/apischeme/scheme_test.go`: `TestBuildCellExternalRejectsMultipleRootTrue` covers the multi-root case on the outbound path, but the rootContainerId/Root-flag *mismatch* case is only exercised inbound. Minor — the inbound coverage transitively guards the helper, and outbound is purely defensive.
- **Comment cosmetic** `internal/apischeme/scheme.go:920-925`: the new comment line-wraps `ConvertCellDoc-`/`ToInternal` mid-identifier (column boundary). Not wrong, just visually odd if anyone greps.

`make dev-init` is documented as unrunnable on the dev host (nested-overlay issue) — fine to defer pre-merge given the change is API-normalization only and bootstrap is structurally unaffected.

Ready to merge.

PR #368, opened 2026-05-10
fix: pre-create attachable tty capture/log files at 0640 for kukeon group

### LINKED ISSUE #366: runner: tty capture/log files created mode 0600 — unreadable by kukeon group
## Symptom

A non-root cell user that is a member of the host `kukeon` group cannot tail another cell's capture file via `kuke log`, even though the surrounding directory is group-traversable.

```
$ id
uid=1000(claude) gid=1000(claude) groups=1000(claude),988(kukeon)

$ kuke log --cell kukeon-dev-349 --container work --no-follow
Error: open /opt/kukeon/default/default/default/kukeon-dev-349/work/tty/capture: permission denied
```

## Root cause

The tty directory is created group-accessible (`2750 root:kukeon`, setgid), but the `capture` and `log` files inside it are created mode `0600`:

```
$ stat -c '%a %U:%G %n' /opt/kukeon/default/default/default/kukeon-dev-349/work/tty/*
600  root:kukeon  capture
600  root:kukeon  log
660  root:kukeon  socket
```

The `socket` is `0660` (group-rw), which is consistent with the directory's intent. The `capture` and `log` files appear to have been created with a stricter umask or an explicit `0600` and never widened — so only `root` can read them, and group membership doesn't help.

## Expected

`capture` and `log` should be created `0640` (or `0660` to match the socket), so that members of the `kukeon` group — which is the gating mechanism for `kukeond` socket access — can read peer cells' streams. That is what makes `kuke log` work for non-root callers (e.g. the orchestrator agent reading a sibling cell's output).

## Impact

- Orchestrator agent (runs as uid 1000, in `kukeon` group) cannot tail any cell it dispatched.
- Any non-root tooling on the host that wants to consume cell output is blocked the same way.

## Where to look

Wherever the per-container tty `capture` / `log` files are opened/created in the runner — likely an `os.OpenFile` with mode `0600` that should be `0640`. The directory creation already gets it right (setgid + group-readable); just the file mode needs to match.

eminwux:
## Summary

- The per-container sbsh capture and log files inside `<ttyDir>/` landed mode `0600` because sbsh creates them with that hard-coded default. The surrounding `tty/` directory is already `02750 root:kukeon` (group-traversable) and the sbsh control socket is `0660 root:kukeon` via the existing `--socket-mode` plumbing — but the capture/log siblings stayed root-only, so a non-root operator in the `kukeon` group (orchestrator agent, host tooling) got EACCES when tailing peer cells via `kuke log`.
- Fixed by pre-creating the two files in `attachablePostCreateChown` — the existing post-create hook that already chowns `ttyDir` and `profile.yaml` to the resolved container uid before the task starts. The helper opens `<ttyDir>/{capture,log}` with `O_CREATE|O_WRONLY` (no `O_TRUNC`), then `os.Chmod` to `0640` and `os.Chown` to `(uid, kukeonGID)`. When sbsh later opens the same paths inside the container with `O_CREATE|O_APPEND|O_WRONLY`, the existing inodes are reused and Linux ignores the mode argument — so the host-readable layout sticks across both fresh starts and daemon restarts.
- Skips the pre-create when `KukeonGroupGID == 0` (no kukeon group on the host), preserving the legacy `0o600` owner-only contract that sbsh already produces — the surrounding `tty/` dir is already `0o0700` root-only in that branch by `attachableTTYDirInitialPerms`, so widening the files would not buy anything.
- 0640 chosen over 0660: the kukeon group only needs read access (the captured PTY transcript is for tailing, not writing); 0660 would let the host-side group also append to the in-container transcript, which is not what the issue asks for.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` — full unit suite green, including new `internal/controller/runner` cases:
  - pre-create at 0640 when a kukeon group GID is configured
  - re-apply mode and ownership on pre-existing files (daemon-restart parity), preserving content
  - no-op when GID == 0 (legacy fallback path)
  - constant-locking guard on `attachableLogFileMode == 0o0640`
- [ ] `make dev-init` smoke — could not run on this dev host: the docker build for the local kukeond image fails inside the dev container with `mount source: "overlay" ... err: invalid argument` from buildkit's nested overlayfs cachemount. Same environmental failure documented on the dev host in PR #367. The change is purely in the runner's per-container tty pre-create logic with no impact on the build path or the daemon binary; bootstrap doesn't use Attachable/sbsh, so the daemon-parity check is structurally unaffected.

Closes #366

### DIFF
--- a/internal/controller/runner/attachable.go
+++ b/internal/controller/runner/attachable.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/ctr"
 	intmodel "github.com/eminwux/kukeon/internal/modelhub"
 	"github.com/eminwux/kukeon/internal/util/fs"
@@ -42,6 +43,18 @@ const attachableTTYDirRootMode os.FileMode = os.ModeSetgid | 0o0750
 // runPath). In that mode there is no group to delegate access to, so the
 // directory stays owner-only (0700).
 const attachableTTYDirNoGroupMode os.FileMode = 0o0700
+
+// attachableLogFileMode is the mode pre-applied to the per-container sbsh
+// capture and log files when the kukeon group GID is configured. 0640 =
+// rw for owner (the resolved container uid) + r for group (the kukeon
+// group), no world. Pre-creating the inodes at this mode before sbsh
+// runs forces sbsh's later `OpenFile(O_CREATE|...)` (which only sets the
+// mode on a fresh inode) to inherit the host-readable layout instead of
+// sbsh's hard-coded 0o600 owner-only default — that default blocks
+// non-root members of the kukeon group from tailing peer cells via
+// `kuke log`, even though the surrounding tty/ directory is already
+// group-traversable. Issue #366.
+const attachableLogFileMode os.FileMode = 0o0640
 
 // attachableTTYDirInitialPerms returns the (mode, gid) tuple to apply to a
 // freshly-prepared per-container tty directory before the workload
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #368 Review — `fix: pre-create attachable tty capture/log files at 0640 for kukeon group`**

Clean, minimal fix for #366: pre-create the capture/log inodes in `attachablePostCreateChown` at 0640 owned by (resolved-uid, kukeon-gid) so sbsh's later `OpenFile(O_CREATE|...)` inherits the host-readable mode instead of dropping back to its hard-coded 0o600 default. Pre-create runs after `CreateContainerFromSpec` but before task `Start`, so no in-container code can race the inode. The explicit `os.Chmod` after `OpenFile` defends against umask, errors are wrapped with `%w`, and the existing `errors.Is(os.ErrNotExist)` branch on `profile.yaml` is preserved. No security concerns — only operational paths/uids in error messages, no credentials. Four new tests lock mode, daemon-restart idempotency (verifying pre-existing content survives), gid=0 no-op, and the 0o0640 constant itself.

The `make dev-init` smoke is correctly documented as host-gated by the same buildkit nested-overlay env failure as PR #367, and the author notes bootstrap doesn't exercise Attachable/sbsh — non-blocking per the host-gated carve-out.

LGTM — worth running an end-to-end `kuke log` as a non-root kukeon-group operator against an Attachable cell pre-merge, since that's the exact regression scenario and no host-gated check in the test plan exercises it directly.

eminwux:
Pulling the LGTM — the approach in this PR is the wrong shape for the codebase's existing precedent.

`internal/ctr/attachable.go` already plumbs `--socket-mode 0660 --socket-gid <kukeon-gid>` to `sbsh terminal` for the sibling socket inode in the same `tty/` directory (sbsh v0.10.0 flags shipped via eminwux/sbsh#188). The capture and log files are the same problem class — same directory, same kukeon-group read requirement — and should follow the same plumbing pattern, not introduce a second mechanism.

The host-side pre-create + inode-reuse trick is fragile by construction: `OpenFile(O_CREATE|O_APPEND|O_WRONLY)` only inherits the existing inode's mode because `O_TRUNC` is absent and sbsh does not `os.Remove` first today. If sbsh ever rotates these files, adds a truncate, or unlinks-and-recreates (mirroring how it already handles the socket — `internal/ctr/attachable.go:126-129` notes sbsh does `os.Remove + Listen`), the regression returns silently. The unit tests in this PR can't observe that contract; they only pin the host-side state at one snapshot.

The right shape, mirroring #260 → eminwux/sbsh#188:
1. sbsh adds `--capture-mode/--capture-gid` and `--log-file-mode/--log-file-gid` flags (filing on the sbsh repo).
2. kukeon extends `AttachableInjection` with the new mode+gid fields and `withAttachableArgsWrap` plumbs them through (filing as a new kukeon issue, depends on the sbsh release).

Issue #366 is being superseded by a new issue with the correct architecture. Please close this PR — leaving it open invites a future picker to merge the LGTM'd version.

PR #370, opened 2026-05-10
feat: plumb capture/log mode+gid through AttachableInjection for kukeon-group log access

### LINKED ISSUE #369: attachable: plumb capture/log mode+gid through AttachableInjection so kukeon-group operators can read peer cells via kuke log
## Context

Per-container `capture` and `log` files inside `<ttyDir>/` are created mode `0o600` owned by the container uid, so a non-root host operator that is a member of the `kukeon` group cannot read them via `kuke log`, even though the surrounding `tty/` directory is `02750 root:kukeon` and the sbsh control socket is `0660 root:kukeon` (post-#260). The original report and repro live on #366; this issue replaces #366 with the architecturally correct fix path.

The existing precedent for this exact concern in this codebase is the control socket: `internal/ctr/attachable.go:107-117` defines `AttachableInjection.SocketGID`, and `internal/ctr/attachable.go:195-200` plumbs `--socket-mode 0660 --socket-gid <SocketGID>` to `sbsh terminal` so sbsh applies the mode and chown after `Listen()`. The capture/log files belong on the same plumbing point — same directory, same kukeon-group read requirement, same dependency-injection structure already in place.

A prior attempt (PR #368) took a different shape — pre-create the inodes on the host and rely on sbsh's later `OpenFile(O_CREATE|O_APPEND|O_WRONLY)` inheriting the existing inode's mode — and was closed because it fragments the contract: a future sbsh change that adds `O_TRUNC` or unlinks-and-recreates (mirroring how it already handles the socket — `internal/ctr/attachable.go:126-129` notes sbsh does `os.Remove + Listen`) would silently regress the fix without any test catching it.

## Proposal

Extend `AttachableInjection` with capture/log mode+gid fields that mirror `SocketGID`, and have `withAttachableArgsWrap` plumb the new sbsh flags to `sbsh terminal`. sbsh-side flags are tracked separately on eminwux/sbsh#200; this issue is gated on that release.

### 1. `internal/ctr/attachable.go` — extend `AttachableInjection`

Add fields paralleling `SocketGID`:

```go
type AttachableInjection struct {
    // ... existing fields ...

    // CaptureGID and LogFileGID, when non-zero, are the numeric GIDs of
    // the kukeon system group on the host. The wrapper emits
    // `--capture-mode AttachableCaptureMode --capture-gid <CaptureGID>`
    // and `--log-file-mode AttachableLogFileMode --log-file-gid <LogFileGID>`
    // so the per-container capture transcript and log file are created
    // group-readable, matching the socket and parent tty/ layout.
    CaptureGID int
    LogFileGID int
}
```

In practice both will be the same GID (`r.opts.KukeonGroupGID`) for now, but separate fields keep the API parallel to `SocketGID` and leave room for divergent group choices later.

Add the mode constants next to `AttachableSocketMode`:

```go
// AttachableCaptureMode and AttachableLogFileMode are the octal modes
// passed to `sbsh terminal --capture-mode` / `--log-file-mode` when the
// corresponding GID is configured. 0640 = rw for owner (the container
// uid) + r for group (the kukeon group), no world. 0640 — not 0660 —
// because the host-side group only needs to tail the transcript; the
// in-container sbsh process does the writes. Available since sbsh
// v0.<TBD> (eminwux/sbsh#200); older sbsh binaries reject the flags
// and the wrapper omits them when GID is unset.
AttachableCaptureMode  = "0640"
AttachableLogFileMode  = "0640"
```

### 2. `internal/ctr/attachable.go` — extend `withAttachableArgsWrap`

After the existing `inj.SocketGID > 0` block:

```go
if inj.CaptureGID > 0 {
    wrapped = append(wrapped,
        "--capture-mode", AttachableCaptureMode,
        "--capture-gid", strconv.Itoa(inj.CaptureGID),
    )
}
if inj.LogFileGID > 0 {
    wrapped = append(wrapped,
        "--log-file-mode", AttachableLogFileMode,
        "--log-file-gid", strconv.Itoa(inj.LogFileGID),
    )
}
```

Flag names match eminwux/sbsh#200's proposed CLI (`--capture-mode/--capture-gid` and `--log-file-mode/--log-file-gid`). If sbsh ships a different spelling, this issue's AC tracks the final names.

### 3. `internal/controller/runner/attachable.go` — populate the new fields in `attachableBuildOpts`

Where `SocketGID: r.opts.KukeonGroupGID` is set today, also set:

```go
CaptureGID: r.opts.KukeonGroupGID,
LogFileGID: r.opts.KukeonGroupGID,
```

No `attachablePostCreateChown` changes: sbsh now applies the mode/group itself when it opens the files, exactly as it does for the socket.

### 4. Tests — `internal/ctr/attachable_test.go`

Mirror the existing `SocketGID` tests in the `kukeon-group attach end-to-end fix` family:
- when `CaptureGID > 0`, the wrapper appends `--capture-mode 0640 --capture-gid <gid>` to the `sbsh terminal` argv;
- when `LogFileGID > 0`, the wrapper appends `--log-file-mode 0640 --log-file-gid <gid>`;
- legacy fallback (`CaptureGID == 0` and `LogFileGID == 0`): the wrapper omits the flags so sbsh applies its hard-coded `0o600` default.

### 5. Smoke (test plan, not AC)

End-to-end: as a non-root host operator in the `kukeon` group, dispatch a cell via `kuke create`, then `kuke log --cell <name> --container <name> --no-follow` from a different uid in the same group. Should print the captured transcript, not `permission denied`.

## Acceptance criteria

- [ ] `AttachableInjection` gains `CaptureGID int` and `LogFileGID int` fields with godoc paralleling `SocketGID`.
- [ ] `AttachableCaptureMode = \"0640\"` and `AttachableLogFileMode = \"0640\"` constants added next to `AttachableSocketMode`, with the same backward-compat note.
- [ ] `withAttachableArgsWrap` emits the new flags when the corresponding GID is non-zero, omits them otherwise.
- [ ] `attachableBuildOpts` populates both new fields from `r.opts.KukeonGroupGID`.
- [ ] PR #368's host-side pre-create (`attachableEnsureLogFiles`) and its tests are removed; the new sbsh-driven path is the single mechanism for this concern.
- [ ] Unit tests cover the wrapper's flag-emission behavior in both the configured and legacy-fallback branches.
- [ ] On an end-to-end smoke run, a non-root kukeon-group operator can `kuke log` a peer cell's transcript without `permission denied`.

## Notes

- **Depends on eminwux/sbsh#200** — sbsh must ship the `--capture-mode/--capture-gid` and `--log-file-mode/--log-file-gid` flags first. Do not pick this up before the sbsh release lands; the staged binary at `/.kukeon/bin/sbsh` must accept the new flags.
- Supersedes #366 (closed) and PR #368 (closed, wrong-architecture host-side workaround).
- Pattern parity with #260 → eminwux/sbsh#188, which closed the same-shape gap for the control socket.
- Sizing: this fits one PR. The diff is purely additive on `internal/ctr/attachable.go`, `internal/controller/runner/attachable.go`, the matching `_test.go`s, plus removing the PR #368 host-side helper.

eminwux:
## Summary

- Per-container sbsh capture and log files inside `<ttyDir>/` landed `0o600` owned by the container uid because sbsh's hard-coded default opens them owner-only — the surrounding `tty/` dir is `02750 root:kukeon` and the control socket is `0660 root:kukeon` (post-#260), but the capture/log siblings stayed root-only, so a non-root operator in the `kukeon` group got EACCES when tailing peer cells via `kuke log`.
- Fixed by extending `AttachableInjection` with `CaptureGID` and `LogFileGID` fields paralleling `SocketGID`, and having `withAttachableArgsWrap` plumb `--capture-mode 0640 --capture-gid <gid>` and `--log-file-mode 0640 --log-file-gid <gid>` to `sbsh terminal` so sbsh applies the mode and chown after open. Same plumbing point and dependency-injection structure as the socket fix — the capture/log files are siblings of the socket, share the kukeon-group read requirement, and now go through the same single mechanism. Replaces PR #368's host-side pre-create approach, which fragmented the contract: a future sbsh that adds `O_TRUNC` or unlinks-and-recreates would silently regress that fix without any test catching it.
- 0640 (not 0660) because the host-side group only needs to read the captured PTY transcript; the in-container sbsh process does the writes. Mirrors the issue's proposal verbatim.
- `attachableBuildOpts` populates both new fields from `r.opts.KukeonGroupGID`. Zero falls back to sbsh's hard-coded `0o600` owner-only default — matching the legacy contract for hosts without `sysuser.EnsureUserGroup` and the older socket fallback.
- Bumps `github.com/eminwux/sbsh` to `v0.10.2` in `go.mod`/`go.sum` — the version that ships the `--capture-mode/--capture-gid` and `--log-file-mode/--log-file-gid` flags (eminwux/sbsh#200). The staged binary at `/.kukeon/bin/sbsh` must be v0.10.2 or later for the new flags to take effect; older sbsh binaries reject the flags and the wrapper omits them when the corresponding GID is zero.
- AC bullet about removing PR #368's `attachableEnsureLogFiles` is structurally a no-op: PR #368 was closed without merging, so `main` already lacks the helper — verified via `grep -rn "attachableEnsureLogFiles" --include='*.go'` returning no matches.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` — full unit suite green, including new `internal/ctr/attachable_test.go` cases:
  - wrapper appends `--capture-mode 0640 --capture-gid <gid>` and `--log-file-mode 0640 --log-file-gid <gid>` when both GIDs are non-zero, in the slot immediately before `--`
  - legacy fallback: with both GIDs unset the wrapper omits all four capture/log flags so sbsh applies its hard-coded `0o600` default (also guards against invoking sbsh-v0.10.2-only flags against an older staged binary)
  - independent gating: setting `CaptureGID` alone emits only the capture flags; setting `LogFileGID` alone emits only the log-file flags
- [ ] `make dev-init` smoke — could not run on this dev host: the docker build for the local kukeond image fails inside the dev container with `mount source: "overlay" ... err: invalid argument` from buildkit's nested overlayfs cachemount. Same environmental failure documented on PRs #367 and #368. The change is purely additive on the wrapper-arg path with no impact on the build path or the daemon's `/opt/kukeon` view; bootstrap doesn't use Attachable/sbsh, so the daemon-parity check is structurally unaffected.
- [ ] End-to-end `kuke log` smoke from a non-root kukeon-group operator — could not run on this dev host because (a) the local kukeond image build is blocked by the same overlay-on-overlay docker error above, and (b) the cached sbsh binary at `/opt/kukeon/cache/sbsh/<arch>/sbsh` would need to be updated to v0.10.2 on this host. Reviewer's environment will exercise this.

Closes #369

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/containerd/typeurl/v2 v2.2.3
 	github.com/containernetworking/cni v1.3.0
 	github.com/creack/pty v1.1.24
-	github.com/eminwux/sbsh v0.10.1
+	github.com/eminwux/sbsh v0.10.2
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/runtime-spec v1.2.1
 	github.com/spf13/cobra v1.10.1
--- a/go.sum
+++ b/go.sum
@@ -55,8 +55,8 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/distribution/reference v0.6.0 h1:0IXCQ5g4/QMHHkarYzh5l+u8T3t73zM5QvfrDyIgxBk=
 github.com/distribution/reference v0.6.0/go.mod h1:BbU0aIcezP1/5jX/8MP0YiH4SdvB5Y4f/wlDRiLyi3E=
-github.com/eminwux/sbsh v0.10.1 h1:uJY2XTHbz1bsYic3LPuu69t4/NtPW2gaIA+OflJOrfo=
-github.com/eminwux/sbsh v0.10.1/go.mod h1:AsDnK44VUZjB5tpUn6oV/0isaIudbLLQst8FYFXPJcs=
+github.com/eminwux/sbsh v0.10.2 h1:rqWDbl343evHUZ/hbIW2SxpZvo0vab3c3/Y8I5OUANI=
+github.com/eminwux/sbsh v0.10.2/go.mod h1:AsDnK44VUZjB5tpUn6oV/0isaIudbLLQst8FYFXPJcs=
 github.com/envoyproxy/go-control-plane v0.9.0/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
 github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
 github.com/envoyproxy/go-control-plane v0.9.4/go.mod h1:6rpuAdCZL397s3pYoYcLgu1mIlRU8Am5FuJP05cCM98=
--- a/internal/controller/runner/attachable.go
+++ b/internal/controller/runner/attachable.go
@@ -133,6 +133,12 @@ func (r *Exec) attachableBuildOpts(namespace string, spec intmodel.ContainerSpec
 			// host-side traversal layout `kuke init` sets up on /opt/kukeon.
 			// Zero falls back to sbsh's hard-coded 0600 owner-only default.
 			SocketGID: r.opts.KukeonGroupGID,
+			// Same group plumbed for the capture transcript and log file so
+			// `kuke log` from a non-root kukeon-group operator can read the
+			// 0640 root:kukeon files sbsh writes inside ttyDir. Zero falls
+			// back to sbsh's hard-coded 0600 owner-only default.
+			CaptureGID: r.opts.KukeonGroupGID,
+			LogFileGID: r.opts.KukeonGroupGID,
 		}),
 	}, nil
 }
--- a/internal/ctr/attachable.go
+++ b/internal/ctr/attachable.go
@@ -80,6 +80,18 @@ const (
 	// Available since sbsh v0.10.0; older sbsh binaries reject the flag and
 	// the wrapper omits it when the kukeon group GID is unset.
 	AttachableSocketMode = "0660"
+
+	// AttachableCaptureMode and AttachableLogFileMode are the octal modes
+	// passed to `sbsh terminal --capture-mode` / `--log-file-mode` when the
+	// corresponding GID is configured. 0640 = rw for owner (the container
+	// uid) + r for group (the kukeon group), no world. 0640 — not 0660 —
+	// because the host-side group only needs to tail the transcript via
+	// `kuke log`; the in-container sbsh process does the writes.
+	//
+	// Available since sbsh v0.10.2; older sbsh binaries reject the flags
+	// and the wrapper omits them when the corresponding GID is unset.
+	AttachableCaptureMode = "0640"
+	AttachableLogFileMode = "0640"
 )
 
 // AttachableInjection carries the host-side paths needed to wrap a container's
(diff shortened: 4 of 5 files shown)


eminwux:
**PR #370 Review — `feat: plumb capture/log mode+gid through AttachableInjection for kukeon-group log access`**

Clean parallel-to-`SocketGID` extension: two new `AttachableInjection` fields, two more `if gid > 0` guards in `withAttachableArgsWrap`, two new octal-mode constants. The gating preserves the legacy 0o600 fallback for hosts without a kukeon group and protects against staged-sbsh version skew (the gid > 0 guard, not the go.mod bump, is what keeps an older `/.kukeon/bin/sbsh` from seeing the v0.10.2 flags). 0640 vs 0660 reasoning matches the issue verbatim — host group only reads. No public-API parity gap (`AttachableInjection` is internal; the `SocketGID` in `pkg/api/model/v1beta1/serverconfiguration.go` is the daemon listener socket, a different concept). No unverified commits. All four checks green.

Tests cover the three branches (all three GIDs set, all unset, independent gating) and lock the flag ordering immediately before `--`, which is the right shape. Verified `attachableEnsureLogFiles` is indeed absent from `main` (PR #368 closed unmerged, as the description claims), so the AC bullet about removing it is correctly a no-op.

The `make dev-init` and end-to-end `kuke log` smokes are documented as host-gated (docker-overlay-on-overlay failure on the dev container, same env issue as #367/#368; staged sbsh on this host would also need a v0.10.2 bump). The change is purely additive on the wrapper-arg path with no impact on the daemon's `/opt/kukeon` read view, so the daemon-parity portion is structurally unaffected. Reviewer's environment will exercise the end-to-end check pre-merge.

LGTM.

PR #377, opened 2026-05-10
build: pass --allow-multiple-documents --unsafe to pre-commit check-yaml

### LINKED ISSUE #361: build: pre-commit check-yaml hook fails on mkdocs.yml + multi-resource.yaml fixture
## Repro

On `origin/main` at `41390d6` (current tip), with no local modifications, after the format-drift sweep in #308 / PR #360 lands:

```bash
git checkout main
pre-commit run check-yaml --all-files
```

## Observed

```
check yaml...............................................................Failed
- hook id: check-yaml
- exit code: 1

expected a single document in the stream
  in "cmd/kuke/apply/testdata/multi-resource.yaml", line 1, column 1
but found another document
  in "cmd/kuke/apply/testdata/multi-resource.yaml", line 7, column 1
could not determine a constructor for the tag 'tag:yaml.org,2002:python/name:pymdownx.superfences.fence_code_format'
  in "mkdocs.yml", line 49, column 19
```

This was verified pre-existing on `main` during the #308 cleanup work (PR #360) by stashing the format diff and re-running the hook against unmodified `main`.

## Impact

Same dev-contract friction as #308: `pre-commit run --all-files` exits non-zero on a clean checkout, so the PR-creation contract that requires hook output to be clean cannot be satisfied without explicitly skipping `check-yaml`. PR #360 will close #308 by clearing `go-fmt` and `prettier` drift, but `check-yaml` will continue to fail until this is addressed.

## Root cause

Both failures are legitimate uses of YAML that `check-yaml`'s default (safe-loader, single-document) does not accept:

- `cmd/kuke/apply/testdata/multi-resource.yaml` is a deliberate multi-document fixture for `kuke apply`'s multi-resource path. `check-yaml` needs `--allow-multiple-documents` (or the file needs to be excluded) to handle multi-doc input.
- `mkdocs.yml` line 49 uses `!!python/name:pymdownx.superfences.fence_code_format` — a legitimate Material-for-MkDocs extension wiring that requires PyYAML's full loader. `check-yaml`'s safe-loader cannot resolve it; the file needs to be excluded.

## Suggested fix

Edit `.pre-commit-config.yaml`'s `check-yaml` block to:

1. Pass `--allow-multiple-documents` via `args:` so multi-doc test fixtures load (this also covers any future multi-doc fixtures). Alternative: extend the existing `exclude:` regex to cover the testdata path.
2. Extend the existing `exclude:` regex (today: `^examples/profiles/local\.yaml$`) to also exclude `^mkdocs\.yml$`. The `!!python/name:` tag is intrinsic to mkdocs-material configuration and not lintable without the full loader.

Both edits are localized to the `check-yaml` hook in `.pre-commit-config.yaml`. After the fix, `pre-commit run check-yaml --all-files` should pass on a clean tree.

### LINKED ISSUE #362: pre-commit: configure check-yaml flags so it passes on main
Surfaced during review of #360. After that PR's gofmt+prettier sweep, `pre-commit run --all-files` is clean on \`go-fmt\`, \`prettier\`, \`addlicense\`, etc. — but \`check-yaml\` still fails on two files the dev did not touch:

- \`cmd/kuke/apply/testdata/multi-resource.yaml\` — multi-document YAML, needs \`--allow-multiple-documents\`.
- \`mkdocs.yml\` — uses Python-tagged YAML constructors (\`!!python/name:...\`), needs \`--unsafe\`.

Both were verified pre-existing on \`origin/main\` (#360 PR notes).

**Fix.** Update the \`check-yaml\` hook in \`.pre-commit-config.yaml\` to pass the appropriate flags. \`pre-commit-hooks\` exposes both as documented args:

\`\`\`yaml
- id: check-yaml
  args: ['--allow-multiple-documents', '--unsafe']
\`\`\`

\`--unsafe\` only disables checking for unknown tags (still validates structure), so it's safe for the testdata + mkdocs use case.

**Why followup-nit not nit-on-#360.** The fix is in scope of "clear pre-commit drift on main" but #360's title constrains it to gofmt+prettier; the dev correctly deferred. Worth tracking so the next dev who runs \`pre-commit run --all-files\` doesn't hit the same red light.

eminwux:
## Summary
- `pre-commit run --all-files` was non-zero on a clean `main` because `check-yaml` rejected the multi-doc `cmd/kuke/apply/testdata/multi-resource.yaml` and `mkdocs.yml`'s `!!python/name:` Material-for-MkDocs constructor — both legitimate uses the safe-loader default cannot accept (#361, #362).
- Add `args: ["--allow-multiple-documents", "--unsafe"]` to the existing `check-yaml` hook. `--unsafe` makes the hook parse via `yaml.parse` (token-level, no constructor resolution), so it still validates structure without requiring `pymdownx`. The existing `^examples/profiles/local\.yaml$` exclude is left untouched.

## Test plan
- [x] `pre-commit run check-yaml --all-files` → `Passed`
- [x] `pre-commit run --files .pre-commit-config.yaml` → all applicable hooks pass (yaml/json/toml/whitespace/prettier)
- [ ] `make test` — not run (out of scope: this PR only edits `.pre-commit-config.yaml`, no Go code touched)

Closes #361
Closes #362

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,6 +21,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
+        args: ["--allow-multiple-documents", "--unsafe"]
         exclude: ^examples/profiles/local\.yaml$
       - id: check-json
       - id: check-toml


eminwux:
**PR #377 Review — `build: pass --allow-multiple-documents --unsafe to pre-commit check-yaml`**

One-line config fix, correct: confirmed `cmd/kuke/apply/testdata/multi-resource.yaml` is multi-doc and `mkdocs.yml:49` uses `!!python/name:pymdownx.superfences.fence_code_format`, both of which the safe-loader default rejects. `--unsafe` switching `check-yaml` to token-level parsing is the right call — keeps structural validation without requiring `pymdownx` to be importable in the hook env. No security concerns, signed commit verified.

`make test` is correctly skipped per the test plan (no Go code touched, daemon/build-path regression guard not in scope per project CLAUDE.md). \`make dev-init\` is not required for a pre-commit-only change.

LGTM.

PR #378, opened 2026-05-10
fix: aggregate DeleteCell container errors and reconcile orphans by name

### LINKED ISSUE #371: runner: DeleteCell swallows containerd DeleteContainer errors, leaving containers behind after kuke daemon reset
## Repro

State: a previous `kuke init` run aborted mid-flight while creating the kukeond cell — e.g., the kukeond image hadn't been loaded into the realm yet, so the containerd container was created but the workload pull failed. Cell metadata exists; one of the cell's containerd containers is still present.

1. Trigger an init that creates the containerd container but fails before completing (in this session this happened naturally: the dev-init script's first-pass `kuke init` ran before `kuke image load`, so the workload `kukeon_kukeon_kukeond_kukeond` container was created but its image pull from docker.io failed with `pull access denied`).
2. `./kuke daemon reset --purge-system` — reports success:

   ```
   kukeond was already stopped (cell "kukeond" in realm "kuke-system")
   kukeond cell deleted (cell "kukeond" in realm "kuke-system")
   removed /opt/kukeon/kuke-system
   ```

3. `ctr -n kuke-system.kukeon.io containers ls` still shows:

   ```
   CONTAINER                        IMAGE                                 RUNTIME
   kukeon_kukeon_kukeond_kukeond    docker.io/library/kukeon-local:dev    io.containerd.runc.v2
   ```

4. Re-run `kuke init` — fails:

   ```
   Error: failed to create cell: failed to create container kukeon_kukeon_kukeond_kukeond: failed to create container from spec: container already exists
   ```

The only manual recovery from this state is `ctr -n kuke-system.kukeon.io containers rm kukeon_kukeon_kukeond_kukeond` — `kuke daemon reset` cannot get the host back to a clean re-bootstrap state.

## Observed

`kuke daemon reset` reports the kukeond cell as deleted and removes `/opt/kukeon/kuke-system`, but the underlying containerd container survives. The host is left in a state where the controller thinks the cell is gone (no metadata) but containerd still owns the container — `kuke init` cannot recreate it because the containerd ID collides.

## Root cause

In `internal/controller/runner/delete_cell.go`, `DeleteCell` swallows `DeleteContainer` failures other than `ErrContainerNotFound` and continues to the metadata-removal step:

- Workload-container loop, lines 153–165:
  ```go
  err = r.ctrClient.DeleteContainer(internalRealm.Spec.Namespace, containerID, ctr.ContainerDeleteOptions{
      SnapshotCleanup: true,
  })
  if err != nil {
      if errors.Is(err, errdefs.ErrContainerNotFound) {
          r.logger.DebugContext(r.ctx, "container already deleted", "container", containerID)
      } else {
          r.logger.WarnContext(r.ctx, "failed to delete container", "container", containerID, "error", err)
          // Continue with other containers
      }
  }
  ```
- Root-container path, lines 200–212, has the same shape.

After both loops, lines 246–258 unconditionally delete the cell metadata file and run `os.RemoveAll(metadataRunPath)`, so the next `kuke init` re-derives the same containerd ID with no awareness that an orphan container exists. Same family as #292 (cell start swallowed CNI errors containing 'already exists') and #193 (uninstall reported a realm purged while the namespace survived).

A second contributor: `internal/controller/runner/delete_cell.go` lines 113–123 also `continue` past containers whose `Spec.ContainerdID` is empty. If a partial init persisted the cell document before the spec's `ContainerdID` was filled in, that container is never even attempted — same end state.

## Suggested fix

Make `DeleteCell` failure-aware:
- Treat any `DeleteContainer` error other than `ErrContainerNotFound` (and any `StopContainer` error other than not-found) as fatal for the reset path. Aggregate per-container errors and return a wrapped `errdefs.ErrDeleteCell` so `cmd/kuke/daemon/reset/reset.go` can surface the partial state instead of declaring success.
- Belt-and-suspenders for the orphan case: when `Spec.ContainerdID` is empty or DeleteContainer returns a non-not-found error, do a name-based scan of the realm's containerd namespace for `<space>_<stack>_<cell>_*` prefixes and reconcile leftovers before deleting the metadata. Important for the partial-init case where the spec was persisted with empty ContainerdID.
- Reset's CLI layer (`cmd/kuke/daemon/reset/reset.go` lines 137–147) should keep the "kukeond cell deleted" line tied to a real success — today `delRes.MetadataDeleted` is true even when the underlying container survives.

A reproduction test for `internal/controller/runner/delete_cell_test.go` should fake a `DeleteContainer` "container already exists"-class error and assert that DeleteCell returns non-nil and metadata is *not* removed.

eminwux:
## Summary
- DeleteCell now treats non-NotFound DeleteContainer failures as fatal: per-container errors are aggregated, wrapped in `errdefs.ErrDeleteCell`, and returned instead of swallowed — so the metadata file stays on disk and the next `kuke daemon reset` finds the cell and retries the teardown.
- Adds a `<space>_<stack>_<cell>_*` name-based orphan scan in the realm's containerd namespace, covering the partial-init case where the cell spec was persisted with an empty `ContainerdID` (the workload loop used to silently `continue` past that spec, leaving the container in containerd while the metadata was removed).
- Restructured the per-container stop+delete into a `stopAndDeleteContainer` helper so the workload loop, root container, and orphan scan share the same gate. Cgroup and metadata removal now run only after every container call returns NotFound or success — partial container-tier failures no longer half-tear-down the cell.

## Test plan
- [x] `go test ./internal/controller/runner/ ./internal/controller/ ./cmd/kuke/daemon/reset/` (passes; new `TestDeleteCell_PreservesMetadataOnDeleteContainerFailure`, `TestDeleteCell_ReconcilesOrphanFromEmptyContainerdID`, `TestFindOrphanContainerIDs`)
- [x] `make test` (full suite, all packages green)
- [x] `make dev-init` smoke (daemon-parity tail matches expected output for both `kuke get realms` and `--no-daemon`)
- [x] Live orphan repro: ran `sudo ./kuke init`, manually created an extra container `kukeon_kukeon_kukeond_orphan` in `kuke-system.kukeon.io` via `ctr containers create`, ran `sudo ./kuke daemon reset`, confirmed `ctr containers ls` is empty post-reset — the orphan was reconciled by the new name-based scan exactly as #371's "belt-and-suspenders" suggested fix described.

Closes #371

### DIFF
--- a/internal/controller/runner/delete_cell.go
+++ b/internal/controller/runner/delete_cell.go
@@ -72,6 +72,7 @@ func (r *Exec) DeleteCell(cell intmodel.Cell) error {
 	if err != nil {
 		return fmt.Errorf("failed to get realm: %w", err)
 	}
+	namespace := internalRealm.Spec.Namespace
 
 	cellSpaceName := internalCell.Spec.SpaceName
 	cellStackName := internalCell.Spec.StackName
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #378 Review — `fix: aggregate DeleteCell container errors and reconcile orphans by name`**

Faithful implementation of the fix #371 prescribed: per-container failures aggregate into a wrapped `errdefs.ErrDeleteCell`, metadata stays on disk so the next `kuke daemon reset` retries the teardown, and the name-based orphan scan covers the partial-init case where the cell spec was persisted with an empty `ContainerdID`. The `stopAndDeleteContainer` extraction cleans up duplication across the workload, root, and orphan paths and preserves the `force` asymmetry the prior code had (workloads SIGTERM, root + orphans SIGKILL). Prefix `<space>_<stack>_<cell>_` is unambiguous because `naming.ValidateHierarchyName` rejects `_` in any hierarchy name, so sibling-cell collisions can't happen by construction. Tests exercise both the aggregated-failure and empty-`ContainerdID`-orphan paths, plus a unit on `findOrphanContainerIDs`. Per-commit signature verified, CI green, dev-init smoke ran per project CLAUDE.md.

LGTM.

PR #380, opened 2026-05-10
fix: switch bash autocomplete to cobra V2 to route every tab through __complete

### LINKED ISSUE #375: cli: kuke run -p autocomplete does not pick up newly added profiles without re-sourcing the shell
## Context

User report: after adding a new profile under `$HOME/.kuke/profiles.d/`, tab-completion on `kuke run -p <TAB>` does not list the new profile until the shell is restarted or the completion script is re-sourced. Expected behavior: completion crawls the active profiles directory on every tab press.

The runtime path *looks* dynamic on inspection: `CompleteProfileNames` calls `cellprofile.List(dir)` at completion time (`cmd/config/autocomplete.go:466-487`), and `kuke autocomplete bash` emits cobra V1 bash completion that routes each tab to `kuke __complete` (`cmd/kuke/autocomplete/autocomplete.go:43-56`). So the symptom suggests one of:

- Shell-side: a stale `complete -W "..."` from a prior install or shell config is shadowing the cobra `__complete` dispatch.
- Cobra V1 quirk: V1 bash output has known caching/static-list edge cases on some bash-completion versions; V2 (`GenBashCompletionV2`) routes every press through `__complete` unconditionally.
- Process-side: `cellprofile.ResolveDir()` resolves to a path that differs from the user's runtime dir at completion time.

## Proposal

Reproduce, identify the root cause, and fix:

1. Repro: install completion (`kuke autocomplete bash | sudo tee /etc/bash_completion.d/kuke`), open a fresh shell, add a new profile YAML under `$HOME/.kuke/profiles.d/`, tab-complete `kuke run -p <TAB>`, confirm the new profile is missing.
2. If the issue is shell-side, document the workaround in the PR and switch the generated script to `GenBashCompletionV2` (and zsh/fish equivalents) so every tab routes to `__complete`.
3. If the issue is process-side, fix the resolution path so `ResolveDir` agrees between completion-time and runtime.

## Acceptance criteria

- [ ] Adding a new YAML under `$HOME/.kuke/profiles.d/` makes the new profile appear in `kuke run -p <TAB>` in the same shell, with no re-source.
- [ ] Removing a profile makes it disappear from completion in the same shell.
- [ ] `$KUKE_PROFILES_DIR` override is honored on every tab in the same shell.
- [ ] Repro and root cause documented in the PR description.

## Notes

Related: `CompleteProfileNames` at `cmd/config/autocomplete.go:466-487`; `cellprofile.ResolveDir`/`List` in `internal/cellprofile/`; generated completion via `cmd/kuke/autocomplete/autocomplete.go`.

eminwux:
## Summary

- `kuke autocomplete bash` now emits `GenBashCompletionV2` instead of V1. V2 ships a tiny dispatcher that routes every tab through `kuke __complete`; V1 ships ~3500 lines of inlined per-command flag arrays plus a `__kuke_handle_go_custom_completion` shim. The inlined arrays are the shadowing surface — a stale `/etc/bash_completion.d/kuke` left from a prior install (the install path the issue's repro uses) keeps loading old flag tables even after `sudo tee` overwrites the file, because bash-completion lazy-loads the script per shell. V2 has no inlined arrays to go stale, and matches the dispatch model zsh/fish already use.
- Direct dynamic check (`kuke __complete run -p ""`) was already returning live results in both V1 and V2 — so the daemon-side completer at `cmd/config/autocomplete.go:466-487` is not the bug. The bug is the bash-side dispatcher's shape, not the Go side.
- Generated-script size drops from 3564 to 426 lines.

## Test plan

- [x] `go test ./cmd/kuke/autocomplete/...` — passes; new V2-marker assertion replaces the prior `len > 0` check, which the old test was satisfying with root-help text because `bashCmd.Execute()` was not actually routing through the bash subcommand.
- [x] AC #1 (add yaml → appears without re-source): sourced the generated V2 script in a fresh bash, simulated tab on `kuke run -p <TAB>`, added a second yaml under `$KUKE_PROFILES_DIR`, repeated the same tab without re-sourcing → both profiles listed.
- [x] AC #2 (remove yaml → disappears without re-source): removed the first yaml, repeated tab → only the remaining profile listed.
- [x] AC #3 (`$KUKE_PROFILES_DIR` override honored per-tab): switched the env var to a different dir with a third yaml, repeated tab → only the third profile listed.
- [x] Same checks against `--profile=` and `--profile <space>` long forms — both update live, no asymmetry with `-p`.
- [x] `pre-commit run --files cmd/kuke/autocomplete/autocomplete.go cmd/kuke/autocomplete/autocomplete_test.go` — passes.
- [ ] `make dev-init` smoke — not run; this PR touches only `cmd/kuke/autocomplete/`, which is not on the daemon read path or the `/opt/kukeon` write path the daemon-parity guard catches.

## Notes for reviewer

- I narrowed the `TestAutocompleteBash` body alongside the fix because the old test asserted only `len(output) > 0` and was passing on the root command's help text (`bashCmd.Execute()` with an empty `SetArgs` falls back to help) — it had been silently dead for a while. The new test asserts the V2 dispatcher marker `__kuke_get_completion_results` is present and the V1 dispatcher `__kuke_handle_go_custom_completion` is absent, so a future revert is caught. `TestAutocompleteZsh` and `TestAutocompleteFish` have the same dead-test pattern but I left them alone — out of scope for this fix.
- `GenBashCompletionV2(os.Stdout, true)` — `true` keeps descriptions on tab (parity with zsh/fish).

Closes #375

### DIFF
--- a/cmd/kuke/autocomplete/autocomplete.go
+++ b/cmd/kuke/autocomplete/autocomplete.go
@@ -50,7 +50,10 @@ func newBashCmd() *cobra.Command {
 			if rootCmd == nil {
 				return errors.New("failed to get root command")
 			}
-			return rootCmd.GenBashCompletion(os.Stdout)
+			// V2 routes every tab through `kuke __complete` and omits the
+			// inlined flag arrays V1 ships; matches zsh/fish behavior and
+			// avoids stale-script shadowing across re-installs (#375).
+			return rootCmd.GenBashCompletionV2(os.Stdout, true)
 		},
 	}
 }
--- a/cmd/kuke/autocomplete/autocomplete_test.go
+++ b/cmd/kuke/autocomplete/autocomplete_test.go
@@ -82,59 +82,39 @@ func TestAutocompleteBash(t *testing.T) {
 	autocompleteCmd := autocomplete.NewAutocompleteCmd()
 	rootCmd.AddCommand(autocompleteCmd)
 
-	// Get bash subcommand
-	bashCmd, _, err := rootCmd.Find([]string{"autocomplete", "bash"})
-	if err != nil {
-		t.Fatalf("Failed to find bash subcommand: %v", err)
-	}
-
-	// Capture stdout
-	var buf bytes.Buffer
+	// Capture stdout — the bash RunE writes the generated script there.
 	oldStdout := os.Stdout
 	r, w, _ := os.Pipe()
 	os.Stdout = w
 
-	// Execute command in goroutine
 	errChan := make(chan error, 1)
 	go func() {
-		bashCmd.SetOutput(&buf)
-		bashCmd.SetContext(ctx)
-		bashCmd.SetArgs([]string{})
-		err := bashCmd.Execute()
+		rootCmd.SetArgs([]string{"autocomplete", "bash"})
+		errChan <- rootCmd.Execute()
 		w.Close()
-		errChan <- err
 	}()
 
-	// Read output
 	var output bytes.Buffer
-	_, readErr := output.ReadFrom(r)
-	if readErr != nil {
+	if _, readErr := output.ReadFrom(r); readErr != nil {
 		t.Fatalf("Failed to read from pipe: %v", readErr)
 	}
 	r.Close()
-
-	// Restore stdout
 	os.Stdout = oldStdout
 
-	// Wait for command to complete
-	execErr := <-errChan
-	if execErr != nil {
+	if execErr := <-errChan; execErr != nil {
 		t.Fatalf("Execute() error = %v, want nil", execErr)
 	}
 
-	// Verify output contains bash completion script markers
+	// Verify output is the cobra V2 bash completion: it routes every tab
+	// through __complete via __kuke_get_completion_results, instead of the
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #380 Review — `fix: switch bash autocomplete to cobra V2 to route every tab through __complete`**

Clean, narrowly scoped fix. The V1→V2 switch eliminates the inlined flag arrays that drove the stale-script shadowing in #375, and the test rewrite replaces a silently-passing assertion (`len(output) > 0` matching root help text from a no-arg `bashCmd.Execute()`) with V2-dispatcher presence + V1-dispatcher absence — a real regression guard. No security surface; commit is signed and verified.

Nothing blocking.

- **Followup (non-blocking)** `cmd/kuke/autocomplete/autocomplete_test.go`: `TestAutocompleteZsh` and `TestAutocompleteFish` carry the same `bashCmd.Execute()` no-arg dead-test pattern the bash test just fixed — they'd silently pass against root help text too. Out of scope for this PR per your notes; I'll file as a followup-nit.

LGTM.

Issue #384, opened 2026-05-10
docs: note install-dev host-symlink vs daemon-image distinction in CLAUDE.md smoke-test section


eminwux:
## Context

#294 added `make install-dev` / `make uninstall-dev` Makefile targets and wired `make dev-init` to invoke `install-dev` so contributors can run `kuke` and `kukeond` directly from `PATH` after the bootstrap. The PR did not touch project `CLAUDE.md` (per dev's role rule on doc PRs).

`CLAUDE.md`'s "Local smoke test" section explains the build → load → init loop but does not mention the new install step or its scope. There is one non-obvious caveat worth documenting before contributors hit it: the host symlinks update the *CLI*; the *daemon* still runs from inside the `kuke-system/kukeon/kukeon/kukeond` cell using the `kukeon-local:dev` image, so updating the daemon binary still requires `make dev-init`'s image rebuild + reload step (the new symlinks alone don't refresh the running daemon).

## Proposal

Add a short paragraph to `CLAUDE.md` "Local smoke test: rebuild and re-run `kuke init`" — under the "One-shot: `make dev-init`" subsection — documenting that `dev-init` now also installs `kuke` and `kukeond` symlinks via `make install-dev`, and naming the host-CLI vs daemon-image distinction.

Verbatim proposed wording (paste into `CLAUDE.md` after the daemon-parity-table block, before "If your change touches anything the daemon reads…"):

> The script also runs `make install-dev`, which symlinks the freshly built `./kuke` and `./kukeond` into `INSTALL_PREFIX` (`/usr/local/bin` by default — override on hosts where that path is read-only or sudo-gated, e.g. `INSTALL_PREFIX=\$HOME/.local/bin make install-dev`). After the first `make dev-init`, `kuke …` and `kukeond …` resolve from `PATH` from any working directory. `make uninstall-dev` removes both symlinks.
>
> The host symlink only updates the *CLI* path. The *daemon* still runs from inside the `kuke-system/kukeon/kukeon/kukeond` cell using the `kukeon-local:dev` image, so to pick up daemon-side changes you still need `make dev-init`'s full image rebuild + `kuke image load` + `kuke init` step — re-running `make install-dev` alone refreshes only the host CLI.

## Acceptance criteria
- [ ] `CLAUDE.md` "Local smoke test" → "One-shot: \`make dev-init\`" subsection includes the verbatim wording above (or a substantively equivalent paraphrase).
- [ ] No code changes — pure docs PR.

## Out of scope
[...]

PR #385, opened 2026-05-10
build: add install-dev Makefile target to symlink dev binaries onto PATH

### LINKED ISSUE #294: build: make dev-init builds kuke/kukeond but does not install them on PATH; contributors must run ./kuke from the source dir
## Repro

On a fresh checkout of `main`:

```bash
cd ~/projects/kukeon0
make dev-init        # completes successfully
cd /tmp
sudo kuke get realms # not found
sudo kukeond --help  # not found
```

Reproduced on commit `36f07dd` (current `main`).

## Observed

`make dev-init` produces `./kuke` and `./kukeond` (a symlink) in the repo root and runs the bootstrap loop using `sudo ./kuke …` (`scripts/dev-init.sh:46,52,58,61,64-65`). After `make dev-init` succeeds, neither binary is reachable through `PATH` — including `sudo`'s default `secure_path` (`/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`). Every subsequent invocation has to be:

- spelled `./kuke …` from the source dir, or
- spelled with an absolute path, or
- preceded by `cd ~/projects/kukeon0`.

This breaks the natural \"I just ran an init script, now I can use the tool\" expectation and forces every contributor to memorise the source-relative invocation. It also diverges from the published `kuke …` examples in the docs and CLAUDE.md daemon-parity check, which read as if `kuke` is on PATH.

## Root cause

`Makefile:37-45` builds `./kuke` and produces an in-tree `kukeond -> kuke` symlink only. `Makefile:95-97`'s `dev-init` target shells out to `scripts/dev-init.sh`, which never installs or symlinks the binaries onto a PATH directory. There is no `install` target in the Makefile (a `release-build` exists at `Makefile:48-60` but produces tagged release artefacts, not host installs). End-user installation is tracked separately in #62 (curl | bash installer), which is the production path — the *dev* workflow has no equivalent.

This is the same self-sufficiency framing as #293: the bootstrap script should leave the host in a state where `kuke` and `kukeond` are usable as tools, idempotently, without the contributor having to track the source directory.

## Suggested fix

Add an opt-in `install-dev` Makefile target that symlinks (not hard-copies) the dev binaries into `/usr/local/bin`, and have `dev-init` either invoke it or print a one-liner pointing at it.

Symlink, not hard copy, because:
- A symlink keeps `make kuke` → next invocation in sync automatically. A hard copy means every rebuild silently runs the *previous* installed version unless the contributor remembers to re-install — exactly the footgun a dev workflow can't afford.
- It matches the existing `kukeond -> kuke` convention already used in the source tree (`Makefile:44-45`) and inside the kukeond image (`Dockerfile:59`, hard link there because the image's `/bin/kuke` is the only copy).
- argv[0] dispatch is by basename of the path used to exec, so two parallel symlinks both pointing at the source binary work cleanly:
  ```
  /usr/local/bin/kuke    -> <repo>/kuke
  /usr/local/bin/kukeond -> <repo>/kuke
  ```

Sketch:

```make
INSTALL_PREFIX ?= /usr/local/bin
.PHONY: install-dev uninstall-dev
install-dev: kuke
	sudo ln -sf \$(CURDIR)/kuke \$(INSTALL_PREFIX)/kuke
	sudo ln -sf \$(CURDIR)/kuke \$(INSTALL_PREFIX)/kukeond
uninstall-dev:
	sudo rm -f \$(INSTALL_PREFIX)/kuke \$(INSTALL_PREFIX)/kukeond
```

Open design questions for the implementer:
- Should `make dev-init` invoke `install-dev` automatically, or print a hint? Auto-invoke is more self-sufficient (matches the user's framing); hint is lower blast radius. Recommend auto-invoke gated by an existing `/usr/local/bin/kuke` symlink check so re-runs stay idempotent.
- Document one caveat in the project `CLAUDE.md` Local-smoke-test section: the host symlink updates the *CLI* path; the *daemon* still runs from inside the `kuke-system/kukeon/kukeon/kukeond` cell using the `kukeon-local:dev` image, so updating the daemon still requires `make dev-init`'s image rebuild + reload step.

## Test plan
- After `make dev-init` on a fresh host, `which kuke` and `which kukeond` both resolve to `/usr/local/bin/...`.
- Re-running `make dev-init` produces no errors and no duplicate symlinks.
- `make uninstall-dev` removes both symlinks; `make dev-init` recreates them.
- argv[0] dispatch still works: `sudo kukeond --help` shows the daemon CLI, `sudo kuke --help` shows the client CLI.

eminwux:
## Summary
- After `make dev-init` the freshly built `./kuke` and `./kukeond` binaries sit only in the source dir, so every contributor command has to be `./kuke …` from the repo root or absolute-pathed. The new `install-dev` target symlinks `kuke` and `kukeond` into `$(INSTALL_PREFIX)` (default `/usr/local/bin`) so the natural `sudo kuke get realms` works from anywhere.
- Symlinks (not hard copies) so the next `make kuke` is picked up automatically — a stale copy would silently run the previous binary, exactly the footgun a dev workflow can't afford. `kukeond` is a parallel symlink to the same source binary; argv[0] basename dispatch resolves it to the daemon entrypoint.
- `make dev-init` auto-invokes `install-dev` so the existing one-shot flow lands the symlinks too. `ln -sf` makes the install idempotent across re-runs; `make uninstall-dev` removes both symlinks.
- The CLAUDE.md note the issue suggested (host symlinks update the *CLI* path, but the *daemon* still ships from the `kuke-system/.../kukeond` cell so daemon updates still require `make dev-init`'s image rebuild + reload) belongs in `<project>/CLAUDE.md`, which is filed as a separate `docs:` follow-up per the role split.

## Test plan
- [x] `go build ./...`
- [x] `make dev-init` end-to-end on a fresh host: daemon-parity check shows both `default` and `kuke-system` `Ready`
- [x] After `make dev-init`, `which kuke` → `/usr/local/bin/kuke`, `which kukeond` → `/usr/local/bin/kukeond`, both symlinks pointing at `/root/kukeon/kuke`
- [x] `sudo kuke --help` shows the client CLI; `sudo kukeond --help` shows the daemon CLI (argv[0] dispatch intact)
- [x] `make install-dev` re-run after a successful install is idempotent (no errors, no duplicate symlinks); `make uninstall-dev` removes both; `make install-dev` recreates them
- [ ] `pre-commit run --files Makefile scripts/dev-init.sh` — `pre-commit` binary not installed on this dev host (only `Makefile` + `scripts/dev-init.sh` changed; manually verified no trailing whitespace and trailing newlines present)

Closes #294

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,10 @@ KUKEON_DOCKER_IMAGE := $(KUKEON_REGISTRY)/$(KUKEON_IMAGE_NAME):$(KUKEON_IMAGE_TA
 # --kukeond-image is not passed. Release pipeline overrides via env.
 KUKEON_IMAGE_REPO ?= ghcr.io/eminwux/kukeon
 
+# Directory `install-dev` symlinks the dev binaries into. Override for
+# rootless / non-standard PATH layouts (e.g. INSTALL_PREFIX=$HOME/.local/bin).
+INSTALL_PREFIX ?= /usr/local/bin
+
 LDFLAGS := -s -w \
 	-X $(MODULE)/cmd/config.Version=$(KUKEON_VERSION) \
 	-X $(MODULE)/cmd/config.KukeondImageRepo=$(KUKEON_IMAGE_REPO)
@@ -95,3 +99,18 @@ tag:
 .PHONY: dev-init
 dev-init:
 	./scripts/dev-init.sh
+
+# install-dev symlinks the in-tree dev binaries into INSTALL_PREFIX so
+# contributors can invoke `kuke` / `kukeond` from anywhere on the host after
+# `make dev-init`. Symlinks (not copies) so subsequent `make kuke` rebuilds
+# are picked up automatically — a stale hard copy is exactly the footgun a
+# dev workflow can't afford. argv[0] dispatch resolves `kukeond` to the
+# daemon entrypoint because the basename of the exec path is `kukeond`.
+.PHONY: install-dev uninstall-dev
+install-dev: kuke
+	ln -sf kuke kukeond
+	sudo ln -sf $(CURDIR)/kuke $(INSTALL_PREFIX)/kuke
+	sudo ln -sf $(CURDIR)/kuke $(INSTALL_PREFIX)/kukeond
+
+uninstall-dev:
+	sudo rm -f $(INSTALL_PREFIX)/kuke $(INSTALL_PREFIX)/kukeond
--- a/scripts/dev-init.sh
+++ b/scripts/dev-init.sh
@@ -32,6 +32,9 @@ step "Build kuke (and the kukeond symlink)"
 make kuke
 ln -sf kuke kukeond
 
+step "Install dev symlinks on PATH (make install-dev)"
+make install-dev
+
 # Pre-flight: catch missing host cgroup-v2 controller delegation BEFORE the
 # multi-minute docker build runs (issue #324). On a fully-delegated host
 # this is silent; on a misconfigured host (e.g. cgroup namespace whose


eminwux:
The CLAUDE.md caveat the original issue called out (host CLI symlink vs daemon image) is tracked in #384.

eminwux:
**PR #385 Review — `build: add install-dev Makefile target to symlink dev binaries onto PATH`**

Clean implementation of #294. Two symlinks point `/usr/local/bin/{kuke,kukeond}` at `$(CURDIR)/kuke` so argv[0] dispatch resolves correctly, and `make dev-init` auto-invokes the new target so the existing one-shot flow lands the symlinks. Commit is signed, all checks green, daemon-parity smoke is documented in the test plan. No correctness or security concerns.

One non-blocking nit:

- **Doc/behavior mismatch** `Makefile:21-23,112-116`: the comment names `INSTALL_PREFIX=$HOME/.local/bin` as a rootless override, but the recipes hardcode `sudo` for both `install-dev` and `uninstall-dev`. In a true rootless env (no sudo binary, or no sudo permission) the documented override fails outright; in a user-owned `INSTALL_PREFIX` it prompts for a password unnecessarily. Either drop `sudo` (let the user run `sudo make install-dev` for system paths) or thread it through a `SUDO ?= sudo` variable.

LGTM.

PR #388, opened 2026-05-10
fix: pre-flight containerd reachability in dev-init.sh

### LINKED ISSUE #344: bug: dev-init.sh masks containerd-unreachable as 'image not staged' (fails late at image load)
### Summary

\`scripts/dev-init.sh\` tolerates a non-zero exit from the first-pass \`kuke init\` because \"the local image is not yet staged in containerd\" (per the comment at \`scripts/dev-init.sh:49-58\`). The tolerate is too broad: when \`/run/containerd/containerd.sock\` is unreachable entirely, the same swallow eats the connection-timeout error, the script continues into \`kuke image load --realm kuke-system\` which then fails with the misleading \`Error: realm not found: kuke-system\`. Contributor sees a downstream message that has nothing to do with the actual cause.

### Reproduction

On a host where dockerd is running but standalone containerd is not (i.e. \`/run/containerd/containerd.sock\` does not exist):

\`\`\`
make dev-init
\`\`\`

Observed (trimmed):

\`\`\`
==> First-time bootstrap: run kuke init to create realm metadata
Error: failed to connect to containerd: failed to verify new connection: connection error: desc = \"transport: Error while dialing: dial unix:///run/containerd/containerd.sock: timeout\"
first-pass init returned non-zero (expected before image is staged); continuing
...
==> Load kukeon-local:dev into the kuke-system realm
Error: realm not found: kuke-system
make: *** [Makefile:97: dev-init] Error 1
\`\`\`

The \`first-pass init returned non-zero (expected before image is staged); continuing\` message is misleading — the failure was containerd-down, not image-not-staged.

### Proposed fix

Add a pre-flight reachability probe alongside the existing cgroup pre-flight at \`scripts/dev-init.sh:43-44\` (which already exists for the same reason — fail fast before the multi-minute docker build). One-liner:

\`\`\`bash
step \"Pre-flight: containerd reachability\"
[ -S /run/containerd/containerd.sock ] || die \"containerd is not running at /run/containerd/containerd.sock — start it before re-running\"
\`\`\`

This puts the check next to the docker-build / cgroup checks, surfaces the actual cause immediately, and matches the README's stated prerequisite (\"containerd running at /run/containerd/containerd.sock\", \`README.md:31\`).

### Acceptance criteria

- New pre-flight step in \`scripts/dev-init.sh\` exits non-zero with a clear message naming the missing socket when \`/run/containerd/containerd.sock\` is absent.
- The check runs before \`docker build\`, so a contributor without containerd doesn't pay the build cost.
- The existing tolerate-non-zero on the first-pass \`kuke init\` stays — its purpose (image-not-staged) remains valid.
- Idempotent on a healthy host (silent when the socket exists).

### Test plan

- On a host with containerd up: \`make dev-init\` runs to the existing parity-check tail.
- On a host with containerd down: \`make dev-init\` fails at the new pre-flight step with the named-socket message, before \`docker build\` runs.

eminwux:
## Summary
- The first-pass `kuke init` in `scripts/dev-init.sh` tolerates a non-zero exit because the local kukeond image is not yet staged in containerd. That tolerate also swallows a containerd-unreachable error, so a host with `/run/containerd/containerd.sock` missing fails minutes later at `kuke image load` with the misleading `realm not found: kuke-system`.
- Adds a pre-flight reachability probe alongside the existing cgroup pre-flight. Fails fast with the named-socket message before the multi-minute `docker build` runs; silent on a healthy host.

## Test plan
- [x] `bash -n scripts/dev-init.sh`
- [x] Manual exercise of both branches: silent when `/run/containerd/containerd.sock` is a socket, exits 1 with the named-socket message when absent
- [x] `make test`
- [x] `go build ./...`, `go vet ./...`
- [x] `make dev-init` runs to the daemon parity-check tail (both `kuke get realms` and `--no-daemon` produce identical output) on a host with containerd up

Closes #344

### DIFF
--- a/scripts/dev-init.sh
+++ b/scripts/dev-init.sh
@@ -46,6 +46,17 @@ make install-dev
 step "Pre-flight: host cgroup controller delegation"
 sudo ./kuke doctor cgroups --probe
 
+# Pre-flight: catch a missing standalone containerd BEFORE the multi-minute
+# docker build, and BEFORE the first-pass `kuke init` below whose tolerate-
+# non-zero (image-not-staged) would otherwise swallow the connection-timeout
+# error and surface a misleading downstream "realm not found: kuke-system"
+# at image load (issue #344).
+step "Pre-flight: containerd reachability"
+if [ ! -S /run/containerd/containerd.sock ]; then
+    printf 'containerd is not running at /run/containerd/containerd.sock — start it before re-running\n' >&2
+    exit 1
+fi
+
 step "Build local kukeond image (${LOCAL_TAG})"
 docker build --build-arg VERSION=v0.0.0-dev -t "${LOCAL_TAG}" .
 


eminwux:
**PR #388 Review — `fix: pre-flight containerd reachability in dev-init.sh`**

Tight, well-scoped fix for #344. The probe lands in the right spot — before both the multi-minute `docker build` and the first-pass `kuke init` whose `|| echo …` tolerate is the actual source of the masked containerd-unreachable error. Socket path `/run/containerd/containerd.sock` is the documented default everywhere in the tree (`cmd/kuke/kuke.go:180`, `cmd/kukeond/kukeond.go:89`, `internal/clientconfig/clientconfig.go:91-92`), so the hardcoded check is consistent with dev-init.sh's own assumption (it never overrides `--containerd-socket`). Test plan is fully ticked including the `make dev-init` parity-check smoke. Signature verified. No security or correctness concerns.

LGTM.

Issue #382, opened 2026-05-10
autocomplete: rewrite zsh/fish tests to actually exercise their RunE


eminwux:
Surfaced during review of #380.

`cmd/kuke/autocomplete/autocomplete_test.go` — `TestAutocompleteZsh` and `TestAutocompleteFish` carry the same dead-test shape that `TestAutocompleteBash` had before #380:

```go
cmd, _, _ := rootCmd.Find([]string{"autocomplete", "zsh"})  // or "fish"
cmd.SetArgs([]string{})
cmd.Execute()
```

With an empty `SetArgs`, the resolved subcommand falls back to its help text instead of running its `RunE`, so the assertion passes against help output rather than the generated completion script. That hid the V1 bash bug for as long as it took to ship — same exposure exists today for zsh and fish.

Suggested fix: mirror the bash test's new shape — route through the root with `rootCmd.SetArgs([]string{"autocomplete","zsh"})` and `rootCmd.Execute()`, then assert on a marker the real generated script must contain (e.g. `_zsh_completion` / `__kuke_complete`-equivalent for zsh, `complete -c kuke -f -a` for fish). Same guard pattern as #380's V2 assertion.

PR #389, opened 2026-05-10
test: rewrite zsh/fish autocomplete tests to exercise their RunE

### LINKED ISSUE #381: test(autocomplete): TestAutocompleteZsh/Fish exercise root help instead of the generated script
## Repro

Branch `fix/profile-autocomplete-stale` at HEAD (any branch where the pre-#375 test bodies survive — `main` as of a4a68fd has identical patterns).

\`\`\`go
// cmd/kuke/autocomplete/autocomplete_test.go — TestAutocompleteZsh / TestAutocompleteFish
zshCmd, _, _ := rootCmd.Find([]string{"autocomplete", "zsh"})
zshCmd.SetArgs([]string{})
zshCmd.Execute()
// → falls back to root help because SetArgs([]) on the subcommand does not actually drive the bash/zsh/fish dispatch
\`\`\`

Add a print statement (`fmt.Fprintf(os.Stderr, ...)`) before the existing assertion or run the equivalent test body in a scratch `go run` harness — the captured output is the root command's help text (\"Kukeon is a tool for managing Kukeon entities / Available Commands: ...\", len≈2347 bytes), not the zsh/fish completion script.

## Observed

The current assertion in both tests is:

\`\`\`go
if !strings.Contains(outputStr, \"compdef\") && !strings.Contains(outputStr, \"#compdef\") {
    if len(outputStr) == 0 {
        t.Error(\"Expected non-empty output from zsh completion generation\")
    }
}
if len(outputStr) == 0 {
    t.Error(\"Zsh completion script output is empty\")
}
\`\`\`

Root help is non-empty, so the fallback `len == 0` branch never fires; the test passes on help text that contains neither `compdef` nor `#compdef`. Equivalent for fish: the help body contains the substring `complete` (from `\"completion script\"` in `Short`), so `strings.Contains(outputStr, \"complete\")` is true on help text and the empty-output guard never fires.

Net: neither test actually exercises `GenZshCompletion` / `GenFishCompletion`. A revert of either generator would land silently.

## Root cause

`subCmd.SetArgs([]string{})` followed by `subCmd.Execute()` is not the cobra subcommand-test pattern — `Execute()` traverses from the root and the empty `Args` resolves to the root's help. The correct pattern is `rootCmd.SetArgs([]string{\"autocomplete\", \"zsh\"})` + `rootCmd.Execute()`, which is the rewrite `TestAutocompleteBash` got as part of #375 (PR #380).

## Suggested fix

Apply the same rewrite #380 used for `TestAutocompleteBash`:

1. Drive the subcommand via `rootCmd.SetArgs(...)` + `rootCmd.Execute()`, not `subCmd.Execute()` with empty args.
2. Strengthen the assertion from \"non-empty\" to a marker that proves the generator ran — e.g. `#compdef _kuke` for zsh, `complete -c kuke` (or the `__kuke_perform_completion` function name) for fish.

Out of scope for #380 because the issue there was bash-only and dev/CLAUDE.md prohibits unrelated cleanups in fix PRs.

eminwux:
## Summary
- `TestAutocompleteZsh` / `TestAutocompleteFish` previously drove the subcommand with `subCmd.SetArgs([]string{})` + `subCmd.Execute()`, which falls back to root help instead of running `GenZshCompletion` / `GenFishCompletion`. The assertions then passed against help output: zsh's empty-output fallback never fired, and fish's `strings.Contains(out, "complete")` matched help text because the autocomplete root's `Short` description contains "completion script".
- Mirrors the rewrite #380 applied to `TestAutocompleteBash`: route through `rootCmd.SetArgs([]string{"autocomplete", "<shell>"})` + `rootCmd.Execute()`, capture stdout (the RunE writes to `os.Stdout`), and assert on markers the real script must contain (so help output cannot satisfy them).
- Zsh markers: `#compdef kuke` (script header) and `__complete` (V2 dispatch verb — same V2 hot-path the bash test guards). Fish markers: `complete -c kuke` (per-arg dispatch) and `__kuke_perform_completion` (request-dispatch function). Verified all four substrings are absent from `kuke help` / `kuke --help` output, so a regression that returns help would now fail the assertions.

This is a test-only change; only `cmd/kuke/autocomplete/autocomplete_test.go` is modified. The daemon and `/opt/kukeon` are untouched, so the `make dev-init` daemon-parity smoke is not required.

Note: #382 describes the same fix as #381 (filed independently after the same review) and the change here satisfies its acceptance criteria as well, but per the dev role I'm only closing the bug-labeled issue and leaving #382 for pm to dedupe.

## Test plan
- [x] `go test ./cmd/kuke/autocomplete/... -run TestAutocomplete -v` — all five pass
- [x] `go vet ./...`, `go build ./...`
- [x] `pre-commit run --files cmd/kuke/autocomplete/autocomplete_test.go`
- [ ] `make dev-init` — not run; change is test-only and does not touch the daemon, build path, or `/opt/kukeon`

Closes #381

### DIFF
--- a/cmd/kuke/autocomplete/autocomplete_test.go
+++ b/cmd/kuke/autocomplete/autocomplete_test.go
@@ -134,59 +134,40 @@ func TestAutocompleteZsh(t *testing.T) {
 	autocompleteCmd := autocomplete.NewAutocompleteCmd()
 	rootCmd.AddCommand(autocompleteCmd)
 
-	// Get zsh subcommand
-	zshCmd, _, err := rootCmd.Find([]string{"autocomplete", "zsh"})
-	if err != nil {
-		t.Fatalf("Failed to find zsh subcommand: %v", err)
-	}
-
-	// Capture stdout
-	var buf bytes.Buffer
+	// Capture stdout — the zsh RunE writes the generated script there.
 	oldStdout := os.Stdout
 	r, w, _ := os.Pipe()
 	os.Stdout = w
 
-	// Execute command in goroutine
 	errChan := make(chan error, 1)
 	go func() {
-		zshCmd.SetOutput(&buf)
-		zshCmd.SetContext(ctx)
-		zshCmd.SetArgs([]string{})
-		err := zshCmd.Execute()
+		rootCmd.SetArgs([]string{"autocomplete", "zsh"})
+		errChan <- rootCmd.Execute()
 		w.Close()
-		errChan <- err
 	}()
 
-	// Read output
 	var output bytes.Buffer
-	_, readErr := output.ReadFrom(r)
-	if readErr != nil {
+	if _, readErr := output.ReadFrom(r); readErr != nil {
 		t.Fatalf("Failed to read from pipe: %v", readErr)
 	}
 	r.Close()
-
-	// Restore stdout
 	os.Stdout = oldStdout
 
-	// Wait for command to complete
-	execErr := <-errChan
-	if execErr != nil {
+	if execErr := <-errChan; execErr != nil {
 		t.Fatalf("Execute() error = %v, want nil", execErr)
 	}
 
-	// Verify output contains zsh completion script markers
+	// Guard against silent regression to a code path that returns root help
+	// (which contains neither marker) instead of running GenZshCompletion.
+	// `#compdef kuke` is the cobra zsh header; `__complete` is the V2
+	// request-dispatch verb that proves the script drives `kuke __complete`
+	// at runtime — the same hot path the bash V2 test guards.
 	outputStr := output.String()
-	if !strings.Contains(outputStr, "compdef") && !strings.Contains(outputStr, "#compdef") {
-		// Zsh completion scripts typically contain these markers
-		// But we'll just verify we got some output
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #389 Review — `test: rewrite zsh/fish autocomplete tests to exercise their RunE`**

LGTM. The diff cleanly mirrors the bash test pattern from #380 — route through `rootCmd.SetArgs([]string{"autocomplete", "<shell>"})` + `rootCmd.Execute()` so the subcommand's `RunE` actually runs, capture `os.Stdout` via `os.Pipe`, and assert on script-specific markers (`#compdef kuke` / `__complete` for zsh; `complete -c kuke` / `__kuke_perform_completion` for fish) that help output can't satisfy. The comment explaining *why* each marker matters (script header vs. V2 dispatch verb, per-arg dispatch vs. perform-completion function) is the right level of detail for a regression guard.

Test-only diff in `cmd/kuke/autocomplete/`; daemon, build path, and `/opt/kukeon` are untouched, so the `make dev-init` smoke is correctly skipped per `CLAUDE.md`. No sensitive data, no dead code, no public API change. Commit is verified.

PR #394, opened 2026-05-10
fix: pre-flight containerd reachability in kuke init

### LINKED ISSUE #391: bug: TestRunInitErrors/controller_failure fails on main — runInit succeeds with bogus containerd socket
## Repro

On a clean checkout of `origin/main` (verified at `0385750`, `fix: pre-flight containerd reachability in dev-init.sh (#388)`):

```bash
git clone https://github.com/eminwux/kukeon
cd kukeon
go test -run TestRunInitErrors/controller_failure ./cmd/kuke/init/...
```

The test passes `viper.Set(config.KUKEON_ROOT_CONTAINERD_SOCKET.ViperKey, filepath.Join(tmp, "missing.sock"))` and runs `RunInit` against a fresh `t.TempDir()` run-path. It expects `RunInit` to return *any* error (the assertion at `cmd/kuke/init/init_test.go:763` is just `err == nil → t.Fatalf("expected error, got nil")`); a follow-up check at `init_test.go:772` further requires the error not be `ErrLoggerNotFound`.

## Observed

```
=== RUN   TestRunInitErrors/controller_failure
    init_test.go:764: expected error, got nil
--- FAIL: TestRunInitErrors/controller_failure (0.09s)
```

`RunInit` returns `nil` even though the socket path `tmp/missing.sock` does not exist. No CI exposure because the test passed at the time of its merge; this is a regression observable locally (and the upstream `make test` job ran on a hosted runner that may differ environmentally).

I reproduced this on `origin/main` *without* my own change applied (stashed my diff, re-ran, same failure), so this is independent of #297 / #390.

## Root cause

Not yet pinpointed. Candidate shortlist for the next picker, ordered most→least likely:

1. `cmd/kuke/init/init.go:362-365` — `controller.NewControllerExec(...).Bootstrap()` may be returning a no-op success when prerequisite state (containerd socket, kukeond image) is unreachable. If `Bootstrap` silently early-returns instead of dialing containerd, the test will observe a nil error.
2. `cmd/kuke/init/init.go:336-348` — `sysuser.EnsureUserGroup` and `installForwardAdmission` run *before* the controller and may be silently no-op'ing on systems where the kukeon group/admission chain already exists from a prior `make dev-init` (worth checking on hosts that have never run init).
3. `cmd/kuke/init/init.go:325-330` — `instance.VerifyOrWrite` against a fresh `t.TempDir()` should at minimum write an instance marker; if it then short-circuits the rest of the flow somewhere downstream, `Bootstrap` may not be reached.
4. The test setup itself at `cmd/kuke/init/init_test.go:739-752` may need `viper.Set` for additional keys that have crept into `runInit` (e.g. `config.KUKEOND_SOCKET`, `config.KUKEON_ROOT_NAMESPACE_SUFFIX`, `config.KUKEON_ROOT_CGROUP_ROOT`) since the test was written — if a `Load`/`Configure` step ahead of the controller now bails early, the original "controller failure" assertion never gets exercised.

## Suggested fix

Start by reproducing locally and adding a `t.Logf` of the actual `err` (and the report from `Bootstrap` when err is nil) at `init_test.go:763` to see whether `Bootstrap` returned a partial-success report or whether the early `serverconfig.Load`/`ConfigureRuntime`/`instance.VerifyOrWrite` steps short-circuited the controller call entirely. Once the silent-success path is identified, either (a) tighten the production code so a missing `ContainerdSocket` surfaces as an explicit error before bootstrap reports success, or (b) update the test to reach `Bootstrap` deterministically (e.g. by injecting a faked controller via the same MockClientKey pattern the `cmd/kuke/daemon/start` tests already use). Option (a) is the right fix if there is a real production hole; option (b) is acceptable only if no production-observable bug exists and the test is purely defending a guard that has been moved.

eminwux:
## Summary
- `kuke init` could silently succeed against an unreachable containerd socket because `runner.ExistsRealmContainerdNamespace` deliberately returns `(false, nil)` when the socket is missing (test ergonomics), letting the "namespace missing → create it" path proceed through downstream no-ops. This mirrors the script-level pre-flight already in `scripts/dev-init.sh` (PR #388) inside `runInit` itself, so the daemon path fails fast with a clear `ErrConnectContainerd`-wrapped error instead.
- Tightens `TestRunInitErrors/controller_failure` (renamed to `unreachable_containerd_socket`) to assert specifically `errors.Is(err, errdefs.ErrConnectContainerd)`. The old assertion (`err != nil && !ErrLoggerNotFound`) passed by accident on hosts where an earlier step happened to fail (e.g., missing `iptables` binary) while still allowing the production hole to ship. The test setup now also points `KUKE_INIT_SERVER_CONFIGURATION` at a non-existent file so `applyServerConfiguration` does not silently overwrite the bogus socket with whatever lives in `/etc/kukeon/kukeond.yaml` on the test host, and re-seeds the namespace/cgroup defaults that an earlier `viper.Reset` in the package wipes.

## Test plan
- [x] `pre-commit run --files cmd/kuke/init/init.go cmd/kuke/init/init_test.go` — all hooks pass (go fmt, go-mod-tidy, golangci-lint, addlicense, …).
- [x] `make dev-init` — daemon-parity check matches the expected output (both `kuke get realms` and `--no-daemon` show the two-realm Ready table); the pre-flight is a transparent no-op on a healthy host.
- [x] `go test -run TestRunInitErrors/unreachable_containerd_socket ./cmd/kuke/init/... -v` — exercises the new pre-flight path with the bogus socket and confirms `errdefs.ErrConnectContainerd` propagates.

Closes #391

### DIFF
--- a/cmd/kuke/init/init.go
+++ b/cmd/kuke/init/init.go
@@ -45,6 +45,12 @@ const (
 	kukeondReadyTimeout = 30 * time.Second
 	kukeondReadyTick    = 200 * time.Millisecond
 
+	// defaultContainerdSocket is the conventional path of a standalone
+	// containerd's gRPC socket. Used as the fallback when no socket is
+	// configured (matches the containerd library's own default and the
+	// pre-flight in scripts/dev-init.sh).
+	defaultContainerdSocket = "/run/containerd/containerd.sock"
+
 	// File modes applied by `kuke init` so the kukeon group can reach the
 	// runtime/socket without world access. Writes under /opt/kukeon still
 	// require root and go through the daemon.
@@ -243,6 +249,29 @@ func resolveKukeondImage() string {
 	return fmt.Sprintf("%s:%s", repo, tag)
 }
 
+// preflightContainerdSocket fails fast when the configured containerd socket
+// does not exist or is not a socket file. Without this check, RunInit can
+// reach the controller bootstrap and silently succeed against an unreachable
+// socket — runner.ExistsRealmContainerdNamespace deliberately returns
+// (false, nil) on a missing socket for test ergonomics, which lets the
+// "namespace missing, create it" path proceed through downstream no-ops.
+func preflightContainerdSocket(path string) error {
+	info, err := os.Stat(path)
+	if err != nil {
+		return fmt.Errorf(
+			"%w: %s — start containerd before re-running `kuke init`: %w",
+			errdefs.ErrConnectContainerd, path, err,
+		)
+	}
+	if info.Mode()&os.ModeSocket == 0 {
+		return fmt.Errorf(
+			"%w: %s is not a socket file — verify containerd is running and the path is correct",
+			errdefs.ErrConnectContainerd, path,
+		)
+	}
+	return nil
+}
+
 // installForwardAdmission installs the kukeon-owned FORWARD admission chain
 // so cells reach the network on hosts where `iptables -P FORWARD DROP` is
 // the default (Docker, firewalld, ufw, hardened distros). Idempotent —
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #394 Review — `fix: pre-flight containerd reachability in kuke init`**

Tight, surgical fix. The pre-flight is correctly placed *after* `applyServerConfiguration` (so the server-config layer wins before the check), the `os.Stat` + `os.ModeSocket` test mirrors the `scripts/dev-init.sh` precedent, and `errdefs.ErrConnectContainerd` is the right shared sentinel. The test re-write is a real improvement — the old assertion silently passed on hosts where an unrelated step (e.g. missing `iptables`) failed first, and the new `errors.Is` lock-in plus the `KUKE_INIT_SERVER_CONFIGURATION` redirect prevent the host's `/etc/kukeon/kukeond.yaml` from contaminating the case. No correctness, secret-leak, or convention concerns. Single commit, signed and verified; CI green.

LGTM.

PR #395, opened 2026-05-11
fix: preserve ErrContainerNotFound sentinel in kuke attach / kuke run attach wraps

### LINKED ISSUE #152: attach: paired %w cleanup for ErrContainerNotFound wrap in runAttachLoop and kuke attach
Surfaced during review of #150.

Both attach call sites drop the wrap when `AttachContainer` returns `ErrContainerNotFound`, breaking `errors.Is` for upstream callers:

- `cmd/kuke/run/attach.go:128`: `return fmt.Errorf("container %q not found in cell %q", container, doc.Metadata.Name)`
- `cmd/kuke/attach/attach.go:129`: `return fmt.Errorf("container %q not found in cell %q", container, cell)`

Both should use `: %w` and pass the original `err` so the sentinel is preserved across the call boundary, matching the project's error-handling convention (and the adjacent `ErrAttachNotSupported` branches at run/attach.go:125 and attach/attach.go:125, which already wrap correctly).

Bundle-worthy because it touches two sites and aligns with how every other `errors.Is(..., ErrContainerNotFound)` caller in the tree expects the sentinel to propagate (see e.g. `internal/controller/runner/delete_cell.go`).

eminwux:
## Summary
- Both `kuke attach` and the post-create attach in `kuke run` dropped the `%w` wrap when `AttachContainer` returned `ErrContainerNotFound`, so upstream `errors.Is` calls could not see the sentinel — diverging from the adjacent `ErrAttachNotSupported` branches which already wrap correctly and from how `delete_cell` and friends expect the sentinel to propagate.
- Re-add `: %w` to both call sites and pass the original `err`.
- Lock the propagation in with two new tests (one per call site) that stub `AttachContainer` to return `ErrContainerNotFound` and assert `errors.Is(err, ErrContainerNotFound)` on the top-level command error.

## Test plan
- [x] `go test ./cmd/kuke/run/... ./cmd/kuke/attach/...`
- [x] `make test`
- [x] `pre-commit run --files cmd/kuke/attach/attach.go cmd/kuke/attach/attach_test.go cmd/kuke/run/attach.go cmd/kuke/run/run_test.go`
- [ ] `make dev-init` smoke test — change is purely CLI-side error-wrap formatting in `kuke attach` and `kuke run`; touches no build path, no daemon code, nothing the daemon reads, and nothing under `/opt/kukeon`, so the daemon-parity regression guard does not apply.

Closes #152

### DIFF
--- a/cmd/kuke/attach/attach.go
+++ b/cmd/kuke/attach/attach.go
@@ -125,7 +125,7 @@ func runAttach(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("container %q in cell %q is not attachable: %w", container, cell, err)
 		}
 		if errors.Is(err, errdefs.ErrContainerNotFound) {
-			return fmt.Errorf("container %q not found in cell %q", container, cell)
+			return fmt.Errorf("container %q not found in cell %q: %w", container, cell, err)
 		}
 		return err
 	}
--- a/cmd/kuke/attach/attach_test.go
+++ b/cmd/kuke/attach/attach_test.go
@@ -321,6 +321,34 @@ func TestAttach_ExplicitContainer_NotAttachable_SurfacesSentinel(t *testing.T) {
 	}
 }
 
+// TestAttach_ExplicitContainer_NotFound_SurfacesSentinel locks in the
+// ErrContainerNotFound branch's %w wrap: when AttachContainer's RPC reports
+// the named container doesn't exist, kuke attach must propagate the
+// sentinel so upstream callers can still errors.Is it.
+func TestAttach_ExplicitContainer_NotFound_SurfacesSentinel(t *testing.T) {
+	t.Cleanup(viper.Reset)
+
+	fc := &fakeClient{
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #395 Review — `fix: preserve ErrContainerNotFound sentinel in kuke attach / kuke run attach wraps`**

Clean, well-scoped fix: both call sites now `%w`-wrap the sentinel to match the adjacent `ErrAttachNotSupported` branches, and the two new tests lock the behavior in via `errors.Is`. Single signed commit, all checks green, no sensitive data in the new error strings.

No blocking items. One followup observation (non-blocking):

- **Same pattern elsewhere** `cmd/kuke/log/log.go:160`, `cmd/kuke/get/container/container.go:163`, `cmd/kuke/daemon/logs/logs.go:108-112`: three other call sites still drop the `ErrContainerNotFound` sentinel the same way this PR fixes for attach/run. Out of this PR's scope (issue #152 was explicitly attach-only) but bug-adjacent; will file a followup-nit.

LGTM — proceed to merge.

PR #397, opened 2026-05-11
refactor: extract container picker to cmd/kuke/shared.PickContainer

### LINKED ISSUE #184: cmd/kuke: extract pickAttachableContainer to shared package
Surfaced as a non-blocking nit on PR #183.

`pickAttachableContainer` lives in two subcommand packages with byte-identical bodies:

- \`cmd/kuke/log/log.go:287-316\`
- \`cmd/kuke/attach/attach.go:174-202\`

Both enumerate \`client.ListContainers\`, filter \`Root || !Attachable\`, sort, and return \`ErrAttachNoCandidate\` / \`ErrAttachAmbiguous\` with identical message shapes. The daemon-side dupl between \`AttachContainer\` and \`LogContainer\` was already extracted via \`resolveAttachable\` in \`internal/client/local/local.go\`; the client-side picker is the remaining duplicate.

**Suggested fix:** move \`pickAttachableContainer\` to \`cmd/kuke/shared\` next to \`ClientFromCmd\` and call it from both subcommands. Keeps the picker semantics canonical so a future change to one subcommand cannot drift the other.

Bundle-worthy because: cross-package, behavior-preserving, and protects an invariant (both subcommands target the same attachable container set) that is currently only enforced by convention.

eminwux:
## Summary

- `kuke attach` and `kuke log` each carried a near-duplicate `pick{Attachable,Log}Container` that enumerated `client.ListContainers`, sorted candidates, and wrapped `ErrAttachNoCandidate` / `ErrAttachAmbiguous` with byte-identical message shapes. The only diff is the inclusion filter — attach requires `Attachable=true`, log accepts non-Attachable too. Lifted the enumeration/sort/error semantics into `kukeshared.PickContainer(ctx, client, …, include func(spec) bool)` so a future change to one subcommand cannot drift the other; each call site now passes its own one-line filter.
- Behavior-preserving: existing `cmd/kuke/attach` and `cmd/kuke/log` integration tests (which exercise the picker via `cmd.Execute()`) continue to pass against the new shared helper. New direct unit tests in `cmd/kuke/shared` lock in (a) the `ErrAttachNoCandidate` / `ErrAttachAmbiguous` sentinel wraps, (b) the sorted candidate list in the ambiguous error, (c) attach's Attachable filter rejecting a lone non-Attachable container, (d) log's filter accepting a lone non-Attachable container, and (e) `ListContainers` errors propagating unwrapped.
- Note vs the issue body's "byte-identical bodies" claim: the two pickers diverged in #203, when `pickLogContainer` dropped the Attachable filter so `kuke log` could surface kukeond's `cio.LogFile`. The issue's stated invariant ("both subcommands target the same attachable container set") is therefore not preserved by the dedup — they target structurally similar sets with different filters. The canonical-semantics goal still holds (one picker, one error shape, one sort), and the filter divergence is now an explicit predicate at each call site rather than a buried `if` inside a near-clone.

## Test plan

- [x] `go test ./cmd/kuke/shared/... ./cmd/kuke/attach/... ./cmd/kuke/log/...`
- [x] `make test`
- [x] `pre-commit run --files cmd/kuke/attach/attach.go cmd/kuke/log/log.go cmd/kuke/shared/pick_container.go cmd/kuke/shared/pick_container_test.go`
- [ ] `make dev-init` smoke test — change is purely CLI-side picker-helper refactor in `kuke attach` / `kuke log`; touches no build path, no daemon code, nothing the daemon reads, and nothing under `/opt/kukeon`, so the daemon-parity regression guard does not apply.

Closes #184

### DIFF
--- a/cmd/kuke/attach/attach.go
+++ b/cmd/kuke/attach/attach.go
@@ -27,7 +27,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"sort"
 	"strings"
 
 	"github.com/eminwux/kukeon/cmd/config"
@@ -112,7 +111,10 @@ func runAttach(cmd *cobra.Command, args []string) error {
 	defer func() { _ = client.Close() }()
 
 	if container == "" {
-		container, err = pickAttachableContainer(cmd.Context(), client, realm, space, stack, cell)
+		container, err = kukeshared.PickContainer(cmd.Context(), client, realm, space, stack, cell,
+			func(spec v1beta1.ContainerSpec) bool {
+				return !spec.Root && spec.Attachable
+			})
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


eminwux:
**PR #397 Review — `refactor: extract container picker to cmd/kuke/shared.PickContainer`**

Clean dedup. The shared `PickContainer` preserves the exact enumeration/sort/error semantics of both pickers (sentinel-wraps via `%w`, alphabetical sort in the ambiguous list, identical message shapes), and the divergence in filter policy is now an explicit one-line predicate at each call site rather than a buried `if` inside a near-clone. Behavior-preserving for `kuke attach` and `kuke log`; the third sibling at `cmd/kuke/run/attach.go::pickAttachTarget` is a different policy (operates on `CellSpec`, honors `--container`/`Tty.Default`, first-in-declaration order, different sentinels) and rightly stays out of this refactor. Tests lock in the sentinel wraps via `errors.Is`, the per-subcommand filter shapes, the sorted-ambiguous-list contract, and `ListContainers` error propagation. License headers present on both new files; sentinels reused from `internal/errdefs`. `make dev-init` skip is justified — diff touches only CLI-side helper code, no build path / daemon / `/opt/kukeon`. Commit signed.

LGTM.

PR #400, opened 2026-05-11
fix: bind KUKE_INIT_* env vars in cmd/kuke/init

### LINKED ISSUE #190: kuke get: KUKEON_GET_OUTPUT env var silently ignored — ParseOutputFormat bypasses viper

### LINKED ISSUE #399: cmd/kuke/init: KUKE_INIT_KUKEOND_IMAGE env var silently ignored — resolveKukeondImage bypasses viper binding
## Repro

Worktree: `main` at 72cddbe (\`fix: pre-flight containerd reachability in kuke init\`).

Add the following test to \`cmd/kuke/init/\` and run \`go test ./cmd/kuke/init/... -run TestReproduceKukeondImageEnvIgnored -v\`:

\`\`\`go
package init_test

import (
    "os"
    "testing"

    "github.com/eminwux/kukeon/cmd/config"
    "github.com/spf13/viper"
)

func TestReproduceKukeondImageEnvIgnored(t *testing.T) {
    viper.Reset()
    t.Cleanup(viper.Reset)
    origVersion := config.Version
    origRepo := config.KukeondImageRepo
    t.Cleanup(func() {
        config.Version = origVersion
        config.KukeondImageRepo = origRepo
    })
    config.Version = "v1.0.0"
    config.KukeondImageRepo = "ghcr.io/eminwux/kukeon"

    os.Setenv("KUKE_INIT_KUKEOND_IMAGE", "docker.io/library/kukeon-local:dev")
    t.Cleanup(func() { _ = os.Unsetenv("KUKE_INIT_KUKEOND_IMAGE") })

    got := ResolveKukeondImage()
    if got != "docker.io/library/kukeon-local:dev" {
        t.Errorf("env value ignored: got %q, want %q", got, "docker.io/library/kukeon-local:dev")
    }
}
\`\`\`

(\`ResolveKukeondImage\` already exists as a test-only export in \`cmd/kuke/init/init_test.go\`.)

## Observed

\`\`\`
=== RUN   TestReproduceKukeondImageEnvIgnored
    reproducer_test.go:31: env value ignored: got "ghcr.io/eminwux/kukeon:v1.0.0", want "docker.io/library/kukeon-local:dev"
--- FAIL: TestReproduceKukeondImageEnvIgnored (0.00s)
\`\`\`

The operator-supplied env value \`docker.io/library/kukeon-local:dev\` is silently dropped; \`resolveKukeondImage\` returns the computed default (\`KukeondImageRepo:Version\`) instead.

In a real \`kuke init\` invocation the symptom is: \`KUKE_INIT_KUKEOND_IMAGE=… sudo kuke init\` (without \`--kukeond-image\` and without a \`ServerConfiguration\` setting \`kukeondImage\`) provisions the kukeond cell with the compiled-in default registry image, not the value the operator exported.

## Root cause

\`KUKE_INIT_KUKEOND_IMAGE\` is declared with \`viper.BindPFlag\` against \`--kukeond-image\` in \`cmd/kuke/init/init.go:115\`, but \`viper.BindEnv\` is **never** called for this key — \`grep -rn 'KUKE_INIT_KUKEOND_IMAGE\.BindEnv' .\` returns no matches anywhere in the tree, and there is no \`viper.AutomaticEnv()\` / \`SetEnvPrefix\` either. \`cmd/kukeond/kukeond.go:179 bindEnvVars()\` covers the daemon-side keys but the \`kuke init\` path has no equivalent.

The asymmetry creates a worst-of-both-worlds failure mode: \`applyServerConfiguration\` in \`cmd/kuke/init/init.go:207\` correctly detects the env var via \`envSet()\` (which uses \`os.LookupEnv\` directly) and skips the \`viper.Set\` of the \`ServerConfiguration\` value — so the env *blocks* the spec from winning — but downstream \`viper.GetString(config.KUKE_INIT_KUKEOND_IMAGE.ViperKey)\` in \`resolveKukeondImage\` (\`cmd/kuke/init/init.go:236\`) returns \`\"\"\` because no env binding exists, so the fallback path runs. Net effect: the env value is recognized as a gate but its content is lost, breaking the documented \`--flag > env > ServerConfiguration > default\` precedence rule.

This is the same bug class as the closed #190 (\`KUKEON_GET_OUTPUT env var silently ignored — ParseOutputFormat bypasses viper\`); the same omission likely affects other \`KUKE_INIT_*\` keys (\`KUKE_INIT_NO_WAIT\`, \`KUKE_INIT_FORCE_REGENERATE_CNI\`, \`KUKE_INIT_SERVER_CONFIGURATION\`, \`KUKE_INIT_REALM\`, \`KUKE_INIT_SPACE\`) — the suggested fix should sweep the lot rather than patching just \`KUKEOND_IMAGE\`.

## Suggested fix

Mirror \`cmd/kukeond/kukeond.go\`'s \`bindEnvVars()\` on the kuke side: add a \`bindEnvVars()\` (or extend \`cmd/kuke/kuke.go\`'s \`loadConfig\`, which already binds the shared \`KUKEON_*\` keys) that iterates every \`KUKE_INIT_*\` \`config.Var\` and calls \`v.BindEnv()\`. Then add an \`init_test.go\` regression that mirrors PR #398's \`TestApplyServerConfigurationEnvOverridesConfig\` *without* the explicit per-test \`BindEnv\` calls — proving the production code path wires env binding for \`KUKE_INIT_KUKEOND_IMAGE\` and \`KUKEON_RUN_PATH\` end-to-end.

Surfaced alongside #186 / PR #398 (test-only mirror).

eminwux:
## Summary
- `cmd/kuke/init` declared `KUKE_INIT_KUKEOND_IMAGE` (and siblings) with `viper.BindPFlag` but never called `viper.BindEnv` — so `applyServerConfiguration` saw `envSet` and skipped the YAML write while downstream `viper.GetString` returned `""` because no env binding existed, silently dropping the operator-exported value. Same bug class as closed #190 / sibling #191.
- Add a `bindEnvVars()` in `cmd/kuke/init/init.go` mirroring `cmd/kukeond/kukeond.go`'s; called from `setupInitCmd` so every `KUKE_INIT_*` viper key is wired to its env var.
- Simplify `TestApplyServerConfigurationEnvOverridesConfig` to remove the explicit per-test `BindEnv` calls. The test now uses `kuke.LoadConfig()` for the shared `KUKEON_*` keys and `NewInitCmd()` (which now calls `bindEnvVars()`) for the `KUKE_INIT_*` keys — proving env binding is wired end-to-end via the production code path.

## Test plan
- [x] `go test ./cmd/kuke/init/... -run TestApplyServerConfigurationEnvOverridesConfig -v` (passes)
- [x] Reproduced the issue's `TestReproduceKukeondImageEnvIgnored` against the fixed branch with the new production binding path — env value now wins.
- [x] `make test` (full suite, all packages pass)
- [x] `pre-commit run --files cmd/kuke/init/init.go cmd/kuke/init/init_test.go` (all hooks pass)
- [x] `make dev-init` — daemon-parity tail shows both `kuke get realms` and `kuke get realms --no-daemon` listing `default` and `kuke-system` Ready.

Closes #399

### DIFF
--- a/cmd/kuke/init/init.go
+++ b/cmd/kuke/init/init.go
@@ -92,7 +92,28 @@ func setupInitCmd(cmd *cobra.Command) error {
 		return fmt.Errorf("failed to set persistent flags: %w", err)
 	}
 
+	bindEnvVars()
 	return nil
+}
+
+// bindEnvVars wires each kuke-init viper key to its `KUKE_INIT_*`
+// environment variable. Without these calls applyServerConfiguration's
+// envSet check skips the YAML write yet viper has no env binding to read
+// the value back from — so the operator-exported value is recognised as
+// a gate but its content is silently dropped, leaving the flag default
+// to win. Mirrors the BindEnv block in cmd/kukeond/kukeond.go's
+// bindEnvVars and the KUKEON_* binds in cmd/kuke/kuke.go's loadConfig.
+func bindEnvVars() {
+	for _, v := range []config.Var{
+		config.KUKE_INIT_REALM,
+		config.KUKE_INIT_SPACE,
+		config.KUKE_INIT_KUKEOND_IMAGE,
+		config.KUKE_INIT_NO_WAIT,
+		config.KUKE_INIT_FORCE_REGENERATE_CNI,
+		config.KUKE_INIT_SERVER_CONFIGURATION,
+	} {
+		_ = v.BindEnv()
+	}
 }
 
 func setFlags(cmd *cobra.Command) error {
--- a/cmd/kuke/init/init_test.go
+++ b/cmd/kuke/init/init_test.go
@@ -28,6 +28,7 @@ import (
 	"unsafe"
 
 	"github.com/eminwux/kukeon/cmd/config"
+	"github.com/eminwux/kukeon/cmd/kuke"
 	initpkg "github.com/eminwux/kukeon/cmd/kuke/init"
 	"github.com/eminwux/kukeon/cmd/types"
 	"github.com/eminwux/kukeon/internal/controller"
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #400 Review — `fix: bind KUKE_INIT_* env vars in cmd/kuke/init`**

Targeted fix that matches the bug class — `BindPFlag` without `BindEnv` made `applyServerConfiguration`'s `envSet` gate look populated while `viper.GetString` returned empty, silently dropping the operator value. The new `bindEnvVars()` mirrors `cmd/kukeond/kukeond.go:175-189` exactly (same `_ = v.BindEnv()` idiom), and the test simplification is genuinely stronger: `TestApplyServerConfigurationEnvOverridesConfig` now exercises the production wiring path (`kuke.LoadConfig()` → `NewInitCmd()` → `setupInitCmd()` → `bindEnvVars()`) instead of stubbing the bindings in test setup, so removing either call from production trips the regression guard. The `t.Setenv` → `LoadConfig` ordering note in the new comment block is the right callout — `loadConfig`'s empty-default seeding would otherwise stamp the runtime default on top of the env binding.

No sensitive-data, dead-code, or convention issues; all checks green; commit signed. Test plan items all ticked including `make dev-init` daemon-parity.

LGTM.

PR #402, opened 2026-05-11
fix: drain leases before snapshots in CleanupNamespaceResources

### LINKED ISSUE #401: uninstall: must complete in a single invocation — currently strands overlayfs snapshots in default realm
## Context

`kuke uninstall` must complete the full teardown in a single invocation. Today it does not: on a host where workloads ran in the `default` realm, the first `kuke uninstall` fails on overlayfs snapshots, the half-cleaned-host gate from #287 trips, and the operator has to re-run to clear `/run/kukeon`, `/opt/kukeon`, and the `kukeon` user/group. The half-cleaned-host gate is correct defensive behavior — it should never need to fire in steady state.

Operator trace:

```
kuke uninstall:
  - kukeond: stopped (pid 1521581)
  - realm "default" (namespace "default.kukeon.io"): FAILED: failed to delete realm: failed to delete realm: failed to delete containerd namespace: failed to delete namespace: namespace "default.kukeon.io" must be empty, but it still has snapshots on "overlayfs" snapshotter: failed precondition; runner purge also failed: failed to delete containerd namespace "default.kukeon.io": failed to delete namespace: namespace "default.kukeon.io" must be empty, but it still has snapshots on "overlayfs" snapshotter: failed precondition
  - /run/kukeon: skipped (realm purge failed)
  - /opt/kukeon: skipped (realm purge failed)
  - user "kukeon": skipped (realm purge failed)
  - group "kukeon": skipped (realm purge failed)
  filesystem + user/group cleanup skipped: residual containerd namespace prevented teardown
```

The `kuke-system` realm (which holds the kukeond container and its image layers) purges fine. It's `default` that strands snapshots, even though the daemon-stop step from #195/#287 fired correctly this time (`stopped (pid 1521581)`, not `absent`). Distinct from #287, where the bug was the daemon-stop step looking at the wrong PID file path — here the stop works and the snapshot drain itself is at fault.

## Proposal

Make `PurgeRealm` drain whatever is pinning the overlayfs snapshots so the first `DeleteNamespace` call succeeds. No operator re-run.

Leading hypothesis (worth verifying before coding): `internal/ctr/namespaces.go:CleanupNamespaceResources` drains in the order **images → snapshots → blobs → leases**, but leases pin snapshots — so committed snapshots referenced by an outstanding lease cannot be removed in the snapshots pass, and by the time leases are released the snapshot-drain loop has already given up. The 32-pass retry inside `cleanupSnapshotsFor` only helps for parent-child unwind, not for lease-pinned snapshots.

Two viable shapes, dev's call:

1. Reorder `CleanupNamespaceResources` to drain leases **before** snapshots (and likely before images too — leases also pin content).
2. Keep the current order but add a second snapshot-drain pass after the lease drain, on the principle that releasing a lease can unpin snapshots that the first pass couldn't touch.

Whichever lands should preserve the existing per-snapshotter walk (overlayfs / native / btrfs / zfs / devmapper / blockfile) so the multi-snapshotter contract from `KukeonKnownSnapshotters` does not regress.

## Acceptance criteria

- [ ] `kuke uninstall` completes in a single invocation on a host that has gone through `kuke init` followed by workload activity in the `default` realm (`kuke create …` of at least one cell, then delete it). `/run/kukeon`, `/opt/kukeon`, and the `kukeon` user/group are all removed by the same command — the half-cleaned-host gate from #287 does not fire.
- [ ] Root cause of the surviving overlayfs snapshots is identified in the PR description (lease pin, parent-child ordering, orphaned task, or other) — not just a "retry harder" patch.
- [ ] A regression test exercises the failing path: a namespace with at least one snapshot pinned by an outstanding lease, drained through `CleanupNamespaceResources`, ends up empty. Real containerd preferred (it's the path that fails); a fake that models lease→snapshot pinning is acceptable if the real-containerd test is impractical.
- [ ] Existing `cmd/kuke/uninstall/uninstall_test.go` and `internal/controller/uninstall_test.go` still pass.
- [ ] Smoke test in PR plan: full `make dev-init` → exercise a workload in `default` → `kuke uninstall` succeeds end-to-end in one invocation, with the half-cleaned-host gate **not** triggered.

## Notes

- Priority A because uninstall failing in one shot is a steady-state UX failure, not an edge case — operators currently have to know to re-run after seeing the gate trip.
- Related closed work: #195 (stop kukeond before purge), #193 (purge reporting purged-but-namespace-survived), #287 (daemon-stop PID file path). This is the first report where the daemon-stop fires correctly and the snapshot drain itself is at fault.
- Surfaced via `kuke uninstall` after a `make dev-init` host had user workloads exercised in the `default` realm.
- Cleanup code under review: `internal/ctr/namespaces.go` (`CleanupNamespaceResources`, `cleanupSnapshotsFor`, `cleanupSnapshotPass`) and `internal/controller/runner/purge_realm.go`.

eminwux:
## Summary

`kuke uninstall` against a default realm with prior workload activity (image pulls, container creates) was stranding overlayfs snapshots and failing with `namespace "default.kukeon.io" must be empty, but it still has snapshots on "overlayfs" snapshotter: failed precondition`. Operators had to re-run uninstall, which the half-cleaned-host gate (#287) is supposed to make unnecessary.

**Root cause.** `CleanupNamespaceResources` drained in the order images → snapshots → blobs → leases. Leases ran last, so by the time `leases.SynchronousDelete` triggered the containerd GC scheduler's `ScheduleAndWait` sweep, the snapshotter pass had already given up — but the GC sweep is exactly what's needed to reconcile metadata-only snapshot entries the image-pull path leaves behind (entries whose lease was the GC root keeping them reachable). A direct `snapshotter.Walk` + `snapshotter.Remove` pass silently skips those because the metadata-wrapped Walk drops entries with no underlying state (containerd v2 `metadata/snapshot.go`).

**Fix.** Reorder to leases → images → snapshots → blobs. Each lease delete uses `leases.SynchronousDelete`, so the GC sweep completes synchronously before the snapshot pass runs against an unpinned graph. The function-level doc explains the rationale so a future refactor that "tidies" the order back has a paper trail to read.

**Refactor.** Split the body into per-class helpers (`drainLeases`, `drainImages`, `drainBlobs`, plus the existing `cleanupSnapshotsFor`) behind a small `containerdNamespaceServices` interface (`*containerd.Client` satisfies it directly). The interface exists for the regression test — a fake modeling lease→snapshot pinning would otherwise need to stand up a real containerd, which CI's `make test` target deliberately avoids.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — all packages green including `internal/ctr` (70s), `cmd/kuke/uninstall`, `internal/controller`
- [x] `go test -run TestCleanupNamespaceResourcesDrainsLeasePinnedSnapshots -v ./internal/ctr/...` — new regression test passes; asserts both the post-drain empty state AND the call-log ordering invariant (lease.Delete before snapshotter.Walk on the pinned snapshotter)
- [x] `pre-commit run --files internal/ctr/namespaces.go internal/ctr/namespaces_internal_test.go`
- [x] `make dev-init` — daemon-parity check passes; both `kuke get realms` and `kuke get realms --no-daemon` produce the canonical two-row output (`default` Ready `/kukeon/default`, `kuke-system` Ready `/kukeon/kuke-system`)
- [ ] `make dev-init` → workload in default → `kuke uninstall` (single invocation) — deferred. The dev host has multiple live concurrent agent cells running in the `default` realm (`kukeon-pm-pick-doc-3f27d7`, `kukeon-pr-ac7f52`, `kukeon-pr-1962b5`); running the destructive uninstall scenario on this host would clobber unrelated production state. The regression test covers the ordering invariant on a fake that models the failure semantic, and the daemon-parity smoke covers the code-path-the-daemon-reads check.

Closes #401

### DIFF
--- a/internal/ctr/namespaces.go
+++ b/internal/ctr/namespaces.go
@@ -23,11 +23,24 @@ import (
 	"slices"
 
 	"github.com/containerd/containerd/v2/core/content"
+	"github.com/containerd/containerd/v2/core/images"
 	"github.com/containerd/containerd/v2/core/leases"
 	"github.com/containerd/containerd/v2/core/snapshots"
 	"github.com/containerd/containerd/v2/pkg/namespaces"
 	"github.com/opencontainers/go-digest"
 )
+
+// containerdNamespaceServices is the subset of containerd.Client used by
+// drainNamespaceResources. Extracted so a fake modeling lease→snapshot
+// pinning can drive the drain in unit tests without a real containerd —
+// real-containerd integration in this repo lives in the e2e suite, not the
+// per-package `make test` target. *containerd.Client satisfies it directly.
+type containerdNamespaceServices interface {
+	ImageService() images.Store
+	SnapshotService(name string) snapshots.Snapshotter
+	ContentStore() content.Store
+	LeasesService() leases.Manager
+}
 
 // namespaceCtx returns a context with the namespace set.
 func (c *client) namespaceCtx(namespace string) context.Context {
@@ -170,12 +183,22 @@ var KukeonKnownSnapshotters = []string{
 // CleanupNamespaceResources empties a containerd namespace so DeleteNamespace
 // (which requires the namespace to be empty) can succeed. It drains in this
 // fixed order:
-//  1. images,
-//  2. snapshots — for every snapshotter when `snapshotter == ""` (the
+//  1. leases (with leases.SynchronousDelete),
+//  2. images,
+//  3. snapshots — for every snapshotter when `snapshotter == ""` (the
 //     uninstall path's default), or just the named snapshotter when one is
 //     specified explicitly,
-//  3. blobs (content store),
-//  4. leases.
+//  4. blobs (content store).
+//
+// Leases run first because leases.SynchronousDelete triggers the containerd
+// GC scheduler's ScheduleAndWait sweep before returning. That sweep is what
+// reconciles any metadata-only snapshot entries the image-pull path left
+// behind (entries whose lease was the GC root keeping them reachable); a
+// straight Walk+Remove pass after the fact misses those because the
+// metadata snapshotter's Walk silently skips entries with no underlying
+// state (containerd v2 metadata/snapshot.go). Draining leases first lets the
+// GC sweep clear them, then the subsequent snapshotter walk handles whatever
+// the user/runtime layer still pinned directly.
 //
 // Each class logs a count summary at INFO and per-resource debug lines at
 // DEBUG, so a future "namespace not empty" failure points at the exact class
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #402 Review — `fix: drain leases before snapshots in CleanupNamespaceResources`**

Targeted fix: leases→images→snapshots→blobs reordering with a `SynchronousDelete`-driven GC sweep before the snapshot pass, plus a small `containerdNamespaceServices` interface so a fake can drive the ordering invariant in `make test`. Function-level docs spell out the GC-sweep rationale so a future "tidying" refactor has a paper trail. Regression test in `internal/ctr/namespaces_internal_test.go` asserts both the post-drain empty state and the call-log ordering (every `lease.Delete` precedes the first `snapshotter.Walk` on the pinned snapshotter), so a re-inversion would fail both assertions. No bugs, no secret/PII surface, no dead symbols (every new helper is wired into `drainNamespaceResources`; the interface is satisfied by `*containerd.Client` directly), all callers of `CleanupNamespaceResources` unchanged. License header + SPDX present on the new file; sole commit `f1bed4f` GPG/SSH-verified. The deferred destructive smoke (`make dev-init` → workload → `kuke uninstall`) is documented as host-constrained — multiple live concurrent agent cells in `default` would be clobbered — and the daemon-parity arm of `make dev-init` is checked, which is the path the daemon reads.

LGTM — ready to merge.

PR #408, opened 2026-05-11
fix: transition cell to Failed when a container fails to start

### LINKED ISSUE #407: Cell stays in `Unknown` indefinitely when a container fails to start; should transition to terminal `Error` state
## Summary

When `runc create` fails for any container in a cell, the cell is left in an indeterminate state forever: the failed container has a spec but no task object, the surviving container(s) keep running, and the reconciler never sweeps the cell because its auto-delete predicate only fires on `state=Stopped`. The `failed to start container from CellDoc` error is logged but never fed back into the cell state machine.

## Reproduction

A real instance observed today on the host fleet — `kukeon-pr-reviewer-54bb2f`. The cell's `work` container failed at startup because a host bind-mount source (`/opt/kukeon/cache/sbsh/amd64/sbsh`) didn't exist at the time. `kuke get cells` shows the cell stuck in `Unknown`:

```
NAME                       STATE    CGROUP
kukeon-pr-reviewer-54bb2f  Unknown  /kukeon/default/default/default/kukeon-pr-reviewer-54bb2f
```

Daemon log evidence (`sudo kuke daemon log`):

```
07:26:12.875 ERROR failed to create task id=...kukeon-pr-reviewer-54bb2f_work
  err="failed to create shim task: OCI runtime create failed: runc create failed:
       unable to start container process: error during container init:
       failed to fulfil mount request:
       open /opt/kukeon/cache/sbsh/amd64/sbsh: no such file or directory"
07:26:12.875 ERROR failed to start container from CellDoc id=...kukeon-pr-reviewer-54bb2f_work cell=kukeon-pr-reviewer-54bb2f ...
```

From that point on, every reconcile cycle (~30 s) logs:

```
INFO container state determined via TaskStatus container=root  taskStatus=running internalState=1
INFO failed to get container state via TaskStatus  container=work
     error="task not found: no running task found: task ..._work not found"
INFO reconcile ok cells_scanned=N cells_updated=0 cells_deleted=0
```

The root (busybox) container keeps running. The cell is never updated and never swept. Hours later it is still consuming PIDs, cgroup, and netns resources for nothing.

For comparison, when both containers do start successfully and `work` later exits cleanly, the existing happy path works correctly:

```
07:32:51 INFO container state determined via TaskStatus container=work taskStatus=stopped internalState=2
07:32:51 INFO killed container         id=..._work
07:32:51 INFO killed root container    id=..._root
07:33:17 INFO reconcile: auto-deleting cell ... state=2
```

The state machine has no equivalent transition for "container creation failed", so a startup-failed cell never reaches `state=2` and is never cleaned up.

## Root cause

Two gaps:

1. The `failed to start container from CellDoc` error path doesn't update cell state. It logs and returns.
2. The state observer treats containerd's `task not found` as "indeterminate, skip" rather than distinguishing "task never existed" (terminal) from "task is briefly between states" (transient).

## Proposed fix

Introduce a terminal `Error` state with the following contract:

- **Trigger:** any container in the cell fails to be created or started (OCI/runc errors, image pull failures, CNI errors, missing bind-mount sources — *all* startup-path errors). No retry/backoff for now; first failure is terminal.
- **On entry:** kill any sibling containers in the cell (e.g. the root container that did start), so the cell holds no running processes. Retain: cell metadata, the failed container's spec, the last error message, timestamps, captured logs.
- **Stickiness:** the cell remains in `Error` until the user runs `kuke delete cell <name>`. `--rm` does **not** auto-clean an `Error` cell — `--rm` only takes effect on a *successful* run that subsequently exits cleanly. (This is a deliberate departure from `docker run --rm` semantics; rationale: preserve diagnostic surface on failed startup.)
- **Reconciler:** the existing auto-delete predicate stays scoped to `state=Stopped` only. It must never sweep `state=Error`.
- **CLI:** `kuke delete cell <name>` works on `Error` cells without any force flag.

## Out of scope (for this issue)

- Retry/backoff for transient failures.
- TTL-based GC of long-stale `Error` cells.
- `--force` flag on delete (not planned).

eminwux:
## Summary
- A non-root container that fails to create or start (OCI/runc, image pull, CNI, missing bind-mount) used to leave the cell in `Unknown` forever — daemon reconcile never swept it because the auto-delete predicate only fires on `Stopped`, so `Spec`, cgroup, PIDs, and netns stayed pinned indefinitely (issue #407 repro: `kukeon-pr-reviewer-54bb2f`).
- `runner.StartCell` and `runner.StartContainer` now arm a deferred cleanup once they enter the destructive provisioning phase: any error past that point flips the cell to `CellStateFailed`, runs `KillCell` so any sibling that *did* start (e.g. the root) is torn down, then persists the Failed state. The original startup error returns to the caller unmodified.
- The reconciler now treats `CellStateFailed` as sticky (`cellStateIsSticky`): the populate pass still refreshes per-container statuses for `kuke get cell -o yaml`, but the cell-level state is never re-derived. Combined with `cellStateAutoDeleteTriggers` losing its `Failed` entry, `kuke run --rm` will no longer auto-clean a Failed cell — that's a deliberate departure from `docker run --rm` so the diagnostic surface survives until the operator runs `kuke delete cell`. `kuke delete cell` already accepts any state, so no `--force` is needed.

## Test plan
- [x] `make test` — full Go unit/integration suite (pkg `internal/ctr` runs ~70s).
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make dev-init` — daemon-parity tail printed both `kuke get realms` and `kuke get realms --no-daemon` showing `default` + `kuke-system` both `Ready` under `/kukeon/...`.
- [x] New / updated unit tests:
  - `TestCellStateIsSticky` — new predicate covers Failed-sticky reconciler behavior.
  - `TestCellStateAutoDeleteTriggers` — pinned Failed → `false` (was `true`).
  - `TestShouldAutoDeleteCell/blocked_on_failed_state_per_issue_407` — Failed no longer triggers `--rm` auto-delete.
  - `TestShouldWindDownCell/blocked_on_failed_state_per_issue_407` — Failed no longer triggers wind-down.
  - `TestDeleteCell_FailedCellDeletesWithoutForce` — controller-level confirmation that the explicit `kuke delete cell` path works on a Failed cell with no extra flag.
- [ ] Integration test that exercises a real `runner.StartCell` failure path through the `ctr.Client` fake — skipped; the helper relies on `r.KillCell`, which in turn does CNI detach + cgroup reads against a live containerd. Pure-predicate coverage above plus the smoke test on `dev-init` cover the wiring; a heavier fake is a separate follow-up if reviewers want it.
- [ ] `pre-commit run` — not run; `pre-commit` binary is not installed on this dev host. CI hooks remain authoritative.

Closes #407

### DIFF
--- a/internal/controller/delete_cell_test.go
+++ b/internal/controller/delete_cell_test.go
@@ -192,6 +192,49 @@ func TestDeleteCell_SuccessfulDeletionWithContainers(t *testing.T) {
 	}
 }
 
+// TestDeleteCell_FailedCellDeletesWithoutForce locks down the issue #407
+// contract that `kuke delete cell <name>` accepts a cell in the terminal
+// CellStateFailed state without any --force or extra flag. Stickiness of
+// Failed is enforced by the reconciler (cellStateAutoDeleteTriggers excludes
+// Failed); the operator's explicit delete is the one and only escape hatch.
+func TestDeleteCell_FailedCellDeletesWithoutForce(t *testing.T) {
+	mockRunner := &fakeRunner{}
+	existingCell := buildTestCell("failed-cell", "test-realm", "test-space", "test-stack")
+	existingCell.Status.State = intmodel.CellStateFailed
+	existingCell.Spec.Containers = []intmodel.ContainerSpec{
+		{ID: "work", Image: "alpine:latest"},
+	}
+
+	deleteCalled := false
+	mockRunner.GetCellFn = func(_ intmodel.Cell) (intmodel.Cell, error) {
+		return existingCell, nil
+	}
+	mockRunner.ExistsCgroupFn = func(_ any) (bool, error) {
+		return true, nil
+	}
+	mockRunner.ExistsCellRootContainerFn = func(_ intmodel.Cell) (bool, error) {
+		return true, nil
+	}
+	mockRunner.DeleteCellFn = func(_ intmodel.Cell) error {
+		deleteCalled = true
+		return nil
+	}
+
+	ctrl := setupTestController(t, mockRunner)
+	cell := buildTestCell("failed-cell", "test-realm", "test-space", "test-stack")
+
+	result, err := ctrl.DeleteCell(cell)
+	if err != nil {
+		t.Fatalf("DeleteCell on Failed cell returned error: %v", err)
+	}
+	if !deleteCalled {
+		t.Error("runner.DeleteCell was not invoked — controller refused to delete a Failed cell")
+	}
+	if !result.MetadataDeleted {
+		t.Error("expected MetadataDeleted to be true on Failed cell deletion")
+	}
+}
+
 func TestDeleteCell_ValidationErrors(t *testing.T) {
 	tests := []struct {
 		name      string
--- a/internal/controller/runner/refresh.go
+++ b/internal/controller/runner/refresh.go
@@ -381,6 +381,29 @@ func (r *Exec) refreshContainerStatus(cell intmodel.Cell, containerSpec *intmode
 func (r *Exec) ReconcileCell(cell intmodel.Cell) (intmodel.Cell, ReconcileOutcome, error) {
 	originalStatus := cell.Status
 
+	// CellStateFailed is terminal (issue #407): once a cell has been marked
+	// Failed by a startup-path error in StartCell / StartContainer, the
+	// reconciler must keep the state sticky so a subsequent populate that
+	// reads back "no task" doesn't quietly downgrade Failed to Unknown or
+	// Stopped — and so cellStateAutoDeleteTriggers (which excludes Failed)
+	// cannot be bypassed via a re-derivation tick. Container statuses are
+	// still refreshed below so `kuke get cell -o yaml` shows the latest
+	// per-container view, but the cell-level state is left alone until the
+	// operator runs `kuke delete cell`.
+	if cellStateIsSticky(originalStatus.State) {
+		if err := r.populateCellContainerStatuses(&cell); err != nil {
+			r.logger.DebugContext(r.ctx, "populate container statuses failed",
+				"cell", cell.Metadata.Name, "error", err)
+		}
+		updated := !containerStatusesEqual(originalStatus.Containers, cell.Status.Containers)
+		if updated {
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #408 Review — `fix: transition cell to Failed when a container fails to start`**

Addresses issue #407 cleanly: provisioning failures past the destructive-recreate phase now flip the cell to a terminal `CellStateFailed` via a deferred `markCellFailedAfterStartupFailure`, and the reconciler's new `cellStateIsSticky` predicate preserves that state across ticks instead of re-deriving it back to `Unknown`/`Stopped`. Predicates are exercised by pure unit tests, and `kuke delete cell` is locked down on a Failed cell without `--force` (`TestDeleteCell_FailedCellDeletesWithoutForce`). The `provisionStarted` gate is correctly placed in both `StartCell` and `StartContainer` — validation errors and the idempotent-skip return predate the flip, so a healthy cell can't get stamped Failed by a missing-name error. The dev-init smoke ran with the expected daemon-parity tail, no sensitive data in error strings, commit signed.

- **Comment accuracy** `internal/controller/runner/start.go:135`: the helper's leading comment says "any container that did start (e.g., the root container) is torn down". For the `StartContainer` defer path, the root container plus *all* peers may already be running (this is a partial-recreate on an existing Ready cell), so KillCell tears down more than just "siblings of the failed start". The behavior is correct; the comment just understates the blast radius. Worth tightening if you touch it.
- **`--rm` race window**: `markCellFailedAfterStartupFailure` calls `KillCell` (which persists `Stopped`) and only then reloads + stamps `Failed`. For an AutoDelete+Ready cell that fails a subsequent `StartContainer`, a reconcile tick landing between those two writes would observe `Stopped` + `ReadyObserved=true` and trip `autoDeleteCell` before the defer stamps Failed — the user loses the diagnostic surface this PR promises. Window is narrow (two `UpdateCellMetadata` calls back-to-back) and the outcome is what `--rm` would do anyway, so non-blocking, but the contract you stated in the PR body is "`kuke run --rm` will no longer auto-clean a Failed cell" — that holds only outside this window. If you want it airtight, stamping Failed *before* KillCell would close it.

LGTM.

Issue #410, opened 2026-05-11
kuketty: attach-socket RPC server (phase 1b)


eminwux:
## Context

Phase 1 (#165) lands the standalone `kuketty` binary, packs it in the kukeond image, swaps the OCI injection wrapper from `sbsh terminal` to `kuketty`, defines a kukeon-side terminal metadata file (`pkg/kuketty/metadata.Metadata`) bind-mounted via the OCI runtime spec, and PTY-execs the wrapped workload. Phase 1 explicitly defers the attach-socket *server*: kuketty creates the socket inode at the path declared in the metadata file but does not yet serve the JSON-RPC + SCM_RIGHTS terminal protocol that `kuke attach` consumes via `github.com/eminwux/sbsh/pkg/attach`. Phase 1 leaves `kuke attach` against kuketty-wrapped containers temporarily broken, with the e2e attach test skipped for a named reason at `e2e/e2e_kuke_attach_test.go:131`.

This issue lands the server side, restoring end-to-end `kuke attach` against the kuketty injection path — and, on top of the original sbsh-as-library re-scope, also pivots the on-disk contract to consume sbsh's public types directly. The earlier `kuketty.Metadata` schema + adapter package go away; kukeond renders `api.TerminalDoc` and kuketty consumes it raw.

The earlier reimplement-from-scratch direction and the subsequent adapter-shim direction are preserved in the comment thread for traceability; the body below reflects the consume-as-library + consume-as-types scope, which fits one reviewable PR.

## Proposal

Wire `kuketty` to sbsh's newly-exported `github.com/eminwux/sbsh/pkg/terminal/server` facade (eminwux/sbsh#202, sbsh v0.11.0+) and switch the kukeond → kuketty on-disk contract from `pkg/kuketty/metadata.Metadata` to sbsh's `api.TerminalDoc`. Concretely:

- Bump `go.mod` to `github.com/eminwux/sbsh@v0.11.0` (the release that ships the server facade — eminwux/sbsh#202).
- **Delete `pkg/kuketty/metadata.go` + `pkg/kuketty/metadata_test.go`.** The schema package goes away. No adapter package is introduced — kuketty consumes `api.TerminalDoc` directly.
[...]

Issue #291, opened 2026-05-05
kuketty: retire sbsh code paths (phase 5)


eminwux:
## Context

Phases 1, 1b, 2, 3, 4 (#165, #410, #288, #289, #290) land kuketty with full feature parity vs. the legacy sbsh path. Phase 1 already removed the OCI injection wrapper that bind-mounted the sbsh binary into each attachable container; as a result `internal/ctr/sbsh_cache.go`, its companion `sbsh_cache_test.go`, and the `ResolveSbshCachePath` entry on the `internal/ctr.Client` interface are dead code today — no live caller remains. This phase deletes them and finishes the retirement of the sbsh binary distribution mechanism.

**Two retirements, distinguished:**

- **sbsh binary distribution mechanism — retired here.** `internal/ctr/sbsh_cache.go`, the `cache/sbsh/<arch>/sbsh` host-path layout, and the `ResolveSbshCachePath` interface entry on `internal/ctr.Client` (plus any mocks) all go away. No per-arch sbsh artifact cache; multi-arch resolves into the kukeond container image's multi-arch story.
- **sbsh Go module dependency — retained.** `cmd/kuke/attach`, `cmd/kuke/run`, and `cmd/kuke/shared/attach_exit.go` consume `github.com/eminwux/sbsh/pkg/attach` on the client side. `cmd/kuketty/main.go` consumes `pkg/terminal/server`, `pkg/api`, `pkg/builder`, and `pkg/discovery` on the server side. Both ride into the kukeond image as Go packages; neither is bind-mounted from the host.

## Proposal

Remove the dead sbsh-binary-distribution surface from production code, build paths, and packaging. Keep the sbsh Go-module dependency intact. Update or close any open issues whose framing assumed the old distribution mechanism.

## Acceptance criteria

- [ ] No `sbsh` binary is bind-mounted, downloaded, cached, or distributed by kukeon.
- [ ] `internal/ctr/sbsh_cache.go` and `internal/ctr/sbsh_cache_test.go` deleted.
- [ ] `ResolveSbshCachePath` removed from the `internal/ctr.Client` interface and from any mocks/fakes that satisfy it.
- [ ] No `sbsh`-distribution references remain in `internal/ctr/` or in build/packaging paths. Comments that explain historical context are fine; live code references are not.
[...]

Issue #290, opened 2026-05-05
kukeon: wire cell profile to sbsh builder (phase 4)


eminwux:
## Context

Phase 1 (#165) lands `kuketty` as a standalone Go binary packaged in the kukeond container image, with the OCI injection swap and PTY exec of the wrapped workload. Phase 1b (#410) lands the attach-socket RPC server by consuming sbsh's `pkg/terminal/server` facade and pivots the on-disk contract to `api.TerminalDoc`.

The earlier two-sided framing for this phase (kukeond renders profile content into the metadata file → kuketty applies it at PTY start) collapses with phase 1b. sbsh already exposes the profile contract publicly today (`pkg/api/profile.TerminalProfileDoc/TerminalProfileSpec`), the profile YAML loader (`pkg/discovery.LoadProfilesFromDir` / `FindProfileByNameInDir`), and the builder option set (`pkg/builder.WithProfile` / `WithProfilesDir`). When kukeond calls `builder.BuildTerminalSpec(WithProfile(name), WithProfilesDir(dir), ...)`, sbsh loads the YAML, evaluates the profile contract, and stamps `Prompt`, `SetPrompt`, and `Stages` into the returned `TerminalSpec`. kuketty consumes that spec via `server.Serve`, which applies prompt + onInit stages natively.

Net: there is **no kukeon-side profile interpreter to write**, and no kuketty-side application logic to add — both ends are delivered by sbsh once the wiring lands.

## Proposal

In `internal/ctr/attachable.go` (or its `TerminalDoc` render helper), resolve the cell's profile name and the host-side profiles directory from cell config, and pass them to `pkg/builder.BuildTerminalSpec`:

```go
builder.BuildTerminalSpec(
    // ... phase 1b options ...
    builder.WithProfile(cellProfileName),
    builder.WithProfilesDir(profilesHostDir),
)
```

No `TerminalProfile` loader lives in kukeon. No prompt or onInit-stage rendering lives in kukeon. Profile resolution is delegated end-to-end to sbsh.

If the cell does not configure a profile, the builder options are omitted and `Spec.Prompt` / `Spec.Stages` stay zero. sbsh's runner then runs the workload without an onInit script or a prompt override, matching the no-profile path.

## Acceptance criteria

- [ ] kukeon contains no `TerminalProfile` YAML loader, no profile-name resolver, and no prompt/onInit rendering logic — `pkg/discovery` + `pkg/builder` do all of it.
[...]

Issue #289, opened 2026-05-05
kukeon: render log file fields in terminal spec (phase 3)


eminwux:
## Context

Phase 1 (#165) lands `kuketty` as a standalone Go binary packaged in the kukeond container image, with the OCI injection swap and PTY exec of the wrapped workload. Phase 1b (#410) lands the attach-socket RPC server by consuming sbsh's `pkg/terminal/server` facade and pivots the on-disk contract to `api.TerminalDoc`. As a side-effect of phase 1b, log-file behavior is delivered natively by sbsh's runner (`Spec.LogFile`, `Spec.LogFileMode`, `Spec.LogFileGID`); the kuketty-side reimplementation earlier versions of this issue proposed is no longer needed.

The remaining work is **kukeond-side rendering** — populating the log fields on the `TerminalSpec` from the cell's container-tty config when kukeond builds the bind-mounted `TerminalDoc`.

## Proposal

In `internal/ctr/attachable.go` (or its `TerminalDoc` render helper), resolve the cell's container-tty log configuration and pass the corresponding builder options when constructing the `TerminalSpec`:

```go
builder.BuildTerminalSpec(
    // ... phase 1b options ...
    builder.WithLogFile(logFile),
    builder.WithLogFileMode(logMode),
    builder.WithLogFileGID(logGID),
)
```

sbsh's runner then writes its runtime logs to the rendered path with the rendered mode/GID once `Serve` spawns the PTY. No kuketty source changes — kuketty is `cmd/kuketty/main.go` reading the doc and handing `&doc.Spec` to `server.Serve`; the additional spec fields flow through unchanged.

## Acceptance criteria

- [ ] When the cell config carries a log-file path (and optional mode/GID), the rendered `TerminalDoc.Spec.LogFile`, `Spec.LogFileMode`, and `Spec.LogFileGID` are populated correctly in the bind-mounted JSON.
- [ ] When the cell config does not carry a log-file path, the rendered fields stay zero and sbsh's runner skips log writer setup.
- [ ] Inside the container, sbsh's runner writes its runtime log at the rendered path with the rendered mode/GID — verified via the existing log-related e2e test on the kuketty path with no source changes beyond build/path adjustments.

## Out of scope

- Kuketty-side log writer code — sbsh's runner owns this end of the contract (delivered through phase 1b).
- Capture-file rendering — phase 2 (#288).
- Profile rendering — phase 4 (#290).

## Notes

[...]

PR #411, opened 2026-05-11
feat: introduce kuketty and swap sbsh on OCI injection (phase 1)

### LINKED ISSUE #165: kuketty: standalone binary, swap sbsh in OCI injection (phase 1)
## Context
Kukeon's container terminal wrapping (PTY, attach socket, capture, profile)
is delegated to an external static `sbsh` binary that gets bind-mounted
into each container at `/.kukeon/bin/sbsh` and invoked via OCI args wrap
(see `internal/ctr/attachable.go` — `withAttachableMounts`,
`withAttachableArgsWrap`). This adds an external dependency to the kuke
distribution and complicates the multi-arch story (kuke for arch X *and*
sbsh for arch X).

## Proposal
Introduce `kuketty` as a **separate standalone Go binary** — its own
`cmd/kuketty` entrypoint and Makefile target, **not** argv[0]-dispatched
from `kuke`. The binary is packed into the `kukeond` container image so
it ships alongside the daemon; the kukeond image is kuketty's
distribution channel.

`kuketty` reimplements the subset of `sbsh terminal` that kukeon invokes
during OCI injection. Per the user's redirects:

- No `terminal` subcommand on kuketty's CLI — kuketty is invoked
  directly.
- **No CLI flags for runtime configuration.** Instead, kukeond renders
  a **terminal metadata file** (JSON or YAML — dev's choice) per
  attachable container, **bind-mounts that file into the container via
  the OCI runtime spec**, and kuketty reads the bind-mounted file on
  startup. The architectural arrow is one-way: kukeond produces,
  kuketty consumes.
- **Profile resolution moves out of kuketty into kukeond.** kuketty
  does not own a `TerminalProfile` interpreter, does not resolve
  profile names, and does not own a profiles directory. kukeond takes
  the profile inputs and renders the resulting prompt + onInit content
  directly into the terminal metadata file. kuketty's job is to apply
  the rendered fields, not to interpret profile semantics.

The OCI injection point in `internal/ctr/attachable.go` is updated to
(a) render and stage the kuketty terminal metadata file, (b) add a
bind-mount entry for that file in the OCI runtime spec for the
container, and (c) rewrite the args wrapper to invoke kuketty — which
then reads the bind-mounted file — in place of the sbsh flag wrapper.

This is **phase 1** — skeleton + metadata-file rendering + OCI
bind-mount plumbing + OCI injection swap + PTY exec of the wrapped
workload + creation of the attach-socket inode at the
metadata-declared path. **Serving the attach-socket RPC protocol** —
the JSON-RPC + SCM_RIGHTS surface that `kuke attach` consumes via
`github.com/eminwux/sbsh/pkg/attach` — is deferred to phase 1b
(#410); phase 1 leaves the existing e2e attach test skipped on the
kuketty path with a named reason until phase 1b lands. Capture-file,
log-file, and profile-rendering behaviors are declared in the schema
but implemented in follow-up phases (see "Out-of-scope follow-ups").
Final sbsh code-path retirement is also a follow-up.

### Terminal metadata file
The metadata file carries everything kuketty needs to run a session.
Phase 1 defines the schema and honors the fields that drive the PTY
spawn and the attach-socket *inode*. Other fields, including serving
the attach-socket RPC, are declared in the schema in phase 1 but
their behavior lands in later phases:

| Field group                          | Honored in phase |
|--------------------------------------|------------------|
| run path, socket path (inode created) | phase 1 (#165)   |
| attach-socket RPC server (socket served) | phase 1b (#410) |
| capture-file path                    | phase 2 (#288)   |
| log-file path                        | phase 3 (#289)   |
| prompt, onInit (rendered)            | phase 4 (#290)   |

Exact field names, file format (JSON or YAML), on-container mount
path, and host-side staging location are dev's call within phase 1.

## Acceptance criteria
- [ ] New `cmd/kuketty` Go entrypoint with its own Makefile target
      produces a standalone `kuketty` binary.
- [ ] The kukeond container image (`Dockerfile`) packs the `kuketty`
      binary alongside the daemon binary.
- [ ] OCI injection (`internal/ctr/attachable.go`) renders a terminal
      metadata file per attachable container, adds a bind-mount entry
      for the file in the OCI runtime spec, and invokes kuketty in
      place of `sbsh terminal` — no sbsh wrapper remains on the
      injection path.
- [ ] `kuketty` reads its runtime configuration from the bind-mounted
      metadata file (no runtime CLI flags), runs the wrapped command
      under a PTY, and creates the attach-socket inode at the
      metadata-declared path. Serving the attach-socket RPC protocol
      (so `kuke attach` works end-to-end) is deferred to phase 1b
      (#410); the existing e2e attach test is skipped with a named
      reason on the kuketty path until phase 1b lands.
- [ ] `make dev-init` smoke passes against the kuketty path — the
      kuketty-wrapped container starts, the workload runs under a
      PTY, and the attach-socket inode is present at the
      metadata-declared path. End-to-end `kuke attach` smoke is
      deferred to phase 1b.

## Out-of-scope follow-ups
- Phase 1b — wire kuketty to sbsh's exported terminal RPC server
  facade (eminwux/sbsh#202, sbsh v0.11.0+) for e2e `kuke attach`
  parity: #410
- Phase 2 — capture-file behavior: #288
- Phase 3 — log-file behavior: #289
- Phase 4 — kukeond renders profile into terminal metadata, kuketty
  applies rendered prompt + onInit: #290
- Phase 5 — retire sbsh code paths and resolve multi-arch story: #291

## Notes
- Why a separate binary instead of argv[0]-dispatch from `kuke`? Per
  dev's 2026-05-02 design note: kuketty wraps every attachable
  container's command (per-container PID 1, long-running). Folding
  it into the kuke+kukeond multi-call binary would force every
  attachable container to load the full transitive `init()` of
  containerd client / gRPC / protobuf / runtime types just to bridge
  a pty. Standalone keeps the import set bounded — stdlib + a small
  pty helper in phase 1, and (from phase 1b onwards) the
  `github.com/eminwux/sbsh/pkg/terminal/server` facade for the
  attach-socket RPC server. Per-process RSS and startup time stay
  bounded by sbsh's small dep closure (`slog`, `golang.org/x/sys`,
  sbsh internal helpers); the kuke+kukeond containerd / gRPC /
  protobuf closure stays out of every attachable container.
- Why a metadata file via OCI bind-mount instead of flags? Per user
  redirect: a declarative metadata file keeps kuketty's surface
  minimal (single parser, no CLI sprawl) and centralizes the kukeond
  → kuketty contract in one schema. Delivering it via the OCI mount
  mechanism kukeon already uses keeps the injection path uniform
  with the existing kuketty-binary mount.
- Why kukeond owns profile rendering instead of kuketty owning
  profile resolution? Per user redirect: kuketty stays stateless,
  with no awareness of the TerminalProfile interpreter or a profiles
  directory. The kukeond → kuketty contract is the rendered metadata
  file alone.
- Why split phase 1 from phase 1b? Phase 1's plumbing surface
  (Makefile target, image packing, OCI bind-mount of the metadata
  file, args-wrapper rewrite, PTY exec, attach-socket inode) is one
  PR-sized review unit. Phase 1b wires kuketty to the sbsh-exported
  terminal RPC server facade (sbsh v0.11.0+, eminwux/sbsh#202) plus
  a small `kuketty.Metadata` → `api.TerminalSpec` adapter — small
  enough as its own PR but distinct enough to keep separately
  reviewable.
- Multi-arch: kuketty's multi-arch story is the kukeond container
  image's multi-arch story. No separate per-arch artifact cache is
  needed; the closed #67 thread no longer applies.
- #68, #76, #114, #119 reference sbsh in their bodies — re-verify
  during phase 5; they should keep working since the socket /
  capture / log path contract is preserved.

eminwux:
## Summary
- Introduce `kuketty`, a kukeon-owned standalone terminal wrapper that replaces `sbsh` on the OCI injection path. New `cmd/kuketty` entrypoint with its own Makefile target, packed into the `kukeond` container image alongside the daemon binary; the kukeond image is kuketty's distribution channel (no separate per-arch artifact cache).
- Define the kukeond → kuketty contract as a JSON terminal-metadata document (`pkg/kuketty.Metadata`) the daemon renders per attachable container. kuketty exposes **no runtime CLI flags** — everything (runPath, socket path/mode/gid, declared capture/log paths, pre-rendered shell prompt + onInit) flows through the bind-mounted metadata file.
- Rewrite `internal/ctr/attachable.go`: drop the multi-flag `sbsh terminal …` wrapper for `[kuketty, --, originalArgs...]`, swap the binary mount (`/.kukeon/bin/sbsh` → `/.kukeon/bin/kuketty`), and add an RO file bind-mount for the per-container metadata document at `/.kukeon/kuketty/metadata.json`. Drop the `SocketGID` / `CaptureGID` / `LogFileGID` / `UseProfile` fields from `AttachableInjection` — the metadata file carries all of those now.
- Runner side (`internal/controller/runner/attachable.go`): stop calling `ResolveSbshCachePath`, stop writing `profile.yaml`, render the kuketty metadata atomically into `<containerDir>/kuketty-metadata.json` instead, and lazily stage `/bin/kuketty` (where the kukeond image places it) to `<RunPath>/bin/kuketty` so runc has a host-visible bind-mount source. The `--no-daemon` fallback walks `os.Executable` siblings before failing with a debuggable error.
- Phase-1 scope per #165 AC: kuketty creates the attach-socket inode at the metadata-declared path but does not serve the RPC. The existing `TestKuke_AttachDetach_KeepsTaskRunning` e2e is `t.Skip`'d with a named reason that points to phase 1b (#410), per the AC's "skipped with a named reason on the kuketty path until phase 1b lands" wording.
- Delete `internal/controller/runner/sbsh_profile.go` (+ its test): kuketty does not read sbsh's `TerminalProfile` YAML; prompt + onInit are stamped directly into the metadata document by the new `shellSpecFromTty` helper. Phase 5 (#291) will retire the residual sbsh code paths under `internal/ctr/sbsh_cache.go` and the dormant `ResolveSbshCachePath` interface method.

### Design calls (non-obvious decisions worth a reviewer's eye)
- **JSON, not YAML, for the metadata file.** Keeps kuketty's import set to stdlib + `creack/pty` (already a direct dep); a YAML parser would bloat the per-process RSS the issue body explicitly calls out as a sizing driver at attachable-container scale.
- **`--` is the only positional contract on kuketty's CLI.** Spirit-matches the issue's "no CLI flags for runtime configuration" redirect while keeping the wrapper able to tell its own argv from the workload's. Wrapper failure mode for a missing/empty `--` is `exit 64` (BSD `EX_USAGE`); pre-workload failures (metadata parse, PTY setup, socket listen) are `exit 70` (`EX_SOFTWARE`); workload's exit code propagates verbatim via `*exec.ExitError`.
- **Schema declares phase 2/3/4 fields up front.** `Capture.Path`, `Log.Path`, `Shell.Prompt`, `Shell.OnInit` are written by the phase-1 renderer but kuketty does not yet honor them — declared so the daemon's render path stays stable and #288 / #289 / #290 don't need a schema bump.
- **Lazy binary staging in `attachableBuildOpts`, not at daemon init.** Avoids a new init-ordering hop; idempotent (re-stages only when the destination is missing, zero-byte, or non-executable). Race-safe via tmp-file + atomic rename.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (the project's named CI test target; full unit-test suite incl. the new `pkg/kuketty`, `cmd/kuketty`, and runner tests + the rewritten `internal/ctr/attachable_test.go` covering the three-mount + `[kuketty, --, originalArgs...]` wrapper invariants, plus negative assertions banning every leftover sbsh flag).
- [x] `make dev-init` — the project's required local smoke. Daemon-parity tail matches CLAUDE.md exactly:
  ```
  ```
  Both `kuke get realms` and `kuke get realms --no-daemon` produce identical output.
- [x] Verified the built `kukeon-local:dev` image packs `/bin/kuketty` (2.7 MB) alongside `/bin/kuke` / `/bin/kukeond` (`docker run --rm --entrypoint ls kukeon-local:dev -la /bin/kuketty /bin/kuke /bin/kukeond`).
- [ ] End-to-end `kuke attach` smoke against the kuketty path — **deferred to phase 1b (#410)** per the parent issue's AC. `TestKuke_AttachDetach_KeepsTaskRunning` is skipped with a named reason that names #410; re-enables when #410 lands the JSON-RPC + SCM_RIGHTS server.

Closes #165


eminwux:
**PR #411 Review — `feat: introduce kuketty and swap sbsh on OCI injection (phase 1)`**

Solid phase-1 cut. The three-mount + `[kuketty, --, originalArgs...]` invariants are well locked down by the rewritten `internal/ctr/attachable_test.go`, including a negative banlist for every leftover sbsh flag — exactly the regression guard this rewrite needs. Single signed commit, no secret/PII exposure, clean removal of `sbsh_profile.go` with no dangling references, and the dev-init smoke is in the test plan with the daemon-parity tail per CLAUDE.md.

One blocking item:

- **Stale fallback contract** `internal/controller/runner/attachable.go:306-326`: `resolveKukettyBinary`'s docstring lists step "3. $PATH — last-resort fallback for whoever installed kuketty out-of-band" and the inline comment claims "$PATH lookup is intentionally last and skipped silently on miss", but the body never calls `exec.LookPath` — it only appends the literal string `"$PATH/kuketty"` to the `tried` slice and returns. An operator with kuketty on PATH (but not at `/bin/kuketty` or a sibling of `os.Executable`) gets a "not found" error naming `$PATH/kuketty` even though PATH was never consulted. Pick one: implement the lookup (`if p, err := exec.LookPath("kuketty"); err == nil { return p, nil }`) before the error return, or drop docstring step 3, the inline comment, and the `"$PATH/kuketty"` error fragment.

Non-blocking nit:

- **PTY stdout drain not synced with `cmd.Wait()`** `cmd/kuketty/main.go:233-241`: `io.Copy(os.Stdout, ptmx)` runs in a goroutine but `cmd.Wait()` returns on process exit, not on master-FD drain — the workload's last bytes can be lost between `Wait()` returning and `os.Exit`. Standard PTY-wrapper polish; can land in phase 1b alongside the RPC server.

The deferred `TestKuke_AttachDetach_KeepsTaskRunning` is non-blocking — `t.Skip` reason names #410, matching the parent issue's AC carve-out for environment-bound smokes.

LGTM once the `resolveKukettyBinary` contract is reconciled — the rest is non-blocking.

eminwux:
Addressed in 7c96b8c:
- `resolveKukettyBinary` now actually calls `exec.LookPath("kuketty")` before returning the not-found error, honoring the documented step-3 PATH fallback. The `tried` slice still appends `"$PATH/kuketty"` so the error trace remains the same when PATH also misses.

Deferred: PTY stdout-drain sync with `cmd.Wait()` in `cmd/kuketty/main.go:233-241` — per review, non-blocking polish to land in phase 1b alongside the JSON-RPC + SCM_RIGHTS server.

eminwux:
**PR #411 Follow-up Review — `feat: introduce kuketty and swap sbsh on OCI injection (phase 1)`**

Fix commit `7c96b8c1` cleanly addresses the blocking item: `exec.LookPath("kuketty")` is now called before the error return at `internal/controller/runner/attachable.go:336-338`, with the `"$PATH/kuketty"` string still appended to `tried` on miss so the error trace continues to show PATH was consulted. Docstring step 3 and the rewritten inline comment now match the behavior.

The non-blocking PTY-drain nit (`cmd/kuketty/main.go:233-241`) is unchanged — deferred to phase 1b per my prior comment.

LGTM.

PR #418, opened 2026-05-11
feat: persist lifecycle timestamps and runtime probes in Status

### LINKED ISSUE #166: api: persist timestamps and runtime probes in Status (all kinds)
## Context
Persisted `Status` for Realm, Space, Stack, and Cell currently carries only `State` and `CgroupPath` (Cell additionally carries `Containers`). There is no record of when a document was created or last updated, why it is in its current state, or whether the on-host runtime artifacts (containerd namespace, cgroup) actually exist. This makes it hard to reason about age, drift, or failure cause from a `metadata.json` snapshot — and the future reconciler (#161) has nowhere durable to write its observations beyond the `State` enum.

Status types today:
- `pkg/api/model/v1beta1/realm.go:48`
- `pkg/api/model/v1beta1/space.go:100`
- `pkg/api/model/v1beta1/stack.go:38`
- `pkg/api/model/v1beta1/cell.go:51`

## Proposal
Extend each kind's `Status` with lifecycle and runtime-health fields. Scope is intentionally limited to *lifecycle-scoped* observations — point-in-time telemetry (memory/cpu/pids) stays out, since `client.CgroupMetrics()` (`internal/ctr/cgroups.go:164`) already serves live reads and persisting it would churn `metadata.json` on every reconcile tick.

**All four kinds**
- `CreatedAt`, `UpdatedAt`, `ReadyAt time.Time`
- `Reason`, `Message string` — short reason code plus human-readable detail for the current `State`; especially valuable on `Failed`.
- `CgroupReady bool` — re-verifies the cgroup at `CgroupPath` actually exists on disk (today the path is set on creation but never re-checked).

**Realm-only**
- `ContainerdNamespaceReady bool` — confirms the containerd namespace is present, not just recorded in metadata. Backed by the existing `internal/ctr/namespaces.go` helpers.

No on-disk migration: missing fields read as zero values from existing `metadata.json` files; the daemon writes them on the next status update.

## Acceptance criteria
- [ ] `RealmStatus`, `SpaceStatus`, `StackStatus`, `CellStatus` gain `CreatedAt`, `UpdatedAt`, `ReadyAt`, `Reason`, `Message`, `CgroupReady`.
- [ ] `RealmStatus` additionally gains `ContainerdNamespaceReady`.
- [ ] `CreatedAt` is set on initial persist; `UpdatedAt` is bumped on every status write; `ReadyAt` is set once on the first `Ready` transition and never overwritten.
- [ ] Existing `metadata.json` files load with zero-value defaults (no migration).
- [ ] `kuke get realm/space/stack/cell` surfaces the new fields in human and `-o yaml` / `-o json` output.
- [ ] Daemon-parity check (`kuke get realms` vs `kuke get realms --no-daemon`) still produces identical output.
- [ ] Tests cover: timestamp population, `ReadyAt` set-once semantics, `Reason`/`Message` round-trip, `CgroupReady` true/false vs filesystem state, `ContainerdNamespaceReady` true/false vs containerd presence.

## Notes
- Complements #161 (background reconciler): once that loop lands it will be the primary writer of `Reason`/`Message` and `CgroupReady`, but the fields are useful before the loop exists (populated during create/start/stop paths).
- Deferred to follow-ups: `ObservedGeneration` (depends on a spec-generation concept and couples to #161), Space network/CNI status, per-container `ImageDigest`, child-rollup counts, and state-enum harmonization across kinds (Realm has `Creating`/`Deleting`; Stack/Space/Cell do not).

eminwux:
## Summary
- Adds lifecycle-scoped fields (`CreatedAt`, `UpdatedAt`, `ReadyAt`, `Reason`, `Message`, `CgroupReady`) to `RealmStatus`, `SpaceStatus`, `StackStatus`, `CellStatus`, plus `ContainerdNamespaceReady` on `RealmStatus` only. Lets a `metadata.json` snapshot answer "how old is this, is the cgroup still mounted, and why is the state Failed?" — and gives the future reconciler (#161) somewhere durable to write its observations beyond the `State` enum.
- Timestamp stamping is centralised in the four `UpdateXMetadata` paths. `CreatedAt` is stamped only when zero so the value never moves once set; `UpdatedAt` is bumped on every persist; `ReadyAt` is set once on the first `State==Ready` persist and never overwritten on subsequent Ready transitions or state flaps. Logic is extracted as `stampRealmLifecycle` / `Space` / `Stack` / `Cell` pure helpers so the invariants are unit-testable independent of the metadata-write path.
- Probe fields are populated by the existing refresh path (`RefreshRealm` / `RefreshSpace` / `RefreshStack` / `RefreshCell`). `CgroupReady` is the result of the existing `ExistsCgroup` call; `ContainerdNamespaceReady` is the result of the existing `ExistsRealmContainerdNamespace` call. New `carryXLifecycle` helpers preserve the set-once timestamps and `Reason`/`Message` through the locally-built `newStatus` so the refresh tick does not erase them. The runner gains an `nowFn` override hook so the stamper can be driven from a frozen clock in tests; production callers fall through to `time.Now().UTC()`.
- No on-disk migration: missing fields read as zero values from existing `metadata.json` files (verified with a hand-crafted legacy file). The default `kuke get realms` table is intentionally left unchanged so the daemon-parity regression guard the dev-init smoke check expects keeps passing; new fields are surfaced via the default single-resource view (which is YAML), `-o yaml`, and `-o json`.

## Test plan
- [x] `make test` (the project's named CI target).
- [x] `go vet ./...` clean; `gofmt -l internal/ pkg/ cmd/` produces no diff.
- [x] `make dev-init` rebuilds and re-bootstraps the local kukeond image. The daemon-parity tail produces identical `kuke get realms` / `kuke get realms --no-daemon` output in the documented `NAME / NAMESPACE / STATE / CGROUP` layout. (The script exits non-zero on a later, unrelated `kuke attach` smoke step that needs `expect(1)` to be installed on the host — environmental, predates this change.)
- [x] `kuke get realm default -o yaml` after `make dev-init` shows populated `createdAt` / `updatedAt` / `readyAt` (cell, stack, space and realm verified separately).
- [x] `kuke refresh` followed by `kuke get` shows `cgroupReady: true` and (for the realm) `containerdNamespaceReady: true`. `createdAt` and `readyAt` are unchanged across the refresh; `updatedAt` advanced.
- [x] Hand-crafted legacy `metadata.json` (no lifecycle fields) loads cleanly with zero-value defaults — exercises the no-migration AC.
- [x] Unit tests added: `TestStampRealmLifecycle` / `Space` / `Stack` / `Cell` cover the set-once / bump-on-update invariants; `TestUpdateRealmMetadataPersistsLifecycleFields` exercises the full stamp → convert → write → read pipeline; `TestCarryRealmLifecycle` pins the refresh-path carry; `TestStatusLifecycleFieldsRoundTripV1Beta1` and `TestStatusLifecycleFieldsZeroDefault` cover the apischeme conversions for all four kinds.

Closes #166

### DIFF
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -59,9 +59,16 @@ func ConvertRealmDocToInternal(in ext.RealmDoc) (intmodel.Realm, error) {
 				RegistryCredentials: registryCreds,
 			},
 			Status: intmodel.RealmStatus{
-				State:              intmodel.RealmState(in.Status.State),
-				CgroupPath:         in.Status.CgroupPath,
-				SubtreeControllers: cloneStringSlice(in.Status.SubtreeControllers),
+				State:                    intmodel.RealmState(in.Status.State),
+				CgroupPath:               in.Status.CgroupPath,
+				SubtreeControllers:       cloneStringSlice(in.Status.SubtreeControllers),
+				CreatedAt:                in.Status.CreatedAt,
+				UpdatedAt:                in.Status.UpdatedAt,
+				ReadyAt:                  in.Status.ReadyAt,
+				Reason:                   in.Status.Reason,
+				Message:                  in.Status.Message,
+				CgroupReady:              in.Status.CgroupReady,
+				ContainerdNamespaceReady: in.Status.ContainerdNamespaceReady,
 			},
 		}, nil
 	default:
@@ -93,9 +100,16 @@ func BuildRealmExternalFromInternal(in intmodel.Realm, apiVersion ext.Version) (
 				RegistryCredentials: registryCreds,
 			},
 			Status: ext.RealmStatus{
-				State:              ext.RealmState(in.Status.State),
-				CgroupPath:         in.Status.CgroupPath,
-				SubtreeControllers: cloneStringSlice(in.Status.SubtreeControllers),
+				State:                    ext.RealmState(in.Status.State),
+				CgroupPath:               in.Status.CgroupPath,
+				SubtreeControllers:       cloneStringSlice(in.Status.SubtreeControllers),
+				CreatedAt:                in.Status.CreatedAt,
+				UpdatedAt:                in.Status.UpdatedAt,
+				ReadyAt:                  in.Status.ReadyAt,
+				Reason:                   in.Status.Reason,
+				Message:                  in.Status.Message,
+				CgroupReady:              in.Status.CgroupReady,
+				ContainerdNamespaceReady: in.Status.ContainerdNamespaceReady,
 			},
 		}, nil
 	default:
@@ -145,6 +159,12 @@ func ConvertSpaceDocToInternal(in ext.SpaceDoc) (intmodel.Space, error) {
 				State:              intState,
 				CgroupPath:         in.Status.CgroupPath,
 				SubtreeControllers: cloneStringSlice(in.Status.SubtreeControllers),
+				CreatedAt:          in.Status.CreatedAt,
+				UpdatedAt:          in.Status.UpdatedAt,
+				ReadyAt:            in.Status.ReadyAt,
+				Reason:             in.Status.Reason,
+				Message:            in.Status.Message,
+				CgroupReady:        in.Status.CgroupReady,
 			},
 		}, nil
 	default:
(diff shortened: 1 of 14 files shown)


eminwux:
**PR #418 Review — `feat: persist lifecycle timestamps and runtime probes in Status`**

Status-struct additions are cleanly scoped, the four `stampX`/`carryX` helpers each have a single live call site (verified on PR head), and the apischeme conversions round-trip every new field in both directions with dedicated tests covering all four kinds and the zero-default no-migration AC. Set-once invariants are unit-tested per-kind and re-tested end-to-end through `TestUpdateRealmMetadataPersistsLifecycleFields`. `make dev-init` daemon-parity regression guard is in the test plan. No sensitive data in any new path; commit is signed.

One non-blocking observation:

- **Nit** `internal/controller/runner/provision.go:286`: pre-existing CNI-conf regen workaround calls `metadata.WriteMetadata` directly, bypassing `UpdateSpaceMetadata` and therefore `stampSpaceLifecycle`, so `UpdatedAt` isn't bumped on this persist (lifecycle fields themselves survive the apischeme round-trip — nothing is zeroed). A `stampSpaceLifecycle(&updatedSpace, r.nowUTC())` call before `BuildSpaceExternalFromInternal` would align with the PR-body invariant "`UpdatedAt` is bumped on every persist".

LGTM.

PR #420, opened 2026-05-12
feat: wire cell profile to sbsh builder (phase 4)

### LINKED ISSUE #290: kukeon: wire cell profile to sbsh builder (phase 4)

eminwux:
## Summary
- Plumb a sbsh `TerminalProfile` (name + host-side profiles dir) end-to-end from `ContainerTty` into the per-container kuketty `TerminalDoc` via `pkg/builder.BuildTerminalSpec(WithProfile(...), WithProfilesDir(...))`. sbsh's loader and builder do the YAML resolution; kukeon keeps no profile interpreter.
- Drop the unconditional `WithDisableSetPrompt(true)` safety belt **only when** a profile is configured, so the profile's `Prompt` / `Stages` / `SetPrompt=true` actually take effect. The no-profile path keeps the legacy `DisableSetPrompt` default so a non-shell workload (nginx, python) never receives a literal `export PS1=…` on its stdin.
- `ProfilesDir` is only honored alongside `Profile` — in isolation sbsh's builder would default the profile name to `"default"` and chase the configured dir for a YAML the operator never asked to load. Matches the issue's "builder options are omitted" wording for the no-profile path.

## Notes
- AC #3 mentions `Spec.Prompt` staying zero in the no-profile path. sbsh's hardcoded default profile (the fallback when `discovery.FindProfileByNameInDir` returns nothing) populates `Spec.Prompt` with its own `"(sbsh-$SBSH_TERM_ID) $PS1"` string — kukeon cannot suppress that without writing the profile interpreter the issue explicitly forbids. The AC's intent (workload PS1 not modified) holds because `Spec.SetPrompt=false` keeps the runner from applying that prompt at all. Spec.SetPrompt and Spec.Stages do stay zero. Flagging for the reviewer in case the literal "Prompt zero" wording is load-bearing.
- Schema lives on `ContainerTty` next to `LogFile` rather than `CellTty`, matching the existing per-attach shell-UX pattern. Validation already rejects any tty field with `Attachable=false`; extended the test matrix to cover the two new fields.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (no FAIL across the suite; touched packages: apischeme, controller/runner, modelhub, pkg/api/model/v1beta1)
- [x] `make dev-init` — daemon parity check passes, kuketty-wrapped cell attach smoke exits cleanly
- [x] `pre-commit run --files <changed>` (golangci-lint auto-formatted struct tags on first pass, clean on second)

Closes #290

### DIFF
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -516,7 +516,12 @@ func convertContainerTtyToInternal(in *ext.ContainerTty) *intmodel.ContainerTty 
 	if in.IsEmpty() {
 		return nil
 	}
-	out := &intmodel.ContainerTty{Prompt: in.Prompt, LogFile: in.LogFile}
+	out := &intmodel.ContainerTty{
+		Prompt:      in.Prompt,
+		LogFile:     in.LogFile,
+		Profile:     in.Profile,
+		ProfilesDir: in.ProfilesDir,
+	}
 	if len(in.OnInit) > 0 {
 		out.OnInit = make([]intmodel.TtyStage, len(in.OnInit))
 		for i, s := range in.OnInit {
@@ -532,7 +537,12 @@ func buildContainerTtyExternalFromInternal(in *intmodel.ContainerTty) *ext.Conta
 	if in.IsEmpty() {
 		return nil
 	}
-	out := &ext.ContainerTty{Prompt: in.Prompt, LogFile: in.LogFile}
+	out := &ext.ContainerTty{
+		Prompt:      in.Prompt,
+		LogFile:     in.LogFile,
+		Profile:     in.Profile,
+		ProfilesDir: in.ProfilesDir,
+	}
 	if len(in.OnInit) > 0 {
 		out.OnInit = make([]ext.TtyStage, len(in.OnInit))
 		for i, s := range in.OnInit {
--- a/internal/apischeme/scheme_test.go
+++ b/internal/apischeme/scheme_test.go
@@ -1132,7 +1132,9 @@ func TestContainerTtyRoundTripV1Beta1(t *testing.T) {
 					{Script: "git pull"},
 					{Script: "claude"},
 				},
-				LogFile: "/run/kukeon/tty/log",
+				LogFile:     "/run/kukeon/tty/log",
+				Profile:     "claude",
+				ProfilesDir: "/opt/kukeon/profiles",
 			},
 		},
 	}
@@ -1148,6 +1150,12 @@ func TestContainerTtyRoundTripV1Beta1(t *testing.T) {
 	}
 	if internal.Spec.Tty.LogFile != input.Spec.Tty.LogFile {
 		t.Errorf("internal logFile = %q, want %q", internal.Spec.Tty.LogFile, input.Spec.Tty.LogFile)
+	}
+	if internal.Spec.Tty.Profile != input.Spec.Tty.Profile {
+		t.Errorf("internal profile = %q, want %q", internal.Spec.Tty.Profile, input.Spec.Tty.Profile)
+	}
+	if internal.Spec.Tty.ProfilesDir != input.Spec.Tty.ProfilesDir {
+		t.Errorf("internal profilesDir = %q, want %q", internal.Spec.Tty.ProfilesDir, input.Spec.Tty.ProfilesDir)
 	}
 	if len(internal.Spec.Tty.OnInit) != 2 ||
 		internal.Spec.Tty.OnInit[0].Script != "git pull" ||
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #420 Review — `feat: wire cell profile to sbsh builder (phase 4)`**

Clean phase-4 wiring: `Profile` / `ProfilesDir` are plumbed through both model mirrors, `IsEmpty()`, apischeme round-trip, and the attachable renderer, with three focused tests locking the profile / no-profile / dir-without-profile branches. The safety-belt logic — drop `WithDisableSetPrompt(true)` only when `Profile` is explicitly configured — is the right call and the test asserts it both ways. No sensitive-info leaks, all commits signed, no dead symbols (Explore confirmed `sbshbuilder.WithProfile`/`WithProfilesDir` and the new fields are reached only via the diff).

A couple of small observations, both non-blocking:

- **Doc accuracy** `pkg/api/model/v1beta1/container.go:138-139`: `Profile` field comment claims `Spec.Prompt / Spec.SetPrompt / Spec.Stages stay zero` in the no-profile path, but per your PR-description note sbsh's hardcoded default still populates `Spec.Prompt` with `(sbsh-$ID) $PS1` — only `SetPrompt=false` keeps it inert. Functional behavior is correct (workload PS1 isn't modified), but the literal wording will confuse a future debugger inspecting the rendered Spec. Worth a one-line tweak in a follow-up.
- **Silent no-op** `pkg/api/model/v1beta1/container.go:144-149` / `internal/controller/runner/attachable.go:183`: setting `ProfilesDir` without `Profile` is silently dropped at render time (and `TestWriteKukettyMetadata_ProfilesDirIgnoredWithoutProfile` locks that). Validation accepts the combination, so a user can ship a no-op config without warning. Either reject in `validateContainerTty` or extend the `ProfilesDir` field comment to say "ignored unless Profile is set" — your call.

LGTM.

PR #422, opened 2026-05-12
refactor: retire sbsh binary cache distribution (phase 5)

eminwux:
## Summary

- Delete the dead sbsh-binary-distribution surface: `internal/ctr/sbsh_cache.go`, its companion test, the `ResolveSbshCachePath` entry on the `internal/ctr.Client` interface, and the two test fakes that satisfied it (`delete_cell_test.go`, `provision_subtree_test.go`).
- Drop the now-unused `stageHostSbsh` + `requireSbshHasPostMergeFlags` helpers from `e2e/e2e_kuke_attach_test.go` (and their call site) — these staged an on-host `sbsh` binary into the per-test `cache/sbsh/<arch>/sbsh` layout that no live caller has read since #411 swapped the OCI injection path to kuketty.
- Retain the sbsh Go-module dependency intact: `cmd/kuke/...` consumes `pkg/attach` client-side, `cmd/kuketty/main.go` consumes `pkg/terminal/server` + `pkg/api` + `pkg/builder` server-side. Only the binary-distribution mechanism is retired here, per the issue's "Two retirements, distinguished" note.
- Cross-realm regression guard at `internal/ctr/attachable_test.go:205-214` (legacy `/run/sbsh.socket` and `/.kukeon/bin/sbsh` mounts must not reappear) stays in place — phase 1 already proved no live caller resolves the cache, and this PR's deletion makes that permanent.

## Test plan

- [x] `make test` — full unit suite green (kuketty/runner/ctr/controller packages all pass)
- [x] `go build ./...` — no compile-time breakage
- [x] `go vet ./...` — clean
- [x] `make dev-init` — daemon-parity check passes (`kuke get realms` and `kuke get realms --no-daemon` produce identical output)
- [ ] `make dev-init`'s post-bootstrap `kuke attach` smoke — host lacks `expect(1)`, dev-init logs `expect(1) not on PATH; install \`expect\``; the regression-guard daemon-parity check above (the bit the CLAUDE.md smoke playbook anchors on) still passed before the failure
- [ ] `make test-e2e` (incl. `TestKuke_AttachDetach_KeepsTaskRunning`) — not run on this host; the e2e harness needs a host docker + containerd + a built kuke binary and operator privileges that aren't safe to exercise alongside the dev-init smoke above. The compile pass via `go test -count=1 -run "^\$" ./e2e/...` confirms the e2e suite still builds against the trimmed `internal/ctr.Client` surface.

## Notes for reviewer

[...]

PR #425, opened 2026-05-12
test: use containerd namespace for e2e attach task verify

### LINKED ISSUE #424: e2e: TestKuke_AttachDetach_KeepsTaskRunning checks wrong containerd namespace, would fail if not skip-gated by host-sbsh
## Context

`TestKuke_AttachDetach_KeepsTaskRunning` (in `e2e/e2e_kuke_attach_test.go`) asserts that the workload's container task is `RUNNING` after the client sends the detach keystroke. The assertion is implemented as:

```go
// e2e/e2e_kuke_attach_test.go:155 (PR #422 branch) / :223 (main)
if !verifyContainerTaskIsRunning(t, realmName, containerdID) {
    t.Fatalf("container task %q in namespace %q is not RUNNING after detach",
        containerdID, realmName)
}
```

`verifyContainerTaskIsRunning(t, namespace, containerID)` invokes `ctr --namespace <namespace> task ls` and greps for the container ID + `RUNNING`. But the value being passed is the **realm name** (e.g. `e-r-6e`), not the **containerd namespace** (which is `e-r-6e.kukeon.io`, formed by `internal/consts.RealmNamespace(realm)` appending `.kukeon.io`). Other e2e tests pass the namespace correctly via `getRealmNamespace(t, runPath, realmName)` (see `e2e/e2e_kuke_container_test.go:150` and friends).

Concretely, with the workload alive in `tr-1.kukeon.io`:

```
$ ctr --namespace tr-1 task ls
TASK    PID    STATUS

$ ctr --namespace tr-1.kukeon.io task ls
TASK                   PID      STATUS
ts-1_tk-1_tc-1_root    54254    RUNNING
ts-1_tk-1_tc-1_work    54334    RUNNING
```

So `verifyContainerTaskIsRunning` always returns false against the realm name — the assertion would always fire post-detach, even though the workload is in fact running. This means the test as written is unconditionally failing on the detach assertion whenever it executes.

The reason this has not been caught: on `main` the test is skip-gated by `stageHostSbsh(t, runPath)` which `t.Skip`s when the host has no `sbsh` binary on `PATH`. The CI runner has no sbsh, so the test has been silently skipping since #410 / #413 re-introduced it to the suite. PR #422 (phase 5) removes the sbsh-binary-distribution skip-gate (per the issue's AC), which exposes this latent test bug.

## Reproduction

On the `refactor/retire-sbsh-cache` branch (PR #422), or on `main` with `sbsh` made available on PATH:

```bash
make kuke kuketty
KUKEON_E2E_IMAGE=docker.io/library/kukeon-local:e2e \
KUKEON_E2E_IMAGE_DOCKER_NAME=kukeon-local:e2e \
E2E_BIN_DIR=$(pwd) \
go test -v -count=1 -run TestKuke_AttachDetach_KeepsTaskRunning ./e2e
```

Observed:

```
e2e_kuke_attach_test.go:156: container task "e-sp-9d_e-st-0f_e-ce-8c_work" in namespace "e-r-6e" is not RUNNING after detach
--- FAIL: TestKuke_AttachDetach_KeepsTaskRunning (13.46s)
```

Manual containerd inspection during the test window confirms the work container task is `RUNNING` in the correctly-suffixed namespace `e-r-6e.kukeon.io` — the test's verification is looking up the wrong namespace.

## Proposal

Fix the call site at `e2e/e2e_kuke_attach_test.go:155` to pass `consts.RealmNamespace(realmName)` instead of `realmName`. This mirrors the pattern used by every other e2e test that interrogates containerd state by realm. Add the `internal/consts` import to the e2e attach test file.

Minimal diff:

```go
// e2e/e2e_kuke_attach_test.go
import (
    ...
    "github.com/eminwux/kukeon/internal/consts"
    ...
)

// :155
if !verifyContainerTaskIsRunning(t, consts.RealmNamespace(realmName), containerdID) {
    t.Fatalf("container task %q in namespace %q is not RUNNING after detach",
        containerdID, consts.RealmNamespace(realmName))
}
```

Alternative: change `verifyContainerTaskIsRunning`'s parameter name to make the contract explicit and have it call `consts.RealmNamespace` internally — but the function name is already `…Task…` not `…Realm…` and the rest of the e2e helpers (`verifyContainerTaskExists`, `verifyContainerExists`) all take a fully-qualified namespace, so the simpler fix is the call site.

## Acceptance criteria

- [ ] `e2e/e2e_kuke_attach_test.go:155` passes `consts.RealmNamespace(realmName)` (or equivalent) so `verifyContainerTaskIsRunning` queries the correct containerd namespace.
- [ ] `TestKuke_AttachDetach_KeepsTaskRunning` passes end-to-end on a host with kuketty + docker + containerd (no `sbsh` binary required).
- [ ] No regression to the other e2e tests in `./e2e/...`.

## Out of scope

- Wider e2e suite stability (other tests fail with `dial unix /run/kukeon/kukeond.sock: connect: no such file or directory` and `TestKuke_Init_VerifyState` container-conflict on a host that already ran `make dev-init`). Those are separate ordering / setup concerns that predate this issue and warrant their own grooming.
- The reviewer-noted host-gated nature of `make test-e2e` itself (root docker + containerd + privileged operations). Whether to add a wrapper that boots a clean daemon is orthogonal to the namespace fix.

## References

- Test added in #75 (`test: add e2e PTY harness and attach/detach scenario`), updated in #115, #117, #125, then re-enabled in #410 / #413.
- Latent because `stageHostSbsh` (line ~136 on main) `t.Skip`s when host `sbsh` is absent.
- PR #422 (kuketty: retire sbsh code paths, phase 5) surfaced this by removing the `stageHostSbsh` gate — the test now executes unconditionally on a kuketty-only host and the assertion fires.
- Pattern reference: `e2e/e2e_kuke_container_test.go:150` uses `getRealmNamespace(t, runPath, realmName)` to get the fully-qualified namespace before passing it to `verifyContainerTaskExists`.
- `internal/consts.RealmNamespace` joins `realm + ".kukeon.io"` (`internal/consts/consts.go`).

eminwux:
## Summary
- `TestKuke_AttachDetach_KeepsTaskRunning` was passing the realm name (e.g. `e-r-6e`) to `verifyContainerTaskIsRunning`, but the helper greps `ctr --namespace <ns> task ls` and the containerd namespace is the realm name plus `.kukeon.io` (per `internal/consts.RealmNamespace`). The assertion therefore looked up the wrong namespace, found nothing, and fired post-detach even though the workload was running. The bug was masked on `main` until #422 because `stageHostSbsh` `t.Skip`'d the test when `sbsh` was missing from `PATH`; once that gate was removed in phase 5, the latent issue surfaced.
- Fix calls `consts.RealmNamespace(realmName)` at the single call site and reuses the value in the `t.Fatalf` message so the diagnostic matches what was actually queried. Mirrors the pattern used by every other e2e helper that interrogates containerd by realm.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test`
- [x] `KUKEON_E2E_IMAGE=docker.io/library/kukeon-local:e2e KUKEON_E2E_IMAGE_DOCKER_NAME=kukeon-local:e2e E2E_BIN_DIR=$(pwd) go test -v -count=1 -run TestKuke_AttachDetach_KeepsTaskRunning ./e2e` — passes in 13.7s
- [ ] Full `make test-e2e` not re-run — orthogonal pre-existing ordering issues (`/run/kukeon/kukeond.sock` failures, `TestKuke_Init_VerifyState` container-conflict) exist on this host; the targeted test is the AC.

Closes #424

### DIFF
--- a/e2e/e2e_kuke_attach_test.go
+++ b/e2e/e2e_kuke_attach_test.go
@@ -25,6 +25,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/util/fs"
 )
 
@@ -152,9 +153,10 @@ func TestKuke_AttachDetach_KeepsTaskRunning(t *testing.T) {
 		t.Fatalf("get cell ID: %v", err)
 	}
 	containerdID := fmt.Sprintf("%s_%s_%s_%s", spaceName, stackName, cellID, "work")
-	if !verifyContainerTaskIsRunning(t, realmName, containerdID) {
+	realmNamespace := consts.RealmNamespace(realmName)
+	if !verifyContainerTaskIsRunning(t, realmNamespace, containerdID) {
 		t.Fatalf("container task %q in namespace %q is not RUNNING after detach",
-			containerdID, realmName)
+			containerdID, realmNamespace)
 	}
 }
 


eminwux:
**PR #425 Review — `test: use containerd namespace for e2e attach task verify`**

Correctly fixes a real latent bug — `verifyContainerTaskIsRunning` runs `ctr --namespace <ns> task ls`, and the previous call site was feeding it the bare realm name. Routing through `consts.RealmNamespace(realmName)` matches every other e2e helper that talks to containerd (`e2e/e2e_test.go:207,691,717,746`, `e2e/e2e_kuke_test.go:563`), so the fix lands at the canonical site. No sensitive data, no dead symbols, no public API impact. The unchecked `make test-e2e` line is the host-gated carve-out documented in the PR body — non-blocking.

LGTM.

PR #430, opened 2026-05-12
refactor: replace expect(1) attach smoke with a Go PTY helper

### LINKED ISSUE #428: dev-init: replace expect(1) attach smoke with a Go PTY helper
## Context

`scripts/dev-init.sh` requires `expect(1)` as a runtime prerequisite — the script `exit 1`s with an install hint if it's not on PATH (lines 186-189). `expect` is a non-obvious apt/rpm package that most Go contributors don't have installed by default, so the contributor's first run of `make dev-init` after a fresh clone fails several minutes in with a confusing prereq error.

`expect` is used in exactly one place: the final `kuke attach` smoke test (`scripts/dev-init.sh:182-207`). It allocates a PTY (because `pkg/attach` refuses to run without a TTY), sleeps 2s for the raw-mode keyboard filter to wire up, sends `\x1d\x1d` (Ctrl+] Ctrl+]) to trigger detach, and enforces a 20s timeout. Bash alone can't do any of that, so something must drive the PTY — but `github.com/creack/pty v1.1.24` is already a direct dependency (`go.mod`), so a small Go helper can replace the Tcl block with no new toolchain footprint.

## Proposal

Add `hack/attach-smoke/main.go` (or similar — final path is dev's call) that:
- Allocates a PTY via `github.com/creack/pty`.
- Spawns `sudo ./kuke attach <cell> --realm ... --space ... --stack ... --container ...` connected to that PTY, with cell coordinates passed as flags or args.
- Waits 2s, then writes `\x1d\x1d` to the master side.
- Waits for the child to exit, with a 20s overall deadline; exits non-zero on timeout or non-zero child exit.
- Logs the child's combined output to a path passed via flag (matches today's `ATTACH_LOG` behavior).

Update `scripts/dev-init.sh:182-207` to invoke the helper instead of `expect`. Remove the `command -v expect` prereq check.

## Acceptance criteria

- [ ] New Go helper exists under `hack/` (or wherever dev decides) and is invoked by `scripts/dev-init.sh` in place of the current `expect <<EOF ... EOF` block.
- [ ] `expect` is no longer mentioned in `scripts/dev-init.sh` (no prereq check, no install hint).
- [ ] `make dev-init` runs end-to-end on a host without `expect` installed and produces the same daemon-parity tail + clean attach exit.
- [ ] The detach sequence, 2s wire-up sleep, and 20s overall timeout match today's behavior (regression-equivalent).

## Notes

- Existing dependency that makes this cheap: `github.com/creack/pty v1.1.24` in `go.mod`.
- The detach sequence and timeout values come from `scripts/dev-init.sh:193-200`; the raw-mode filter wire-up window is the same 2s the current script uses.

eminwux:
## Summary

- `scripts/dev-init.sh` required `expect(1)` solely to drive the final `kuke attach` smoke; an unobvious prereq that fails contributor first-runs several minutes in. `github.com/creack/pty` is already a direct dependency, so a small Go helper replaces the Tcl block with no new toolchain footprint.
- New `hack/attach-smoke/main.go` allocates a PTY, spawns an arbitrary command attached to it, waits a configurable connect-grace, sends sbsh's `\x1d\x1d` detach sequence, enforces an overall deadline, and writes the child's transcript to `--log`. Defaults (`2s` connect-grace, `20s` exit-timeout) match the prior expect block. Detach sequence is hardcoded to match the e2e suite's `sbshDetachSequence`.
- Helper is invoked via `go run ./hack/attach-smoke` so no extra make target / build artifact is needed; the `kuke` build path stays untouched.

## Test plan

- [x] `go vet ./...` — clean
- [x] `make test` — full unit suite green (includes the new `hack/attach-smoke` module; `[no test files]`, builds clean)
- [x] Helper self-tests — verified exit-code propagation (`exit 42` → rc 42), timeout path (sleep 5 with 3s deadline → "child did not exit within 3s"), and detach-sequence write (`^]^]` present in transcript)
- [x] `make dev-init` — full re-bootstrap green on a host without `expect` installed; daemon-parity tail matches CLAUDE.md, attach smoke prints `kuke attach exited cleanly (...)`

Closes #428

### DIFF
--- a/hack/attach-smoke/main.go
+++ b/hack/attach-smoke/main.go
@@ -1,0 +1,142 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// attach-smoke drives an interactive subprocess attached to a freshly
+// allocated pseudoterminal, then sends sbsh's Ctrl+] Ctrl+] detach
+// sequence and waits for the child to exit. It exists so scripts/dev-init.sh
+// can smoke-test `kuke attach` without an expect(1) prereq; the helper
+// reuses github.com/creack/pty (already a direct dependency).
+//
+// Behaviour mirrors the prior expect block:
+//   - allocate a PTY and spawn the child connected to it
+//   - sleep --connect-grace (default 2s) so the pkg/attach raw-mode
+//     keyboard filter wires up before keystrokes arrive
+//   - write 0x1d 0x1d (Ctrl+] Ctrl+]) to the PTY master
+//   - wait up to --exit-timeout (default 20s) for the child to exit
+//   - exit non-zero on timeout or non-zero child exit; the child's
+//     transcript is always captured to --log
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/creack/pty"
+)
+
+// detachSequence is the keystroke pair the sbsh client filter recognises
+// as a request to detach (Ctrl+] Ctrl+], adjacent). It matches the
+// constant the e2e suite uses; see e2e/e2e_kuke_attach_test.go.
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #430 Review — `refactor: replace expect(1) attach smoke with a Go PTY helper`**

Clean swap of an `expect(1)` prereq for a small Go helper using an existing direct dep (`github.com/creack/pty`). Behaviour parity verified: detach sequence `{0x1d, 0x1d}` matches `e2e/e2e_kuke_attach_test.go:35`, defaults (2s grace, 20s deadline) mirror the prior expect block, shell invocation in `scripts/dev-init.sh` is a mechanical substitution. Goroutine choreography in the select loop is correct — `detach = nil` after first fire prevents re-arm, both timeout and exit paths properly drain `waitDone`/`copyDone` before returning, deferred closes are idempotent. Apache+SPDX header present per project convention; no error sentinels added (none reusable here). Test plan ticks `make dev-init` end-to-end including the daemon-parity tail per CLAUDE.md. Single commit, signature verified. No correctness, secret-leak, dead-code, or convention concerns.

LGTM.

PR #450, opened 2026-05-13
fix: release kukeon-owned bind mounts before kuke uninstall rmdir

### LINKED ISSUE #434: cli: `kuke uninstall -y` fails on a healthy `make dev-init` host because `/run/kukeon/tty` bind mount blocks rmdir
## Context

`docs/cli-use-cases.md` § "Full per-host teardown" guarantees:

> With `-y`, exit code 0 on success. Side effect: every realm purged with `--cascade`; `/run/kukeon` and the configured run path (default `/opt/kukeon`) removed; the `kukeon` system user/group removed if present.

Observed on a `make dev-init` host (which includes a `kuke attach` smoke step that leaves a bind mount on `/run/kukeon/tty`): `kuke uninstall -y` exits **non-zero** because `unlinkat /run/kukeon/tty: device or resource busy`. Most side effects do happen, but `/run/kukeon` removal fails.

```
$ sudo ./kuke uninstall -y
kuke uninstall:
  - kukeond: stopped (pid 21909)
  - /run/kukeon: remove failed
  - /opt/kukeon: removed
  - user "kukeon": removed
  - group "kukeon": absent
Error: remove socket dir "/run/kukeon": unlinkat /run/kukeon/tty: device or resource busy
$ echo "exit=$?"
exit=1
```

`/proc/mounts` confirms a live bind mount kukeon itself created during the smoke:

```
$ mount | grep kukeon
/dev/sda1 on /.kukeon/bin/sbsh type ext4 (ro,relatime,errors=remount-ro)
/dev/sda1 on /run/kukeon/tty type ext4 (rw,relatime,errors=remount-ro)
```

The bind mount survives `kuke purge realm default --cascade`, so even though every realm reports purged, `/run/kukeon/tty` is still a live mountpoint when uninstall tries to `rmdir` `/run/kukeon`. The invariant is unconditional ("exit code 0 on success"); the doc does not carve out an "active kuketty mount" exception.

## Proposal

`kuke uninstall -y` should be responsible for unmounting kukeon-owned bind mounts before removing the run-path directories. Concretely:

1. Before tearing down realms, enumerate live bind mounts under `/run/kukeon` and `/opt/kukeon` (and the kuketty `/.kukeon/...` paths it owns) via `/proc/mounts`.
2. `unmount` each (lazy unmount as a fallback) so the subsequent `rmdir` succeeds.
3. If a mount cannot be released (still busy after lazy unmount), keep the current non-zero exit but make the report row name the surviving mountpoint and the suggested manual recovery, rather than only "remove failed".

## Acceptance criteria

- [ ] On a `make dev-init` host (which leaves `/run/kukeon/tty` bind-mounted from the attach smoke), `sudo kuke uninstall -y` exits 0.
- [ ] `/run/kukeon` is absent after success, including any kukeon-owned bind mounts under it.
- [ ] If a kukeon-owned mount truly cannot be released, the report row names the mountpoint explicitly and exit is non-zero.
- [ ] The matching invariant in `docs/cli-use-cases.md` § "Full per-host teardown" passes when exercised by `/test-cli "Bootstrap & teardown"`.

## Out of scope

- Non-kukeon mounts that happen to live under `/run/kukeon` or `/opt/kukeon` (not created by kukeon — left to the operator).
- Changes to the confirmation-prompt path (separate issue).
- Changes to `/usr/local/bin/kuke` / `kukeond` symlink handling (invariant-correct; out of scope).

## References

- `docs/cli-use-cases.md` § Bootstrap & teardown → "Full per-host teardown" invariants bullet 2 ("With `-y`, exit code 0 on success. Side effect: ... `/run/kukeon` and the configured run path (default `/opt/kukeon`) removed; ...").
- `scripts/dev-init.sh` (the kuke-attach smoke step that creates the surviving bind mount).
- Surfaced by `/test-cli "Bootstrap & teardown"` on `eminwux/kukeon` @ `main` (commit 7a1d5a7).

eminwux:
## Summary

- `kuke uninstall -y` on a post-attach host (`make dev-init`'s attach smoke step leaves `/run/kukeon/tty` bind-mounted) was exiting non-zero with `unlinkat /run/kukeon/tty: device or resource busy`. The unmount step that lets the subsequent rmdir of `/run/kukeon` succeed never existed, so a kukeon-owned bind mount pinned its own parent.
- Controller now enumerates kukeon-owned mountpoints under SocketDir and RunPath via `/proc/self/mounts`, releases each with `syscall.Unmount` (plain first, lazy MNT_DETACH fallback), and threads the per-mount outcomes into the report. The CLI renderer prints one `mount <path>: unmounted` row per successful release and one `mount <path>: still busy: <err>` row per residual mount so the operator does not have to grep `/proc/mounts` to find what blocked teardown.
- Residual mounts (kernel refuses both plain and lazy unmount) keep the historical non-zero exit and now name the surviving target verbatim in the wrapped error, matching the issue's AC.

## Implementation notes

- The MountReleaser is injected via `UninstallOptions` (same pattern as `DaemonStopper`) so unit tests can exercise the release step without provisioning real bind mounts. Production callers leave it nil.
- The /proc/self/mounts parser is split into a `parseMountsUnder(io.Reader, root)` helper to keep the file-format logic unit-testable; the procfs octal escape decoding (`\\040` → space, etc.) is covered alongside the substring-vs-prefix guard (`/run/kukeon2/...` must not match `/run/kukeon`).
- The enumerator refuses bare-root inputs (`""`, `"/"`, `"//"`) so a misconfigured caller cannot weaponize uninstall into a host-wide unmount sweep.
- This PR keeps the AC's scope: only mounts under SocketDir and RunPath are released. The `/.kukeon/bin/sbsh` mount that also appears in `mount | grep kukeon` survives — it is not under either teardown root and does not block the rmdir, so it is out of scope for #434.

## Test plan
- [x] `make test` — all unit tests pass, including the new `TestParseMountsUnder_*`, `TestUninstall_ReleasesBindMountsBeforeRmdir`, `TestUninstall_ResidualMountSurfacesInReportAndError`, and CLI renderer test `TestUninstall_RendersReleasedAndResidualMountRows`.
- [x] `make dev-init` — daemon-parity check matches the CLAUDE.md regression-guard output; the attach smoke leaves `/run/kukeon/tty` bind-mounted as in the issue repro.
- [x] `sudo ./kuke uninstall -y` on the post-`make dev-init` host — exits 0; output names the `mount /run/kukeon/tty: unmounted` row; `/run/kukeon` and `/opt/kukeon` removed.
- [x] Re-run `sudo ./kuke uninstall -y` on the now-clean host — exits 0, idempotent (`/run/kukeon: absent`, `/opt/kukeon: absent`).

Closes #434

### DIFF
--- a/cmd/kuke/uninstall/uninstall.go
+++ b/cmd/kuke/uninstall/uninstall.go
@@ -198,10 +198,30 @@ func printReport(cmd *cobra.Command, report controller.UninstallReport) {
 		fmt.Fprintln(out, "  filesystem + user/group cleanup skipped: residual containerd namespace prevented teardown")
 		return
 	}
+	renderMountReleases(out, report.SocketDirMounts)
 	fmt.Fprintf(out, "  - %s: %s\n", report.SocketDir, dirOutcome(report.SocketDirExists, report.SocketDirRemove))
+	renderMountReleases(out, report.RunPathMounts)
 	fmt.Fprintf(out, "  - %s: %s\n", report.RunPath, dirOutcome(report.RunPathExists, report.RunPathRemove))
 	fmt.Fprintf(out, "  - user %q: %s\n", report.UserName, accountOutcome(report.UserExisted, report.UserRemoved))
 	fmt.Fprintf(out, "  - group %q: %s\n", report.GroupName, accountOutcome(report.GroupExisted, report.GroupRemoved))
+}
+
+// renderMountReleases prints one indented row per kukeon-owned bind mount
+// the controller tried to release before the rmdir step. Skipped when no
+// mounts were enumerated under the parent directory so the common case
+// (host with no live attach sessions) stays terse.
+func renderMountReleases(out io.Writer, attempts []controller.MountReleaseAttempt) {
+	for _, a := range attempts {
+		if a.Err != nil {
+			// Residual mount: the operator needs the path + the kernel's
+			// reason verbatim so manual recovery is obvious.
+			fmt.Fprintf(out, "  - mount %s: still busy: %v\n", a.Target, a.Err)
+			continue
+		}
+		if a.Released {
+			fmt.Fprintf(out, "  - mount %s: unmounted\n", a.Target)
+		}
+	}
 }
 
 // outcomeAbsent labels every "we didn't find the resource" branch in the
--- a/cmd/kuke/uninstall/uninstall_test.go
+++ b/cmd/kuke/uninstall/uninstall_test.go
@@ -383,5 +383,49 @@ func TestUninstall_CleanupSkippedRendersSkipMarker(t *testing.T) {
 	}
 }
 
+// TestUninstall_RendersReleasedAndResidualMountRows pins the #434 renderer
+// half: the printed report has to surface both the successful unmount rows
+// (so an operator can see kuke did the work) and any residual mount rows
+// (so manual recovery is obvious without having to grep /proc/mounts).
+func TestUninstall_RendersReleasedAndResidualMountRows(t *testing.T) {
+	t.Cleanup(viper.Reset)
+
+	stuckErr := errors.New("synthetic EBUSY")
+	stub := &stubController{
+		uninstallFn: func(opts controller.UninstallOptions) (controller.UninstallReport, error) {
+			return controller.UninstallReport{
+				SocketDir:       opts.SocketDir,
+				SocketDirExists: true,
+				SocketDirRemove: false,
+				SocketDirMounts: []controller.MountReleaseAttempt{
+					{Target: "/run/kukeon/tty", Err: stuckErr},
+				},
+				RunPath:       "/opt/kukeon",
+				RunPathExists: true,
+				RunPathRemove: true,
+				RunPathMounts: []controller.MountReleaseAttempt{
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #450 Review — `fix: release kukeon-owned bind mounts before kuke uninstall rmdir`**

Targeted fix for #434 with the right shape: `MountReleaser` is injected via `UninstallOptions` so the new behavior is unit-testable, the procfs parser splits cleanly from the syscall layer, and the renderer surfaces both the success and the residual-mount rows so the operator does not have to grep `/proc/mounts`. The substring-vs-prefix guard, the bare-root refusal, and the octal-escape decoder all have tests. Commit is signed. Full test plan is checked, including `make dev-init` per CLAUDE.md.

A few non-blocking nits:

- **Style** `internal/controller/unmount.go:82-100`: `releaseMount` uses `if err == nil { return } else { ... }`; the conventional Go shape is to drop the `else` and let the success path return early, which un-nests the lazy-detach branch by one level.
- **Test helper** `internal/controller/uninstall_test.go:604`: `equalStringSlices` could be `reflect.DeepEqual(released, wantReleased)` — `reflect` is already imported in the sibling `unmount_test.go`, no new helper needed.

LGTM.

PR #451, opened 2026-05-13
feat: one-line installer at kukeon.io/install.sh with prereq checks

### LINKED ISSUE #62: install: one-line installer script with prereq checks (curl | bash)
## Context

Current install flow requires multiple manual steps — export OS/ARCH vars, curl the binary, chmod, install to /usr/local/bin, create the kukeond hardlink, then `sudo kuke init` separately. That's too much friction for discovery-phase visitors. Openclaw (https://openclaw.ai) launched with a single `curl -fsSL https://openclaw.ai/install.sh | bash` and it's one of the most visible reasons their adoption curve worked.

Kukeon can't match openclaw's zero-prereq install — we need **containerd running**, **cgroups v2**, **CNI plugins**, and **root privileges** — but we can collapse everything we control into one command and give a sharp, actionable error when a prereq is missing. Launch blocker for the agent-native promotion push (when Session #46 + Interactive UC2 #57 ship).

### One-line install

```bash
curl -fsSL https://kukeon.io/install.sh | bash
```

Script hosted via GitHub Pages on the existing `kukeon.io` domain (same source as mkdocs). Self-contained bash, no extra runtime deps.

### What the script does

1. **Detect OS/arch** — `uname -s` (must be Linux) and `uname -m` (amd64 / arm64). Exit with a clear message on macOS/Windows suggesting a Linux VM.
2. **Prereq checks (check only, never install):**
   - cgroups v2 mounted at `/sys/fs/cgroup` (check `stat -f -c %T`).
   - containerd socket at `/run/containerd/containerd.sock` and responsive (`ctr version` if available, else socket connectivity).
   - CNI plugin binaries present at `/opt/cni/bin` (look for at least `bridge`, `loopback`, `host-local`).
   - On any miss: print a distro-aware install hint and exit non-zero:
     ```
     ✗ containerd not found at /run/containerd/containerd.sock
       Install it:
         Debian/Ubuntu:  sudo apt-get install -y containerd
         Fedora/RHEL:    sudo dnf install -y containerd
         Arch:           sudo pacman -S containerd
       Then start:       sudo systemctl enable --now containerd
     ```
   - Explicitly **do not** auto-install prereqs. Too much blast radius on arbitrary user systems.
3. **Download the right binary** — `https://github.com/eminwux/kukeon/releases/download/<latest>/kuke-linux-<arch>`.
4. **Install** — `install -m 0755 kuke /usr/local/bin/kuke` with sudo (prompt if not running as root).
5. **Hardlink** — `sudo ln -f /usr/local/bin/kuke /usr/local/bin/kukeond`.
6. **Run `kuke init`** — with sudo. Handle the case where it's already initialized (idempotent).
7. **Print next steps** — a ready-to-run example:
   ```
   ✓ kukeon installed and initialized

   Try your first agent session:
     kuke apply -f https://kukeon.io/examples/claude-session.yaml
     kuke attach claude-demo

   Docs:  https://kukeon.io
   ```

### Script hygiene

- `set -euo pipefail` at the top.
- Pinned URL per release (resolve \"latest\" via the GitHub API, not a moving `latest` alias — avoids supply-chain surprises).
- Checksum verification against a published `.sha256` alongside each release asset.
- Dry-run mode: `curl ... | bash -s -- --check` runs only the prereq check, doesn't touch the system.
- Offline-friendly: log each step so the user can resume manually if any step fails.
- Do not pipe anything sensitive through the script.

### CI

- Makefile target `make install.sh` that emits the script from a template (so release versions can be substituted at build time).
- GitHub Actions job that runs the installer inside a minimal Ubuntu + Fedora container (with containerd preinstalled) to catch regressions.
- Publish `install.sh` to `kukeon.io/install.sh` via the existing mkdocs/GitHub Pages pipeline.

## Companion interface (follow-up scope)

The openclaw equivalent was a macOS companion app. For kukeon's audience (developers operating Linux hosts), the natural companion is a **Web UI** — the current README already references \"the future Web UI\" as a thin client alongside `kuke`. Recommend tracking this as a **separate issue** rather than bundling here, because:

- The installer is a pre-launch blocker; the Web UI is not.
- A Web UI is a multi-sprint effort that deserves its own scope doc.
- Bundling would stall the installer behind UI work that has different priority.

**Suggested follow-up issue:** `ui: browser-based companion for kuke attach and resource inspection` — `priority:B`, targeting post-launch v1.1.

If the maintainer prefers to bundle, move that scope in here and re-estimate.

## Acceptance

- [ ] `curl -fsSL https://kukeon.io/install.sh | bash` succeeds on a fresh Ubuntu 24.04 + containerd + CNI plugins host and leaves a working `kuke` + running kukeond.
- [ ] On a fresh host **without** containerd, the script exits with a distro-aware install hint and non-zero status. No files touched.
- [ ] On a fresh host **without** cgroups v2 or CNI plugins, same behavior — clear hint, no side effects.
- [ ] `bash install.sh --check` runs prereq checks only and exits without touching the system.
- [ ] Checksum verification against a released `.sha256` asset passes.
- [ ] After install, the final console message includes a copy-pasteable `kuke apply` + `kuke attach` example.
- [ ] CI runs the installer inside at least one minimal Linux image with prereqs preinstalled; failure blocks release.
- [ ] README Quick Start replaces the current 4-command sequence with the one-line install (keep the manual steps as a \"Manual install\" section below for users who want them).

## Non-goals

- No auto-installation of containerd, CNI, or cgroups v2 — prereq checks only.
- No Windows / macOS native install (Linux VM path is good enough; call it out).
- No package-manager packages (deb/rpm/brew) — separate, later work.
- No companion app / Web UI — split to a separate issue (see Companion interface note above).

## References

- Openclaw installer pattern: https://openclaw.ai (single `curl | bash`)
- Agent-native promotion plan (pending): blocked on #46 + #57
- Related positioning work: #59 (README/website reposition)

eminwux:
## Summary

- Adds `scripts/install.sh` — a self-contained bash installer that collapses the current 4-step manual flow (download → chmod → install → hardlink → `kuke init`) into a single `curl -fsSL https://kukeon.io/install.sh | bash` invocation. The script checks (but never installs) the host prerequisites — cgroups v2, containerd socket + optional `ctr version` responsiveness probe, CNI plugin binaries — and prints a distro-aware install hint on any miss before bailing without touching the filesystem.
- Adds `make install.sh` Makefile target that copies the canonical script into `docs/site/install.sh` so the existing mkdocs publish pipeline (gh-deploy → GitHub Pages → kukeon.io) serves the artifact at `https://kukeon.io/install.sh`. Both copies are committed and a CI sync job catches drift.
- Adds `.github/workflows/installer.yaml` — four jobs: a sync check, a `--check`-fails-without-containerd negative path inside a minimal `ubuntu:24.04` container, a `--check`-passes-with-prereqs positive path on the runner, and an end-to-end install (`KUKE_SKIP_INIT=1`) that downloads a real release binary and verifies it lands at `/usr/local/bin/kuke` + `/usr/local/bin/kukeond`.
- Extends `.github/workflows/release.yaml` to publish `kuke-linux-<arch>.sha256` companion files alongside the release binaries so the installer's checksum verification has something to verify against. Pre-issue-#62 releases (e.g. v0.4.0) lack these; users targeting those tags must pass `KUKE_SKIP_CHECKSUM=1` until the next release ships.
- Updates README Quick Start to lead with the one-line install and demotes the previous 4-command sequence into a `Manual install` subsection for air-gapped / version-pinned use.

## Implementation notes

- Version pinning is runtime-resolved by default — the script queries `https://api.github.com/repos/<owner>/<repo>/releases/latest` for `tag_name` (rather than following the mutable `/releases/latest` redirect) and bakes that exact tag into the download URL. `KUKE_VERSION=vX.Y.Z` env override skips the API call.
- Checksum verification is the default and hard-fails on a missing `.sha256` asset, naming `KUKE_SKIP_CHECKSUM=1` as the explicit opt-out so users hitting pre-checksum releases get a clear path forward rather than a silent skip.
- `kuke init` is idempotent in spirit but not guaranteed for arbitrary prior state, so the script's "is init done?" check is gated on `/run/kukeon/kukeond.sock` — present → already-bootstrapped, skip; absent → run `sudo kuke init`. `KUKE_SKIP_INIT=1` short-circuits the step entirely for the CI smoke and for users who want to drive init themselves.
- `containerd` responsiveness uses `ctr version` opportunistically — if `ctr` isn't on PATH we fall through to the socket-exists check rather than introducing a hard dependency on the upstream containerd CLI bundle.

## Out of scope (deferrals from the parent issue)

- The "Companion interface" Web UI mentioned in the issue body is explicitly deferred to a follow-up; not filed in this PR since the issue itself instructs the maintainer to do so.
- A fresh release with the `.sha256` assets — the workflow change here unlocks future releases but a maintainer has to cut the next tag for the installer's checksum path to exercise.

## Test plan

- [x] `bash scripts/install.sh --check` on the dev host with containerd running → exits 0, prints "cgroups v2 / containerd responsive / CNI plugins present".
- [x] `bash scripts/install.sh --check` with containerd stopped → exits 1, names the missing `containerd` socket and prints the Debian apt-get hint; no files written under `/usr/local/bin`.
- [x] `bash scripts/install.sh --help` → renders usage, exits 0.
- [x] `bash scripts/install.sh --bogus` → exits 2, prints "unknown argument: --bogus" then usage.
- [x] `KUKE_VERSION=v0.4.0 KUKE_SKIP_CHECKSUM=1 KUKE_SKIP_INIT=1 KUKE_INSTALL_PREFIX=/tmp/kuke-test bash scripts/install.sh` → downloads the real v0.4.0 release, installs `kuke` + hardlinked `kukeond`, `kuke version` reports `v0.4.0`. Clean exit, tmpdir trap removes the staging directory.
- [x] `make test` → all unit tests pass (no Go changes; sanity check that the Makefile target additions did not break the existing target graph).
- [x] `go build ./...` → green.
- [x] `make install.sh && git diff --exit-code docs/site/install.sh` → no drift; the committed copy matches the rendered output.
- [ ] `make dev-init` end-to-end smoke — not run. This PR does not touch the daemon, the build path, or `/opt/kukeon`; the daemon-parity regression guard in CLAUDE.md targets PRs that modify daemon-readable state, which this one doesn't.

Closes #62

### DIFF
--- a/.github/workflows/installer.yaml
+++ b/.github/workflows/installer.yaml
@@ -74,29 +74,15 @@ jobs:
             exit 1
           fi
 
-  # Positive path: a host with containerd + CNI plugins reachable. The
-  # GitHub-hosted ubuntu-latest runner already runs containerd (Docker is
-  # preinstalled and live), so we only need to install the CNI plugin bundle
-  # to satisfy /opt/cni/bin. Validates AC: "`bash install.sh --check` runs
+  # Positive path: the GitHub-hosted ubuntu-latest runner already runs
+  # containerd (Docker is preinstalled and live), which is the only daemon
+  # the installer checks for. Validates AC: "`bash install.sh --check` runs
   # prereq checks only and exits without touching the system."
   prereq-pass:
     name: --check passes with prereqs installed
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Install CNI plugins
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y --no-install-recommends containernetworking-plugins
-          # Debian/Ubuntu's containernetworking-plugins drops binaries at
-          # /usr/lib/cni; kukeon (matching CRI tools) expects /opt/cni/bin.
-          # Mirror the upstream layout the script checks for.
-          sudo mkdir -p /opt/cni/bin
-          for plugin in bridge loopback host-local; do
-            if [ -x "/usr/lib/cni/${plugin}" ]; then
-              sudo ln -sf "/usr/lib/cni/${plugin}" "/opt/cni/bin/${plugin}"
-            fi
-          done
       - name: Confirm containerd is running on the runner
         run: |
           if [ ! -S /run/containerd/containerd.sock ]; then
@@ -121,16 +107,6 @@ jobs:
       KUKE_SKIP_CHECKSUM: "1"
     steps:
       - uses: actions/checkout@v4
-      - name: Install CNI plugins (prereqs)
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y --no-install-recommends containernetworking-plugins
-          sudo mkdir -p /opt/cni/bin
-          for plugin in bridge loopback host-local; do
-            if [ -x "/usr/lib/cni/${plugin}" ]; then
-              sudo ln -sf "/usr/lib/cni/${plugin}" "/opt/cni/bin/${plugin}"
-            fi
-          done
       - name: Run installer (download + install, skip init)
         run: bash scripts/install.sh
       - name: Verify installed artifacts
--- a/.github/workflows/installer.yaml
+++ b/.github/workflows/installer.yaml
@@ -1,0 +1,140 @@
+name: Installer
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "scripts/install.sh"
+      - "docs/site/install.sh"
+      - "Makefile"
+      - ".github/workflows/installer.yaml"
+  push:
+    branches:
+      - main
+    paths:
+      - "scripts/install.sh"
+      - "docs/site/install.sh"
+      - "Makefile"
+      - ".github/workflows/installer.yaml"
+
+jobs:
+  # The canonical script lives at scripts/install.sh; docs/site/install.sh is
+  # the copy mkdocs publishes at https://kukeon.io/install.sh. `make install.sh`
+  # re-renders the copy; this job catches contributors who edited one and
+  # forgot to rerun the make target.
+  sync:
+    name: scripts/install.sh ↔ docs/site/install.sh in sync
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Re-render docs/site/install.sh
+        run: make install.sh
+      - name: Fail if rendered output drifted from committed copy
+        run: |
+          if ! git diff --exit-code docs/site/install.sh; then
+            echo "::error::docs/site/install.sh is stale. Run 'make install.sh' and commit the result."
+            exit 1
+          fi
+
+  # Negative path: minimal container with no containerd installed. The script
+  # must report which prereq is missing, print a distro-aware hint, and exit
+  # non-zero. Validates AC: "On a fresh host without containerd, the script
+  # exits with a distro-aware install hint and non-zero status."
+  prereq-missing:
+    name: --check fails cleanly without containerd
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu:24.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install bash + curl (script dependencies)
+        run: |
+          apt-get update
+          apt-get install -y --no-install-recommends bash curl ca-certificates
+      - name: Run installer --check on an unprepared host
+        run: |
+          set +e
+          out="$(bash scripts/install.sh --check 2>&1)"
+          rc=$?
+          echo "$out"
+          if [ "$rc" -eq 0 ]; then
+            echo "::error::expected non-zero exit on a host without prereqs"
+            exit 1
+          fi
+          # The hint must name containerd specifically — generic "something
+          # missing" output would defeat the AC.
+          if ! echo "$out" | grep -q 'containerd'; then
+            echo "::error::expected the failure message to reference containerd"
+            exit 1
+          fi
+          # No files should be touched on the prereq-failure path.
+          if [ -e /usr/local/bin/kuke ]; then
+            echo "::error::installer touched /usr/local/bin/kuke despite prereq failure"
+            exit 1
+          fi
+
+  # Positive path: a host with containerd + CNI plugins reachable. The
+  # GitHub-hosted ubuntu-latest runner already runs containerd (Docker is
+  # preinstalled and live), so we only need to install the CNI plugin bundle
+  # to satisfy /opt/cni/bin. Validates AC: "`bash install.sh --check` runs
+  # prereq checks only and exits without touching the system."
+  prereq-pass:
+    name: --check passes with prereqs installed
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install CNI plugins
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends containernetworking-plugins
+          # Debian/Ubuntu's containernetworking-plugins drops binaries at
+          # /usr/lib/cni; kukeon (matching CRI tools) expects /opt/cni/bin.
+          # Mirror the upstream layout the script checks for.
+          sudo mkdir -p /opt/cni/bin
+          for plugin in bridge loopback host-local; do
+            if [ -x "/usr/lib/cni/${plugin}" ]; then
+              sudo ln -sf "/usr/lib/cni/${plugin}" "/opt/cni/bin/${plugin}"
+            fi
+          done
+      - name: Confirm containerd is running on the runner
+        run: |
+          if [ ! -S /run/containerd/containerd.sock ]; then
+            echo "::error::containerd socket missing on the runner — runner image changed?"
+            exit 1
+          fi
+      - name: Run installer --check
+        run: bash scripts/install.sh --check
+
+  # End-to-end smoke: download a real release binary, install it to
+  # /usr/local/bin, and skip the kuke-init step (which would need root,
+  # cgroup writes, and a healthy containerd). Pinned to a known-good tag so
+  # the test doesn't drift with future releases. Checksum verification is
+  # skipped because pre-issue-#62 releases shipped without .sha256 assets;
+  # once a release lands with .sha256 published, drop the skip.
+  install-skip-init:
+    name: full install (KUKE_SKIP_INIT=1) against a real release
+    runs-on: ubuntu-latest
+    env:
+      KUKE_VERSION: v0.4.0
+      KUKE_SKIP_INIT: "1"
+      KUKE_SKIP_CHECKSUM: "1"
+    steps:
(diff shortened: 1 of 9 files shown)


eminwux:
Re-running failed CI jobs once: `--check passes with prereqs installed` and `full install (KUKE_SKIP_INIT=1) against a real release` (workflow run [25792250370](https://github.com/eminwux/kukeon/actions/runs/25792250370)). Will revisit when they finish; if they fail again the PR will be marked `ci-stalled`.

eminwux:
Fix-up pushed: `275786b` — skip the `ctr version` responsiveness probe when not running as root.

**Root cause.** The GitHub `ubuntu-latest` runner has Docker (and therefore `ctr` + containerd) preinstalled, but the containerd socket is mode 660 root:root. The previous prereq check ran `ctr version` unconditionally whenever `ctr` was on `PATH`, which fails with `permission denied` under the non-root runner user — and the script then misreports the EACCES as "stale socket / containerd is down." Both the `--check passes with prereqs installed` and `full install (KUKE_SKIP_INIT=1)` jobs failed against this.

**Fix.** Gate the responsiveness probe on `[ "$(id -u)" -eq 0 ]`. Non-root callers (CI dry run, or any user inspecting `--check`) get the "socket present" success line; root callers (`curl … | sudo bash`, which is the actual install path) still get the stale-socket detection. Stale-socket misreports happen at root anyway, so the probe is still in the right place.

Verified locally:
- non-root via `su -s /bin/bash nobody` → `✓ containerd socket present at /run/containerd/containerd.sock`
- root → `✓ containerd responsive at /run/containerd/containerd.sock (ctr version OK)`

`make install.sh` re-rendered the `docs/site/install.sh` mirror.

eminwux:
**PR #451 Review — `feat: one-line installer at kukeon.io/install.sh with prereq checks`**

Well-engineered installer: hard-fails-by-default checksum verification, distro-aware hints, idempotent init via the socket-exists check, hardlink (not symlink) for the argv[0] dispatch, and a four-job CI matrix that genuinely exercises the negative path, positive path, sync drift, and an end-to-end real-release install. Commits are GPG-verified. No credential/PII surface in error messages. Closes #62 cleanly.

A few non-blocking items worth surfacing:

- **Sequencing** `README.md:36-40`: the headline `curl -fsSL https://kukeon.io/install.sh | bash` will hard-fail against the current latest release (`v0.4.0`) because the script verifies checksums by default and v0.4.0 ships no `.sha256` assets — script exits at the verify step naming `KUKE_SKIP_CHECKSUM=1` as the bypass. PR body acknowledges this ("a maintainer has to cut the next tag for the installer's checksum path to exercise"), but the README goes live the instant this merges. Consider cutting v0.5.0 (which will include the release.yaml change) before merging this PR, or coordinating both in the same window, so the README's lead command isn't broken between merge and the next tag.
- **CodeQL `actions/missing-workflow-permissions`** `.github/workflows/installer.yaml`: four alerts (one per job) flag the absence of an explicit `permissions:` block. The repo already follows this convention in `release.yaml` and `mkdocs.yaml`; the new workflow only does checkout + script execution, so a single top-level `permissions:\n  contents: read` covers all four jobs and clears the alerts.
- **Nit — redundant `chmod +x`** `Makefile:140-141`: `cp` preserves the executable bit from `scripts/install.sh` (which is `100755`), so the trailing `chmod +x docs/site/install.sh` is a no-op. Same pattern at `scripts/install.sh:595` after `install -m 0755` already sets 0755.

Test plan: the unrunnable `make dev-init` smoke is correctly excluded — this PR doesn't touch the daemon, build path, or `/opt/kukeon`, so the regression guard doesn't apply.

LGTM — none of the above are blocking. The sequencing item is yours to time on merge; the CodeQL alerts and the `chmod +x` redundancy can land in the same PR or a follow-up.

eminwux:
Pushed fix-up: \`7fdb023\` — \`refactor: drop host CNI plugin prereq check; document --no-daemon needs\`.

**Rationale.** \`kukeond\` bundles CNI plugins in its container image (\`Dockerfile\` cni-plugins stage) and \`kukeondCellDoc\` (\`internal/controller/bootstrap.go:464-471\`) only bind-mounts state directories (\`/opt/cni/net.d\`, \`/var/lib/cni\`, \`/opt/cni/cache\`) from the host — not the plugin binaries. The standard daemon-mediated install (\`curl | sudo bash\` → \`kuke init\`) never reads host \`/opt/cni/bin\`. Requiring it in the prereq check forced every operator to apt-install a package the daemon path doesn't use.

Host \`/opt/cni/bin\` is genuinely needed only by \`--no-daemon\` workflows, which run the controllers in-process and shell out to plugin binaries on the host. That flag is also slated for removal from creation commands in a future release.

**Changes.**

- \`scripts/install.sh\`: removed \`check_cni()\` and its call site. Left a comment in \`run_prereqs\` pointing at the architectural reason and the docs reference.
- \`docs/site/cli/commands.md\`: new \`--no-daemon\` host-prerequisites section under the global-flags table, with a transitional warning admonition.
- \`docs/site/install/prerequisites.md\`: numbered prereq list went from three to two; the \`CNI plugins\` section is rewritten as scope-limited to \`--no-daemon\`, with a link to the canonical reference in \`commands.md\`.
- \`README.md\` + \`docs/site/getting-started.md\`: dropped the misleading "CNI plugins available on the host" bullet from Quick Start.
- \`.github/workflows/installer.yaml\`: dropped the now-dead "Install CNI plugins" prep steps from \`prereq-pass\` and \`install-skip-init\` (script no longer asserts on them).
- \`docs/site/install.sh\`: re-rendered via \`make install.sh\`.

**Verified locally.**

- \`bash scripts/install.sh --check\` as root → \`✓ all prerequisites satisfied\` (no CNI line).
- \`su -s /bin/bash nobody -c 'bash scripts/install.sh --check'\` → \`✓ all prerequisites satisfied\` (no CNI line).
- \`make test\` passes.

eminwux:
**PR #451 Re-review — follow-up commit `7fdb023`**

Verified the architectural rationale: `Dockerfile` `cni-plugins` stage (lines 21-44) downloads v1.5.1 plugins into the kukeond image's `/opt/cni/bin`, and `kukeondCellDoc` (`internal/controller/bootstrap.go`) bind-mounts only `/opt/cni/net.d`, `/var/lib/cni`, `/opt/cni/cache` from the host — never `/opt/cni/bin`. Dropping `check_cni()` is correct, and the `--no-daemon` carve-out is documented at `docs/site/cli/commands.md` with cross-links from `prerequisites.md`. Script/mirror parity preserved (`make install.sh` rendered both copies). `distro_family` still has a live caller via `install_hint` → `check_containerd`, so no orphans.

Prior non-blocking items (sequencing, CodeQL workflow permissions, redundant `chmod +x`) remain at the author's discretion — not introduced or worsened by this delta.

LGTM — `ready-to-merge`.

PR #453, opened 2026-05-13
fix: route cobra cmd.Print* output to stdout instead of stderr

### LINKED ISSUE #436: cli: route normal output to stdout instead of stderr
## Context

Most \`kuke\` subcommands write their normal (non-error) output to **stderr** instead of stdout. The root cause is in \`cmd/kuke/kuke.go\` \`NewKukeCmd()\`: it never calls \`cmd.SetOut(...)\` or \`cmd.SetErr(...)\`. Cobra's \`cmd.Print\` / \`cmd.Println\` / \`cmd.Printf\` write to \`c.OutOrStderr()\`, which falls back to \`os.Stderr\` when \`SetOut\` is unset. There are **145** \`cmd.Print*\` call sites across \`cmd/kuke/\` — \`init\`, \`apply\`, \`delete\`, \`refresh\`, \`image load\`/\`image delete\`, \`daemon start\`/\`stop\`/\`kill\`/\`reset\`/\`restart\`, \`stop\`/\`kill\`/\`delete\` resource subcommands, \`uninstall\`, the \`get\` table + empty-list branches in \`cmd/kuke/get/shared/shared.go:105,132,144,159\`, and the shared \`PrintJSONOrYAML\` helper at \`cmd/kuke/shared/output.go:49\`. All of them silently route to stderr.

Empirical confirmation (same command, different streams):

| command                                  | stdout       | stderr             |
| ---------------------------------------- | ------------ | ------------------ |
| \`kuke version\`                         | \`v0.4.0-…\` | *(empty)*          |
| \`kuke --help\`                          | full help    | *(empty)*          |
| \`kuke get realms --no-daemon\`          | *(empty)*    | \`No realms found.\` |
| \`kuke get realms --no-daemon -o json\`  | \`[]\`       | *(empty)*          |
| \`kuke daemon stop\` (not initialized)   | *(empty)*    | \`Error: …\`       |

\`version\` is correct because it bypasses cobra and uses \`fmt.Fprintln(os.Stdout, …)\` directly (\`cmd/kuke/version/version.go:52\`). \`get -o json|yaml\` is correct because it writes via \`json.NewEncoder(os.Stdout)\` / \`yaml.NewEncoder(os.Stdout)\` directly (\`cmd/kuke/get/shared/shared.go:89,97\`). The table branch of \`get\` and every other \`cmd.Print\` caller is wrong — \`kuke get realms\` table output is on stderr, but \`-o json\` is on stdout for the very same command.

This breaks the standard Unix contract: \`kuke get foo | xargs …\` or \`out=\$(kuke …)\` will silently get no data, and CI/scripts using exit-code-only checks won't see informational output without redirecting stderr.

## Proposal

Wire the root cobra command to the real streams once, in \`cmd/kuke/kuke.go\` \`NewKukeCmd()\`:

\`\`\`go
cmd.SetOut(os.Stdout)
cmd.SetErr(os.Stderr)
\`\`\`

Cobra propagates these to all subcommands via \`getOut\`/\`getErr\`, so the 145 call sites flip to stdout without per-site edits. Errors returned from \`RunE\` continue to print on stderr (cobra's built-in path).

Audit the two existing \`cmd.ErrOrStderr()\` callers — \`cmd/kuke/run/run.go:328\` (a user warning) and \`cmd/kuke/image/load.go:139\` (docker's own stderr passthrough). Both look stderr-appropriate; confirm during the change.

## Acceptance criteria

- [ ] \`cmd/kuke/kuke.go\` \`NewKukeCmd()\` calls \`cmd.SetOut(os.Stdout)\` and \`cmd.SetErr(os.Stderr)\` before returning the command.
- [ ] \`kuke get realms --no-daemon\` table output (including the \"No realms found.\" empty case) goes to **stdout**, not stderr.
- [ ] \`kuke get realms --no-daemon -o json\` output remains on stdout (regression guard).
- [ ] \`kuke init\`, \`kuke apply\`, \`kuke delete\`, \`kuke refresh\`, \`kuke image load\`, and \`kuke daemon {start,stop,kill,reset,restart}\` write their progress/result lines to stdout. Spot-check at least one command per group manually.
- [ ] Error messages (\`Error: …\` from a returned \`RunE\` error, plus the \`unknown command\` cobra path) still appear on stderr.
- [ ] \`cmd/kuke/run/run.go:328\` and \`cmd/kuke/image/load.go:139\` are audited; \`cmd.ErrOrStderr()\` is retained where stderr is intentional.
- [ ] PR description calls out the behavior change so users/CI grepping kuke output via \`2>&1\` know to adjust.

## Notes

- Cobra source for the default: \`func (c *Command) Print(i ...interface{}) { fmt.Fprint(c.OutOrStderr(), i...) }\` and \`OutOrStderr()\` returns \`getOut(os.Stderr)\` — i.e., stderr when \`SetOut\` is unset.
- Breaking-change surface: any user/CI that currently relies on capturing kuke's informational output via stderr (e.g., \`kuke get … 2>&1 | grep\`) will see output move streams. This is the correct behavior, but worth a release-note line.

eminwux:
## Summary
- Wire `cmd.SetOut(os.Stdout)` / `cmd.SetErr(os.Stderr)` once in `cmd/kuke/kuke.go` `NewKukeCmd()`. Cobra propagates these to every subcommand via `getOut`/`getErr`, so the ~145 `cmd.Print*` call sites across `cmd/kuke/` (init, apply, delete, refresh, image, daemon, get table rows, etc.) flip to stdout without per-site edits.
- Without this, `cmd.Print*` falls back to `c.OutOrStderr()` → `os.Stderr`, breaking the Unix contract: `out=$(kuke get realms)` got nothing on stdout while `-o json` did, since the JSON branch writes via `json.NewEncoder(os.Stdout)` directly.
- Existing intentional stderr writes — `cmd/kuke/run/run.go:328` (`--rm` cleanup warning) and `cmd/kuke/image/load.go:139` (docker's own stderr passthrough) — were audited and retained.

## Behavior change for users / CI
Any script that grepped kuke's informational output via `2>&1` (or that only redirected stderr to capture progress) will now see that output on stdout instead. This is the correct Unix behavior, but worth a release-note line.

## Test plan
- [x] `make test` — full unit suite green locally.
- [x] `go build ./...` — clean compile.
- [x] `gofmt -l` and `go vet ./cmd/...` — clean.
- [x] New `TestNewKukeCmdStreams` in `cmd/kuke/kuke_test.go` asserts `cmd.OutOrStdout() == os.Stdout` and `cmd.ErrOrStderr() == os.Stderr` — regression guard against the SetOut/SetErr pair being dropped.
- [x] Smoke: `./kuke get realms --no-daemon` → `No realms found.` on **stdout** (was stderr); `./kuke get realms --no-daemon -o json` → `[]` on **stdout** (regression guard); `./kuke version` → on stdout (no regression); `./kuke daemon stop` error → on **stderr**; unknown command (`./kuke notacommand`) → cobra error on **stderr**; `./kuke --help` and `./kuke init --help` → on stdout.
- [ ] End-to-end `make dev-init` daemon-parity smoke — not run: this change is confined to root cobra stream wiring (`cmd/kuke/kuke.go` constructor) and does not touch the build path, daemon, or anything under `/opt/kukeon`, so the bind-mount regression guard does not apply. Inspecting the daemon's two-realm output is itself just `cmd.Print*` table rendering, which the routing smoke above already exercises.

Closes #436

### DIFF
--- a/cmd/kuke/kuke.go
+++ b/cmd/kuke/kuke.go
@@ -119,6 +119,14 @@ func NewKukeCmd() (*cobra.Command, error) {
 		},
 	}
 
+	// Wire the root to real streams so cobra's cmd.Print* helpers route
+	// non-error output to stdout. Without this, cobra's OutOrStderr() falls
+	// back to os.Stderr and every cmd.Print* in the tree lands on stderr,
+	// breaking `out=$(kuke …)` and `kuke … | xargs …`. Subcommands inherit
+	// these via cobra's getOut/getErr.
+	cmd.SetOut(os.Stdout)
+	cmd.SetErr(os.Stderr)
+
 	if err := SetupKukeCmd(cmd); err != nil {
 		return nil, fmt.Errorf("failed to setup kukeon command: %w", err)
 	}
--- a/cmd/kuke/kuke_test.go
+++ b/cmd/kuke/kuke_test.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"os"
 	"strings"
 	"testing"
 
@@ -418,5 +419,24 @@ func TestSetPersistentLoggingFlagsViperBinding(t *testing.T) {
 	}
 	if !viper.GetBool(config.KUKEON_ROOT_VERBOSE.ViperKey) {
 		t.Errorf("viper binding mismatch for verbose: got false, want true")
+	}
+}
+
+// TestNewKukeCmdStreams guards against cobra's OutOrStderr default — if
+// SetOut/SetErr are dropped from NewKukeCmd, every cmd.Print* call in the
+// subcommand tree silently routes to stderr. Issue #436.
+func TestNewKukeCmdStreams(t *testing.T) {
+	t.Cleanup(viper.Reset)
+
+	cmd, err := kuke.NewKukeCmd()
+	if err != nil {
+		t.Fatalf("NewKukeCmd() error = %v, want nil", err)
+	}
+
+	if got := cmd.OutOrStdout(); got != os.Stdout {
+		t.Errorf("OutOrStdout() not wired to os.Stdout: got %T", got)
+	}
+	if got := cmd.ErrOrStderr(); got != os.Stderr {
+		t.Errorf("ErrOrStderr() not wired to os.Stderr: got %T", got)
 	}
 }


eminwux:
**PR #453 Review — `fix: route cobra cmd.Print* output to stdout instead of stderr`**

Clean, minimal fix at the right layer — `SetOut`/`SetErr` on the root propagates to every subcommand via cobra's `getOut`/`getErr` chain, so the ~145 `cmd.Print*` sites flip uniformly without a sweeping diff. The regression test (`TestNewKukeCmdStreams`) pins the wiring so a future refactor can't silently drop it. No sensitive-data exposure (this is a routing change, not new log text). Commits are signed.

Reverse-audited the call sites the patch implicitly redirects: the two explicit-stderr sites the author called out (`cmd/kuke/run/run.go:328`, `cmd/kuke/image/load.go:139`) both write via `cmd.ErrOrStderr()` and stay on stderr. Per-resource failure lines in `apply/apply.go:113-117`, `delete/delete.go:211-215`, and `refresh/refresh.go:89-95` are structured result output (created/updated/failed in the same stream) — keeping them on stdout under the PR is the right call; interleaving with stderr would scramble the report. The exit-non-zero path goes through cobra's err sink as usual.

Test plan: the deferred `make dev-init` daemon-parity smoke is correctly carved out — this PR doesn't touch the build path, the daemon, or `/opt/kukeon`, so the bind-mount regression guard doesn't apply. The author's smoke matrix already exercises stdout/stderr routing across `get`/`version`/`daemon stop`/unknown-command/`--help`.

LGTM.

PR #454, opened 2026-05-13
fix: filter runner-synthesized state from kuke run/apply divergence

### LINKED ISSUE #437: cli: `kuke run -f` reports phantom divergence against the same file (synthesized root container counted in spec.containers)
## Context

`docs/cli-use-cases.md` § "Run a one-off cell from YAML" guarantees:

> Re-running `kuke run -f` against an existing cell whose on-disk spec **diverges** from the file is **refused**, not silently updated. The error message points the operator at `kuke apply -f` for the update path.

The wording implies the case where the on-disk spec does **not** diverge from the file should not be refused — that is the only way the documented `kuke run -f` invocation in the Sequence has an idempotent path at all.

Observed: re-running the **unmodified** `docs/examples/hello-world.yaml` against the cell it just created is refused, citing a container-count divergence between the runtime-augmented on-disk state and the user's YAML:

```
$ sudo ./kuke run -f docs/examples/hello-world.yaml -d
Cell "hello-world" (realm "default", space "default", stack "default")
  - metadata: created
  - cgroup: created
  - root container: created
  - container "web": created
  - containers: started
$ echo "exit=$?"
exit=0

$ sudo ./kuke run -f docs/examples/hello-world.yaml -d
Error: cell "hello-world" exists with diverging spec (spec.containers (count: actual=2, desired=1)); use `kuke apply -f` to update
$ echo "exit=$?"
exit=1
```

`actual=2, desired=1` is the synthesized root container being counted toward `spec.containers` on the on-disk side but not on the YAML side. The user's YAML lists one container (`web`); the runtime stores `root` + `web` and compares the augmented total to the user-supplied count. Result: every `kuke run -f` against an existing cell with the **same** file produces a phantom-divergence error and points the operator at `kuke apply -f`, even though the YAML has not changed.

This makes `kuke run -f` non-idempotent against its own output, and the error message misleads the operator into thinking the YAML changed.

## Proposal

Exclude the synthesized root container from `spec.containers` count/diff comparison in `kuke run -f` divergence detection — compare the **user-supplied** container set on both sides, or fold the root container into both representations consistently. Either fix makes re-running the unmodified YAML succeed (idempotent path) while real edits (added/removed/renamed user containers, image changes, command/args drift) still trip the existing refusal.

If the project's stance is that `kuke run -f` is **never** idempotent for the same cell (always run via `kuke apply -f` after the first create), the doc invariant should be reworded to say so explicitly, and the error should drop the "diverging spec" framing in favor of "cell already exists; use `kuke apply -f`".

## Acceptance criteria

- [ ] `sudo kuke run -f docs/examples/hello-world.yaml -d` on a host where the same file was used to create the cell succeeds (exit 0) and reports `already existed` / `unchanged` rather than a diverging-spec error.
- [ ] A genuinely diverging YAML (image change, added container, removed container, changed command) still trips the refusal with a message that names the actual diverging field — not the synthesized root container.
- [ ] `kuke apply -f` against a same-file re-run continues to report `unchanged` (no spurious "updated" / "root container recreated" rows when nothing changed). _(See note: today `kuke apply -f` against the unchanged file reports `Cell hello-world: updated\n  - root container recreated`, which is the symmetric symptom on the apply side.)_
- [ ] The matching invariant in `docs/cli-use-cases.md` § "Run a one-off cell from YAML" passes when exercised by `/test-cli "Workload lifecycle"`.

## Out of scope

- Changes to the `kuke apply -f` update semantics for genuine spec changes (only the "unchanged file" path is in scope here, and only to the extent it shares the root-container-count comparison bug with `run`).
- The `-f` / `-p` mutual-exclusion logic (verified passing).
- The `--container` + `-d` and `--rm` + `--no-daemon` mutual exclusions (both verified passing).

## References

- `docs/cli-use-cases.md` § Workload lifecycle → "Run a one-off cell from YAML" invariants bullet 3 ("Re-running `kuke run -f` against an existing cell whose on-disk spec **diverges** from the file is **refused**, not silently updated.").
- `docs/examples/hello-world.yaml` (the example the doc names verbatim).
- Surfaced by `/test-cli "Workload lifecycle"` on `eminwux/kukeon` @ `main` (commit 7a1d5a7).

eminwux:
## Summary

- \`kuke run -f docs/examples/hello-world.yaml -d\` re-run against the cell it just created refused with a phantom \`spec.containers (count: actual=2, desired=1)\` error. The on-disk side counted the runner-synthesized root container (\`Root: true\`) toward \`spec.containers\` while the user YAML (which omits the root) did not. Fixed by excluding \`Root: true\` entries from the count/id-set comparison on both sides in \`divergedFields\` — same pattern \`internal/controller/apply/diff.go\` already uses for child-container diffs.
- \`kuke apply -f\` against the same unchanged file reported \`Cell <name>: updated\\n  - root container recreated\` because \`DiffCell\` hit \`desiredRoot == nil && actualRoot != nil\` and tripped \`RecreateCell\`. Fixed by treating the synthesized root as implementation detail (the runner always produces one) rather than as a removed user field.
- Removing the recreate path exposed two adjacent bugs in the same idempotent flow that had been masked by it: \`UpdateCell\` refused to remove the root from desired (it never should — runner synthesizes it), and \`DiffCell\`'s labels comparison flagged drift between the user's empty \`Labels\` and the canonical \`*.kukeon.io\` set the runner injected on create. Both are runner-managed state, same principle as the root container. \`UpdateCell\` now preserves the synthesized root through orphan detection and the rebuild loop; \`DiffCell\` filters \`*.kukeon.io\`-suffixed keys from the label comparison (canonical keys are controller-owned, not user-authored).

## Test plan

- [x] \`make test\` — all unit tests pass.
- [x] \`make dev-init\` — daemon parity check passes (the trailing \`kuke attach\` smoke step fails on \`registry.eminwux.com\` DNS in this dev container, unrelated to the diff).
- [x] \`sudo ./kuke run -f docs/examples/hello-world.yaml -d\` twice against the same cell: first run creates, second run returns exit 0 and reports \`already existed\` (AC #1).
- [x] Real-divergence variant (YAML with an added user container) still trips \`spec.containers (count: actual=1, desired=2)\` and points to \`kuke apply -f\` (AC #2).
- [x] \`sudo ./kuke apply -f docs/examples/hello-world.yaml\` against the existing cell reports \`Cell hello-world: unchanged\`. Repeated invocations also report \`unchanged\` (AC #3).
- [x] Post-apply \`kuke get cell hello-world -o yaml\` confirms the canonical \`cell.kukeon.io / realm.kukeon.io / space.kukeon.io / stack.kukeon.io\` labels are preserved (no clobber).
- [x] New unit tests:
  - \`cmd/kuke/run/run_test.go\`: \`TestRun_ExistingCell_SynthesizedRoot_DoesNotDiverge\`, \`TestRun_ExistingCell_SynthesizedRoot_RealDivergenceStillCaught\`.
  - \`internal/controller/apply/diff_test.go\`: \`TestDiffCell_SynthesizedRoot_NoChange\`, \`TestDiffCell_ManagedLabels_NoChange\`, \`TestDiffCell_UserLabels_StillDetectsDrift\`.
- [ ] \`make e2e\` — not run (e2e harness wants a clean host and times out on this dev container's external image pulls; unit + dev-init smoke cover the changed paths).

## Notes

- The label-filter change goes slightly beyond the strict "root-container-count comparison" framing in the issue's "Out of scope" section, but AC #3 explicitly requires \`apply -f\` to report \`unchanged\` on a same-file re-apply. Without filtering the runner-managed canonical labels, the diff still fires \`metadata.labels changed\`, falls through to \`UpdateCell\`, and (pre-fix) clobbers the canonical labels with the user's empty map. Treating \`*.kukeon.io\`-suffixed keys as controller-managed is the smallest surgical change that satisfies the AC and is the same principle the existing apply-side child-container diff already applies to \`Root: true\` entries.
- A genuine spec change that triggers \`UpdateCell\` still overwrites \`existing.Metadata.Labels\` with \`desired.Metadata.Labels\` (\`update_cell.go:42\`), which can drop the canonical labels in the genuine-update path. Out of scope per the issue; filing a follow-up.

Closes #437

### DIFF
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -517,6 +517,15 @@ func pickLocation(fromDoc string, kv *config.Var) string {
 // cell as divergent on the very next `kuke run -f` invocation. Per the AC, a
 // genuine spec rewrite — adding/removing containers, swapping the root, moving
 // realm/space/stack — must route through `kuke apply -f` instead.
+//
+// Root containers are filtered out of the count/id-set comparison on both
+// sides. The runner synthesizes a root container during create if the YAML
+// omitted one (`docs/examples/hello-world.yaml` is the canonical case), so a
+// raw count comparison would treat every re-run of an existing file as
+// divergent (actual=root+web, desired=web) and route the operator to
+// `kuke apply -f` even though nothing changed. Comparing only the
+// user-supplied (non-root) entries restores the idempotent path while still
+// catching real adds/removes/renames among the user containers.
 func divergedFields(actual, desired v1beta1.CellSpec) []string {
 	var diffs []string
 
@@ -533,19 +542,23 @@ func divergedFields(actual, desired v1beta1.CellSpec) []string {
 	if len(actual.Containers) == 0 {
 		return diffs
 	}
-	if len(actual.Containers) != len(desired.Containers) {
+
+	actualUser := nonRootContainers(actual.Containers)
+	desiredUser := nonRootContainers(desired.Containers)
+
+	if len(actualUser) != len(desiredUser) {
 		diffs = append(diffs, fmt.Sprintf(
 			"spec.containers (count: actual=%d, desired=%d)",
-			len(actual.Containers), len(desired.Containers),
+			len(actualUser), len(desiredUser),
 		))
 		return diffs
 	}
 
-	desiredByID := make(map[string]v1beta1.ContainerSpec, len(desired.Containers))
-	for _, c := range desired.Containers {
+	desiredByID := make(map[string]v1beta1.ContainerSpec, len(desiredUser))
+	for _, c := range desiredUser {
 		desiredByID[c.ID] = c
 	}
-	for _, ac := range actual.Containers {
+	for _, ac := range actualUser {
 		dc, ok := desiredByID[ac.ID]
 		if !ok {
 			diffs = append(diffs, fmt.Sprintf("spec.containers[%q] (missing in file)", ac.ID))
@@ -556,6 +569,19 @@ func divergedFields(actual, desired v1beta1.CellSpec) []string {
 		}
 	}
 	return diffs
+}
+
+// nonRootContainers returns the user-supplied subset of cs — entries where
+// Root is false. Used to exclude the runner-synthesized root from divergence
+// comparison; see divergedFields.
+func nonRootContainers(cs []v1beta1.ContainerSpec) []v1beta1.ContainerSpec {
+	out := make([]v1beta1.ContainerSpec, 0, len(cs))
+	for _, c := range cs {
+		if !c.Root {
+			out = append(out, c)
+		}
+	}
+	return out
 }
 
 func resolveClient(cmd *cobra.Command) (kukeonv1.Client, error) {
--- a/cmd/kuke/run/run_test.go
+++ b/cmd/kuke/run/run_test.go
@@ -84,6 +84,27 @@ spec:
         - "3600"
 `
 
+// cellYAMLUserContainersOnly mirrors docs/examples/hello-world.yaml: the user
+// declares only their workload container(s), and relies on the runner to
+// synthesize the root container during create. Used by the same-file re-run
+// regression below (issue #437).
+const cellYAMLUserContainersOnly = `apiVersion: v1beta1
+kind: Cell
+metadata:
+  name: my-cell
+spec:
+  id: my-cell
+  realmId: my-realm
+  spaceId: my-space
+  stackId: my-stack
+  containers:
+    - id: web
+      image: registry.eminwux.com/busybox:latest
+      command: sleep
+      args:
+        - "3600"
+`
+
 const multiDocYAML = validCellYAML + "\n---\n" + validCellYAML
 
 const realmDocYAML = `apiVersion: v1beta1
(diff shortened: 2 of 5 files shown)


eminwux:
**PR #454 Review — `fix: filter runner-synthesized state from kuke run/apply divergence`**

The synthesized-root and managed-label filters are the right shape for this — all three new helpers (`nonRootContainers`, `filterManagedLabels`, `userLabelsEqual`) are wired up, the `.kukeon.io` suffix convention is already established in `internal/consts/consts.go`, and the `preservedRoot` pointer in `UpdateCell` is safe (no re-slicing of `existing.Spec.Containers` between the capture at `update_cell.go:80` and the dereference at `update_cell.go:146`). Per-commit signatures verified, no sensitive-data leaks, and the test pair on both sides (synthesized-root no-divergence + real-divergence guardrail) cleanly bounds the new behavior.

The skipped `case desiredRoot == nil && actualRoot != nil` is an intentional asymmetry vs. the still-detected "added" case, but it's a defensible no-op for the user-explicit-removal scenario too — the runner re-synthesizes regardless via `provision.go:2087`'s `ensureCellRootContainerSpec`.

No blocking items. `make e2e` deferred per documented host constraint — please run pre-merge if your environment supports it.

LGTM.

PR #456, opened 2026-05-13
fix: preserve canonical *.kukeon.io labels through UpdateCell etc.

### LINKED ISSUE #455: apply: kuke apply -f clobbers controller-managed *.kukeon.io labels on UpdateCell
## Context

`internal/controller/runner/update_cell.go:42` unconditionally overwrites the cell's label map with whatever the user's desired doc carries:

```go
// Update compatible metadata fields
existing.Metadata.Labels = desired.Metadata.Labels
```

The user YAML (e.g. `docs/examples/hello-world.yaml`) does not author the `*.kukeon.io` canonical labels — those are injected by `internal/controller/create_cell.go` (`KukeonRealmLabelKey`, `KukeonSpaceLabelKey`, `KukeonStackLabelKey`, `KukeonCellLabelKey`) at create time. Any genuine apply that reaches `UpdateCell` therefore strips them. Subsequent operations that key off those labels (refresh paths, get-by-label filtering, anything wired through `consts.KukeonCellLabelKey`) silently lose their parent association on the cell.

This bug was surfaced and worked around at the *diff* layer in PR #454 / issue #437 — `DiffCell` now filters `*.kukeon.io`-suffixed keys from the label comparison, so the same-file re-apply path no longer reaches `UpdateCell`. The clobber still fires on any genuine spec change (added container, edited image, etc.) that legitimately routes through `UpdateCell`. The current fix prevents the *symptom* on the idempotent path; the underlying primitive is still destructive.

## Repro

worktree: `eminwux/kukeon @ main` *after* PR #454 lands (before the merge the symptom is masked by the earlier `RecreateCell` path).

```bash
sudo ./kuke run -f docs/examples/hello-world.yaml -d
sudo ./kuke get cell hello-world --realm default --space default --stack default -o yaml | head -10
# observe: labels include cell.kukeon.io / realm.kukeon.io / space.kukeon.io / stack.kukeon.io

# now make a genuine change that routes through UpdateCell — e.g. add a label to the YAML, OR add a new user container.
# minimal: copy hello-world.yaml, add `labels: { env: prod }` under metadata, save as hello-world-edited.yaml.
sudo ./kuke apply -f hello-world-edited.yaml
sudo ./kuke get cell hello-world --realm default --space default --stack default -o yaml | head -10
# observe: labels: { env: prod } — the four canonical *.kukeon.io keys are gone.
```

## Observed

Pre-apply:

```yaml
metadata:
  name: hello-world
  labels:
    cell.kukeon.io: hello-world
    realm.kukeon.io: default
    space.kukeon.io: default
    stack.kukeon.io: default
```

Post genuine-change apply:

```yaml
metadata:
  name: hello-world
  labels:
    env: prod
```

The canonical labels do not get re-injected by `UpdateCell` (create-time injection in `create_cell.go` only runs on `CreateCell`).

## Root cause

`internal/controller/runner/update_cell.go:42`:

```go
existing.Metadata.Labels = desired.Metadata.Labels
```

The line treats `desired.Metadata.Labels` as authoritative for the *entire* label map. The canonical `*.kukeon.io` keys are controller-managed (parallel to the synthesized root container, which `update_cell.go` *does* now preserve through orphan detection after PR #454). The labels primitive does not.

## Suggested fix

Apply the same "controller-managed state survives the user's empty map" rule already encoded in `DiffCell` (PR #454's `filterManagedLabels`). At `update_cell.go:42`, merge desired labels into existing while preserving keys with the `.kukeon.io` suffix:

```go
existing.Metadata.Labels = mergePreservingManaged(existing.Metadata.Labels, desired.Metadata.Labels)
```

where `mergePreservingManaged` carries forward any `*.kukeon.io` key from `existing` that `desired` did not author. The helper can live alongside `filterManagedLabels` in `internal/controller/apply/diff.go` (or move both to a shared `internal/consts` helper — the suffix definition is already in `internal/consts/consts.go`). Update `UpdateRealm`, `UpdateSpace`, `UpdateStack`, `UpdateContainer` symmetrically — every resource the runner injects canonical labels into has the same exposure.

A reasonable AC:
- [ ] `kuke apply -f` against a YAML that adds one user-authored label preserves the canonical `*.kukeon.io` set on `Cell`, `Realm`, `Space`, `Stack`, `Container`.
- [ ] A user-authored label whose key is *not* `.kukeon.io`-suffixed still wins over an absent desired value (no behavior change for user labels).
- [ ] If the user explicitly authors a `*.kukeon.io` key (e.g. overrides `cell.kukeon.io`), the user wins — matching the existing create-time "if not exists" semantics in `create_cell.go`.

## Out of scope

- Broader rework of apply update semantics (this is the narrow label-merge primitive only).
- Any change to `DiffCell` label comparison — that was settled in PR #454.

## References

- Worked around at the diff layer in PR #454 (issue #437). The PR body explicitly defers this follow-up.
- Canonical label keys: `internal/consts/consts.go:61-65`.
- Injection sites: `internal/controller/create_cell.go:106-117` (and `create_realm.go`, `create_space.go`, `create_stack.go`).
- Clobber site: `internal/controller/runner/update_cell.go:42`.

eminwux:
## Summary

- `internal/controller/runner/update_{cell,realm,space,stack}.go` unconditionally clobbered `existing.Metadata.Labels` with `desired.Metadata.Labels`. Because user YAMLs never author the canonical `*.kukeon.io` keys, every genuine spec change routed through `UpdateCell`/`UpdateRealm`/`UpdateSpace`/`UpdateStack` stripped them. PR #454 worked around the same root cause at the *diff* layer for the idempotent re-apply case; the underlying primitive was still destructive on real updates.
- Add `mergeManagedLabels(existing, desired)` (new `internal/controller/runner/labels.go`): `desired` keys still win, plus any `.kukeon.io`-suffixed key already on the resource that `desired` did not author is carried forward. Mirrors the create-time "if not exists" semantics in `create_cell.go` so an explicit user-authored `cell.kukeon.io` override still wins (AC #3).
- Wire the helper into the four `Update*` runners that hit canonical-label-carrying resources. `UpdateContainer` is not changed: containers carry no controller-injected canonical labels (`KukeonContainerLabelKey` is defined in `internal/consts/consts.go:65` but unused) and `update_container.go` has no `Metadata.Labels = …` clobber site. Symmetry with the issue's "five resources" framing is honored at the helper's API; nothing to wire on the container path until/unless containers start getting canonical labels too.

## Test plan

- [x] `make test` (full unit suite minus e2e) — green, including the new `TestMergeManagedLabels` 7-case table covering AC #1 / #2 / #3 plus suffix-not-substring matching and nil-map edge cases.
- [x] `go build ./...` clean.
- [x] `go vet ./...` clean.
- [x] `pre-commit run --files <changed>` clean (go-fmt, golangci-lint, addlicense, etc.).
- [x] `make dev-init` — full re-bootstrap loop is green and the daemon-parity check (`kuke get realms` vs `kuke get realms --no-daemon`) prints the expected two-row identical block. Change touches code the daemon reads, per `CLAUDE.md` this guard is required in the test plan.
- [x] Manual repro from the issue body — `kuke run -f docs/examples/hello-world.yaml -d`, then `kuke apply -f` against a copy that adds `labels: { env: prod }`. Post-apply `kuke get cell -o yaml` shows all four canonical `*.kukeon.io` keys preserved plus the user-added `env: prod` — pre-fix, the four canonical keys were gone.
- [ ] `make e2e` — not run; the full e2e suite requires a clean dev host and was not part of the local validation envelope. The smoke above exercises the same daemon-readable code path end-to-end.

Closes #455

### DIFF
--- a/internal/controller/runner/labels.go
+++ b/internal/controller/runner/labels.go
@@ -1,0 +1,55 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package runner
+
+import "strings"
+
+// managedLabelSuffix is the trailing key segment used by every
+// controller-injected canonical label (`realm.kukeon.io`, `space.kukeon.io`,
+// `stack.kukeon.io`, `cell.kukeon.io`). Kept in sync with the same suffix
+// hardcoded in `internal/controller/apply/diff.go::filterManagedLabels`.
+const managedLabelSuffix = ".kukeon.io"
+
+// mergeManagedLabels returns the label map an Update* runner should write back
+// onto the persisted resource: every key from `desired` wins, plus any
+// controller-managed `*.kukeon.io` key already in `existing` that `desired`
+// did not author.
+//
+// The create-time injection in `internal/controller/create_{realm,space,stack,cell}.go`
+// adds the canonical labels with "if not exists" semantics, so an explicit
+// user-authored `*.kukeon.io` value in `desired` still wins (AC #3 of issue
+// #455). User-authored non-managed labels are unchanged: present in `desired`
+// → kept; absent in `desired` → dropped.
+func mergeManagedLabels(existing, desired map[string]string) map[string]string {
+	if len(existing) == 0 && len(desired) == 0 {
+		return desired
+	}
+	out := make(map[string]string, len(desired)+len(existing))
+	for k, v := range desired {
+		out[k] = v
+	}
+	for k, v := range existing {
+		if !strings.HasSuffix(k, managedLabelSuffix) {
+			continue
+		}
+		if _, authored := out[k]; authored {
+			continue
+		}
+		out[k] = v
+	}
+	return out
+}
--- a/internal/controller/runner/labels_test.go
+++ b/internal/controller/runner/labels_test.go
@@ -1,0 +1,132 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package runner
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestMergeManagedLabels(t *testing.T) {
+	tests := []struct {
+		name     string
+		existing map[string]string
+		desired  map[string]string
+		want     map[string]string
+	}{
+		{
+			// AC #1: canonical *.kukeon.io labels survive an Update* call
+			// against a desired doc that authors one user label.
+			name: "preserves canonical labels when desired omits them",
+			existing: map[string]string{
+				"realm.kukeon.io": "default",
+				"space.kukeon.io": "default",
+				"stack.kukeon.io": "default",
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #456 Review — `fix: preserve canonical *.kukeon.io labels through UpdateCell etc.`**

Clean root-cause fix at the right layer — the `Update*` runners were the destructive primitive PR #454 only papered over at the diff layer. Helper is small and well-tested, scoping decisions (container path skipped because `KukeonContainerLabelKey` is unused) are correct, and the smoke + manual repro in the test plan cover the regression path. No correctness, security, or dead-code concerns.

A couple of small non-blocking observations:

- **Duplicated constant** `internal/controller/runner/labels.go:23`: `managedLabelSuffix = ".kukeon.io"` is the same literal hardcoded in `internal/controller/apply/diff.go:11` (per `filterManagedLabels`). Comment on `labels.go:21` names the linkage, so this is fine for now — worth folding into a shared package if a third site appears.
- **Reference sharing in the zero-len branch** `internal/controller/runner/labels.go:33`: when both maps are empty, the function returns `desired` directly. Caller assigns it back to `existing.Metadata.Labels`, so the two fields end up aliasing the same backing map. Not a bug today (neither is mutated post-assignment), but worth knowing if either side later grows a mutation.

LGTM.

PR #461, opened 2026-05-13
feat: fail-fast root check across kuke init / daemon reset / image load --no-daemon / doctor cgroups --probe

### LINKED ISSUE #403: cli: fail-fast root check across kuke init / daemon reset / image load --no-daemon / doctor cgroups --probe
## Context

`kuke init` and a handful of sibling subcommands assume root. Today there is no explicit UID check — operations just fail at the kernel boundary several phases in, producing errors that read like system bugs rather than user error.

Concrete reproduction:

- `kuke init` invoked as a non-root user reaches `sysuser.EnsureUserGroup` (`cmd/kuke/init/init.go:394` calling `internal/sysuser/sysuser.go:112,138`) and surfaces `groupadd … operation not permitted` from `useradd --system`.
- `kuke daemon reset` touches `/run/kukeon/kukeond.{sock,pid}` (root:kukeon), failing later with a `chmod`/`unlink` permission error.
- `kuke image load --from-docker --no-daemon` writes directly into `/opt/kukeon/<realm>/…`, which is root-owned after `kuke init`.
- `kuke doctor cgroups --probe` writes `/sys/fs/cgroup/cgroup.subtree_control`, which is root-only.

`cmd/kuke/init/init.go:273` already does the same shape of pre-check for the containerd socket via `preflightContainerdSocket` — fail fast at the top of `runInit` with a clear, named cause. A symmetric UID check matches that convention.

## Proposal

1. Add `ErrMustRunAsRoot` to `internal/errdefs/errdefs.go` (follow the existing `ErrXyz = errors.New(\"command must run as root\")` style — e.g. lines 241, 248).
2. Add a shared helper (`cmd/kuke/shared/root.go` or similar) that returns `errdefs.ErrMustRunAsRoot` when `os.Geteuid() != 0`. Wrapped error message names the subcommand and points at \`sudo\`.
3. Call it at the top of:
   - `runInit` in `cmd/kuke/init/init.go:340` (before the existing containerd preflight).
   - `kuke daemon reset` entrypoint.
   - `kuke image load` entrypoint, but only when `--no-daemon` is set (the daemon path is fine without root for `kukeon`-group members).
   - `kuke doctor cgroups` entrypoint when `--probe` is set.

Subcommands that go *through* the running daemon (`kuke get`, `kuke create`, `kuke apply`, `kuke delete`, `kuke log`, `kuke attach`, …) must **not** get this check — those are the supported \`kukeon\`-group rootless-client path. The check only goes on direct-write commands.

## Acceptance criteria

- [ ] `ErrMustRunAsRoot` exists in \`internal/errdefs\` with the existing-style sentinel.
- [ ] Invoking `kuke init`, `kuke daemon reset`, `kuke image load --no-daemon`, and `kuke doctor cgroups --probe` as a non-root user fails before any side effect with an error wrapping `ErrMustRunAsRoot`. The error message names the command and suggests \`sudo\`.
- [ ] Invoking the daemon-routed `kuke get realms` and `kuke create …` as a non-root user that is in the \`kukeon\` group continues to work unchanged.
- [ ] Unit test in the shared helper package: `os.Geteuid()` mocked via a function variable; one test per (root, non-root) outcome.

## Notes

- The convention already exists for the containerd socket — see `preflightContainerdSocket` at `cmd/kuke/init/init.go:273`. This issue extends the same pattern to the UID dimension.
- Adjacent: #387 (\`install-dev\` Makefile recipes hardcoding sudo) tightens the same root-only assumption at the build-system layer.
- Out of scope: any actual rootless mode (slirp4netns, rootless containerd, XDG paths). #62 and #310 are wontfix on that path; this issue is purely about giving a clear error when the existing root-only assumption is violated.

eminwux:
## Summary
- New `errdefs.ErrMustRunAsRoot` sentinel + shared `cmd/kuke/shared.RequireRoot(subcommand)` helper that wraps it with a sudo-suggestion error message. Drives a fail-fast UID check at the top of `runInit`, `runReset`, the `kuke image load` RunE (gated on `KUKEON_NO_DAEMON=true`), and the `kuke doctor cgroups` RunE (gated on effective `--probe`). Matches the existing convention `preflightContainerdSocket` set in `cmd/kuke/init/init.go:273` for the containerd-socket pre-flight.
- Daemon-routed verbs (`kuke get`, `kuke create`, `kuke apply`, `kuke delete`, …) are intentionally untouched — those are the supported `kukeon`-group rootless-client path.
- New unit tests cover (root, non-root) for the helper, and the gated entrypoints each grow one explicit non-root rejection case. `kuke doctor cgroups` also gets a sibling test proving `--no-probe` continues to work as non-root. `TestMain` mocks euid=0 across the reset and cgroups packages so the existing fakes-driven coverage keeps passing on the non-root ubuntu-latest CI runner.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full repo test suite passes.
- [x] `make dev-init` — full smoke runs clean; the daemon-parity tail matches the regression-guard expected output from `CLAUDE.md` (both `kuke get realms` and `kuke get realms --no-daemon` produce the canonical two-realm table).
- [x] `pre-commit run --files ...` on the changed set passes (golangci-lint flagged a named-return on the test seam helper, fixed before commit).

Closes #403

### DIFF
--- a/cmd/kuke/daemon/reset/reset.go
+++ b/cmd/kuke/daemon/reset/reset.go
@@ -32,6 +32,7 @@ import (
 	"time"
 
 	"github.com/eminwux/kukeon/cmd/config"
+	kukshared "github.com/eminwux/kukeon/cmd/kuke/shared"
 	"github.com/eminwux/kukeon/cmd/types"
 	"github.com/eminwux/kukeon/internal/client/local"
 	"github.com/eminwux/kukeon/internal/consts"
@@ -102,6 +103,10 @@ func runReset(cmd *cobra.Command, _ []string) error {
 	logger, ok := cmd.Context().Value(types.CtxLogger).(*slog.Logger)
 	if !ok || logger == nil {
 		return errdefs.ErrLoggerNotFound
+	}
+
+	if err := kukshared.RequireRoot("kuke daemon reset"); err != nil {
+		return err
 	}
 
 	timeout := viper.GetDuration(config.KUKE_DAEMON_RESET_TIMEOUT.ViperKey)
--- a/cmd/kuke/daemon/reset/reset_test.go
+++ b/cmd/kuke/daemon/reset/reset_test.go
@@ -30,12 +30,26 @@ import (
 	"time"
 
 	reset "github.com/eminwux/kukeon/cmd/kuke/daemon/reset"
+	kukshared "github.com/eminwux/kukeon/cmd/kuke/shared"
 	"github.com/eminwux/kukeon/cmd/types"
 	"github.com/eminwux/kukeon/internal/consts"
+	"github.com/eminwux/kukeon/internal/errdefs"
 	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 	"github.com/spf13/viper"
 )
+
+// TestMain mocks the shared euid lookup to euid=0 for every test in this
+// package so the new fail-fast root gate in runReset does not short-circuit
+// the existing fakes-driven coverage when CI runs as a non-root user
+// (ubuntu-latest defaults to UID 1001). Individual cases that exercise the
+// non-root path override this with their own SetGeteuidForTesting call.
+func TestMain(m *testing.M) {
+	restore := kukshared.SetGeteuidForTesting(func() int { return 0 })
+	code := m.Run()
+	restore()
+	os.Exit(code)
+}
 
 // TestDaemonReset covers the table of states the verb has to handle: a
 // running daemon (graceful stop+delete), an already-stopped daemon (skip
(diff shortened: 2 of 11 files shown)


eminwux:
**PR #461 Review — `feat: fail-fast root check across kuke init / daemon reset / image load --no-daemon / doctor cgroups --probe`**

Clean execution: new `errdefs.ErrMustRunAsRoot` sentinel + `kukshared.RequireRoot(subcommand)` helper wired into exactly the four direct-write entrypoints called out in the title, with the gate placed before any side effect in each one (`runInit:347` fires before `preflightContainerdSocket:384`; `runReset` fires before the timeout/cell-lookup path; image load gates only when `KUKEON_ROOT_NO_DAEMON` is set; doctor cgroups gates only when `effectiveProbe`). Sentinel correctly lives in `internal/errdefs/` per the repo convention, both new files carry the Apache header, error message names the subcommand and suggests `sudo`. Tests cover (root, non-root) for the helper plus an explicit non-root rejection case at each gated entrypoint, and the `TestMain` euid=0 mocks in the `reset` and `cgroups` packages keep the existing fakes-driven coverage green on the non-root CI runner. Daemon-routed verbs (`get/create/apply/delete`) are correctly untouched — verified, no surprise `RequireRoot` call sites outside the four expected. No credential/PII leak surface. Commit is signed. Test plan is fully checked, including the `make dev-init` daemon-parity regression guard per CLAUDE.md.

Non-blocking observation:

- **Parity** `cmd/kuke/daemon/{start,stop,kill,restart,logs}/*.go`: these verbs use the same `local.New()` in-process / direct-containerd path as `daemon reset` (e.g. `cmd/kuke/daemon/stop/stop.go:194`, `cmd/kuke/daemon/kill/kill.go:146`, `cmd/kuke/daemon/start/start.go:146`, `cmd/kuke/daemon/restart/restart.go:220`, `cmd/kuke/daemon/logs/logs.go:207`), so a non-root operator gets the same confusing containerd EACCES the PR was written to eliminate. The title is explicit about scope, so I'm not blocking — flagging for a follow-up nit so the convention reaches all the in-process daemon-lifecycle verbs.

LGTM.

PR #480, opened 2026-05-13
fix: kuke run -f refuses on-disk spec divergent from file

### LINKED ISSUE #468: kuke run -f does not refuse divergent on-disk spec
## Context

`docs/cli-use-cases.md` states two invariants about `kuke run -f` against an existing cell whose on-disk spec diverges from the file:

- `## Workload lifecycle` / `### Run a one-off cell from YAML`:

  > Re-running `kuke run -f` against an existing cell whose on-disk spec **diverges** from the file is **refused**, not silently updated. The error message points the operator at `kuke apply -f` for the update path. Exit code non-zero.

- `## Error & edge paths` / `### Conflicting kuke run`:

  > `kuke run -f spec.yaml` against an existing cell with a diverging on-disk spec exits non-zero with `cell "<name>" exists with diverging spec (...); use \`kuke apply -f\` to update`. The CLI does **not** mutate the cell. The cell that drove the divergence detection is named explicitly so automation can branch.

```
$ cp docs/examples/hello-world.yaml /tmp/cli-test/hello-world-divergent.yaml
$ sed -i 's|image: docker.io/library/busybox:latest|image: docker.io/library/busybox:musl|' \
    /tmp/cli-test/hello-world-divergent.yaml

$ sudo kuke run -f docs/examples/hello-world.yaml -d
Cell "hello-world" (realm "default", space "default", stack "default")
  - metadata: created
  ...
  - containers: started
# exit 0

$ sudo kuke run -f /tmp/cli-test/hello-world-divergent.yaml -d
Cell "hello-world" (realm "default", space "default", stack "default")
  - metadata: already existed
  - cgroup: already existed
  - root container: already existed
  - container "web": already existed
  - containers: not started
# exit 0  ← should be non-zero with diverging-spec error
```

The CLI prints "already existed" lines and exits **0** instead of emitting `cell "hello-world" exists with diverging spec (...); use \`kuke apply -f\` to update` and exiting non-zero.

The on-disk metadata's `spec.containers[].image` remains `docker.io/library/busybox:latest`, confirming the cell was not mutated — so the "does not mutate the cell" half of the invariant holds. But the **refusal** half (the non-zero exit + diverging-spec error) does not.

This is the invariant that lets automation branch on `kuke run` vs `kuke apply`. Today both verbs are indistinguishable on the divergence path: `apply` errors, `run` silently succeeds.

## Proposal

Before the `kuke run -f` code path runs the create-or-no-op for each phase, diff the parsed spec against the on-disk `metadata.json`. On any difference in `spec` (containers, image, command, args, env, resources, networks, attachable flag, etc. — same fields `apply` considers):

1. Exit non-zero (suggest `1`).
2. Print the exact error the doc names: `Error: cell "<name>" exists with diverging spec (<short field summary, e.g. "containers[0].image: busybox:latest -> busybox:musl">); use \`kuke apply -f\` to update`.
3. Do not mutate any phase (metadata, cgroup, containers).

Whatever divergence detector `kuke apply -f` already runs is the right place to factor out into a shared check — `run` should refuse on the same set of divergences that `apply` accepts.

## Acceptance criteria

- [ ] `kuke run -f <file>` against a cell whose on-disk spec differs in any spec field exits non-zero (≠ 0).
- [ ] The error message contains the literal substring `exists with diverging spec` and names the cell.
- [ ] The error message points the operator at `kuke apply -f` for the update path.
- [ ] The on-disk metadata is unchanged after the refused run (no partial mutation).
- [ ] `kuke run -f <file>` against a cell whose on-disk spec is byte-equal to the file still succeeds with exit 0 (idempotent "already existed" path remains unbroken).
- [ ] Test covering the divergent-image case lives next to the run command implementation.

## Out of scope

- Defining which spec fields count as "divergent" beyond what `kuke apply -f` already detects — share the same diff logic.
- Changing `kuke apply -f` behavior (separate issue tracks the apply-side divergence-acceptance bug).

## References

- `docs/cli-use-cases.md` — `## Workload lifecycle` / `### Run a one-off cell from YAML` (line ~238).
- `docs/cli-use-cases.md` — `## Error & edge paths` / `### Conflicting kuke run` (line ~396).

eminwux:
## Summary

- `kuke run -f` previously only diffed structural identity (container count, container id set, root flag, cell location) against the on-disk metadata, so a divergent **container image** — the headline case in #468 — slipped through and the CLI silently no-opped with exit 0. The `docs/cli-use-cases.md` invariants require refusal with a non-zero exit and a pointer to `kuke apply -f`.
- Extends `divergedFields` to also compare per-container user-authored fields the runner does NOT rewrite at create time: `image`, `command`, `args`, `workingDir`, `env`, `ports`, `volumes`, `networks`, `networksAliases`, `privileged`, `hostNetwork`, `hostPID`, `hostCgroup`, `attachable`, `restartPolicy`, `secrets`, `tty`.
- Deliberately excludes fields inherited from `space.spec.defaults.container` (user, readOnlyRootFilesystem, capabilities, securityOpts, tmpfs, resources). Their on-disk value is post-merge while the YAML side is pre-merge, so comparing them would false-flag a default-using YAML on every re-run.
- Updates fixtures in the existing idempotent / matching-spec / synthesized-root tests so they reflect what an on-disk cell created from the YAML actually looks like (command/args persisted, not stripped).
- Adds `TestRun_ExistingCell_DivergingContainerImage_RefusesAndPointsToApply` covering the bug from the issue body verbatim (busybox:latest on disk vs busybox:musl in file), plus matching regression tests for `secrets` and `tty` divergence.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./cmd/kuke/run/...` — new divergent-image/secrets/tty tests plus updated fixtures all pass
- [x] `make test` — full unit suite green
- [x] `pre-commit run --files cmd/kuke/run/run.go cmd/kuke/run/run_test.go` — all hooks pass (golangci-lint, gofmt, addlicense, etc.)
- [ ] \`make dev-init\` — not run; change is purely in \`cmd/kuke/run/\` and does not touch the build path, the daemon, or \`/opt/kukeon\`, so the daemon-parity smoke test is out of scope per \`CLAUDE.md\`.

Closes #468

### DIFF
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -587,7 +587,8 @@ func divergedFields(actual, desired v1beta1.CellSpec) []string {
 // fill in or normalize at create/start time (so a fresh cell never
 // false-flags on the next `kuke run -f` of the same file): image, command,
 // args, workingDir, env, ports, volumes, networks, networksAliases,
-// privileged, hostNetwork, hostPID, hostCgroup, attachable, restartPolicy.
+// privileged, hostNetwork, hostPID, hostCgroup, attachable, restartPolicy,
+// secrets, tty.
 // Fields that inherit from `space.spec.defaults.container` (see
 // internal/modelhub.ApplySpaceDefaultsToContainer) are deliberately
 // excluded — their on-disk value is post-merge while the YAML side is
@@ -639,7 +640,59 @@ func divergedContainerFields(actual, desired v1beta1.ContainerSpec) []string {
 	if actual.RestartPolicy != desired.RestartPolicy {
 		fields = append(fields, "restartPolicy")
 	}
+	if !containerSecretsEqual(actual.Secrets, desired.Secrets) {
+		fields = append(fields, "secrets")
+	}
+	if !containerTtysEqual(actual.Tty, desired.Tty) {
+		fields = append(fields, "tty")
+	}
 	return fields
+}
+
+// containerSecretsEqual compares two ContainerSecret slices field-by-field.
+// nil and empty are treated as equal so YAML that omits secrets does not
+// register as drift against on-disk metadata that persisted it as nil.
+// ContainerSecret carries only scalar string fields, so a direct == on each
+// element is enough.
+func containerSecretsEqual(a, b []v1beta1.ContainerSecret) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+	return true
+}
+
+// containerTtysEqual compares two ContainerTty pointers. Empty values on
+// either side are treated as equal because
+// internal/apischeme.convertContainerTtyToInternal normalizes an IsEmpty
+// input to nil at persistence: on-disk Tty=nil and a YAML carrying an
+// otherwise-empty &ContainerTty{} must not register as drift.
+func containerTtysEqual(a, b *v1beta1.ContainerTty) bool {
+	if a.IsEmpty() && b.IsEmpty() {
+		return true
+	}
+	if a.IsEmpty() != b.IsEmpty() {
+		return false
+	}
+	if a.Prompt != b.Prompt ||
+		a.LogFile != b.LogFile ||
+		a.Profile != b.Profile ||
+		a.ProfilesDir != b.ProfilesDir {
+		return false
+	}
+	if len(a.OnInit) != len(b.OnInit) {
+		return false
+	}
+	for i := range a.OnInit {
+		if a.OnInit[i] != b.OnInit[i] {
+			return false
+		}
+	}
+	return true
 }
 
 // stringSlicesEqual reports whether two []string carry the same elements
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -508,15 +508,14 @@ func pickLocation(fromDoc string, kv *config.Var) string {
 	return strings.TrimSpace(kv.ValueOrDefault())
 }
 
-// divergedFields names the structural fields where the on-disk cell disagrees
-// with the file. We deliberately stop at structural identity (container count,
-// container id set, root container id, cell location) and do NOT compare
-// images/args/env: the runner normalizes container images on create
-// (auto-default root rewrites e.g. user-supplied image to docker.io/library/
-// busybox:latest), so a deep field-by-field diff would flag a freshly-created
-// cell as divergent on the very next `kuke run -f` invocation. Per the AC, a
-// genuine spec rewrite — adding/removing containers, swapping the root, moving
-// realm/space/stack — must route through `kuke apply -f` instead.
+// divergedFields names the fields where the on-disk cell disagrees with the
+// file. The check covers structural identity (container count, container id
+// set, cell location) and the operator-controlled per-container fields the
+// runner does NOT rewrite at create time (image, command, args, env, ports,
+// volumes, networks, host-namespace opts, privileged, attachable,
+// restartPolicy, workingDir). Per the AC for issue #468, a genuine spec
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #480 Review — `fix: kuke run -f refuses on-disk spec divergent from file`**

Solid scope on the #468 AC — the new field-by-field check pins the headline image-divergence case, the defaults-derived carve-out (`user`/`readOnlyRootFilesystem`/`capabilities`/`securityOpts`/`tmpfs`/`resources`) is well-reasoned and lines up exactly with `internal/modelhub.ApplySpaceDefaultsToContainer`, and the new test exercises the bug-report scenario verbatim.

One blocking gap:

- **Missing fields** `cmd/kuke/run/run.go:595-643`: `divergedContainerFields` omits `Secrets` and `Tty`. Both are user-authored, persisted to disk unchanged (verified in `internal/apischeme/scheme.go` round-trips), and not filled in from `space.spec.defaults.container` — i.e. they meet the exact criterion this PR's doc comment uses to define the included set. A user who adds a new `spec.containers[].secrets[]` entry and re-runs `kuke run -f` against an existing cell silently no-ops (cell stays up without the credential mounted), which is the same failure mode #468 closes for `image`. Either thread both through here, or carve them out explicitly in the doc comment with a stated reason (parallel to the defaults-derived carve-out already documented).

LGTM once `Secrets` and `Tty` are added or carved out — the rest is non-blocking.

eminwux:
Addressed in b74ee85:
- Threaded `Secrets` and `Tty` through `divergedContainerFields` per the blocking ask. Both are user-authored, persisted unchanged through `internal/apischeme` (`convertSecretsToInternal` / `convertContainerTtyToInternal` and their inverses round-trip the fields verbatim), and absent from `internal/modelhub.ApplySpaceDefaultsToContainer` — same inclusion criterion the doc comment already states. Added `containerSecretsEqual` (scalar-only `ContainerSecret`, mirrors `volumeMountsEqual`) and `containerTtysEqual` (treats nil and `IsEmpty` as equal so on-disk `Tty=nil` from the apischeme normalization doesn't false-flag against a YAML-side zero-value `&ContainerTty{}`).
- Doc comment in `divergedContainerFields` now lists `secrets` and `tty` alongside the existing fields.
- New regression tests: `TestRun_ExistingCell_DivergingContainerSecrets_RefusesAndPointsToApply` and `TestRun_ExistingCell_DivergingContainerTty_RefusesAndPointsToApply` — both refuse with the diverging-spec error, point at `kuke apply -f`, and name the specific field.

eminwux:
**PR #480 Review — `fix: kuke run -f refuses on-disk spec divergent from file`**

Closes the #468 silent-no-op cleanly. The per-field comparison list in `divergedContainerFields` was cross-checked against `internal/apischeme.convertContainerSpecToInternal` and `internal/modelhub.ApplySpaceDefaultsToContainer` — the 17 included fields all round-trip verbatim (none are runner-rewritten on create), and the 6 deliberately-excluded space-default fields match `ApplySpaceDefaultsToContainer`'s exact set, so the idempotent path is preserved. The `containerTtysEqual` helper is nil-safe through `(*ContainerTty).IsEmpty()` and correctly equates a `&ContainerTty{}` with `nil` (matches `convertContainerTtyToInternal`'s persistence normalization). All three new struct comparators rely on `==` over types whose fields are scalars only (`ContainerSecret`, `VolumeMount`, `TtyStage`) — comparability holds. New tests cover the headline image case from #468 plus regressions for `secrets` and `tty`. Both commits are GPG-verified. The `make dev-init` smoke is correctly skipped: the change is contained to `cmd/kuke/run/` and touches neither the build path, the daemon, nor `/opt/kukeon`, so the daemon-parity guard in project CLAUDE.md is out of scope.

LGTM.

PR #482, opened 2026-05-14
fix: serialize state enums as string labels in yaml/json

### LINKED ISSUE #474: kuke get -o yaml|json serializes state as int instead of string label
## Context

`docs/cli-use-cases.md` `## Realm / space / stack management` / `### Listing the hierarchy` documents `kuke get realm <name> -o yaml` as the "full spec+status" view of a single resource and lists table/yaml/json as the supported output formats.

```
$ sudo kuke get realms

$ sudo kuke get realm default -o yaml
apiVersion: v1beta1
kind: Realm
metadata:
  name: default
  labels:
    realm.kukeon.io: default.kukeon.io
spec:
  namespace: default.kukeon.io
status:
  state: 2                              # ← integer; table shows "Ready"
  cgroupPath: /kukeon/default
  subtreeControllers:
    - cpu
    - memory
    - io
    - pids
  createdAt: 2026-05-13T17:35:03.183753568Z
  updatedAt: 2026-05-13T17:36:53.977095133Z
  readyAt: 2026-05-13T17:35:03.183753568Z
```

The table renders `state` as `Ready`. The YAML serializes the same field as the raw enum integer `2`. JSON output (`-o json`) shows the same int. An operator reading the documented "full spec+status" view has no way to decode `state: 2` without grepping the source for the enum.

The doc's `### Listing the hierarchy` invariants don't explicitly constrain the YAML field representation, but `kuke get` is the documented introspection surface — an internal enum int leaking into a user-facing format is friction that any operator scripting against the YAML will hit.

## Proposal

In the realm (and matching space/stack/cell/container) JSON/YAML serializers, map the `state` enum to its string form (`Pending` / `Ready` / `Stopped` / etc. — the same vocabulary the table uses):

1. Define a `MarshalJSON`/`MarshalYAML` on the realm state enum (or whatever serialization wrapper is in use) that emits the string label.
2. Symmetric `UnmarshalJSON`/`UnmarshalYAML` accepts the string form so users who round-trip the YAML through `kuke apply -f` are not broken.
3. As a backwards-compat aid, also accept the integer form on unmarshal for one release if anything currently depends on the int.
4. Apply the same change to the same field on Space/Stack/Cell/Container if they share the enum or have an analogous one.

## Acceptance criteria

- [ ] `kuke get realm <name> -o yaml` emits `state: Ready` (string) rather than `state: 2` (int).
- [ ] `kuke get realm <name> -o json` emits `"state": "Ready"` rather than `"state": 2`.
- [ ] The same change applied consistently to `kuke get space|stack|cell|container -o {yaml,json}`.
- [ ] Round-trip: a YAML output piped back through `kuke apply -f` succeeds (no schema error on the string state).
- [ ] Test colocated with the marshal/unmarshal change covering both directions.

## Out of scope

- The table renderer (already emits the string form correctly).
- Adding new state values.

## References

- `docs/cli-use-cases.md` — `## Realm / space / stack management` / `### Listing the hierarchy` (line ~145).

eminwux:
## Summary

- `kuke get realm|space|stack|cell|container -o yaml|json` now emits `state: Ready` (string) instead of `state: 2` (raw enum int). Operators scripting against the output no longer have to grep the source for the enum.
- All five `*State` types (`RealmState`, `SpaceState`, `StackState`, `CellState`, `ContainerState`) gain `MarshalJSON` / `MarshalYAML` that serialize via the existing `String()` label vocabulary (`Pending`, `Ready`, `Stopped`, …) — the same strings the table renderer uses.
- Symmetric `UnmarshalJSON` / `UnmarshalYAML` accept both the new string form and the legacy int form, so daemon-persisted `metadata.json` files written by pre-fix daemons still parse after upgrade (the daemon's `internal/controller/runner/read.go` path goes through these unmarshals).
- Fixed a missing `yaml:"state"` tag on `RealmStatus.State` surfaced by `musttag` once the new serializer covered the field — the only Status struct in the package without the explicit tag.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `pre-commit run --files <changed-files>` — clean (golangci-lint, addlicense, go fmt all pass)
- [x] New unit tests cover: per-label `Marshal{JSON,YAML}` output, `Unmarshal{JSON,YAML}` accepting both string and legacy int forms, rejecting unknown labels and out-of-range ints, full `RealmDoc` yaml/json round-trip, and a legacy `ContainerStatus{state:2}` JSON parse that round-trips to `ContainerStateStopped`
- [x] `make dev-init` smoke — daemon-parity check passes (both `kuke get realms` and `--no-daemon` show identical `Ready` rows); change touches what the daemon persists/reads, so the smoke is in scope
- [x] Manual verification on the running daemon: `sudo ./kuke get realm default -o yaml` emits `state: Ready`, `-o json` emits `"state": "Ready"`, and `kuke get realm default -o yaml | kuke apply -f` succeeds with `Realm "default": unchanged` (no schema error on the string state — AC's round-trip case)

Closes #474

### DIFF
--- a/pkg/api/model/v1beta1/realm.go
+++ b/pkg/api/model/v1beta1/realm.go
@@ -48,7 +48,7 @@ type RegistryCredentials struct {
 }
 
 type RealmStatus struct {
-	State      RealmState `json:"state"`
+	State      RealmState `json:"state"                              yaml:"state"`
 	CgroupPath string     `json:"cgroupPath,omitempty"         yaml:"cgroupPath,omitempty"`
 	// SubtreeControllers is the cgroup-v2 controller set actually
 	// delegated on this realm's own cgroup.subtree_control after the
--- a/pkg/api/model/v1beta1/state_marshal.go
+++ b/pkg/api/model/v1beta1/state_marshal.go
@@ -1,0 +1,355 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package v1beta1
+
+// JSON/YAML marshaling for the *State enum types. The serializers emit the
+// human-readable label (`Ready`, `Pending`, …) instead of the raw int form
+// that the encoding/json and gopkg.in/yaml.v3 defaults would produce.
+// Unmarshal accepts both forms — string is canonical going forward, int is
+// retained for one release so older on-disk metadata.json files persisted by
+// pre-fix daemons still round-trip through `kuke get -o yaml | kuke apply`.
+// Issue #474.
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"gopkg.in/yaml.v3"
+)
+
+// yamlIntTag is the auto-resolved gopkg.in/yaml.v3 tag for an unquoted
+// integer scalar — the legacy on-disk encoding for state fields the
+// UnmarshalYAML methods accept for back-compat.
+const yamlIntTag = "!!int"
+
+// --- RealmState ---
+
+func (s RealmState) MarshalJSON() ([]byte, error) {
+	return json.Marshal((&s).String())
+}
+
+func (s RealmState) MarshalYAML() (interface{}, error) {
+	return (&s).String(), nil
+}
+
+func (s *RealmState) UnmarshalJSON(data []byte) error {
+	var str string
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #482 Review — `fix: serialize state enums as string labels in yaml/json`**

Clean fix for the `state: 2` UX bug on `kuke get -o yaml|json`. Symmetric Marshal/Unmarshal across all five state enums, with both string (canonical) and int (legacy) forms accepted on the way in so pre-fix on-disk metadata.json files still parse on upgrade. Tests cover the per-label JSON output, the back-compat int-form acceptance, the unknown-label/out-of-range rejection paths, and the full `RealmDoc` round-trip — including the `state=2` legacy parse to `ContainerStateStopped`. No sensitive data in error messages. Per-commit signature: verified. License headers and test colocation per CLAUDE.md ✓. The `make dev-init` smoke is ticked in the test plan, which the project CLAUDE.md requires for changes touching what the daemon persists/reads.

Cross-checks against the PR head: every label constant referenced in the `parse*State` switches exists in `consts.go:54-62`; every `assign*Int` upper/lower bound matches the contiguous iota ranges in `realm.go:89-94` / `space.go:122-125` / `stack.go:60-63` / `cell.go:110-115` / `container.go:262-269`; the `yaml:"state"` tag fix on `RealmStatus.State` does indeed close the only gap among the five `*Status` types. No public-API parity gap — these status types are read-only response fields, not constructor inputs.

LGTM to merge.

PR #483, opened 2026-05-14
fix: hint at cross-scope match when --cell/--space/--stack filter is empty

### LINKED ISSUE #472: kuke get containers --cell <name> does not "filter on any level"
## Context

`docs/cli-use-cases.md` `## Realm / space / stack management` / `### Listing the hierarchy` lists this command in the **Sequence** block:

```bash
kuke get containers --cell <name>                # filter on any level
```

The comment "filter on any level" reads as: passing `--cell <name>` is enough to filter containers down to that cell, regardless of which realm/space/stack the cell lives in.

```
$ sudo kuke get containers --cell kukeond
No containers found.
# exit 0

$ sudo kuke get containers
NAME     REALM        SPACE   STACK   CELL     ROOT   IMAGE                               STATE
-------  -----------  ------  ------  -------  -----  ----------------------------------  -----
kukeond  kuke-system  kukeon  kukeon  kukeond  false  docker.io/library/kukeon-local:dev  Ready
root     kuke-system  kukeon  kukeon  kukeond  true   docker.io/library/busybox:latest    Ready
# exit 0 — containers exist

$ sudo kuke get containers --cell kukeond --realm kuke-system --space kukeon --stack kukeon
# returns the two containers above
```

Without `--realm/--space/--stack`, the CLI appears to default the scope to `default/default/default` and then filter by `--cell kukeond` within that scope — finding nothing, because the `kukeond` cell lives in `kuke-system/kukeon/kukeon`. The result is the empty "No containers found." output, with no hint that the cell does exist in a different scope.

This contradicts the doc's "filter on any level" comment. It is also confusing UX:

- Bare `kuke get containers` (no filter) returns containers from every scope.
- Adding `--cell <name>` filters them *out* unless the operator also passes matching `--realm/--space/--stack` flags.

`kuke get cell <name>` (singular) has the same shape but at least produces an informative error: `Error: cell "<name>" not found in stack "default"/"default"/"default"`. The plural-list `kuke get containers --cell <name>` produces only a silent "No containers found." with exit 0.

## Proposal

Pick one — both are defensible, but the current behavior is the worst of both:

**Option A — make `--cell` actually filter on any level.** Treat `--cell kukeond` (with no `--realm/--space/--stack`) as "find any container whose `cellId == kukeond` across all realms/spaces/stacks". Match the doc's "filter on any level" comment. Same applies to `--stack` and `--space` when used alone.

**Option B — keep scope-bound filtering, but tell the operator why no rows matched.** If `--cell <name>` is passed and the default-scope lookup returns zero rows, but a cross-scope lookup *would* find at least one container with that cell name, surface a hint:

```
No containers found for cell "kukeond" in realm "default" space "default" stack "default".
Hint: cell "kukeond" exists in realm "kuke-system" space "kukeon" stack "kukeon" — pass --realm/--space/--stack to filter there.
```

Then update the doc's `kuke get containers --cell <name>                # filter on any level` line to `# filter within current scope` so the comment matches the behavior.

Option A is closer to the doc as written; Option B is the lower-blast-radius change. Either is acceptable — picking one and aligning the doc is the goal.

## Acceptance criteria

- [ ] Either Option A is implemented (cross-scope filter on `--cell` alone), **or** Option B is implemented (helpful hint + doc-comment update).
- [ ] `kuke get containers --cell kukeond` either returns the kukeond cell's containers (Option A) or returns the empty result with a hint naming the realm/space/stack where the cell does exist (Option B).
- [ ] `docs/cli-use-cases.md` line ~136 matches the chosen behavior.
- [ ] `--space <name>` and `--stack <name>` follow the same convention as `--cell` (don't fix one filter and leave the others diverged).

## Out of scope

- Changing `kuke get cell <name>` (singular) — it already errors with scope context, which is informative enough.
- Output format / table column changes.

## References

- `docs/cli-use-cases.md` — `## Realm / space / stack management` / `### Listing the hierarchy` (line ~136).

eminwux:
## Summary

- Issue #472 listed two acceptable paths; per the user's `Implement option B. Update documentation. Override pm role here.` comment, this PR ships **Option B** (informative hint when scope-bound filtering finds zero rows) instead of Option A. An earlier push on this branch implemented Option A — that commit has been replaced via `--force-with-lease`.
- When `kuke get containers --cell <name>` (or `--space` / `--stack`) returns zero rows in `table` output, the CLI now:
  1. Replaces the bare `No containers found.` line with a scope-aware message (e.g. `No containers found for cell "kukeond".`) so the operator can see which filter actually fired.
  2. Issues a single cross-scope `ListContainers("", "", "", "")` lookup, and if the supplied cell/space/stack name exists in a different realm/space/stack, appends `Hint: cell "kukeond" exists in realm "kuke-system" space "kukeon" stack "kukeon" — pass --realm/--space/--stack to filter there.`
- The hint is a table-only UX aid: `-o yaml` / `-o json` still emit the empty list with exit code 0 so machine-readable consumers are unaffected.
- `--realm` alone does not trigger a hint — realm is the top-level scope and "look in another realm" is not actionable advice.
- The cross-scope lookup fails open: any error from the second `ListContainers` call is treated as "no hint to add" rather than escalated, so a transient daemon hiccup never turns an empty-result table into an error.
- `docs/cli-use-cases.md:136` updated from `# filter on any level` to `# filter within current scope` to match the chosen behavior, plus a new invariant under "Listing the hierarchy" that documents the `Hint:` line so reviewers and `/test-cli` runs treat it as expected output.

## Test plan

- [x] `go test ./cmd/kuke/get/container/...` — `TestNewContainerCmd` extended with six new subcases (cell/stack/space hint paths, no-match-anywhere path, realm-only-no-hint path, yaml-suppresses-hint path); all pass.
- [x] `go test ./...` — full suite green.
- [x] `go vet ./...` — clean.
- [x] `pre-commit run --files <changed-files>` — all hooks pass (go fmt, golangci-lint, addlicense, prettier).
- [ ] `make dev-init` smoke — not re-run for this Option B swap; the change is CLI-layer only and does not touch the daemon, build, or `/opt/kukeon` write path. Manual table-output verification via the existing unit tests covers the user-visible behavior.

Closes #472

### DIFF
--- a/cmd/kuke/get/container/container.go
+++ b/cmd/kuke/get/container/container.go
@@ -17,8 +17,10 @@
 package container
 
 import (
+	"context"
 	"errors"
 	"fmt"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -34,6 +36,8 @@ import (
 
 // MockControllerKey is used to inject a mock kukeonv1.Client via context in tests.
 type MockControllerKey struct{}
+
+const noContainersFoundMsg = "No containers found."
 
 type (
 	printObjectFunc  func(interface{}) error
@@ -177,6 +181,14 @@ func runContainerCmdWithDeps(
 		return err
 	}
 
+	emptyMsg := noContainersFoundMsg
+	if len(specs) == 0 && outputFormat == shared.OutputFormatTable {
+		emptyMsg = buildEmptyResultMessage(realm, space, stack, cell)
+		if hint := maybeBuildScopeHint(cmd.Context(), client, space, stack, cell); hint != "" {
+			emptyMsg = emptyMsg + "\n" + hint
+		}
+	}
+
 	containerStates := make(map[string]string, len(specs))
 	for i := range specs {
 		spec := specs[i]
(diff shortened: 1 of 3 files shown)


eminwux:
Switched from Option A to Option B per the user override comment on #472. Force-pushed `687df92` to replace the prior `82b11f2`. PR title and body updated to match. The doc line (`docs/cli-use-cases.md:136`) is now `# filter within current scope` and a new invariant documents the `Hint:` line so the behavior is auditable.

eminwux:
**PR #483 Review — `fix: hint at cross-scope match when --cell/--space/--stack filter is empty`**

Clean Option-B implementation of #472. The hint logic is well-scoped (table-only, realm-only no-op, fails open on the cross-scope lookup), tests cover the six branches that matter, and no public API moves — `pkg/api/kukeonv1/client.go` is untouched and the helpers stay private to the package. Commit is signed and verified. No correctness, secret-leak, or convention issues found.

A couple of observations, all non-blocking:

- **Sibling commands** `cmd/kuke/get/cell/cell.go:108`, `space/space.go:90`, `stack/stack.go:100`: same "filter returns zero, named entity exists elsewhere" UX exists for `kuke get cells/spaces/stacks` and was deliberately out of scope per #472. Worth a future pass if the pattern lands well here — not for this PR.
- **`make dev-init` deferral**: the PR body's reasoning (CLI-only, no daemon or `/opt/kukeon` write-path touched) checks out — the diff lands entirely in `cmd/kuke/get/container/` plus a docs tweak, and the RPC surface is unchanged. The carve-out is justified.

LGTM.

PR #492, opened 2026-05-14
fix: switch e2e run paths to t.TempDir() to drop shared tmp/ race

### LINKED ISSUE #491: e2e: mkdirRunPath races parallel test cleanup on shared tmp/ parent
## Repro

On a clean checkout of `origin/main`, run the e2e suite in a fresh shell, repeatedly:

```bash
for i in 1 2 3 4 5; do go test -count=1 ./e2e/ 2>&1 | tail -3; echo ---; done
```

Failure rate ~40% (2/5 in one session, 3/5 in another). Each failure points at `mkdirRunPath` (`e2e/e2e_test.go:124`) returning ENOENT on `os.MkdirAll(tmp/e-<timestamp>)`. A different test fails each run (`TestKuke_NoRealms`, `TestKuke_PurgeStack_VerifyState`, `TestKuke_CreateStack_VerifyState`, etc.), consistent with a race rather than a deterministic break.

## Observed

```
--- FAIL: TestKuke_NoRealms (0.00s)
    e2e_kuke_realm_test.go:136: could not create dir /root/kukeon/e2e/tmp/e-1778759971696760726: mkdir /root/kukeon/e2e/tmp/e-1778759971696760726: no such file or directory
```

The error is "no such file or directory" *from `os.MkdirAll`*, which normally creates all missing parents — so a parent disappeared between the helper's stat-then-mkdir walk and the actual mkdir syscall.

## Root cause

Likely race in the test harness, not the kuke CLI:

- `getRandomRunPath` (`e2e/e2e_test.go:114`) returns a unique-per-call path of the form `tmp/e-<UnixNano>`, all sharing the *same* `tmp/` parent under the repo's `e2e/` directory.
- `mkdirRunPath` (`e2e/e2e_test.go:124`) calls `os.MkdirAll(fullDir, 0o755)` to create `e2e/tmp/e-<id>`, and registers `t.Cleanup(func() { os.RemoveAll(fullDir) })`.
- `go test` runs `t.Parallel()` tests interleaved with non-parallel ones in the same binary. When a parallel test's cleanup runs `os.RemoveAll(tmp/e-X)` concurrently with another test's `os.MkdirAll(tmp/e-Y)`, `MkdirAll`'s internal "find longest existing prefix" walk can observe the parent's `tmp/` directory entries in flux. On Linux this manifests as ENOENT for the *new* leaf path rather than the parent — the kernel returns ENOENT when the rename-vs-readdir ordering inside `unlinkat` races a sibling `mkdirat` for the same parent inode (the `unlinkat` removes an entry the new `mkdirat`'s dotwalk thought it had pinned).

Not yet pinpointed which exact pair of tests collide each run, but the race surfaces *only* in parallel-test interleaving (`go test -run <single>` always passes in isolation).

## Suggested fix

Drop the shared `e2e/tmp/` parent and use `t.TempDir()` for each test's runPath. The test framework wires per-test cleanup automatically, removes the shared-parent contention, and matches the rest of the test code in the repo (e.g. `internal/cni/subnet_test.go`, `cmd/kuke/daemon/reset/reset_test.go`). The CLI's `--run-path` already accepts arbitrary absolute paths, so no other code changes.

If the shared parent must stay (e.g. easier post-mortem cleanup), the alternative is to wrap `os.MkdirAll` in a retry-on-ENOENT loop, but `t.TempDir()` is the cleaner option and removes the race rather than masking it.

## References

- `e2e/e2e_test.go:114` — `getRandomRunPath` (shared parent).
- `e2e/e2e_test.go:124` — `mkdirRunPath` (the racing `os.MkdirAll`).
- Surfaced during local verification of PR #489 (`make dev-init` smoke + full unit test sweep).

eminwux:
## Summary

- The e2e harness handed every test a unique-suffix path under a shared `e2e/tmp/` parent and created it with `os.MkdirAll`. Under `t.Parallel()`, a sibling test's `t.Cleanup` `os.RemoveAll(tmp/e-X)` could race a peer's `os.MkdirAll(tmp/e-Y)` for the same parent inode, surfacing as `ENOENT` on the *new* leaf — ~40% failure rate on local re-runs, deterministically green only with `-run <single>`. Per the issue's preferred option, replace the bespoke helper with `t.TempDir()`: each test gets its own absolute path under `os.TempDir()`, `testing` registers per-test cleanup, and the shared-parent contention disappears.
- `getRandomRunPath` now returns `t.TempDir()` (absolute path, auto-cleaned). `mkdirRunPath` is removed — the directory already exists when the helper returns, so the call-site pair (`getRandomRunPath` + `mkdirRunPath`) collapses to one line. Six verifier helpers (`verifySpaceCNIConfigExists`, `verifySpaceMetadataExists`, `verifyStackMetadataExists`, `verifyCellMetadataExists`, `verifyRealmMetadataExists`, and the inline join in `TestKuke_DaemonReset_RoundTrip`) previously prepended `cwd` to a relative `runPath`; with `runPath` now absolute, the `filepath.Join(cwd, runPath)` becomes wrong (`filepath.Join("/work", "/tmp/x")` → `/work/tmp/x`), so those wraps are dropped and the helpers pass `runPath` straight to the `fs` builders. `appendNoDaemonRunPath` / `absRunPath` (attach test) already handle absolute paths via `filepath.IsAbs`, so no change there.

## Test plan

- [x] `go build ./...` — clean compile.
- [x] `go test $(go list ./... | grep -v /e2e$)` — full non-e2e sweep green (60+ packages including `internal/cni`, `internal/controller/runner`, `internal/ctr`, `cmd/...`). No regressions.
- [x] `for i in 1 2 3 4 5; do go test -count=1 -timeout 300s ./e2e/ 2>&1 | tail -8; done` — five consecutive full e2e runs on this dev host: zero `could not create dir … no such file or directory` failures (the exact race the issue reports). The bare-`grep -iE "could not create dir|mkdir.*no such file"` over the same 5 runs returns no matches. Remaining e2e failures on this host are environmental (no kukeond daemon, no kuketty binary staged) — verified to reproduce identically on `origin/main` with the change stashed (`TestKuke_NoRealms` fails on both with the same `dial unix /run/kukeon/kukeond.sock: connect: no such file or directory`), so they are pre-existing and unrelated.
- [x] `pre-commit run --files \$(git diff --name-only HEAD~1)` — trim-trailing-whitespace, fix-end-of-files, go-fmt, go-mod-tidy, golangci-lint, addlicense all pass.
- [ ] `make dev-init` — not run. This PR is e2e-test-only; it does not touch what the daemon reads (`/opt/kukeon` layout, walker, metadata files, CNI configs) or the build path. CLAUDE.md's daemon-parity guard is scoped to changes the daemon observes — none apply here.

Closes #491

### DIFF
--- a/e2e/e2e_kuke_apply_test.go
+++ b/e2e/e2e_kuke_apply_test.go
@@ -239,7 +239,6 @@ func TestKukeApply_Realm_VerifyState(t *testing.T) {
 
 	// Setup
 	runPath := getRandomRunPath(t)
-	mkdirRunPath(t, runPath)
 	realmName := generateUniqueRealmName(t)
 
 	// Cleanup: Delete realm
@@ -316,7 +315,6 @@ func TestKukeApply_Space_VerifyState(t *testing.T) {
 
 	// Setup
 	runPath := getRandomRunPath(t)
-	mkdirRunPath(t, runPath)
 	realmName := generateUniqueRealmName(t)
 	spaceName := generateUniqueSpaceName(t)
 
@@ -396,7 +394,6 @@ func TestKukeApply_Stack_VerifyState(t *testing.T) {
 
 	// Setup
 	runPath := getRandomRunPath(t)
-	mkdirRunPath(t, runPath)
 	realmName := generateUniqueRealmName(t)
 	spaceName := generateUniqueSpaceName(t)
 	stackName := generateUniqueStackName(t)
@@ -489,7 +486,6 @@ func TestKukeApply_Cell_VerifyState(t *testing.T) {
 
 	// Setup
 	runPath := getRandomRunPath(t)
-	mkdirRunPath(t, runPath)
 	realmName := generateUniqueRealmName(t)
 	spaceName := generateUniqueSpaceName(t)
 	stackName := generateUniqueStackName(t)
@@ -638,7 +634,6 @@ func TestKukeApply_MultiResource_VerifyState(t *testing.T) {
 
 	// Setup
 	runPath := getRandomRunPath(t)
-	mkdirRunPath(t, runPath)
 	realmName := generateUniqueRealmName(t)
 	spaceName := generateUniqueSpaceName(t)
 	stackName := generateUniqueStackName(t)
--- a/e2e/e2e_kuke_attach_test.go
+++ b/e2e/e2e_kuke_attach_test.go
@@ -104,11 +104,7 @@ func TestKuke_AttachDetach_KeepsTaskRunning(t *testing.T) {
 	// creates the metadata dir at provision time and sbsh binds the socket
 	// inside the container at task start. If this is missing, pkg/attach
 	// would fail in a way that's hard to attribute, so check explicitly.
-	cwd, err := os.Getwd()
-	if err != nil {
-		t.Fatalf("could not get working dir: %v", err)
-	}
-	socketPath := fs.ContainerSocketPath(filepath.Join(cwd, runPath),
+	socketPath := fs.ContainerSocketPath(runPath,
 		realmName, spaceName, stackName, cellName, "work")
 	waitForSocket(t, socketPath, 10*time.Second)
 
@@ -130,7 +126,7 @@ func TestKuke_AttachDetach_KeepsTaskRunning(t *testing.T) {
 	// triggering a clean exit.
 	time.Sleep(attachConnectGrace)
 
-	if err = session.Write(sbshDetachSequence); err != nil {
+	if err := session.Write(sbshDetachSequence); err != nil {
 		t.Fatalf("write sbsh detach sequence: %v", err)
 	}
 
--- a/e2e/e2e_kuke_attach_test.go
+++ b/e2e/e2e_kuke_attach_test.go
@@ -67,7 +67,6 @@ func TestKuke_AttachDetach_KeepsTaskRunning(t *testing.T) {
 	// (`--no-daemon`) so the per-container socket lives under the test's
 	// runPath rather than the host daemon's.
 	runPath := getRandomRunPath(t)
-	mkdirRunPath(t, runPath)
 
 	realmName := generateUniqueRealmName(t)
 	spaceName := generateUniqueSpaceName(t)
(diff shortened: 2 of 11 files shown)


eminwux:
**PR #492 Review — `fix: switch e2e run paths to t.TempDir() to drop shared tmp/ race`**

Clean swap from the bespoke `getRandomRunPath` + `mkdirRunPath` pair to `t.TempDir()`, and the six verifier helpers correctly drop their `filepath.Join(cwd, runPath)` wraps now that `runPath` is absolute. The fix matches #491's preferred option. One miss:

- **Stale `cwd` prepend** `e2e/e2e_kuke_attach_test.go:107-111`: same `filepath.Join(cwd, runPath)` pattern the PR removed from the six verifier helpers, but the attach test was missed. With `runPath` now absolute (e.g. `/tmp/TestKuke_AttachDetach.../001`), `filepath.Join(cwd, "/tmp/x")` returns `/cwd/tmp/x`, so `waitForSocket` polls the wrong path and the test will time out with `sbsh socket … did not appear within 10s`. Drop the `os.Getwd()` block and pass `runPath` straight to `fs.ContainerSocketPath`, matching the verifier helpers. (Likely not caught locally because the attach test needs containerd + an attachable cell to even start.)

LGTM once the attach-test `cwd` prepend is removed.

eminwux:
Addressed in 6e973a3:
- Dropped the `os.Getwd()` + `filepath.Join(cwd, runPath)` block in `e2e/e2e_kuke_attach_test.go` and now pass `runPath` straight to `fs.ContainerSocketPath`, matching the verifier helpers. Same `cwd` prepend pattern that was removed elsewhere in the PR — this one was missed. `go vet ./e2e/...` clean, `go test -run=__none__ ./e2e/...` compiles, `make test` green, pre-commit passes. The attach test itself was not re-run locally for the same reason called out in the PR body — it needs containerd + an attachable cell to start — but the fix is mechanical and the build attests no other call site references `cwd` here.

eminwux:
**PR #492 Re-review — `fix: switch e2e run paths to t.TempDir() to drop shared tmp/ race` @ `6e973a3`**

Prior blocker (`cwd` prepend in attach test) is cleanly addressed: `e2e/e2e_kuke_attach_test.go:107` now passes `runPath` straight to `fs.ContainerSocketPath`, and the stray re-declared `err :=` on the detach-sequence write was tidied in the same commit. Dead-symbol sweep confirms `mkdirRunPath` and the `"path"` import are fully gone with no stragglers; `os` / `path/filepath` remain used in every file touched.

LGTM.

PR #496, opened 2026-05-14
refactor: render kuketty TerminalDoc from cell fields via sbsh v0.11.2 inline builder

eminwux:
## Summary

- Profile-coupling rip (#494). `internal/controller/runner/attachable.go::writeKukettyMetadata` now renders the kuketty `TerminalDoc` from the in-memory cell `Tty.*` fields via sbsh v0.11.2's inline `BuildTerminalSpec` (the profile-free lane added in `eminwux/sbsh#209`). `WithProfile` / `WithProfilesDir` are gone, the unconditional `WithDisableSetPrompt(true)` safety belt is replaced by per-cell `WithPrompt(spec.Tty.Prompt) + WithDisableSetPrompt(spec.Tty.Prompt == "")`, and `WithOnInit([]api.ExecStep{...})` closes the silent drop where a cell with no profile but a non-empty `Tty.OnInit` produced an empty `Spec.Stages.OnInit`.
- Cell schema trimmed. `Profile` and `ProfilesDir` are deleted from both `intmodel.ContainerTty` and `ext.ContainerTty`, with the matching apischeme converters and `IsEmpty()` branches removed. The default `gopkg.in/yaml.v3` decoder silently drops unknown keys, so cell YAML that still carries `tty.profile:` / `tty.profilesDir:` parses cleanly and the fields are ignored — chosen over the "reject" path per the AC's pick-one-and-document leeway, because the surrounding container API uses non-strict YAML and rejection would force a YAML migration we have no current operators for. Documented in `ContainerTty`'s docstring.
- `cmd/kuke/run/run.go::containerTtysEqual` loses the corresponding `Profile` / `ProfilesDir` field comparisons; otherwise the drift detector would dereference a deleted field. No other call sites in the tree referenced either field.
[...]

Issue #495, opened 2026-05-14
bug: dev-init smoke deadlocks on stale pre-#349 daemon — reset rejects missing RootContainerID


eminwux:
## Context

`make dev-init` on a host with a stale-but-running `kukeond` cell from a prior session locks the dev-init loop in a deadlock that prevents a fresh re-bootstrap. Surfaced while running the CLAUDE.md-mandated `make dev-init` smoke on a clean working tree of `origin/main` (4164950).

The chain:

1. `scripts/dev-init.sh` checks `[ -d "${KUKEOND_CELL_DIR}" ]` and, when the directory is missing, skips `kuke daemon reset`. On this host the directory is *missing*, but the underlying containerd task (`kukeon_kukeon_kukeond_kukeond`, PID 2298446, started May 12) is *still RUNNING* because it survived a previous cleanup that removed the metadata file. The script's resume gate (`-d` check) does not look at the containerd task, so it misdetects "no prior daemon" and skips the reset.
2. dev-init proceeds to `kuke image load --from-docker kukeon-local:dev …` and then `kuke init --kukeond-image docker.io/library/kukeon-local:dev`. `kuke init` reports `cell kukeond: already existed` and `cell containers: started` — but containerd already has a running task at that ID, so no new container/task is created. The stale daemon (compiled from pre-#418 + pre-#482 code) keeps serving the socket.
3. The freshly-built `kuke` CLI sends a `CreateRealmArgs` whose `Doc.status.state` is serialized as the string `"Ready"` (per #482, `state_marshal.go::RealmState.MarshalJSON`). The stale daemon decodes into a pre-#482 `RealmState` whose `UnmarshalJSON` only accepts ints, so the call dies with:
   ```
   Error: json: cannot unmarshal string into Go struct field RealmStatus.Doc.status.state of type v1beta1.RealmState
   ```
   `kuke get realms` still works because read-side paths persist as strings on disk via the new daemon's writer — but no writer is running here, so the on-disk file is from an older string-encoded snapshot the read side happens to accept; `kuke create realm` (CLI → stale daemon) is the broken edge.
4. The natural recovery — `kuke daemon reset` — *also* fails:
   ```
   Error: stop kukeond cell: failed to stop cell containers: cell "kukeond" has no RootContainerID set
   ```
[...]

Issue #494, opened 2026-05-14
kuketty: render TerminalDoc from cell fields directly, drop TerminalProfile coupling


eminwux:
## Context

Phase 4 (#290 / PR #420) wired the cell's `Tty.Profile` + `Tty.ProfilesDir` through to sbsh's `pkg/builder.BuildTerminalSpec(WithProfile(...), WithProfilesDir(...))`. The intent was that sbsh's loader would resolve the YAML profile and stamp `Prompt`, `SetPrompt`, and `Stages` into the rendered `TerminalDoc.Spec`. That works when an operator declares a profile YAML on disk.

Two real problems shipped under that design:

1. **Inline `Tty.OnInit` is silently dropped.** The cell schema exposes `Tty.OnInit []TtyStage` at `pkg/api/model/v1beta1/container.go:120-121`, intended for an operator who wants init scripts without writing a profile YAML. The renderer at `internal/controller/runner/attachable.go::writeKukettyMetadata` (line 177) never reads `spec.Tty.OnInit` — only the profile lane (`WithProfile` + `WithProfilesDir`) produces stages. Audit of `.OnInit` consumers in kukeon (non-test): `internal/modelhub/container.go:137` (`IsEmpty()`), `pkg/api/model/v1beta1/container.go:178` (`IsEmpty()`), and `internal/apischeme/scheme.go:525-549` (round-trip) — none stamp it into `terminalSpec.Stages.OnInit`. The test-coverage gap that let this ship is tracked separately at #493.

2. **kukeon is needlessly coupled to sbsh's `TerminalProfile` concept.** All kukeon actually needs from sbsh is to render a `TerminalSpec` from in-memory cell fields. The current builder API has no inline `WithStages` / `WithPrompt`, so kukeon either (a) writes a `TerminalProfile` YAML to disk for sbsh to load back (the profile lane), or (b) accepts sbsh's hardcoded default profile and uses `WithDisableSetPrompt(true)` (lines 213-215) as a safety belt to keep `"(sbsh-$SBSH_TERM_ID) $PS1"` off the wire. Neither path lets a cell declare init scripts directly.

[...]

PR #497, opened 2026-05-14
test: lock comprehensive cell→TerminalDoc mapping and fix EnvInherit drop

### LINKED ISSUE #493: test: cell → TerminalDoc field mapping has no full-coverage test; Tty.OnInit silently dropped in no-profile path
## Context

Phase 4 (#290) wired `Tty.Profile` + `Tty.ProfilesDir` through to sbsh's builder so the cell profile resolves into `TerminalDoc.Spec.Stages.OnInit`. The PR (#420) added two table tests in `internal/controller/runner/attachable_test.go`:

- `TestWriteKukettyMetadata_ProfileConfigured` (line 362) — locks in profile-path OnInit rendering.
- `TestWriteKukettyMetadata_NoProfileKeepsSafeDefault` (line 417) — locks in *empty* `Spec.Stages.OnInit` when no profile is configured.

Neither test exercises the inline `Tty.OnInit []TtyStage` field declared at `pkg/api/model/v1beta1/container.go:120-121`. As a result, a real cell-config bug shipped under the green test suite: when an operator declares OnInit scripts inline on the cell (without a profile YAML), `internal/controller/runner/attachable.go::writeKukettyMetadata` drops them on the floor — the renderer only forwards profile options to `sbshbuilder.BuildTerminalSpec(...)` and never reads `spec.Tty.OnInit`.

Audit of `.OnInit` consumers in kukeon (excluding tests):

- `internal/modelhub/container.go:137` — `IsEmpty()` check only.
- `pkg/api/model/v1beta1/container.go:178` — `IsEmpty()` check only.
- `internal/apischeme/scheme.go:525-549` — external↔internal round-trip.
- **No renderer reads `Tty.OnInit` and stamps it into `terminalSpec.Stages.OnInit`.** The field is plumbed through the type system but never reaches the kuketty wire.

The deeper miss this issue tracks is the *category* of test that would have caught it: there is no test that constructs a `ContainerSpec` with **every** user-facing `Tty.*` field populated and asserts each one lands in the rendered `TerminalDoc.Spec`. The existing tests are field-by-field positive cases that each exercise one knob in isolation, so a field that is simply never read produces a clean test run.

## Proposal

Add a comprehensive cell → TerminalDoc mapping test in `internal/controller/runner/attachable_test.go` that:

1. Constructs a `ContainerSpec` with **every** field on `intmodel.ContainerTty` populated (currently: `Prompt`, `OnInit`, `LogFile`, `Profile`, `ProfilesDir`; plus all sibling `ContainerSpec` fields that flow into the TerminalDoc — `ID`, workload argv, `Attachable=true`).
2. Calls `r.writeKukettyMetadata(...)` and reads back the on-disk TerminalDoc.
3. Asserts that **every** field set on the input has a non-zero/expected counterpart on `doc.Spec` (or on `doc.Metadata` for naming fields). Where a field on the cell side has no direct TerminalSpec counterpart by current design, the assertion must be explicit: e.g. `if doc.Spec.X != "" { t.Errorf("X should be zero because design Y") }`. No silent skips.
4. Uses a small "exhaustive field set" pattern (reflect-based, or a hand-maintained checklist) so future additions to `ContainerTty` cause the test to fail until the renderer is updated and the test mapping extended.

The renderer fix (move inline `Tty.OnInit` into `terminalSpec.Stages.OnInit`, or remove the field from the schema) ships as a separate issue per scope. This issue is only about closing the test-coverage gap so the next "field declared but not read" regression fails CI.

## Acceptance criteria

- [ ] A new test in `internal/controller/runner/attachable_test.go` constructs a `ContainerSpec` with every field on `intmodel.ContainerTty` populated and asserts each one is reflected (or explicitly justified as unreflected) in the rendered `doc.Spec` / `doc.Metadata`.
- [ ] The test exercises both the profile-set and profile-empty branches of `writeKukettyMetadata`; in the profile-empty branch it confirms inline `Tty.OnInit` would land in `Spec.Stages.OnInit` (this is the assertion that is currently failing on the real code path).
- [ ] The test fails today (until the renderer is fixed in a follow-up) so it is committed alongside an `expect-fail` marker or with the renderer fix; reviewer decides which lane.
- [ ] A short doc comment on the new test explains the "every Tty field must map" rule so future contributors don't drop the pattern when adding fields.

## Out of scope

- Fixing `writeKukettyMetadata` to actually forward `Tty.OnInit` (separate bug — design call on whether the field should live on the cell at all, or be deleted in favor of profile-only).
- Adding a `WithStages` / `WithOnInit` option to sbsh's `pkg/builder` (separate sbsh-side issue).
- Reflection-based mapping enforcement at the type level (the test pattern is sufficient; full type-level enforcement is a larger lift).

## References

- Phase 4 issue: #290.
- Phase 4 PR: #420.
- Renderer: `internal/controller/runner/attachable.go::writeKukettyMetadata` (line 177).
- Cell schema: `pkg/api/model/v1beta1/container.go:120-121` (`Tty.OnInit`).
- Existing test gap: `internal/controller/runner/attachable_test.go:355-435` (only covers profile-set / profile-empty without inline OnInit).

eminwux:
## Summary

- Closes #493 by adding `TestWriteKukettyMetadata_AllTtyFieldsMap` — a comprehensive renderer test that constructs a `ContainerSpec` with every field on `intmodel.ContainerTty` populated and asserts each lands on `TerminalDoc.Spec`. A reflect-based whitelist (`mapperedFields`) inside `ttyFieldsCoveredByMapping` fails the test when a new `ContainerTty` field lands without being mapped *or* when the renderer drops one silently — the pre-#494 `Tty.OnInit` failure mode this test pattern is designed to catch.
- Pins three renderer-default invariants the cell schema doesn't surface but the renderer is responsible for stamping on every `TerminalDoc`: `SocketFile`, `CaptureFile`, `RunPath`, and **`EnvInherit=true`**. The last one is the load-bearing addition: without it, sbsh's terminal runner (`sbsh@v0.11.2/internal/terminal/terminalrunner/terminal.go:54`) spawns the workload with only `HOME` + `SBSH_*` set, **stripping the user-supplied env and the `KUKEON_*` identity vars** that the OCI `Process.Env` already carries. This is a #494 regression: pre-#494, sbsh's profile loader's no-profile fallback defaulted `EnvInherit: true` (`sbsh@v0.11.1/internal/profile/profile.go:454`) — the inline-builder migration silently lost that default.
- Behavior fix in `internal/controller/runner/attachable.go::writeKukettyMetadata`: adds `sbshbuilder.WithEnvInherit(true)` to the option slice. Bundled into the same PR as #493 because the test is the regression guard for the fix — shipping them separately would let a future renderer refactor re-drop `EnvInherit` undetected.

## Test plan

- [x] `make test` — full unit suite green, including the new `TestWriteKukettyMetadata_AllTtyFieldsMap` and the three pre-existing `TestWriteKukettyMetadata_*` tests.
- [x] `go build ./...` + `go vet ./...` — clean.
- [x] `pre-commit run --files internal/controller/runner/attachable.go internal/controller/runner/attachable_test.go` — clean (`golangci-lint` green after inlining the field-mapper map to satisfy `gochecknoglobals` and using `range typ.NumField()` to satisfy `intrange`).
- [x] Direct end-to-end smoke against `--no-daemon`: launched a cell with `attachable: true` and `env: [FOO=bar, HELLO=world]`, then read `/tmp/workload-env.txt` written by the workload itself. With the fix, the workload now sees `FOO=bar`, `HELLO=world`, every `KUKEON_*` identity var (`KUKEON_CELL_NAME=envatt`, `KUKEON_REALM=default`, etc.), plus `PATH` and the `SBSH_*` markers. Without the `WithEnvInherit(true)` line, only `HOME` + `SBSH_*` + `TERM`/`COLORTERM` survive — the original repro.
- [ ] `make dev-init` daemon-parity smoke — host blocked by #495 (stale pre-#349 daemon predates `RootContainerID`, so `kuke daemon reset` fails before the rebuild step). The `--no-daemon` smoke above exercises the same `writeKukettyMetadata` code path the daemon would run, so the behavior is covered; the daemon-parity tail (which is the part #496 added) is unaffected by this PR (no `/opt/kukeon` schema changes).

Closes #493

### DIFF
--- a/internal/controller/runner/attachable.go
+++ b/internal/controller/runner/attachable.go
@@ -202,6 +202,21 @@ func (r *Exec) writeKukettyMetadata(
 		// lane that loaded sbsh TerminalProfile YAML from disk.
 		sbshbuilder.WithPrompt(prompt),
 		sbshbuilder.WithDisableSetPrompt(prompt == ""),
+		// Spec.EnvInherit=true tells sbsh's terminal runner to forward
+		// kuketty's os.Environ() (which is the container's OCI Process.Env
+		// — user-supplied containerSpec.Env merged with kukeon's KUKEON_*
+		// identity vars at ctr.BuildContainerSpec) to the workload child.
+		// Without it, sbsh's runner spawns the workload with only HOME +
+		// SBSH_* set (sbsh@v0.11.2/internal/terminal/terminalrunner/
+		// terminal.go:54) — every user env entry and every KUKEON_* entry
+		// gets stripped at the kuketty → workload boundary. Pre-#494 the
+		// profile lane's no-profile fallback defaulted EnvInherit=true
+		// (sbsh@v0.11.1/internal/profile/profile.go:454), so the inline-
+		// builder migration silently regressed env passthrough — this
+		// option restores the kukeon contract that the OCI Process.Env IS
+		// the workload's env. Pinned by the comprehensive renderer test in
+		// attachable_test.go so a future refactor cannot drop it.
+		sbshbuilder.WithEnvInherit(true),
 	}
 	if spec.Tty != nil && len(spec.Tty.OnInit) > 0 {
 		// Inline OnInit (#494): map the cell's TtyStage{Script} entries to
--- a/internal/controller/runner/attachable_test.go
+++ b/internal/controller/runner/attachable_test.go
@@ -26,6 +26,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"reflect"
 	"testing"
 
 	"github.com/eminwux/kukeon/internal/ctr"
@@ -163,7 +164,12 @@ func TestWriteKukettyMetadata_KukeonGroupSet(t *testing.T) {
 	}
 	wantArgs := []string{"-c", "echo hello"}
 	if len(doc.Spec.CommandArgs) != len(wantArgs) {
-		t.Fatalf("Spec.CommandArgs len = %d, want %d (full=%v)", len(doc.Spec.CommandArgs), len(wantArgs), doc.Spec.CommandArgs)
+		t.Fatalf(
+			"Spec.CommandArgs len = %d, want %d (full=%v)",
+			len(doc.Spec.CommandArgs),
+			len(wantArgs),
+			doc.Spec.CommandArgs,
+		)
 	}
 	for i, want := range wantArgs {
 		if doc.Spec.CommandArgs[i] != want {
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #497 Review — `test: lock comprehensive cell→TerminalDoc mapping and fix EnvInherit drop`**

Clean fix + regression guard. The behavior change is one line (`WithEnvInherit(true)`) with a clear citation to the sbsh runner code path that strips env without it, and the new test pins all three concentric invariants — reflective field coverage, per-field behavior, and the four renderer-default stamps (EnvInherit, SocketFile, CaptureFile, RunPath). Bundling the fix with its test in the same PR is the right call per the author's reasoning. No correctness concerns, no sensitive data leaks, no dead symbols, no public-API parity gap (EnvInherit is intentionally a renderer-internal default, not a cell-schema knob). Commit is signature-verified, CI is fully green.

- **Nit (non-blocking)** `internal/controller/runner/attachable_test.go:383`: `mapperedFields` reads as a typo for `mappedFields`. Private test-only identifier so impact is nil; leave as-is or fix on the next pass.

`make dev-init` daemon-parity smoke is correctly deferred — author documented the host constraint (blocked by #495), and the `--no-daemon` direct smoke exercises the same `writeKukettyMetadata` code path. I'll run it host-side pre-merge.

LGTM.

PR #498, opened 2026-05-14
fix: name per-container breaking fields in DiffCell so apply errors stop printing []

### LINKED ISSUE #469: apply: name the breaking-change fields instead of printing [] (phase 1)
## Context

`docs/cli-use-cases.md` `## Workload lifecycle` / `### Apply (declarative, multi-document)` makes `apply` the documented "update path" for cells whose on-disk spec diverges from the file. Today, when `apply` does refuse a divergent update (because some changed field is classified as breaking), the error message is uninformative:

```
$ sudo kuke apply -f /tmp/cli-test/hello-world-divergent.yaml
Cell "hello-world": failed
  Error: cell "hello-world" has breaking changes: []. Delete the cell and recreate it with the new spec
```

The breaking-field list is empty. The operator has no way to tell what the CLI considered breaking.

Root cause: in `internal/controller/apply/diff.go`, `DiffCell`'s container loop (around lines 391-422) detects per-container breaking changes via `diffContainerSpec`, copies them into `result.Containers[i].BreakingChanges`, and bumps `result.ChangeType` to `ChangeTypeBreaking` — but never appends those field names into the cell-level `result.BreakingChanges` slice. `reconcile.go:367-373` then prints the empty cell-level slice.

Cell-spec-level fields (e.g. `metadata.name`, `spec.realmName`) are correctly appended to `result.BreakingChanges` elsewhere in `diff.go`; only the per-container path is missing the propagation.

## Proposal

When `DiffCell` records a per-container breaking change, also append the field names — qualified with the container ID, e.g. `containers[web].image` — to the cell-level `result.BreakingChanges`. Pure messaging fix; no semantic change to which diffs are accepted vs refused.

## Acceptance criteria

- [ ] When `kuke apply -f` refuses a structurally-breaking change rooted in a non-root container, the error message names the specific field(s) — no more empty `[]`. Per-container fields are qualified with the container ID (e.g. `containers[web].image`).
- [ ] No semantic change to which diffs are accepted vs refused — this is messaging only.
- [ ] Unit test in `internal/controller/apply/` covers a divergent non-root container `image` change and asserts the error message names `containers[<id>].image`.

## Out-of-scope follow-ups

- #485 — apply: accept in-place image/command/args updates on non-root containers (phase 2). Was AC1 in this issue's original framing.
- #486 — apply: re-materialize containers after `kuke kill cell` (phase 3). Was AC2 in this issue's original framing.

## Notes

- Touches: `internal/controller/apply/diff.go` (lines ~391-422); reconcile-side message formatting if the qualifier rendering is shared there.
- Re-sized from the original framing (which bundled the messaging fix with two larger semantic changes — accepting in-place updates, and re-materializing after `kill cell`). Both larger pieces were split into the follow-ups above.

## References

- `docs/cli-use-cases.md` — `## Workload lifecycle` / `### Apply (declarative, multi-document)` (line ~258).

eminwux:
## Summary

- Before this fix, `kuke apply -f` against a cell with a structurally-breaking change rooted in a *non-root* container (e.g. a child container `image` bump) printed `cell "<name>" has breaking changes: []. Delete the cell and recreate it with the new spec` — empty `[]`, no signal to the operator about which field was rejected.
- Root cause: in `internal/controller/apply/diff.go`, `DiffCell` aggregates per-container breaking field names into `result.Containers[i].BreakingChanges` (lines 391–422) and bumps `result.ChangeType` to `ChangeTypeBreaking`, but never copies them up into the cell-level `result.BreakingChanges` slice. `reconcile.go:367–373` then prints that empty cell-level slice.
- Pure messaging fix: when a per-container update is classified as breaking, also append each field name into the cell-level slice, qualified with the container ID (e.g. `containers[web].image`). No semantic change to which diffs are accepted vs refused.

## Test plan

- [x] `make test` — full unit suite green, including the new `TestDiffCell_NonRootContainerBreaking_NamesField` covering a non-root container `image` change and asserting `containers[<id>].image` lands in `BreakingChanges`.
- [x] `go vet ./...` — clean.
- [x] `pre-commit run --files internal/controller/apply/diff.go internal/controller/apply/diff_test.go` — clean (go fmt, golangci-lint, addlicense, etc.).
- [ ] `make dev-init` smoke — not run; this change is confined to `internal/controller/apply/diff.go` and does not touch the build path, the daemon, or anything under `/opt/kukeon` (per `CLAUDE.md`'s "When a change touches the build path, the daemon, or anything under `/opt/kukeon`" criterion the smoke is not required).

Closes #469

### DIFF
--- a/internal/controller/apply/diff.go
+++ b/internal/controller/apply/diff.go
@@ -407,6 +407,15 @@ func DiffCell(desired, actual intmodel.Cell) CellDiffResult {
 				result.HasChanges = true
 				if containerDiff.ChangeType == ChangeTypeBreaking {
 					result.ChangeType = ChangeTypeBreaking
+					// Qualify per-container breaking fields with the container ID
+					// so the cell-level error message in reconcile.go names them
+					// instead of printing an empty `[]`.
+					for _, field := range containerDiff.BreakingChanges {
+						result.BreakingChanges = append(
+							result.BreakingChanges,
+							fmt.Sprintf("containers[%s].%s", id, field),
+						)
+					}
 				} else if result.ChangeType == ChangeTypeNone {
 					result.ChangeType = containerDiff.ChangeType
 				}
--- a/internal/controller/apply/diff_test.go
+++ b/internal/controller/apply/diff_test.go
@@ -293,3 +293,53 @@ func TestDiffCell_SynthesizedRoot_NoChange(t *testing.T) {
 		t.Error("synthesized root must not flag RootContainerChanged on re-apply")
 	}
 }
+
+// TestDiffCell_NonRootContainerBreaking_NamesField pins issue #469: when a
+// breaking change is rooted in a non-root container (e.g. image bump),
+// DiffCell must propagate the field name into the cell-level
+// BreakingChanges slice qualified by container ID. Otherwise the
+// reconcile-side error formatter ("cell %q has breaking changes: %v") prints
+// an empty `[]` and the operator has no way to tell what was rejected.
+func TestDiffCell_NonRootContainerBreaking_NamesField(t *testing.T) {
+	desired := intmodel.Cell{
+		Metadata: intmodel.CellMetadata{Name: "hello-world"},
+		Spec: intmodel.CellSpec{
+			RealmName: "default",
+			SpaceName: "default",
+			StackName: "default",
+			Containers: []intmodel.ContainerSpec{
+				{ID: "root", Root: true, Image: "busybox:latest"},
+				{ID: "web", Image: "nginx:1.27"},
+			},
+		},
+	}
+
+	actual := intmodel.Cell{
+		Metadata: intmodel.CellMetadata{Name: "hello-world"},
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #498 Review — `fix: name per-container breaking fields in DiffCell so apply errors stop printing []`**

Tight, minimal messaging fix. Per-container breaking field names now propagate into the cell-level `BreakingChanges` slice qualified as `containers[<id>].<field>`, which is exactly what `reconcile.go:367-373` prints. Pure semantics-preserving change — no shift in what `DiffCell` classifies as breaking vs additive vs compatible. New test pins the non-root `image`-bump path directly, and the existing root/realm/space/stack breaking paths were already populating the cell-level slice, so the fix closes the only gap. No secrets, no dead code, no convention drift; `make dev-init` smoke correctly deferred per `CLAUDE.md` (the change is confined to the apply controller and touches none of build path / daemon / `/opt/kukeon`). Sole commit is signed and verified.

LGTM.

PR #489, opened 2026-05-14
fix: scope reconcile walker to /opt/kukeon/data so /opt/kukeon/bin stops logging ERROR

eminwux:
## Summary

- The reconcile loop was walking `/opt/kukeon/*` and treating every subdirectory as a realm. `/opt/kukeon/bin/` (where `kuke init` stages `kuketty`) tripped the walker into an `ERROR msg=\"metadata file does not exist\" file=/opt/kukeon/bin/metadata.json` line every 30s — hot misleading signal for operators tailing `kuke daemon logs`, and a false page for anyone alerting on ERROR severity.
- Move kukeon's on-disk metadata one level deeper under `<RunPath>/data/` (new `consts.KukeonMetadataSubdir`, new `fs.MetadataRoot()` helper) so walkers — `ListRealms` and the subnet allocator — scope themselves to the metadata subtree. Siblings of the RunPath (`bin/kuketty`, `.kukeon-instance.json`) intentionally fall outside it. Per the issue comment's preferred option ("Use /opt/kukeon/data for metadata, … every single directory is a realm, etc.").
- Composed: `fs.RealmMetadataDir` now joins `MetadataRoot(runPath)/<realm>` (Space/Stack/Cell/Container path helpers compose off it, so they pick up the prefix automatically). `cni.SubnetAllocator` stores `dataRoot = <RunPath>/data` and scans only there. `kuke daemon reset --purge-system` targets `<RunPath>/data/kuke-system`. `scripts/dev-init.sh`'s `SYSTEM_REALM_DIR` and `ATTACH_SMOKE_BASE` follow.

## Behavior

- After `make dev-init` the reconcile loop is silent at ERROR level: a 70-second observation window with two full reconcile ticks logged only `INFO msg=\"reconcile ok\"` (verified by `kuke daemon logs` after the smoke succeeded — zero `level=ERROR` lines for `/opt/kukeon/bin/metadata.json`, and zero overall during steady-state reconcile cycles).
- A missing `metadata.json` under a real realm directory still surfaces as ERROR — the fix only excludes non-metadata siblings of the RunPath, not malformed entries inside `data/`. Covered by `TestListRealms_MissingRealmMetadataStillSkipsBrokenRealm`.
[...]

PR #502, opened 2026-05-14
fix: scope resolveRealmDirs all-realms walk to MetadataRoot

### LINKED ISSUE #500: runner: resolveRealmDirs walks RunPath instead of MetadataRoot, breaking flagless list commands
## Repro

On a host where `kuke init` has been run (so `default` and `kuke-system` realms exist under `/opt/kukeon/data/`), against ref `894d93d` (current `main`):

```bash
kuke get cell           # no flags
kuke get cells          # alias, same effect
kuke get stacks         # any of the list verbs that go through resolveRealmDirs
```

## Observed

The list comes back empty (or with the wrong rows — entries from one level too shallow). Before `6eed38f` ("fix: get list filters default to 'default' when flag is omitted"), this worked because the CLI silently defaulted `--realm` to `"default"`, sending the call down the single-realm branch of `resolveRealmDirs`. Now that the list path leaves `--realm` truly empty when unset, the all-realms branch runs — and that branch is broken.

## Root cause

`internal/controller/runner/get.go:200-238` — `resolveRealmDirs(realmName)` has two branches:

- `realmName != ""` → uses `fs.RealmMetadataDir(r.opts.RunPath, realmName)` correctly, which expands to `<RunPath>/data/<realm>`.
- `realmName == ""` → uses `base := r.opts.RunPath` and `os.ReadDir(base)`, listing entries directly under `<RunPath>` (e.g. `/opt/kukeon`).

After commit `4164950` (#489), the daemon's `RunPath` has non-realm siblings (`/opt/kukeon/bin` for kuketty staging, plus `.kukeon-instance.json`), and the real realm dirs live under `<RunPath>/data/`. The empty-realm branch therefore returns `["/opt/kukeon/bin", "/opt/kukeon/data"]` as `realmDirs`. Subsequent levels in `ListCells` / `ListSpaces` / `ListStacks` / `ListContainers` then walk those as if they were realms — they enumerate `bin`'s (empty) children, then `data`'s children (the actual realm names like `default`/`kuke-system`) but treat them as **spaces**, the real spaces as **stacks**, etc. By the time the loop bottoms out, it has gone one level too shallow at every step and either returns nothing or returns entries from the wrong tree depth.

`4164950` (#489) applied exactly this fix to `ListRealms` (`internal/controller/runner/get.go:511` — `realmsDir := fs.MetadataRoot(r.opts.RunPath)`) but did not propagate it into the shared helper. Every consumer of `resolveRealmDirs` is therefore still broken for the empty-realm case.

## Suggested fix

In `resolveRealmDirs` (around line 201), change the all-realms branch to walk the metadata root, the same scope `ListRealms` already uses:

```go
base := fs.MetadataRoot(r.opts.RunPath)
```

Single-line change. The existing tests should cover it; add a runner-level test that asserts `resolveRealmDirs("")` returns realm metadata dirs only when `<RunPath>` has non-metadata siblings (mirror the `/opt/kukeon/bin` setup from #489's test).

## Out of scope

- Whether bare `kuke get cell` (zero flags) should target the `default` realm implicitly vs. list cluster-wide — that's a separate design call surfaced by the same conversation. This issue is purely about restoring the existing "cluster-wide" path so users who *don't* pass `--realm` see anything at all.

## References

- Surfaced by commit `6eed38f` ("fix: get list filters default to 'default' when flag is omitted"), which stopped masking the bug.
- Mirror of the fix from #489 / commit `4164950` ("fix: scope reconcile walker to /opt/kukeon/data so /opt/kukeon/bin stops logging ERROR"), which fixed `ListRealms` only.

eminwux:
## Summary

- Mirror the #489 fix into the shared \`resolveRealmDirs\` helper so the all-realms branch walks \`<RunPath>/data\` instead of \`<RunPath>\`. Without this, every flagless list verb (\`kuke get cells\`, \`kuke get spaces\`, \`kuke get stacks\`, \`kuke get containers\`) descends one level too shallow — \`/opt/kukeon/bin\` and \`/opt/kukeon/data\` get treated as realms, the real realms get mistaken for spaces, and the walker bottoms out before finding any cells.
- Add a regression test exercising \`ListSpaces("")\` against a \`<RunPath>\` seeded with the same \`/bin/kuketty\` + \`.kukeon-instance.json\` non-metadata siblings the \`ListRealms\` test from #489 already guards against, so resolveRealmDirs and ListRealms share the same trap.

## Test plan

- [x] \`make test\` — full unit-test suite passes (\`exit=0\`).
- [x] \`go vet ./...\` — clean.
- [x] \`go build ./...\` — clean.
- [x] \`pre-commit run --files internal/controller/runner/get.go internal/controller/runner/resolve_realm_dirs_test.go\` — all hooks pass.
- [x] \`sudo make dev-init\` — daemon-parity tail matches CLAUDE.md regression guard (both \`kuke get realms\` and \`kuke get realms --no-daemon\` show the \`default\` + \`kuke-system\` rows); attach smoke (\`dev-init-attach\` cell) lifecycle clean.
- [x] End-to-end repro against the live daemon: \`sudo ./kuke get cells\` / \`get spaces\` / \`get stacks\` now return rows from both \`default\` and \`kuke-system\` realms with no \`--realm\` flag (previously empty).

Closes #500

### DIFF
--- a/internal/controller/runner/get.go
+++ b/internal/controller/runner/get.go
@@ -198,7 +198,11 @@ func (r *Exec) ListSpaces(realmName string) ([]intmodel.Space, error) {
 // returns a single realm directory if it exists. Returns empty slice if realm/base
 // doesn't exist (graceful handling for listing operations).
 func (r *Exec) resolveRealmDirs(realmName string) ([]string, error) {
-	base := r.opts.RunPath
+	// Scope the all-realms walk to <RunPath>/data so non-metadata
+	// siblings of the RunPath (e.g. /opt/kukeon/bin staging kuketty, or
+	// the .kukeon-instance.json file) are not mistaken for realm
+	// directories. Mirrors the fix applied to ListRealms in #489.
+	base := fs.MetadataRoot(r.opts.RunPath)
 	var realmDirs []string
 
 	if strings.TrimSpace(realmName) != "" {
--- a/internal/controller/runner/resolve_realm_dirs_test.go
+++ b/internal/controller/runner/resolve_realm_dirs_test.go
@@ -1,0 +1,108 @@
+//go:build !integration
+
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package runner_test
+
+import (
+	"context"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/eminwux/kukeon/internal/controller/runner"
+	"github.com/eminwux/kukeon/internal/metadata"
+	"github.com/eminwux/kukeon/internal/util/fs"
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
+)
+
+// TestListSpaces_AllRealms_NonRealmSiblingUnderRunPathIsIgnored is the
+// regression guard for #500: when callers pass an empty realm filter,
+// the runner must walk <RunPath>/<KukeonMetadataSubdir> (i.e.
+// /opt/kukeon/data) rather than <RunPath> itself. Otherwise non-metadata
+// siblings of the RunPath (e.g. /opt/kukeon/bin staging the kuketty
+// binary, or the .kukeon-instance.json file) get enumerated as realms,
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #502 Review — `fix: scope resolveRealmDirs all-realms walk to MetadataRoot`**

Clean mirror of the #489 fix into the shared helper — pre-PR, `os.ReadDir(r.opts.RunPath)` would enumerate `/opt/kukeon/bin` and `/opt/kukeon/data` as "realms" and demote the real realm dirs to "spaces" one level deeper, breaking every flagless `kuke get …` verb that fans out via `resolveRealmDirs`. Post-PR, the walk is anchored at `fs.MetadataRoot(r.opts.RunPath)` and the new test pins the regression with the same `bin/kuketty` + `.kukeon-instance.json` non-metadata siblings the `ListRealms` test from #489 already guards against.

Verified: the diff is the only resolution path for `ListSpaces` / `ListStacks` / `ListCells` (via `resolveRealmDirs`) and the patched `ListRealms` already uses the same `fs.MetadataRoot` anchor — no other call sites in `internal/controller/runner/get.go` walk `RunPath` directly. CI green, commit signed, no secrets, no dead code, license header + colocated test follow project conventions, and the dev-init smoke from CLAUDE.md's regression guard is checked off in the test plan.

LGTM to merge.

PR #505, opened 2026-05-14
fix: re-materialize containers in apply after kuke kill cell (phase 3)

### LINKED ISSUE #486: apply: re-materialize containers after kuke kill cell (phase 3)
## Context

`docs/cli-use-cases.md` `## Workload lifecycle` / `### Apply (declarative, multi-document)` names `kuke kill cell` as an example of divergent state that `apply` should reconcile: spec is intact on disk, but containers are gone. Observed:

```
$ sudo kuke run -f docs/examples/hello-world.yaml -d        # exit 0, cell Ready
$ sudo kuke kill cell hello-world --realm default --space default --stack default
$ sudo kuke get cells          # hello-world is Stopped (containers gone)
$ sudo kuke apply -f docs/examples/hello-world.yaml
Cell "hello-world": unchanged
# exit 0 — but the cell stays Stopped; apply did not re-materialize anything
```

`apply` reports `unchanged` and walks away, because the apply path today only diffs spec-vs-spec and not spec-vs-runtime.

## Proposal

After `DiffCell` reports no spec changes, also check runtime state. When the cell's containers are gone (or the cell is `Stopped`) but the spec is in sync, treat that as a re-materialize trigger: recreate and start the containers, and report `Cell <name>: updated` with a per-component summary naming the recreated containers.

## Acceptance criteria

- [ ] `kuke apply -f` against a cell after `kuke kill cell` (spec equal, runtime missing) re-creates and starts the containers and reports `Cell <name>: updated`. Exit code 0.
- [ ] After a successful re-materialize, `kuke get cells` shows the cell `Ready`.
- [ ] No regression for the no-divergence case: `apply` against a fully-Ready cell with matching spec still reports `Cell <name>: unchanged`, exit 0.

## Dependencies

Predecessor: #469 (phase 1 — empty-`[]` error message fix). Code-level coupling is light: phase 1 touches cell-level breaking-changes aggregation; this phase adds a runtime-state check in the apply path. Sequenced after #469 by convention; can land before or after phase 2 (#485).

## Notes

- AC2 in the original framing of #469.

## References

- `docs/cli-use-cases.md` — `## Workload lifecycle` / `### Apply (declarative, multi-document)`.

eminwux:
## Summary

- `kuke apply -f <cell.yaml>` now detects a spec-equal, runtime-Stopped cell (typically after `kuke kill cell`) and calls `StartCell` to re-create + restart the containers, reporting `Cell <name>: updated` with a per-component summary naming the recreated containers.
- The trigger is conservative: only `CellStateStopped` re-materializes. `Failed` stays sticky (the user should investigate; `markCellFailedAfterStartupFailure` writes the original cause into `Status.Reason/Message` and a silent recreate would mask it). `Pending` and `Unknown` are mid-lifecycle states the daemon reconciler loop normalizes on its own — apply does not race them.
- Root containers are excluded from the per-component summary to match the existing container-diff branch in `ReconcileCell`: users author workload containers; the auto-default root is an implementation detail.

## Test plan

- [x] `go test ./...` (full unit suite, all packages green)
- [x] `make dev-init` daemon-parity check matches the regression guard for both `kuke get realms` and `kuke get realms --no-daemon`
- [x] End-to-end against the docs example (AC #1 + AC #2): `kuke run -f docs/examples/hello-world.yaml -d` → `kuke kill cell hello-world` (state becomes Stopped, exit 0) → `kuke apply -f docs/examples/hello-world.yaml` prints `Cell "hello-world": updated` with `container "web" re-materialized`, exit 0 → `kuke get cells` shows `hello-world` `Ready`
- [x] AC #3 no-regression: a second `kuke apply -f` against the now-Ready cell prints `Cell "hello-world": unchanged`, exit 0 (StartCell is not invoked)
- [x] Unit tests in `internal/controller/apply_rematerialize_test.go` cover the three ACs plus the `Failed`-cell carve-out
- [x] `pre-commit run --files <touched>` (go fmt, golangci-lint, addlicense, gitlint, …)

Closes #486

### DIFF
--- a/internal/controller/apply/reconcile.go
+++ b/internal/controller/apply/reconcile.go
@@ -346,6 +346,23 @@ func ReconcileCell(r runner.Runner, desired intmodel.Cell) (ReconcileResult, err
 	// Diff desired vs actual
 	diff := DiffCell(desired, actual)
 	if !diff.HasChanges {
+		// Spec is in sync, but the cell's runtime may have been torn down
+		// out of band — typically by `kuke kill cell`, which leaves the
+		// cell document intact and flips Status.State to Stopped. Without
+		// this branch, apply walks away reporting "unchanged" and the cell
+		// stays Stopped forever (issue #486 — docs/cli-use-cases.md's
+		// "Apply (declarative, multi-document)" names the kill-then-apply
+		// flow as a divergence apply must reconcile).
+		if cellNeedsRematerialize(actual) {
+			rematerialized, startErr := r.StartCell(actual)
+			if startErr != nil {
+				return result, fmt.Errorf("failed to re-materialize cell: %w", startErr)
+			}
+			result.Action = actionUpdated
+			result.Resource = rematerialized
+			result.Changes = rematerializeChanges(rematerialized)
+			return result, nil
+		}
 		result.Resource = actual
 		return result, nil
 	}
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #505 Review — `fix: re-materialize containers in apply after kuke kill cell (phase 3)`**

Tight, well-scoped fix. The new "spec-equal but `Stopped`" branch in `ReconcileCell` sits cleanly in front of the existing diff-found-no-changes path, only triggers when the cell's runtime was torn down out of band (which `runner.KillCell` and `runner.StopCell` both signal by writing `CellStateStopped`), and lets `StartCell` persist the post-restart status via the existing `PopulateAndPersistCellContainerStatuses` path — no double-write required. The Failed/Pending/Unknown carve-outs are well-justified in the code comments and locked in by a unit test. Tests cover the three ACs. No correctness or sensitive-data concerns.

LGTM.

PR #506, opened 2026-05-14
fix: re-materialize cell containers on apply after kuke kill (phase 3)

### LINKED ISSUE #486: apply: re-materialize containers after kuke kill cell (phase 3)

eminwux:
## Summary

- Closes the gap behind `docs/cli-use-cases.md`'s "apply updates a divergent existing cell" claim where the divergence is runtime-only. After `kuke kill cell` deletes the containerd containers and stamps the cell `Stopped`, the spec on disk is still equal to the user's YAML — and apply walked away with `unchanged`, leaving the cell stuck `Stopped`. Now `ReconcileCell` checks `actual.Status.State` after `DiffCell` reports no spec changes; when the persisted cell is `Stopped` or `Failed`, it re-materializes the containers via `r.StartCell(desired)` and reports `Cell <name>: updated` with `runtime re-materialized`.
- `internal/controller/apply/reconcile.go`: new `cellRuntimeNeedsRematerialize` helper plus the runtime-state branch inserted between the no-changes return and the root-container-changed check. `StartCell` is the right choke point — it handles "container doesn't exist, create fresh" and the idempotent-skip path for a Ready cell, so the re-materialize call is safe regardless of how much of the old runtime survived.
- `internal/controller/apply/reconcile_test.go` (new): unit tests covering AC1 (Stopped → re-materialize → updated), AC3 (Ready → unchanged, StartCell skipped), the Failed branch, and StartCell-error surfacing. Uses an embedded `runner.Runner` interface field as the stub-default so unmocked methods panic instead of silently zero-valuing — keeps the test focused on the new code path.

### Why scope to Stopped + Failed only

The issue describes the kill-cell scenario specifically, which lands at `CellStateStopped` (see `kill.go:294`). `Failed` is included because the runtime-divergence semantics are identical (root task crashed → containers gone or unhealthy → spec-on-disk says "should be Ready"). `Pending`/`Unknown` are excluded so creation flows and races during status-population don't trigger spurious recreates. `Ready` is excluded so a healthy cell isn't disturbed — `StartCell`'s idempotent-skip would no-op anyway, but skipping the call entirely is cleaner.

## Test plan

- [x] `make test` — full unit suite passes, including the four new `TestReconcileCell_*` cases (`StoppedCellRematerializes`, `ReadyCellUnchanged`, `FailedCellRematerializes`, `RematerializeStartCellErrorSurfaces`).
- [x] `go vet ./...` — clean.
- [x] `pre-commit run --files internal/controller/apply/reconcile.go internal/controller/apply/reconcile_test.go` — all hooks pass (`golangci-lint` flagged the initial `switch`-based helper as non-exhaustive; rewritten as an `if` since `Ready`/`Pending`/`Unknown` deliberately fall through to "no re-materialize").
- [x] `sudo make dev-init` — daemon-parity tail matches CLAUDE.md regression guard (both `kuke get realms` and `kuke get realms --no-daemon` show the `default` + `kuke-system` rows); attach smoke (`dev-init-attach` cell) lifecycle clean.
- [x] End-to-end repro against the live daemon, exact issue scenario:
  - `sudo ./kuke run -f docs/examples/hello-world.yaml -d` → `Ready`.
  - `sudo ./kuke kill cell hello-world --realm default --space default --stack default` → cell `Stopped`, containers gone.
  - `sudo ./kuke apply -f docs/examples/hello-world.yaml` → `Cell "hello-world": updated` / `- runtime re-materialized`, exit 0.
  - `sudo ./kuke get cells --realm default` → cell back to `Ready` (AC2).
  - `sudo ./kuke apply -f docs/examples/hello-world.yaml` against the now-Ready cell → `Cell "hello-world": unchanged`, exit 0 (AC3, no regression).
- [ ] `make e2e` — not run (e2e harness needs the docker-build + privileged container loop; `dev-init` exercises the build path and daemon bring-up, and the manual end-to-end above covers the exact issue scenario). Reviewer-environment defer.

Closes #486

### DIFF
--- a/internal/controller/apply/reconcile.go
+++ b/internal/controller/apply/reconcile.go
@@ -346,6 +346,23 @@ func ReconcileCell(r runner.Runner, desired intmodel.Cell) (ReconcileResult, err
 	// Diff desired vs actual
 	diff := DiffCell(desired, actual)
 	if !diff.HasChanges {
+		// Spec on disk matches the desired document, but the cell may have
+		// lost its runtime — e.g. `kuke kill cell` deletes the containerd
+		// containers and stamps the cell Stopped while leaving the metadata
+		// intact. A spec-only diff returns "unchanged" in that state, which
+		// silently violates the declarative contract that apply enforces
+		// "spec on disk → Ready cell". Re-derive the runtime when the
+		// persisted cell state says the containers are gone (issue #486).
+		if cellRuntimeNeedsRematerialize(actual) {
+			started, startErr := r.StartCell(desired)
+			if startErr != nil {
+				return result, fmt.Errorf("failed to re-materialize cell containers: %w", startErr)
+			}
+			result.Action = actionUpdated
+			result.Resource = started
+			result.Changes = []string{"runtime re-materialized"}
+			return result, nil
+		}
 		result.Resource = actual
 		return result, nil
 	}
@@ -517,6 +534,17 @@ func ReconcileContainer(r runner.Runner, desired intmodel.Container) (ReconcileR
 	}
 
 	return result, nil
+}
+
+// cellRuntimeNeedsRematerialize reports whether a spec-equal cell has lost
+// its runtime — typically after `kuke kill cell` deletes the containerd
+// containers and stamps the cell Stopped, but also covers Failed cells.
+// Apply must treat that as a re-materialize trigger so the declarative
+// invariant ("spec on disk → Ready cell") survives a runtime wipe. Issue
+// #486.
+func cellRuntimeNeedsRematerialize(cell intmodel.Cell) bool {
+	state := cell.Status.State
+	return state == intmodel.CellStateStopped || state == intmodel.CellStateFailed
 }
 
 // ReconcileResult represents the result of reconciling a resource.
--- a/internal/controller/apply/reconcile_test.go
+++ b/internal/controller/apply/reconcile_test.go
@@ -1,0 +1,255 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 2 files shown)


eminwux:
Duplicate of #505

PR #508, opened 2026-05-14
fix: flip cell to Failed with Reason/Message on CreateCell/Start* errors

### LINKED ISSUE #504: runner: failed CreateCell/StartCell/StartContainer leaves cell in derived Stopped state with empty Reason/Message (and blocks AutoDelete)
## Context

A cell whose provisioning fails partway through `provisionNewCell` (or whose `StartCell` / `StartContainer` fails) is not surfaced to the operator as a failed cell. It either:

- (a) Stays in derived `Stopped` state with no Reason/Message — when the failure happens **before** `StartCell` (i.e., inside `runner.CreateCell` → `provisionNewCell` → `createCellCgroup` / `createCellContainers`), so the `#407` Failed-state defer on `StartCell` never gets a chance to run; or
- (b) Transitions to `Failed`, but `Status.Reason` and `Status.Message` are left empty — because `markCellFailedAfterStartupFailure` writes only `Status.State`, even though it receives the `cause error` and logs it.

In either case, the operator has no on-cell breadcrumb explaining why the cell ended up the way it did, and case (a) additionally interacts with `Spec.AutoDelete=true`: the `ReadyObserved` latch (intentional safety gate, `refresh.go:616-635`) never closes, so the reconciler refuses to auto-delete the orphaned cell. `kuke g ce` shows a perfectly normal-looking `Stopped` cell.

## Reproduction (real-world case)

`kuke run --rm` against a profile whose root container image pull or container-create step fails inside `createCellContainers` (image registry hiccup, naming collision, etc.). The cell I caught this on:

```
$ ./kuke g ce kukeon-dev-root-865fc2 -o json | jq .status
{
  \"state\": \"Stopped\",
  \"cgroupPath\": \"\",
  \"network\": {},
  \"containers\": [
    { \"name\": \"root\", \"state\": \"Unknown\", ... },
    { \"name\": \"work\", \"state\": \"Stopped\", ... }
  ],
  \"createdAt\": \"2026-05-14T12:35:45Z\",
  \"updatedAt\": \"2026-05-14T12:36:04Z\",
  \"readyAt\": \"0001-01-01T00:00:00Z\"
}
```

Spec has `autoDelete: true`. Notice:

- `state: Stopped`, **not** `Failed`
- `cgroupPath: \"\"` even though `/sys/fs/cgroup/kukeon/default/default/default/kukeon-dev-root-865fc2/` **exists on host**
- `readyAt` is zero, `readyObserved` is missing (i.e., `false`)
- `reason` and `message` are absent
- Journal has shim entries for sibling cells (`kukeon-dev-root-59d8cc`, `kukeon-dev-root-6b98bc`) created seconds apart, but **no shim was ever created for `865fc2`** — meaning `createCellContainers` never reached `TaskCreate`/`TaskStart`.

The trail of evidence shows `provisionNewCell` performed its stub write at line 1082 (`State=\"\"`, `CgroupPath=\"\"`), `createCellCgroup` succeeded (cgroup is on host), then `createCellContainers` at line 1115 failed, returning before `markCellReady` (line 1120) and the final `UpdateCellMetadata` (line 1123). The reconciler later derives `Stopped` from \"no task in containerd\" — that's the `Stopped` the operator sees.

### Gap 1 — no Failed transition on the CreateCell path

`internal/controller/runner/start.go:148 markCellFailedAfterStartupFailure` is only wired into `StartCell` (line 197) and `StartContainer` (line 815) via deferred error handlers. There is no equivalent on `provisionNewCell` / `createCellCgroup` / `createCellContainers`. When any of those return an error, `runner.CreateCell` bubbles it up to `controller.CreateCell`, which wraps it in `errdefs.ErrCreateCell` and returns to the RPC client. The cell metadata file is left exactly as the stub write at `provision.go:1082` produced it.

### Gap 2 — `Status.Reason` / `Status.Message` are declared but never written

The fields exist (`pkg/api/model/v1beta1/cell.go:94-95`):

\`\`\`go
Reason  string \`json:\"reason,omitempty\"  yaml:\"reason,omitempty\"\`
Message string \`json:\"message,omitempty\" yaml:\"message,omitempty\"\`
\`\`\`

The refresh carry helpers preserve them (`refresh.go:116-117,193-194,241-242,313-314 — next.Reason = orig.Reason; next.Message = orig.Message`), but nothing in the codebase ever **writes** them:

\`\`\`
$ grep -rn -E '(\.Reason|\.Message)\s*=' internal pkg --include='*.go' | grep -v _test
# only the four carry-forward lines — no writers anywhere
\`\`\`

Even `markCellFailedAfterStartupFailure`, which receives the failure `cause error` and logs it (start.go:153,168,173), only sets `Status.State = CellStateFailed`. It never copies the cause onto `Reason` / `Message`.

### Interaction with `AutoDelete=true`

Once the cell ends up in `Stopped` (case a above) without `ReadyObserved=true`, `shouldAutoDeleteCell` (`refresh.go:641`) blocks cleanup:

\`\`\`go
func shouldAutoDeleteCell(autoDelete bool, newState intmodel.CellState, readyObserved bool) bool {
    return autoDelete && cellStateAutoDeleteTriggers(newState) && readyObserved
}
\`\`\`

`ReadyObserved` is the latch designed to prevent reaping an *in-flight* `CreateCell` before its root container has come up (`latchReadyObserved`, refresh.go:616-635). It correctly stays open for this cell — but since the cell will never come up, the latch never closes, and the cell sits forever as an orphan with a host cgroup and an on-disk stub. The operator has to `kuke delete cell` it manually with no on-cell hint as to why.

## Expected behavior

A cell that fails during `CreateCell` / `StartCell` / `StartContainer` should:

1. Have `Status.State == Failed` persisted, regardless of *which* stage failed.
2. Have `Status.Reason` populated with a short, machine-readable label (e.g., `CreateCellFailed`, `StartCellFailed`, `StartContainerFailed`, or a more specific sub-reason like `CreateContainersFailed`).
3. Have `Status.Message` populated with the wrapped error string from `cause` (single-line, bounded length).

Then `kuke g ce -o yaml` shows the operator something actionable:

\`\`\`yaml
status:
  state: Failed
  reason: CreateCellContainersFailed
  message: \"pull registry.eminwux.com/busybox:latest: connection refused\"
\`\`\`

`Failed` is already sticky (`cellStateIsSticky`, refresh.go:595-597) and is already excluded from `cellStateAutoDeleteTriggers` (refresh.go:599-614) — so an explicit Failed transition on the CreateCell path also gives `Spec.AutoDelete=true` cells a defined terminal state that does **not** silently disappear, matching the intent of `#407`.

## Proposal

1. Add a `markCellFailedAfterCreateFailure(cell, cause)` helper (or generalize `markCellFailedAfterStartupFailure` so both call sites share a writer), and:
   - Wire it into a deferred error handler at the top of `provisionNewCell` (`provision.go:1080`), armed after the line-1082 stub write succeeds — mirroring the `provisionStarted` gate pattern in `StartCell`.
   - Best-effort `KillCell` for any container that did start.
   - Reload from disk, then set `Status.State = Failed`, `Status.Reason`, `Status.Message`, and persist via `UpdateCellMetadata`. Log a WARN on persist failure (don't mask the original cause), same posture as today's helper.

2. Update both Failed-state writers (`markCellFailedAfterStartupFailure` and the new CreateCell counterpart) so they populate `Status.Reason` and `Status.Message` from the `cause error`:
   - `Reason`: a stable label like `CreateCellFailed` / `StartCellFailed` / `StartContainerFailed`. Could also be derived from the wrapped error sentinel (`errors.As(cause, errdefs.ErrXxx)`) for finer-grained reasons.
   - `Message`: `cause.Error()`, truncated to a sane bound (e.g., 256 chars) to keep YAML readable.

3. Audit `kuke get cell` / `kuke describe cell` output paths so the new `reason` / `message` fields render in the default (table) and `-o yaml` / `-o json` views.

## Acceptance criteria

- [ ] A cell whose `createCellContainers` returns an error ends up with `Status.State == Failed`, non-empty `Status.Reason`, non-empty `Status.Message` on disk.
- [ ] A cell whose `StartCell` / `StartContainer` fails (issue #407 path) also gets `Status.Reason` / `Status.Message` populated — closing Gap 2 even on the path the helper already exists for.
- [ ] `kuke g ce <name> -o yaml` surfaces both `reason` and `message`.
- [ ] `Status.Reason` / `Status.Message` are preserved by the refresh-carry helpers (they already are — verify the new writers don't get clobbered on the next reconciler tick).
- [ ] An `AutoDelete=true` cell that fails during create is **not** silently auto-deleted: it stays as `Failed` with a Reason+Message until the operator runs `kuke delete cell`. This matches today's `#407` posture on `StartCell`-stage failures.
- [ ] Unit test in `internal/controller/runner/start_test.go` (or new `provision_test.go`): inject a `createCellContainers` failure, assert the persisted cell has `State=Failed`, `Reason` non-empty, `Message` containing the wrapped cause.

## Out of scope

- Auto-cleanup of the host cgroup left behind by the failed `createCellCgroup` step. The cgroup is already harmless (no tasks attached) and `kuke delete cell` reaps it; pairing cgroup-cleanup with the Failed transition is a separate concern.
- New CLI surface for clearing `Failed` cells in bulk (`kuke delete cell --state failed`, etc.). Today's `kuke delete cell` per-name flow is sufficient.
- Re-classifying `Failed` as auto-deletable when `AutoDelete=true`. The current exclusion in `cellStateAutoDeleteTriggers` is deliberate (#407 preserves Failed cells for diagnostics); operators opt in via `kuke delete cell`.

## References

- `internal/controller/runner/provision.go:1080-1128` — `provisionNewCell`, the unguarded CreateCell-stage path.
- `internal/controller/runner/start.go:133-175,197,815` — existing `markCellFailedAfterStartupFailure` helper and its two call sites.
- `internal/controller/runner/refresh.go:595-642` — `cellStateIsSticky`, `cellStateAutoDeleteTriggers`, `latchReadyObserved`, `shouldAutoDeleteCell` — the gates this issue is reasoning about.
- `pkg/api/model/v1beta1/cell.go:72-97` — `CellStatus` including the unused `Reason` / `Message` fields.
- Issue #407 — Failed-state semantics on `StartCell` / `StartContainer` (the prior art this issue extends to the CreateCell path).

eminwux:
## Summary

- A cell that fails during \`CreateCell\` (e.g. \`createCellContainers\` returns a registry/pull error) now ends up on disk as \`state: Failed\`, with \`status.reason\` = \`CreateCellFailed\` and \`status.message\` = the wrapped cause — instead of the derived \`Stopped\` with empty fields the operator was seeing per issue #504. \`StartCell\` and \`StartContainer\` failures (issue #407 path) likewise populate \`reason\` (\`StartCellFailed\` / \`StartContainerFailed\`) and \`message\`, closing Gap 2 in the same writer.
- The Failed transition on the CreateCell path is wired into \`provisionNewCell\` via an inner-closure pattern with a \`provisionStarted\` gate armed right after the line-1082 stub write — a pre-stub \`UpdateCellMetadata\` failure leaves no metadata file to stamp Failed onto, so the helper is skipped. Closure was used (instead of a named \`retErr\` + defer like StartCell / StartContainer) because \`nonamedreturns\` is enabled for new code via \`golangci-lint --new-from-rev HEAD\`; the existing two functions are grandfathered.
- The existing \`markCellFailedAfterStartupFailure\` helper was renamed to \`markCellFailed\` and now also writes \`Status.Reason\` (caller-supplied stable label) plus \`Status.Message\` via the new \`truncateFailureMessage\` helper: collapses control whitespace to spaces, joins runs, caps at 256 bytes with a \`...\` indicator, and trims on a valid UTF-8 boundary so registry refs with non-ASCII characters don't render as broken runes in YAML.
- Because \`Failed\` is already sticky (\`cellStateIsSticky\`) and excluded from \`cellStateAutoDeleteTriggers\`, an \`autoDelete: true\` cell that fails during create now stays as \`Failed\` until the operator runs \`kuke delete cell\` — AC #5. The existing \`carryCellLifecycle\` already copies Reason/Message through the refresh path, so the new writes survive subsequent reconciler ticks (AC #4); a regression test asserts the carry remains in place.

## Test plan

- [x] \`go test ./...\` (full unit suite green, including the four new tests in \`internal/controller/runner/start_test.go\`: \`TestTruncateFailureMessage\`, \`TestMarkCellFailed_PersistsReasonAndMessage\`, \`TestMarkCellFailed_LongCauseIsTruncated\`, \`TestMarkCellFailed_PreservesFieldsAcrossRefreshCarry\`)
- [x] \`make test\` green
- [x] \`go vet ./...\` clean
- [x] \`pre-commit run --files internal/controller/runner/{provision,start,start_test}.go\` clean (go fmt, golangci-lint, addlicense, etc.)
- [x] \`make dev-init\` smoke test passes, daemon-parity check matches the regression guard for both \`kuke get realms\` and \`kuke get realms --no-daemon\`
- [x] Manual end-to-end repro of AC #1 / #2 / #3 — \`kuke run -f /tmp/bad-cell.yaml -d\` against a profile pointing at \`registry.invalid/does-not-exist:latest\` produces:

  \`\`\`
  status:
    state: Failed
    reason: CreateCellFailed
    message: 'failed to create container default_default_failtest_worker: failed to create container from spec: failed to pull image registry.invalid/does-not-exist:latest: failed to resolve reference ... He...'
  \`\`\`

  Truncated at 256 bytes with the \`...\` indicator, single-line.
- [x] Manual AC #5 verification — same repro with \`autoDelete: true\` on the spec: the cell stays as \`Failed\` after 3s (no auto-delete fired), purged manually with \`kuke purge cell --cascade\`.

Closes #504

### DIFF
--- a/internal/controller/runner/provision.go
+++ b/internal/controller/runner/provision.go
@@ -1078,53 +1078,74 @@ func cellSpaceBridgeName(cell intmodel.Cell) (string, error) {
 }
 
 func (r *Exec) provisionNewCell(cell intmodel.Cell) (intmodel.Cell, error) {
-	// Update cell metadata
-	if err := r.UpdateCellMetadata(cell); err != nil {
-		return intmodel.Cell{}, fmt.Errorf("%w: %w", errdefs.ErrUpdateCellMetadata, err)
+	// The inner closure owns the actual provisioning work. The outer
+	// function captures provisionStarted via closure and, on any error
+	// past the stub write, flips the on-disk record to Failed with a
+	// Reason and Message so `kuke get cell -o yaml` shows the operator
+	// the failing stage instead of a derived Stopped (issue #504).
+	// A pre-stub UpdateCellMetadata failure leaves no metadata file to
+	// stamp Failed onto, so provisionStarted stays false and the helper
+	// is skipped. Closure pattern (instead of a named retErr + defer
+	// like StartCell / StartContainer) keeps the new code lint-clean
+	// against nonamedreturns.
+	var provisionStarted bool
+	out, err := func() (intmodel.Cell, error) {
+		// Update cell metadata
+		if err := r.UpdateCellMetadata(cell); err != nil {
+			return intmodel.Cell{}, fmt.Errorf("%w: %w", errdefs.ErrUpdateCellMetadata, err)
+		}
+
+		// Past the stub write: any subsequent error must flip the
+		// on-disk record to Failed.
+		provisionStarted = true
+
+		// Create cell cgroup
+		cgroupPath, subtreeControllers, err := r.createCellCgroup(cell)
+		if err != nil {
+			return intmodel.Cell{}, err
+		}
+
+		// Update internal model with cgroup path and effective subtree
+		// controllers (issue #328 — surfaces the result of
+		// enableCellControllers so operators can confirm the cell-level
+		// delegation from `kuke get cells -o yaml`).
+		cell.Status.CgroupPath = cgroupPath
+		cell.Status.SubtreeControllers = subtreeControllers
+
+		// Persist the host-side bridge this cell's space lands on.
+		// Computing it here (instead of the read path) keeps the cell
+		// self-describing for debug tooling and survives daemon
+		// restarts because the cell metadata file is rewritten in
+		// UpdateCellMetadata below.
+		if bridge, brErr := cellSpaceBridgeName(cell); brErr == nil {
+			cell.Status.Network.BridgeName = bridge
+		} else {
+			// Bridge name is best-effort metadata; failing to derive
+			// it must not block provisioning (the cell can still come
+			// up — the iface is created by CNI from the conflist
+			// regardless of what we record here).
+			r.logger.WarnContext(r.ctx, "could not derive cell bridge name",
+				"cell", cell.Metadata.Name, "error", brErr)
+		}
+
+		// Create pause container and all containers for the cell
+		if _, err = r.createCellContainers(&cell); err != nil {
+			return intmodel.Cell{}, err
+		}
+
+		markCellReady(&cell)
+
+		// Update cell metadata
+		if err = r.UpdateCellMetadata(cell); err != nil {
+			return intmodel.Cell{}, fmt.Errorf("%w: %w", errdefs.ErrUpdateCellMetadata, err)
+		}
+
+		return cell, nil
+	}()
+	if err != nil && provisionStarted {
+		r.markCellFailed(cell, "CreateCellFailed", err)
 	}
-
-	// Create cell cgroup
-	cgroupPath, subtreeControllers, err := r.createCellCgroup(cell)
-	if err != nil {
-		return intmodel.Cell{}, err
-	}
-
-	// Update internal model with cgroup path and effective subtree
-	// controllers (issue #328 — surfaces the result of enableCellControllers
-	// so operators can confirm the cell-level delegation from `kuke get
-	// cells -o yaml`).
-	cell.Status.CgroupPath = cgroupPath
-	cell.Status.SubtreeControllers = subtreeControllers
-
-	// Persist the host-side bridge this cell's space lands on. Computing it
-	// here (instead of the read path) keeps the cell self-describing for
-	// debug tooling and survives daemon restarts because the cell metadata
-	// file is rewritten in UpdateCellMetadata below.
-	if bridge, brErr := cellSpaceBridgeName(cell); brErr == nil {
-		cell.Status.Network.BridgeName = bridge
-	} else {
-		// Bridge name is best-effort metadata; failing to derive it must
-		// not block provisioning (the cell can still come up — the iface
-		// is created by CNI from the conflist regardless of what we
-		// record here).
-		r.logger.WarnContext(r.ctx, "could not derive cell bridge name",
-			"cell", cell.Metadata.Name, "error", brErr)
-	}
-
-	// Create pause container and all containers for the cell
-	_, err = r.createCellContainers(&cell)
-	if err != nil {
-		return intmodel.Cell{}, err
-	}
-
-	markCellReady(&cell)
-
-	// Update cell metadata
-	if err = r.UpdateCellMetadata(cell); err != nil {
-		return intmodel.Cell{}, fmt.Errorf("%w: %w", errdefs.ErrUpdateCellMetadata, err)
-	}
-
-	return cell, nil
+	return out, err
 }
 
 func (r *Exec) createCellCgroup(cell intmodel.Cell) (string, []string, error) {
--- a/internal/controller/runner/start.go
+++ b/internal/controller/runner/start.go
@@ -130,22 +130,59 @@ func cellTasksAllRunningFn(
 	return true
 }
 
-// markCellFailedAfterStartupFailure transitions the cell to the terminal
-// CellStateFailed state (issue #407) when a non-validation, containerd-touching
-// step in StartCell / StartContainer returns an error. The cleanup itself is:
+// maxFailureMessageLen caps Status.Message so a long, multi-line wrap from
+// `fmt.Errorf("%w: %w", …)` chains doesn't blow up `kuke get cell -o yaml`.
+// 256 bytes is comfortably wider than any error sentinel string in the repo
+// while still rendering as a single readable line.
+const maxFailureMessageLen = 256
+
+// truncateFailureMessage flattens cause's text into the single-line, bounded
+// form Status.Message expects: newlines/tabs collapsed to spaces, trimmed of
+// runs of whitespace, and capped at maxFailureMessageLen with a trailing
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #508 Review — `fix: flip cell to Failed with Reason/Message on CreateCell/Start* errors`**

LGTM. Closes the issue #504 gap cleanly: CreateCell failures now persist `state: Failed` with a stable `Reason` label and a single-line, 256-byte-capped `Message`, parallel to the StartCell / StartContainer paths from issue #407. The inner-closure pattern in `provisionNewCell` is well-justified in the comment (lint constraint), the `provisionStarted` gate correctly skips the helper on pre-stub UpdateCellMetadata failures, and the four new tests in `start_test.go` cover the persist, truncation, and `carryCellLifecycle` regression paths. UTF-8-aware truncation in `truncateFailureMessage` (`internal/controller/runner/start.go:159`) is a nice touch for registry refs. No correctness, leak, dead-code, or convention issues; commits all signed; CI all green; `make dev-init` listed checked in the test plan.

PR #514, opened 2026-05-15
refactor: drop dead errors.As branch in cgroupcheck.readControllers

### LINKED ISSUE #326: cgroupcheck: drop dead errors.As branch in readControllers (or implement the actionable error)
Surfaced during review of #325.

`internal/cgroupcheck/cgroupcheck.go:343-347`:

```go
data, err := os.ReadFile(path)
if err != nil {
    // Return a more actionable error for the common "kuke doctor
    // cgroups" misuse where the operator pointed --root at something
    // that isn't a cgroup-v2 hierarchy.
    var pathErr *fs.PathError
    if errors.As(err, &pathErr) {
        return nil, err
    }
    return nil, err
}
```

Both arms return the same `nil, err` — the `errors.As` check has no effect, and the comment promises a "more actionable error" that isn't synthesized. Either wrap the `*fs.PathError` with the cgroup-v2-hierarchy hint the comment describes (so `--root` pointing at a non-cgroup directory yields a structured remediation), or drop the dead branch + comment.

eminwux:
## Summary
- Both arms of the `errors.As(err, &pathErr)` check at `internal/cgroupcheck/cgroupcheck.go:489` returned `nil, err` identically — the branch had no runtime effect, and the comment promised a "more actionable error" that was never synthesized. Dropped the dead branch + misleading comment (lower-blast-radius of the two resolutions the issue offered).
- `io/fs` was only used by that branch — dropped the import. `errors` stays (still used by the `errors.Is` calls at lines 300, 305, 307 in the probe-error classifier).
- Behaviorally a no-op: `readControllers` still returns `nil, err` on `os.ReadFile` failure, and the two `Check()` call sites at lines 248 and 252 still wrap with `fmt.Errorf("read %s: %w", path, err)` so the path appears in the operator-facing error.

## Test plan
- [x] `go build ./...`
- [x] `make test` (full package suite — `internal/cgroupcheck` passes; no regressions elsewhere)
- [x] `pre-commit run --files internal/cgroupcheck/cgroupcheck.go` (go fmt / golangci-lint / addlicense clean)
- [ ] `make dev-init` — not run; this PR is a pure dead-code removal in `internal/cgroupcheck` (the daemon read path uses only `cgroupcheck.CellResourceControllers()`, a constant slice — `readControllers`/`Check()` are reached exclusively via `cmd/kuke/doctor/cgroups`). The CLAUDE.md smoke-test trigger ("anything the daemon reads") does not apply.

Closes #326

### DIFF
--- a/internal/cgroupcheck/cgroupcheck.go
+++ b/internal/cgroupcheck/cgroupcheck.go
@@ -32,7 +32,6 @@ package cgroupcheck
 import (
 	"errors"
 	"fmt"
-	"io/fs"
 	"os"
 	"path/filepath"
 	"sort"
@@ -489,13 +488,6 @@ func HostAdvertised(hostRoot string) ([]string, error) {
 func readControllers(path string) ([]string, error) {
 	data, err := os.ReadFile(path)
 	if err != nil {
-		// Return a more actionable error for the common "kuke doctor
-		// cgroups" misuse where the operator pointed --root at something
-		// that isn't a cgroup-v2 hierarchy.
-		var pathErr *fs.PathError
-		if errors.As(err, &pathErr) {
-			return nil, err
-		}
 		return nil, err
 	}
 	out := strings.Fields(string(data))


eminwux:
**PR #514 Review — `refactor: drop dead errors.As branch in cgroupcheck.readControllers`**

Clean dead-code removal closing #326. Both arms of the `errors.As` branch returned `nil, err` identically, the comment promised an actionable error that was never synthesized, and `io/fs` was only used by that branch. Behavior is preserved — all three `readControllers` callers (`Check` at `cgroupcheck.go:247,251`, `HostAdvertised` at `cgroupcheck.go:478`) wrap with `fmt.Errorf("read %s: %w", path, err)`, so the operator-facing error still names the path. Daemon read paths use only the `CellResourceControllers()` constant slice (confirmed in `provision.go`, `ctr/cgroups.go`, `modelhub/cell.go`); `readControllers` is reached only via `cmd/kuke/doctor/cgroups`, so the `make dev-init` carve-out is correct.

No leaks, no convention drift, signature verified.

LGTM.

PR #518, opened 2026-05-15
fix: pre-stamp Failed before KillCell to close markCellFailed --rm race

### LINKED ISSUE #409: runner: close --rm race between KillCell's Stopped-write and Failed-stamp in markCellFailedAfterStartupFailure
Surfaced during review of #408.

`(*Exec).markCellFailedAfterStartupFailure` (`internal/controller/runner/start.go:148`) currently calls `KillCell` first (which writes `CellStateStopped` via its own `UpdateCellMetadata`), then reloads, then stamps `CellStateFailed` and persists again. Between those two writes there is a brief window where the cell on disk is `Stopped` with `ReadyObserved=true` (for the `StartContainer` failure path on a previously-Ready cell).

If the daemon's reconcile loop ticks during that window on an `AutoDelete=true` cell, `ReconcileCell` will:
- see `originalStatus.State == Stopped` (not sticky, falls through the new `cellStateIsSticky` branch),
- re-derive `newState == Stopped` (KillCell removed the containers),
- `shouldAutoDeleteCell` returns true (AutoDelete && Stopped && ReadyObserved),
- call `autoDeleteCell`, which kills + `DeleteCell`'s the cell before the original defer's final `UpdateCellMetadata` ever lands.

The user loses the diagnostic surface the PR explicitly promises:

> Combined with `cellStateAutoDeleteTriggers` losing its `Failed` entry, `kuke run --rm` will no longer auto-clean a Failed cell — that's a deliberate departure from `docker run --rm` so the diagnostic surface survives until the operator runs `kuke delete cell`.

The contract holds outside the race window; inside it, the cell vanishes.

**Suggested fix** (`internal/controller/runner/start.go:148`): stamp `Failed` *before* calling `KillCell` (so the persisted state is sticky from the start), then call `KillCell`, then re-stamp `Failed` after the reload to overwrite KillCell's `Stopped` write. The `cellStateIsSticky` branch in `ReconcileCell` will then short-circuit any reconcile tick that lands in either window.

Window is narrow (two back-to-back `UpdateCellMetadata` calls); only affects `StartContainer`-failure on `--rm` cells that already reached Ready. The `StartCell` path is unaffected (cell never reached Ready, `ReadyObserved=false`, autoDelete trigger blocked).

eminwux:
## Summary
- `markCellFailed` now reloads + writes `State=Failed` *before* calling `KillCell`, then re-stamps `Failed` after the post-kill reload. The pre-kill write makes the on-disk state sticky (`cellStateIsSticky(Failed)`) for the duration of `KillCell`, so a reconcile tick that lands on an `AutoDelete=true` cell that previously reached Ready no longer trips `shouldAutoDeleteCell` and reaps the cell before the operator-facing `Failed` breadcrumb persists. Issue #409.
- The race window does not fully vanish — `KillCell`'s own `PopulateAndPersistCellContainerStatuses` writes `CellStateStopped` at the end of its run, so the window from that write to the helper's post-kill restamp remains. The fix matches the issue body's suggested approach: narrow the window dramatically by making the disk state sticky from the start, with the post-kill restamp closing the residual gap. Fully eliminating the window would require a deeper refactor of `KillCell`'s state-write contract and was out of scope here.

## Test plan
- [x] `make test` — passes (the `internal/controller/runner` package picks up the new `TestMarkCellFailed_PreStampsBeforeKillCell` regression test, which counts `UpdateCellMetadata` calls via the `nowFn` clock probe; verified the test fails on the pre-fix code by stashing the `start.go` change and re-running).
- [x] `pre-commit run --files internal/controller/runner/start.go internal/controller/runner/start_test.go` — passed.
- [x] `make dev-init` smoke — daemon-parity check (`kuke get realms` with and without `--no-daemon`) renders the expected two-realm Ready table; attach smoke against a kuketty-wrapped cell exits cleanly.

Closes #409

### DIFF
--- a/internal/controller/runner/start.go
+++ b/internal/controller/runner/start.go
@@ -165,13 +165,20 @@ func truncateFailureMessage(cause error) string {
 // (issue #407 for the Start*-stage paths; issue #504 for the CreateCell-stage
 // path). The cleanup itself is:
 //
+//   - reload from metadata and stamp State=Failed plus Status.Reason /
+//     Status.Message *before* KillCell runs, so the on-disk state is sticky
+//     from the start of the cleanup (issue #409). cellStateIsSticky(Failed)
+//     short-circuits ReconcileCell, so any reconcile tick that lands during
+//     KillCell — including the one tripped by `--rm`'s AutoDelete + a cell
+//     that already reached Ready — observes Failed and refuses to reap the
+//     cell.
 //   - best-effort KillCell on the cell so any container that did start (e.g.,
 //     the root container) is torn down — the cell holds no running processes
-//     once it enters Failed;
-//   - reload from metadata so we overwrite KillCell's own Stopped-write, then
-//     stamp State=Failed plus Status.Reason / Status.Message so `kuke get
-//     cell -o yaml` shows the operator the failing stage and the wrapped
-//     cause, and persist via UpdateCellMetadata.
+//     once it enters Failed. KillCell internally writes CellStateStopped at
+//     the end of its run, briefly re-exposing the pre-#409 race window;
+//   - reload again and re-stamp State=Failed + Reason + Message so the
+//     operator-facing breadcrumb survives KillCell's Stopped-write. `kuke get
+//     cell -o yaml` then shows the failing stage and the wrapped cause.
 //
 // The original startup error is returned unmodified by the caller; this helper
 // only writes the new state so the reconciler can later observe it as terminal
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #518 Review — `fix: pre-stamp Failed before KillCell to close markCellFailed --rm race`**

The pre-kill `Failed` stamp closes the dominant slice of the issue #409 window cleanly. `cellStateIsSticky(CellStateFailed)` (`internal/controller/runner/refresh.go:606-609`) short-circuits `ReconcileCell` at line 482, and `cellStateAutoDeleteTriggers` (line 624-627) excludes `Failed`, so a reconcile tick landing between the pre-stamp and `KillCell`'s internal `Stopped`-write cannot trip `shouldAutoDeleteCell` on an `AutoDelete + ReadyObserved` cell. `KillCell` ignores the input `cell.Status.State` (it reloads via `GetCell`), so pre-mutating the in-memory cell is safe. The regression test counts `UpdateCellMetadata` via the `nowFn` probe and pins the 2-write contract; verified the call graph (`stampCellLifecycle` calls `nowFn` once per persist; `GetCell` does not). No correctness, secret-leak, dead-code, or convention issues found. The residual window from `KillCell`'s `Stopped`-write to the post-kill restamp is documented in the PR body as out of scope, which is the right call.

LGTM to merge.

PR #529, opened 2026-05-15
fix: route kuke attach through SUN_PATH-safe socket symlink

### LINKED ISSUE #521: bug: kuke attach fails with "connect: invalid argument" on long runPaths (SUN_PATH overflow)
## Context

`make e2e TestKuke_AttachDetach_KeepsTaskRunning` fails on `origin/main` (verified at 6bd245b) with:

```
Error: error attaching client: ping failed: dial unix /tmp/TestKuke_AttachDetach_KeepsTaskRunning2015614058/001/data/e-r-86/e-sp-52/e-st-4f/e-ce-c4/work/tty/socket: connect: invalid argument
```

Re-discovered while resuming the parked re-run task from #119 (whose original premise — `waitForSocket` timeout on the sbsh-era socket — is stale after the kuketty phases #413/#415/#417/#420 landed). The socket now appears as expected; the failure has moved from socket-staging to the host-side `connect(2)` call.

**Root cause** (verified). Linux's `sockaddr_un` `sun_path` limit is **108 bytes including the terminating NUL** (`UNIX_PATH_MAX`). When the path passed to `connect()` exceeds that limit, the kernel returns `EINVAL`, which `net.Dial("unix", path)` surfaces verbatim as `connect: invalid argument`. The host-side socket path the e2e test connects to is computed by `fs.ContainerSocketPath` (`internal/util/fs/metadata.go:138`) as:

```
<runPath>/data/<realm>/<space>/<stack>/<cell>/<container>/tty/socket
```

For the e2e scenario, `getRandomRunPath` (`e2e/e2e_test.go:119`) backs `runPath` with `t.TempDir()`, which Go names after the test function: `/tmp/<TestName><rand>/NNN`. `TestKuke_AttachDetach_KeepsTaskRunning` is the longest test name in the e2e suite at 39 characters, and the combined path comes out to **109 bytes** — one over the limit:

```
$ echo -n "/tmp/TestKuke_AttachDetach_KeepsTaskRunning2015614058/001/data/e-r-86/e-sp-52/e-st-4f/e-ce-c4/work/tty/socket" | wc -c
109
```

Inside the container the path is `/run/kukeon/tty/socket` (well under 108) so `kuketty`'s `Listen` works; the same inode is reachable on the host *only* via the long bind-mount source path, and that's where the SUN_PATH overflow bites at `kuke attach` time.

This is a regression hazard beyond the e2e suite: any production runPath whose layout pushes `<runPath>/data/.../tty/socket` over 108 bytes (e.g. operators who place `--run-path` deeper than `/opt/kukeon`, or whose realm/space/stack/cell IDs are longer than the test fixture's `e-r-XX`) will hit the same failure under `kuke attach` with no clear diagnostic.

## Repro

On `origin/main` (6bd245b) with docker + containerd running and `make kuke && make kuketty` already done:

```
$ docker build --build-arg VERSION=v0.0.0-e2e -t kukeon-local:e2e .
$ HOME=$HOME E2E_BIN_DIR=$PWD KUKEON_E2E_IMAGE=docker.io/library/kukeon-local:e2e \
  KUKEON_E2E_IMAGE_DOCKER_NAME=kukeon-local:e2e \
  go test -v -run '^TestKuke_AttachDetach_KeepsTaskRunning$' ./e2e
...
    e2e_kuke_attach_test.go:139: kuke attach exited with code 1 (want 0)
        output:
        Error: error attaching client: ping failed: dial unix /tmp/TestKuke_AttachDetach_KeepsTaskRunning.../001/data/e-r-XX/e-sp-XX/e-st-XX/e-ce-XX/work/tty/socket: connect: invalid argument
--- FAIL: TestKuke_AttachDetach_KeepsTaskRunning (13.79s)
```

`waitForSocket` (`e2e/e2e_kuke_attach_test.go:215`) passes — the per-container socket inode is created — so the failure is purely at host-side `connect()`.

## Proposal

The clean fix is to make the host-side socket path SUN_PATH-safe by construction, instead of letting it inherit the full metadata layout depth. Two options:

1. **Host-side stable shortname.** Maintain a per-container short symlink (e.g. `<runPath>/s/<short-id>` where `<short-id>` is the existing 8-char cellID) that resolves to the same socket inode. `kuke attach` connects via the short symlink; the deep metadata path stays as-is for everything else. The symlink would be staged at the same point the daemon currently materializes the tty directory.
2. **Shorter base layout.** Drop the `/data/<realm>/<space>/<stack>/<cell>/<container>/tty/` segment in favour of `<runPath>/sockets/<cellID>-<container>` for the socket only. The metadata-dir layout stays for files that aren't size-constrained; only `ContainerSocketPath` moves.

Option 1 is the lower-blast-radius call — only `ContainerSocketPath`'s callers on the connect side change, and the symlink target stays the bind-mount source so `kuketty` inside the container is untouched. Option 2 ripples more (every reader of the host-side socket path moves), but it eliminates the indirection.

Either way, the daemon should also bound-check the resolved host-side socket path at provision time and refuse with a clear error if it would exceed 107 bytes (108 minus NUL), so future deepening of the layout fails loud rather than at `kuke attach`.

Finally, separately scope the e2e test: `getRandomRunPath` should fall back to a shorter prefix (e.g. `os.MkdirTemp("/tmp", "ke-")`) when `t.TempDir()` would produce a SUN_PATH-overflowing socket path. This keeps the suite green even before the architectural fix lands, and the harness check protects against test names accidentally drifting longer.

## Acceptance criteria

- [ ] Host-side path returned by `fs.ContainerSocketPath` (or its replacement) for any well-formed `<runPath, realm, space, stack, cell, container>` tuple is `<= 107` bytes for all `runPath` under 60 bytes — verified by a unit test in `internal/util/fs/`.
- [ ] Daemon refuses container provisioning with a clear `socket path %q exceeds SUN_PATH` error when the resolved host-side socket path would overflow, instead of deferring the failure to `kuke attach`.
- [ ] `e2e/e2e_test.go:getRandomRunPath` produces a runPath short enough that `TestKuke_AttachDetach_KeepsTaskRunning` (the longest test name in the suite) is SUN_PATH-safe.
- [ ] `make e2e TestKuke_AttachDetach_KeepsTaskRunning` passes end-to-end on `origin/main` after the fix: socket appears, `kuke attach` connects, Ctrl+] Ctrl+] sequence is consumed by the in-process filter, `kuke attach` exits 0, and the underlying container task remains RUNNING.

## Out of scope

- Lifting the workload `tty/` layout itself — the deep layout is fine for everything that isn't a unix socket; only the socket inode needs a SUN_PATH-safe handle.
- Renaming or merging `internal/util/fs/metadata.go`'s helper set; the constraint is on the resolved socket path, not the structure of the metadata directory.
- Backporting the bound check to merged daemon snapshots that pre-date this issue; the dev-init smoke catches stale-daemon problems on the operator's host.

## References

- #119 — parked re-run of `TestKuke_AttachDetach_KeepsTaskRunning`, closed-as-outdated when this issue lands (per the owner's instruction in https://github.com/eminwux/kukeon/issues/119#issuecomment).
- Kuketty injection-path phases: #413, #415, #417, #420.
- `internal/ctr/attachable.go:54` (`AttachableSocketPath` — the in-container peer of the host path).
- `internal/util/fs/metadata.go:138` (`ContainerSocketPath` — the host-side socket location, single source of truth).
- `e2e/e2e_test.go:119` (`getRandomRunPath` — produces the t.TempDir-rooted runPath that surfaces the overflow).
- Linux `man 7 unix` (`UNIX_PATH_MAX = 108`).

eminwux:
## Summary
- Linux's `sockaddr_un.sun_path` caps at 108 bytes (incl. NUL); the deep host-side socket path (`<RunPath>/data/<realm>/<space>/<stack>/<cell>/<container>/tty/socket`) overflowed on the longest e2e test name, surfacing as `connect: invalid argument` from `kuke attach`.
- The daemon now stages a per-RunPath SUN_PATH-safe symlink at `<RunPath>/s/<sha256[:8]>` pointing at the deep socket inode; `AttachContainer` returns that handle so the kernel-level `connect(2)` never sees a sun_path longer than `consts.KukeonMaxSocketPath = 107` (single-letter parent dir is intentional — spends the budget on the short id, not the directory).
- Provision-time fail-fast: `ensureAttachableSocketSymlink` refuses to stage when the resolved symlink path would still overflow, returning the new `errdefs.ErrSocketPathTooLong` with the offending path named — so a future operator-configured RunPath that overflows fails loud at provision time instead of deferring to first `kuke attach`.
- Cleanup wired in `DeleteCell` and `DeleteContainer` so the symlinks under `<RunPath>/s/` don't accumulate as containers come and go (idempotent — missing dirent is fine).
- e2e: `getRandomRunPath` falls back to a short `os.MkdirTemp("/tmp", "ke-")` prefix when `t.TempDir()` would push the *deep* path over `SUN_PATH`. The architectural fix routes attach through the short symlink so this fallback is defense-in-depth: a future regression that re-introduces the deep dial fails loudly here, not in CI.
- Unit tests in `internal/util/fs/metadata_test.go` pin the layout (`<RunPath>/s/<short>`), determinism (sha256-based ID round-trips), distinct-per-container property (incl. the pipe-separator concat-collision pair), and the SUN_PATH budget for a 60-byte RunPath with 64-byte realm/space/stack/cell/container names.

## Test plan
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `make test` — full unit suite excluding e2e passes (`pkg/api/...`, `internal/...`, `cmd/...`)
- [x] `pre-commit run --files $(git diff --name-only origin/main..HEAD)` — addlicense, golangci-lint, go-fmt, go-mod-tidy, end-of-file, trailing-whitespace all pass
- [ ] `make e2e TestKuke_AttachDetach_KeepsTaskRunning` (AC #4) — **deferred**: this dev session is itself running inside a kuke-managed container (visible mounts: `/run/kukeon/tty`, `/.kukeon/bin/kuketty`, `/.kukeon/kuketty/metadata.json`). The prior session that attempted to drive e2e from this same environment broke the outer host with connectivity loss and disk IO churn — diagnosis below. Per explicit user direction this session, the e2e + dev-init smoke runs need to happen on a clean host outside the kuketty wrapper. The non-running docker + containerd daemons are NOT the toolchain — both `which docker` and `which containerd` resolve — they are intentionally not being brought up here.
- [ ] `make dev-init` daemon-parity smoke (project CLAUDE.md §Local smoke test) — same defer + reason as `make e2e`. The diff touches `internal/controller/runner/` so the daemon does read this path; the smoke is required by project conventions but cannot safely run from this nested environment.

## Diagnosis of the prior host-break incident (per user request)
The prior dev tick that staged this same diff also attempted the full e2e suite from inside the kuke-managed dev container. Three compounding factors broke the host:

1. **Container-in-container Docker build.** `make test-e2e` first runs `docker build -t kukeon-local:e2e .`, which pulls a full base image and writes a fresh overlay tree to `/var/lib/docker` *inside* the dev container's overlay rootfs. The outer host's overlay-on-overlay shows up as sustained write IO.
2. **Nested containerd with no isolation budget.** The e2e harness brings up a host-side containerd at `/run/containerd/containerd.sock` and stages every cell/container/CNI bridge into it; per-test cleanup runs `t.Cleanup`, but a partial failure (which `TestKuke_AttachDetach_KeepsTaskRunning` did exhibit pre-fix due to this very SUN_PATH bug) leaves CNI bridges, veth pairs, and containerd namespaces dangling. CNI bridge accumulation is what the user observed as "connectivity issues" — each leaked bridge consumes a `/30` and an `iptables` chain in the outer host's view.
3. **No `kuke daemon reset` between attempts.** Re-running the suite without resetting `/opt/kukeon` re-uses any half-staged state from the prior attempt and bind-mounts fresh content over it, producing the high IO observed.

**Mitigation taken this session:** no docker/containerd daemons were brought up. Code review + unit suite + lint were sufficient to validate the fix; e2e is deferred via the `ci-stalled` label below. Future runs from a kuke-wrapped dev environment should either (a) run e2e from outside the wrapper, or (b) gate `make test-e2e` on a `KUKEON_E2E_HOST_OK` env so the wrapped agent fails fast instead of silently churning the outer host. The latter is a candidate `chore:` follow-up worth filing separately if the user agrees.

## Notes on coverage
- AC #1 (path ≤ 107 for any well-formed tuple under a 60-byte RunPath) — `TestContainerSocketSymlinkPath_FitsSUNPath` in `internal/util/fs/metadata_test.go`.
- AC #2 (daemon refuses provisioning with `ErrSocketPathTooLong` when the path would overflow) — `ensureAttachableSocketSymlink` in `internal/controller/runner/attachable.go`; error surfaced via `errdefs.ErrSocketPathTooLong`.
- AC #3 (`getRandomRunPath` produces a SUN_PATH-safe path for the longest test) — `deepSocketPathFits` fallback in `e2e/e2e_test.go`.
- AC #4 (`make e2e TestKuke_AttachDetach_KeepsTaskRunning` passes end-to-end) — deferred per the test plan above; needs verification on a clean host.

Closes #521

### DIFF
--- a/e2e/e2e_kuke_attach_test.go
+++ b/e2e/e2e_kuke_attach_test.go
@@ -104,7 +104,11 @@ func TestKuke_AttachDetach_KeepsTaskRunning(t *testing.T) {
 	// creates the metadata dir at provision time and sbsh binds the socket
 	// inside the container at task start. If this is missing, pkg/attach
 	// would fail in a way that's hard to attribute, so check explicitly.
-	socketPath := fs.ContainerSocketPath(runPath,
+	// Polls the SUN_PATH-safe symlink the runner stages (issue #521) — the
+	// same handle `kuke attach` connects through. `os.Stat` follows the
+	// symlink, so the loop succeeds once kuketty creates the deep socket
+	// inode at the resolved target.
+	socketPath := fs.ContainerSocketSymlinkPath(runPath,
 		realmName, spaceName, stackName, cellName, "work")
 	waitForSocket(t, socketPath, 10*time.Second)
 
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -109,16 +109,52 @@ func runBinary(t *testing.T, env []string, command string, args ...string) (int,
 	return exitCode, []byte(stdoutBuf.String()), []byte(stderrBuf.String())
 }
 
-// getRandomRunPath returns a unique, absolute run path for the test, backed
-// by t.TempDir(). The directory is created up front and removed automatically
-// when the test ends, so callers do not need a separate create-or-cleanup
-// step. Each call returns its own path, which avoids contention on a shared
-// parent directory (issue #491: MkdirAll on a shared parent races
-// concurrent t.Cleanup RemoveAll from sibling parallel tests, surfacing as
-// ENOENT on the new leaf).
+// getRandomRunPath returns a unique, absolute run path for the test. When
+// the t.TempDir-rooted path would push the *deep* per-container kuketty
+// socket path (`<runPath>/data/<realm>/<space>/<stack>/<cell>/<container>/
+// tty/socket`) over Linux's UNIX_PATH_MAX, the helper falls back to a
+// short /tmp prefix and registers a cleanup so the suite stays SUN_PATH-
+// safe regardless of how Go names the t.TempDir parent (issue #521).
+//
+// The architectural fix (issue #521) routes `kuke attach` through a short
+// symlink, so the deep path length no longer breaks `connect(2)` —
+// but the fallback is deliberate defense-in-depth: a future regression
+// that accidentally re-introduces the deep-path dial would fail loudly
+// here instead of in CI.
+//
+// Each call returns its own path so parallel-sibling tests don't race
+// MkdirAll/Cleanup-RemoveAll on a shared parent (issue #491's regression
+// class).
 func getRandomRunPath(t *testing.T) string {
 	t.Helper()
-	return t.TempDir()
+	tempDir := t.TempDir()
+	if deepSocketPathFits(tempDir) {
+		return tempDir
+	}
+	// t.TempDir() is the standard, but here it is exactly what overflows
+	// SUN_PATH because Go derives the temp basename from the test
+	// function name (`TestKuke_AttachDetach_KeepsTaskRunning` is 39
+	// bytes). Drop to a short, name-free /tmp prefix instead.
+	short, err := os.MkdirTemp("/tmp", "ke-") //nolint:usetesting // intentional shorter prefix; see comment
+	if err != nil {
+		t.Fatalf("MkdirTemp(/tmp, ke-): %v", err)
+	}
+	t.Cleanup(func() { _ = os.RemoveAll(short) })
+	return short
+}
+
+// deepSocketPathFits reports whether the longest plausible per-container
+// kuketty socket path under runPath fits SUN_PATH. The longest e2e suffix
+// is `/data/<realm>/<space>/<stack>/<cell>/<container>/tty/socket` with
+// realm/space/stack/cell names produced by generateUnique*Name (`e-r-XX`,
+// `e-sp-XX`, `e-st-XX`, `e-ce-XX`) and the container name fixed to
+// "work" by the attach fixture — 52 bytes of suffix all in. The check is
+// against the deep path, not the SUN_PATH-safe symlink the runner stages,
+// because deep-path regressions are the failure mode this fallback
+// defends against (the symlink budget is comfortably wider).
+func deepSocketPathFits(runPath string) bool {
+	const deepSocketSuffixLen = len("/data/e-r-XX/e-sp-XX/e-st-XX/e-ce-XX/work/tty/socket")
+	return len(runPath)+deepSocketSuffixLen <= consts.KukeonMaxSocketPath
 }
 
 // buildKukeRunPathArgs builds --run-path flag arguments.
--- a/internal/client/local/local.go
+++ b/internal/client/local/local.go
@@ -956,13 +956,20 @@ func controllerImageToWire(img controller.ImageInfo) kukeonv1.ImageInfo {
 // sbsh control-socket path. Bytes never traverse this RPC — the caller
 // (`kuke attach`) opens HostSocketPath directly and runs the sbsh client
 // loop against it.
+//
+// Returns the SUN_PATH-safe symlink path (fs.ContainerSocketSymlinkPath)
+// rather than the deep socket inode path so `connect(2)` never sees a
+// sun_path longer than 107 bytes regardless of how deep the resolved
+// metadata layout is (issue #521). The runner stages the symlink at
+// provision time so a re-derived path on a freshly-spawned `kuke attach`
+// is always pre-materialised.
 func (c *Client) AttachContainer(_ context.Context, doc v1beta1.ContainerDoc) (kukeonv1.AttachContainerResult, error) {
 	spec, err := c.resolveAttachable(doc)
 	if err != nil {
 		return kukeonv1.AttachContainerResult{}, err
 	}
 	return kukeonv1.AttachContainerResult{
-		HostSocketPath: fs.ContainerSocketPath(
+		HostSocketPath: fs.ContainerSocketSymlinkPath(
 			c.ctrl.RunPath(),
 			spec.RealmName, spec.SpaceName, spec.StackName, spec.CellName, spec.ID,
 		),
--- a/internal/consts/consts.go
+++ b/internal/consts/consts.go
@@ -49,6 +49,27 @@ const (
 	// injected by Attachable=true specs.
 	KukeonContainerSocketFile = "socket"
 
+	// KukeonSocketSymlinkSubdir is the basename of the per-RunPath directory
+	// that holds SUN_PATH-safe symlinks to the deep per-container kuketty
+	// sockets (`<RunPath>/data/<realm>/<space>/<stack>/<cell>/<container>/
+	// tty/socket`). Linux `connect(2)` caps the path passed in
+	// sockaddr_un.sun_path at 108 bytes including the terminator (issue #521),
+	// and the deep metadata-rooted path overflows on long RunPaths or
+	// long realm/space/stack/cell IDs. The daemon stages a short symlink at
+	// `<RunPath>/s/<short>` at provision time; `kuke attach` connects via
+	// the symlink so the literal sun_path stays SUN_PATH-safe regardless of
+	// how deep the resolved target lives. Single-letter basename to spend
+	// the budget on the per-container short id, not the directory.
+	KukeonSocketSymlinkSubdir = "s"
+
+	// KukeonMaxSocketPath is Linux's UNIX_PATH_MAX minus the terminating
+	// NUL — the maximum number of bytes a path passed to `connect(2)` on a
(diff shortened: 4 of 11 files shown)


eminwux:
**PR #529 Review — `fix: route kuke attach through SUN_PATH-safe socket symlink`**

Architecturally sound — short symlink at `<RunPath>/s/<sha256[:8]>` keeps `connect(2)` SUN_PATH-safe regardless of metadata depth, lifecycle is wired symmetrically in cell + container delete paths, the provision-time fail-fast is the right place to surface future-RunPath overflow, and the e2e fallback to `/tmp/ke-` is correct defense-in-depth. CI green; commit verified; no sensitive-data exposure. One blocking item:

- **Hash collision via pipe-in-name** `internal/util/fs/metadata.go:192-194`: `containerSocketShortID` concatenates the tuple with `|` as the separator, but `ValidateHierarchyName`/`ValidateRealmName` (`internal/util/naming/hierarchyname.go:51`, `realmname.go:43`) reject only `_` and `/` — `|` is a legal name character. So realm `a|b` + space `c` and realm `a` + space `b|c` (identical stack/cell/container) hash to the same 16-hex id and overwrite each other's symlink at provision time, silently routing `kuke attach` for one container to the other's socket. Fix: switch the separator to `_` or `/` (both already rejected by both validators) — the PR's existing `concat` row in `TestContainerSocketSymlinkPath_DistinctPerContainer` would still pass, and the failure case is currently uncovered.

A non-blocking nit:

- **Missing failure-path test** `internal/controller/runner/attachable.go:596-599`: AC #2 (refuse to provision with `ErrSocketPathTooLong`) is asserted by inspection only. A unit-level test that constructs a >107-byte resolved path would round out coverage.

The deferred `make e2e TestKuke_AttachDetach_KeepsTaskRunning` and `make dev-init` smokes are documented host-gated items per CLAUDE.md "When NOT to say LGTM" — non-blocking, but the user should run both pre-merge.

LGTM once the separator collision is closed — the rest is non-blocking.

eminwux:
Addressed in 24b3f88:
- Separator collision (blocking): switched `containerSocketShortID` from `|` to `_`. `_` is rejected by both `naming.ValidateRealmName` and `naming.ValidateHierarchyName`, so the join is unambiguous regardless of any byte values in the names. The reviewer-flagged pair `realm "a|b" + space "c"` vs `realm "a" + space "b|c"` no longer hashes to the same id. New regression row `"pipe-in-name"` in `TestContainerSocketSymlinkPath_DistinctPerContainer` pins this — verified it fails on HEAD~1 (old `|` separator) and passes on HEAD.

Deferred:
- Failure-path unit test for `ensureAttachableSocketSymlink` returning `ErrSocketPathTooLong` (non-blocking nit) — left for a focused follow-up. The current AC#2 attestation remains inspection-only.
- `make e2e TestKuke_AttachDetach_KeepsTaskRunning` and `make dev-init` daemon-parity smoke — still host-gated, deferred to a clean host per the PR body's test plan.

eminwux:
**PR #529 Re-review — fixes look good**

Commit `24b3f88` closes the blocker: separator switched to `_` (rejected by both `ValidateRealmName` and `ValidateHierarchyName`, so no name can contain it — the join is now unambiguous), and a `pipe-in-name` row added to `TestContainerSocketSymlinkPath_DistinctPerContainer` matching the exact collision case from the prior review. CI green; both commits verified.

LGTM — unconditional. Reminder for the user: `make e2e TestKuke_AttachDetach_KeepsTaskRunning` and `make dev-init` are still deferred per the test plan (documented host-gated) and should be run pre-merge.

Issue #532, opened 2026-05-15
bug: kuke doctor cgroups passes green on no-swap + no-userspace-OOM hosts (would have flagged the 2026-05-15 incident host)


eminwux:
## Repro

Filed alongside the unlimited-`memory.max` bug. The 2026-05-15 host-wedge incident on `kukeon-dev` (QEMU VM, 19 GiB RAM, no swap, no systemd-oomd, no earlyoom) would have been preventable if the doctor pre-flight had warned about this combination. Reproduction:

```bash
# On a host with no swap and no userspace OOM:
swapon --show           # empty
systemctl is-enabled systemd-oomd  # disabled
pgrep earlyoom          # nothing

sudo kuke doctor cgroups --probe
# → passes green; emits no warning about the host being one cgroup-unlimited
#   container away from a hard wedge.
```

## Observed

`grep -n "swap\|OOM\|earlyoom\|systemd-oomd" cmd/kuke/doctor/cgroups/cgroups.go` returns empty. The probe enforces cgroup v2 mode and the required-controller list (cpu, memory, io, pids, …), which is necessary but not sufficient — having the `memory` controller delegated does nothing if (a) no `memory.max` ever gets written (companion bug) and (b) the kernel has no swap to cushion overshoot and no userspace OOM (systemd-oomd / earlyoom) to kill the offending workload before the host degrades.

## Root cause

`cmd/kuke/doctor/cgroups/cgroups.go` is scoped to cgroup-controller availability; it has no notion of "is the host viable as a hard tenancy boundary?" That broader check is exactly the gap the incident reporter named: *"have kuke doctor cgroups --probe warn when host has no swap and no userspace OOM guard (systemd-oomd / earlyoom), because in that configuration any container without a memory cap can wedge the whole VM."*

## Suggested fix

Extend `cmd/kuke/doctor/cgroups/cgroups.go` (or split into a sibling `kuke doctor host` subcommand if cgroup-scope purity matters) to check, in order:

1. **Swap availability.** Read `/proc/swaps`; if it has only the header line, emit a high-severity warning naming the file and the missing line.
2. **Userspace OOM guard.** Probe for `systemd-oomd` (`systemctl is-active systemd-oomd`), `earlyoom` (`pgrep earlyoom`), or `oomd` from the cgroup-v2 freezer toolchain. If none are active, emit a high-severity warning.
[...]

PR #533, opened 2026-05-15
fix: add daemon-default memory cap fallback for cgroup-unlimited containers

### LINKED ISSUE #531: bug: kukeond admits containers with no memory.max → host wedges on cgroup-unlimited workloads
## Repro

Filed after a verified incident on `kukeon-dev` (QEMU VM, 16 vCPU, 19 GiB RAM, no swap) on 2026-05-15 around 22:47 CEST. Two pr-reviewer cells (`default_default_kukeon-pr-reviewer-21a0c9`, `default_default_crew-pr-reviewer-2f6fff`) ran on the `default.kukeon.io` realm without `ContainerSpec.Resources.MemoryLimitBytes` populated. While a debug session was using the still-running `kukeon-dev-root-2e6b1a` cell, `systemd-journald` logged:

```
22:47:28 systemd-journald: Under memory pressure, flushing caches.
```

…and then complete silence (no kernel OOM trace persisted, `/sys/fs/pstore/` empty, host required hard reset). A `nsenter ... curl https://github.com` from inside the cell netns observed at 22:43:38 took ~4 min to return — the page-cache thrash that precedes a full memory wedge.

To reproduce on a fresh host:

```bash
# 19 GiB no-swap host with no systemd-oomd and no earlyoom (matches incident host)
swapon --show     # empty
systemctl is-enabled systemd-oomd  # disabled

# Create any cell whose ContainerSpec has Resources == nil OR Resources.MemoryLimitBytes == nil
# Inside it, allocate progressively until the host runs out of RAM
cat /sys/fs/cgroup/<containerd-ns>/<containerd-id>/memory.max
# → "max" (no limit). The kernel cannot scope the OOM to the workload.
```

## Observed

`grep` across the tree confirms there is no daemon-wide default, no host-fraction fallback, and no cell-level inheritance for memory limits:

- `internal/ctr/spec.go:578-588` — OCI build step is fully opt-in:

  ```go
  if spec.Resources != nil {
      if spec.Resources.MemoryLimitBytes != nil && *spec.Resources.MemoryLimitBytes > 0 {
          opts = append(opts, oci.WithMemoryLimit(uint64(*spec.Resources.MemoryLimitBytes)))
      }
      ...
  }
  ```

  When `Resources == nil` or `MemoryLimitBytes == nil`, the cgroup's `memory.max` stays at `max`.
- `internal/modelhub/container.go:194-199` — `ContainerResources` is the only memory-limit knob; no `CellResources` defaults exist that would cascade down.
- `internal/modelhub/merge.go:93-111` — `cloneResources` is a deep-copy; no defaulting.
- `grep -rn "DefaultResources\|defaultMemoryLimit\|HostMemoryFraction"` returns empty.
- `kukeond` startup (`internal/daemon/`) does not stamp a daemon-wide cap.

Pr-reviewer-class workloads run Claude/Node/Go toolchains that routinely allocate multi-GB. On a no-swap host with no userspace OOM guard, an unlimited container can consume enough memory to evict journald — exactly what the silenced 22:47:44 journal line and the hard-reset requirement show.

## Root cause

The OCI plumbing is correct *when* `MemoryLimitBytes` is set, but the daemon never enforces a sane fallback. The hot path that admits any externally-orchestrated container (e.g. the crew/agents orchestrator) cannot trust upstream specs to always carry a memory cap — the gap turns "spec lacks limit" into "host can be killed by any tenant."

Not the same root cause as #312 (decorative cgroup tree, closed): #312 was about whose cgroup the process *lives in*; this is about whether `memory.max` *gets written at all*, independent of placement.

## Suggested fix

Either (or both, the two compose cleanly):

1. **Daemon-default cap.** Add a `kukeond serve --default-memory-limit-bytes <N>` flag (and a matching `KUKEON_DEFAULT_MEMORY_LIMIT_BYTES` env). When set, `BuildContainerSpec` in `internal/ctr/spec.go:578` applies it whenever `spec.Resources.MemoryLimitBytes` is nil. Operators in unmanaged environments (no orchestrator stamping limits) can set this once at `kuke init` time. Default empty = current behavior, so the change is opt-in but documented.

2. **Refuse unlimited cells when the host lacks a safety net.** When `kukeond` starts (or when a cell is created), check `/proc/swaps` and `systemctl is-enabled systemd-oomd` / `pgrep earlyoom`. If swap is absent AND no userspace OOM is active AND the incoming `ContainerSpec.Resources.MemoryLimitBytes` is nil, reject the container with a clear error naming this issue. Operators who genuinely want unlimited cells on a no-swap host pass an explicit override flag (`--allow-unlimited-memory`).

Option 1 is the lower-blast-radius landing — pure plumbing. Option 2 is the policy escalation that closes the host-wedge class entirely. I'd ship #1 first and gate #2 behind a follow-up.

A unit test in `internal/ctr/spec_security_test.go` already covers the populated path (`MemoryLimitBytes: ptrInt64(4 * 1024 * 1024 * 1024)`); the new tests pin the nil-default behavior end-to-end with the daemon-default flag.

## Out of scope

- Per-cell or per-stack resource budgets — useful, separately shippable.
- CPU and pids caps — same opt-in pattern in the same code block; once memory has a default, the analogous fixes are mechanical follow-ups.
- The page-cache-thrash → CNI-curl latency at 22:43:38 — almost certainly a downstream symptom of memory pressure, not a separate CNI bug; once memory limits land, the latency disappears with it.

## References

- #312 (closed) — different concern (cgroup-tree placement vs. limit-not-written).
- Companion issue covering the `kuke doctor cgroups` blind spot that would have warned about this host configuration before the incident.
- `internal/ctr/spec.go:578-588` — the missing-default site.
- `internal/modelhub/container.go:194-199`; `pkg/api/model/v1beta1/container.go:232` — the spec shape that needs a default companion field or daemon-wide fallback.

eminwux:
## Summary
- Adds an opt-in daemon-wide memory-cap fallback so `kukeond` writes `memory.max` on every admitted container whose `ContainerSpec.Resources.MemoryLimitBytes` is unset or zero. Closes the host-wedge gap the 2026-05-15 incident named: on no-swap, no-userspace-OOM hosts an unbounded workload can evict journald and require a hard reset.
- New surfaces: `kukeond serve --default-memory-limit-bytes <N>`, `KUKEON_DEFAULT_MEMORY_LIMIT_BYTES` env, and `ServerConfiguration.Spec.DefaultMemoryLimitBytes` (YAML). Precedence is the same as every other kukeond knob: `--flag > env > ServerConfiguration YAML > built-in default (0 = disabled)`.
- Explicit per-container `Resources.MemoryLimitBytes` always wins: the daemon-default never clobbers an operator-supplied cap. Zero / negative arguments are a no-op so the option can be passed unconditionally; serve.go clamps negatives with a warning before they reach `uint64`.
- Wiring path: `cmd/kukeond` flag/env binding → `daemon.Options.Controller` → `controller.Options.DefaultMemoryLimitBytes` → `runner.Options.DefaultMemoryLimitBytes` → `r.daemonDefaultBuildOpts()` → `ctr.WithDefaultMemoryLimit(...)` BuildOption. Applied at every `BuildContainerSpec` / `BuildRootContainerSpec` call site in the runner (start.go and provision.go for both root and non-root creation paths), so the cap reaches both workload containers and cell root containers when the daemon is configured with one.
- The default ServerConfiguration template (`internal/serverconfig/serverconfig.go`) now documents the new field at `0` so fresh hosts see it commented in `/etc/kukeon/kukeond.yaml` with the no-swap rationale.

## Behavior preserved
- Zero default (the new built-in default) = existing behavior bit-for-bit. No existing call site changes the OCI spec it produces when the flag is unset.
- `BuildRootContainerSpec` gains a variadic `...BuildOption` parameter; the previous two-argument form keeps working because Go variadics are backwards-compatible.
- `kuke ... --no-daemon` (CLI/local controller path) does not plumb the option, so non-daemon flows stay unchanged.

## Implementation calls worth flagging
- The AC's issue body proposed two fixes (daemon-default cap, plus refusal-when-host-unsafe) and explicitly said *"I'd ship #1 first and gate #2 behind a follow-up."* This PR is #1 only — the host-safety doctor probe is the companion issue (#532) and a separate landing.
- The same fallback applies to **root containers** as well as workload containers (mirrors BuildContainerSpec in BuildRootContainerSpec). Rationale: the root container ID/path is the one most likely to be tiny in practice, but the issue body's framing ("daemon never enforces a sane fallback ... any tenant") argues the cap should be universal once the operator sets it. Operators must therefore size the default high enough not to wedge the `kukeond` cell itself on reconcile-driven recreate. `kuke init`'s bootstrap path runs before any daemon-flag exists, so the initial kukeond-cell creation is unaffected.
- `gosec` G115 (int64→uint64) is suppressed with `//nolint:gosec` and the bounding rationale on the two call sites — matches the bounded-conversion pattern in `internal/cni/subnet.go:116-117`.

## Test plan
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `go test \$(go list ./... | grep -v /e2e) -count=1 -timeout 600s` — full unit suite passes (the `internal/ctr` package runs an existing containerd-dialing test that takes ~70s to fail open against the absent host containerd; same behavior on `origin/main`, not introduced by this change)
- [x] `pre-commit run --files \$(git diff --name-only origin/main..HEAD)` — addlicense, golangci-lint (gosec, govet, etc.), go-fmt, go-mod-tidy, end-of-file, trailing-whitespace all pass
- [ ] `make e2e` — **deferred** to the reviewer's environment: this dev session runs inside a kuke-managed container (`/run/kukeon/tty`, `/.kukeon/bin/kuketty`, `/.kukeon/kuketty/metadata.json` are visible mounts) and a prior session that drove `make e2e` from inside this same wrapper broke the outer host with overlay-on-overlay IO and leaked CNI bridges. Per the role playbook's "verification step can't safely run on the current host" section, the e2e check needs a clean host. The change adds no new e2e contract — it only writes `memory.max` from an additional source — so the existing e2e suite is the right vehicle once it runs on a clean host.
- [ ] `make dev-init` daemon-parity smoke (project CLAUDE.md §Local smoke test) — same defer + same reason. The diff touches `internal/controller/runner/`, so the daemon does read this path; reviewer should run the smoke on a clean host before merge.

## New tests
- `internal/ctr/spec_default_memory_test.go` covers all branches of the new logic for both `BuildContainerSpec` and `BuildRootContainerSpec`:
  - `Resources == nil` + default set → daemon default written.
  - `Resources != nil` with `MemoryLimitBytes == nil` → daemon default written.
  - `Resources.MemoryLimitBytes == ptrInt64(0)` → daemon default written (the existing `> 0` guard skipped this case; the fallback closes it).
  - Explicit positive `MemoryLimitBytes` → explicit value wins, default ignored.
  - No option, `WithDefaultMemoryLimit(0)`, `WithDefaultMemoryLimit(-1)` → no Memory.Limit emitted (current-behavior preservation).
- `cmd/kukeond/kukeond_test.go` extends the YAML-precedence tests for the new field and adds a `--default-memory-limit-bytes` flag-presence check plus an env-overrides-YAML precedence test.

Closes #531

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -121,6 +121,10 @@ var (
 	KUKEOND_RECONCILE_INTERVAL = DefineKV(
 		"KUKEOND_RECONCILE_INTERVAL", "kukeond/reconcileInterval", "30s",
 	)
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKEON_DEFAULT_MEMORY_LIMIT_BYTES = DefineKV(
+		"KUKEON_DEFAULT_MEMORY_LIMIT_BYTES", "kukeond/defaultMemoryLimitBytes", "0",
+	)
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_INIT_REALM = DefineKV("KUKE_INIT_REALM", "kuke/init/realm")
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -122,8 +122,8 @@ var (
 		"KUKEOND_RECONCILE_INTERVAL", "kukeond/reconcileInterval", "30s",
 	)
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
-	KUKEON_DEFAULT_MEMORY_LIMIT_BYTES = DefineKV(
-		"KUKEON_DEFAULT_MEMORY_LIMIT_BYTES", "kukeond/defaultMemoryLimitBytes", "0",
+	KUKEOND_DEFAULT_MEMORY_LIMIT_BYTES = DefineKV(
+		"KUKEOND_DEFAULT_MEMORY_LIMIT_BYTES", "kukeond/defaultMemoryLimitBytes", "0",
 	)
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
--- a/cmd/kukeond/kukeond.go
+++ b/cmd/kukeond/kukeond.go
@@ -165,6 +165,20 @@ func NewKukeondCmd() (*cobra.Command, error) {
 		return nil, err
 	}
 
+	cmd.PersistentFlags().Int64(
+		"default-memory-limit-bytes", 0,
+		"Daemon-wide fallback memory cap (bytes) for any admitted container "+
+			"whose Resources.MemoryLimitBytes is unset. 0 disables; an explicit "+
+			"per-container limit always wins. Recommended on hosts without swap "+
+			"and without a userspace OOM guard. (issue #531)",
+	)
+	if err := viper.BindPFlag(
+		config.KUKEON_DEFAULT_MEMORY_LIMIT_BYTES.ViperKey,
+		cmd.PersistentFlags().Lookup("default-memory-limit-bytes"),
+	); err != nil {
+		return nil, err
+	}
+
 	bindEnvVars()
 
 	cmd.AddCommand(newServeCmd())
@@ -186,6 +200,7 @@ func bindEnvVars() {
 		config.KUKEOND_SOCKET,
 		config.KUKEOND_SOCKET_GID,
 		config.KUKEOND_RECONCILE_INTERVAL,
+		config.KUKEON_DEFAULT_MEMORY_LIMIT_BYTES,
 	} {
 		_ = v.BindEnv()
 	}
@@ -229,6 +244,11 @@ func applyServerConfiguration(cmd *cobra.Command, spec v1beta1.ServerConfigurati
 		!flagChanged(cmd, "cgroup-root") &&
 		!envSet(config.KUKEON_ROOT_CGROUP_ROOT) {
 		viper.Set(config.KUKEON_ROOT_CGROUP_ROOT.ViperKey, spec.CgroupRoot)
+	}
+	if spec.DefaultMemoryLimitBytes > 0 &&
+		!flagChanged(cmd, "default-memory-limit-bytes") &&
+		!envSet(config.KUKEON_DEFAULT_MEMORY_LIMIT_BYTES) {
+		viper.Set(config.KUKEON_DEFAULT_MEMORY_LIMIT_BYTES.ViperKey, spec.DefaultMemoryLimitBytes)
 	}
 }
 
--- a/cmd/kukeond/kukeond.go
+++ b/cmd/kukeond/kukeond.go
@@ -173,7 +173,7 @@ func NewKukeondCmd() (*cobra.Command, error) {
 			"and without a userspace OOM guard. (issue #531)",
 	)
 	if err := viper.BindPFlag(
-		config.KUKEON_DEFAULT_MEMORY_LIMIT_BYTES.ViperKey,
+		config.KUKEOND_DEFAULT_MEMORY_LIMIT_BYTES.ViperKey,
 		cmd.PersistentFlags().Lookup("default-memory-limit-bytes"),
 	); err != nil {
 		return nil, err
@@ -200,7 +200,7 @@ func bindEnvVars() {
 		config.KUKEOND_SOCKET,
 		config.KUKEOND_SOCKET_GID,
 		config.KUKEOND_RECONCILE_INTERVAL,
-		config.KUKEON_DEFAULT_MEMORY_LIMIT_BYTES,
+		config.KUKEOND_DEFAULT_MEMORY_LIMIT_BYTES,
 	} {
 		_ = v.BindEnv()
 	}
@@ -247,8 +247,8 @@ func applyServerConfiguration(cmd *cobra.Command, spec v1beta1.ServerConfigurati
 	}
 	if spec.DefaultMemoryLimitBytes > 0 &&
 		!flagChanged(cmd, "default-memory-limit-bytes") &&
-		!envSet(config.KUKEON_DEFAULT_MEMORY_LIMIT_BYTES) {
-		viper.Set(config.KUKEON_DEFAULT_MEMORY_LIMIT_BYTES.ViperKey, spec.DefaultMemoryLimitBytes)
+		!envSet(config.KUKEOND_DEFAULT_MEMORY_LIMIT_BYTES) {
+		viper.Set(config.KUKEOND_DEFAULT_MEMORY_LIMIT_BYTES.ViperKey, spec.DefaultMemoryLimitBytes)
 	}
 }
 
--- a/cmd/kukeond/kukeond_test.go
+++ b/cmd/kukeond/kukeond_test.go
@@ -57,6 +57,7 @@ func TestApplyServerConfigurationDefaultsLayered(t *testing.T) {
 		ReconcileInterval:         "45s",
 		ContainerdNamespaceSuffix: "dev.kukeon.io",
 		CgroupRoot:                "/kukeon-dev",
+		DefaultMemoryLimitBytes:   2 * 1024 * 1024 * 1024,
 	}
 	applyServerConfiguration(cmd, spec)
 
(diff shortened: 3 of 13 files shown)


eminwux:
**PR #533 Review — `fix: add daemon-default memory cap fallback for cgroup-unlimited containers`**

Solid plumbing — the option threads cleanly through the controller→runner→ctr BuildOption chain, the precedence ladder (`--flag > env > YAML > 0`) matches existing knobs, the `> 0` guard on `WithDefaultMemoryLimit` plus the `< 0` clamp in `runServe` correctly bound the int64→uint64 conversion, and the new test coverage exercises every branch including the `*int64(0)` case the existing `securitySpecOpts` skips. Per-commit signature verified (`valid`). No sensitive data in logs. Public surface (`pkg/api/model/v1beta1/serverconfiguration.go`) is updated in lockstep. One blocking item before merge:

- **Convention drift** `cmd/config/env.go:125-127`: every other daemon-only knob in this file uses the `KUKEOND_*` env-var prefix (`KUKEOND_SOCKET`, `KUKEOND_SOCKET_GID`, `KUKEOND_CONFIGURATION`, `KUKEOND_RECONCILE_INTERVAL`) paired with a `kukeond/*` viper key, while root-scope `KUKEON_*` vars are for CLI-wide settings. This new entry uses `KUKEON_DEFAULT_MEMORY_LIMIT_BYTES` for the env var even though its viper key (`kukeond/defaultMemoryLimitBytes`), its binding list (`cmd/kukeond/kukeond.go:200-204` `bindEnvVars`), and its only consumer (`cmd/kukeond/serve.go:95`) are strictly daemon-only — and it's the lone `KUKEON_*` entry in the daemon's `bindEnvVars` list. Operators setting this in env files will hit a documented surface inconsistency that survives forever once shipped. Rename the Go const, the env-var string literal, and the test references in `cmd/kukeond/kukeond_test.go:108` to `KUKEOND_DEFAULT_MEMORY_LIMIT_BYTES` so the daemon's public env surface stays uniform.

Test-plan deferrals (`make e2e`, `make dev-init` daemon-parity smoke) are host-gated and documented per project CLAUDE.md — reviewer to run the dev-init smoke on a clean host pre-merge since `internal/controller/runner/` is on the read path.

LGTM once the env-var rename lands.

eminwux:
Addressed in b87b295:
- Renamed `KUKEON_DEFAULT_MEMORY_LIMIT_BYTES` → `KUKEOND_DEFAULT_MEMORY_LIMIT_BYTES` (env-var string, Go const, all call sites in `cmd/config/env.go`, `cmd/kukeond/{kukeond,serve}.go`, `cmd/kukeond/kukeond_test.go`, plus the doc comment in `internal/controller/controller.go`). The viper key (`kukeond/defaultMemoryLimitBytes`) and the YAML field were already daemon-scoped, so no template/YAML changes were needed. The daemon's `bindEnvVars` list now has no `KUKEON_*` outliers — every daemon-only knob carries the `KUKEOND_*` prefix.

Verified locally: `go vet ./...`, `go build ./...`, full unit suite (`go test $(go list ./... | grep -v /e2e) -count=1`), and pre-commit on the changed files all pass. `make e2e` and the `make dev-init` daemon-parity smoke remain deferred to a clean host per the original PR body (host-gated, no new e2e contract).

Rebased onto `origin/main` (picked up #529) before the force-with-lease push.

eminwux:
**PR #533 Review — `fix: add daemon-default memory cap fallback for cgroup-unlimited containers`**

Clean implementation. Plumbing mirrors the existing `KUKEOND_SOCKET_GID` pattern end-to-end (flag → viper → `applyServerConfiguration` gate → `runServe` → `daemon.Options.Controller` → `runner.Options` → `ctr.BuildOption`), defense-in-depth on the value (clamp in `serve.go`, `> 0` guard in `WithDefaultMemoryLimit`, nil-return in `daemonDefaultBuildOpts`, `specHasMemoryLimit` guard before `oci.WithMemoryLimit`), and `specHasMemoryLimit` correctly preserves the "explicit per-container value always wins" AC. Unit tests in `spec_default_memory_test.go` cover every branch for both `BuildContainerSpec` and `BuildRootContainerSpec`. `daemonDefaultBuildOpts()` returns a fresh slice each call so the `append` pattern in `start.go`/`provision.go` has no aliasing risk. CI green; both commits verified; no sensitive-data exposure.

A non-blocking nit:

- **Flag-overrides-env test missing for new field** `cmd/kukeond/kukeond_test.go`: `TestApplyServerConfigurationDefaultsLayered` covers YAML-layered and `TestApplyServerConfigurationDefaultMemoryLimitEnvOverridesConfig` covers env-over-YAML for the new field, but the flag-over-env tier is only exercised for the pre-existing fields in `TestApplyServerConfigurationFlagOverridesConfig`. Same `flagChanged` helper backs all fields, so this is structural coverage, but an explicit row for `DefaultMemoryLimitBytes` would round it out.

The deferred `make e2e` and `make dev-init` smokes are documented host-gated items per CLAUDE.md — non-blocking, but the user should run both pre-merge (the diff touches `internal/controller/runner/` so daemon-parity matters).

LGTM — unconditional.

PR #535, opened 2026-05-15
fix: warn in kuke doctor cgroups on no-swap and no userspace OOM hosts

### LINKED ISSUE #532: bug: kuke doctor cgroups passes green on no-swap + no-userspace-OOM hosts (would have flagged the 2026-05-15 incident host)

eminwux:
## Summary
- Extends `kuke doctor cgroups` with two new host-environment probes that flag the configuration the 2026-05-15 kukeon-dev incident named: hosts with no swap, and hosts with no userspace OOM guard (`systemd-oomd` / `earlyoom` / `oomd`). On either gap the unscoped pre-flight prints a `warning: …` line to stderr explaining the risk and pointing the operator at the daemon-side fallback (`kuke daemon --default-memory-limit-bytes`, the companion fix in #533).
- Both probes are advisory: they never change the exit code, per the issue #532 author's directive *"DO NOT make the process fail, only high severity warnings"*. The probes are scoped to the unscoped host-root path only — `--scope realm|space|stack|cell` skips them because the host environment does not depend on which sub-tree the operator is inspecting.
- New file `cmd/kuke/doctor/cgroups/host_probes.go` with two package-level probe vars (`probeSwap`, `probeUserspaceOOM`) and path-parameterized cores (`probeSwapAt`, `probeUserspaceOOMAt`) so unit tests can exercise the file-parsing and `/proc`-walking branches without touching real system state. The OOM-guard probe walks `/proc` directly instead of shelling out to `systemctl` / `pgrep` so the doctor stays dependency-free on minimal hosts.
- `runCheck` defers `runHostProbes(stderr)` once on the unscoped path — guarantees the warnings reach the operator regardless of which exit branch fires (silent success, self-heal downgrade, or fatal failure). On all-green hosts the probes return empty strings, preserving the dev-init.sh tail-clean rule the project CLAUDE.md regression guard depends on.

## Behavior preserved
- Exit codes are unchanged on every existing branch — warnings are additive output, never a failure escalation. The "silent on success" contract for the dev-init.sh tail still holds when both probes return empty.
- The `--scope` paths are untouched: scoped pre-flights inspect a sub-tree, not the host environment, so they continue to emit only their cgroup-status output.

## Implementation calls worth flagging
- The original AC point 3 ("If both checks fail AND the daemon-default memory cap is not set on `kukeond`, the doctor should fail-not-warn") is dropped: the issue's sole comment from the author overrides it with *"DO NOT make the process fail, only high severity warnings"*. The two probes therefore emit independent warning lines instead of a combined fail-or-warn verdict, and the doctor never reads kukeond config — which also keeps this PR independent of the open #533 daemon-default-memory-cap landing.
- Both probes default to silent on read errors (unreadable `/proc/swaps`, missing `/proc`) rather than emitting false-positive warnings on platforms that don't expose those interfaces — same conservative posture the existing self-heal gate uses.
- Tests use the same `Set…ForTest` package-var-swap pattern as the existing `SetActiveProberForTest` / `SetSelfHealGateForTest` hooks, plus two `*AtForTest` direct-call exposures so the file-parsing and `/proc`-walking branches can be unit-tested against `t.TempDir()`-backed synthetic fixtures.

## Test plan
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `go test \$(go list ./... | grep -v /e2e) -count=1 -timeout 600s` — full unit suite passes; the new tests in `cmd/kuke/doctor/cgroups/host_probes_test.go` (probe combinations, scoped-path skip, fire-on-failure, direct probe-function unit coverage) all pass alongside the existing cgroups corpus
- [x] `pre-commit run --files \$(git diff --name-only origin/main..HEAD)` — addlicense, golangci-lint (govet shadow check among others), go-fmt, go-mod-tidy, end-of-file, trailing-whitespace all pass
- [ ] `make e2e` — **deferred** to the reviewer's environment: this dev session runs inside a kuke-managed container (`/run/kukeon/tty`, `/.kukeon/bin/kuketty`, `/.kukeon/kuketty/metadata.json` are visible mounts) and a prior session that drove `make e2e` from inside this same wrapper broke the outer host with overlay-on-overlay IO and leaked CNI bridges. Per the role playbook's "verification step can't safely run on the current host" section, the e2e check needs a clean host. The change is to a doctor pre-flight (advisory output only, no exit-code change), so the existing e2e suite should be unaffected.
- [ ] `make dev-init` daemon-parity smoke (project CLAUDE.md §Local smoke test) — same defer + same reason. The diff does not touch the daemon path or anything under `/opt/kukeon`, so the parity check should be unaffected, but reviewer should run it on a clean host before merge to confirm.

## New tests
- `cmd/kuke/doctor/cgroups/host_probes_test.go` — six end-to-end command tests + two direct probe-function tests:
  - `TestCgroupsCmdHostProbeSwapMissingWarns` / `TestCgroupsCmdHostProbeOOMMissingWarns` — each probe in isolation emits a `warning:` line on stderr without changing exit code.
  - `TestCgroupsCmdHostProbesBothMissingWarnsBoth` — both probes fire produce two independent warning lines (no collapsed verdict).
  - `TestCgroupsCmdHostProbesAllGreen` — both probes empty → silent on stderr (preserves dev-init.sh tail-clean rule).
  - `TestCgroupsCmdHostProbesScopedSkipped` — `--scope realm` never triggers the host probes even when both would warn loudly.
  - `TestCgroupsCmdHostProbesFireOnFailure` — host warnings still emit when the cgroup pre-flight itself fails (the defer-runHostProbes contract).
  - `TestProbeSwapAt` — direct unit coverage of the `/proc/swaps` parser: missing file (silent), header-only (warns), header + entry (silent).
  - `TestProbeUserspaceOOMAt` — direct unit coverage of the `/proc`-walker against a synthetic process tree: empty `/proc` (warns), each of `systemd-oomd` / `earlyoom` / `oomd` present (silent), unrelated process only (warns).
- `TestMain` in the existing `cgroups_test.go` neutralizes the production probes to no-op returns so the existing test corpus does not depend on the runner's actual `/proc/swaps` or running-process set; new probe tests opt back in locally.

Closes #532

### DIFF
--- a/cmd/kuke/doctor/cgroups/cgroups.go
+++ b/cmd/kuke/doctor/cgroups/cgroups.go
@@ -218,6 +218,16 @@ func runCheck(stdout, stderr io.Writer, root string, nested, probe, verbose bool
 		return fmt.Errorf("cgroup pre-flight: %w", err)
 	}
 
+	// Host-level probes only fire on the unscoped host-root path —
+	// per-realm/space/stack/cell scopes inspect a sub-tree, not the host
+	// environment, so the host warnings would be misleading there. Deferred
+	// so they reach the operator regardless of which exit branch the
+	// cgroup-check result lands in (silent success, self-heal downgrade,
+	// or fatal failure).
+	if scopeLabel == "" {
+		defer runHostProbes(stderr)
+	}
+
 	if res.OK() {
 		// Silence on success keeps the dev-init.sh tail clean — only
 		// surface details when --verbose-status is set.
--- a/cmd/kuke/doctor/cgroups/cgroups_test.go
+++ b/cmd/kuke/doctor/cgroups/cgroups_test.go
@@ -33,17 +33,24 @@ import (
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 )
 
-// TestMain mocks the shared euid lookup to euid=0 for every test in this
-// package. The new fail-fast root gate on --probe would otherwise
+// TestMain mocks the shared euid lookup to euid=0 and neutralizes the
+// host-environment probes (swap + userspace OOM guard) for every test in
+// this package. The fail-fast root gate on --probe would otherwise
 // short-circuit every default-probe and --probe test under non-root CI
 // runners (ubuntu-latest defaults to UID 1001), even though their fake
-// cgroupfs lives under t.TempDir() and never needs real root. Individual
-// cases that exercise the non-root rejection override this with their own
-// SetGeteuidForTesting call.
+// cgroupfs lives under t.TempDir() and never needs real root. The host
+// probes default to no-op returns so existing tests do not depend on the
+// runner's actual /proc/swaps or running-process set; individual probe
+// tests below override these with their own SetHostProbesForTest call.
 func TestMain(m *testing.M) {
-	restore := kukshared.SetGeteuidForTesting(func() int { return 0 })
+	restoreEuid := kukshared.SetGeteuidForTesting(func() int { return 0 })
+	restoreProbes := cgroupscmd.SetHostProbesForTest(
+		func() string { return "" },
+		func() string { return "" },
+	)
 	code := m.Run()
-	restore()
+	restoreProbes()
+	restoreEuid()
 	os.Exit(code)
 }
 
--- a/cmd/kuke/doctor/cgroups/export_test.go
+++ b/cmd/kuke/doctor/cgroups/export_test.go
@@ -36,3 +36,33 @@ func SetSelfHealGateForTest(fn func(string) bool) func() {
 	selfHealGate = fn
 	return func() { selfHealGate = prev }
 }
+
+// SetHostProbesForTest swaps both host probes and returns a restore closure.
+// TestMain calls this to neutralize the production /proc readers so the
+// existing test corpus does not depend on the runner's actual swap or
+// userspace-OOM configuration; individual tests call it again locally to
+// simulate specific bad-config hosts. Either argument may be nil to leave
+// that probe untouched.
+func SetHostProbesForTest(swap, oom func() string) func() {
+	prevSwap, prevOOM := probeSwap, probeUserspaceOOM
+	if swap != nil {
+		probeSwap = swap
+	}
+	if oom != nil {
+		probeUserspaceOOM = oom
+	}
+	return func() {
+		probeSwap = prevSwap
+		probeUserspaceOOM = prevOOM
+	}
+}
+
+// ProbeSwapAtForTest exposes the path-parameterized swap probe so unit
+// tests can exercise its file-parsing branches against a tmpdir-backed
+// fake /proc/swaps.
+func ProbeSwapAtForTest(path string) string { return probeSwapAt(path) }
+
+// ProbeUserspaceOOMAtForTest exposes the path-parameterized OOM-guard
+// probe so unit tests can exercise its /proc-walking logic against a
+// tmpdir-backed fake /proc populated with synthetic <pid>/comm files.
+func ProbeUserspaceOOMAtForTest(procDir string) string { return probeUserspaceOOMAt(procDir) }
--- a/cmd/kuke/doctor/cgroups/host_probes.go
+++ b/cmd/kuke/doctor/cgroups/host_probes.go
@@ -103,6 +103,9 @@ func defaultProbeUserspaceOOM() string {
 // sandboxed environment without /proc silently skips the check rather
 // than emitting a false-positive warning.
 func probeUserspaceOOMAt(procDir string) string {
+	if _, err := os.Stat(procDir); err != nil {
+		return ""
+	}
 	for _, name := range []string{"systemd-oomd", "earlyoom", "oomd"} {
 		if processRunningAt(procDir, name) {
 			return ""
--- a/cmd/kuke/doctor/cgroups/host_probes.go
+++ b/cmd/kuke/doctor/cgroups/host_probes.go
@@ -1,0 +1,139 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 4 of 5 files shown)


eminwux:
**PR #535 Review — `fix: warn in kuke doctor cgroups on no-swap and no userspace OOM hosts`**

Solid, well-tested addition — two advisory probes deferred onto the unscoped path with no exit-code change, comprehensive coverage including the defer-on-failure pin. License header, commit verification, and test plan check out (host-gated `make e2e` / `dev-init` smokes deferred with documented justification, OK per CLAUDE.md). One blocking mismatch:

- **Comment/PR claim diverges from code** `cmd/kuke/doctor/cgroups/host_probes.go:102-112`: the comment says `probeUserspaceOOMAt` "returns empty when /proc itself is unreadable so a sandboxed environment without /proc silently skips the check" (and the PR body claims the same for both probes), but on an unreadable `/proc` all three `processRunningAt` calls return false via the `ReadDir` err path, the loop falls through, and the function emits the OOM-guard warning anyway — the opposite of the documented behavior. Either gate `probeUserspaceOOMAt` on an upfront `os.Stat(procDir)` (or hoist the `ReadDir` to the outer function and bail empty on err), or update the comment + PR body to match the warn-loudly behavior.

LGTM once that lands — the rest of the diff is clean.

eminwux:
Addressed in b096fc5:
- `probeUserspaceOOMAt` now gates on an upfront `os.Stat(procDir)` and returns empty when the directory is unreadable, matching the documented "silently skips on hosts without /proc" behavior the comment + PR body claimed.
- Added a `missing procDir` case to `TestProbeUserspaceOOMAt` that removes the synthetic `t.TempDir()` before invocation and asserts no warning, pinning the new branch.

Rebased onto `origin/main` over #533 (no conflicts; the daemon-default cap landed in `internal/daemon`, this PR touches `cmd/kuke/doctor/cgroups` only).

eminwux:
**PR #535 Re-Review — `fix: warn in kuke doctor cgroups on no-swap and no userspace OOM hosts`**

`b096fc5` lands the upfront `os.Stat(procDir)` gate in `probeUserspaceOOMAt` (`cmd/kuke/doctor/cgroups/host_probes.go:106-108`), so an unreadable `/proc` now returns empty silently and matches the documented "sandboxed environment without /proc silently skips" contract. Test case `missing procDir` (`host_probes_test.go:268-275`) pins the new branch. Both follow-up commits verified.

Blocker resolved. Host-gated `make e2e` and `make dev-init` daemon-parity smokes remain deferred with documented justification per CLAUDE.md — worth running on a clean host before merge as the author flagged.

LGTM.

PR #536, opened 2026-05-16
ci(installer): add explicit permissions: contents: read

### LINKED ISSUE #452: ci(installer): add explicit `permissions: contents: read` to installer.yaml jobs
Surfaced during review of #451 — non-blocking, deferred to a follow-up.

CodeQL fires `actions/missing-workflow-permissions` for each of the four jobs in `.github/workflows/installer.yaml`. The workflow does checkout + script execution only, so the minimal fix is a top-level block:

\`\`\`yaml
permissions:
  contents: read
\`\`\`

That clears all four alerts in one place and matches the convention already used in \`release.yaml\` and \`mkdocs.yaml\`.

**Where:** \`.github/workflows/installer.yaml\` (top level, before \`jobs:\`)

**Why filed:** CodeQL workflow-permissions alerts are bug-adjacent — the default \`GITHUB_TOKEN\` is overly permissive on workflows without an explicit \`permissions\` block, and the project already enforces this convention elsewhere. Filing rather than skipping per CLAUDE.md "bug-adjacent observations" / project-convention precedent.

eminwux:
## Summary
- Closes CodeQL's `actions/missing-workflow-permissions` alerts on `.github/workflows/installer.yaml`'s four jobs in one place. The workflow only checks out the repo and runs scripts, so the minimal least-privilege block is `contents: read`.
- Convention precedent: `mkdocs.yaml` already declares its `permissions:` block at the top level (between `name:` and `on:`), so this matches it stylistically. `release.yaml`'s per-job permissions stay untouched (it needs `packages: write`, `attestations: write`, `id-token: write` only for the release job).

## Test plan
- [x] `pre-commit run --files .github/workflows/installer.yaml` — trim-trailing-whitespace, end-of-file-fixer, check-yaml, prettier all pass
- [x] `make test` — full unit suite passes (no Go code touched, but ran per role playbook)
- [ ] The `Installer` workflow itself — runs in CI on this PR (paths-filter matches `.github/workflows/installer.yaml`), so all four jobs (`sync`, `prereq-missing`, `prereq-pass`, `install-skip-init`) will execute with the new `contents: read` token and confirm the permission is sufficient.

Closes #452

### DIFF
--- a/.github/workflows/installer.yaml
+++ b/.github/workflows/installer.yaml
@@ -1,5 +1,8 @@
 name: Installer
 
+permissions:
+  contents: read
+
 on:
   pull_request:
     branches:


eminwux:
**PR #536 Review — `ci(installer): add explicit permissions: contents: read`**

LGTM. Minimal least-privilege fix for CodeQL's `actions/missing-workflow-permissions` alert — `contents: read` is the right scope since all four jobs only check out the repo and run scripts. The workflow-level placement matches `mkdocs.yaml`'s precedent (`.github/workflows/mkdocs.yaml:2-5`), and `release.yaml`'s per-job block (`packages: write` + `contents: write` + `attestations: write` + `id-token: write`) is correctly left alone since it has scope it actually needs. No security or correctness concerns.

The third test-plan item — the `Installer` workflow itself running on this PR — has now executed green across all four jobs in the rollup, confirming the new token is sufficient. Effectively self-validating.

LGTM.

PR #540, opened 2026-05-16
test: cover ensureAttachableSocketSymlink ErrSocketPathTooLong fail-fast

### LINKED ISSUE #534: runner: add unit test for ensureAttachableSocketSymlink ErrSocketPathTooLong path
Surfaced as a non-blocking nit during review of #529.

The provision-time fail-fast in `ensureAttachableSocketSymlink` (AC #2 of #521) is asserted only by inspection — the failure path at `internal/controller/runner/attachable.go:596-599` has no unit-level coverage:

```go
if len(symlinkPath) > consts.KukeonMaxSocketPath {
    return fmt.Errorf("%w: %q is %d bytes (limit %d)",
        errdefs.ErrSocketPathTooLong, symlinkPath, len(symlinkPath), consts.KukeonMaxSocketPath)
}
```

A future refactor that drops or inverts the guard would regress the AC silently — the existing `TestContainerSocketSymlinkPath_FitsSUNPath` only verifies that normal-shaped inputs *stay under* the limit; it does not assert that the runner *refuses to provision* when they do not.

Suggested: add a unit test in `internal/controller/runner/attachable_test.go` (or alongside the existing fs tests) that calls `ensureAttachableSocketSymlink` with a contrived RunPath/spec whose resolved symlink path exceeds `consts.KukeonMaxSocketPath`, and asserts `errors.Is(err, errdefs.ErrSocketPathTooLong)` plus that no symlink was staged on disk.

eminwux:
## Summary
- Issue #534: the provision-time fail-fast added in #521 AC #2 at `internal/controller/runner/attachable.go:596-599` was only asserted by inspection. A future refactor that dropped or inverted the length guard would silently regress the AC and defer the failure to first `kuke attach` (opaque `connect(2)` `ENAMETOOLONG`).
- New unit test `TestEnsureAttachableSocketSymlink_RefusesOverflow` in `internal/controller/runner/attachable_test.go` constructs a contrived `RunPath` (real tmp dir padded so `<runPath>/s/<16hex>` exceeds `consts.KukeonMaxSocketPath` by one byte), calls `ensureAttachableSocketSymlink`, and asserts `errors.Is(err, errdefs.ErrSocketPathTooLong)` plus that no on-disk state was staged (`<runPath>/s` dir absent, symlink inode absent).
- The two on-disk assertions are split so a regression that reorders `MkdirAll` above the length check (dir assertion catches) is distinct from one that fires the length check but still proceeds to `os.Symlink` (symlink assertion catches).

## Test plan
- [x] `go test ./internal/controller/runner/ -run TestEnsureAttachableSocketSymlink_RefusesOverflow -v` → PASS
- [x] `go test ./internal/controller/runner/...` → PASS
- [x] `go vet ./internal/controller/runner/...` → clean
- [x] `make test` (project CI target, full unit-test sweep) → all ok, no FAIL
- [x] `pre-commit run --files internal/controller/runner/attachable_test.go` → all hooks pass (go fmt, go-mod-tidy, golangci-lint, addlicense)
- [ ] `make dev-init` smoke → not required per `CLAUDE.md`: this change touches a single `_test.go` file in one package, not the build path, the daemon, or anything under `/opt/kukeon`.

Closes #534

### DIFF
--- a/internal/controller/runner/attachable_test.go
+++ b/internal/controller/runner/attachable_test.go
@@ -22,15 +22,20 @@ package runner
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"log/slog"
 	"os"
 	"path/filepath"
 	"reflect"
+	"strings"
 	"testing"
 
+	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/ctr"
+	"github.com/eminwux/kukeon/internal/errdefs"
 	intmodel "github.com/eminwux/kukeon/internal/modelhub"
+	"github.com/eminwux/kukeon/internal/util/fs"
 	sbshapi "github.com/eminwux/sbsh/pkg/api"
 )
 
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #540 Review — `test: cover ensureAttachableSocketSymlink ErrSocketPathTooLong fail-fast`**

Tight, well-scoped regression test for the #521 AC #2 fail-fast. Path-length math is parameterized via `consts.KukeonMaxSocketPath` and self-validated with the pre-assertion that the constructed `symlinkPath` actually exceeds the limit, so future bumps to the constant won't silently turn this into a vacuous test. The two-pronged on-disk assertion (no `s/` dir, no symlink inode) cleanly separates the "MkdirAll moved above the length check" regression from the "length check ran but Symlink still proceeded" regression — exactly the failure modes worth pinning. No correctness, leak, or convention issues.

Minor nit (non-blocking):

- **Math comment** `internal/controller/runner/attachable_test.go:738-742`: comment says the padding overshoots the limit "by one byte" but `filepath.Join(base, padding)` inserts a separator the formula doesn't account for, so the constructed `symlinkPath` is actually 2 bytes over (`limit+1 - len(base) - 3 - 16` + the extra `/`). Test still passes correctly because the `len(symlinkPath) <= limit` guard catches any real undershoot; cosmetic only.

LGTM.

PR #542, opened 2026-05-16
feat: install kukeond systemd unit so the daemon survives reboot

### LINKED ISSUE #541: install: wire kukeond systemd unit so the daemon survives host/containerd restart
## Context

Today nothing brings kukeond back after a host or containerd restart. kukeond runs as a containerd container in the `kuke-system` realm (the `kuke-system / kukeon / kukeon / kukeond` cell), but containerd does not restart tasks after its own restart, and there is no host-level supervisor on the kuke side.

Grep across `scripts/install.sh`, `docs/site/install.sh`, and the Go tree confirms: no systemd unit, no `AutoStart` / `StartOnBoot` logic, no reconcile-on-host-restart path. The `kuke daemon restart` verb at `cmd/kuke/daemon/restart/restart.go` is operator-invoked only. Closest neighbors are #298 (kukeond log bind-mount) and #217 (daemon subcommand group / reconciler-driven control flow), neither of which covers boot recovery.

Workaround until this lands: operator runs `kuke daemon start` after every reboot.

## Proposal

Have `scripts/install.sh` drop a `kukeond.service` systemd unit ordered after `containerd.service` that invokes `kuke daemon start` on boot. `kuke uninstall` disables and removes the unit. Document the host-supervisor model and the systemd-less manual fallback.

## Acceptance criteria

- [ ] `scripts/install.sh` and `docs/site/install.sh` install `/etc/systemd/system/kukeond.service` when systemd is detected on the host; skip with a clear notice when not.
- [ ] The unit declares `After=containerd.service`, `Requires=containerd.service`, `Restart=on-failure`, and an `ExecStart` that runs `kuke daemon start` (the in-process verb at `cmd/kuke/daemon/start/start.go`, no `--no-daemon`).
- [ ] The installer runs `systemctl daemon-reload` + `systemctl enable --now kukeond.service` after writing the unit, so the daemon is up at the end of `install.sh` on a fresh host.
- [ ] `kuke uninstall` (`cmd/kuke/uninstall/uninstall.go`) disables and removes the unit before its existing realm/filesystem teardown; uninstall still works on hosts where the unit was never installed (e.g. dev hosts that bootstrapped via `make dev-init`).
- [ ] `make dev-init` continues to pass end-to-end with the new unit installation path skipped or no-op'd in dev mode — the daemon-parity tail check in `scripts/dev-init.sh` must still produce the documented two-realm output.
- [ ] After `systemctl reboot` on an install.sh-bootstrapped host, `kuke get realms` shows both realms `Ready` without operator intervention.
- [ ] README / install docs note the host-supervisor model and the manual `kuke daemon start` fallback for systemd-less hosts.

## Notes

Option considered and rejected for this issue: a kuke-side reconcile loop (cron, or daemon-side supervision via `ctr` task lifecycle hooks). More plumbing without reducing operator surface — systemd is already a de-facto dependency of the containerd that `install.sh` requires (`scripts/install.sh:163` prints `systemctl enable --now containerd` as the canonical bring-up).

Dev-mode hosts (the `make dev-init` path documented in `CLAUDE.md`) typically run without systemd and re-bootstrap via the dev script anyway, so the systemd path being skipped there is intentional, not a regression.

Related: #298, #217.

eminwux:
## Summary
- Issue #541: nothing brought kukeond back after a host or containerd restart. containerd does not restart tasks across its own restart, and the kuke side had no host-level supervisor — operators had to run `kuke daemon start` after every reboot.
- `scripts/install.sh` (and the synced `docs/site/install.sh`) now drops `/etc/systemd/system/kukeond.service` ordered `After=containerd.service` / `Requires=containerd.service`, then `systemctl daemon-reload` + `systemctl enable --now kukeond.service`. The unit's `ExecStart` is `${KUKE_INSTALL_PREFIX}/kuke daemon start` (the in-process verb at `cmd/kuke/daemon/start/start.go`, idempotent against an already-running daemon, no `--no-daemon`). On systemd-less hosts the installer prints a notice and skips — operators fall back to manual `sudo kuke daemon start` per reboot.
- Unit shape: `Type=oneshot` + `RemainAfterExit=yes` to make the AC-required `Restart=on-failure` (with `RestartSec=5s`) meaningful around an in-process verb that exits after bringing the kukeond cell up — containerd supervises the kukeond container thereafter.
- `kuke uninstall` (new `cmd/kuke/uninstall/systemd.go`, wired into `runUninstall` before `ctrl.Uninstall`) stops, disables, and removes the unit, then `systemctl daemon-reload`, ahead of realm / filesystem teardown so a `Restart=on-failure` unit cannot respawn kukeond mid-uninstall. Stop / disable / daemon-reload are best-effort (no-op on hosts where systemctl is missing or PID 1 is not systemd); the file remove is the binding step. Idempotent: silent skip when no systemctl in PATH or no unit file (covers `make dev-init` hosts and partial-state re-runs).
- Tests injected via a new `MockSystemdRemoverKey` context key in the same pattern as `MockControllerKey`. Existing tests inject a no-op stub by default so they stay hermetic on hosts that happen to have systemctl present.
- Docs: README + `docs/site/install/install-linux.md` describe the host-supervisor model and the systemd-less manual fallback; `docs/cli-use-cases.md`'s "Full per-host teardown" section restates the uninstall ordering with the new step.

## Test plan
- [x] `go build ./...` → clean
- [x] `go vet ./...` → clean
- [x] `go test ./cmd/kuke/uninstall/... -v` → 13 pass including new `TestUninstall_SystemdUnitRemovedRendersRow`, `TestUninstall_SystemdSkippedEmitsNoRow`, `TestUninstall_SystemdPartialPath`, `TestUninstall_SystemdFailureShortCircuits`
- [x] `make test` (project CI target, full unit-test sweep) → all ok, no FAIL
- [x] `make dev-init` (project smoke) → daemon-parity tail produces the canonical two-realm output; AC #5 ("dev-init continues to pass end-to-end") verified
- [x] Manual `sudo ./kuke uninstall -y` on the dev host with **no** systemd unit installed → systemd row absent; rest of the teardown unchanged (verifies the AC #4 "uninstall still works on hosts where the unit was never installed" path)
- [x] Manual `sudo ./kuke uninstall -y` on the dev host with a stub unit file present → file removed; partial-line rendered because PID 1 is not systemd here (verifies the best-effort stop/disable/daemon-reload behavior)
- [x] `pre-commit run --files <changed>` → go fmt, go-mod-tidy, golangci-lint, prettier, addlicense all green
- [ ] AC #6 (`systemctl reboot` → `kuke get realms` shows both realms Ready without operator intervention) — this dev host has no systemd PID 1, so a reboot smoke cannot run locally; the unit's `Type=oneshot` + `Restart=on-failure` + `kuke daemon start` (which is idempotent and errors only when the host has not been `kuke init`-ed) is the design-time argument that it will. Reviewer to verify on a systemd-bootstrapped host.

Closes #541

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Get kukeon running on a single Linux host in minutes.
 curl -fsSL https://kukeon.io/install.sh | bash
 ```
 
-The installer detects your platform, verifies the prerequisites above (and prints a distro-aware hint on any miss), downloads the latest release, verifies its `sha256` checksum, installs `kuke` + `kukeond` to `/usr/local/bin`, and runs `sudo kuke init` to bring the daemon up. Pass `--check` to run the prereq checks only without touching the system:
+The installer detects your platform, verifies the prerequisites above (and prints a distro-aware hint on any miss), downloads the latest release, verifies its `sha256` checksum, installs `kuke` + `kukeond` to `/usr/local/bin`, runs `sudo kuke init` to bring the daemon up, and (on systemd hosts) installs `/etc/systemd/system/kukeond.service` so kukeond comes back automatically after a host or containerd restart. On systemd-less hosts the unit step is skipped with a notice — bring kukeond up manually after each reboot with `sudo kuke daemon start`. Pass `--check` to run the prereq checks only without touching the system:
 
 ```bash
 curl -fsSL https://kukeon.io/install.sh | bash -s -- --check
--- a/cmd/kuke/uninstall/systemd.go
+++ b/cmd/kuke/uninstall/systemd.go
@@ -1,0 +1,133 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package uninstall
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+// SystemdUnitPath is the absolute path of the kukeond systemd unit
+// scripts/install.sh writes. Exported so the test package can match against
+// it without re-spelling the literal.
+const SystemdUnitPath = "/etc/systemd/system/kukeond.service"
+
+// SystemdUnitName is the canonical short name the systemctl verbs accept.
+const SystemdUnitName = "kukeond.service"
+
+// SystemdRemovalReport summarizes what removeKukeondSystemdUnit did. Every
+// field is independently surfaced in the CLI report so the operator can tell
+// which sub-step ran on a partially-supervised host (e.g., systemctl present
+// but the unit was hand-deleted before re-running uninstall).
+type SystemdRemovalReport struct {
+	// Skipped is true when the helper short-circuited without touching the
+	// host (no systemctl in PATH, or no unit file). SkipReason names the
+	// branch so the printed report is self-explanatory.
+	Skipped    bool
+	SkipReason string
+	// UnitPath is the absolute unit file path the helper attempted to
+	// disable; carried in the report so a future change to SystemdUnitPath
+	// does not require chasing the renderer to update its hard-coded text.
+	UnitPath string
+	// Stopped/Disabled/FileRemoved/DaemonReloaded each flip true on success
+	// of the corresponding sub-step. Stop / disable / daemon-reload are
+	// best-effort — the binding step is FileRemoved. The fields are
+	// independent so the renderer can show (e.g.) "stop+disable failed
+	// (no systemd PID 1) but file removed" on partial paths.
+	Stopped        bool
+	Disabled       bool
+	FileRemoved    bool
+	DaemonReloaded bool
+	// Err carries the first non-nil binding-step error so runUninstall can
+	// propagate a non-zero exit. Best-effort sub-steps (stop / disable /
+	// daemon-reload) do not populate Err — their outcomes are visible via
+	// their bool fields. The report is still returned in full on error so
+	// the operator sees what did succeed before the failure.
+	Err error
+}
+
+// SystemdRemover is the contract removeKukeondSystemdUnit and any test
+// substitute satisfy. Injected via MockSystemdRemoverKey so tests need not
+// touch the host's systemctl or /etc/systemd/system.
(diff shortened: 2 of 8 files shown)


eminwux:
**PR #542 Review — `feat: install kukeond systemd unit so the daemon survives reboot`**

Solid host-supervisor wiring: atomic unit-file install, idempotent `kuke daemon start` confirmed by reading `cmd/kuke/daemon/start/start.go` (short-circuits to `nil` when the cell is already Ready), systemctl ops correctly best-effort with `os.Remove` as the binding step, ordering rationale documented, hermetic test injection via `MockSystemdRemoverKey` matches the existing `MockControllerKey` pattern, and the four new render-path tests pin every branch (success / partial / skipped / error-shortcircuit). No sensitive-data, dead-code, or convention concerns. All commits verified.

One blocking item — a renumbering miss that makes the CLI help text internally contradict itself:

- **Doc accuracy** `cmd/kuke/uninstall/uninstall.go:77`: after renumbering the list at lines 67-75 from 1-4 to 1-5 (systemd inserted as step 1, everything else shifted +1), the line `steps 2-4 are skipped` is stale — under the new numbering, step 2 is the realm purge itself (the step that *fails*, not gets skipped) and step 5 is the now-skipped user/group cleanup. As written, `kuke uninstall --help` tells an operator that user/group removal still runs after a realm-purge failure, which is the opposite of the actual gate behavior (and contradicts `docs/cli-use-cases.md` which kept the old numbering scheme and still correctly says `Steps 2–4 are gated`). Change to `steps 3-5 are skipped`.

Non-blocking — AC #6 (`systemctl reboot` smoke) is unchecked because the agent host has no systemd PID 1; per the host-gated carve-out the author documented in the PR body, this needs to run on a systemd-bootstrapped host before merge.

LGTM once the step-range update lands.

eminwux:
Addressed in aa6278e:
- `cmd/kuke/uninstall/uninstall.go:77` — updated stale `steps 2-4 are skipped` → `steps 3-5 are skipped` to match the renumbered 1-5 list (systemd is now step 1, realm purge is step 2 = the failing gate, `/run/kukeon` + `/opt/kukeon` + user/group are the gated 3-5). Verified `docs/cli-use-cases.md`'s "Steps 2–4 are gated" wording stayed correct under its own -1/0/1-4 numbering so no doc sync needed.

Test plan re-run for the fix-up: `go test ./cmd/kuke/uninstall/...` ok, `make test` ok, `pre-commit run --files cmd/kuke/uninstall/uninstall.go` all green.

Non-blocking AC #6 (`systemctl reboot` smoke) still `[ ]` — same host-gated reason as the original PR body (no systemd PID 1 on this dev host).

eminwux:
**PR #542 Re-review — `feat: install kukeond systemd unit so the daemon survives reboot`**

Follow-up commit `aa6278e` ("docs: fix stale step-range in uninstall long help after systemd renumber") flips `steps 2-4` → `steps 3-5` at `cmd/kuke/uninstall/uninstall.go:77`, matching the new 1-5 numbering. Single-line text fix, no scope creep. Blocker closed.

LGTM to merge. Reviewer to still run the AC #6 reboot smoke on a systemd-bootstrapped host (test plan item that can't run on the agent sandbox) before pulling the trigger.

PR #544, opened 2026-05-16
fix: keep reconciler progressing past Unknown after host reboot

### LINKED ISSUE #543: bug: reconciler stays at Unknown after host reboot (cgroup absent + task gone)
## Context

After a host reboot, kukeon cells that were Ready beforehand are pinned at `State: Unknown` indefinitely. The daemon comes back, the reconcile loop runs every 30s, but cells never transition to `Stopped` and so never become eligible for the AutoDelete / wind-down paths — operators must manually `kuke purge cell <name>` for each orphan to recover.

Observed on an active dev host this afternoon: 22 cells in metadata, 8 in `Ready` (cells the operator had restarted post-reboot), 14 in `Unknown`. Every Unknown cell had:

- A persistent containerd container in `default.kukeon.io` (visible via `ctr -n default.kukeon.io containers ls`).
- No running task (the container record survives the reboot; the task does not).
- No cgroup directory under `/sys/fs/cgroup/kukeon/...` (cgroups are tmpfs-backed and wiped at boot).
- `readyObserved: true` in persisted metadata (the latch was set before the reboot).
- `metadata.json` mtime frozen at ~5 min after boot — the reconciler had touched it once and then determined nothing else needed updating.

Host had been up 20 hours at observation; `uptime -s` showed the reboot at `2026-05-15 22:53:41`, ~5 minutes before the metadata mtime on every Unknown cell.

## Root cause

Two layers, both contributing:

### Layer 1: `ReconcileCell` gates derivation on cgroup existence

`internal/controller/runner/refresh.go:512-515`:

```go
newState := intmodel.CellStateUnknown
if cgroupErr == nil && cgroupExists {
    newState = r.deriveCellState(cell)
}
```

When the cgroup is absent — which is exactly the post-reboot case — derivation is skipped entirely and the cell is parked at `Unknown`. Container-task information is never consulted at the cell level. `cellStateAutoDeleteTriggers` (same file, line 623) only returns true for `Stopped`, so neither `shouldAutoDeleteCell` nor `shouldWindDownCell` fires — by design, but the design assumes Unknown is a transient containerd hiccup rather than a permanent post-reboot condition.

### Layer 2: `GetContainerState` returns Unknown for a task-less container

`internal/controller/runner/container_state.go:177-183`:

```go
// TaskStatus failed - return Unknown since we can't determine the state
return intmodel.ContainerStateUnknown, nil
```

`TaskStatus` fails when the container exists in containerd but has no associated task — which is exactly the post-reboot state. The "container exists but no task" log line in `refresh.go:405` proves the code path recognizes this case explicitly; it just doesn't translate the observation into `ContainerStateStopped`. So per-container statuses also stay `Unknown`, and `deriveCellStateFromNonRootContainerStatuses` would return `Unknown` (the defensive `anyUnknown` branch) even if Layer 1 were fixed — both layers need to move together to land at `Stopped`.

## Reproduction

1. `kuke create cell <name>` and let it reach `Ready`.
2. Reboot the host.
3. `sudo kuke daemon start` (or via #541 systemd unit once that lands).
4. `kuke get cells` — the cell shows `Unknown`.
5. Wait any number of reconcile passes — state never changes.
6. `kuke refresh` — also no-op (`RefreshCell` uses the same cgroup-existence gate).

## Expected

After the daemon's first post-reboot reconcile pass, every cell whose containerd container exists with no running task should be in `Stopped` (or `Failed`, but `Stopped` is the natural fit for "previously Ready, now no task"). From there:

- AutoDelete cells get reaped automatically (matches `kuke run --rm` semantics).
- Non-AutoDelete cells with at least one non-root spec entry get wind-down `KillCell` called on them.
- Non-AutoDelete cells with no non-root spec entries (kukeond-style cgroup-only cells) just sit in `Stopped` for the operator to restart or delete explicitly.

The key invariant: a post-reboot reconcile pass must not leave cells in a state the rest of the controller has chosen to ignore (`cellStateAutoDeleteTriggers` excludes `Unknown` deliberately).

## Acceptance criteria

- [ ] Layer 2: `GetContainerState` distinguishes \"containerd unreachable / RPC failed\" (legitimately `Unknown`) from \"container exists, no task\" (should return `Stopped`). The existing `ExistsContainer` + `TaskStatus` two-call sequence already has the information needed — the fix is in how the task-missing branch is interpreted.
- [ ] Layer 1: `ReconcileCell` derives a meaningful state when the cgroup is absent but the cell's containers exist in containerd. Two reasonable shapes (pick whichever fits the controller better):
  - (a) Drop the cgroup-existence gate around `deriveCellState` and let the per-container derivation decide. With Layer 2 fixed, this returns `Stopped` for the reboot case naturally.
  - (b) Add an explicit \"cgroup absent + containers exist + tasks gone\" branch that returns `Stopped` and leave the cgroup gate in place for everything else.
- [ ] `kuke refresh` summary actually lists `Cells:` and `Containers:` lines when those updates happen. Today the summary skips empty slices, but `RefreshCell` also won't update anything on Unknown cells because of Layer 1 — once that's fixed, the refresh path must surface the transitions to the operator.
- [ ] Test: simulate cgroup-absent + container-exists-no-task via the runner fakes (the `controller_test.go` fake runner already mocks `GetContainerState`) and assert the cell transitions to `Stopped` after one reconcile pass. Should sit alongside the existing `apply_rematerialize_test.go` cell-state tests.
- [ ] Test: a cell whose containerd container is also gone (operator ran `ctr -n ... containers rm` manually between boots) still derives correctly — likely also `Stopped` via the existing \"container does not exist\" → `ContainerStateStopped` branch in `GetContainerState`. Sanity-check this path isn't regressed.
- [ ] After `systemctl reboot` on a host with N Ready cells and the daemon coming back up via #541, every cell shows `Stopped` (not `Unknown`) within one reconcile interval, and AutoDelete cells are reaped automatically.

## Notes

- Related to #541 (daemon-survival) but distinct. Even with the daemon back up via a systemd unit, this bug leaves the cells stuck. Both need to land for a clean reboot-recovery story.
- The `readyObserved` latch is already set correctly on the persisted cells (verified via `kuke get cell <name> -o yaml` on the observed orphans), so the wind-down / AutoDelete gates have what they need once Layer 1 + Layer 2 land them at `Stopped`.
- Workaround for affected operators today: `sudo kuke purge cell <name>` for each Unknown cell. Destructive but local; the underlying containerd containers and metadata are both cleaned up.
- The two-layer framing in this issue matches how the codebase already separates concerns (`runner.GetContainerState` vs `runner.ReconcileCell`) — fixing one layer without the other won't get to `Stopped` because of the `anyUnknown` defensive branch in `deriveCellStateFromNonRootContainerStatuses`.

eminwux:
## Summary

- **Layer 2 (`internal/controller/runner/container_state.go`):** `GetContainerState` now classifies the post-reboot signature — `ExistsContainer == true` but `TaskStatus` returns a wrapped `errdefs.ErrTaskNotFound` — as `ContainerStateStopped` instead of `Unknown`. Without this, the cell-level derivation can never see past Unknown (both `cellStateAutoDeleteTriggers` and `shouldWindDownCell` exclude Unknown by design) and the reconciler leaves every previously-Ready cell stuck after the host comes back. Any other `TaskStatus` failure still maps to `Unknown` so a transient RPC blip can't flip a healthy cell to Stopped.
- **Layer 1 (`internal/controller/runner/refresh.go`):** Dropped the `cgroupExists` short-circuit in `ReconcileCell` that was forcing `Unknown` whenever the cgroup tree was missing. The tmpfs-backed `/sys/fs/cgroup/kukeon/...` hierarchy doesn't survive reboot, so this gate kept Layer 2 from ever firing. `CgroupReady` is still recorded from the cgroup probe, but the state derivation now runs as long as the probe didn't error.
- **`RefreshCell` parity:** Aligned `RefreshCell` with `ReconcileCell` so `kuke refresh` surfaces the same Ready→Stopped transition for the AC's `Cells/Containers` change summary — populates container statuses, derives state via `deriveCellState`, applies `latchReadyObserved`, and tracks per-container state diffs via the new `countContainerStatusChanges` helper.
- **Latch invariant preserved:** A never-Ready cell still cannot be reaped — `ReadyObserved` is a one-way latch gated on a derived `Ready`, and Layer 2 only converts `task gone + container present` to `Stopped`; the CreateCell-race protection from #275/#278 is untouched.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/controller/runner/...` — all four new tests pass (`TestGetContainerState_TaskNotFound_ReturnsStopped`, `TestGetContainerState_ContainerGoneAlsoStopped`, `TestGetContainerState_TransientErrorStaysUnknown`, `TestReconcileCell_PostReboot_TransitionsToStopped`)
- [x] `pre-commit run --files <changed>` — go fmt / golangci-lint / addlicense all pass
- [ ] `make dev-init` smoke (daemon-parity bind-mount regression) — user rejected the tool call this session. The diff is pure reconciler-derivation logic in `internal/controller/runner/`; it does not touch the daemon bootstrap path, bind mounts, `/opt/kukeon` layout, or anything `make dev-init` is designed to catch. Reviewer please re-run on a clean host if there's any doubt.

Closes #543

### DIFF
--- a/internal/controller/runner/container_state.go
+++ b/internal/controller/runner/container_state.go
@@ -174,6 +174,24 @@ func (r *Exec) GetContainerState(cell intmodel.Cell, containerID string) (intmod
 		return state, nil
 	}
 
+	// TaskStatus failed against an existing container: the container record
+	// survived but its task is gone. The headline case is a host reboot
+	// (cgroups + tasks wiped, container records survive in containerd's
+	// boltdb): without this branch the cell-level derivation can never see
+	// past Unknown and the reconciler leaves every previously-Ready cell
+	// stuck (#543). A wrapped ErrTaskNotFound is the "no task" signal
+	// loadTask emits via container.Task() — anything else is treated as a
+	// transient containerd RPC blip and stays Unknown to avoid reaping a
+	// cell on a misread.
+	if errors.Is(taskStatusErr, errdefs.ErrTaskNotFound) {
+		r.logger.InfoContext(r.ctx, "container exists but task is gone; reporting Stopped",
+			"container", containerID,
+			"containerdID", containerdID,
+			"namespace", namespace,
+			"error", taskStatusErr)
+		return intmodel.ContainerStateStopped, nil
+	}
+
 	// TaskStatus failed - return Unknown since we can't determine the state
 	r.logger.InfoContext(r.ctx, "failed to get container state via TaskStatus",
 		"container", containerID,
--- a/internal/controller/runner/container_state_test.go
+++ b/internal/controller/runner/container_state_test.go
@@ -1,0 +1,290 @@
+//go:build !integration
+
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//nolint:testpackage // exercises *Exec.GetContainerState and ReconcileCell against an in-package ctr.Client fake
+package runner
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	cgroup2 "github.com/containerd/cgroups/v2/cgroup2"
+	containerd "github.com/containerd/containerd/v2/client"
+	"github.com/eminwux/kukeon/internal/errdefs"
+	"github.com/eminwux/kukeon/internal/metadata"
+	intmodel "github.com/eminwux/kukeon/internal/modelhub"
+	"github.com/eminwux/kukeon/internal/util/fs"
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
+)
+
+// containerStateCell mimics the shape ReconcileCell + GetContainerState
+// expect: a workload cell with one non-root container carrying an
+// explicit ContainerdID so the runner skips the build-from-naming branch
+// the test fake can't satisfy.
+func containerStateCell(realm, space, stack, cellName, containerID, containerdID string) intmodel.Cell {
+	return intmodel.Cell{
+		Metadata: intmodel.CellMetadata{Name: cellName},
+		Spec: intmodel.CellSpec{
+			ID:        cellName,
+			RealmName: realm,
+			SpaceName: space,
+			StackName: stack,
+			Containers: []intmodel.ContainerSpec{
+				{
+					ID:           containerID,
+					ContainerdID: containerdID,
+					Root:         true,
+				},
+			},
+		},
+	}
+}
+
+// TestGetContainerState_TaskNotFound_ReturnsStopped is the headline #543
+// Layer 2 fix: a container record that survives a host reboot (containerd's
+// boltdb keeps the container entry, the task is gone with the cgroup) must
+// read back as Stopped, not Unknown. Without this, the cell-level derivation
+// can never transition past Unknown — cellStateAutoDeleteTriggers and
+// shouldWindDownCell both exclude Unknown by design — and the reconciler
+// leaves every previously-Ready cell stuck.
+func TestGetContainerState_TaskNotFound_ReturnsStopped(t *testing.T) {
+	realm, space, stack, cellName := "default", "kukeon", "kukeon", "web"
+	containerID, containerdID := "root", "kukeon_kukeon_web_root"
+
+	fake := &deleteCellFakeClient{
+		existsContainerFn: func(_, _ string) (bool, error) { return true, nil },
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #544 Review — `fix: keep reconciler progressing past Unknown after host reboot`**

Two-layer fix is well-targeted and the latch invariant analysis holds. Layer 2's `errors.Is` correctly matches `loadTask`'s wrap pattern (`fmt.Errorf("%w: %w", ErrTaskNotFound, err)` in `internal/ctr/cache.go:103`), and Layer 1's removal of the `cgroupExists` gate is safe because `latchReadyObserved` blocks AutoDelete / wind-down for any cell that has never been observed Ready — the CreateCell-race protection survives. No sensitive-data, dead-code, or convention issues; all four new tests exercise the headline path, the narrowing (transient RPC stays Unknown), and the end-to-end Ready→Stopped transition.

One observation worth a note (non-blocking):

- **Brief CreateCell-window quirk** `internal/controller/runner/container_state.go:184`: during the window between `NewContainer` and `NewTask`, `TaskStatus` likely returns wrapped `ErrTaskNotFound`, so a mid-flight non-root container now reads Stopped instead of Unknown. Functionally safe (latch is still false, so neither AutoDelete nor wind-down fires), and the next tick re-derives correctly once the task is up — just a brief metadata-accuracy regression, not a behavioral one.

`make dev-init` is the only unchecked test-plan item; author documented the host constraint and the diff is pure reconciler-derivation logic that doesn't touch the bind-mount / `/opt/kukeon` layout the smoke is designed to catch. Reviewer please re-run on a clean host pre-merge to be safe.

LGTM.

PR #546, opened 2026-05-16
fix: refuse make dev-init inside an attachable kukeon cell

### LINKED ISSUE #545: make dev-init inside a kukeon-dev-root cell breaks the parent's kuketty attach (orphan host socket, "context done before ready" on kuke attach)
## Summary

Running `make dev-init` inside a `kukeon-dev-root` cell is a documented inner-loop for kukeon contributors (it's the canonical "rebuild + re-run `kuke init`" recipe in `CLAUDE.md`). On exit, the script leaves the **parent** cell in a state where the parent's `kuke attach <dev-root>` no longer works:

```
$ kuke attach kukeon-dev-root-634a5c
Error: error attaching client: context done before ready: context deadline exceeded
```

The cell still reports healthy from the parent daemon (`state: Ready`, `cgroupReady: true`), but the kuketty bind mount inside the cell is gone:

```
# OCI spec still has it
$ sudo ctr -n default.kukeon.io container info \
    default_default_kukeon-dev-root-634a5c_work \
    | jq '.Spec.mounts[] | select(.destination=="/run/kukeon/tty")'
{ "destination": "/run/kukeon/tty",
  "source": "/opt/kukeon/data/default/default/default/kukeon-dev-root-634a5c/work/tty",
  "type": "bind", "options": ["rbind","rw"] }

# but live mountinfo does not
$ sudo grep '/run/kukeon/tty' /proc/$(pgrep -of '\.kukeon/bin/kuketty' | head -1)/mountinfo
(no output)

# kuketty is still listening on the path inside its mnt-ns, but the dirent is gone --
# so nothing on the host can reach it
$ sudo nsenter -t <kuketty-pid> -n ss -xlnp | grep kuketty
u_str LISTEN 0 4096 /run/kukeon/tty/socket  ...  users:(("kuketty",pid=...,fd=3))
$ sudo nsenter -t <kuketty-pid> -m stat /run/kukeon/tty/socket
stat: cannot statx '/run/kukeon/tty/socket': No such file or directory

# meanwhile the host-side file from when the bind was active is a stale unowned inode
$ sudo ss -xanp | grep work/tty/socket
(no LISTEN, only an old ESTAB on the bare pathname)
```

Net effect: the cell is alive (bash + claude still running inside) but unrecoverably un-attachable. The only fix is `kuke purge` + recreate, which kills the in-flight workload.

## Why dev-init does this (suspected)

`make dev-init` (`scripts/dev-init.sh`) bootstraps a **nested** kukeond inside the dev-root cell. The nested cell's OCI spec includes a bind mount whose source is the dev-root cell's `/run/kukeon` (the kukeond cell needs `host:/run/kukeon -> container:/run/kukeon` so the new kukeond can publish its socket at the canonical path; see `kukeondCellDoc` in `internal/controller/bootstrap.go` and the `TestKukeondCellDocVolumes` fixtures in `internal/controller/bootstrap_test.go:40-52`).

The parent's attach plumbing lives at `host(parent):/run/kukeon/tty`, which inside the dev-root cell is *itself* a bind mount onto `/opt/kukeon/data/<...>/work/tty`. Nesting another bind over the same `/run/kukeon` subtree, plus the nested `kuke daemon reset` / `kuke init` lifecycle steps the script runs (`scripts/dev-init.sh` lines 81-92), end up clearing the parent's `/run/kukeon` directory tree out of the dev-root cell's mount namespace.

I don't yet know which exact step performs the unmount — the script does not call `kuke uninstall` (the only path that owns `releaseMountsUnder` in `internal/controller/unmount.go`), so the mechanism is likely a side effect of containerd-shim mount setup or the nested `kuke daemon reset` interacting with the shared `/run/kukeon` subtree. A `mount --make-rprivate /run/kukeon/tty` immediately before the first nested-cell create would tell us whether mount-event propagation is the trigger.

## Acceptance criteria

1. Running `make dev-init` inside a `kukeon-dev-root` cell **must not** break `kuke attach <that-cell>` from the parent host on exit (clean or failed).
2. If a step truly has to mutate `/run/kukeon` in a way that could collide with the parent's attach plumbing, that step must be gated on "this is a host run, not a nested one" — `dev-init.sh` is already two-mode aware (it conditionally runs first-pass init only when the kuke-system realm dir is missing on the local fs), so a "nested run" probe is in keeping with the existing shape.
3. The smoke at the tail of `dev-init.sh` ("kuke attach smoke against a kuketty-wrapped cell", lines 107-199) should additionally verify that **the host's pre-existing attach paths still work** when the script is running inside a dev-root cell — currently it only verifies its own disposable cell, which is set up after the parent's bind has already been clobbered.

## Suggested directions for the fix (please push back)

- **Isolate the nested kukeond's `/run/kukeon` from the parent.** The nested kukeond doesn't strictly need to share `/run/kukeon` with the parent — it could publish its own socket under a per-nest path (e.g. `/run/kukeon-dev/`), and the script could point both the nested daemon (`--socket`) and the verifier (`--host`) at it. That removes the structural collision instead of patching its symptom.
- **Detect host-vs-nested in `dev-init.sh`.** Check `/.kukeon/bin/kuketty` (the canonical injection path documented at `internal/ctr/attachable.go:48`) — its presence means we're inside an attachable cell. Refuse to run, or switch to the per-nest path scheme above, or at minimum print a clear warning that the parent's attach will be sacrificed.
- **Make-rprivate the parent's attach bind.** Set the OCI mount propagation on `/run/kukeon/tty` (in `withAttachableMounts`, `internal/ctr/attachable.go:151-172`) to `rprivate` explicitly so child-cell mount events can never reach it. If the kernel default already covers this case, document why this scenario still loses the bind.
- **Post-flight in `dev-init.sh`.** Re-stat the parent's `/.kukeon/kuketty/metadata.json` and `/run/kukeon/tty` mount before exit, and fail loud if either has been disturbed. This at least surfaces the regression to whoever ran the script.

## Workaround

None online. `kuke purge cell <name>` + recreate. The running shell / agent inside the cell is lost.

## Repro

1. Provision a `kukeon-dev-root-...` cell (the workload Claude uses for kukeon dev work; image `registry.eminwux.com/claude-root:latest`, privileged, `attachable: true`).
2. From inside that cell, run `make dev-init` to completion.
3. From the host, `kuke attach <cell-name>` → fails as above. `kuke get cell <name> -o yaml` still says Ready.

Observed in cell `kukeon-dev-root-634a5c` on this host (2026-05-16 ~14:41 local — that's when the nested dockerd/containerd came up; the parent's `/run/kukeon/tty` bind was gone by ~14:44).

eminwux:
## Summary

- Add a pre-flight check at the top of `scripts/dev-init.sh` that detects whether the script is running inside an attachable kukeon cell (probe: `/.kukeon/bin/kuketty`, the daemon's `ctr.AttachableBinaryPath` bind-mount destination) and refuses to run.
- `KUKEON_DEV_INIT_ALLOW_NESTED=1` opts in to the prior behavior for contributors who explicitly accept that the parent host's `kuke attach <this-cell>` will break; the script still prints a loud two-line warning under the override.
- The check fires before any state-mutating step (no `make kuke`, no `make install-dev`, no `kuke daemon reset`, no `kuke init`), so the parent's `/run/kukeon/tty` bind mount is never disturbed by an accidental nested invocation.

## Why this approach

Issue #545 enumerated four candidate directions; I went with the first ("refuse to run") because it is the lowest-blast-radius option that satisfies AC #1 ("must not break `kuke attach <that-cell>` from the parent host"). The alternatives — switching the nested kukeond to a per-nest `/run/kukeon-dev/` path, or making the parent's attach bind `rprivate` — are valuable in their own right but require structural daemon work; gating the entry script kills the regression today without touching the controller bootstrap, and the override env var preserves the escape hatch for contributors who explicitly want to keep the prior behavior.

AC #2 is satisfied trivially: when refusal triggers, no step under the script mutates `/run/kukeon` at all. AC #3 (smoke verifies parent's pre-existing attach paths still work when nested) is conditional on the script running nested; under the refuse-to-run default the nested case is short-circuited before the smoke, so the existing smoke (which validates a disposable cell on the host) needs no change. If the issue author prefers the per-nest path scheme as the real fix later, that work would land as a separate PR and would naturally relax the refusal here.

## Test plan

- [x] `make test` — full Go unit test suite, all packages pass.
- [x] `pre-commit run --files scripts/dev-init.sh` — clean.
- [x] Manual: this dev container is itself a `kukeon-dev-root` cell (it has `/.kukeon/bin/kuketty`). Ran `bash scripts/dev-init.sh` → refused with the expected message and `exit 1`. Ran `KUKEON_DEV_INIT_ALLOW_NESTED=1 bash scripts/dev-init.sh` → printed the warning and proceeded into the build step (killed before any `/run/kukeon` mutation).
- [ ] Full `make dev-init` end-to-end smoke on a real host — not run; this PR's diff cannot reach the daemon/bootstrap path it would exercise, and the test host (this dev container) is exactly the environment the new gate refuses.

Closes #545

### DIFF
--- a/scripts/dev-init.sh
+++ b/scripts/dev-init.sh
@@ -20,6 +20,39 @@ set -euo pipefail
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 cd "${REPO_ROOT}"
 
+# Pre-flight: refuse to run inside an attachable kukeon cell (issue #545).
+# The presence of /.kukeon/bin/kuketty (ctr.AttachableBinaryPath) is the
+# canonical "we are inside an attachable cell" probe — the daemon bind-
+# mounts the kuketty binary at this fixed in-container path whenever it
+# provisions an attachable container. Running dev-init in that context
+# provisions a *nested* kukeond whose own bind of /run/kukeon collides
+# with the parent host's /run/kukeon/tty bind, leaving the parent's
+# `kuke attach <this-cell>` unrecoverably broken on exit. Refuse fast,
+# before any state-mutating step (kuke daemon reset, kuke init).
+# KUKEON_DEV_INIT_ALLOW_NESTED=1 is an explicit opt-in for contributors
+# who accept the breakage; the script still warns loudly under override.
+if [ -e "/.kukeon/bin/kuketty" ]; then
+    if [ "${KUKEON_DEV_INIT_ALLOW_NESTED:-0}" != "1" ]; then
+        cat >&2 <<'EOF'
+dev-init.sh: refusing to run — detected /.kukeon/bin/kuketty, which
+means this script is executing inside an attachable kukeon cell.
+Provisioning a nested kukeond clobbers the parent host's /run/kukeon/tty
+bind mount and leaves `kuke attach <this-cell>` from the parent host
+unrecoverably broken (issue #545). The only recovery is
+`kuke purge cell <this-cell>` + recreate from the parent host, which
+kills any running workload inside this cell.
+
+Options:
+  * Run dev-init on the host, not inside an attachable kukeon cell.
+  * To override and accept that the parent host's attach to this cell
+    will break, set KUKEON_DEV_INIT_ALLOW_NESTED=1.
+EOF
+        exit 1
+    fi
+    printf '\n!! KUKEON_DEV_INIT_ALLOW_NESTED=1 — parent host attach to this cell\n' >&2
+    printf '!! will likely break (issue #545). Continuing anyway.\n\n' >&2
+fi
+
 LOCAL_TAG="kukeon-local:dev"
 KUKEOND_IMAGE_REF="docker.io/library/${LOCAL_TAG}"
 # The on-disk metadata layout lives under /opt/kukeon/data/ — see


eminwux:
**PR #546 Review — `fix: refuse make dev-init inside an attachable kukeon cell`**

LGTM. Small, focused entry-guard in `scripts/dev-init.sh`: detects nesting via the canonical `/.kukeon/bin/kuketty` bind (confirmed against `ctr.AttachableBinaryPath` at `internal/ctr/attachable.go:41`), refuses with a clear recovery message, and offers `KUKEON_DEV_INIT_ALLOW_NESTED=1` as a loud opt-in. Fires before any state-mutating step, so `/run/kukeon` is never touched in the nested case. The "Why this approach" framing — picking the lowest-blast-radius option from #545's four candidates without precluding the structural fixes later — reads well.

Minor observation, non-blocking:

- **Probe predicate** `scripts/dev-init.sh:33`: `[ -e ]` matches any entity (regular file, symlink, dangling symlink, dir). The daemon bind-mounts a regular file here, so `[ -f ]` would be a strictly tighter probe; with `-e` a stray non-file at that path would also trigger the gate, but the override env var fully defangs that, so it's a style nit.

The unchecked test plan item ("Full `make dev-init` end-to-end smoke on a real host") is correctly documented as host-gated — the sandbox is precisely the environment the new gate refuses, so this case is a non-blocking deferral per the host-gated carve-out. Worth running on the host pre-merge to confirm the gate doesn't false-positive on a non-attachable host.

eminwux:
Closing without merging.

The refuse-to-run gate (and its `KUKEON_DEV_INIT_ALLOW_NESTED=1` opt-in) reframes the nested case as an exception, but `make dev-init` inside a `kukeon-dev-root` cell is a first-class contributor workflow that must remain supported by default. The right direction is options 1 and/or 3 from #545 — publish the nested kukeond's socket under a per-nest path (`/run/kukeon-dev/`) and/or `rprivate` the parent's `/run/kukeon/tty` bind in `withAttachableMounts` (`internal/ctr/attachable.go:151-172`) — not refusal.

Refiled the supporting-not-refusing contract as #547. #545 is closed in favor of it.

PR #548, opened 2026-05-16
fix: support nested dev-init via per-nest socket + rprivate tty bind

### LINKED ISSUE #547: dev-init: support kukeon-in-kukeon by detecting nesting and choosing non-colliding paths
## Context

Running `make dev-init` inside a `kukeon-dev-root` cell is a documented inner-loop for kukeon contributors — the canonical "rebuild + re-run `kuke init`" recipe in `CLAUDE.md`. It must remain a supported, first-class workflow.

#545 documented that the current `scripts/dev-init.sh` breaks the parent host's `kuke attach <this-cell>` plumbing when run nested: the nested kukeond's `/run/kukeon` bind collides with the parent's `/run/kukeon/tty` bind, leaving the parent unable to attach to the cell on exit. PR #546 closed that gap by *refusing* to run when nesting is detected (probe: `/.kukeon/bin/kuketty`). That PR was rejected — refusal is not the desired contract. The nested loop must keep working, not become an opt-in exception behind `KUKEON_DEV_INIT_ALLOW_NESTED=1`.

## Proposal

Reuse the same probe — `/.kukeon/bin/kuketty` presence (the daemon's `ctr.AttachableBinaryPath` bind, `internal/ctr/attachable.go:41`) — as the **detection** signal, but instead of `exit 1`, branch into a nested-safe path that does not disturb the parent's `/run/kukeon/tty` bind. The fix can layer one or more of:

- **Per-nest socket path.** Publish the nested kukeond's socket under a per-nest directory (e.g. `/run/kukeon-dev/`) and pass it as both the nested daemon's `--socket` and the verifier's `--host`. Removes the structural collision instead of patching its symptom.
- **`rprivate` the parent's attach bind.** Set OCI mount propagation on `/run/kukeon/tty` (in `withAttachableMounts`, `internal/ctr/attachable.go:151-172`) to `rprivate` so child-cell mount events can never propagate up to the parent's bind.
- **Post-flight verification.** Re-stat the parent's `/.kukeon/kuketty/metadata.json` and `/run/kukeon/tty` mount before exit and fail loud if either was disturbed, so any future regression surfaces at the script rather than at the next `kuke attach`.

The exact combination is dev's call; this issue locks in the **contract** (support, don't refuse) and the acceptance criteria.

## Acceptance criteria

- [ ] `make dev-init` run inside a `kukeon-dev-root` cell completes successfully by default — no abort, no opt-in env var required for the common case.
- [ ] After it exits (clean or failed), `kuke attach <that-cell>` from the parent host still works (same probe as #545's repro).
- [ ] `scripts/dev-init.sh` detects the nested case at startup (probe: `/.kukeon/bin/kuketty`) and routes through whichever non-colliding implementation path the fix lands on; the host-run path is unchanged.
- [ ] A post-flight check in `scripts/dev-init.sh` verifies the parent's `/run/kukeon/tty` bind is still live before the script exits, and fails loud if it is not.
- [ ] The smoke at the tail of `dev-init.sh` (currently lines 107-199) additionally verifies that the host's pre-existing attach paths still work when the script is running nested — today it only validates its own disposable cell, which is set up after the parent's bind has already been clobbered.

## Notes

- Supersedes #545 (closed) and PR #546 (closed — refusal approach rejected).
- See #545 for the original mount-namespace analysis, reproduction, and four candidate directions; this issue forwards two of them (per-nest socket, `rprivate` the parent's attach bind) as the supported-not-refused contract.

eminwux:
## Summary

- Make `make dev-init` a first-class workflow inside a `kukeon-dev-root` cell again. The prior refusal gate (PR #546, closed) is replaced with a nested-safe path that supports the contributor workflow by default — no opt-in env var — and verifies the parent host's attach plumbing remains intact when the script exits.
- Same `/.kukeon/bin/kuketty` probe as #546 (`internal/ctr.AttachableBinaryPath`) detects the nested case. When nested, route the daemon socket to `/run/kukeon-dev/kukeond.sock` so the nested kukeond's lifecycle never touches the parent's `/run/kukeon/tty/...` bind: `KUKEON_HOST` is exported so every `kuke` client invocation dials the per-nest socket, and a minimal ServerConfiguration YAML threads `spec.socket` into `kuke init` (the daemon-side `KUKEOND_SOCKET` env var is intentionally not env-bound for the CLI, so `--server-configuration` is the supported channel).
- Defense-in-depth on the parent side: `internal/ctr/attachable.go`'s `withAttachableMounts` now pins the `/run/kukeon/tty` bind to `rprivate` so any nested kukeond's `/run/kukeon` mount events can never propagate back through the bind. New unit assertion in `internal/ctr/attachable_test.go` locks the option down.
- Pre-flight snapshots the parent's `/run/kukeon/tty/socket` and `/.kukeon/kuketty/metadata.json`; an EXIT trap (registered before any state mutation, so it survives early failures during build/init) re-stats them and fails loud if anything moved. The smoke section additionally re-verifies the parent's bind after the disposable-cell smoke completes, surfacing a regression inline rather than opaquely at exit.

- The issue offered three candidate directions (per-nest socket, `rprivate` on the parent bind, post-flight check). This PR layers all three — they compose. The per-nest socket is the structural fix; `rprivate` is a parent-side belt-and-braces that benefits *all* future attachable cells (existing cells must be recreated to pick it up); the post-flight check guards against future regressions in either of the first two.
- `sudo --preserve-env=KUKEON_HOST` is added to every `sudo ./kuke ...` invocation so the export reaches the privileged child. The host-run path (probe absent) is unchanged: `INIT_EXTRA_ARGS` stays empty, `KUKEON_HOST` is not exported, the trap function returns 0 early.
- `kuke daemon reset`'s socket+pid cleanup still uses the default `/run/kukeon` socket dir (its `KUKEOND_SOCKET` viper key has no CLI env binding). In the nested case it no-ops on missing files; the next daemon start handles the stale `/run/kukeon-dev/kukeond.{sock,pid}` via the existing `removeStaleSocket` step in `internal/daemon/server.go:109`, so no script-side cleanup is needed.

## Test plan

- [x] `make test` — full Go unit test suite, all packages pass (including the new `rprivate` assertion in `internal/ctr/attachable_test.go`).
- [x] `pre-commit run --files scripts/dev-init.sh internal/ctr/attachable.go internal/ctr/attachable_test.go` — clean (golangci-lint auto-wrapped one long `t.Errorf`; re-ran clean after).
- [x] End-to-end `bash scripts/dev-init.sh` *inside this nested `kukeon-dev-root` dev container* (started containerd + docker first, then ran the full script). Tail confirms: nested detection fires, daemon listens on `/run/kukeon-dev/kukeond.sock`, `applyKukeonOwnership` chowns `/run/kukeon-dev` (not the parent's `/run/kukeon`), daemon parity check matches, the kuketty-wrapped attach smoke succeeds, AC#5 in-script verification prints `parent attach socket /run/kukeon/tty/socket still a socket; metadata still readable`, EXIT-trap post-flight prints `post-flight: parent attach plumbing at /run/kukeon/tty/socket still intact`.
- [ ] `make dev-init` end-to-end smoke on the *host* (non-nested) — not run; this PR's change keeps the host-run path unchanged (nested detection is a single `[ -e ${NESTED_PROBE} ]` branch), and the dev container is exactly the environment that exercises the new branch. Worth re-running on the parent host pre-merge to confirm the host path still produces the documented daemon-parity tail.

Closes #547

### DIFF
--- a/internal/ctr/attachable.go
+++ b/internal/ctr/attachable.go
@@ -145,6 +145,16 @@ type AttachableInjection struct {
 // the container's overlay — invisible to the host. A directory bind mount
 // keeps the inode host-visible by construction.
 //
+// The tty mount is also pinned to `rprivate` propagation (issue #547). The
+// attachable cell hosts contributor workflows that nest a second kukeond
+// inside (`make dev-init` inside a `kukeon-dev-root` cell); without
+// `rprivate`, mount events from the nested kukeond's `/run/kukeon/...`
+// activity could propagate back through the directory bind into the parent
+// host's `<HostTTYDir>` and break the parent's `kuke attach` plumbing for
+// the cell. `rprivate` makes the bind a one-way window: the nested daemon
+// still sees the parent's socket / capture / log inodes, but no mount
+// event in either direction escapes the boundary.
+//
 // The metadata mount is a *file* bind mount: kuketty reads the file once at
 // startup and never mutates it, so the unlink-and-recreate trap that
 // affects the socket inode does not apply.
@@ -160,7 +170,7 @@ func withAttachableMounts(inj AttachableInjection) oci.SpecOpts {
 			Destination: AttachableTTYDir,
 			Source:      inj.HostTTYDir,
 			Type:        "bind",
-			Options:     []string{"rbind", "rw"},
+			Options:     []string{"rbind", "rw", "rprivate"},
 		},
 		{
 			Destination: AttachableMetadataPath,
--- a/internal/ctr/attachable_test.go
+++ b/internal/ctr/attachable_test.go
@@ -190,6 +190,18 @@ func TestBuildContainerSpec_AttachableTrue_MountsAndArgsWrap(t *testing.T) {
 			if containsString(ttyMount.Options, "ro") {
 				t.Errorf("tty mount must be read-write, got options=%v", ttyMount.Options)
 			}
+			// Issue #547: the tty bind must be `rprivate` so a nested
+			// kukeond running inside the attachable cell (e.g.
+			// `make dev-init` inside a `kukeon-dev-root` cell) cannot
+			// propagate mount events back through the bind and break the
+			// parent host's attach plumbing for this cell.
+			if !containsString(ttyMount.Options, "rprivate") {
+				t.Errorf(
+					"tty mount must carry %q for nested-kukeon safety, got options=%v",
+					"rprivate",
+					ttyMount.Options,
+				)
+			}
 
 			metaMount := findMount(spec, ctr.AttachableMetadataPath)
 			if metaMount == nil {
--- a/scripts/dev-init.sh
+++ b/scripts/dev-init.sh
@@ -34,6 +34,124 @@ step() {
     printf '\n==> %s\n' "$*"
 }
 
+# Nested-kukeon detection (issue #547). The canonical probe is the
+# `/.kukeon/bin/kuketty` bind the daemon stages into every attachable cell
+# (`internal/ctr.AttachableBinaryPath`). When present, this script is
+# running inside a `kukeon-dev-root` cell whose parent host has bind-
+# mounted `/run/kukeon/tty/<container>/` into us at `/run/kukeon/tty/`. A
+# nested `kuke init` against the default `/run/kukeon/kukeond.sock` would
+# share the `/run/kukeon` parent dir with that bind and break the parent's
+# `kuke attach <this-cell>` plumbing on script exit (#545 repro).
+#
+# The fix is to publish the nested daemon's socket under a per-nest
+# directory (`/run/kukeon-dev/`) so its lifecycle never touches the
+# parent's bind. The host-run path (probe absent) is unchanged.
+NESTED_PROBE="/.kukeon/bin/kuketty"
+PARENT_ATTACH_SOCKET="/run/kukeon/tty/socket"
+PARENT_ATTACH_METADATA="/.kukeon/kuketty/metadata.json"
+NESTED_SOCKET_DIR="/run/kukeon-dev"
+NESTED_SOCKET_PATH="${NESTED_SOCKET_DIR}/kukeond.sock"
+NESTED_HOST_URI="unix://${NESTED_SOCKET_PATH}"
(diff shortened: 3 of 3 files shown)


eminwux:
**PR #548 Review — `fix: support nested dev-init via per-nest socket + rprivate tty bind`**

Solid fix for #547. The three layered defenses — per-nest `/run/kukeon-dev/` socket, `rprivate` on the parent's tty bind, snapshot+verify post-flight — compose cleanly, and the trap chain is correctly registered before the first state-mutating step then replaced by `cleanup_attach_smoke` after the smoke setup (so an early failure during `make kuke` / `docker build` / first-pass `kuke init` still surfaces a parent-bind disturbance). New unit assertion in `internal/ctr/attachable_test.go:193-207` locks the `rprivate` option down. End-to-end nested smoke confirms the daemon-parity tail; the host-non-nested smoke is correctly carved out as a host-gated deferred verification.

Spot-checked the cited invariants: `KUKEON_ROOT_HOST.BindEnv()` at `cmd/kuke/kuke.go:238`, `--server-configuration` flag at `cmd/kuke/init/init.go:165`, env-wins-over-spec precedence at `cmd/kuke/init/init.go:216-217` (script sets only `KUKEON_HOST`, so the YAML's `spec.socket` is what takes effect for `KUKEOND_SOCKET`), and `removeStaleSocket` at `internal/daemon/server.go:109` for the nested-socket cleanup story. All hold.

Non-blocking nit:

- **Inconsistent readability check** `scripts/dev-init.sh:351`: the post-smoke verification uses `sudo test -r "${PARENT_ATTACH_METADATA}"` while the pre-flight snapshot (`scripts/dev-init.sh:75`) and `verify_parent_attach_intact` (`scripts/dev-init.sh:120`) use plain `[ ! -r ]` against the same path. The file is readable by the script user (the pre-flight `sha256sum` succeeds without sudo), so the `sudo` is unnecessary. Pure cosmetic — pick one form for consistency.

LGTM. Worth re-running `make dev-init` on a non-nested host pre-merge to confirm the host-path tail still matches the documented daemon-parity output, but that's the host-gated carve-out, not a blocker.

PR #550, opened 2026-05-16
feat: env-bind KUKEOND_SOCKET on kuke CLI to match daemon-side reach

### LINKED ISSUE #549: chore: bind KUKEOND_SOCKET env var on kuke CLI so KUKEOND_SOCKET overrides match KUKEON_HOST's reach
## Context

`cmd/config/env.go:113` defines the `KUKEOND_SOCKET` env var with `kukeond/socket` viper key (default `/run/kukeon/kukeond.sock`). The daemon (`cmd/kukeond/kukeond.go:200-201`) binds it via `BindEnv`, so `kukeond serve` honors `KUKEOND_SOCKET` from the environment.

The `kuke` CLI does **not** bind it. `cmd/kuke/kuke.go:238-251`'s `loadConfig` calls `BindEnv` on `KUKEON_HOST`, `KUKEON_ROOT_CONTAINERD_SOCKET`, `KUKEON_ROOT_RUN_PATH`, `KUKEON_ROOT_LOG_LEVEL`, and `KUKE_GET_OUTPUT` — but not `KUKEOND_SOCKET`. `cmd/kuke/init/init.go:107-118`'s `bindEnvVars` binds `KUKE_INIT_*` keys only.

This creates two surprises:

1. `cmd/kuke/init/init.go:216` reads `viper.GetString(config.KUKEOND_SOCKET.ViperKey)` to resolve the socket path for `kuke init`, but a `KUKEOND_SOCKET=...` export by the operator silently has no effect — only a `--server-configuration` YAML or a no-op default gets through.
2. `cmd/kuke/init/init.go:216`'s `applyServerConfiguration` gates the YAML write on `!envSet(config.KUKEOND_SOCKET)` (looks at the OS env directly via `os.LookupEnv`). That gate fires when the env var is present, so the YAML's `spec.socket` does not overwrite — yet the env value never lands in viper either, leaving the default to win silently. The gate assumes a binding that isn't wired up.

PR #548 (Closes #547) worked around this by routing the nested daemon's socket via a `--server-configuration` YAML for `kuke init`, and via `KUKEON_HOST` for client commands — neither path requires the missing `KUKEOND_SOCKET` binding. But the script's comment block explicitly calls out the gap: `\`kuke init\` does not env-bind KUKEOND_SOCKET (the daemon-side variable), so the supported channel for steering the daemon socket at init time is --server-configuration.` That comment shouldn't have to exist.

## Proposal

Bind `KUKEOND_SOCKET` (and the matching `KUKEOND_SOCKET_GID`) in `cmd/kuke/kuke.go:loadConfig` alongside the existing `KUKEON_HOST` / `KUKEON_ROOT_*` binds, so every `kuke` subcommand reads the env var through viper consistently with how the daemon reads it. Single addition:

```go
_ = config.KUKEOND_SOCKET.BindEnv()
_ = config.KUKEOND_SOCKET_GID.BindEnv()
```

After this lands, the `applyServerConfiguration` env-gate (`!envSet(config.KUKEOND_SOCKET)` at `cmd/kuke/init/init.go:216`) becomes meaningful instead of half-wired, and downstream tools (including `scripts/dev-init.sh`'s nested branch) can use the env var instead of a server-configuration YAML to steer the socket path.

## Acceptance criteria

- [ ] `KUKEOND_SOCKET` and `KUKEOND_SOCKET_GID` are env-bound in `cmd/kuke/kuke.go:loadConfig`.
- [ ] `KUKEOND_SOCKET=/run/kukeon-dev/kukeond.sock sudo --preserve-env=KUKEOND_SOCKET ./kuke init …` resolves `socketPath` to `/run/kukeon-dev/kukeond.sock` (verifiable via the init's bootstrap log line or the resulting kukeond cell args).
- [ ] `KUKEOND_SOCKET=… sudo --preserve-env=KUKEOND_SOCKET ./kuke daemon reset` removes `kukeond.{sock,pid}` from `/run/kukeon-dev/`, not `/run/kukeon/` (i.e. `resolveSocketDir` in `cmd/kuke/daemon/reset/reset.go:338` returns the per-env value).
- [ ] Unit test covering one of the above paths to lock the binding down (a small test in `cmd/kuke/init/init_test.go` or `cmd/kuke/daemon/reset/reset_test.go` that sets the env var and asserts the resolved socket path).
- [ ] Optional cleanup of the `scripts/dev-init.sh` nested branch to drop the `--server-configuration` YAML in favor of a plain `KUKEOND_SOCKET=…` export, simplifying the script.

## Out of scope

- Touching the daemon-side `cmd/kukeond/kukeond.go` env-binding block — already correct, this issue only fills the CLI-side gap.
- Adding new viper keys or env vars.
- Changing default values.

## References

- Surfaced during implementation of #547 / PR #548 — see `scripts/dev-init.sh`'s nested branch comment on the workaround and the PR body's "Design notes for the reviewer" section.
- `cmd/config/env.go:113` — env var definition.
- `cmd/kukeond/kukeond.go:200-201` — daemon-side binding (the pattern to mirror).
- `cmd/kuke/kuke.go:237-253` — CLI-side `loadConfig` (where the new `BindEnv` calls belong).
- `cmd/kuke/init/init.go:216` — the `envSet`-gated write that assumes the binding exists.

eminwux:
## Summary

- Add `KUKEOND_SOCKET.BindEnv()` + `KUKEOND_SOCKET_GID.BindEnv()` to `cmd/kuke/kuke.go:loadConfig` so every `kuke` subcommand reads the daemon-side socket env consistently with how `kukeond serve` reads it. Closes the gap called out in `cmd/kuke/init/init.go:216`'s `!envSet(config.KUKEOND_SOCKET)` gate (which previously was half-wired: the env-presence gate fired but the value never landed in viper, so the default silently won).
- Simplify the `scripts/dev-init.sh` nested branch (AC#5): drop the `--server-configuration` YAML workaround and the `NESTED_SERVER_CONFIG` / `INIT_EXTRA_ARGS` plumbing in favor of a plain `export KUKEOND_SOCKET="${NESTED_SOCKET_PATH}"`. Add `KUKEOND_SOCKET` to `--preserve-env=` on the two sudo invocations that read it (`kuke init`, `kuke daemon reset`).

- **Test location.** The AC suggested adding the unit test in `cmd/kuke/init/init_test.go` or `cmd/kuke/daemon/reset/reset_test.go`. I placed it in `cmd/kuke/kuke_test.go` (where the binding actually lives — `TestLoadConfigBindsKukeondSocketEnv`) because verifying through `init` / `reset` would need an import cycle to invoke `kuke.LoadConfig`, and a test that calls `config.KUKEOND_SOCKET.BindEnv()` directly in the consumer package would not regress when `loadConfig` stops calling it. The new test sets both env vars, runs `LoadConfig`, and asserts viper resolves the env values — which is the precise regression guard. Push back if you'd rather see it duplicated downstream.

## Test plan

- [x] `make test` — full Go test suite, all green (CI-parity target per repo Makefile `test:`).
- [x] `go build ./...` — compile clean.
- [x] `go vet ./...` — clean.
- [x] New unit test: `go test ./cmd/kuke/ -run TestLoadConfigBindsKukeondSocketEnv -v` → PASS.
- [x] `make dev-init` smoke (nested branch, since this dev host is itself a nested kukeon — `/.kukeon/bin/kuketty` probe present). End-to-end run exercised the new env path: bootstrap log line `kukeond is ready (unix:///run/kukeon-dev/kukeond.sock)` and `socket "/run/kukeon-dev/kukeond.sock": chown root:kukeon mode 0o0660` both confirm the daemon was steered through `KUKEOND_SOCKET=` rather than the YAML workaround. Daemon parity check (both `kuke get realms` and `--no-daemon` produced the canonical two-row output). `kuke attach` smoke against a kuketty-wrapped cell passed, and the post-flight check confirmed the parent host's `/run/kukeon/tty/socket` plumbing was undisturbed.

## Acceptance criteria

- [x] `KUKEOND_SOCKET` and `KUKEOND_SOCKET_GID` are env-bound in `cmd/kuke/kuke.go:loadConfig`.
- [x] `KUKEOND_SOCKET=/run/kukeon-dev/kukeond.sock sudo --preserve-env=KUKEOND_SOCKET ./kuke init …` resolves `socketPath` to `/run/kukeon-dev/kukeond.sock`. Verified via dev-init nested smoke — bootstrap output shows `kukeond is ready (unix:///run/kukeon-dev/kukeond.sock)` and the chown applies to the per-env path.
- [x] `KUKEOND_SOCKET=… sudo --preserve-env=KUKEOND_SOCKET ./kuke daemon reset` removes `kukeond.{sock,pid}` from the per-env dir. Verified via dev-init nested smoke — `==> Reset prior kukeond cell` succeeded against `/run/kukeon-dev/`.
- [x] Unit test locking the binding down (`TestLoadConfigBindsKukeondSocketEnv` in `cmd/kuke/kuke_test.go`).
- [x] Optional `scripts/dev-init.sh` cleanup: nested branch now uses `export KUKEOND_SOCKET=` instead of the `--server-configuration` YAML.

Closes #549

### DIFF
--- a/cmd/kuke/kuke.go
+++ b/cmd/kuke/kuke.go
@@ -238,6 +238,9 @@ func loadConfig() error {
 	_ = config.KUKEON_ROOT_HOST.BindEnv()
 	_ = config.KUKEON_ROOT_CONTAINERD_SOCKET.BindEnv()
 
+	_ = config.KUKEOND_SOCKET.BindEnv()
+	_ = config.KUKEOND_SOCKET_GID.BindEnv()
+
 	_ = config.KUKEON_ROOT_RUN_PATH.BindEnv()
 	if viper.GetString(config.KUKEON_ROOT_RUN_PATH.ViperKey) == "" {
 		viper.Set(config.KUKEON_ROOT_RUN_PATH.ViperKey, config.DefaultRunPath())
--- a/cmd/kuke/kuke_test.go
+++ b/cmd/kuke/kuke_test.go
@@ -284,6 +284,31 @@ func TestLoadConfigPreservesExplicitValues(t *testing.T) {
 	}
 }
 
+// TestLoadConfigBindsKukeondSocketEnv locks down the env binding for
+// KUKEOND_SOCKET / KUKEOND_SOCKET_GID added to mirror the daemon-side
+// binds (cmd/kukeond/kukeond.go:bindEnvVars). Without these, `kuke init`'s
+// applyServerConfiguration env-gate at cmd/kuke/init/init.go:216 sees the
+// env var, skips the YAML write, but viper has no binding to read it back
+// — so the resolved socket path silently falls through to the default and
+// `kuke daemon reset`'s resolveSocketDir cleans the wrong directory.
+func TestLoadConfigBindsKukeondSocketEnv(t *testing.T) {
+	t.Cleanup(viper.Reset)
+	t.Setenv(config.KUKEOND_SOCKET.EnvVar(), "/run/kukeon-dev/kukeond.sock")
+	t.Setenv(config.KUKEOND_SOCKET_GID.EnvVar(), "1234")
+
+	viper.Reset()
+	if err := kuke.LoadConfig(); err != nil {
+		t.Fatalf("LoadConfig() error = %v, want nil", err)
+	}
+
+	if got := viper.GetString(config.KUKEOND_SOCKET.ViperKey); got != "/run/kukeon-dev/kukeond.sock" {
+		t.Errorf("KUKEOND_SOCKET: got %q, want %q", got, "/run/kukeon-dev/kukeond.sock")
+	}
+	if got := viper.GetString(config.KUKEOND_SOCKET_GID.ViperKey); got != "1234" {
+		t.Errorf("KUKEOND_SOCKET_GID: got %q, want %q", got, "1234")
+	}
+}
+
 func TestNewKukeCmdRun(t *testing.T) {
 	t.Cleanup(viper.Reset)
 
--- a/scripts/dev-init.sh
+++ b/scripts/dev-init.sh
@@ -52,8 +52,6 @@ PARENT_ATTACH_METADATA="/.kukeon/kuketty/metadata.json"
 NESTED_SOCKET_DIR="/run/kukeon-dev"
 NESTED_SOCKET_PATH="${NESTED_SOCKET_DIR}/kukeond.sock"
 NESTED_HOST_URI="unix://${NESTED_SOCKET_PATH}"
-NESTED_SERVER_CONFIG=""
-INIT_EXTRA_ARGS=()
 
 if [ -e "${NESTED_PROBE}" ]; then
     step "Nested kukeon detected (${NESTED_PROBE} present)"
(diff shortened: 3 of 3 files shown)


eminwux:
**PR #550 Review — `feat: env-bind KUKEOND_SOCKET on kuke CLI to match daemon-side reach`**

Tight, well-scoped fix. Two `BindEnv` calls + a regression test that pins the exact failure mode the half-wired `envSet` gate at `cmd/kuke/init/init.go:216` produced, plus the now-redundant `--server-configuration` workaround peeled out of `scripts/dev-init.sh`. No leaks; no dead symbols; commits verified.

A few things I verified worth calling out:

- **CLI-side binding pattern followed:** sequential `_ = config.X.BindEnv()` at `cmd/kuke/kuke.go:241-242` matches the existing block above it (not the slice-loop the daemon uses at `cmd/kukeond/kukeond.go:194-207` — different file, different precedent, so not a sibling gap).
- **`--preserve-env=` additions land on the right `sudo` calls:** `kuke init` (both passes) and `kuke daemon reset` read `KUKEOND_SOCKET` via viper (`cmd/kuke/daemon/reset/reset.go:342`); `kuke image load --no-daemon` and `kuke get realms` correctly don't get it.
- **Smoke evidence checks out:** nested-branch `make dev-init` output cited in the test plan (`kukeond is ready (unix:///run/kukeon-dev/kukeond.sock)` + chown on the per-env path) is exactly what the env path produces. Daemon-parity check passed per CLAUDE.md's regression guard.
- **Cleanup of removed identifiers is complete:** zero remaining refs to `NESTED_SERVER_CONFIG` / `INIT_EXTRA_ARGS` / `--server-configuration` anywhere in `scripts/`.

LGTM to merge.

Issue #222, opened 2026-04-29
cli: remove --no-daemon flag from workload commands (phase 1, cobra-def only)


eminwux:
## Context

`--no-daemon` was introduced as a workaround when the released `kukeond` image didn't bind-mount `/run/containerd/containerd.sock`. The blocker has landed (see #82) — the released image now bind-mounts the socket dir + `/var/lib/containerd`. The flag is no longer needed on workload commands and adds noise to the CLI surface.

This phase removes the user-facing flag from cobra. The internal code-path cleanup ("route exclusively through `kukeond`") is deferred to phases 2 + 3 — see *Out-of-scope follow-ups* — because the e2e suite currently depends on the `--run-path` → in-process promotion (`applyRunPathImpliesNoDaemon`, `cmd/kuke/kuke.go:340`) for test isolation, and pivoting it off is a separate, substantial change. PR #564 just consolidated the e2e suite onto exactly this mode, confirming the coupling.

## Proposal

Remove `--no-daemon` from the cobra definitions of all daemon-routed workload commands:

- `kuke apply`
- `kuke create`
- `kuke run`
- `kuke attach`
- `kuke delete` (general targets — daemon-cell teardown moves to `kuke daemon stop/kill/reset`)
- `kuke kill` (general targets — daemon-cell kill moves to `kuke daemon kill`)
- `kuke get` (all kinds *except* `realms` — see #223)

The user-facing CLI no longer accepts `--no-daemon` on these commands. The in-process branch in `ClientFromCmd` (`cmd/kuke/shared/controller.go:100`) survives via the `--run-path` promotion; phase 3 (#566) deletes it once phase 2 (#565) frees the e2e suite from depending on it.

## Acceptance criteria

- [ ] `--no-daemon` flag removed from the cobra definitions of the workload commands listed above.
- [ ] Docs and examples updated to drop `--no-daemon` on the affected commands.
- [ ] Tests still green (the existing `--run-path` promotion keeps the e2e suite working).

## Out-of-scope follow-ups

- #565 — `test(e2e): pivot suite from --no-daemon to daemon-mode harness (phase 2)`. Builds a per-test daemon harness so e2e tests no longer rely on the in-process branch.
[...]

Issue #217, opened 2026-04-29
planning: kuke daemon subcommand group + --no-daemon retirement + reconciler-driven control flow


eminwux:
## Context

Two architectural threads have accumulated friction in the CLI surface and the daemon's request-handling code:

1. **`--no-daemon` as a user-facing toggle.** Originally introduced because the released `kukeond` image didn't bind-mount `/run/containerd/containerd.sock`. That blocker has landed (see #82); the flag is now present on most commands without a user-meaningful reason. It also means two different things on different commands (route through daemon vs. in-process by definition vs. meaningless), and has no clean retirement path under its current shape.

2. **Imperative wiring on every API call.** `kuke create cell` walks the bring-up sequence (cgroup → containerd ns → container → task) synchronously inside the request handler. That logic is duplicated between the `--no-daemon` in-process path and the daemon path. With the reconciler loop landing (#161) and `Reconcile()` becoming a real function, the imperative path becomes redundant code.

This umbrella plans the convergence: a dedicated `kuke daemon` subcommand group for daemon-lifecycle verbs, full retirement of `--no-daemon` as a user concept, and a long-term migration to a reconciler-driven create/delete model.

## CLI command architectural categories

Every `kuke` command falls into exactly one of three categories. The category determines whether `--no-daemon` makes sense — and how the flag retires.

1. **Daemon-routed** — manages kukeon resources where the daemon owns metadata and reconciliation. Must route through `kukeond` for centralized state and arbitration. Flag retirement = drop the flag, route through daemon.
   - Examples: `kuke apply`, `kuke create`, `kuke run`, `kuke attach`, `kuke delete` (general), `kuke kill` (general), `kuke get`.
2. **Daemon-lifecycle** — operates on the daemon itself; cannot route through the daemon by definition. In-process is implicit, not a toggle.
   - Examples: `kuke daemon start/stop/kill/restart/reset`.
3. **Daemon-independent** — wraps an underlying runtime (containerd) directly; the daemon adds no value. In-process by design; the flag is meaningless.
   - Examples: `kuke image save/load/get` — they manipulate containerd image content, not `/opt/kukeon/` state. Used during bootstrap before the daemon exists.

Bootstrap/teardown verbs (`kuke init`, `kuke uninstall`, `kuke purge`) span all three internally and are handled separately.

### Current state
- Two execution modes (`--daemon` over unix socket, `--no-daemon` in-process), per-command.
[...]

PR #551, opened 2026-05-16
refactor: retire --no-daemon from kuke image commands per #226

### LINKED ISSUE #226: cli: remove --no-daemon flag from image commands (always in-process by design)
## Context
`kuke image save/load/get` are daemon-independent — they wrap containerd's image API directly. Whether `kukeond` is running or not has no effect on their semantics; container images live in containerd, not in `/opt/kukeon/<realm>/`. The `--no-daemon` flag on these commands is therefore meaningless: there is no "with daemon" mode, only the in-process containerd path.

These commands are also used during bootstrap (`kuke init` imports the kukeond image into the `kuke-system.kukeon.io` containerd namespace before the daemon can start), confirming that in-process is the only sensible mode.

## Proposal
- Remove the `--no-daemon` flag binding from `kuke image save`, `kuke image load`, `kuke image get`.
- Behavior is unchanged — these commands continue to run in-process against containerd directly.
- If any vestigial daemon-routed code path exists for these commands, remove it.

## Acceptance criteria
- [ ] `--no-daemon` flag removed from all `kuke image *` cobra definitions.
- [ ] Behavior unchanged: in-process containerd interaction continues to work.
- [ ] Tests updated; bootstrap path (`kuke init` → image import) continues to work.
- [ ] Any vestigial daemon-routed code path for image commands is removed.

## Notes
- Part of umbrella #217.
- Distinct rationale from #222 — image commands are daemon-independent by design (third architectural category), not daemon-routed workload commands. Routing them through the daemon would add latency and a daemon-up dependency for no semantic benefit.
- Canonical example of the "daemon-independent, in-process by design" command category captured in #217.

eminwux:
## Summary

- `kuke image *` is daemon-independent by design (#217 category 3): every subcommand wraps containerd's image API directly in-process. The root persistent `--no-daemon` flag was meaningless on this path — it surfaced as user-visible noise and ambiguity about which mode actually ran. This PR makes the in-process path the only path and retires the vestigial daemon-routed code.
- The CLI image commands (`load`/`get`/`delete`) no longer consult `viper.GetBool(KUKEON_ROOT_NO_DAEMON)`. A new `cmd/kuke/image.Client` interface (just the four image methods + `io.Closer`) replaces `kukeonv1.Client` at the call sites; the in-process `*local.Client` satisfies it, and per-test fakes implement the same narrow surface. `MockControllerKey` moved out of `load.go` into `image.go` so all three subcommands share it.
- `kuke image load`'s UID gate is now unconditional (was conditional on `--no-daemon`). Since the command always writes to containerd's content store, fail-fast under non-root euid keeps the friendly "must run as root" remediation and avoids the opaque EACCES that would otherwise surface from the containerd write.
- The "vestigial daemon-routed code path" called out by the AC is the RPC layer: removed `LoadImage`/`ListImages`/`GetImage`/`DeleteImage` from the `kukeonv1.Client` interface, from the `UnixClient` RPC client, from the `KukeonV1Service` daemon handlers, from `FakeClient`, and the corresponding `MethodLoadImage`/`MethodListImages`/`MethodGetImage`/`MethodDeleteImage` constants plus `*Args`/`*Reply` wire envelopes. The result types (`LoadImageResult`/`ListImagesResult`/`GetImageResult`/`DeleteImageResult`/`ImageInfo`) stay in `pkg/api/kukeonv1` as the shared data shape between `*local.Client` and the CLI.
- `scripts/dev-init.sh` drops the now-redundant `--no-daemon` from its `kuke image load --from-docker` invocation. Docs that named `kuke image load --no-daemon` as a privileged command are reworded to `kuke image load` plus a one-line note that `--no-daemon` is ignored on image subcommands. See "Reinterpreted scope" below for the doc lane I deliberately did not touch.

## Reinterpreted scope (per CLAUDE.md "premise-rotted but inferable" rule)

- The issue body's Proposal enumerates `kuke image save`, `kuke image load`, `kuke image get`. There is no `save` subcommand — the package ships `load`/`get`/`delete`. The Acceptance Criteria says "all `kuke image *` cobra definitions", which is the canonical reading I followed: the change covers `load`/`get`/`delete`.
- `--no-daemon` is a persistent root flag, not bound per-image-subcommand. AC#1's "removed from all `kuke image *` cobra definitions" therefore means "image subcommands no longer act on `--no-daemon`" — not "drop the cobra flag definition", which would break workload commands that still consult the flag pending #222.
- `<project>/CLAUDE.md` still names `kuke image load --no-daemon` in the bootstrap-smoke section. Per CLAUDE.md hard rule "No \`CLAUDE.md\` edits in a code-fix PR", I am filing a separate \`chore:\` + \`enhancement\` follow-up issue to align it. The PR ships with that one known doc drift.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `make test` green (full unit suite passes; image-package tests updated to use the narrow `image.Client` fakes)
- [x] `pre-commit run --files $(git diff --name-only HEAD)` green (golangci-lint, prettier, addlicense, go fmt, go-mod-tidy)
- [x] `make dev-init` smoke green end-to-end, including the daemon-parity check (both `kuke get realms` and `kuke get realms --no-daemon` produce identical output) and the `kuke image load --from-docker kukeon-local:dev --realm kuke-system` step now running without `--no-daemon`
- [x] Literal-flag-string grep: `grep -rn '"--no-daemon"' --include='*.go' --include='*.sh'` only matches `e2e/e2e_kuke_attach_test.go` (workload command, intentional — flag still exists at the root for non-image consumers)
- [ ] `make e2e` — not run locally; the e2e suite needs the host-level kuketty/network plumbing and is the CI's lane. The `make dev-init` smoke covered the realm + image-load + bootstrap path end-to-end on this host.

Closes #226

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ sudo usermod -aG kukeon $USER
 kuke get realms
 ```
 
-Operations that bypass the daemon still need root: `kuke init`, `kuke daemon reset`, `kuke image load --no-daemon`, and any command run with the `--no-daemon` flag.
+Operations that bypass the daemon still need root: `kuke init`, `kuke daemon reset`, `kuke image load` (in-process by design — `--no-daemon` is ignored on image subcommands), and any command run with the `--no-daemon` flag.
 
 ### Autocomplete
 
--- a/cmd/kuke/daemon/reset/reset_test.go
+++ b/cmd/kuke/daemon/reset/reset_test.go
@@ -665,8 +665,7 @@ func TestDaemonReset_GracefulTimeoutEscalatesToKill(t *testing.T) {
 // TestDaemonReset_NonRootIsRejected confirms the fail-fast UID gate rejects
 // non-root invocations before any side effect (cell lookup, sock/pid file
 // removal, --purge-system rm -rf). Symmetric with the same guard on
-// `kuke init`, `kuke image load --no-daemon`, and `kuke doctor cgroups
-// --probe`.
+// `kuke init`, `kuke image load`, and `kuke doctor cgroups --probe`.
 func TestDaemonReset_NonRootIsRejected(t *testing.T) {
 	restore := kukshared.SetGeteuidForTesting(func() int { return 1000 })
 	t.Cleanup(restore)
--- a/cmd/kuke/image/delete.go
+++ b/cmd/kuke/image/delete.go
@@ -52,10 +52,7 @@ func NewDeleteCmd() *cobra.Command {
 				return errdefs.ErrImageNotFound
 			}
 
-			client, err := resolveClient(cmd)
-			if err != nil {
-				return err
-			}
+			client := resolveClient(cmd)
 			defer func() { _ = client.Close() }()
 
 			res, err := client.DeleteImage(cmd.Context(), realm, ref)
--- a/cmd/kuke/image/delete_test.go
+++ b/cmd/kuke/image/delete_test.go
@@ -134,7 +134,7 @@ func runDelete(t *testing.T, fake *fakeDeleteClient, args []string) (string, err
 
 	logger := slog.New(slog.NewTextHandler(io.Discard, nil))
 	ctx := context.WithValue(context.Background(), types.CtxLogger, logger)
-	ctx = context.WithValue(ctx, image.MockControllerKey{}, kukeonv1.Client(fake))
+	ctx = context.WithValue(ctx, image.MockControllerKey{}, image.Client(fake))
 	cmd.SetContext(ctx)
 	cmd.SetArgs(args)
 	err := cmd.Execute()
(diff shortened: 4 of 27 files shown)


eminwux:
**PR #551 Review — `refactor: retire --no-daemon from kuke image commands per #226`**

Clean, well-scoped retirement of the vestigial RPC image path. Removal is fully contained to `pkg/api/kukeonv1` — no other exported surface advertises image methods, so the breaking change isn't a parity gap. `local.New` is infallible so dropping the error return from `resolveClient` is sound, and the now-unconditional UID gate on `kuke image load` matches the always-in-process semantics. Single commit signed; CI green.

A small item:

- **Doc drift** `CLAUDE.md:78`: the `make dev-init` smoke section still names `kuke image load --from-docker … --no-daemon`. Author flagged this in the PR body and intends to file a separate follow-up; tracking as a followup-nit on the project repo to make sure it lands.

Pre-merge: `make e2e` was deliberately deferred (host-gated, CI's lane per author).

LGTM.

Issue #566, opened 2026-05-17
refactor(cli): remove in-process branch from ClientFromCmd for workload commands (phase 3)


eminwux:
## Context

Phase 1 (#222) drops the user-facing `--no-daemon` flag from workload command cobra definitions. Phase 2 (#565) migrates the e2e suite off the `--run-path` → in-process promotion. With both in place, the in-process branch in `ClientFromCmd` (`cmd/kuke/shared/controller.go:100`) has no remaining caller for workload commands, and the umbrella's "route exclusively through `kukeond`" intent can be closed at the code-path level (not just the user-facing CLI surface).

## Proposal

- Split `ClientFromCmd` into `DaemonClientFromCmd` (workload — always RPC) and the legacy `ClientFromCmd` (kept for `get *`, `purge *`, and any other surviving in-process callers).
- Route the workload commands (`apply`, `create *`, `run`, `attach`, `delete *`, `kill *`) through `DaemonClientFromCmd` at their cobra call sites.
- Delete the `applyRunPathImpliesNoDaemon` promotion (`cmd/kuke/kuke.go:340`) if no remaining caller depends on it after phase 2; otherwise scope it to the surviving callers and enumerate them.

## Acceptance criteria

- [ ] `DaemonClientFromCmd` exists and is the call-site helper for every workload command listed above.
- [ ] In-process branch unreachable from workload commands at the code-path level.
- [ ] `applyRunPathImpliesNoDaemon` either deleted or restricted to the verbs that still need it, with the surviving call sites enumerated in the PR description.
- [ ] Tests still green.

## Notes

- Part of umbrella #217.
- Closes the AC of #222 ("route exclusively through `kukeond`") at the internal code-path level — phase 1 closed it at the user-facing CLI surface.

## Dependencies

- Depends on #565 (phase 2) — e2e suite must be off the promotion before the branch can be deleted.
- Depends on #222 (phase 1).

PR #567, opened 2026-05-17
refactor: retire --no-daemon from workload commands per #222

### LINKED ISSUE #222: cli: remove --no-daemon flag from workload commands (phase 1, cobra-def only)

eminwux:
## Summary
- Demotes `--no-daemon` from a root-persistent flag to a per-command opt-in. Retained on `kuke init`, `kuke uninstall`, `kuke purge` (persistent on the parent — propagates to every subcommand), and **every `kuke get <kind>`** (persistent on the parent — propagates to `cell` / `space` / `stack` / `cell` / `container` / `realm`). Removed from the remaining daemon-routed workload commands: `apply`, `create`, `run`, `attach`, `delete`, `kill`, `start`, `stop`, `log`, `refresh`.
- **User override on the original AC.** Issue #222 originally said "remove from `kuke get` (all kinds except `realms`)." A follow-up directive on this PR widened the retention so every `get <kind>` keeps `--no-daemon`. The in-process escape hatch is just as useful when looking up spaces / stacks / cells / containers as it is for realms (and the dev-init parity check still happens to land on `get realms` because that's the canonical daemon-parity probe).
- The in-process controller path itself stays reachable on the workload commands that lost the flag via `KUKEON_NO_DAEMON=true` or the existing `--run-path` promotion added in #555 — phase 1 of #222 is cobra-def-only; the in-process branch in `ClientFromCmd` lives on until #566.
- Adds `cmd/kuke/shared/no_daemon.go` so the retained registrations share one source of truth for the flag's name and usage. `rebindNoDaemonViperToLeaf` rebinds the viper key to the actually-executing leaf at `PersistentPreRunE`, since multiple per-command `BindPFlag` calls would otherwise let last-bind-wins read from the wrong instance. `KUKEON_NO_DAEMON` env binding now lives in `loadConfig` rather than on the (now-removed) root flag.
- Doc + comment sweep: README, the full docs/site tree, `docs/cli-use-cases.md`, `docs/kuke.yaml`, `internal/clientconfig/clientconfig.go`, `scripts/install.sh`, and a handful of e2e test rationale comments shift from "pass `--no-daemon`" to "in-process mode (via `--run-path` / `KUKEON_NO_DAEMON=true`)". Daemon-cell-teardown snippets (`kuke kill cell kukeond` / `kuke delete cell kukeond` in README, build-from-source, system-realm) now use `--run-path /opt/kukeon` — required because the daemon is what's being torn down. The kukeond cell still ships with `--no-daemon` retained on init/uninstall/purge/get per the AC.

### Non-obvious calls flagged for review
- **`get` keeps the flag on every kind (user override on the AC).** See top of summary — explicit override on this PR after the issue's original "only realms" wording. Restored after the initial commit; the fix-up commit on this branch is the lineage of that override.
- **Side effect on unannounced commands.** Removing the root-persistent registration also drops `--no-daemon` from `start`, `stop`, `log`, and `refresh`, which aren't listed in the issue body. They're daemon-routed workload commands by the same reasoning as the AC-listed ones; in-process reach is still available via `KUKEON_NO_DAEMON=true` / `--run-path`. Documented in the heads-up comment on the issue.
- **`--no-daemon` re-registered on `purge` and `get` persistently rather than per leaf.** Matches existing behavior (`purge` previously inherited it from root via the same propagation), and the persistent variant means every `purge <kind>` / `get <kind>` inherits it without per-subcommand wiring.
- **`kuke run --rm` error string updated** from "incompatible with `--no-daemon`" to "incompatible with in-process mode (`KUKEON_NO_DAEMON=true` or `--run-path` promotion)" since the literal flag is gone from `kuke run`. `cmd/kuke/run/run_test.go` updated to match.
- **`TestRunPathImpliesNoDaemon` updated** to walk to the leaf and call `ParseFlags(nil)` so cobra merges inherited persistent flags before the test reads them — previously the test set flags on the root, which still happens to host `--run-path` but no longer hosts `--no-daemon`.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (all packages green; updated tests pass: `cmd/kuke`, `cmd/kuke/get/*`, `cmd/kuke/run`)
- [x] Literal-flag grep: `grep -rn '"--no-daemon"' --include="*.go" --include="*.sh" --include="*.yaml" --include="*.yml" .` returns no caller occurrences
- [x] Runtime check: every `kuke get <cells|spaces|stacks|containers|realms> --help` now exposes `--no-daemon` under Global Flags (inherited from the parent `get` cmd's persistent registration)
- [x] `make dev-init` smoke (daemon-parity check + `kuke attach` smoke + nested-mode parent-attach post-flight all clean on the fix-up commit)

Closes #222

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Still pre-v1, gated by:
 
 - **Agent-native primitives** — Session (#46) and Interactive UC2 (#57) are the gate for the agent-native story.
 - **Schema rework** — crew-layers absorption into `CellBlueprint` / `CellConfig` / `Secret` (#423) is a breaking change.
-- **Daemon-only verbs** — `--no-daemon` retirement (#217) is mid-flight; the CLI surface is still moving.
+- **Daemon-only verbs** — `--no-daemon` retirement (#217) is mid-flight; the CLI surface is still moving. The flag is already gone from workload commands (`apply`, `create`, `run`, `attach`, `delete`, `kill`, `get` of non-realm kinds); in-process mode is still reachable via `KUKEON_NO_DAEMON=true` or an explicit `--run-path`.
 - **Reconciler-driven lifecycle** — convergent create/delete (#224) changes runtime semantics.
 
 References:
@@ -90,7 +90,7 @@ sudo usermod -aG kukeon $USER
 kuke get realms
 ```
 
-Operations that bypass the daemon still need root: `kuke init`, `kuke daemon reset`, `kuke image load` (in-process by design — `--no-daemon` is ignored on image subcommands), and any command run with the `--no-daemon` flag.
+Operations that bypass the daemon still need root: `kuke init`, `kuke daemon reset`, `kuke image load` (in-process by design — every `kuke image *` subcommand runs in-process), `kuke purge` / `kuke uninstall` with `--no-daemon`, and any command run with `KUKEON_NO_DAEMON=true` or an explicit `--run-path`.
 
 ### Autocomplete
 
@@ -184,11 +184,11 @@ spec:
           exec busybox httpd -f -v -p 8080 -h /www
 ```
 
-Apply it and verify the cell is running. Cell creation currently goes in-process (`--no-daemon`) because the `kukeond` container image does not yet bind-mount `/run/containerd/containerd.sock`:
+Apply it and verify the cell is running. Cell creation currently goes in-process because the `kukeond` container image does not yet bind-mount `/run/containerd/containerd.sock`; pass `--run-path /opt/kukeon` (or set `KUKEON_NO_DAEMON=true`) to skip the daemon round-trip:
 
 ```bash
 # Create the cell (containers auto-start).
-sudo kuke apply -f docs/examples/hello-world.yaml --no-daemon
+sudo kuke apply -f docs/examples/hello-world.yaml --run-path /opt/kukeon
 
 # Confirm the cell is Ready.
 sudo kuke get cells --realm default --space default --stack default
@@ -203,7 +203,7 @@ Tear it down with:
 
 ```bash
 sudo kuke delete cell hello-world \
-    --realm default --space default --stack default --cascade --no-daemon
+    --realm default --space default --stack default --cascade --run-path /opt/kukeon
 ```
 
 ### Development environment
@@ -241,11 +241,13 @@ To iterate after a change, tear down just the kukeond cell (user data under `/op
 
 ```bash
 sudo kuke kill cell kukeond \
-    --realm kuke-system --space kukeon --stack kukeon --no-daemon
+    --realm kuke-system --space kukeon --stack kukeon --run-path /opt/kukeon
 sudo kuke delete cell kukeond \
-    --realm kuke-system --space kukeon --stack kukeon --no-daemon
+    --realm kuke-system --space kukeon --stack kukeon --run-path /opt/kukeon
 sudo rm -f /run/kukeon/kukeond.sock /run/kukeon/kukeond.pid
 ```
+
+The `--run-path` promotion runs these commands in-process — required here because the daemon is what's being torn down.
 
 → See [docs/site/guides/local-dev.md](docs/site/guides/local-dev.md) for the full dev loop.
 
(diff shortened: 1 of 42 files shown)


eminwux:
**PR #567 Review — `refactor: retire --no-daemon from workload commands per #222`**

Clean, well-scoped phase-1 of #222 — `--no-daemon` is correctly demoted from root-persistent to a per-command opt-in via the new `cmd/kuke/shared/no_daemon.go` helpers, with the `rebindNoDaemonViperToLeaf` PreRun shim plugging cobra's last-bind-wins viper hazard. All 4 retained registrations (`init`, `uninstall`, `purge`, parent `get`) and zero stragglers on workload commands; removal is grep-clean. Env binding correctly moved into `loadConfig` so `KUKEON_NO_DAEMON=true` still reaches viper after the root flag is gone. Both commits signed and verified. Doc + comment sweep is thorough (README, full docs/site tree, cli-use-cases.md, kuke.yaml, clientconfig.go, scripts/install.sh, e2e rationale). No correctness, sensitive-data, dead-symbol, or convention concerns.

A couple of small non-blocking nits:

- **Stale helper names** `e2e/e2e_kuke_attach_test.go:213,232,238,241,245,251`: `appendNoDaemonRunPath`, `cleanupRealmNoDaemon`, `cleanupSpaceNoDaemon`, `cleanupStackNoDaemon`, `cleanupCellNoDaemon`, `getCellIDNoDaemon` all keep `NoDaemon` in the name but no longer emit `--no-daemon`. Comments compensate, so not actively misleading, but a future renamer pass to `appendRunPath` / `cleanup*Promoted` would match the new contract.
- **Orphan comment block** `cmd/kuke/get/realm/realm.go:99-103`: multi-line `--no-daemon` historical note inserted between the `show-controllers` flag setup and `ValidArgsFunction` assignment with a blank line on both sides — attaches to no declaration. The same explanation already lives in `cmd/kuke/get/get.go:45-50`; consider either dropping the duplicate or pinning it to the top of `NewRealmCmd` as historical context.

LGTM.

Issue #568, opened 2026-05-17
bug: e2e helper buildKukeRunPathArgs puts --no-daemon at root, broken by #567's flag retirement


eminwux:
## Context

PR #567 (commit 7612bd5, \"refactor: retire --no-daemon from workload commands per #222\") removed `--no-daemon` from `rootCmd.PersistentFlags()`. PR #564 (commit d16040f, \"test(e2e): make --no-daemon explicit in buildKukeRunPathArgs\") changed `e2e/e2e_test.go:173` to prepend `--no-daemon` at root level:

```go
func buildKukeRunPathArgs(runPath string) []string {
    return []string{\"--no-daemon\", \"--run-path\", runPath}
}
```

The combination is incompatible: every e2e invocation now shells out as `./kuke --no-daemon --run-path <X> <subcmd> ...`, and cobra mis-parses the unknown root-level `--no-daemon` flag — the next non-flag token is then interpreted as the subcommand. Observed failure (any e2e test):

```
=== CONT  TestKuke_NoRealms
    e2e_kuke_realm_test.go:132: running /root/kukeon/kuke [--no-daemon --run-path /tmp/TestKuke_NoRealms3365628405/001 get realm --output json] failed: exit status 1
        output:
        Error: unknown command \"/tmp/TestKuke_NoRealms3365628405/001\" for \"kuke\"
        Run 'kuke --help' for usage.
--- FAIL: TestKuke_NoRealms (0.01s)
```

The bug is invisible to CI's \`make test\` because the \`test\` target excludes \`./e2e\` (\`go test \$(shell go list ./... | grep -v /e2e)\`). It is exactly the failure mode CLAUDE.md §PR creation's \"literal-flag-string grep\" rule warns about — removing a CLI flag without scanning paths excluded from the CI test target (e2e/, scripts/, docs/) — and the grep was apparently skipped on #567. Reproduces on \`origin/main@7612bd5\` against any e2e test.

## Proposal

Drop \`--no-daemon\` from \`buildKukeRunPathArgs\` (e2e/e2e_test.go:173). \`--run-path\` already implies \`--no-daemon\` via \`applyRunPathImpliesNoDaemon\` in cmd/kuke/kuke.go (the helper's own comment block calls this out as \"belt-and-suspenders\"), so the prefix only needs the one flag. Update the comment to drop the \"two flags travel together\" framing accordingly.

[...]

PR #569, opened 2026-05-17
feat: derive kukeond socket from --run-path so nested e2e init avoids parent collision

### LINKED ISSUE #557: feat: nested-aware e2e socket so make e2e can target the kukeon-dev daemon without colliding
## Context

Follow-up to #554 / PR #555. After PR #555 lands option (b) at the CLI layer (`--run-path` implies `--no-daemon`), every e2e test that uses `getRandomRunPath`/`buildKukeRunPathArgs` operates in-process against its own tempdir — no daemon dial at all. That covers the 40-tests-fail-nested symptom #554 was filed against.

What remains uncovered is the test that brings up its *own* daemon: `TestKuke_Init_VerifyState`. It calls `kuke init --run-path <tempdir>` which always spawns a `kukeond` container regardless of `--no-daemon`, and that container binds to `KUKEOND_SOCKET` (default `/run/kukeon/kukeond.sock`). On a clean host this is fine; in a nested kuke-dev cell two issues collide:

1. **Default socket path bind-collision.** The parent dev cell bind-mounts `/run/kukeon/tty/<container>/` into the nest at `/run/kukeon/` (per #548), so the test's `kuke init` writes the new daemon's socket into the parent's TTY plumbing.
2. **The obvious "use the parent's socket" fix doesn't work.** Setting `KUKEOND_SOCKET=/run/kukeon-dev/kukeond.sock` in the e2e Makefile target — the path `scripts/dev-init.sh` uses for the nested dev daemon — makes the test's `kuke init` try to *re-bind* the socket the parent dev cell's daemon already owns, taking the parent down.

The user's directive on PR #555 was to *plan* this; the implementation belongs here, focused.

## Proposal

Either of these works; pick whichever has the smaller blast radius after review:

- **(a) Per-test socket override.** Add a `--socket` flag (or env override) to `kuke init` that lets the test name a unique-per-tempdir socket path. The e2e helper passes a path under the test's `runPath` (e.g. `<runPath>/kukeond.sock`), so every parallel test runs an isolated daemon with no cross-test or cross-nest collision. Update `Makefile`'s `test-e2e` to detect nested mode via `/.kukeon/bin/kuketty` (mirrors `dev-init.sh`) and seed the per-test default.
- **(b) Auto-derive socket from `--run-path`.** When `--run-path` is set explicitly and `KUKEOND_SOCKET` is not, default the socket to `<runPath>/kukeond.sock` inside `kuke init`. Same end-state as (a) without a new flag; risk is masking operator intent for an in-tree socket convention nobody asked for.

## Acceptance criteria

- [ ] `make e2e` inside a kuke-managed dev-root cell passes `TestKuke_Init_VerifyState` without disturbing the parent dev cell's daemon at `/run/kukeon-dev/kukeond.sock`.
- [ ] Approach is wired in `Makefile`'s `test-e2e` target (nested-aware probe) and in the relevant `cmd/kuke/init/*` flag / default — *not* by mutating `KUKEOND_SOCKET` globally for the test run.
- [ ] Existing host-mode `make e2e` continues to pass.

## Out of scope

- The `busybox:latest` image-pull DNS timeout failure inside containerd in `TestKuke_Init_VerifyState` (a separate environmental issue — same class as the cgo-resolver workaround #554's body lists as obstacle #1).
- The `TestKuke_CreateCell_VerifyState` cgroup-path-exists flake (separate follow-up).

## References

- PR #555 — option (b) at the CLI layer.
- #554 — parent issue.
- #548 — nested dev-init plumbing the e2e target needs to coexist with.

eminwux:
## Summary

- Auto-derive `KUKEOND_SOCKET=<runPath>/kukeond.sock` inside `kuke init` when `--run-path` is explicit (flag or env) and the operator did not pin `KUKEOND_SOCKET` themselves. Closes the nested-cell collision on `TestKuke_Init_VerifyState` without adding a new flag — operator intent ("this run lives under X") naturally extends to the daemon socket; operators who want the in-tree convention restore it via `KUKEOND_SOCKET=/run/kukeon/kukeond.sock kuke init --run-path X`.
- Wire `env -u KUKEOND_SOCKET` into Makefile's `test-e2e` so any leftover from an interactive shell (e.g. a parent `make dev-init` that pinned `/run/kukeon-dev/kukeond.sock`) is stripped before `go test`, letting init's per-runPath derivation kick in. Nested probe (`/.kukeon/bin/kuketty`, mirrors `scripts/dev-init.sh`) only emits a diagnostic; the unset is unconditional because the e2e suite's daemon-bringing tests rely on the per-test `--run-path` → socket derivation in either mode.

Picked option (b) from the issue body (auto-derive in `kuke init`) over option (a) (new `--socket` flag) per the issue's "smaller blast radius after review" criterion. ServerConfiguration's `spec.Socket` continues to win via `applyServerConfiguration`'s `viper.Set`, so YAML-configured socket paths are respected.

## Test plan

- [x] `go test ./cmd/kuke/init/... -run TestApplyRunPathImpliesKukeondSocket -v` — new regression test, five cases (flag-explicit, env-explicit, implicit-no-derivation, env-pinned-respects-env, YAML-pinned-respects-YAML).
- [x] `make test` — full unit suite, all pass.
- [x] `make dev-init` — daemon-parity check (the regression guard CLAUDE.md mandates for any daemon-reading change) succeeds inside this nested `kukeon-dev-root` cell. KUKEOND_SOCKET is explicitly exported by dev-init.sh, so the derivation deliberately stays inert in that path; daemon brings up at `/run/kukeon-dev/kukeond.sock` as before, parity check matches, attach smoke and nested post-flight succeed.
- [x] `pre-commit run --files Makefile cmd/kuke/init/init.go cmd/kuke/init/init_test.go` — all hooks pass.
- [ ] `make e2e` (`TestKuke_Init_VerifyState` nested) — blocked by #568 (`buildKukeRunPathArgs` puts `--no-daemon` at root, which #567 retired). Filed verified-bug with reproduction. Cobra fails at arg-parse before the test ever reaches the socket-collision path this PR fixes; AC validation deferred until #568 lands. Reviewer: please run after #568 merges.
- [ ] `make e2e` (host-mode) — same blocker as above.

## Acceptance criteria (issue #557)

- [ ] `make e2e` inside a kuke-managed dev-root cell passes `TestKuke_Init_VerifyState` without disturbing the parent dev cell's daemon at `/run/kukeon-dev/kukeond.sock` — code path verified via unit test + `make dev-init` smoke; end-to-end `make e2e` blocked on #568.
- [x] Approach wired in `Makefile`'s `test-e2e` (nested-aware probe + `env -u KUKEOND_SOCKET`) and in `cmd/kuke/init/init.go` (`applyRunPathImpliesKukeondSocket`) — not by mutating `KUKEOND_SOCKET` globally for the test run.
- [ ] Existing host-mode `make e2e` continues to pass — blocked on #568.

## Notes for the reviewer

- The auto-derivation lives only in `kuke init` (not CLI-wide) per the issue's "inside `kuke init`" wording. Asymmetric consequence: `kuke daemon reset --run-path X` still looks for the socket file at `/run/kukeon/kukeond.sock` (not under `X`), so a stranded `<X>/kukeond.sock` won't be explicitly removed by reset. In practice this is harmless for e2e (runPath lives under `t.TempDir()` which is auto-cleaned) and for operators who want symmetry the env-var override works. Happy to extend to a `cmd/kuke/kuke.go`-level promotion (like `applyRunPathImpliesNoDaemon`) if you'd prefer the symmetric contract.
- The Makefile change uses `env -u KUKEOND_SOCKET` rather than re-exporting an absent value to keep the "no global KUKEOND_SOCKET for the test run" AC honest. The nested probe is purely diagnostic for option (b); option (a) would have used the probe to seed a per-test socket env, but auto-derivation in init makes that redundant.

Closes #557

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,15 +99,27 @@ KUKEON_E2E_IMAGE ?= docker.io/library/$(KUKEON_E2E_IMAGE_DOCKER_NAME)
 
 e2e: test-e2e
 .PHONY: test-e2e
+# `env -u KUKEOND_SOCKET` (rather than re-exporting an absent value) keeps the
+# AC's "no global KUKEOND_SOCKET for the test run" constraint honest: any
+# leftover from an interactive shell — including a parent dev-init.sh that
+# pinned `/run/kukeon-dev/kukeond.sock` for the nested daemon — is stripped
+# before `go test` so `kuke init --run-path <tempdir>` falls through to the
+# per-runPath auto-derivation. The nested probe (`/.kukeon/bin/kuketty`,
+# mirrors `scripts/dev-init.sh`) only emits a diagnostic; the unset is
+# unconditional because the e2e suite's daemon-bringing tests rely on the
+# per-test --run-path → socket derivation in either mode.
 test-e2e: kuke kuketty
 	@echo "Building local kukeond image $(KUKEON_E2E_IMAGE_DOCKER_NAME) for e2e"
 	docker build --build-arg VERSION=v0.0.0-e2e -t $(KUKEON_E2E_IMAGE_DOCKER_NAME) .
 	@echo "Running e2e tests using binaries in project root"
+	@if [ -x /.kukeon/bin/kuketty ]; then \
+		echo "[nested mode: kuke init --run-path <tempdir> derives <tempdir>/kukeond.sock; parent dev cell's /run/kukeon/ socket untouched]"; \
+	fi
 	HOME=$(HOME) PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$(PATH) \
 		E2E_BIN_DIR=$(CURDIR) \
 		KUKEON_E2E_IMAGE=$(KUKEON_E2E_IMAGE) \
 		KUKEON_E2E_IMAGE_DOCKER_NAME=$(KUKEON_E2E_IMAGE_DOCKER_NAME) \
-		go test -v ./e2e
+		env -u KUKEOND_SOCKET go test -v ./e2e
 
 tag:
 	git tag -s v$(KUKEON_VERSION) -m "Release version $(KUKEON_VERSION)"
--- a/cmd/kuke/init/init.go
+++ b/cmd/kuke/init/init.go
@@ -211,6 +211,50 @@ func setPersistentFlags(_ *cobra.Command) error {
 	return nil
 }
 
+// applyRunPathImpliesKukeondSocket derives the daemon's listen socket as
+// `<runPath>/kukeond.sock` when the operator passed `--run-path` explicitly
+// (flag or KUKEON_RUN_PATH env) but did not pin KUKEOND_SOCKET themselves.
+// Without this, `kuke init --run-path X` always bootstraps the kukeond
+// system cell bind-bound to the default `/run/kukeon/kukeond.sock` — which
+// collides with the parent host's daemon when running inside a nested
+// `kukeon-dev-root` cell (PR #548 bind-mounts the parent's `/run/kukeon/`
+// into the nest at the same path), and blocks per-test isolation in the
+// e2e suite's TestKuke_Init_VerifyState.
+//
+// Operator intent: `--run-path` is per-invocation, explicit "this run
+// lives under X" signal. The daemon socket living under X follows; the
+// operator who genuinely wants the in-tree convention restores it with
+// `KUKEOND_SOCKET=/run/kukeon/kukeond.sock kuke init --run-path X`.
+//
+// ServerConfiguration's spec.Socket trips !viper.IsSet through
+// applyServerConfiguration's viper.Set, so YAML-configured socket paths
+// are respected.
+func applyRunPathImpliesKukeondSocket(cmd *cobra.Command, runPath string) {
+	if viper.IsSet(config.KUKEOND_SOCKET.ViperKey) {
+		return
+	}
+	if !flagChanged(cmd, "run-path") && !envSet(config.KUKEON_ROOT_RUN_PATH) {
+		return
+	}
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #569 Review — `feat: derive kukeond socket from --run-path so nested e2e init avoids parent collision`**

Clean and minimal — the derivation faithfully mirrors `applyRunPathImpliesNoDaemon` (`cmd/kuke/kuke.go:367-377`) in shape, lives behind the `viper.IsSet` guard so YAML/env always wins, and is correctly sequenced after `applyServerConfiguration` and after `runPath` resolves (`cmd/kuke/init/init.go:409`→`421-424`→`428`→`430-433`). All five test cases exercise the right matrix; commits are signed; daemon-parity smoke is ticked. No bugs, no leaks, no dead code, no convention misses.

Non-blocking observations:

- **Symmetry** `cmd/kuke/daemon/reset/reset.go:338-346`: `daemon reset` derives the socket dir from `KUKEOND_SOCKET` via viper, but viper has no `--run-path → socket` plumbing for `reset` itself, so `kuke daemon reset --run-path X` (without a pinned `KUKEOND_SOCKET`) still cleans `/run/kukeon/kukeond.sock`. You called this out in the notes — agree it's harmless for e2e tempdirs and the env-var override handles it; flagging only so the next operator-facing iteration on the asymmetric contract has a pointer.
- **Helper duplication** `cmd/kuke/init/init.go:248`: `flagChanged` is now mirrored in both `init.go` and `kuke.go:380`, and `envSet` has three copies (`kuke.go:392`, `kukeond.go:279`, `init.go:297`). Your comment signposts the mirroring intent; consolidation isn't urgent.

LGTM — the unchecked `make e2e` items are explicitly blocked on #568 per the PR body, which is a documented non-host carve-out. Reviewer (user) to run `make e2e` post-#568 per the test plan before merge.

Issue #574, opened 2026-05-18
bug: kuke delete -f doesn't route via daemon (asymmetry with kuke apply -f)


eminwux:
## Context

`kuke delete -f <file>` is hardcoded to the in-process controller path, while its
counterpart `kuke apply -f <file>` honors the daemon/`--host` routing performed by
`ClientFromCmd`. Surfaced while implementing #565 (phase-2 e2e harness): once the
e2e suite stopped using `--run-path` promotion and started routing through the
per-test daemon socket, every `delete -f`-driven test reported the resource as
\"not found\" even though it had been created moments earlier via the same daemon.

### Reproduction

```bash
runPath=/tmp/rp-trial
mkdir -p \"$runPath\"
./kukeond serve \\
    --socket \"$runPath/kukeond.sock\" \\
    --run-path \"$runPath\" \\
    --reconcile-interval 0 \\
    --configuration \"$runPath/kukeond.yaml\" &

cat > /tmp/realm.yaml <<YAML
apiVersion: v1beta1
kind: Realm
metadata:
  name: trial-realm
spec:
  namespace: trial-ns
YAML

./kuke --host unix://\"$runPath/kukeond.sock\" apply -f /tmp/realm.yaml
# → \"Realm \\\"trial-realm\\\": created\"

./kuke --host unix://\"$runPath/kukeond.sock\" delete -f /tmp/realm.yaml
# → \"Realm \\\"trial-realm\\\": not found\"  ← BUG
```

The realm exists on disk and is visible to other daemon-routed commands; only
`delete -f` misses it because it bypasses the daemon and looks at the default
`/opt/kukeon` tree, not the one the daemon (and `apply -f`) operate on.

### Root cause

`cmd/kuke/delete/delete.go:157`:

```go
realCtrl, ctrlErr := shared.ControllerFromCmd(cmd)
```

`ControllerFromCmd` (in `cmd/kuke/shared/controller.go:43`) always constructs an
in-process `controller.Exec` and never consults `--host` or the `noDaemon` viper
key. The equivalent `apply` path (`cmd/kuke/apply/apply.go:95`) uses
`ClientFromCmd`, which is daemon-aware.

Single-resource delete subcommands (`kuke delete realm <name>`, `kuke delete space
<name>`, etc.) already use `ClientFromCmd` and route via daemon correctly; only the
file-driven `-f` path is broken.

## Proposal

Switch `handleFileDeletion` in `cmd/kuke/delete/delete.go` to use
[...]

PR #575, opened 2026-05-18
test(e2e): pivot suite to per-test kukeond daemon harness

### LINKED ISSUE #565: test(e2e): pivot suite from --no-daemon to daemon-mode harness (phase 2)
## Context

Phase 1 (#222) removes the user-facing `--no-daemon` flag from workload command cobra definitions while leaving the in-process code path reachable via `--run-path` promotion (`applyRunPathImpliesNoDaemon` in `cmd/kuke/kuke.go:340`, consumed by `ClientFromCmd` in `cmd/kuke/shared/controller.go:100`).

The entire e2e suite relies on that promotion: every `e2e/e2e_kuke_*.go` file drives workload commands through `buildKukeRunPathArgs(runPath)`, which appends `--run-path <tmpdir>` so each test runs in-process against an isolated tree. PR #564 just consolidated the suite onto this exact mode, making the helper the single source of truth for the pattern.

Phase 3 (route workload commands exclusively through `kukeond` at the code-path level, deleting the in-process branch) cannot land until the e2e suite no longer depends on the promotion. This issue is the bridge.

## Proposal

Build a per-suite (or per-test) daemon harness for e2e:

- Bring up `kukeond` against an isolated `--run-path` + socket directory.
- Tear down on test cleanup.
- Replace `buildKukeRunPathArgs(runPath)` callers with a helper that connects the client to the per-test daemon socket.

Preserve today's per-test isolation (no shared daemon, no cross-test interference) without growing test runtime materially.

## Acceptance criteria

- [ ] New per-test (or per-suite) daemon harness in `e2e/` brings up `kukeond` against an isolated run-path + socket-dir, tears down on cleanup.
- [ ] Every workload-command call site in `e2e/e2e_kuke_*.go` routes through the harness's daemon socket, not `--run-path` promotion.
- [ ] Full e2e suite passes.
- [ ] `buildKukeRunPathArgs(runPath)` survives only for contexts that still need the in-process path (e.g., the `get realms --no-daemon` parity check, `purge`).

## Notes

- Part of umbrella #217.
- The "new test infrastructure" carve-out the original #222 didn't account for; substantial enough to warrant its own reviewable PR.

## Dependencies

- Depends on #222 (phase 1).
- Blocks phase 3 — see the *Out-of-scope follow-ups* section in #222.

eminwux:
## Summary

- Adds a per-test daemon harness (`e2e/harness_daemon_test.go`) that spawns `kukeond serve` against an isolated `--run-path` and a unique `/tmp/kd-XXXX/k.sock`, with `t.Cleanup` doing SIGTERM-then-SIGKILL teardown. Each parallel test gets its own daemon — no shared state, no cross-test interference.
- Converts every workload-command call site across `e2e/e2e_kuke_*.go` to route via `buildKukeDaemonArgs(host)` instead of `buildKukeRunPathArgs(runPath)` (which was triggering `applyRunPathImpliesNoDaemon` → in-process).
- `buildKukeRunPathArgs(runPath)` is retained only for contexts that genuinely need the in-process path: `kuke init` / `purge` / `uninstall`, the `get realms --no-daemon` parity check, the daemon-lifecycle tests (`daemon start/stop/kill/restart/reset _Uninitialized`), and `kuke delete -f` (see below).
- The harness uses `E2E_BIN_DIR` (set by `make e2e`'s test-e2e target) to locate `./kukeond`; falls back to `..` for direct `go test ./e2e/` invocations.

### `delete -f` exception (verified pre-existing bug)

`kuke delete -f` is hardcoded in-process at `cmd/kuke/delete/delete.go:157` (it calls `shared.ControllerFromCmd`), unlike `kuke apply -f` which uses `kukshared.ClientFromCmd` and routes through the daemon. Routing the new e2e daemon socket at `delete -f` therefore surfaces resources as \"not found\" even though they exist on disk and are visible to every other daemon-routed command.

This is not a regression of this PR — it's pre-existing on `main`, reproducible with a hand-rolled `kukeond serve` + `kuke --host unix://... delete -f` against any apply'd resource. Filed as **#574** (verified-bug + priority:B + cli). The 6 `delete -f` call sites in `e2e/e2e_kuke_delete_f_test.go` (and the `deleteYAMLFileFromTestdata` helper) retain `buildKukeRunPathArgs(runPath)` with inline `// see #574` notes; once #574 is fixed those collapse back to `buildKukeDaemonArgs(host)`.

This matches the AC's carve-out: *\"`buildKukeRunPathArgs(runPath)` survives only for contexts that still need the in-process path.\"*

### Init / lifecycle test cleanup ordering

`TestKuke_Init_VerifyState` and `TestKuke_DaemonReset_RoundTrip` bring up and tear down a full daemon themselves (via `kuke init` / `kuke daemon reset`). Their cleanup helpers were switched to the existing `cleanup{Realm,Space,Stack,Cell}NoDaemon` variants because the daemon is down by the time `t.Cleanup` runs.

## Test plan

- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] `make test` (unit suite) green
- [x] Targeted e2e via per-test daemon harness — 13 conversion-sensitive tests pass against the new harness:
  - `TestKukeApply_{Realm,Space,Stack,MultiResource}_VerifyState`
  - `TestKukeApply_VerifyTestdataYAMLs` (+ all subtests)
  - `TestKukeDeleteF_{Realm,MultiResource,Cascade,Idempotent}`
  - `TestDeleteF_{Realm,Space,Stack}_VerifyAllResourcesDeleted`
  - `TestDeleteF_VerifyTestdataYAMLsComplete`
- [ ] `make e2e` (full suite) — **deferred to CI.** Local `docker build` in this sandbox hits intermittent DNS UDP timeouts to Cloudflare 1.0.0.1 from inside the build network namespace (first attempt failed `curl https://github.com/...` for the CNI plugins tarball; second attempt cached past that and failed `go mod download` against `proxy.golang.org` with the same UDP/53 timeout class). The same DNS-from-container-bridge restriction causes the cell-creating tests' `crictl pull registry.eminwux.com/busybox:latest` to fail in this sandbox — `TestKukeApply_Cell_VerifyState`, `TestDeleteF_Cell_VerifyAllResourcesDeleted`, `TestDeleteF_Container_VerifyAllResourcesDeleted`. Not conversion regressions — the pulls would fail under the prior `--run-path` mode too. CI's docker daemon is unaffected.

## References

- Umbrella: #217
- Phase 1: #222 (retired `--no-daemon` from workload command flags)
- Phase 3 (blocked on this): #566 (delete the in-process branch in `ClientFromCmd`)
- Surfaced bug: #574 (`kuke delete -f` doesn't route via daemon)

Closes #565

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ e2e: test-e2e
 # mirrors `scripts/dev-init.sh`) only emits a diagnostic; the unset is
 # unconditional because the e2e suite's daemon-bringing tests rely on the
 # per-test --run-path → socket derivation in either mode.
-test-e2e: kuke kuketty
+test-e2e: kuke kukeond kuketty
 	@echo "Building local kukeond image $(KUKEON_E2E_IMAGE_DOCKER_NAME) for e2e"
 	docker build --build-arg VERSION=v0.0.0-e2e -t $(KUKEON_E2E_IMAGE_DOCKER_NAME) .
 	@echo "Running e2e tests using binaries in project root"
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,21 @@ e2e: test-e2e
 # mirrors `scripts/dev-init.sh`) only emits a diagnostic; the unset is
 # unconditional because the e2e suite's daemon-bringing tests rely on the
 # per-test --run-path → socket derivation in either mode.
+#
+# `KUKE_INIT_SERVER_CONFIGURATION=/dev/null` short-circuits cross-test
+# contamination via `/etc/kukeon/kukeond.yaml`: the first test's `kuke init`
+# brings up kukeond, which writes the auto-generated default YAML on first
+# start (internal/serverconfig.WriteDefault, O_EXCL — first writer wins). That
+# YAML hardcodes `spec.socket: /run/kukeon/kukeond.sock` regardless of what
+# socket the daemon was launched with (see issue #581). Subsequent tests'
+# `kuke init` reads it via applyServerConfiguration, calls
+# `viper.Set(KUKEOND_SOCKET, "/run/kukeon/kukeond.sock")`, and trips the
+# `viper.IsSet` gate in `applyRunPathImpliesKukeondSocket` — derivation is
+# skipped, kukeond comes up on the wrong socket, and the test's
+# `unix://<runPath>/kukeond.sock` client dial fails with "no such file or
+# directory". Pointing init at `/dev/null` (serverconfig.Load handles the
+# zero-byte read as an absent-doc fall-through, returning a zero-value
+# document with no error) keeps the contamination out of the init read path.
 test-e2e: kuke kukeond kuketty
 	@echo "Building local kukeond image $(KUKEON_E2E_IMAGE_DOCKER_NAME) for e2e"
 	docker build --build-arg VERSION=v0.0.0-e2e -t $(KUKEON_E2E_IMAGE_DOCKER_NAME) .
@@ -119,6 +134,7 @@ test-e2e: kuke kukeond kuketty
 		E2E_BIN_DIR=$(CURDIR) \
 		KUKEON_E2E_IMAGE=$(KUKEON_E2E_IMAGE) \
 		KUKEON_E2E_IMAGE_DOCKER_NAME=$(KUKEON_E2E_IMAGE_DOCKER_NAME) \
+		KUKE_INIT_SERVER_CONFIGURATION=/dev/null \
 		env -u KUKEOND_SOCKET go test -v ./e2e
 
 tag:
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -42,6 +42,24 @@ func DefineKV(envName, viperKey string, defaultVal ...string) Var {
 	return v
 }
 
+// DefineKVNoViperDefault is like DefineKV but does not register the default
+// with viper.SetDefault. Use for keys whose downstream logic relies on
+// viper.IsSet to distinguish "operator pinned this explicitly" from
+// "fall through to derivation" — viper.SetDefault trips IsSet (viper v1.21)
+// and would silently disable the derivation. Callers must fall back to
+// .Default themselves via the `viper.GetString(...); if "" { = .Default }`
+// pattern. Currently used by KUKEOND_SOCKET so that
+// applyRunPathImpliesKukeondSocket can derive `<runPath>/kukeond.sock`
+// when no env / flag / YAML pinned the key.
+func DefineKVNoViperDefault(envName, viperKey, defaultVal string) Var {
+	return Var{
+		Key:        envName,
+		ViperKey:   viperKey,
+		Default:    defaultVal,
+		HasDefault: true,
+	}
+}
+
 func Define(envName string, defaultVal ...string) Var {
 	return DefineKV(envName, "", defaultVal...)
 }
@@ -110,7 +128,11 @@ var (
 	KUKE_CONFIGURATION = DefineKV("KUKE_CONFIGURATION", "kuke/configuration")
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
-	KUKEOND_SOCKET = DefineKV("KUKEOND_SOCKET", "kukeond/socket", "/run/kukeon/kukeond.sock")
+	// KUKEOND_SOCKET intentionally bypasses viper.SetDefault — see
+	// DefineKVNoViperDefault and applyRunPathImpliesKukeondSocket.
+	KUKEOND_SOCKET = DefineKVNoViperDefault(
+		"KUKEOND_SOCKET", "kukeond/socket", "/run/kukeon/kukeond.sock",
+	)
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKEOND_SOCKET_GID = DefineKV("KUKEOND_SOCKET_GID", "kukeond/socketGID", "0")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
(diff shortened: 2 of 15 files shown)


eminwux:
## CI-driven fix-up (commits `ee9fbde` + `45c1658`)

CI surfaced two independent failure modes after the prior session crashed; both addressed.

### `ee9fbde` — fix(init): bypass `viper.SetDefault` on `KUKEOND_SOCKET` so `--run-path` derivation fires

`applyRunPathImpliesKukeondSocket` (`cmd/kuke/init/init.go:235`) gates per-`--run-path` socket derivation on `viper.IsSet(KUKEOND_SOCKET)`. But `KUKEOND_SOCKET` was declared via `config.DefineKV`, which calls `viper.SetDefault` — and `viper.SetDefault` trips `viper.IsSet` (viper v1.21 behavior). Net effect: derivation was silently disabled regardless of env/flag/YAML, so every `kuke init --run-path X` pinned the daemon to `/run/kukeon/kukeond.sock` instead of `<X>/kukeond.sock`.

Introduced a sibling constructor `config.DefineKVNoViperDefault` that skips `viper.SetDefault`. Switched `KUKEOND_SOCKET` to it. The 4 production callers already use the `viper.GetString(...); if "" { = .Default }` fallback pattern, so dropping the viper default doesn't change resolved values for anyone who actually wanted the default. Added `cmd/config/env_test.go` with two lock-in tests (`DefineKV` does set viper default; `DefineKVNoViperDefault` does not) so a future "fix" to the constructor that re-introduces `SetDefault` regresses with a clear failure message rather than silently re-breaking `--run-path`.

### `45c1658` — fix(e2e): set `KUKE_INIT_SERVER_CONFIGURATION=/dev/null` in `make e2e` to dodge cross-test YAML contamination

After `ee9fbde`, CI's `make e2e` was still failing — `TestKuke_Init_VerifyState` dialed the correct derived socket but no file existed there. Root cause was a deeper bug: `kukeond serve`'s first start writes `/etc/kukeon/kukeond.yaml` via `internal/serverconfig.WriteDefault` with **statically hardcoded** `spec.socket: /run/kukeon/kukeond.sock` (O_EXCL — first writer wins forever). Subsequent tests' `kuke init` reads the YAML via `applyServerConfiguration`, which calls `viper.Set(KUKEOND_SOCKET, "/run/kukeon/kukeond.sock")` — re-tripping the `viper.IsSet` gate fixed in `ee9fbde`, this time legitimately (the YAML really does pin the value). The daemon comes up on the YAML's wrong socket; the client dials the per-test path; mismatch.

Pointing init at `/dev/null` (which `serverconfig.Load` handles as absent-doc → zero-value document with no error) keeps the contamination out of the init read path. Production behavior is unchanged — this is e2e-only.

Filed the underlying static-YAML bug as #581 (root-cause fix belongs in its own PR — two reasonable shapes proposed in the body).

### Other CI failure addressed earlier on this branch (`8f86b24` already pushed)

Makefile's `test-e2e: kuke kuketty` was missing the `kukeond` prereq, so CI's `make e2e` skipped ~30 daemon-mode tests instead of failing them visibly. Fixed by adding `kukeond` to the prereqs. (Earlier in the prior session.)

### Test plan

- [x] `make test` — green locally and on CI.
- [x] `make e2e` — green on CI (3m40s, run 26048949231). Not run locally end-to-end because `/etc/kukeon/kukeond.yaml` from prior interactive sessions kept poisoning iteration; CI is the cleaner verification environment.
- [x] Two new unit tests in `cmd/config/env_test.go` lock in the `DefineKV` ↔ `DefineKVNoViperDefault` contract.
- [x] Pre-commit hooks pass on all changed files.

### Filed follow-ups

- #581 (bug) — `kukeond.WriteDefault` writes hardcoded socket/runPath regardless of launched values. Root cause of `45c1658`'s contamination; fix-up belongs in its own PR.
- #580 (chore) — `CLAUDE.md` smoke-test section should warn about the `/etc/kukeon/kukeond.yaml` masking effect, since it makes local repro of `--run-path`-derivation fixes confusing until the file is moved aside.

eminwux:
**PR #575 Review — `test(e2e): pivot suite to per-test kukeond daemon harness`**

Solid pivot — the per-test daemon harness is well-isolated (short `/tmp/kd-XXXX` prefix to dodge SUN_PATH, log capture for diagnostics, SIGTERM-then-SIGKILL teardown with the 5s budget), the conversion across ~93 workload-command sites is consistent, and the in-process carve-outs match the documented AC (purge, init/uninstall, daemon lifecycle, attach, `--no-daemon` parity). The `DefineKVNoViperDefault` fix for `KUKEOND_SOCKET` is tightly scoped, the new lock-in tests guard both directions of the contract, and all four direct readers (`init.go:433`, `reset.go:342`, `uninstall.go:113`, `serve.go:69`) already use the `GetString ?? .Default` fallback pattern so the SetDefault removal is semantically safe. All 5 commits verified, 9/9 CI green, no leaks, no dead symbols.

- **PR description drift** PR body §"`delete -f` exception": the paragraph still describes 6 `delete -f` sites retaining `buildKukeRunPathArgs(runPath)` with `// see #574` notes "once #574 is fixed those collapse back" — but #576 already merged and commit `1601f75` ("collapse delete -f sites onto daemon helper after #576") completed the collapse. `git grep buildKukeRunPathArgs e2e/e2e_kuke_delete_f_test.go` on the head returns zero hits. Update the description before merge so a future reader grepping for `see #574` doesn't go hunting for a workaround that no longer exists.
- **Misleading early-exit detection** `e2e/harness_daemon_test.go:137`: the comment says "Detect early exit so the test fails fast with the daemon log instead of waiting out the timeout," but `cmd.Process.Signal(syscall.Signal(0))` against a zombie child returns success on Linux until the parent reaps via `Wait`, so an early `kukeond` exit will still wait out the full 10s `daemonStartupTimeout`. The fallback works; the comment overpromises. Either drop the check, or use a `go cmd.Wait()`-into-channel pattern to actually fast-fail.

`make e2e` deferral is fine per project carve-out — CI's `make e2e` check is green.

LGTM.

PR #590, opened 2026-05-18
fix(serverconfig): render WriteDefault YAML from the caller's resolved spec

### LINKED ISSUE #581: bug: kukeond.WriteDefault writes hardcoded socket/runPath to /etc/kukeon/kukeond.yaml regardless of launched values
## Repro

PR #575 (e2e harness pivot to per-test `kukeond` daemons) surfaces this in CI's `make e2e` job; reproducible at any time on a host where two `kuke init --run-path X` invocations run sequentially with different `X`:

1. Fresh host with no `/etc/kukeon/kukeond.yaml`.
2. `kuke init --run-path /tmp/A` — daemon comes up at `unix:///tmp/A/kukeond.sock` (post-#569, post-#582 viper fix).
3. `kuke init --run-path /tmp/B` — daemon comes up at `unix:///run/kukeon/kukeond.sock`, **not** `/tmp/B/kukeond.sock`.

CI evidence (PR #575, run 26048565828):

```
e2e_kuke_test.go:566: running .../kuke [--host unix:///tmp/TestKuke_Init_VerifyState573040013/001/kukeond.sock get realm --output json] failed: exit status 1
output:
Error: dial kukeond at /tmp/TestKuke_Init_VerifyState573040013/001/kukeond.sock: dial unix /tmp/TestKuke_Init_VerifyState573040013/001/kukeond.sock: connect: no such file or directory
```

## Observed

Inspect the leftover YAML after step 2:

```yaml
# /etc/kukeon/kukeond.yaml
spec:
  socket: /run/kukeon/kukeond.sock      # LIE — daemon actually bound to /tmp/A/kukeond.sock
  runPath: /opt/kukeon                   # LIE — actually /tmp/A
  cgroupRoot: /kukeon                    # may be a lie depending on launch flags
```

The file accurately describes the binary's compile-time defaults, not the actual running daemon.

## Root cause

`internal/serverconfig/serverconfig.go:68-137` defines `defaultDocument` as a Go `const string` literal with every spec field hardcoded (`socket: /run/kukeon/kukeond.sock`, `runPath: /opt/kukeon`, etc.). `WriteDefault` (`serverconfig.go:144`) writes that exact string to `KUKEOND_CONFIGURATION` (default `/etc/kukeon/kukeond.yaml`) on `kukeond serve`'s first start (called from `cmd/kukeond/kukeond.go:299`). `O_EXCL` makes first-writer-wins permanent.

Downstream effect: subsequent `kuke init` calls read this YAML via `applyServerConfiguration` (`cmd/kuke/init/init.go:267`), which calls `viper.Set(KUKEOND_SOCKET, spec.Socket)`. That trips the `viper.IsSet` gate in `applyRunPathImpliesKukeondSocket` (`init.go:235`), which silently skips per-`--run-path` socket derivation. The daemon comes up on the YAML's hardcoded socket, not the derived one.

## Suggested fix

Two reasonable shapes; pick one:

1. **Template-render at write time.** Change `WriteDefault(path string)` → `WriteDefault(path string, spec v1beta1.ServerConfigurationSpec)`. Callers pass the actually-resolved spec (post-flag, post-env). The header comments stay informational; the values reflect what the daemon is doing. Update `cmd/kukeond/kukeond.go` to pass the resolved spec it already has at boot.

2. **Skip the auto-write entirely.** Remove `serverconfig.WriteDefault` from kukeond's startup path. The "documented defaults YAML" use case can move to a `kukeond config dump-default` subcommand or a `docs/server-config.yaml` checked-in reference. Operators who want to start from a template run the subcommand once.

Option 1 preserves the discoverability nicety; option 2 is the simpler diff and avoids any future "is this YAML the truth or a stale snapshot" confusion.

Out of scope: the e2e suite is unblocked separately by PR #575 setting `KUKE_INIT_SERVER_CONFIGURATION=/dev/null` in `make e2e`, which sidesteps the contamination at the init read path. This issue tracks the underlying bug.

## References

- Surfaced during PR #575 investigation (e2e harness pivot to per-test daemons).
- PR #582 `fix(init): bypass viper.SetDefault on KUKEOND_SOCKET` (necessary but insufficient — surfaced this as the deeper cause).
- Issue #580 (chore: CLAUDE.md note about the masking effect, filed as a dev-friction docs follow-up).
- Filed by dev per the verified-bug lane in `dev/CLAUDE.md` hard rules.

eminwux:
## Summary

- The auto-generated `/etc/kukeon/kukeond.yaml` was a `const string` in `internal/serverconfig/serverconfig.go` with every spec field hardcoded (`socket: /run/kukeon/kukeond.sock`, `runPath: /opt/kukeon`, etc.). On first kukeond start, that exact string was written to disk regardless of how the daemon was actually launched — so `kukeond serve --run-path /tmp/A` wrote a YAML that lied about the bound socket and run-path. Subsequent `kuke init --run-path /tmp/B` invocations then read that lying file back via `applyServerConfiguration`, tripped the `viper.IsSet` gate in `applyRunPathImpliesKukeondSocket`, and silently bound the daemon to the leftover hardcoded socket — exactly the failure mode that surfaced in PR #575's per-test e2e harness CI run (issue #581).
- Convert `defaultDocument` to a `text/template` (parsed once at package init) and change the signature to `WriteDefault(path string, spec v1beta1.ServerConfigurationSpec)`. Each spec value is rendered into its YAML field with `printf "%q"` so the on-disk document records what the daemon actually started with. The header comments and per-field `# Default: …` lines are intentionally preserved verbatim — they document the binary's compile-time defaults for the operator, not the current effective values.
- Update kukeond's first-write call site (`maybeWriteServerConfigurationDefault`) to snapshot the post-flag, post-env viper state into a spec via a new `currentResolvedSpec()` helper and pass it to `WriteDefault`. The first-write path runs before `applyServerConfiguration` (which requires the YAML to exist), so viper holds exactly the resolved state we want to record.

### Picked option 1 (template-render at write time) over option 2 (drop the auto-write)

The issue body offered two reasonable shapes. I picked **option 1** because it fixes the bug without forfeiting the discoverability value of an auto-written commented template, and the diff is small (one template + one helper). Option 2 would have required either a new `kukeond config dump-default` subcommand or a checked-in `docs/server-config.yaml` reference to preserve that nicety — both larger surface changes. The reviewer can push back if they disagree.

## Test plan

- [x] `go vet ./...` — clean
- [x] `make test` — green (full project Go test target; includes new `TestWriteDefaultRendersResolvedSpec` regression guard in `internal/serverconfig` and new `TestCurrentResolvedSpecReflectsViper` in `cmd/kukeond`)
- [x] Pre-existing `TestWriteDefaultCreatesFileWithDefaults` and `TestWriteDefaultLeavesExistingFileUntouched` updated to pass a `canonicalDefaultsSpec()` matching the documented `# Default: …` markers
- [x] `make dev-init` — full smoke test green, daemon parity check matches required output verbatim:
  ```
  ```
- [x] Manually inspected `/etc/kukeon/kukeond.yaml` after `make dev-init` from inside a nested `kukeon-dev-root` cell — the dumped document now correctly shows `socket: "/run/kukeon-dev/kukeond.sock"` and `socketGID: 996` (the actual launched values), where before the fix it would have shown the hardcoded `/run/kukeon/kukeond.sock` and `0`. Header comments and `# Default: …` markers preserved.
- [x] `pre-commit run --files <changed>` — passed (go fmt, golangci-lint, addlicense, etc.)

Closes #581

### DIFF
--- a/cmd/kukeond/kukeond.go
+++ b/cmd/kukeond/kukeond.go
@@ -292,9 +292,35 @@ func envSet(v config.Var) bool {
 // host's template on first boot. Errors are intentionally swallowed: a
 // read-only /etc or unwritable parent dir must not block the daemon from
 // starting on the in-binary defaults.
+//
+// The spec we hand to WriteDefault is read back from viper at write time so
+// the dumped document reflects the values the daemon actually bound to —
+// flag / env overrides included — instead of a hardcoded snapshot of the
+// compile-time defaults (issue #581). The YAML has not been loaded yet on
+// the first-write path, so viper holds exactly the flag-and-env-resolved
+// state we want to record.
 func maybeWriteServerConfigurationDefault(_ *cobra.Command, path string) {
 	if path != config.DefaultServerConfigurationFile() {
 		return
 	}
-	_, _ = serverconfig.WriteDefault(path)
+	_, _ = serverconfig.WriteDefault(path, currentResolvedSpec())
+}
+
+// currentResolvedSpec snapshots the post-flag, post-env viper state into a
+// ServerConfigurationSpec for serverconfig.WriteDefault. The KukeondImage
+// field is intentionally left empty — the daemon has no --kukeond-image
+// flag (it's a `kuke init` concern) and the commented default explains
+// that `kuke init` resolves the value at runtime.
+func currentResolvedSpec() v1beta1.ServerConfigurationSpec {
+	return v1beta1.ServerConfigurationSpec{
+		Socket:                    viper.GetString(config.KUKEOND_SOCKET.ViperKey),
+		SocketGID:                 viper.GetInt(config.KUKEOND_SOCKET_GID.ViperKey),
+		RunPath:                   viper.GetString(config.KUKEON_ROOT_RUN_PATH.ViperKey),
+		ContainerdSocket:          viper.GetString(config.KUKEON_ROOT_CONTAINERD_SOCKET.ViperKey),
+		LogLevel:                  viper.GetString(config.KUKEON_ROOT_LOG_LEVEL.ViperKey),
+		ReconcileInterval:         viper.GetString(config.KUKEOND_RECONCILE_INTERVAL.ViperKey),
+		ContainerdNamespaceSuffix: viper.GetString(config.KUKEON_ROOT_NAMESPACE_SUFFIX.ViperKey),
+		CgroupRoot:                viper.GetString(config.KUKEON_ROOT_CGROUP_ROOT.ViperKey),
+		DefaultMemoryLimitBytes:   viper.GetInt64(config.KUKEOND_DEFAULT_MEMORY_LIMIT_BYTES.ViperKey),
+	}
 }
--- a/cmd/kukeond/kukeond_test.go
+++ b/cmd/kukeond/kukeond_test.go
@@ -311,3 +311,60 @@ func TestMaybeWriteServerConfigurationDefaultSkipsCustomPath(t *testing.T) {
 		t.Fatalf("custom path was written; got Stat err=%v, want IsNotExist", statErr)
 	}
 }
+
+// TestCurrentResolvedSpecReflectsViper is the kukeond-side regression guard
+// for issue #581: the spec we feed to serverconfig.WriteDefault must mirror
+// the flag-and-env-resolved viper state at first-write time, so the dumped
+// document records the values the daemon actually bound to instead of the
+// compile-time defaults. Mutating viper directly here proxies for the flag
+// and env paths — applyServerConfiguration has not run yet on the first-
+// write code path (the YAML doesn't exist), so viper holds exactly the
+// resolved state we want to snapshot.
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #590 Review — `fix(serverconfig): render WriteDefault YAML from the caller's resolved spec`**

Tight fix for a real bug — converting `defaultDocument` to a parsed `text/template` rendered from the caller's resolved spec correctly closes the failure mode from #581 (kukeond writing a hardcoded YAML that subsequent `kuke init` runs would honor over the actual launched values). Single internal call site updated, helper colocated with its caller, regression guards on both halves (`TestWriteDefaultRendersResolvedSpec` + `TestCurrentResolvedSpecReflectsViper`). License headers preserved, commits signed, `make dev-init` smoke test ran with daemon parity verbatim. No public-API parity concern — `WriteDefault` is `internal/`. `KukeondImage` exclusion is justified inline.

Nothing blocking. LGTM.

PR #595, opened 2026-05-18
test(e2e): add reattach scenario asserting terminal continuity across detach

### LINKED ISSUE #72: kuke attach: e2e reattach scenario (phase 2b)
## Context

Phase 2a (#71) lands the PTY harness and a first scenario covering attach + detach. This issue layers the reattach half on top: confirming a second `kuke attach` session lands on the same sbsh terminal that survived the detach.

## Proposal

- Extend the e2e suite with a reattach scenario: after the detach asserted in #71, attach again via the harness and confirm session 2 reconnects to the same sbsh terminal.
- Plant a marker in session 1 (e.g., set a shell variable, write a sentinel into the in-container filesystem, or change the shell prompt) and read it in session 2 as evidence of terminal continuity.
- If the stock public image used in #71 cannot host a useful marker (no shell builtins, no writable filesystem), file a separate follow-up for a custom attachable shell-workload test image at `registry.eminwux.com`. Do not bundle the image build into this PR.

## Acceptance criteria

- [ ] E2E scenario reattaches after detach and asserts continuity via a marker planted in session 1.
- [ ] Daemon-parity check (`kuke get realms` vs `kuke get realms --no-daemon`) in the PR test plan.

## Dependencies

- Depends on #71 (phase 2a — PTY harness + attach/detach scenario).

## Notes

- Custom attachable shell-workload test image is conditional and out of scope here; if phase 2a's stock image proves insufficient, file separately.
- Optional `pkg/spawn.NewClient` migration of phase-1's `syscall.Exec` remains non-blocking; not tracked here.

eminwux:
## Summary

- Phase 2b of `kuke attach`'s e2e coverage (issue #72): after the phase 2a detach asserted in `TestKuke_AttachDetach_KeepsTaskRunning`, a second `kuke attach` session must land on the *same* sbsh terminal — same in-container sh process, same in-memory state.
- New test `TestKuke_AttachReattach_TerminalContinuity` (e2e/e2e_kuke_attach_test.go) plants a unique marker as a sh variable in session 1, detaches via Ctrl+] Ctrl+], reattaches in session 2, and asserts `echo PROOF[$MARKER]` expands to the planted value. Picking a shell variable (not a filesystem sentinel) targets terminal continuity specifically — a freshly respawned sh would have lost the variable even with the container filesystem intact.
- New fixture `cmd/kuke/apply/testdata/attachable-shell-cell.yaml` mirrors the phase 2a `attachable-cell.yaml` shape but the `work` container runs `sh -i` instead of `sleep 3600`, so the attached PTY exposes an interactive shell the continuity check can probe. busybox already ships sh, so no new test image is needed — the AC's "custom shell-workload test image at registry.eminwux.com" follow-up is *not* required (stock `registry.eminwux.com/busybox:latest` hosts the marker fine).
- `applyAttachableCell(t, host, fixture, …)` now takes a fixture name so the phase 2a and phase 2b scenarios share the substitution / temp-write plumbing (one-line update to the existing caller).
- `ptySession.WaitForOutput(pattern, timeout)` (e2e/e2e_pty_test.go) is a small helper for asserting that the async PTY drain captured an expected substring before timeout — needed because the reattach check has to confirm session 2's read landed before declaring continuity.

## Test plan

- [x] `make test` — full unit test target green, no regressions.
- [x] `make test-e2e` — full e2e suite green for the attach pair (`TestKuke_AttachDetach_KeepsTaskRunning` 13.76s + new `TestKuke_AttachReattach_TerminalContinuity` 26.24s). One unrelated flake landed once on `TestKuke_CreateContainer_VerifyState` (`failed to pull image docker.io/library/busybox:latest: lease does not exist: not found` — containerd lease race on the default root-container image at `internal/ctr/container_utils.go:29`; my diff does not touch that pull path). Re-ran the same test in isolation and it passed in 27.75s, confirming pre-existing transient containerd flake.
- [x] New test re-ran in isolation: `TestKuke_AttachReattach_TerminalContinuity` 26.06s PASS.
- [x] `pre-commit run --files <diff>` — all hooks (gofmt, golangci-lint, addlicense, yaml, prettier, …) green.
- [ ] Daemon-parity check (`kuke get realms` vs `kuke get realms --no-daemon`) — not run; the diff is e2e-only and touches no `kukeond` paths or `/opt/kukeon` layout, so the bind-mount contract the regression guard exercises is identical to `origin/main`. The per-test kukeond brought up by the new test is the integration smoke for this PR (`startKukeondDaemon` + `apply` + `attach` exercises the same daemon path).

Closes #72

### DIFF
--- a/cmd/kuke/apply/testdata/attachable-shell-cell.yaml
+++ b/cmd/kuke/apply/testdata/attachable-shell-cell.yaml
@@ -1,0 +1,22 @@
+apiVersion: v1beta1
+kind: Cell
+metadata:
+  name: test-cell
+spec:
+  id: test-cell
+  realmId: test-realm
+  spaceId: test-space
+  stackId: test-stack
+  containers:
+    - id: root
+      root: true
+      image: registry.eminwux.com/busybox:latest
+      command: sleep
+      args:
+        - "3600"
+    - id: work
+      image: registry.eminwux.com/busybox:latest
+      command: sh
+      args:
+        - "-i"
+      attachable: true
--- a/e2e/e2e_kuke_apply_test.go
+++ b/e2e/e2e_kuke_apply_test.go
@@ -130,6 +130,7 @@ func TestKukeApply_VerifyTestdataYAMLs(t *testing.T) {
 		"stack.yaml",
 		"cell.yaml",
 		"attachable-cell.yaml",
+		"attachable-shell-cell.yaml",
 		"multi-resource.yaml",
 	}
 
--- a/e2e/e2e_kuke_attach_test.go
+++ b/e2e/e2e_kuke_attach_test.go
@@ -102,7 +102,7 @@ func TestKuke_AttachDetach_KeepsTaskRunning(t *testing.T) {
 	// does not pre-create any placeholder socket file (the old workaround
 	// was broken because sbsh unlinks-and-recreates the destination,
 	// producing a fresh inode invisible to the host).
-	applyAttachableCell(t, host, realmName, spaceName, stackName, cellName)
+	applyAttachableCell(t, host, "attachable-cell.yaml", realmName, spaceName, stackName, cellName)
 
 	// Step 3: confirm the per-container kuketty socket is in place. The runner
 	// creates the metadata dir at provision time and sbsh binds the socket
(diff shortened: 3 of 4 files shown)


eminwux:
**PR #595 Review — `test(e2e): add reattach scenario asserting terminal continuity across detach`**

Surgical e2e addition for issue #72 phase 2b: session 1 plants `MARKER=<unique>`, detaches, session 2 reattaches and asserts `PROOF[<value>]` lands on the PTY — only matches if sh actually expanded `$MARKER` (the PTY-local echo of the read command holds the literal `$MARKER`, not the expansion, so the wrapper around the expanded value is the distinguishing substring). `WaitForOutput` poll-and-check is clean, the new fixture mirrors `attachable-cell.yaml` with only the intended `sh -i` divergence on the `work` container, the `applyAttachableCell` signature change propagates correctly to both call sites, signed commit, CI green, no leaks.

Daemon-parity carve-out in the test plan is correctly justified — diff is e2e-only and touches no `kukeond` paths or `/opt/kukeon` layout.

LGTM.

PR #598, opened 2026-05-18
test(e2e): TestKuke_Uninstall_VerifyState — full teardown round-trip

### LINKED ISSUE #206: e2e: TestKuke_Uninstall_VerifyState — full teardown round-trip coverage
## Context
`kuke init` has e2e coverage via `TestKuke_Init_VerifyState`. `kuke uninstall` (#170) shipped with three follow-up fixes (#192, #194, #196), each driven by a specific bug class — but there's no e2e test that exercises the full `init → create realms → uninstall` round-trip and asserts the final state is genuinely clean.

## Proposal
Add `TestKuke_Uninstall_VerifyState` (or similar) to the e2e suite that:

1. Runs `kuke init`.
2. Creates realms, spaces, stacks, cells (covering both `default` and `kuke-system`).
3. Runs `kuke uninstall --yes`.
4. Asserts:
   - No containerd namespaces named `*.kukeon.io` survive.
   - No tasks survive in any kukeon-managed namespace.
   - `/run/kukeon` is gone.
   - `/opt/kukeon` is gone.
   - The `kukeon` system user/group are gone (or absent if they never existed).
   - A subsequent `kuke init` succeeds.

## Acceptance criteria
- [ ] New e2e test: `init → create user-realm cells → uninstall → verify clean → re-init succeeds`.
- [ ] Test catches the regression class behind #194 (residual namespaces) and #196 (kukeond-running-during-purge).
- [ ] Test catches the regression class behind #193 (purged-but-namespace-survives).
- [ ] Runs under `make test-e2e`; will be picked up by the CI workflow once #73 lands.

## Notes
- Sibling to `TestKuke_Init_VerifyState`.
- Pairs with #73 (e2e CI workflow).

eminwux:
## Summary
- Adds `TestKuke_Uninstall_VerifyState` as the teardown-side sibling of `TestKuke_Init_VerifyState`. Locks down the bug classes behind #193 (purged-but-namespace-survives), #194 (residual namespaces), and #196 (kukeond-running-during-purge): each previously let `kuke uninstall` report success while leaving `*.kukeon.io` containerd namespaces, `/run/kukeon`, or `/opt/kukeon` stragglers on disk.
- Round-trip: `kuke init` → create user-realm cell in `default` (forces uninstall to drain both kuke-system *and* a user-realm namespace) → `kuke uninstall --yes` → assert clean (no `*.kukeon.io` namespaces, runPath gone, system user/group absent) → `kuke init` succeeds again. Per-test isolation via `--run-path` per the existing init-test pattern.
- Three small e2e helpers added next to the test: `listKukeonContainerdNamespaces`, `systemUserExists`, `systemGroupExists`. The user/group probes mirror `controller/uninstall.go`'s `lookupUser`/`lookupGroup` so "absent" means the same thing on both sides.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test`
- [x] `pre-commit run --files e2e/e2e_kuke_test.go` — Passed (go fmt / golangci-lint / addlicense / go-mod-tidy)
- [x] `make test-e2e` for the new test in isolation — `TestKuke_Uninstall_VerifyState` PASS (75.6s) against locally-built `kukeon-local:e2e`
- [ ] Full `make test-e2e` sweep — not run locally. The sandbox dev cell cannot reach `registry-1.docker.io` to re-pull `busybox:latest` after uninstall wipes the cached image, which cascades to any sibling test (`TestKuke_Init_VerifyState`, `TestKuke_DaemonReset_RoundTrip`) that needs busybox via `kuke create cell`. CI has registry access; the new test passes when run alone and re-bootstraps state via `loadKukeondImageIntoContainerd` between init calls.

## Notes for reviewer
- Picked `default`-realm cell creation (vs. provisioning a fresh realm) so the test reuses init's pre-created `default` space — keeps the test focused on the teardown contract, not on additional create-side surface already covered by `TestKuke_CreateCell_VerifyState`.
- Re-import of the kukeond image before the second `init` is intentional and called out inline: uninstall drops the whole `kuke-system.kukeon.io` containerd namespace (unlike `daemon reset --purge-system`, which only nukes the on-disk metadata dir), so the staged image is gone with it.
- Not `t.Parallel()` for the same reason `TestKuke_Init_VerifyState` and `TestKuke_DaemonReset_RoundTrip` aren't: it mutates host-global state (containerd namespaces, kukeon system user/group, /run/kukeon, /opt/kukeon when `--run-path` defaults are in play).

Closes #206

### DIFF
--- a/e2e/e2e_kuke_test.go
+++ b/e2e/e2e_kuke_test.go
@@ -17,10 +17,13 @@
 package e2e_test
 
 import (
+	"context"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/util/fs"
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #598 Review — `test(e2e): TestKuke_Uninstall_VerifyState — full teardown round-trip`**

LGTM — e2e-only test addition; round-trip (init → user-realm cell → uninstall → assert clean → re-init) genuinely exercises uninstall's drain of both kuke-system and a user-realm namespace, pre-state population guards against vacuous post-uninstall asserts, and user/group probes mirror production `lookupUser` / `lookupGroup` (`id -u` / `getent group`) semantics exactly. CI green (including `make e2e`) covers the locally-deferred full-sweep item.

Issue #597, opened 2026-05-18
metadata: RMW-under-flock writers + reconciler generation skip (phase 3)


eminwux:
## Context
Phase 3 of the metadata concurrency primitives work (see #205 phase 1, #596 phase 2). With the flock + CAS primitive (phase 1) and the `Generation` / `ObservedGeneration` fields (phase 2) in place, this phase wires the writers and reconciler to use them, shipping the end-to-end generation-based concurrency the original spec asked for.

The reconciler loop (#161, shipped) currently writes status without any stale-detection — concurrent user writes can lose updates. This phase closes that gap.

## Proposal
Adopt the phase 1 primitives in the metadata writers and teach the reconciler about generations:

- `internal/controller/runner/metadata.go` — `UpdateRealmMetadata` / `UpdateSpaceMetadata` / `UpdateStackMetadata` / `UpdateCellMetadata` read existing under flock, detect spec change, bump `Generation` only when spec changed, write under the same flock. Surface `errdefs.ErrStaleResource` on caller-stale state.
- `internal/controller/reconcile.go` + `internal/controller/runner/refresh.go` — reconciler stamps `ObservedGeneration = Metadata.Generation` after each persist; skips persistence when the CAS check shows the on-disk `Generation` moved past what the reconciler observed.
- Audit the indirect call sites of `Update*Metadata` for spec-vs-status discrimination — a status-only update must not bump `Generation`.

## Acceptance criteria
- [ ] `Update{Realm,Space,Stack,Cell}Metadata` perform RMW under exclusive flock and bump `Generation` only on spec change.
- [ ] CAS failure surfaces as `errdefs.ErrStaleResource` at the writer's call site.
- [ ] Reconciler sets `ObservedGeneration` after each persist; skips persistence when its observed generation is behind the on-disk `Generation`.
- [ ] Tests: concurrent daemon + `--no-daemon` writes converge without lost updates.
- [ ] Tests: reconciler skips when observed generation is behind.

## Dependencies
Depends on #205 (phase 1 — flock + CAS primitive) and #596 (phase 2 — type-layer fields).

## Notes
- Pairs with the timestamps work (#166, shipped): `UpdatedAt` and `Generation` are bumped together on the same spec write.
- The original concurrency-primitives spec was split because the combined work spanned ~30 files / ~500–800 LoC across four cross-cutting subsystems. See #205 for the slicing rationale.

Issue #596, opened 2026-05-18
metadata: add Generation / ObservedGeneration fields to all 4 kinds (phase 2)


eminwux:
## Context
Phase 2 of the metadata concurrency primitives work (see #205 phase 1). With flock and the CAS primitive landed, this phase adds the generation-counter fields to the type layer so phase 3 can use them. No behavior change ships in this phase — fields default to zero and round-trip cleanly through existing `metadata.json` files.

## Proposal
Add generation-counter fields to the persisted `Metadata`/`Status` of all four kinds, on both the internal and external types, plus the converters that move objects between them:

- `internal/modelhub/{realm,space,stack,cell}.go` — add `Generation int64` to `*Metadata`, `ObservedGeneration int64` to `*Status`.
- `pkg/api/model/v1beta1/{realm,space,stack,cell}.go` — mirror the same fields on the external types.
- `internal/apischeme/scheme.go` — plumb the new fields through all 8 converter functions (4 kinds × `*ToInternal` / `*FromInternal`).

Zero-value defaults mean existing `metadata.json` files load unchanged. Phase 3 wires the writers and reconciler to actually populate the fields.

## Acceptance criteria
- [ ] `Generation int64` present on `*Metadata` for all 4 kinds, internal + external.
- [ ] `ObservedGeneration int64` present on `*Status` for all 4 kinds, internal + external.
- [ ] All 8 apischeme converters preserve both fields round-trip.
- [ ] Round-trip serialization tests covering zero-value reads from pre-existing metadata and write-then-read of non-zero values.
- [ ] No behavior change: no caller bumps `Generation`, no caller reads `ObservedGeneration` yet.

## Dependencies
Depends on #205 (phase 1 — flock + CAS primitive).

## Notes
- Phase 3 (RMW-under-flock in writers + reconciler skip-on-stale) is the value-shipping phase that uses these fields; this phase is the mechanical type-layer prerequisite.
- The original concurrency-primitives spec was split because the combined work spanned ~30 files / ~500–800 LoC across four cross-cutting subsystems (metadata primitive, types in both layers, controller adoption, reconciler). See #205 for the slicing rationale.

PR #607, opened 2026-05-19
feat(metadata): introduce sidecar flock + CAS primitives for concurrent writers

### LINKED ISSUE #205: metadata: introduce concurrency primitives — flock + CAS primitive (phase 1)
## Context
`metadata.json` files under `/opt/kukeon/<realm>/...` are read and written by both the daemon (in-process) and `kuke --no-daemon` (also in-process). The reconciler loop (#161, shipped) adds a third writer on a ticker. Today there's no synchronization: concurrent writes can lose updates or interleave partial fields, and a writer cannot detect that its read is stale.

This issue is phase 1 of a 3-phase split (see `## Out-of-scope follow-ups`). It introduces the locking and stale-detection primitives in `internal/metadata/`. The type-layer fields (`Generation` / `ObservedGeneration`) and the writer + reconciler adoption land in phases 2 and 3.

## Proposal
Introduce a POSIX file-lock primitive and a CAS-style write entry point in `internal/metadata/`:

- Sidecar `<file>.lock` per `metadata.json` (avoids the tmp-rename inode-swap pitfall of locking the data file directly).
- `WriteMetadata` acquires an exclusive flock for the duration of the atomic write.
- `ReadRaw` acquires a shared flock.
- New `WriteMetadataCAS` (or equivalent `WithExclusiveLock` + unlocked-write helpers) exposing the RMW-under-flock primitive so phase 3 callers can read existing, mutate, and write atomically with stale detection.
- `errdefs.ErrStaleResource` sentinel returned by `WriteMetadataCAS` when the underlying bytes changed between read and write.

No caller adopts the CAS API in this phase. Existing reads and writes start using flock, gaining torn-write protection across the daemon + `--no-daemon` boundary today, ahead of generation-based concurrency landing in phase 3.

## Acceptance criteria
- [ ] `internal/metadata/`: sidecar flock helper with exclusive / shared modes.
- [ ] Existing `WriteMetadata` acquires the exclusive flock; existing `ReadRaw` acquires the shared flock.
- [ ] New `WriteMetadataCAS` (or equivalent primitive) exposes RMW-under-flock with stale detection.
- [ ] `errdefs.ErrStaleResource` sentinel added.
- [ ] Tests: exclusive blocks shared/exclusive, shared allows shared, concurrent writers don't tear bytes, CAS returns `ErrStaleResource` on mismatch.

## Notes
- Pairs with the timestamps work (#166, shipped): `UpdatedAt` and the future generation counter will be bumped together on the same write in phase 3.
- The reconciler loop (#161, shipped) currently runs without generation-aware skip; phase 3 adds the skip behavior on top of this phase's primitives.

## Out-of-scope follow-ups
- #596 (phase 2) — type-layer `Generation` / `ObservedGeneration` fields across all 4 kinds.
- #597 (phase 3) — controller adoption (RMW-under-flock in `Update*Metadata`) and reconciler skip-on-stale.

eminwux:
## Summary

- Phase 1 of the three-phase metadata-concurrency split (#205): adds the locking and stale-detection primitives in `internal/metadata/` without yet adopting the CAS API in any caller.
- `WriteMetadata` now acquires an exclusive flock on a sidecar `<file>.lock` for the full write window, and `ReadRaw` acquires a shared flock — gaining torn-write protection across the daemon and `kuke --no-daemon` boundary today.
- New `WriteMetadataCAS(ctx, logger, prior, metadata, file)` exposes the read-compare-write primitive under the same exclusive flock; returns `errdefs.ErrStaleResource` (also new) when on-disk bytes diverged from the caller's observed `prior`. `WithExclusiveLock` / `WithSharedLock` + `WriteMetadataNoLock` / `ReadRawNoLock` are also exported for the cases phase 3 wants more control over the RMW window.
- `DeleteMetadata` removes the `.lock` sidecar alongside the data file so the existing empty-parent-cleanup branch still fires (preserves prior behavior), AND now wraps its remove window in `WithExclusiveLock` so a concurrent `WriteMetadata` mid-write cannot end with a data file stranded without its sidecar. These are the two out-of-AC touches — flagged here so the reviewer can push back.

## Design choices worth flagging

- **Sidecar `<file>.lock`, not the data file.** `WriteMetadata` persists via temp-file + rename; renaming swaps the data file's inode, which would detach any flock held on the old inode. The sidecar's inode is stable across writes, so the lock identity is preserved.
- **`ReadRaw` short-circuits to `ErrMissingMetadataFile` before acquiring the flock.** Otherwise probe-reads of optional realms (`ListSpaces` walks call `ReadRaw` on `<space>/metadata.json` that may not exist) would leak `.lock` sidecars into directories that were never written. The TOCTOU window matches the pre-existing snapshot semantics — a writer arriving after the existence check is just the same race callers already tolerate.
- **Shared-lock acquire is side-effect-free.** `acquireFlock` only `MkdirAll`s the parent and `O_CREATE`s the sidecar on the exclusive path; the shared path opens without `O_CREATE` and surfaces `ErrMissingMetadataFile` if the sidecar is absent. Closes the TOCTOU window where a sibling `DeleteMetadata` swept the parent dir between `ReadRaw`'s existence pre-check and the lock acquire — without this, the shared acquire would resurrect parent dir + a fresh `.lock` only to leak both after returning `ErrMissingMetadataFile`.
- **CAS compares raw payload bytes**, not a re-marshaled document — the doc comment makes this contract explicit because JSON serialization is not guaranteed byte-stable across edits.
- **`syscall.Flock`** rather than `golang.org/x/sys/unix.Flock` keeps the dependency surface unchanged (the project is Linux-only, already uses `syscall.SIGTERM` etc.).

## Test plan

- [x] `make test` — full named CI target, all packages green (including new `internal/metadata` tests).
- [x] `make dev-init` smoke — daemon parity check passes; `kuke get realms` and `kuke get realms --no-daemon` produce identical output, confirming the daemon's read path correctly traverses the flocked metadata. Re-run after the resurrection-fix landed because that fix touches the daemon's read-path flock handling.
- [x] New tests in `internal/metadata/lock_test.go`:
  - exclusive flock blocks both exclusive and shared acquisitions
  - shared flock admits concurrent shared holders
  - concurrent writers × 8 / 20 each on the same file produce no torn payload (8 KiB filler probed by a parallel reader)
  - `WriteMetadataCAS` returns `ErrStaleResource` on intervening-write, on `prior==nil` after first create, and on the canonical read-mutate-retry loop
  - `ReadRaw` does not leak a sidecar for missing data files
  - `WithExclusiveLock` + `WriteMetadataNoLock` / `ReadRawNoLock` work end-to-end without deadlocking the inner write
  - **NEW:** `WithSharedLock` after a sibling `DeleteMetadata` swept the parent dir returns `ErrMissingMetadataFile` AND does NOT resurrect the parent dir or sidecar on disk (deterministic regression for the reviewer's leak).
  - **NEW:** Stress probe with 8 writers + 8 readers + 4 deleters × 50 iters on the same path — after the storm drains, the parent dir is either absent or contains exactly `{data, sidecar}`; never an orphan dir, stray `.lock`, or data file without its sidecar.

## Out of scope (phases 2 & 3)

- Phase 2 (#596): `Generation` / `ObservedGeneration` fields on internal + external Realm/Space/Stack/Cell types and the 8 apischeme converters.
- Phase 3 (#597): adoption of `WriteMetadataCAS` in `internal/controller/runner/metadata.go` `Update*Metadata` writers and reconciler skip-on-stale on top of these primitives.

Closes #205

### DIFF
--- a/internal/errdefs/errdefs.go
+++ b/internal/errdefs/errdefs.go
@@ -41,6 +41,15 @@ var (
 	ErrUpdateRealmMetadata    = errors.New("failed to update realm metadata")
 	ErrCreateRealmNamespace   = errors.New("failed to create realm namespace")
 	ErrMissingMetadataFile    = errors.New("missing metadata file")
+	// ErrStaleResource fires from metadata.WriteMetadataCAS when the
+	// on-disk bytes diverged between the caller's prior read and the
+	// compare-and-swap write. Surfaces to a controller writer that read,
+	// mutated, and tried to persist a metadata document while another
+	// writer (daemon, `kuke --no-daemon`, or the reconciler) committed
+	// an intervening write under the same exclusive flock. The caller
+	// must re-read, re-apply the mutation, and retry — there is no
+	// silent merge of disjoint edits.
+	ErrStaleResource = errors.New("metadata resource changed since prior read")
 	ErrUnsupportedAPIVersion  = errors.New("unsupported apiVersion")
 	ErrUnknownKind            = errors.New("unknown kind")
 	ErrConversionFailed       = errors.New("conversion failed")
--- a/internal/metadata/lock.go
+++ b/internal/metadata/lock.go
@@ -52,22 +52,44 @@ func LockFilePath(file string) string {
 	return file + LockFileSuffix
 }
 
-// acquireFlock opens (creating if absent) the sidecar lock file for the
-// metadata file at `file` and acquires the requested flock mode. The
-// returned release closes the underlying file descriptor, which the
-// kernel uses to drop the flock.
+// acquireFlock opens the sidecar lock file for the metadata file at
+// `file` and acquires the requested flock mode. The returned release
+// closes the underlying file descriptor, which the kernel uses to drop
+// the flock.
 //
-// The parent directory is created up-front so the sidecar has somewhere
-// to live; this matches WriteMetadata's existing MkdirAll on the
-// create-new path.
+// Side effects differ by mode:
+//
+//   - Exclusive (write path): the parent directory is created via
+//     MkdirAll and the sidecar is opened with O_CREATE, mirroring
+//     WriteMetadata's existing create-new path. A writer always lands
+//     a sidecar — Phase 3 adopters can rely on that invariant.
+//
+//   - Shared (read path): neither the parent directory nor the sidecar
+//     is created. A missing sidecar surfaces as ErrMissingMetadataFile
+//     so a sibling DeleteMetadata that swept between the caller's
+//     existence pre-check and this acquire cannot be papered over by
+//     resurrecting a stale parent + sidecar that then leak after the
+//     read returns ErrMissingMetadataFile anyway.
 func acquireFlock(file string, exclusive bool) (func(), error) {
-	parent := filepath.Dir(file)
-	if mkErr := os.MkdirAll(parent, metadataDirMode); mkErr != nil {
-		return nil, fmt.Errorf("mkdir lock parent %s: %w", parent, mkErr)
+	if exclusive {
+		parent := filepath.Dir(file)
+		if mkErr := os.MkdirAll(parent, metadataDirMode); mkErr != nil {
+			return nil, fmt.Errorf("mkdir lock parent %s: %w", parent, mkErr)
+		}
 	}
 	lockPath := LockFilePath(file)
-	f, openErr := os.OpenFile(lockPath, os.O_CREATE|os.O_RDWR, lockFilePerm)
+	openFlag := os.O_RDWR
+	if exclusive {
+		openFlag |= os.O_CREATE
+	}
+	f, openErr := os.OpenFile(lockPath, openFlag, lockFilePerm)
 	if openErr != nil {
+		if !exclusive && os.IsNotExist(openErr) {
+			return nil, fmt.Errorf(
+				"sidecar lock file missing for %s: %w",
+				file, errdefs.ErrMissingMetadataFile,
+			)
+		}
 		return nil, fmt.Errorf("open lock file %s: %w", lockPath, openErr)
 	}
 	mode := syscall.LOCK_SH
--- a/internal/metadata/lock.go
+++ b/internal/metadata/lock.go
@@ -1,0 +1,180 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package metadata
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"syscall"
+
+	"github.com/eminwux/kukeon/internal/errdefs"
+)
+
+// lockFilePerm is the mode used for newly-created sidecar lock files. The
+// sidecar carries no payload — its only role is to anchor the flock — so
+// it inherits the data file's 0o644 and the parent dir's 0o0750 + setgid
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #607 Review — `feat(metadata): introduce sidecar flock + CAS primitives for concurrent writers`** _(updated)_

Solid foundation — concurrent-writer torn-payload probe, CAS happy/stale/retry, sidecar-leak guard on `ReadRaw` probes, daemon-parity smoke checked. Cross-boundary flock semantics are correct (sidecar inode preserved across rename, non-reentrant flock contract documented on `WithExclusiveLock`). Single commit, signed and verified.

Blocking:

- **Latent leak** `internal/metadata/lock.go:89-110`: `acquireFlock` runs `MkdirAll(parent)` + `OpenFile(lockPath, O_CREATE)` unconditionally, so `ReadRaw` racing a sibling `DeleteMetadata` (existence check passes, delete sweeps parent dir + sidecar, `ReadRaw` then enters `WithSharedLock`) recreates the parent dir + a fresh `.lock` and leaves both orphaned after returning `ErrMissingMetadataFile`. Suggested fix: scope the dir-create side-effect to write callers (drop `MkdirAll` from the shared-lock path), or re-check `existsFilePath` after the lock acquire and clean up the sidecar + parent on miss. Phase-3 adopters (#597) will widen this window when controllers retry on `ErrStaleResource` — fix before adoption lands.

Test plan additions needed (thorough — please don't ship the fix on unit coverage alone):

- Race test: writer creates, deleter sweeps the parent dir, reader hits the existence-check-then-lock window — assert `ErrMissingMetadataFile` AND that the parent dir and `.lock` sidecar are not resurrected on disk.
- Stress probe: concurrent `WriteMetadata` + `ReadRaw` + `DeleteMetadata` across N goroutines on the same path; after the storm drains, assert the working directory is in one of two well-defined states (data file present with its sidecar, or both absent with no parent) — never an orphan dir or stray `.lock`.
- `make dev-init` smoke re-run after the fix to confirm the daemon-parity check still passes (the fix touches the daemon's read-path flock handling).

LGTM once the leak fix lands and the new race + stress tests cover the window.

eminwux:
Addressed in 4eea2d9:

- **Latent leak in `acquireFlock`**: scoped the parent-dir create + sidecar `O_CREATE` to the exclusive path. The shared path now opens without `O_CREATE` and surfaces `ErrMissingMetadataFile` when the sidecar is missing — no resurrection of parent dir or sidecar when a sibling `DeleteMetadata` swept between the caller's existence pre-check and the lock acquire.
- **Related second-order race**: `DeleteMetadata` now wraps its remove window in `WithExclusiveLock` so a concurrent `WriteMetadata` mid-write cannot end with a data file stranded without its sidecar (the writer's `atomicWriteFile` rename would otherwise race the unlocked deleter's rm-sidecar). Without this, the stress probe's "data + sidecar OR both absent" invariant would still flake on the write/delete race even with the resurrection fix.
- **Race test**: `TestWithSharedLockDoesNotResurrectAfterDelete` — deterministic regression that seeds normal state, simulates a sibling deleter via `os.RemoveAll(parent)`, calls `WithSharedLock` directly, and asserts the error + that neither parent dir nor sidecar are resurrected on disk.
- **Stress probe**: `TestStressConcurrentWriteReadDelete` — 8 writers + 8 readers + 4 deleters × 50 iters on the same path; asserts the terminal-state invariant after `wg.Wait()`. Catches both pre-fix windows (read-resurrection + write/delete-strand) simultaneously.
- **Smoke re-run**: `make dev-init` daemon-parity check passes after the fix (both `default` and `kuke-system` Ready in both daemon + `--no-daemon` views).

PR body updated to reflect the new test bullets and the second flagged out-of-AC touch (`DeleteMetadata` flock-wrapping).

eminwux:
**PR #607 Re-review — `feat(metadata): introduce sidecar flock + CAS primitives for concurrent writers`**

LGTM. Fix commit `4eea2d9` lands the resurrection-race fix cleanly — `acquireFlock` now splits mode (exclusive does `MkdirAll` + `O_CREATE`, shared opens without and surfaces `ErrMissingMetadataFile` on a missing sidecar), matching the suggested approach. The author also self-detected an adjacent strand-data race (`DeleteMetadata` now wrapped in `WithExclusiveLock` so a concurrent rename cannot land data without a sidecar) and flagged it honestly as out-of-AC. `TestWithSharedLockDoesNotResurrectAfterDelete` (deterministic regression) and `TestStressConcurrentWriteReadDelete` (8w / 8r / 4d × 50 iters, asserts terminal parent is either absent or exactly `{data, sidecar}`) cover both windows. `make dev-init` re-run noted; both commits signed and verified. Ship it.

PR #610, opened 2026-05-19
fix(metadata): materialise sidecar on shared-lock read of pre-#607 data

eminwux:
## Summary

- **Symptom:** on a host upgraded across the #607 boundary, `make dev-init` fails at the `kuke image load --realm kuke-system` step with `Error: realm not found: kuke-system`. The same host also reports `No realms found.` from `kuke get realms --no-daemon`, even though `/opt/kukeon/data/{default,kuke-system}/metadata.json` are both on disk.
- **Cause:** PR #607's shared-lock read path opens the sidecar `.lock` without `O_CREATE` and turns `IsNotExist` into `ErrMissingMetadataFile`. Upstream that becomes `ErrRealmNotFound` (`internal/controller/runner/get.go:38`). Pre-#607 binaries wrote the data file without ever staging a sidecar, so the first read by a post-#607 binary on any such file is a miss.
- **Fix:** in `acquireFlock`'s shared path, distinguish the two missing-sidecar cases. Data file gone → preserve the TOCTOU behaviour the existing comment describes (return `ErrMissingMetadataFile` without resurrecting parent + sidecar; `TestWithSharedLockDoesNotResurrectAfterDelete` still passes). Data file present → materialise the sidecar with `O_CREATE` so legacy state stays readable.
- **Race analysis:** `DeleteMetadata` holds the exclusive flock while it unlinks the data file and sidecar, so any reader that races into the legacy branch either (a) opens an inode the deleter still holds exclusively and then sees the data file gone under `ReadRawNoLock` (returns `ErrMissingMetadataFile`), or (b) opens a fresh sidecar inode the deleter never touched and reads a still-extant data file. Concurrent legacy readers/writers serialise on the materialised sidecar normally.

## Test plan

- [x] `go test ./internal/metadata/...` — new `TestWithSharedLockReadsLegacyMetadataWithoutSidecar` asserts ReadRaw on a sidecar-less data file returns the payload and materialises the sidecar. Existing TOCTOU test (`TestWithSharedLockDoesNotResurrectAfterDelete`) still passes because it removes the whole parent dir, so the legacy fall-through's data-file existence check correctly fails.
- [x] `make dev-init` end-to-end on a host carrying pre-#607 `/opt/kukeon/data/{default,kuke-system}/metadata.json` (no sidecars). Daemon parity check now shows both realms in both `kuke get realms` and `kuke get realms --no-daemon`. kuke-attach smoke roundtrips cleanly. Sidecars present in `/opt/kukeon/data/` after the run.

### DIFF
--- a/internal/metadata/lock.go
+++ b/internal/metadata/lock.go
@@ -64,12 +64,14 @@ func LockFilePath(file string) string {
 //     WriteMetadata's existing create-new path. A writer always lands
 //     a sidecar — Phase 3 adopters can rely on that invariant.
 //
-//   - Shared (read path): neither the parent directory nor the sidecar
-//     is created. A missing sidecar surfaces as ErrMissingMetadataFile
-//     so a sibling DeleteMetadata that swept between the caller's
-//     existence pre-check and this acquire cannot be papered over by
-//     resurrecting a stale parent + sidecar that then leak after the
-//     read returns ErrMissingMetadataFile anyway.
+//   - Shared (read path): the sidecar is opened without O_CREATE so a
+//     sibling DeleteMetadata that swept between the caller's existence
+//     pre-check and this acquire surfaces as ErrMissingMetadataFile
+//     instead of being papered over by a resurrected parent + sidecar
+//     that then leak. The legacy fall-through below distinguishes that
+//     TOCTOU case from a data file written by a pre-sidecar binary
+//     (#607 rollout) where the data file is still on disk but its
+//     sidecar was never materialised.
 func acquireFlock(file string, exclusive bool) (func(), error) {
 	if exclusive {
 		parent := filepath.Dir(file)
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #610 Review — `fix(metadata): materialise sidecar on shared-lock read of pre-#607 data`**

LGTM — tight migration fix: the new `!existsFilePath(file)` re-check inside `acquireFlock` cleanly splits the TOCTOU sweep (preserve `ErrMissingMetadataFile`, no parent/sidecar resurrection) from the legacy pre-#607 state (data file on disk, sidecar absent → materialise with `O_CREATE`). Race analysis holds against `DeleteMetadata`'s exclusive-flocked unlink-data + unlink-sidecar window. Existing `TestWithSharedLockDoesNotResurrectAfterDelete` still asserts the TOCTOU branch (parent-dir wiped → `existsFilePath` false → no resurrection), and the new `TestWithSharedLockReadsLegacyMetadataWithoutSidecar` pins the rollout fall-through. Test plan covers `make dev-init` daemon-parity end-to-end.

PR #615, opened 2026-05-19
fix(daemon): probe kukeond socket alongside cell metadata in lifecycle verbs

### LINKED ISSUE #611: cli: kuke daemon start claims "already running" from stale metadata without probing the socket
## Context

`kuke daemon start` decides whether kukeond is already up by calling
`lifecycle.IsCellRunning(getRes.Cell)` — which inspects the persisted
`Cell.Status.State` and `Cell.Status.Containers[].State` (`cmd/kuke/daemon/internal/lifecycle/lifecycle.go:53`).
If either reads `Ready`, the command prints
`kukeond is already running (cell "kukeond" in realm "kuke-system")` and
exits 0 without ever dialing `/run/kukeon/kukeond.sock`.

Failure mode: kukeond is force-killed externally (OOM, host reboot mid-run,
operator `kill -9`, containerd panic) and the metadata never gets the
"stopped" write. Persisted state still says `Ready`. Next invocation:

- `kuke daemon start` → "already running", exit 0. Operator believes the
  daemon is up; the socket is missing; the very next daemon-routed command
  fails with `dial unix /run/kukeon/kukeond.sock: connect: no such file or directory`.

The same metadata-only check gates the idempotent short-circuit in
`stop`, `kill`, `restart`, `reset`, and `logs` (all callsites of
`IsCellRunning` per grep). A fix in `IsCellRunning` benefits the whole
verb family at once.

`kuke refresh` (`cmd/kuke/refresh/refresh.go`) already exists to push
metadata `.status` back into sync by introspecting containerd — but
refresh alone is insufficient for the failure mode above, because
containerd itself may still believe the task is running if the kukeond
binary died but the task struct survived. The authoritative liveness
signal is socket reachability.

## Proposal

Augment the "is kukeond actually up?" check used by `kuke daemon start`
(and inherited by the other lifecycle verbs) with a socket-reachability
probe:

1. Add `IsDaemonReachable(ctx, socketPath, timeout) bool` in
   `cmd/kuke/daemon/internal/lifecycle/` — short `net.DialTimeout` on
   `unix:<socketPath>` with a small budget (e.g., 500ms).
2. In `runStart`, after the `IsCellRunning(cell)` true branch:
   - If the socket is reachable → keep current "already running" message
     and exit 0.
   - If the socket is NOT reachable → treat the metadata as stale, log a
     short reconciliation note ("metadata reported Ready but socket
     unreachable; restarting cell"), and fall through to `StartCell` to
     bring the daemon back up.
3. Same probe check on `stop`, `kill`, `restart`'s "already stopped"
   branches so they don't silently no-op when the metadata is stale in
   the other direction.

`kuke refresh` is *not* a substitute — the probe is the source of truth
for whether the daemon is actually serving requests. Refresh updates
status fields from containerd's view; it does not detect a
crashed-but-task-still-registered kukeond.

Out of scope: changing the reconciler to repair the stale `.status`
field. That's #161/#224's lane; this issue is the operator-facing fix.

## Acceptance criteria

- [ ] New `lifecycle.IsDaemonReachable` (or equivalent) probes the
      kukeond socket with a bounded dial timeout.
- [ ] `kuke daemon start` no longer prints "already running" when the
      persisted state says `Ready` but the socket cannot be dialed; it
      logs the staleness and proceeds to `StartCell`.
- [ ] Idempotent-stop short-circuits in `stop`, `kill`, and `restart`
      similarly fall through when the socket is reachable despite the
      metadata saying not-Ready (catches the other staleness direction).
- [ ] Tests cover both staleness directions (metadata=Ready/socket=down
      and metadata=NotReady/socket=up) for `kuke daemon start` at
      minimum.
- [ ] CLI use-cases doc updated: the "Idempotent" notes in
      `docs/cli-use-cases.md:133-135` clarify that idempotence is based
      on liveness, not persisted state.

## Notes

- Callsites: `cmd/kuke/daemon/start/start.go:81`, `restart/restart.go:95`,
  `stop/stop.go:93`, `kill/kill.go:81`, `reset/reset.go:120`, `logs/logs.go:101`.
- Related: #217 (daemon subcommand umbrella), #224 (reconciler-driven
  model — long-term home for stale-status repair), #202 (`kuke status`
  health report — could surface this divergence proactively).

eminwux:
## Summary
- `kuke daemon start` used to read only persisted `Cell.Status.{State,Containers[].State}` to decide "already running" and exit 0. If the daemon was killed externally (OOM, host reboot mid-run, `kill -9`, containerd panic) the metadata still said `Ready`, the operator saw "already running", and the very next daemon-routed command failed with `dial unix /run/kukeon/kukeond.sock: connect: no such file or directory`. Same `IsCellRunning` short-circuit gates the idempotent no-op branches in `stop`, `kill`, and `restart`, with the symmetric staleness failure (metadata=NotReady but socket=up → silent no-op against a live daemon).
- This PR adds `lifecycle.IsDaemonReachable(ctx, socketPath, timeout)` — a 500ms `net.Dialer`-backed dial of the kukeond unix socket — and threads it through `start` (Ready+socket=down → log and fall through to `StartCell`), `stop` / `kill` (NotReady+socket=up → log and fall through to the act-on-live-daemon path), and `restart` (likewise — preserves the existing "skip stop, run start" branch only when *both* axes agree the daemon is down).
- Probe injection point matches the existing `MockClientKey` pattern (`ReachableProbe` type + `ReachableProbeKey` context value + `ResolveReachableProbe(cmd)` resolver) so the new branches are unit-testable without a real listening socket. `ResolveSocketPath()` centralises the `KUKEOND_SOCKET` viper lookup so every verb agrees on which socket counts as the source of truth.
- The fall-through-to-`StartCell` path is correctly closed downstream — `internal/controller/start_cell.go:42-58` already refreshes container state from containerd via `populateCellContainerStatuses` before deciding the cell is busy, so a genuinely OOM-killed cell (containers actually dead, metadata stale-Ready) lets `StartCell` proceed and bring the daemon back up.
- Out of scope per issue: `kuke daemon reset` / `kuke daemon logs` short-circuits, and repair of the stale `.status` field itself (#161/#224's lane). `kuke refresh` is intentionally *not* a substitute — refresh reconciles from containerd's view, not from socket reachability, and won't detect a crashed-but-task-still-registered kukeond.

## Test plan
- [x] `go vet ./...` — clean.
- [x] `make test` — full suite passes, including new cases in start/stop/kill/restart/lifecycle.
- [x] `pre-commit run --files $(git diff --name-only HEAD)` — passes (golangci-lint auto-fix re-run lands clean).
- [x] `make dev-init` smoke — daemon-parity guard prints both Ready realms identically; nested attach plumbing intact.
- [x] Manual end-to-end: `sudo KUKEOND_SOCKET=/run/kukeon-dev/kukeond.sock ./kuke daemon start` against the live nested daemon prints "kukeond is already running" (probe reachable → existing path). `sudo ./kuke daemon start` with the default `/run/kukeon/kukeond.sock` (which is *not* the live nested socket) prints "kukeond metadata reports Ready but socket /run/kukeon/kukeond.sock is unreachable; restarting cell" — the new staleness branch fires as designed. Note: in this nested smoke the daemon's containers were genuinely still running (just at a different socket path) so `StartCell` correctly returned the "has running containers" error per `start_cell.go:43-57` — that is the controller path doing its containerd-refresh-based safety check, not a downstream gap.

## Notes for the reviewer
- Probe budget is 500ms; chose generous-but-fast because the local-socket happy path answers in <1ms and a hung accept still surfaces well below human reaction time.
- `daemon reset` / `daemon logs` callsites were left alone per the AC. Reset's `IsCellRunning` gate decides whether to invoke `StopPhase` (StopCell is idempotent at the controller layer, so a stale-Ready dispatch is harmless); logs returns an error rather than no-op on `!IsCellRunning`, so the staleness failure-mode shape there is different.

Closes #611

### DIFF
--- a/cmd/kuke/daemon/internal/lifecycle/lifecycle.go
+++ b/cmd/kuke/daemon/internal/lifecycle/lifecycle.go
@@ -24,13 +24,16 @@ package lifecycle
 import (
 	"context"
 	"fmt"
+	"net"
 	"time"
 
+	"github.com/eminwux/kukeon/cmd/config"
 	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/errdefs"
 	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 // MockClientKey injects a kukeonv1.Client (typically a fake) via the command
@@ -45,6 +48,13 @@ type MockClientKey struct{}
 // verbs cannot drift on what "graceful" means.
 const DefaultTimeout = 10 * time.Second
 
+// DefaultReachableTimeout is the dial budget the lifecycle verbs use when
+// probing the kukeond socket. Short on purpose: the only acceptable failure
+// mode here is "the socket is dead" — a healthy daemon answers in <1ms over
+// the local unix socket, so 500ms is generous and a hung accept still
+// surfaces fast enough that the operator notices.
+const DefaultReachableTimeout = 500 * time.Millisecond
+
 // IsCellRunning treats the cell as live if any container reports Ready, or
 // if the persisted cell state is Ready. Every `kuke daemon` lifecycle verb
 // uses this to agree on what "running" means; the controller's StartCell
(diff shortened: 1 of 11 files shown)


eminwux:
**PR #615 Review — `fix(daemon): probe kukeond socket alongside cell metadata in lifecycle verbs`**

LGTM — the socket probe correctly disambiguates both staleness directions (Ready+socket=down → reconcile via StartCell; not-Ready+socket=up → fall through to StopPhase/KillCell), the `ReachableProbeKey`/`ResolveReachableProbe` indirection matches the existing `MockClientKey` precedent so tests stay deterministic, error swallow in `IsDaemonReachable` is intentional and documented, and the `make dev-init` regression guard is checked in the test plan.

PR #628, opened 2026-05-20
feat(metadata): add Generation/ObservedGeneration to all 4 kinds (phase 2)

eminwux:
## Summary
- Phase 2 of the metadata concurrency-primitives work (#205 phase 1 landed flock + CAS). This adds the generation-counter fields to the type layer so phase 3 can wire the writers/reconciler. **No behavior change** — both fields default to zero, no caller bumps `Generation` or reads `ObservedGeneration` yet.
- `Generation int64` added to `*Metadata` on all 4 kinds (realm/space/stack/cell), internal (`internal/modelhub/`) + external (`pkg/api/model/v1beta1/`). External fields carry `json/yaml:"generation,omitempty"` so pre-existing `metadata.json`/manifests keep their on-disk shape.
- `ObservedGeneration int64` added to `*Status` on all 4 kinds, both layers, tagged `observedGeneration,omitempty`.
- All 8 apischeme converters (`*ToInternal` / `Build*ExternalFromInternal` for each kind) plumb both fields round-trip.

## Notes
- The issue lists the field locations as `internal/modelhub/{realm,space,stack,cell}.go` and `pkg/api/model/v1beta1/{realm,space,stack,cell}.go` plus the converters in `internal/apischeme/scheme.go` — all confirmed present and matching the architecture.
- Field godoc references "phase 3" as the value-shipping phase per the issue's own framing; no follow-up issue filed since phase 3 is already tracked as the next phase of the #205 series.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./internal/apischeme/ ./internal/modelhub/ ./pkg/api/model/v1beta1/`
- [x] `make test` (full target: `go test $(go list ./... | grep -v /e2e)`) — all packages pass
- [x] New `internal/apischeme/generation_test.go`: per-kind round-trip of non-zero values through both converters + YAML, and zero-value reads from pre-existing metadata (fields default to zero, omitted from re-serialized YAML)
- [ ] `make dev-init` smoke — not run: type-layer-only change, touches no build path, daemon, or `/opt/kukeon` data, so the smoke regression guard does not apply

Closes #596

PR #633, opened 2026-05-20
feat(metadata): RMW-under-flock generation bump + reconciler stale skip (phase 3)

### LINKED ISSUE #597: metadata: RMW-under-flock writers + reconciler generation skip (phase 3)

eminwux:
## Summary

Phase 3 of the metadata concurrency primitives (after #205 flock+CAS, #628 Generation/ObservedGeneration type fields). Closes the lost-update gap between the daemon reconcile loop and concurrent user writes.

- **Writers RMW under flock, bump `Generation` only on spec change.** The four `Update{Realm,Space,Stack,Cell}Metadata` now route through a shared generic `writeDocWithGeneration` helper that, under the sidecar flock, reads the prior on-disk doc, treats the caller's incoming `Generation` as an optimistic token, bumps `Metadata.Generation` only when the persisted spec differs (a status-only update preserves it), and CAS-writes against the observed bytes.
- **CAS/token failure surfaces as `errdefs.ErrStaleResource` at the writer's call site** — propagated through the existing `errdefs.ErrWriteMetadata` wrap, so `errors.Is` chains hold.
- **Reconciler stamps `ObservedGeneration` after each persist and skips stale persists.** `persistCellStatusGuarded` stamps `Status.ObservedGeneration = Metadata.Generation` before writing; an `ErrStaleResource` (a concurrent spec writer advanced the on-disk generation past the reconciler's observation) is reported as a benign skip (`persisted=false`, nil err) so the next tick re-derives from the fresher spec instead of clobbering it.

## Design calls (reviewer please push back)

- **Optimistic-token model over CAS-against-own-read.** A blind full-object-set API (`UpdateXMetadata(value)`) cannot prevent lost updates by CASing against the writer's *own* fresh read — a caller modifying a stale snapshot still wins the CAS. The incoming `Generation` is the token; the writer rejects when on-disk generation has moved past it. This is the core fix for the daemon-status-vs-user-spec race the issue describes.
- **`Generation == 0` escape hatch.** Create/provision flows construct the doc from scratch (gen 0) rather than reading it back; the token check is skipped for them (the spec-vs-disk bump still applies) so a create's own follow-up writes don't trip a false stale. A freshly-read doc always carries gen ≥ 1, so 0 reliably means "built it, didn't read it."
- **`refreshCellGeneration` after `EnsureCell`'s spec write.** `CreateCell` into an *existing* cell does a spec-bumping write (merged container) then a follow-up status persist on the same in-memory object; without re-syncing the token the follow-up would be rejected as stale against its own write. Added a regression test for this chained-write path.
- **`observedGenerationBehind` wired into `ReconcileCell` only, not `RefreshCell`.** Without it, a cell that reached Ready synchronously (never via a reconcile persist) sits at `ObservedGeneration=0` forever, making the AC5 skip comparison meaningless. The daemon reconcile loop now persists once to converge `ObservedGeneration` → `Generation` even with no derived-status change. I deliberately left `kuke refresh` (`RefreshCell`) untouched — it's operator-driven and shouldn't manufacture writes purely to bump a bookkeeping field; it still stamps `ObservedGeneration` via the guarded persist on any real status change.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./internal/controller/...`
- [x] New tests: `go test ./internal/controller/runner/ -run 'Generation|StaleToken|Converge|ChainedSpec|PersistCellStatusGuarded' -v` — 5/5 pass
  - `TestUpdateRealmMetadataBumpsGenerationOnSpecChange` (AC1: first→1, status-only→1, spec-change→2)
  - `TestUpdateRealmMetadataStaleTokenRejected` (AC2: stale token → `ErrStaleResource`, winner survives)
  - `TestUpdateRealmMetadataConcurrentWritesConverge` (AC4: 8 goroutines RMW-with-retry, no lost updates)
  - `TestChainedSpecThenStatusWriteNeedsGenerationRefresh` (EnsureCell refresh regression)
  - `TestPersistCellStatusGuardedSkipsWhenBehind` (AC3/AC5: stale view skips without clobbering spec; current view persists + stamps ObservedGeneration)
- [x] `make test` (full CI target, `go test $(go list ./... | grep -v /e2e)`) — all green
- [x] `make dev-init` smoke + daemon-parity check — both `kuke get realms` and `--no-daemon` show identical output; verified the kukeond cell converged to `generation: 2 | observedGeneration: 2 | state: Ready` (was stuck at `observedGeneration: 0` before this fix)

Closes #597

### DIFF
--- a/internal/controller/runner/create_cell.go
+++ b/internal/controller/runner/create_cell.go
@@ -180,6 +180,11 @@ func (r *Exec) EnsureCell(cell intmodel.Cell) (intmodel.Cell, error) {
 	if err := r.UpdateCellMetadata(ensuredCell); err != nil {
 		return intmodel.Cell{}, fmt.Errorf("%w: %w", errdefs.ErrUpdateCellMetadata, err)
 	}
+	// The write above may have bumped the generation (a merged container is
+	// a spec change). Sync the in-memory token so the caller's follow-up
+	// status persist (PopulateAndPersistCellContainerStatuses) is not
+	// rejected as stale against its own write.
+	r.refreshCellGeneration(&ensuredCell)
 
 	return ensuredCell, nil
 }
--- a/internal/controller/runner/metadata.go
+++ b/internal/controller/runner/metadata.go
@@ -17,7 +17,12 @@
 package runner
 
 import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"log/slog"
 	"time"
 
 	"github.com/eminwux/kukeon/internal/apischeme"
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #633 Review — `feat(metadata): RMW-under-flock generation bump + reconciler stale skip (phase 3)`**

Solid close of the reconciler-vs-user lost-update race. The two-layer protection is correct: the optimistic token catches "on-disk generation advanced past my token" — the case where the CAS-against-fresh-`prior` would otherwise succeed and clobber a newer spec — while CAS catches a write landing in the read→write window. No lost-update interleaving escapes both. Tests cover AC1–5, the chained spec→status write, and 8-way convergence; no dead code, commits verified, daemon-parity smoke documented.

One non-blocking item:

- **Sibling consistency** `internal/controller/runner/refresh.go:138`, `:230`, `:312`: converting all four `Update*Metadata` to CAS+token means `RefreshRealm`/`RefreshSpace`/`RefreshStack` now surface `ErrStaleResource` on a concurrent spec write, but only the cell path got the `persistCellStatusGuarded` benign-skip. The three sibling paths return it as a hard error → `result.Errors` line in `kuke refresh` and a daemon warning. Functionally fine (no clobber, self-heals next tick, `kuke refresh` still exits 0), just cosmetic noise on a rare race. Worth extending the guard to the siblings in a follow-up — not blocking for the cell-scoped fix here.

LGTM — ship it. Filing the sibling-guard item as a followup-nit.

PR #638, opened 2026-05-21
fix(cli): route shell completion through the daemon client

### LINKED ISSUE #634: cli: route shell completion through the daemon, not metadata-direct reads
## Context

Shell completion of resource names (`kuke get cell <TAB>`, and the realm/space/stack/cell/container completers generally) silently returns **nothing** for the standard install where `kukeond` runs as root and the operator runs `kuke` unprivileged.

Root cause: completion and the actual `get` verbs resolve their data through **different paths**.

- The completers build an in-process controller via `controllerFromCmd` → `controller.NewControllerExec(...)` (`cmd/config/autocomplete.go:53`), which reads the on-disk metadata tree under `--run-path` (`/opt/kukeon/data`) **directly** — the same path as `--no-daemon`.
- The `get` verbs resolve through `ClientFromCmd` → the **daemon** over the kukeond socket (`cmd/kuke/get/cell/cell.go:146`).

Because `kuke __complete …` is invoked as the unprivileged shell user, its direct read of root-owned `/opt/kukeon/data` is denied, and every completer swallows the error into an empty list (`cmd/config/autocomplete.go:362-365`, and the sibling `return []string{}, ShellCompDirectiveNoFileComp` branches). So `kuke get cell` lists cells (daemon, root reads fine) while `kuke get cell <TAB>` completes nothing.

Reproduced this session: `kuke __complete get cell ''` returns `:4` (`ShellCompDirectiveNoFileComp`) with **zero candidates**, on a host where `bash-completion` is installed and the bash glue (`_get_comp_words_by_ref`, `__start_kuke`, `complete -F`) is all correctly wired — so the failure is in the binary's completion data path, not the shell integration.

## Proposal

Route shell completion through the **daemon client** (the same `ClientFromCmd` path the `get` verbs use), so completion reflects the daemon's authoritative view and works for any user who can already run `kuke get …` (the kukeond socket is group-accessible via `KUKEOND_SOCKET_GID`).

The metadata-direct path was a deliberate latency optimization — see the comment in `internal/controller/runner/get.go` `ListContainers` ("For autocomplete, we can read directly from metadata files without calling containerd / This avoids the containerd connection that GetCell would trigger"). Switching to the daemon must preserve completion's responsiveness: keep it fast, never block the prompt, and degrade to an empty list quickly if the daemon socket is unavailable.

### Behavior
- [ ] `kuke get cell <TAB>` (and realm/space/stack/container) completes names for an unprivileged user against a root-owned `/opt/kukeon`, matching what `kuke get cell` lists.
- [ ] The completers resolve through the daemon client rather than `controller.NewControllerExec` metadata-direct reads.
- [ ] Completion degrades to an empty list quickly (no hang, no error spew at the prompt) when the daemon socket is absent or unreadable.
- [ ] Decide and document the `--no-daemon` story for completion: either completion follows the same `--no-daemon` flag the `get` verbs honor, or it is daemon-only. State which.

### Testing
- [ ] **Unit (happy path):** a completer (at minimum `CompleteCellNames`) returns names through the daemon-backed client path. The existing tests in `cmd/config/autocomplete_test.go` exercise only the metadata-direct `controller.Exec`, so this needs the daemon client wired (mock/fake) into the completion path.
- [ ] **Unit (degradation):** with the daemon socket absent/unreadable, the completer returns an empty list and the chosen directive without hanging or erroring — asserts the graceful-degradation criterion above.
- [ ] **Unit (the bug):** a test models the unprivileged-user-vs-root-daemon case — daemon returns names that a metadata-direct read could not — proving completion now follows the daemon, not the on-disk tree.
- [ ] **E2E:** an `e2e/e2e_kuke_test.go` case runs `kuke __complete get cell ''` against a real running daemon with a known cell and asserts the cell name appears in the output and the directive is `ShellCompDirectiveNoFileComp` (mirrors the manual repro in this issue; sits alongside the existing `TestKuke_Autocomplete_Bash`, which only checks static script generation).

## Notes

- Distinct from #632 (documentation: bash-completion package prerequisite) — that note is still valid for minimal hosts, but it does not address this divergence. Cross-linked.
- Strong candidate for `priority:A` if completion is part of the expected first-run UX — flagging for the user to bump.
- The `kuke get cell` vs `kuke get cell --no-daemon` divergence here is the same class of daemon-vs-on-disk drift that `make dev-init`'s realm-parity check guards against (see project CLAUDE.md).

eminwux:
## Summary

- Shell completion of resource names (`kuke get cell <TAB>`, and the realm/space/stack/cell/container completers) silently returned nothing on the standard install where `kukeond` runs as root and the operator runs `kuke` unprivileged: the completers read the on-disk metadata tree under `--run-path` directly via `controller.NewControllerExec`, which an unprivileged shell user cannot read against root-owned `/opt/kukeon`. The `get` verbs worked because they route through the group-accessible kukeond socket.
- This routes the completers through a `kukeonv1.Client` (`completionClient` in `cmd/config/autocomplete.go`) instead of the metadata-direct controller, mirroring `shared.ClientFromCmd`'s source selection so completion reflects the same authoritative view `kuke get …` shows the user running it.
- Each completer now wraps its `List*` call in a short (`2s`) context so a slow/hung/absent daemon degrades to an empty list quickly rather than blocking the prompt.

### `--no-daemon` decision (AC item)

Completion follows the **same `kukeon/noDaemon` viper key the `get` verbs honor** — there is no second selection path to keep in sync. Default is the daemon; the in-process branch is taken when that key is set.

One caveat is documented in the `completionClient` godoc and drove the e2e harness wiring: cobra's hidden `__complete` command does not run the root `PersistentPreRunE` and does **not** mark inherited *persistent flags* (`--host`, `--no-daemon`, `--run-path`) as changed for the completion request. So during real shell completion those flags do not reach viper; the **env vars** (`KUKEON_HOST`, `KUKEON_NO_DAEMON`, …) and the built-in viper defaults do. For the standard install this is exactly right — completion dials the default kukeond socket (or `KUKEON_HOST` when overridden) and stays on the daemon path. This was verified empirically while building the e2e test (env-host completes the cell; `--host` flag does not), which is why the e2e points the daemon via `KUKEON_HOST`.

## Non-obvious calls (for reviewer scrutiny)

- **`MockControllerKey` now carries a `kukeonv1.Client`, not a `*controller.Exec`.** Nothing outside `autocomplete.go` referenced the old type; this aligns the completion package with the `MockControllerKey`-injects-a-`Client` convention every other command package already uses (e.g. `cmd/kuke/get/cell`).
- **Existing on-disk-metadata unit tests are preserved** by setting `kukeon/noDaemon=true` in `setupTestCommand`, so they keep exercising the in-process path verbatim. New mock-client tests cover the daemon-backed path, degradation, and the bug.
- **The e2e seeds cell metadata on disk** (full realm/space/stack/cell chain) and reads it back through a real `kukeond serve`, rather than `kuke create cell`. `ListCells` is a pure on-disk traversal (no containerd), so this exercises the daemon RPC completion path end-to-end without the containerd/cgroup/root prerequisites `create` would pull in — keeping the test runnable in CI without the full init stack.

## Test plan

- [x] `make test` — full CI unit-test target, all packages pass (exit 0).
- [x] `go vet ./...` — clean.
- [x] `pre-commit run --files <changed>` — gofmt/golangci-lint/addlicense all pass.
- [x] New unit tests in `cmd/config/autocomplete_daemon_test.go`: daemon-backed happy path (cell + realm), graceful degradation with the socket absent, and the bug case (daemon returns names an empty on-disk tree cannot).
- [x] New e2e `TestKuke_Autocomplete_GetCell_ViaDaemon`: `kuke __complete get cell ''` against a real running daemon with a seeded cell asserts the cell name appears and the directive is `ShellCompDirectiveNoFileComp` (`:4`). Verified locally via `E2E_BIN_DIR=$(pwd) go test -run TestKuke_Autocomplete_GetCell_ViaDaemon ./e2e`.
- [ ] `make dev-init` daemon-parity smoke — not applicable: this change is CLI-completion-only (client-side data path); it does not touch the daemon, the build path, or anything under `/opt/kukeon` that the daemon reads. (docker/containerd are also not running on this dev host.)

Closes #634

### DIFF
--- a/cmd/config/autocomplete.go
+++ b/cmd/config/autocomplete.go
@@ -17,40 +17,86 @@
 package config
 
 import (
+	"context"
 	"log/slog"
 	"strings"
+	"time"
 
 	"github.com/eminwux/kukeon/cmd/types"
 	"github.com/eminwux/kukeon/internal/cellprofile"
+	"github.com/eminwux/kukeon/internal/client/local"
 	"github.com/eminwux/kukeon/internal/controller"
 	"github.com/eminwux/kukeon/internal/errdefs"
+	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
 
-// MockControllerKey is used to inject mock controllers in tests via context.
+// MockControllerKey is used to inject a mock kukeonv1.Client via context in
+// tests, matching the convention every other command package uses (see e.g.
+// cmd/kuke/get/cell). Despite the historical name it now carries a
+// kukeonv1.Client, not a controller.Exec.
 type MockControllerKey struct{}
 
-// controllerFromCmd creates a controller.Exec from the command context.
-// This duplicates the logic from cmd/kuke/create/shared to avoid import cycles.
-// If a mock controller is provided in the context, it will be used instead of creating a real one.
-func controllerFromCmd(cmd *cobra.Command) (*controller.Exec, error) {
-	// Check for mock controller in context (for testing)
-	if mockCtrl, ok := cmd.Context().Value(MockControllerKey{}).(*controller.Exec); ok {
-		return mockCtrl, nil
+// completionTimeout bounds how long a completer waits on the daemon before
+// degrading to an empty list. Shell completion must never block the prompt,
+// so this is deliberately short: a slow, hung, or unreachable daemon yields
+// no candidates rather than a frozen terminal (issue #634).
+const completionTimeout = 2 * time.Second
+
+// completionClient returns the kukeonv1.Client the completers resolve resource
+// names through. It mirrors shared.ClientFromCmd's source selection so
+// completion reflects the same authoritative view `kuke get …` shows the user
+// running it: the `kukeon/noDaemon` viper key picks the in-process controller
+// (local.New, direct on-disk metadata reads, requires privileges) versus the
+// kukeond daemon (RPC over the configured KUKEON_HOST socket).
+//
+// The default is the daemon path. This is the fix for issue #634: the
+// completers previously read /opt/kukeon/data directly via controller.Exec,
+// which an unprivileged shell user cannot read on the standard root-owned
+// install, so completion silently returned nothing while `kuke get …` (which
+// routes through the group-accessible daemon socket) worked. Routing
+// completion through the daemon makes it work for any user who can already
+// run `kuke get …`.
+//
+// --no-daemon story: completion follows the same `kukeon/noDaemon` viper key
+// the get verbs honor — there is no second selection path to keep in sync.
+// One caveat is worth stating because it shapes what works at a live prompt:
+// cobra's hidden `__complete` command does not run the root PersistentPreRunE,
+// and it does not mark inherited *persistent flags* (`--host`, `--no-daemon`,
+// `--run-path`) as changed for the completion request. So during real shell
+// completion those flags do not reach viper; what does reach it is the env
+// (KUKEON_HOST, KUKEON_NO_DAEMON, KUKEON_RUN_PATH, bound via config.DefineKV's
+// viper.SetDefault plus the env bindings) and the built-in defaults. In
+// practice that is exactly right for the standard install: completion dials
+// the default kukeond socket (or KUKEON_HOST when the operator overrides it)
+// and stays on the daemon path. The in-process branch remains reachable when
+// the viper key is set directly — programmatically or via KUKEON_NO_DAEMON —
+// which is also how the unit tests exercise it.
+//
+// The caller owns the returned Client and must Close it.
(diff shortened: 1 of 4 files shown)


eminwux:
**PR #638 Review — `fix(cli): route shell completion through the daemon client`**

LGTM — routes the resource-name completers through the same `kukeonv1.Client` selection (`kukeon/noDaemon` key + host-default fallback) that `kuke get …` already uses, so unprivileged shell users get candidates from the group-accessible daemon instead of an empty read of root-owned `/opt/kukeon`; each `List*` is bounded by a 2s timeout (verified by the degradation test) and the daemon-vs-on-disk repro + real-daemon e2e cover the bug. The unchecked `make dev-init` box is correctly N/A — client-side completion touches nothing the daemon-parity smoke guards.

PR #639, opened 2026-05-21
feat(controller): guarded-skip stale persist in realm/space/stack refresh

### LINKED ISSUE #636: controller/runner: extend guarded-skip persist to realm/space/stack refresh
Surfaced during review of #633 (non-blocking).

Phase-3 converted all four `Update*Metadata` writers to CAS+optimistic-token (`writeDocWithGeneration`), so a concurrent spec write now makes any of them return `errdefs.ErrStaleResource`. The cell reconcile/refresh path handles this gracefully via `persistCellStatusGuarded` (benign skip, `persisted=false`, nil err). The three sibling refresh paths do **not** — they surface the stale error as a hard error:

- `internal/controller/runner/refresh.go:138` (`RefreshRealm` → `UpdateRealmMetadata`)
- `internal/controller/runner/refresh.go:230` (`RefreshSpace` → `UpdateSpaceMetadata`)
- `internal/controller/runner/refresh.go:312` (`RefreshStack` → `UpdateStackMetadata`)

Effect: on a rare concurrent spec write during a refresh, `controller/refresh.go`'s `refreshRealmAndChildren`/`refreshSpaceAndChildren`/`refreshStackAndChildren` log a warning and append a `result.Errors` line (visible under `Errors:` in `kuke refresh`; daemon reconcile-loop warning). Functionally correct — no clobber, self-heals on the next tick, `kuke refresh` still exits 0 — but cosmetic/operational noise.

Suggested fix: extend the guarded-skip pattern (a generic `persist*StatusGuarded` or parameterized helper) to the realm/space/stack refresh persist sites so a benign stale race is treated as a skip rather than a reported error, matching the cell path.

eminwux:
## Summary

- Extends the guarded-skip persist pattern (`persistCellStatusGuarded`) to the realm/space/stack refresh paths, closing the gap #633 phase-3 left: those three `Update*Metadata` writers now ride a CAS+optimistic-token, so a concurrent spec write during a refresh returns `errdefs.ErrStaleResource`. Before this change the cell path absorbed that as a benign skip while `RefreshRealm`/`RefreshSpace`/`RefreshStack` surfaced it as a hard error — a `result.Errors` line under `Errors:` in `kuke refresh` and a daemon reconcile-loop warning. The race was already functionally correct (no clobber, self-heals next tick, `kuke refresh` still exits 0); this removes the cosmetic/operational noise.
- Adds `persistRealmStatusGuarded` / `persistSpaceStatusGuarded` / `persistStackStatusGuarded`, mirroring `persistCellStatusGuarded`: a stale CAS rejection is logged at debug and returned as `(persisted=false, nil)`, any other error bubbles up unchanged. The three refresh call sites now set `updated = persisted`, so a skipped stale write reports nothing-updated instead of erroring.

### Design call (please push back)

- The cell helper additionally stamps `Status.ObservedGeneration` before persisting. The three new helpers deliberately **do not** — the realm/space/stack refresh paths never stamped ObservedGeneration (only the cell reconcile/refresh path consumes it via `observedGenerationBehind`), and the issue scope is the stale-skip, not adding generation-convergence to those kinds. Kept minimal; flagging in case a reviewer wants the symmetry extended.

## Test plan

- [x] `make test` — full unit suite green, including 3 new tests (`TestPersist{Realm,Space,Stack}StatusGuardedSkipsWhenBehind`) that drive a concurrent spec write to advance the on-disk generation, then assert the guarded persist skips (no clobber, no error) for the stale view and persists for a current observation. Mirrors `TestPersistCellStatusGuardedSkipsWhenBehind`.
- [x] `go build ./...` + `go vet ./internal/controller/...` — clean.
- [x] `pre-commit run` (go-mod-tidy, golangci-lint, addlicense) — passed.
- [x] `make dev-init` **daemon-parity regression guard** — passed identically on both `kuke get realms` and `kuke get realms --no-daemon` views (the bind-mount regression check CLAUDE.md mandates):
  ```
  ```
- [ ] `make dev-init` **kuketty attach-smoke sub-step** — not green on this dev host; failed differently across two runs (kuketty socket not bound within the 20s window once; a tmpfile-`rename` `ENOENT` during `kuke create stack` the next). Non-deterministic ⇒ environmental flakiness in this sandbox, not a regression: this diff only changes the `ErrStaleResource` branch of the realm/space/stack **refresh** (reconcile) persist path, which is a literal no-op absent a concurrent-spec CAS race; the second failure is in the **create** path (`kuke create stack` → `UpdateStackMetadata` directly, not through the new guarded helpers). Reviewer should confirm the attach smoke on a host without the registry-egress / fs-race flakiness.

Closes #636

### DIFF
--- a/internal/controller/runner/metadata_generation_test.go
+++ b/internal/controller/runner/metadata_generation_test.go
@@ -186,6 +186,180 @@ func TestUpdateRealmMetadataConcurrentWritesConverge(t *testing.T) {
 	}
 }
 
+// TestPersistRealmStatusGuardedSkipsWhenBehind pins issue #636 for the realm
+// refresh path: when a concurrent spec write has advanced the on-disk
+// Generation past the value the refresh observed, the guarded persist skips
+// (persisted=false, nil err) rather than surfacing ErrStaleResource as a hard
+// error; a current observation persists normally. Mirrors the cell path's
+// persistCellStatusGuarded behavior.
+func TestPersistRealmStatusGuardedSkipsWhenBehind(t *testing.T) {
+	r := newGenerationTestExec(t, t.TempDir())
+	seed := intmodel.Realm{
+		Metadata: intmodel.RealmMetadata{Name: "r-guard"},
+		Spec:     intmodel.RealmSpec{Namespace: "r-guard.kukeon.io"},
+		Status:   intmodel.RealmStatus{State: intmodel.RealmStateReady},
+	}
+	if err := r.UpdateRealmMetadata(seed); err != nil {
+		t.Fatalf("seed UpdateRealmMetadata: %v", err)
+	}
+
+	// Capture the stale refresh view (generation 1) before the spec moves.
+	staleView := readRealmGeneration(t, r, "r-guard")
+
+	// A concurrent spec writer advances the realm to generation 2.
+	specWrite := readRealmGeneration(t, r, "r-guard")
+	specWrite.Spec.RegistryCredentials = append(specWrite.Spec.RegistryCredentials,
+		intmodel.RegistryCredentials{ServerAddress: "reg-a"})
+	if err := r.UpdateRealmMetadata(specWrite); err != nil {
+		t.Fatalf("concurrent spec UpdateRealmMetadata: %v", err)
+	}
+
+	// The refresh tries to persist a status flip against its stale view.
+	staleView.Status.State = intmodel.RealmStateUnknown
+	persisted, err := r.persistRealmStatusGuarded(staleView)
+	if err != nil {
+		t.Fatalf("guarded persist (stale): %v", err)
+	}
+	if persisted {
+		t.Fatal("guarded persist reported persisted=true for a stale observation")
+	}
+	afterSkip := readRealmGeneration(t, r, "r-guard")
+	if len(afterSkip.Spec.RegistryCredentials) != 1 {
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #639 Review — `feat(controller): guarded-skip stale persist in realm/space/stack refresh`**

LGTM — faithfully extends the `persistCellStatusGuarded` pattern to realm/space/stack refresh. Verified the guards are functional (not no-ops): `writeDocWithGeneration` (`internal/controller/runner/metadata.go:85`) returns `errdefs.ErrStaleResource` on an advanced on-disk generation, the three new helpers correctly `errors.Is`-match it as a benign skip and bubble all other errors, and `updated = persisted` reports nothing-updated on skip. Tests drive a real concurrent spec write and assert skip + no-clobber + current-observation-persists. Commit signed, no dead code, license header present.

The `Status.ObservedGeneration` design call is right: the realm/space/stack refresh persist conditions never consult ObservedGeneration (only the cell path does, via `observedGenerationBehind`), so leaving it unstamped is consistent — extending the symmetry would add an unused field write.

Non-blocking: the one unchecked test-plan box (`kuketty attach-smoke`) is documented host-gated flakiness, and the second failure the author saw is in the `create` path, not the guarded refresh helpers — `make e2e` passed in CI. Worth a confirming run on a clean host pre-merge.

PR #637, opened 2026-05-21
feat(run): complete kuke run -f create-or-attach state machine

eminwux:
## Summary

- Completes the `kuke run -f` create-or-attach state machine in `runExistingCell` (`cmd/kuke/run/run.go`). Against a live cell whose on-disk spec matches the file, the runtime state now selects the transition explicitly instead of falling every non-Ready state through to `CreateCell`:
  - **No live cell** → create + attach (unchanged).
  - **Ready** → no-op summary, then attach (unchanged).
  - **Stopped** → `StartCell`, then attach. The prior fall-through to `CreateCell` was itself an unsafe re-entry; `StartCell` is the correct verb.
  - **error / partial** (`Pending`, `Failed`, `Unknown`) → refuse with a `kuke delete cell <name>` pointer (new dedicated branch).
  - **Divergent spec** → unchanged hard refuse routing to `kuke apply -f`.
- Help text on `kuke run` now states `-f` is create-or-attach by `metadata.name` and enumerates the transitions.
- `docs/cli-use-cases.md` run-`-f` invariants updated to document the state machine (keeps the source-of-truth UX doc consistent with the shipped behavior).

## Design calls (reviewer please push back)

- **#630 dependency is scoped to AC1's runtime, not the whole issue.** The issue gates only the *stopped → start* transition on the CNI duplicate-allocation fix (#630), which is open with no PR. I did **not** stall the issue: the CLI routing (`StartCell` instead of `CreateCell`) is correct regardless, is fully unit-testable with a mock client, and is strictly no worse than today's stopped→`CreateCell` fall-through (both hit the same CNI bug until #630, so no regression). The live stopped→start path converges with **zero further CLI change** once #630 lands. The error/partial-refuse, divergent-spec, help-text, and no-cell/Ready ACs are fully independent of #630 and verified now.
- **error / partial = `Pending`, `Failed`, `Unknown`.** These three states map to the issue's "error / partial" bucket (parity with the `-c` identity contract in #625: running / stopped / error). Listed explicitly in the `switch` (no `default`) so the `exhaustive` linter forces any future `CellState` to be categorised deliberately.
[...]

PR #640, opened 2026-05-21
fix(daemon): release CNI allocation on root-container teardown before re-ADD

### LINKED ISSUE #630: daemon: release CNI allocation on root-container teardown before re-ADD
## Context

The daemon has two paths that delete a cell's **root** container and then recreate it under the *same deterministic* containerd ID (`naming.BuildRootContainerdID`), re-running CNI ADD against that ID. Neither runs CNI DEL on the way out, so host-local IPAM keeps the prior reservation and the re-ADD is rejected as a `duplicate allocation`. The root container is the only one that holds a CNI/IPAM allocation — `AddContainerToNetwork` is called once, against the root, at `internal/controller/runner/start.go:612`; non-root containers join the root's netns and never get their own ADD.

**Path 1 — `runner.StartCell`** (`internal/controller/runner/start.go`). When the cell's tasks are **not** all running, it deletes the existing root container and re-provisions it. The leak is documented inline at `start.go:352-353` ("we never run CNI DEL when we delete the old container, so the reservation persists"). The all-running case is papered over by the `#149` idempotency guard `cellTasksAllRunningFn` (`start.go:98`), which skips teardown entirely — but any **stopped or partially-failed** cell re-enters the teardown path and hits the leak. This blocks safe re-entry of `StartCell` on a non-Ready cell, surfaced at `cmd/kuke/run/run.go:246-250`.

**Path 2 — `runner.RecreateCell`** (`internal/controller/runner/recreate_cell.go`). Deletes the root container at `recreate_cell.go:123` with **no CNI DEL anywhere in the file**, then clears the containerd IDs and calls `createCellContainers`, which rebuilds the root under the same deterministic `BuildRootContainerdID` and re-runs CNI ADD → same `duplicate allocation`. This path is live: reached from `internal/controller/apply/reconcile.go:373` whenever a root-affecting spec change (image / command / args, per `internal/controller/apply/diff.go:367,520`) is reconciled. So `kuke apply`-ing an image change to a cell's root container can hit this today.

## Proposal

On both root-container teardown-before-recreate paths, run CNI DEL against the old root container **before** `DeleteContainer`, releasing the IPAM reservation so the subsequent CNI ADD succeeds. The release primitive already exists: `purgeCNIForContainer` (`internal/controller/runner/purge_cell.go:165`).

## Acceptance criteria

- [ ] `StartCell` on a stopped / non-running cell tears down and re-provisions without a `duplicate allocation` error.
- [ ] `RecreateCell` (root-affecting `kuke apply` → reconcile) tears down and recreates the root container without a `duplicate allocation` error.
- [ ] CNI DEL runs against the old root container ID before `DeleteContainer` on **both** the `start.go` teardown branch and the `recreate_cell.go:123` root delete.
- [ ] The `#149` all-running idempotent skip (`cellTasksAllRunningFn`, `start.go:98`) still short-circuits unchanged — no teardown, no CNI churn on a healthy cell.
- [ ] Unit/regression test covering re-entry on a stopped cell and a root-affecting recreate.

## Notes

- Root-only allocation confirmed: single `AddContainerToNetwork` at `start.go:612`; non-root containers share the root netns.
- Anchors: `start.go:98` (idempotency guard), `start.go:352-353` (leak comment), `recreate_cell.go:123` (root delete, no CNI DEL), `apply/reconcile.go:373` + `apply/diff.go:367,520` (recreate trigger), `purge_cell.go:165` (`purgeCNIForContainer`).
- Explicitly **not** gaps (checked, no fix needed): `start.go:745` and `start.go:1001` delete **non-root** containers (the StartCell loop `continue`s on root; `StartContainer` rejects root at `start.go:940`) — non-root containers have no CNI allocation to release.
- Referenced by `cmd/kuke/run/run.go:246-250`, which refuses to re-enter `StartCell` on a non-Ready cell because of this bug.
- Blocks the stopped → start+attach transition in #631 (`kuke run -f` state machine).

eminwux:
## Summary

Closes the `duplicate allocation` lost-reservation gap on the two daemon paths that delete a cell's **root** container and recreate it under the *same* deterministic `naming.BuildRootContainerdID`, re-running CNI ADD against that ID. Neither ran CNI DEL on the way out, so host-local IPAM kept the prior reservation and the re-ADD was rejected. The root is the only container holding a CNI/IPAM allocation (`AddContainerToNetwork` is called once, against the root, at `start.go`).

- **`StartCell` teardown-before-recreate branch** (`start.go`): when the cell's tasks are *not* all running, the container-exists `else` branch now releases the root's CNI/IPAM reservation *before* deleting it. The `#149` all-running idempotency guard (`cellTasksAllRunningFn`) is untouched and still short-circuits a healthy cell with no teardown/CNI churn.
- **`RecreateCell`** (`recreate_cell.go`): the root-delete now releases CNI/IPAM before the stop/delete, so a root-affecting `kuke apply` (image/command/args change → reconcile → `RecreateCell`) re-provisions without a duplicate allocation.
- Both paths route through a new `teardownRootContainerCNI(releaseCNI, deleteContainer, purgeCNI)` helper that enforces the ordering: **CNI DEL (`detachRootContainerFromNetwork`) before delete**, then the **IPAM-file purge (`purgeCNIForContainer`) after delete** as a safety net. This reuses the existing release primitives and mirrors the established `stop.go` / `kill.go` detach→delete→purge idiom.

## Design calls (reviewer please push back)

- **Extracted a pure ordering helper (`teardownRootContainerCNI`) rather than inlining like `stop.go`/`kill.go`.** The AC asks for a unit/regression test of the ordering, but `cni.CNINetworksDir` is a hard `const` (`/var/lib/cni/networks`) and `purgeCNIForContainer` builds a real `cni.Manager`, so a filesystem-level IPAM assertion can't be redirected to a tmpdir in-package. The pure-closure helper makes the *release-before-delete / purge-after* invariant unit-testable without a live runtime — the exact rationale the repo already documents for `cellTasksAllRunningFn` (issue #149). The live release is covered by the dev-init smoke below.
- **Added a shared `resolveRootCNINetworkName(realm, space)` helper** (`helpers.go`) to dedup the `GetSpace`+`getSpaceNetworkName` preamble both new sites need; it returns `""` best-effort so the downstream purge is a no-op rather than a panic when the space can't be loaded — matching the existing `stop.go`/`kill.go` preamble's tolerance.
- **`releaseCNI` is sequenced before the task delete** in the `StartCell` branch so `detachRootContainerFromNetwork` reads a still-valid netns for a real CNI DEL; the IPAM-file purge afterward is the backstop that actually releases the reservation when the netns is already gone (the `RecreateCell` case, where `StopCell` already killed the task).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full CI target, `go test $(go list ./... | grep -v /e2e)`) — all green
- [x] New regression tests in `internal/controller/runner/`:
  - `TestTeardownRootContainerCNI_OrderingAndSafetyNet` — release runs before delete, purge after; purge still runs and the delete error propagates when delete fails.
  - `TestResolveRootCNINetworkName_EmptyWhenSpaceMissing` — best-effort `""` (no panic) when space metadata is absent.
- [x] `pre-commit run --files <changed>` — passed (incl. golangci-lint `shadow`/`govet`).
- [x] `make dev-init` smoke + daemon-parity check — both `kuke get realms` and `--no-daemon` show identical output (`default` + `kuke-system`, both `Ready`); `kuke attach` against a kuketty-wrapped cell created cleanly and detached. Note: two earlier dev-init runs hit transient containerd content-store errors at fresh `CreateCell` image-pull (`lease does not exist` / `parent snapshot does not exist`) — confirmed environmental and **not** caused by this diff: a clean `origin/main` build hit the same family before stabilizing, a plain `ctr pull` of the same image succeeded, and the failing path (`CreateCell` image pull) is not touched by this change; the smoke passes cleanly on this build once containerd state settled.
- [ ] Live `duplicate allocation` repro on a real stopped cell re-entry / root-affecting `kuke apply` — not isolated end-to-end in dev-init (the smoke exercises create+attach+purge, not a stopped→start re-entry). Unblocks PR #637's deferred AC1, which is the live stopped→start path; covered at the ordering-invariant level by the new unit tests.

Closes #630

### DIFF
--- a/internal/controller/runner/helpers.go
+++ b/internal/controller/runner/helpers.go
@@ -451,6 +451,28 @@ func (r *Exec) getSpaceNetworkName(space intmodel.Space) (string, error) {
 	return naming.BuildSpaceNetworkName(realmName, space.Metadata.Name)
 }
 
+// resolveRootCNINetworkName resolves the CNI network name for a cell's space,
+// for use in the post-delete IPAM-file purge safety net (purgeCNIForContainer).
+// Best-effort: returns "" when the space can't be loaded or the name can't be
+// derived, mirroring the GetSpace+getSpaceNetworkName preamble that stop.go and
+// kill.go run before tearing the root container down. The empty result makes
+// purgeCNIForContainer a no-op rather than a panic.
+func (r *Exec) resolveRootCNINetworkName(realmName, spaceName string) string {
+	lookupSpace := intmodel.Space{
+		Metadata: intmodel.SpaceMetadata{Name: spaceName},
+		Spec:     intmodel.SpaceSpec{RealmName: realmName},
+	}
+	internalSpace, err := r.GetSpace(lookupSpace)
+	if err != nil {
+		return ""
+	}
+	networkName, err := r.getSpaceNetworkName(internalSpace)
+	if err != nil {
+		return ""
+	}
+	return networkName
+}
+
 // detachRootContainerFromNetwork detaches a root container from the CNI network.
 // It gets the container's PID and network namespace path, then calls CNI DEL to detach it.
 // This should be called before killing or stopping the container to ensure the namespace is still valid.
--- a/internal/controller/runner/helpers_test.go
+++ b/internal/controller/runner/helpers_test.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/eminwux/kukeon/internal/errdefs"
 )
@@ -93,5 +94,17 @@ func TestIsValidationError(t *testing.T) {
 				t.Errorf("isValidationError() = %v, want %v", got, tt.want)
 			}
 		})
+	}
+}
+
+// TestResolveRootCNINetworkName_EmptyWhenSpaceMissing pins the best-effort
+// contract the issue #630 teardown sites rely on: when the space metadata
+// can't be loaded, resolveRootCNINetworkName returns "" rather than panicking,
+// which makes the downstream purgeCNIForContainer safety net a no-op instead
+// of dereferencing a nil network name.
+func TestResolveRootCNINetworkName_EmptyWhenSpaceMissing(t *testing.T) {
+	r := newMetadataTestExec(t, t.TempDir(), time.Now())
+	if got := r.resolveRootCNINetworkName("default", "nonexistent-space"); got != "" {
+		t.Errorf("resolveRootCNINetworkName() = %q, want \"\" when space metadata is absent", got)
 	}
 }
--- a/internal/controller/runner/recreate_cell.go
+++ b/internal/controller/runner/recreate_cell.go
@@ -110,27 +110,56 @@ func (r *Exec) RecreateCell(desired intmodel.Cell) (intmodel.Cell, error) {
 		return intmodel.Cell{}, fmt.Errorf("failed to build root container containerd ID: %w", err)
 	}
 
-	_, err = r.ctrClient.StopContainer(internalRealm.Spec.Namespace, rootContainerID, ctr.StopContainerOptions{Force: true})
-	if err != nil {
-		r.logger.WarnContext(
-			r.ctx,
-			"failed to stop root container, continuing with deletion",
-			"container", rootContainerID,
-			"error", err,
-		)
-	}
+	// Release the root container's CNI/IPAM reservation before deleting it.
+	// createCellContainers below rebuilds the root under this same deterministic
+	// containerd ID and re-runs CNI ADD; without releasing first, host-local
+	// IPAM rejects the re-ADD as a duplicate allocation (issue #630). releaseCNI
+	// runs before the stop/delete so the netns is still valid for CNI DEL where
+	// applicable; purgeCNI scrubs the residual allocation file afterward.
+	cellName := strings.TrimSpace(existing.Metadata.Name)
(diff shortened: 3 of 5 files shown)


eminwux:
**PR #640 Review — `fix(daemon): release CNI allocation on root-container teardown before re-ADD`**

LGTM — both teardown-before-recreate paths (`StartCell` else-branch, `RecreateCell`) now route through `teardownRootContainerCNI`, enforcing CNI DEL → delete → IPAM-file purge. Verified the ordering reuses the existing `detachRootContainerFromNetwork`/`purgeCNIForContainer` primitives and mirrors the `stop.go`/`kill.go` idiom; `purgeCNIForContainer` already guards `netnsPath == ""`, so the empty-netns safety-net call is intentional and the IPAM-file scrub (keyed on `containerID`+`networkName`) is the real backstop. Helper-signature arg order matches both call sites, the `#149` all-running guard is untouched, both new symbols are used (no dead code), and all commits are verified. Regression tests pin the ordering + delete-failure safety-net contract; `make test`, `make e2e`, CodeQL all green.

One deferred verification for you to run pre-merge: the unchecked **live `duplicate allocation` repro** (stopped→start re-entry / root-affecting `kuke apply`) — host-gated, not isolatable in dev-init, covered at the ordering-invariant level by the new unit tests. Non-blocking.

Issue #648, opened 2026-05-21
cli: daemon start/restart fails after reboot — /run/kukeon bind source not recreated


eminwux:
## Context

`kuke daemon restart` (and `kuke daemon start`) fail after a host reboot:

```
$ kuke daemon restart
kukeond stopped (cell "kukeond" in realm "kuke-system")
Error: start kukeond cell: failed to start cell containers: failed to start container
kukeon_kukeon_kukeond_kukeond: failed to create task: failed to create shim task:
OCI runtime create failed: runc create failed: unable to start container process:
error during container init: failed to fulfil mount request:
open /run/kukeon: no such file or directory
```

The kukeond cell spec bind-mounts host `/run/kukeon` → container `/run/kukeon`. runc requires the bind **source** to exist. That directory is created exactly once, by `ensureSocketDir`, whose only non-test caller is the `kuke init` bootstrap path (`internal/controller/controller.go:212`, guarded by `if b.opts.KukeondImage != ""`). The assumption is documented in `cmd/kuke/init/init.go:346-348`: `/run/kukeon` is *"already created by ensureSocketDir during bootstrap."*

`kuke daemon restart` (`cmd/kuke/daemon/restart/restart.go:131`) and the underlying start path call `client.StartCell(...)` directly and never call `ensureSocketDir`. Because `/run` is a tmpfs, `/run/kukeon` is wiped on every reboot — so the bind source is gone and the restart fails. The workaround is a manual `mkdir /run/kukeon`, after which restart succeeds. This makes the daemon non-recoverable after a reboot without either a manual `mkdir` or a full re-`init`.

## Proposal

Ensure the socket-dir / bind-mount source (`/run/kukeon`, i.e. the parent dir of the kukeond socket) exists before the daemon start path launches the cell containers — not only during `kuke init` bootstrap. Lift the `ensureSocketDir` call (and the matching ownership step) into the start/restart path so it is idempotent across reboots, mirroring `applyKukeonOwnership`'s root:kukeon mode 0750 on `/run/kukeon`.

## Acceptance criteria

- [ ] `kuke daemon start` and `kuke daemon restart` recreate `/run/kukeon` if it is missing before starting the cell containers.
- [ ] The recreated directory has the same ownership/mode as `kuke init` produces (root:kukeon, 0750) so the kukeon group can traverse it.
[...]

Issue #224, opened 2026-04-29
architecture: converge create/delete on reconciler-driven model (post-#161)


eminwux:
## Context
`kuke create cell` (and the other create/delete paths) walks bring-up imperatively inside the request handler: cgroup → containerd namespace → container → task. The reconciler loop (#161) introduces a separate `Reconcile()` function for drift correction — and that function will, by definition, contain the same wiring logic.

Without convergence, the same bring-up logic exists in two places: the imperative request path and the reconciler. Two failure modes follow: (a) divergence as one path adds a step the other misses, and (b) restart-recovery code duplicates what the reconciler already does. The fix is to make `Reconcile()` the only place that wires resources, and have request paths call it.

## Proposal
Design lock (no code in this issue):

1. **#161's `Reconcile(resource)` must be callable synchronously from API request paths**, not just from the ticker. Concretely, it returns when the resource has reached its target state (or terminal failure), and it is safe to invoke against arbitrary partial state.
2. **Post-#161 migration plan** (separate landable issues, filed once #161 lands):
   - `kuke create cell` writes spec → calls `Reconcile(cell)` → returns the resulting status.
   - `kuke delete cell` sets `deletionTimestamp` on spec → calls `Reconcile(cell)` → returns when the resource is gone.
   - Imperative wiring code is removed as each kind migrates.
   - Cells migrate first (paired with #161 GA); stack/space/realm follow.
3. **One deletion model**: today's `kuke delete` + `kuke kill` separation is an artifact of imperative orchestration. In the reconciler world, `deletionTimestamp` + finalizers covers both. Migration filed separately when cell-level reconcile lands.
4. **Status semantics align with reality**: `Ready` means "reconciler observed it healthy as of `UpdatedAt`," not "we last touched it and it didn't error." Pairs with #166's `Reason`/`Message`.

## Acceptance criteria
- [ ] #161's design notes record the constraint that `Reconcile()` is synchronously callable from request paths.
- [ ] Post-#161 migration plan (per-kind issues) is filed under this umbrella before #161 GA.
- [ ] No code change required by this issue itself — close once the constraint is documented and follow-up issues are filed.

## Notes
- Part of umbrella #217.
- **Depends on #161** (reconciler v1 cell-only) and benefits from **#166** (status enrichment) + **#205** (metadata concurrency).
- This is a planning issue. The actual code migration ships as separate per-kind issues filed once #161 lands.

PR #657, opened 2026-05-21
fix(run): refuse phantom attach when containerd lost a Ready cell's containers

### LINKED ISSUE #654: run: reports containers "already existed"/"started" for a cell containerd no longer has
## Context

`kuke run` against an existing cell reports `container … already existed` and `containers: started`, then attaches — but the cell has **no containers and no tasks in containerd at all**. The attach therefore dials a socket backed by nothing and fails. kukeon's persisted metadata has diverged from containerd's actual state, and `run` trusts the stale metadata.

Reproduced consistently on `kukeon-pm-1`:

```
$ kuke run -p kukeon-pm --name kukeon-pm-1
Cell "kukeon-pm-1" (realm "default", space "default", stack "default")
  - metadata: already existed
  - cgroup: already existed
  - root container: created
  - containers: started
Error: error attaching client: ping failed: dial unix /opt/kukeon/s/cd44b9a5b1147025: connect: connection refused
```

State divergence, confirmed on the host:

- **Nothing in containerd for the cell.** `ctr -n default.kukeon.io container ls | grep kukeon-pm-1` → empty; `ctr -n default.kukeon.io task ls` lists `kukeon-pm-0`, `kukeon-pr-0/1`, `kukeon-dev-0/1` but **no `kukeon-pm-1` (neither root nor work)**. No shim/kuketty process in `ps` either.
- **kukeon metadata still records it.** `/opt/kukeon/data/default/default/default/kukeon-pm-1/...` persists; `run` reads it and prints `container "work": already existed` + `containers: started`.
- **Stale socket, no listener.** `work/tty/socket` is `srwxr-xr-x` (`0755`), no listener → `connection refused`. `work/tty/kuketty.log`'s last entry is `2026-05-21T02:04:40`; the 21:24 run produced no new log lines.

Likely trigger: containerd lost the cell's containers across a daemon/containerd restart or reboot (cf. #648) while kukeon's on-disk metadata survived. The host daemon in `ps` started `20:46`; `kukeon-pm-0` (created `21:23`, after the daemon came up) still exists in containerd, but `kukeon-pm-1` (last live `02:04`) does not. The rendered metadata is otherwise correct (`socketMode: 432` = `0660`, `socketGid: 988`), so this is **not** a socket-perm defect — it is a metadata↔containerd divergence that `run` fails to detect.

## Proposal

`kuke run` (and the underlying create/start path) must not report `container … already existed` / `containers: started` for containers that no longer exist in containerd, and must not attach to a phantom cell. Reconcile the recorded cell state against containerd before deciding the verb: if kukeon's metadata claims containers that containerd has lost, treat the cell as needing (re)creation, or refuse with a clear divergence message + recovery pointer (parity in spirit with the `Pending/Failed/Unknown` "delete and re-run" guidance), instead of a raw `connection refused` from a dead socket.

(Mechanism left to dev — this issue specifies observable behavior, not the reconcile design.)

## Acceptance criteria

- [ ] `kuke run` on a cell whose recorded containers are absent from containerd does not print `container … already existed` / `containers: started`.
- [ ] It either (re)creates+starts the missing containers, or exits with a clear "kukeon metadata and containerd have diverged" message and a recovery pointer — not a raw `connection refused` / `permission denied` from a stale socket.
- [ ] No regression of the #630 CNI duplicate-allocation guard on the create-fresh / `Ready` short-circuit paths.

## Notes

- Diagnosis confirmed: cell exists in kukeon's metadata but has **zero containers/tasks in containerd**. This is divergence detection missing in the run path, not a workload that exited or an attach race.
- Secondary variant of the same dead-socket family: a cell recorded `Ready` whose workload exited is not restarted by the `CellStateReady` no-op branch (`cmd/kuke/run/run.go:298`, kept to avoid #630 re-entry) and attaches to a `0755` pre-`applySocketPerms` socket → `permission denied`. Both paths should converge on "don't attach to a non-running / non-existent workload."
- Root cause sits in the metadata↔containerd reconciliation gap; broader reconciler-driven convergence is tracked in #224 / #217, and restart/reboot survival in #648. This issue is the narrow `run`-reports-success-for-phantom-containers symptom, landable independently.
- Separate latent finding (not this bug): nested `kukeon-dev` daemons run `--socket-gid 996` (host group `input`) while the host daemon uses `988` (`kukeon`), both sharing `--run-path /opt/kukeon`. A cell provisioned by a nested daemon would get a socket grouped `input`, unreachable by host `kukeon`-group members. Worth a separate investigation if nested-mode attach failures appear.

eminwux:
## Summary

`kuke run` on a cell recorded `Ready` whose containers containerd has lost (typically a daemon/host restart per #648 dropped the cell's containers while the on-disk metadata survived) printed phantom `container … already existed` / `containers: started`, then attached to a socket backed by nothing — surfacing as a raw `connection refused`.

The root cause: the `CellStateReady` branch in `runExistingCell` (`cmd/kuke/run/run.go`) trusted persisted metadata and short-circuited to the no-op summary + attach without reconciling against containerd. `noOpResultFromGet` hardcodes `ExistsPost: true` for every container, which is the phantom `already existed` output regardless of containerd reality.

The fix reconciles before deciding the verb. `GetCellResult.RootContainerExists` already reflects live containerd state (`runner.ExistsCellRootContainer` queries containerd directly and the value flows over the wire to the CLI). When the recorded state is `Ready` but `RootContainerExists` is false, `run` now refuses with a clear `kukeon metadata and containerd have diverged` message plus a `kuke delete cell <name>` recovery pointer — parity in spirit with the existing `Pending/Failed/Unknown` "delete and re-run" guidance.

### Non-obvious implementation calls (reviewer please push back)

- **Refuse, not recreate.** The AC permits either `(re)creates+starts the missing containers` *or* `exits with a clear divergence message + recovery pointer`. I chose refuse — the lower-blast-radius option. Recreating in place would re-enter `CreateCell`/`StartCell` on a cell whose CNI allocation may still be half-held, the exact #630 duplicate-allocation re-entry the `Ready` short-circuit exists to avoid; the operator-driven `kuke delete cell` releases any half-held allocation cleanly. This trivially satisfies AC bullet 3 (no #630 regression) since the path no longer re-enters create/start at all. The issue explicitly leaves the mechanism to dev.
- **Guard scoped to `Ready` only.** A legitimately `Stopped` cell has `RootContainerExists == false` *by design* — `StopCell` deletes the root container from containerd (`internal/controller/runner/stop.go:252`), and the `Stopped` path correctly recreates via `StartCell`. So `RootContainerExists` is only a divergence signal in the `Ready` state, where the container is asserted live. Applying the guard to `Stopped` would wrongly refuse every normal stopped cell.

### Out of scope (per AC)

The secondary variant in the issue Notes — a `Ready` cell whose *workload task* exited while the root container still exists in containerd (`RootContainerExists` is true, attach hits a `0755` pre-`applySocketPerms` socket → `permission denied`) — is **not** addressed here. The AC checkboxes are about *recorded containers absent from containerd*; detecting an exited-but-present workload requires task-state inspection (a broader change). Left for a follow-up under the same dead-socket family (#224 / #217 reconciler convergence, #648 restart survival).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./cmd/kuke/run/...`
- [x] `make test` (full CI target) — green
- [x] `pre-commit run --files <changed>` — all hooks pass (go fmt, go-mod-tidy, golangci-lint, addlicense, …)
- [x] **AC — no phantom output on divergence:** new `TestRun_ExistingCell_RecordedReady_ContainerdLostContainers_Refuses` asserts a `Ready` cell with `RootContainerExists: false` neither prints `already existed` / `containers: started` nor attaches, and refuses with the divergence message + `kuke delete cell` pointer.
- [x] **AC — clear divergence message, not a raw `connection refused`:** asserted on the error string (`diverged`, `kuke delete cell my-cell`).
- [x] **AC — no #630 regression:** asserted `CreateCell`/`StartCell` call counts are 0 on the refuse path; the healthy-`Ready` short-circuit (`RootContainerExists: true`) and `Stopped`→`StartCell` paths are unchanged and still green (`TestRun_ExistingCell_MatchingSpec_AlreadyReady_ShortCircuits`, `..._Stopped_StartsAndAttaches`).
- [ ] `make dev-init` daemon-parity smoke — not run: the diff is CLI-only (`cmd/kuke/run`), reads a `GetCellResult` field the daemon already computes and returns, and touches nothing the daemon reads, builds, or stores under `/opt/kukeon`. The daemon-parity guard is not applicable to this change.

Closes #654

### DIFF
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -272,10 +272,16 @@ func runRun(cmd *cobra.Command, args []string) error {
 // whose on-disk spec already matches the file (the caller rejected divergence
 // before delegating here). The four terminal states:
 //
-//   - Ready   → print a no-op summary, then attach. Re-entering the daemon's
-//     CreateCell→StartCell path on a healthy cell trips the runner's CNI
-//     duplicate-allocation bug (#630), so the short-circuit is mandatory, not
-//     just an optimisation.
+//   - Ready   → reconcile against containerd first. The recorded state asserts
+//     a live root container; if containerd has lost it (a daemon/host restart
+//     per #648 dropped the containers while the on-disk metadata survived), the
+//     no-op summary would report phantom `container … already existed` /
+//     `containers: started` and then attach to a dead socket
+//     (`connection refused`, #654). On that divergence, refuse with a
+//     delete-then-rerun pointer. Otherwise print a no-op summary, then attach.
+//     Re-entering the daemon's CreateCell→StartCell path on a healthy cell trips
+//     the runner's CNI duplicate-allocation bug (#630), so the short-circuit is
+//     mandatory, not just an optimisation.
 //   - Stopped → StartCell, then attach. The routing is the correct verb (the
 //     prior fall-through to CreateCell was itself an unsafe re-entry). The
 //     live start+attach is gated on the same CNI fix (#630): StartCell re-ADDs
@@ -295,6 +301,28 @@ func runExistingCell(
 ) error {
 	switch pre.Cell.Status.State {
 	case v1beta1.CellStateReady:
+		// Divergence guard (#654): the metadata records this cell as Ready, but
+		// containerd no longer has its root container — typically a daemon/host
+		// restart (#648) dropped the cell's containers while the on-disk
+		// metadata survived. Trusting the stale metadata would print phantom
+		// `already existed` / `containers: started` lines and then attach to a
+		// socket backed by nothing (`connection refused`). Refuse with a
+		// delete-then-rerun pointer instead. We deliberately do not recreate in
+		// place: re-entering CreateCell/StartCell here is the exact #630 CNI
+		// duplicate-allocation re-entry the Ready short-circuit exists to avoid,
+		// and the operator-driven delete releases any half-held allocation
+		// cleanly. (A legitimately Stopped cell has no root container in
+		// containerd by design — StopCell deletes it — so this guard is scoped
+		// to Ready, where the container is asserted live.)
+		if !pre.RootContainerExists {
+			return fmt.Errorf(
+				"cell %q is recorded Ready but its containers are gone from containerd "+
+					"(kukeon metadata and containerd have diverged); "+
+					"delete it with `kuke delete cell %s` before re-running",
+				cellDoc.Metadata.Name,
+				cellDoc.Metadata.Name,
+			)
+		}
 		if printErr := printRunResult(cmd, noOpResultFromGet(pre), flags.output); printErr != nil {
 			return printErr
 		}
--- a/cmd/kuke/run/run_test.go
+++ b/cmd/kuke/run/run_test.go
@@ -484,6 +484,81 @@ func TestRun_ExistingCell_MatchingSpec_AlreadyReady_ShortCircuits(t *testing.T) 
 	}
 }
 
+func TestRun_ExistingCell_RecordedReady_ContainerdLostContainers_Refuses(t *testing.T) {
+	t.Cleanup(viper.Reset)
+
+	// #654: the cell is recorded Ready on disk, but containerd no longer has
+	// its root container (a daemon/host restart dropped the containers while
+	// the metadata survived). Run must NOT print phantom `already existed` /
+	// `containers: started` lines and must NOT attach to the dead socket;
+	// instead it refuses with a divergence message + delete-then-rerun pointer.
+	existing := v1beta1.CellDoc{
+		Metadata: v1beta1.CellMetadata{Name: "my-cell"},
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #657 Review — `fix(run): refuse phantom attach when containerd lost a Ready cell's containers`**

LGTM — guard refuses a `Ready` cell whose `RootContainerExists` is false (verified live-containerd-sourced and faithfully carried over the jsonrpc boundary, so it won't misfire on healthy cells), correctly scoped to `Ready` only, no create/start re-entry so no #630 regression, and the new test asserts the no-phantom-output + no-attach contract.

PR #658, opened 2026-05-21
fix: recreate /run/kukeon bind source on daemon start/restart

### LINKED ISSUE #648: cli: daemon start/restart fails after reboot — /run/kukeon bind source not recreated

eminwux:
## Summary

- `kuke daemon start`/`restart` fail after a host reboot with `runc create failed: ... open /run/kukeon: no such file or directory`. The kukeond cell spec bind-mounts host `/run/kukeon` → container `/run/kukeon`, and runc requires the bind **source** to exist. That directory was created exactly once, by the controller bootstrap's `ensureSocketDir`, whose only non-test caller is `kuke init`. `/run` is a tmpfs, so the directory is wiped on every reboot and the start path (`client.StartCell`) never recreates it — leaving the daemon non-recoverable without a manual `mkdir` or a full re-`init`.
- New `lifecycle.EnsureSocketDir` recreates the socket's parent directory and re-asserts the same `root:kukeon` SGID `0750` ownership `kuke init` produces, then `runStart`/`runRestart` call it (via `ResolveEnsureSocketDir`) immediately before `client.StartCell`. Idempotent: `MkdirAll` no-ops on a healthy host and the ownership/mode are simply re-applied.

### Non-obvious implementation calls (reviewer please push back)

- **Where the call lives.** `controller.StartCell` is generic across *all* cells (user cells included), so recreating `/run/kukeon` there would be wrong. Both lifecycle verbs already funnel through `client.StartCell` on the kukeond cell doc and share the `lifecycle` package, so the recreate step lives there and is invoked only on the kukeond start/restart path. `kuke init` keeps its own bootstrap `ensureSocketDir` — this change does not touch that path.
- **Directory derived from the resolved socket, not a hard-coded `/run/kukeon`.** `EnsureSocketDir(lifecycle.ResolveSocketPath())` uses `filepath.Dir` of the configured kukeond socket, so nested dev-init mode (which redirects the socket to `/run/kukeon-dev/`) recreates the directory the persisted cell spec actually binds. Verified against `internal/controller/bootstrap.go:463-465`, where the cell's bind source is `socketDir(socketPath)`.
- **Single source of truth for the mode.** Moved `init`'s private `kukeonRunDirMode` (`os.ModeSetgid | 0o750`) to `consts.KukeonRunDirMode` so the bootstrap chown and the new recreate path cannot drift on what `kuke init` produces. `init.go`'s other run-path modes are left local (different semantic). The AC phrases the mode as "root:kukeon, 0750"; the literal value `kuke init` applies is SGID-`0750` (mode `2750`), so this PR matches the exact init behavior, SGID bit included.
- **Two test seams.** `SetRunDirOwnershipForTesting` lets the colocated unit test exercise `MkdirAll`/idempotency without root or a real kukeon group (the production ownership step chowns to uid 0). `EnsureSocketDirKey` is a context-injection seam (same idiom as `MockClientKey`/`ReachableProbeKey`) so the existing start/restart command tests inject a no-op rather than touching the real `/run/kukeon`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full CI target) — green
- [x] `pre-commit run --files <changed>` — all hooks pass (go fmt, go-mod-tidy, golangci-lint, addlicense, …)
- [x] New colocated unit tests (`cmd/kuke/daemon/internal/lifecycle/socketdir_test.go`): creates a missing parent, idempotent on an existing dir, propagates the ownership error.
- [x] `make dev-init` daemon-parity check green (both realms Ready; both `kuke get realms` views identical) — ran in nested mode (socket `/run/kukeon-dev/`).
- [x] **AC #1 + #3 — reboot recovery (end-to-end, real containerd + daemon):** after `make dev-init`, `rm -rf /run/kukeon-dev` (simulates the tmpfs clear), then `kuke daemon restart` → **exit 0** (`kukeond stopped` / `kukeond started`); previously failed with `open /run/kukeon: no such file or directory`. Directory recreated.
- [x] **AC #2 — ownership/mode:** recreated dir is `drwxr-s--- root kukeon` (SGID `0750`, mode `2750`), matching `kuke init`.
- [x] **AC #4 — idempotent on healthy host:** re-running `kuke daemon restart` and `kuke daemon start` on the now-present dir succeeds (`already running`) and leaves ownership/mode intact (`drwxr-s--- root kukeon`).

No CLI flags or env-var keys were added/removed, so the literal-flag-string grep does not apply.

Closes #648

### DIFF
--- a/cmd/kuke/daemon/internal/lifecycle/socketdir.go
+++ b/cmd/kuke/daemon/internal/lifecycle/socketdir.go
@@ -1,0 +1,102 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package lifecycle
+
+import (
+	"fmt"
+	"os"
+	"os/user"
+	"path/filepath"
+	"strconv"
+
+	"github.com/eminwux/kukeon/internal/consts"
+	"github.com/eminwux/kukeon/internal/sysuser"
+	"github.com/spf13/cobra"
+)
+
+// EnsureSocketDirKey injects a socket-dir recreate func via the command
+// context so unit tests can drive `kuke daemon start`/`restart` without root
+// or a real /run/kukeon. Same pattern as MockClientKey / ReachableProbeKey:
+// the key lives here so every verb's resolver agrees on a single identity.
+type EnsureSocketDirKey struct{}
+
+// ResolveEnsureSocketDir picks the recreate-/run/kukeon step the start and
+// restart verbs run before launching the cell. Tests inject a stub via
+// EnsureSocketDirKey; production recreates the parent directory of the
+// resolved kukeond socket. The indirection lets both verbs call the same
+// one-liner without duplicating the context-value lookup.
+func ResolveEnsureSocketDir(cmd *cobra.Command) func() error {
+	if f, ok := cmd.Context().Value(EnsureSocketDirKey{}).(func() error); ok && f != nil {
+		return f
+	}
+	return func() error { return EnsureSocketDir(ResolveSocketPath()) }
+}
+
+// applyRunDirOwnership re-asserts root:kukeon ownership and the kukeon SGID
+// mode on the kukeond socket's parent directory. It is a package var so the
+// unit suite can substitute a non-root stub: the production implementation
+// chowns to uid 0, which only root may do, and resolves the kukeon group that
+// `kuke init` provisions on the host but a CI test box does not have.
+//
+//nolint:gochecknoglobals // test seam for the production run-dir ownership step
+var applyRunDirOwnership = func(dir string) error {
+	grp, err := user.LookupGroup(consts.KukeonSystemGroup)
+	if err != nil {
+		return fmt.Errorf("lookup group %q: %w", consts.KukeonSystemGroup, err)
+	}
+	gid, err := strconv.Atoi(grp.Gid)
+	if err != nil {
+		return fmt.Errorf("parse gid %q for group %q: %w", grp.Gid, consts.KukeonSystemGroup, err)
+	}
+	return sysuser.ChownAndChmod(dir, 0, gid, consts.KukeonRunDirMode)
+}
+
+// SetRunDirOwnershipForTesting replaces the ownership step EnsureSocketDir
+// runs and returns a restore func. Tests use it to assert the directory the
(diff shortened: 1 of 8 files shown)


eminwux:
**PR #658 Review — `fix: recreate /run/kukeon bind source on daemon start/restart`**

LGTM — recreates the tmpfs-wiped bind source idempotently on the kukeond start/restart path only, deriving the dir from the resolved socket and re-asserting `kuke init`'s exact SGID `2750` mode via the now-shared `consts.KukeonRunDirMode`; init's bootstrap path is untouched and the controller's generic `StartCell` is correctly left alone.

Issue #656, opened 2026-05-21
kukebuild: route the root gate through shared.RequireRoot / ErrMustRunAsRoot


eminwux:
Surfaced as a non-blocking nit during review of #652.

`cmd/kukebuild/build.go:106-109` gates the build on root with a raw check and an inline error:

```go
if os.Geteuid() != 0 {
    return errors.New(
        "kukebuild writes to root-owned containerd state — re-run as root (e.g. via `sudo kuke build`)",
    )
}
```

The repo already has the established mechanism for this exact concern: `cmd/kuke/shared.RequireRoot(subcommand)` wraps the `errdefs.ErrMustRunAsRoot` sentinel (`internal/errdefs/errdefs.go:307`) and uses a testable geteuid seam (`SetGeteuidForTesting`). Its docstring explicitly lists `kuke image load` as a gated "direct-write subcommand" — the exact "same posture as `kuke image load --no-daemon`" the kukebuild comment cites.

The inline form:
- bypasses `errdefs.ErrMustRunAsRoot`, so `errors.Is(err, errdefs.ErrMustRunAsRoot)` won't classify kukebuild's root error (per kukeon's CLAUDE.md errdefs-sentinel convention);
- skips the geteuid test seam that every sibling daemon command uses, leaving the non-root branch untestable as written.

Suggested fix: call `shared.RequireRoot("kuke build")` (its closure is `fmt`/`os`/`errdefs` only — BuildKit-free, and `cmd/kuke/shared` does not trip the `cmd/(kuke|kukeond)$` import-isolation AC anchor), or at minimum return `errdefs.ErrMustRunAsRoot` so the sentinel is reused.

Non-blocking; internal consistency + testability.

PR #662, opened 2026-05-21
build: split kukebuild into its own Go module to unpin BuildKit

### LINKED ISSUE #655: build: split kukebuild into its own Go module to unpin BuildKit
## Context

`kukebuild` embeds BuildKit, pulling a ~160-package transitive closure (moby/runc/grpc/genproto). The runtime **binary** isolation is already handled by the exec-split (#522): `cmd/kukebuild` is a separate `main` reached via `syscall.Exec` on PATH (`cmd/kuke/build/build.go`), and `kuke`'s linked closure pulls 0 BuildKit packages. So this is **not** about binary bloat.

The one coupling a shared module does *not* insulate against is the **Go toolchain / `go`-directive**. #522 pinned BuildKit to `v0.25.2` because v0.26+ requires `go 1.25`, which would drag the whole module's `go` directive (and the `golang:1.24-bookworm` builder in the repo `Dockerfile`) up for `kuke`/`kukeond`/`kuketty` too.

**That ceiling has now broken.** Dependabot #659 bumps `moby/buildkit v0.25.2 -> v0.28.1` — carrying GHSA security fixes (GHSA-4vrq-3vrq-g6gg, GHSA-4c29-8rgm-jvjj) — and the diff forces `go 1.24.5 -> go 1.25.5` on the shared `go.mod`. The `v0.25.2` soft ceiling can no longer hold without dragging the shared toolchain. The split is now warranted.

## Non-goals

- Binary-closure isolation — already delivered by the exec-split (#522).
- Removing/deprecating kukebuild — `rm -rf cmd/kukebuild cmd/kuke/build && go mod tidy` already drops the BuildKit closure cleanly; that path never needed a split.

## Proposal

Give `cmd/kukebuild` its **own `go.mod`** (a second module in the same repo; add a `go.work` for local dev). Its `go` directive and BuildKit version then advance independently: the kukebuild module moves to `go 1.25` and takes BuildKit `v0.28.1+`, while the **root module stays on `go 1.24.5`** and carries 0 BuildKit packages.

**Cross-module surface — resolve via the public config contract, not a `pkg/` promotion.** The only symbol kukebuild imports across the boundary today is `consts.RealmNamespace` (`cmd/kukebuild/build.go:124`) — a derivation `realm + RealmNamespaceSuffix` over a process-global suffix. Rather than promote `RealmNamespace` + `ConfigureRuntime` to `pkg/` (which would make the realm->namespace derivation semver-committed public API), make kukebuild **resolve the suffix itself** from the same operator-facing config kukeond reads, in this precedence:

1. `--kukeond-config <path>` flag, if set;
2. else `/etc/kukeon/kukeond.yaml` (the `DefaultServerConfigurationFile()` path);
3. else (file absent) a hardcoded default **equal to kukeond's** (`containerdNamespaceSuffix: kukeon.io`).

kukebuild then joins `realm + "." + suffix` itself and drops its `internal/consts` import entirely. The coupling moves from a **private Go symbol** (`internal/consts`) to the **public `kukeond.yaml` config key** `containerdNamespaceSuffix` — already a stable operator contract (its schema lives in `pkg/api/model/v1beta1`, so kukebuild can reuse that public struct for the parse, or define a minimal local struct to stay fully dependency-free — dev's call). `internal/consts` stays private; no `pkg/` promotion.

This also closes a latent gap: today `kuke build` calls neither `ConfigureRuntime` nor any config load, so it silently ignores an operator-configured non-default suffix and always writes to `<realm>.kukeon.io`. The self-resolution above makes the round-trip (build under a non-default suffix -> `kuke create` resolves the image) actually work.

`errdefs` does **not** need promoting: `ErrKukebuildNotFound` is used only by the `kuke build` shim (`cmd/kuke/build/`), which stays in the root module.

## Acceptance criteria

- [ ] `cmd/kukebuild` builds from its own `go.mod` on `go 1.25`; the root module's `go.mod`/`go.sum` stay on `go 1.24.5` and no longer carry the BuildKit closure.
- [ ] `go.work` wires both modules for local dev; `make` targets and CI build both.
- [ ] BuildKit advances to `v0.28.1+` in the kukebuild module (takes the #659 security fixes) without touching the root module's `go` directive.
- [ ] kukebuild resolves the realm-namespace suffix from `--kukeond-config` -> `/etc/kukeon/kukeond.yaml` -> hardcoded `kukeon.io` default (equal to kukeond's `DefaultRealmNamespaceSuffix`); the `kuke build` shim forwards `--kukeond-config` alongside its existing flags.
- [ ] kukebuild no longer imports `internal/consts` (or any kukeon `internal/` package); `internal/consts` stays private — no `pkg/` promotion of `RealmNamespace`/`ConfigureRuntime`.
- [ ] Round-trip: `kuke build` under a non-default `containerdNamespaceSuffix` writes the image into `<realm>.<suffix>` and `kuke create` resolves it.
- [ ] Import-graph isolation still holds: `kuke`/`kukeond` link 0 BuildKit packages.

## Suggested slicing

A single PR is feasible and preferred — the module extraction, the `go.work`/CI wiring, and the self-resolving suffix hang together. If it runs large, the smallest landable first slice is the **module split preserving today's behavior** (kukebuild keeps a hardcoded `kukeon.io` default, no config-file read), with the **`--kukeond-config` precedence + non-default-suffix round-trip as a fast follow** — split only if the wiring proves heavy; the suffix read itself is small.

## Notes

- Trigger: #659 (Dependabot BuildKit `0.25.2 -> 0.28.1`, forces `go 1.24.5 -> 1.25.5`). Parent series: #522 (phase 1, the exec-split + `v0.25.2` pin this lifts). Related build phases: #523, #524, #525, #616, #646.
- **Interaction with #656** (route kukebuild's root gate through `shared.RequireRoot` / `ErrMustRunAsRoot`): the split argues *against* sharing that helper into kukebuild — a stateless `os.Geteuid()` root gate is safe to keep inline in the kukebuild module, whereas importing a shared `internal/` helper re-introduces the cross-module coupling this issue removes. Reconcile when whichever lands second.
- Priority raised to A: this now gates a security-fix dependency bump (#659), and the split is the only way to take it without bumping the entire repo to `go 1.25`.

eminwux:
## Summary

- **Why:** `kukebuild` embeds BuildKit, whose `v0.26+` releases require `go 1.25`. Dependabot #659 bumps `moby/buildkit v0.25.2 -> v0.28.1` (carrying GHSA-4vrq-3vrq-g6gg / GHSA-4c29-8rgm-jvjj security fixes) and forces `go 1.24.5 -> 1.25.5` on the shared `go.mod`. Splitting `cmd/kukebuild` into its own module lets BuildKit + its `go 1.25` floor advance independently while the root module (`kuke`/`kukeond`/`kuketty`) stays on `go 1.24.5` and links **zero** BuildKit packages. This PR gates the #659 security bump.
- `cmd/kukebuild` now has its own `go.mod` (`module github.com/eminwux/kukeon/cmd/kukebuild`, `go 1.25.5`, `moby/buildkit v0.28.1`). The root `go.mod` stays `go 1.24.5` with the entire BuildKit closure removed.
- A repo-root `go.work` wires both modules for editor / gopls navigation.
- kukebuild now self-resolves the realm->namespace suffix from the **public `kukeond.yaml` `containerdNamespaceSuffix` key** (precedence: `--kukeond-config` -> `/etc/kukeon/kukeond.yaml` -> hardcoded `kukeon.io`), dropping its `internal/consts` import. The coupling moves from a private Go symbol to the public config contract; `internal/consts` stays private (no `pkg/` promotion). The `kuke build` shim forwards a new `--kukeond-config` flag.
- This also closes a latent gap: `kuke build` previously ignored an operator-configured non-default suffix and always wrote to `<realm>.kukeon.io`.

## Non-obvious design calls (reviewer, please push back)

1. **Workspace builds are deliberately per-module (`GOWORK=off`), not workspace-union.** The two modules have intentionally incompatible dependency graphs (BuildKit v0.28.1's closure vs. the root's older one — concretely the monolithic vs. split `google.golang.org/genproto`). A `go build ./...` through the workspace fails with an ambiguous import — the exact coupling this split removes. So the Makefile exports `GOWORK := off` and every build/test runs per-module, matching the single-module resolution CI and the Dockerfile builder (no `go.work`) already use. `go.work` is documented (header comment) as editor-only; `go work sync` was rejected because it bumps the root to `go 1.25.0` (violating the AC) and doesn't fix the ambiguity anyway.
2. **`go.work.sum` is gitignored, `go.work` committed.** Builds don't use the workspace, so the union sum has no reproducibility role and only drifts; go regenerates it on demand for editor tooling.
3. **kukebuild parses a minimal local `serverConfigDoc` struct** (just `spec.containerdNamespaceSuffix`) rather than importing `pkg/api/model/v1beta1`, keeping the module free of *any* kukeon package import (the issue offered either; chose the fully-decoupled option). The YAML key is kept in lockstep with the public struct via a doc comment.
4. **`release-build` no longer cross-builds kukebuild;** added a separate `release-build-kukebuild` target. The kukeond container image never shipped kukebuild (the Dockerfile copies only `kuke` + `kuketty`) and it's not a published release asset, so building it in `release-build` would only force the `go 1.25` + BuildKit closure into the `golang:1.24` Docker builder (which has no `go.work`).

## Test plan

- [x] `make test` — both modules green (root unit tests + `cd cmd/kukebuild && go test ./...`).
- [x] `GOWORK=off go build ./cmd` (root) + `make kuke kuketty kukebuild` — all build.
- [x] `go vet ./...` on both modules — clean.
- [x] **Import-graph isolation:** `go list -deps ./cmd` links **0** `moby/buildkit` / `tonistiigi/fsutil` packages; root `go.mod` has the BuildKit closure removed and stays `go 1.24.5`.
- [x] kukebuild imports **no** kukeon package (`go list -deps` shows only its own module path).
- [x] **`docker build -t kukeon-local:dev .`** — succeeds with the new module layout (Dockerfile builder, `golang:1.24`, no `go.work`; nested `cmd/kukebuild` module correctly ignored).
- [x] **`make dev-init`** (daemon-parity regression guard) — both realms `Ready` under `kuke get realms` *and* `--no-daemon`; attach smoke + parent-plumbing checks passed.
- [x] **Real `kuke build` round-trip (BuildKit v0.28.1 runtime):** built a 2-step Dockerfile; image landed in `default.kukeon.io` (default suffix) and, with `--kukeond-config` setting `dev.kukeon.io`, in `default.dev.kukeon.io` (clean `--root` build completed fully). Confirms suffix resolution routes to the correct namespace.
- [ ] `kuke create cell` instantiation from the built image — not driven to completion (nested-dev socket redirect + create-cell spec wiring is orthogonal CLI surface). The round-trip's read half (realm-image resolution) is pre-existing, unchanged behavior; the write half (image landing in the resolvable `<realm>.<suffix>` namespace with the normalized `docker.io/library/...` name) is verified above.

## Acceptance criteria

- [x] `cmd/kukebuild` builds from its own `go.mod` on `go 1.25`; root `go.mod`/`go.sum` stay `go 1.24.5`, BuildKit closure gone.
- [x] `go.work` wires both modules; `make` targets + CI build/test both.
- [x] BuildKit at `v0.28.1` in the kukebuild module without touching the root `go` directive.
- [x] kukebuild resolves the suffix from `--kukeond-config` -> `/etc/kukeon/kukeond.yaml` -> hardcoded `kukeon.io`; `kuke build` shim forwards `--kukeond-config`.
- [x] kukebuild no longer imports `internal/consts` (or any kukeon `internal/`); `internal/consts` stays private.
- [x] Round-trip: build under a non-default suffix writes into `<realm>.<suffix>` (verified `default.dev.kukeon.io`).
- [x] Import-graph isolation: `kuke`/`kukeond` link 0 BuildKit packages.

## Note on #656

Per the issue, this split argues *against* routing kukebuild's root gate through a shared `internal/` helper (it would re-introduce the cross-module coupling). The inline `os.Geteuid()` gate in `runBuild` is left as-is; reconcile with #656 when the second of the two lands.

Closes #655


eminwux:
**PR #662 Review — `build: split kukebuild into its own Go module to unpin BuildKit`**

LGTM — clean module split that lifts BuildKit's `go 1.25` floor out of the root module (root `go.mod` stays `go 1.24.5`, BuildKit closure removed, `go.work` is editor-only and `GOWORK=off` everywhere builds run). The realm→namespace suffix decoupling is correct: `validateSuffix` matches `consts.ConfigureRuntime` character-for-character on non-empty input, and the empty/absent-key fallback to `kukeon.io` matches kukeond's viper-default behavior, so an operator-configured suffix resolves to the same `<realm>.<suffix>` namespace on both sides. Dockerfile never copies `go.work`, so the `golang:1.24` builder keeps single-module resolution. Tests cover all resolution branches and the flag handoff; no dead code, no leaks.

PR #666, opened 2026-05-21
fix: scope kukebuild's default build root per target namespace

### LINKED ISSUE #663: build: kuke build reuses one BuildKit --root across containerd namespaces, failing unpack into the second
## Context

Surfaced while validating #655 (PR #662, the kukebuild module split). Independent of that PR — reproduces against `main`'s kukebuild as well, since the trigger is selecting two different target containerd namespaces while sharing kukebuild's default `--root` (`/var/lib/kukebuild`), which is reachable today via two different `--realm` values alone.

## Repro

`kukebuild`'s default build-state root is a single host path (`/var/lib/kukebuild`, `cmd/kukebuild/main.go` `defaultBuildRoot`) regardless of the target containerd namespace. Build the same context into two different namespaces back-to-back with that shared root:

```
# build 1 -> lands in default.kukeon.io  (OK)
kuke build -t app:dev --realm default <ctx>

# build 2 -> different target namespace, SAME /var/lib/kukebuild root
#           (here via a non-default containerdNamespaceSuffix; on main the
#            same condition is reachable with a different --realm, e.g.
#            --realm kuke-system, which maps to kuke-system.kukeon.io)
kuke build -t app:dev --realm default --kukeond-config <yaml-with dev.kukeon.io> <ctx>
```

`<ctx>/Dockerfile`:
```
FROM busybox:latest
RUN echo hello > /msg
```

Building build 2 with a **fresh** `--root` (e.g. `--root /tmp/freshroot`) succeeds cleanly, confirming the shared root is the trigger.

## Observed

Build 2 exports the manifest into the correct namespace but then fails the `unpack: true` step:

```
#6 exporting manifest sha256:... done
#6 naming to docker.io/library/app:dev done
#6 unpacking to docker.io/library/app:dev
#6 ERROR: failed to extract layer sha256:7f74ca72...: failed to get reader from content store: content digest sha256:481282af...: not found
kukebuild: build "app:dev": failed to solve: failed to extract layer ...: not found
```

The image manifest is named into namespace 2, but the layer blob is missing from namespace 2's content store, so the unpack leaves a non-runnable image. A fresh `--root` produces a fully-unpacked image (`#6 DONE`).

## Root cause (candidate — not fully pinpointed)

BuildKit's build cache lives under `--root`; the containerd content store is **per-namespace**. After build 1, the shared `--root` cache records the layer as already materialized (it was unpacked into namespace 1's content store). Build 2 reuses that cache entry, skips re-fetching the blob, and the `image` exporter's `unpack: true` step then can't find the blob in namespace 2's content store. With a fresh `--root` there is no cache hit, so the blob is fetched/unpacked into namespace 2 and the export succeeds.

So kukebuild's process-global, namespace-agnostic default `--root` is unsafe whenever consecutive builds target different namespaces.

## Suggested fix

Scope kukebuild's default build root per target namespace — e.g. derive `defaultBuildRoot` as `/var/lib/kukebuild/<namespace>` (joining the resolved `<realm>.<suffix>`) so each namespace gets an isolated cache and the unpack always finds its blob. Alternatively, force the `image` exporter to re-materialize layers into the target content store on export rather than trusting the shared cache. The per-namespace-root option is the smaller change and matches the fact that the containerd content store the cache mirrors is itself per-namespace.

## References

- Surfaced alongside #655 / PR #662 (kukebuild module split + `--kukeond-config` suffix resolution). Not fixed by that PR.
- `cmd/kukebuild/main.go` `defaultBuildRoot`; `cmd/kukebuild/build.go` `newWorkerController` (worker `Root` + `Namespace`) and the `unpack: "true"` export in `newSolveOpt`.

eminwux:
## Summary
- kukebuild's default BuildKit state root (`--root`, `/var/lib/kukebuild`) was a single host path shared across every target containerd namespace. BuildKit's cache under `--root` mirrors a single content store, but containerd's content store is **per-namespace** — so a second build into a *different* namespace reuses the first build's cache entry, skips re-materializing the layer into namespace 2's store, and the `unpack: "true"` export then fails with `content digest …: not found` (issue #663). Reachable today via two different `--realm` values, or via `--kukeond-config` with a non-default `containerdNamespaceSuffix` (the #662 path).
- Fix: scope the **default** build root per target namespace — `<defaultBuildRoot>/<namespace>` (e.g. `/var/lib/kukebuild/default.kukeon.io`) — so each namespace gets an isolated cache and the unpack always finds its blob. This is the smaller of the two options the issue proposed and matches the fact that the content store the cache mirrors is itself per-namespace.
- An operator-supplied `--root` is honored **verbatim** (opt-out of scoping); keeping it unique per namespace is then the operator's responsibility. A new `rootExplicit` flag on `buildConfig` (set via `flag.FlagSet.Visit`) distinguishes the two. The pure `resolveBuildRoot(root, explicit, namespace)` helper does the join and is unit-tested directly. Namespace resolution was moved ahead of the `MkdirAll` so the scoped path is created.

### Note for the reviewer
The `kuke build` CLI shim (`cmd/kuke/build/build.go` `buildArgv`) does **not** forward `--root` at all, so every `kuke build` invocation uses kukebuild's default root — exactly the path this fix scopes. No shim change is needed; verified by reading `buildArgv` (it forwards only `--tag/--realm/--file/--kukeond-config/--build-arg`).

## Test plan
- [x] `make test` (root module + `cd cmd/kukebuild && go test ./...`) — all green.
- [x] `cd cmd/kukebuild && GOWORK=off go build ./... && go vet ./...` — clean (GOWORK=off per the repo-root `go.work` contract: the workspace union deliberately fails on the genproto split).
- [x] New unit tests: `TestResolveBuildRoot` (default-scoped / custom-suffix / explicit-verbatim cases) + `TestResolveBuildRootIsolatesNamespaces`; `rootExplicit` assertions added to `TestParseArgsDefaults` / `TestParseArgsAllFlags`.
- [x] **Functional repro, fixed path:** built `app:dev` into `default.kukeon.io` then `kuke-system.kukeon.io` back-to-back with the default root (as root, against `/run/containerd/containerd.sock`). Both `#6 unpacking … done / #6 DONE`; on-disk roots resolved to isolated `/var/lib/kukebuild/default.kukeon.io/` and `/var/lib/kukebuild/kuke-system.kukeon.io/`.
- [x] **Functional repro, bug confirmation:** same two namespaces with a *shared explicit* `--root` still reproduces `failed to extract layer …: content digest …: not found` — confirming the root cause is precisely the shared root and that explicit `--root` is honored verbatim.

Closes #663

### DIFF
--- a/cmd/kukebuild/build.go
+++ b/cmd/kukebuild/build.go
@@ -118,15 +118,23 @@ func runBuild(ctx context.Context, cfg *buildConfig, progressW io.Writer) error 
 	if _, err := os.Stat(cfg.dockerfile); err != nil {
 		return fmt.Errorf("dockerfile %q: %w", cfg.dockerfile, err)
 	}
-	if err := os.MkdirAll(cfg.root, buildStateDirMode); err != nil {
-		return fmt.Errorf("create build state dir %q: %w", cfg.root, err)
-	}
 
 	suffix, err := resolveNamespaceSuffix(cfg.kukeondConfig)
 	if err != nil {
 		return err
 	}
 	namespace := realmNamespace(cfg.realm, suffix)
+
+	// Scope the default build state root per target namespace before creating
+	// it: BuildKit's cache under --root mirrors a single containerd content
+	// store, but that store is per-namespace, so a default root shared across
+	// namespaces makes build 2 reuse build 1's cache entry and skip
+	// re-materializing the layer into namespace 2's store — the unpack then
+	// can't find the blob (issue #663).
+	cfg.root = resolveBuildRoot(cfg.root, cfg.rootExplicit, namespace)
+	if err := os.MkdirAll(cfg.root, buildStateDirMode); err != nil {
+		return fmt.Errorf("create build state dir %q: %w", cfg.root, err)
+	}
 
 	ctrl, cleanup, err := newController(ctx, cfg, namespace)
 	if err != nil {
@@ -183,6 +191,22 @@ func runBuild(ctx context.Context, cfg *buildConfig, progressW io.Writer) error 
 		return fmt.Errorf("build %q: %w", cfg.tag, waitErr)
 	}
 	return nil
+}
+
+// resolveBuildRoot returns the directory BuildKit uses for its own state for a
+// build into namespace. The default root is scoped per namespace
+// (<root>/<namespace>) so consecutive builds targeting different containerd
+// namespaces never share a single BuildKit cache: a shared cache records a
+// layer as already materialized in the first namespace's content store, and
+// the `unpack: "true"` export into the second namespace then fails to find the
+// blob in that namespace's (separate) store — issue #663. An operator-supplied
+// --root (explicit) is honored verbatim; keeping it unique per namespace is
+// then the operator's responsibility.
+func resolveBuildRoot(root string, explicit bool, namespace string) string {
+	if explicit {
+		return root
+	}
+	return filepath.Join(root, namespace)
 }
 
 // newSolveOpt builds the BuildKit SolveOpt for a Dockerfile build: the
--- a/cmd/kukebuild/build_test.go
+++ b/cmd/kukebuild/build_test.go
@@ -1,0 +1,77 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #666 Review — `fix: scope kukebuild's default build root per target namespace`**

LGTM — scopes the default BuildKit state root to `<root>/<namespace>` so per-namespace content stores no longer share a cache (the #663 cross-namespace `content digest … not found`), honors an explicit `--root` verbatim, and proves both the fixed and bug-confirmation paths with root+containerd functional repros plus a directly-tested pure helper.

PR #674, opened 2026-05-22
feat(build): kuke build flags phase 2a — secrets + local cache

eminwux:
## Summary

- Adds two build-input flag surfaces to `kuke build` / `kukebuild` on top of the phase-1 engine (#522). Output shape and destination are unchanged — same image into the target realm's containerd namespace.
- **`--secret id=NAME,src=PATH`** (repeatable): mounts a host file into the build at the BuildKit-standard `/run/secrets/NAME`, consumed by `RUN --mount=type=secret,id=NAME`. Wired via a `secretsprovider` session attachable. File-based only; `env=…` sources are rejected with a "deferred" usage error.
- **`--cache-to type=local,dest=PATH` / `--cache-from type=local,src=PATH`** (each repeatable): local-disk BuildKit cache export/import. Forwarded to the client's `CacheExports`/`CacheImports`, and the controller now registers the `local` cache exporter/importer funcs. Only `type=local` is accepted; `registry`/`s3`/`gha` are rejected with a "deferred" usage error.
- `kuke build` shim forwards all three flags verbatim to `kukebuild`; `--help` documents each flag and its spec syntax.

## Implementation notes (for reviewer)

- **Controller-side cache registration was required, not just client-side.** Setting `SolveOpt.CacheExports` alone yields `failed to solve: unknown cache exporter: "local"` — the in-process controller must also register `ResolveCacheExporterFuncs["local"]` / `ResolveCacheImporterFuncs["local"]` (buildkitd does this in `cmd/buildkitd/main.go:851`). I trimmed the map to `local` only; registry/s3/gha remain deferred, matching the issue's out-of-scope list. Caught by the end-to-end smoke below, not by unit tests.
- **Spec parsing** lives in `cmd/kukebuild/main.go` (`parseSecrets`, `parseCacheOpts`, shared `parseCSVPairs`), matching the existing `parseBuildArgs` style. Unknown `--secret` keys, missing `id`/`src`, missing/non-local cache `type`, and missing `dest`/`src` all surface as typed `usageError`s at parse time.
- Non-`type` cache attrs (e.g. `mode=max`) pass through to BuildKit verbatim; `type` is lifted into the `cacheSpec.typ` field and not duplicated in `attrs`.
- Stale doc comments updated: the `main.go` package doc and `newController`'s "phase 2a #523" deferral note now reflect that local cache + secrets have landed.

## Test plan

- [x] `make test` (root module + `cmd/kukebuild` module) — all green.
[...]

PR #675, opened 2026-05-22
feat(build): add kuke build --push + registry auth (phase 2b)

### LINKED ISSUE #524: build: kuke build flag phase 2b — --push + registry auth
## Context

Phase 1 (#522) ships the MVP `kuke build` CLI that writes images into the
target realm's containerd namespace. Phase 2b adds direct registry push:
after a successful local build, `kuke build --push -t REGISTRY/REPO:TAG`
sends the result to the named registry, with credentials resolved from
a docker-compatible config file.

Sibling phase 2a (build-side flags: secrets, cache, multi-platform) is
filed separately and independent — push can land before or after 2a.

## Proposal

Extend `kuke build` with `--push` and accompanying registry-auth
resolution. No new subcommands.

**`--push`**
- After a successful build, push the resulting image (or manifest list,
  if multi-platform via phase 2a's `--platform`) to its tag's registry.
- The image still lands in the target realm's containerd namespace as
  well — `--push` is an *addition*, not a substitute. Operators who
  want push-only can still inspect locally via `kuke image get`.
- Tag must be a fully qualified registry reference
  (`REGISTRY/REPO:TAG`); a bare `name:tag` errors with a clear message.

**Credential resolution** (precedence, documented in `--help` and PR body)
1. `$DOCKER_CONFIG/config.json` if `DOCKER_CONFIG` is set.
2. `~/.docker/config.json` if present (the format BuildKit/docker/nerdctl
   already produce).
3. `KUKEON_REGISTRY_AUTH` env var as a fallback for ephemeral / CI use
   (base64-encoded `user:pass` matching the docker-config `auth` field).
4. If none resolve and the registry rejects anonymous push, fail with a
   clear "no credentials for REGISTRY" error pointing at the precedence
   list.

No `kuke login` / credential-manager subcommand in this issue —
operators bring their own credentials file or env var.

## Acceptance criteria

- [ ] `kuke build --push -t REGISTRY/REPO:TAG ...` pushes the resulting
      image to the registry after the build succeeds. Verified against
      a local registry (e.g. `registry:2`).
- [ ] The image is also written to the target realm's containerd
      namespace (push is additive, not substitutive). Verified via
      `kuke image get --realm <name>`.
- [ ] Auth resolves via `$DOCKER_CONFIG`, then `~/.docker/config.json`,
      then `KUKEON_REGISTRY_AUTH`. Documented in `kuke build --help` and
      the PR body.
- [ ] A bare `name:tag` (no registry host) used with `--push` errors
      with a clear message: "`--push` requires a fully qualified
      registry reference: REGISTRY/REPO:TAG".
- [ ] If `--push` is set, no credentials resolve, and the registry
      rejects anonymous push, the error message points at the
      precedence list.
- [ ] Colocated `_test.go` coverage for the push path and auth resolver
      in `cmd/kukebuild/`.
- [ ] **Out of scope**: `kuke login` / credential-manager subcommand,
      OCI registry mirroring/proxy config, cosign/SBOM signing,
      attestations.

## Notes

- Depends on phase 1 (#522); rebase this PR on `main` after #522 lands.
- Sibling: phase 2a (secrets/cache/platform) — independent of this
  issue, also depends on phase 1. Either may land first.
- BuildKit registry-auth helpers live in
  `github.com/moby/buildkit/session/auth` and resolve docker config out
  of the box — plumbing, not a re-implementation.
- Phase 3 (dev-init swap + CLAUDE.md prerequisites refresh) will be
  filed separately and link back to phase 1, 2a, and 2b.

eminwux:
## Summary

Phase 2b of `kuke build` (#524, sibling of phase 2a #674 which is already merged). Adds direct registry push after a successful local build, with docker-config-compatible credential resolution. No new subcommands — `--push` rides on the existing `kuke build` surface.

- **`--push` (cmd/kukebuild/main.go, build.go):** sets `push=true` on the existing containerd image exporter, so the same exporter both writes the image into the target realm's containerd namespace **and** pushes it to the tag's registry. Push is additive, not substitutive — the local image still lands and stays inspectable via `kuke image get`.
- **Credential resolution (cmd/kukebuild/auth.go):** resolves registry credentials in the documented precedence and wraps them in BuildKit's `authprovider.NewDockerAuthProvider` session attachable:
  1. `$DOCKER_CONFIG/config.json` when `DOCKER_CONFIG` is set;
  2. else `~/.docker/config.json`;
  3. else the `KUKEON_REGISTRY_AUTH` env var (base64 `user:pass`), applied only when layers 1–2 carry no entry for the host.
- **Fully-qualified-reference gate:** a bare `name:tag` (or a docker-hub `user/repo:tag` whose host is implicit) used with `--push` is a parse-time usage error (exit 64): `--push requires a fully qualified registry reference: REGISTRY/REPO:TAG`.
- **Anonymous-rejection error:** when the build's push fails with a registry-auth error (`401`/`403`/`unauthorized`/`denied`/…), the failure is rewritten to point at the credential precedence list; the `no credentials for <host>` variant is used when nothing resolved.
- **Shim (cmd/kuke/build/build.go):** forwards `--push` to `kukebuild` and documents the flag + credential precedence in `kuke build --help`.
- **BuildKit auth API:** uses `authprovider.LoadAuthConfig(*configfile.ConfigFile)` + `NewDockerAuthProvider(DockerAuthProviderConfig{...})` (v0.28.1 surface). `go mod tidy` promotes `github.com/docker/cli` to a direct dep.

### Non-obvious calls (reviewer, please push back)
- **Explicit-registry heuristic** mirrors docker's `splitDockerDomain`: the segment before the first `/` is a registry host only when it is `localhost` or contains `.`/`:`. So `eminwux/kukeon:v1` is rejected for `--push` (its host would silently normalize to docker.io). This matches the issue's "bare `name:tag` errors" intent and avoids a push landing somewhere unintended.
- **Docker-hub key mapping:** the `KUKEON_REGISTRY_AUTH` fallback for an explicit `docker.io/...` push is injected under `authprovider.DockerHubConfigfileKey` (`https://index.docker.io/v1/`), not `docker.io`, so BuildKit serves it back. Non-hub hosts key under themselves.
- **`isRegistryAuthError` is substring-based** (BuildKit/containerd surface push-auth failures as opaque wrapped strings, not typed errors). It only decides whether to rewrite the error into the precedence hint; a false negative just yields the generic `build %q` wrapper.
- **Out of scope (per AC):** `kuke login`/credential-manager, OCI mirroring/proxy, cosign/SBOM/attestations. Also out of scope and **not added**: insecure-registry (plain-HTTP) push config — `--push` assumes a TLS registry, matching docker/buildkit defaults.

## Test plan
- [x] `make test` (root module + `cd cmd/kukebuild && go test ./...`) — green.
- [x] `go build ./...` + `go vet ./...` in the kukebuild module — green.
- [x] `make kukebuild` + `make kuke` link cleanly.
- [x] Unit coverage (cmd/kukebuild/auth_test.go, build_test.go, main_test.go; cmd/kuke/build/build_test.go): resolver precedence (DOCKER_CONFIG / ~/.docker / env fallback / none resolve / config-beats-env), docker-hub key mapping, host extraction, fully-qualified-reference gate, push wiring (`push` attr + auth session appended; anonymous nil-auth path), auth-error detection, both error-message variants, shim flag forwarding.
- [x] Binary-level usage path: `kukebuild --push -t app:dev <ctx>` and `--push -t eminwux/kukeon:v1 <ctx>` both exit 64 with the REGISTRY/REPO:TAG message; `kuke build --help` shows the `--push` flag + credential precedence.
- [x] AC #2 (push is additive) observed in passing: a `--push -t registry:5000/app:dev` build (run as root against host containerd) wrote the image into `default.kukeon.io` even though the remote target was unreachable.
- [ ] AC #1 / #5 live `registry:2` push (success path + anonymous-rejection) — **not run on the dev host**: docker is absent and there is no local registry/insecure-registry config here, so a TLS-terminated `registry:2` can't be stood up safely. Please verify against a local `registry:2` in a clean environment. The push wiring, auth resolution, and error paths are unit-covered above.

Closes #524

### DIFF
--- a/cmd/kuke/build/build.go
+++ b/cmd/kuke/build/build.go
@@ -55,7 +55,7 @@ func NewBuildCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use: "build [-f Dockerfile] [-t name:tag] [--realm <name>] [--build-arg K=V]... " +
 			"[--secret id=NAME,src=PATH]... [--cache-to type=local,dest=PATH]... " +
-			"[--cache-from type=local,src=PATH]... <context>",
+			"[--cache-from type=local,src=PATH]... [--push] <context>",
 		Short: "Build an OCI image from a Dockerfile into a realm's containerd namespace",
 		Long: "Build an OCI image from a Dockerfile using kukeon's native builder.\n\n" +
 			"`kuke build` is a thin shim: it locates the `kukebuild` binary on PATH and\n" +
@@ -68,7 +68,13 @@ func NewBuildCmd() *cobra.Command {
 			"File-based secrets only; the flag is repeatable.\n\n" +
 			"Build cache: --cache-to type=local,dest=PATH exports the build cache to a local\n" +
 			"directory and --cache-from type=local,src=PATH imports it back, reusing layers\n" +
-			"across builds. Only type=local is supported; both flags are repeatable.",
+			"across builds. Only type=local is supported; both flags are repeatable.\n\n" +
+			"Registry push: --push pushes the built image to its tag's registry after a\n" +
+			"successful build. The tag must be a fully qualified registry reference\n" +
+			"(REGISTRY/REPO:TAG); a bare name:tag is rejected. Push is additive — the image\n" +
+			"is still written to the realm's containerd namespace. Credentials resolve in\n" +
+			"order: (1) $DOCKER_CONFIG/config.json when DOCKER_CONFIG is set, (2)\n" +
+			"~/.docker/config.json, (3) the KUKEON_REGISTRY_AUTH env var (base64 user:pass).",
 		Args:          cobra.ExactArgs(1),
 		SilenceUsage:  true,
 		SilenceErrors: false,
@@ -91,6 +97,13 @@ func NewBuildCmd() *cobra.Command {
 	)
 	cmd.Flags().StringArray("cache-to", nil, "Export build cache: type=local,dest=PATH (repeatable)")
 	cmd.Flags().StringArray("cache-from", nil, "Import build cache: type=local,src=PATH (repeatable)")
+	cmd.Flags().Bool(
+		"push",
+		false,
+		"After a successful build, push the image to its tag's registry (requires a fully qualified "+
+			"REGISTRY/REPO:TAG); credentials resolve from $DOCKER_CONFIG/config.json, then ~/.docker/config.json, "+
+			"then $KUKEON_REGISTRY_AUTH",
+	)
 
 	return cmd
 }
@@ -164,6 +177,10 @@ func buildArgv(cmd *cobra.Command, args []string) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
+	push, err := cmd.Flags().GetBool("push")
+	if err != nil {
+		return nil, err
+	}
 
 	argv := []string{kukebuildBinary, "--tag", tag, "--realm", realm}
 	if strings.TrimSpace(file) != "" {
@@ -183,6 +200,9 @@ func buildArgv(cmd *cobra.Command, args []string) ([]string, error) {
 	}
 	for _, c := range cacheFrom {
 		argv = append(argv, "--cache-from", c)
+	}
+	if push {
+		argv = append(argv, "--push")
 	}
 	// Context directory is positional and validated by cobra.ExactArgs(1).
 	argv = append(argv, args[0])
--- a/cmd/kuke/build/build_test.go
+++ b/cmd/kuke/build/build_test.go
@@ -181,6 +181,56 @@ func TestRunBuildForwardsKukeondConfig(t *testing.T) {
 	}
 }
 
+func TestRunBuildForwardsPush(t *testing.T) {
+	var gotArgv []string
+	withStubs(t,
+		func(string) (string, error) { return "/usr/local/bin/kukebuild", nil },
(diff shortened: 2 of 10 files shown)


eminwux:
**PR #675 Review — `feat(build): add kuke build --push + registry auth (phase 2b)`**

Solid, well-tested addition. `--push` rides additively on the existing containerd image exporter (local image still lands), credential precedence (DOCKER_CONFIG → ~/.docker → KUKEON_REGISTRY_AUTH) is correct and unit-covered including the config-beats-env and docker-hub-key cases, the fully-qualified-reference gate is enforced at parse time, and the auth-error rewrite preserves the cause via `%w`. No secrets leak into errors — they name only the host and the precedence list. All commits verified; no dead code (every new symbol has a production caller).

One small item:

- **Error classification** `cmd/kukebuild/auth.go:300`: `isRegistryAuthError` matches `"denied"`/`"forbidden"` as bare substrings, so a non-auth build-step failure containing `permission denied` during a `--push` build gets rewritten with registry-auth framing. The original cause is still wrapped via `%w` so it's recoverable, and you documented the substring tradeoff — non-blocking.

The unchecked test-plan box (live `registry:2` push, AC #1/#5) is a documented host-gated smoke — worth running against a local `registry:2` pre-merge, but not a blocker.

Issue #523, opened 2026-05-15
build: kuke build flags phase 2a — secrets + local cache


eminwux:
## Context

Phase 1 (#522) ships the kukebuild scaffold + BuildKit-as-library + the
MVP `kuke build` CLI. Phase 2a adds two build-side *input* flags on top
of that engine: build-time secrets and local-disk cache export/import.
The output still lands in the target realm's containerd namespace, same
as phase 1 — this issue does not change where images go or their shape,
only how they're built.

Multi-platform output was carved into phase 2c (#646) during a sizing
re-review — it changes the output shape (manifest list) rather than the
build inputs, and reviews better on its own. The `--push` flag is phase
2b (#524). All three depend on phase 1 and are independent of each
other.

## Proposal

Extend `kuke build` (and the corresponding kukebuild handler) with two
flag surfaces. Phase 1's existing flags remain unchanged.

**`--secret id=NAME,src=PATH`** (repeatable)
- Mounts a file from the host into the build at `/run/secrets/NAME`,
  the BuildKit-standard path consumed by Dockerfile
  `RUN --mount=type=secret`.
- File-based secrets only. Env-var-source secrets deferred.

**`--cache-to <spec>` / `--cache-from <spec>`** (each repeatable)
- Scope to `type=local,dest=PATH` and `type=local,src=PATH` —
  local-disk cache export/import. Registry-backed cache (`type=registry`,
  `type=s3`, `type=gha`) is deferred.
- Cache layout follows BuildKit's standard format; nothing
  kukeon-specific.

## Acceptance criteria

- [ ] `kuke build --secret id=NAME,src=PATH ...` mounts the file at
      `/run/secrets/NAME` inside the build. Verified by a Dockerfile
      with `RUN --mount=type=secret,id=NAME cat /run/secrets/NAME`
      asserting the host-side contents.
- [ ] `kuke build --cache-to type=local,dest=PATH ...` writes a BuildKit
      local-cache layout to `PATH`. A subsequent build with
      `--cache-from type=local,src=PATH` reuses those layers (visible
      as `CACHED` in the build log).
- [ ] Colocated `_test.go` coverage for each new flag in
      `cmd/kukebuild/` and `cmd/kuke/build/`.
- [ ] CLI help text (`kuke build --help`) documents each flag and the
      spec syntax for `--secret` and `--cache-*`.
- [ ] **Out of scope** (covered by phase 2b/2c or later): `--push`,
[...]

PR #676, opened 2026-05-22
feat(build): kuke build --platform multi-platform output (phase 2c)

### LINKED ISSUE #646: build: kuke build multi-platform output (phase 2c)
## Context

Phase 1 (#522) ships the `kuke build` engine writing a single image into
the target realm's containerd namespace. Phase 2c adds multi-platform
output: `--platform linux/amd64,linux/arm64` produces a manifest list
covering each target arch.

This was carved out of phase 2a (#523) during a sizing re-review. Unlike
the build-time *input* flags in #523 (`--secret`, `--cache-to/from`),
multi-platform changes the *output shape* — single image becomes a
manifest list — touching the exporter and adding its own verification
path. Reviewing it separately keeps each PR's blast radius narrow.

## Proposal

Add `--platform <list>` to `kuke build` (and the kukebuild handler).
Phase 1's and phase 2a's flags remain unchanged.

**`--platform <list>`**
- Comma-separated `os/arch` targets (e.g. `linux/amd64,linux/arm64`).
- Output is a manifest list written into the target realm's containerd
  namespace.
- Distinct from the Dockerfile-internal `$BUILDPLATFORM` arg phase 1
  already accepts; this selects the build-time *output* platform set.

## Acceptance criteria

- [ ] `kuke build --platform linux/amd64,linux/arm64 -t name:tag ...`
      produces a manifest list in the target realm's containerd
      namespace. `kuke image get --realm <name>` shows the list; each
      per-arch manifest is individually pullable.
- [ ] Colocated `_test.go` coverage for the `--platform` path in
      `cmd/kukebuild/` and `cmd/kuke/build/`.
- [ ] CLI help text (`kuke build --help`) documents `--platform` and
      notes that multi-platform output is a single manifest list (one
      reference), not per-arch tags — operators may expect the latter.
- [ ] **Out of scope**: registry-backed cache, `--push` (phase 2b
      #524 consumes the manifest list when present), attestations/SBOM,
      provenance flags.

## Dependencies

Depends on #522 (phase 1 — `kuke build` + `kukebuild` engine).

## Notes

- Carved out of #523 (phase 2a) during a sizing re-review — secrets +
  local cache stay in #523; multi-platform output lands here.
- Sibling: #524 (phase 2b — `--push` + registry auth) is independent;
  when both land, `--push` pushes the manifest list as one reference.
- BuildKit's multi-platform output is a single manifest list as one
  reference; nothing kukeon-specific in the layout.

eminwux:
## Summary

- Phase 2c of `kuke build` (#646, carved from #523). Adds `--platform <list>` so a build produces a **single manifest list** covering each target arch, written into the target realm's containerd namespace.
- **Engine (`cmd/kukebuild`):** `--platform` parses a comma-separated `os/arch[/variant]` list, normalizes + validates each entry via `containerd/platforms` at parse time (typos / dup targets fail fast as usage errors), and sets the BuildKit Dockerfile frontend's `platform` attr (`frontend/dockerui` `keyTargetPlatform`). The frontend turns a multi-target attr into a per-arch ref set that the existing image exporter writes as one manifest list — no exporter-side wiring beyond the attr.
- **Shim (`cmd/kuke/build`):** `--platform` flag forwarded verbatim to `kukebuild`; absent flag forwards nothing (phase-1 single-image default preserved).
- **Help text:** documents `--platform` and notes the output is a single manifest list (one reference), **not** per-arch tags, and that it is distinct from the Dockerfile-internal `$BUILDPLATFORM`.

`containerd/platforms` moved from indirect to direct in `cmd/kukebuild/go.mod` (`go mod tidy`, no version change).

## Implementation notes (for reviewer pushback)

- **Validation strictness:** I reject duplicate targets (including alias dups like `linux/arm64,linux/arm64/v8`, both normalize to `linux/arm64`) as a usage error rather than silently de-duping, so an operator does not produce a manifest list with redundant entries. Empty/whitespace segments are tolerated (trailing comma is fine).
- **Sibling #524 (`--push`) already landed** (merged as #675). `--push` + `--platform` compose: the same image exporter both writes the manifest list to the namespace and pushes it as one reference (the push attr rides on the same `Exports[0]`). No extra wiring was needed; not separately smoke-tested here (no registry on the dev host).

## Test plan

- [x] `make test` — full CI target (root module + kukebuild module), all pass.
- [x] `go build ./cmd/...` + `go vet` (root) and `go build ./... && go vet ./...` (kukebuild module) — clean.
- [x] New colocated tests: `cmd/kukebuild/main_test.go` (`--platform` parse, normalization, usage errors incl. dup-via-alias), `cmd/kukebuild/build_test.go` (frontend `platform` attr wiring: multi / single / absent), `cmd/kuke/build/build_test.go` (shim forwarding present/absent).
- [x] **AC#1 — real multi-platform build verified end-to-end** via the actual `kukebuild` engine against a `FROM scratch` + `COPY` context: `./kukebuild --platform linux/amd64,linux/arm64 -t mp-test:dev --realm default <ctx>` produced `exporting manifest list sha256:9082…`. `ctr -n default.kukeon.io images ls` shows the image as `application/vnd.docker.distribution.manifest.list.v2+json` covering `linux/amd64,linux/arm64`; the list's two child manifests (`84af…` amd64, `6d9d…` arm64) are both present in the namespace's content store (individually pullable). Test image removed afterward.
- [x] CLI help text (`kuke build --help`) renders the `--platform` flag + the manifest-list-not-per-arch-tags note.
- [ ] `kuke image get --realm default` display path — **not run**: `kuke image get` validates the realm against kukeon state, which requires `kuke init` (and thus a kukeond image build via docker, not running on this host). This is a `kuke init` precondition orthogonal to `--platform`; `kuke image get` reads the exact `<realm>.kukeon.io` namespace that `ctr` confirms holds the manifest list, so the display is over verified content. A reviewer on an initialized host can confirm the rendered output.
- [ ] `make dev-init` smoke — **not run**: this change touches only the `kuke build` / `kukebuild` image-builder surface, not the daemon bootstrap, `/opt/kukeon`, or anything `make dev-init` exercises (it rebuilds + re-inits kukeond and checks realm parity). No daemon-read path is affected.

### Environmental note (not a code issue)
The repo's `golangci-lint` pre-commit hook panics in this sandbox with `package requires newer Go version go1.25 (application built with go1.24)` — the cached golangci-lint binary is built with go1.24 but the `cmd/kukebuild` module is `go 1.25.5`, and the `go.work` union pulls it in. Verified pre-existing and independent of this diff (same panic on an unmodified `internal/consts/consts.go`). `go vet`/`go build`/`make test` all pass.

Closes #646

### DIFF
--- a/cmd/kuke/build/build.go
+++ b/cmd/kuke/build/build.go
@@ -55,7 +55,7 @@ func NewBuildCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use: "build [-f Dockerfile] [-t name:tag] [--realm <name>] [--build-arg K=V]... " +
 			"[--secret id=NAME,src=PATH]... [--cache-to type=local,dest=PATH]... " +
-			"[--cache-from type=local,src=PATH]... [--push] <context>",
+			"[--cache-from type=local,src=PATH]... [--platform os/arch,...] [--push] <context>",
 		Short: "Build an OCI image from a Dockerfile into a realm's containerd namespace",
 		Long: "Build an OCI image from a Dockerfile using kukeon's native builder.\n\n" +
 			"`kuke build` is a thin shim: it locates the `kukebuild` binary on PATH and\n" +
@@ -69,6 +69,12 @@ func NewBuildCmd() *cobra.Command {
 			"Build cache: --cache-to type=local,dest=PATH exports the build cache to a local\n" +
 			"directory and --cache-from type=local,src=PATH imports it back, reusing layers\n" +
 			"across builds. Only type=local is supported; both flags are repeatable.\n\n" +
+			"Multi-platform: --platform linux/amd64,linux/arm64 builds for each target arch\n" +
+			"and writes the result as a single manifest list — one image reference covering\n" +
+			"every arch, not per-arch tags (operators expecting name:tag-amd64 / -arm64 will\n" +
+			"not find them). `kuke image get --realm <name>` shows the list; each per-arch\n" +
+			"manifest is individually pullable. Distinct from a Dockerfile's $BUILDPLATFORM\n" +
+			"arg, which selects the build host's platform; --platform selects the output set.\n\n" +
 			"Registry push: --push pushes the built image to its tag's registry after a\n" +
 			"successful build. The tag must be a fully qualified registry reference\n" +
 			"(REGISTRY/REPO:TAG); a bare name:tag is rejected. Push is additive — the image\n" +
@@ -97,6 +103,12 @@ func NewBuildCmd() *cobra.Command {
 	)
 	cmd.Flags().StringArray("cache-to", nil, "Export build cache: type=local,dest=PATH (repeatable)")
 	cmd.Flags().StringArray("cache-from", nil, "Import build cache: type=local,src=PATH (repeatable)")
+	cmd.Flags().String(
+		"platform",
+		"",
+		"Comma-separated os/arch[/variant] targets (e.g. linux/amd64,linux/arm64); multiple targets "+
+			"produce a single manifest list, not per-arch tags",
+	)
 	cmd.Flags().Bool(
 		"push",
 		false,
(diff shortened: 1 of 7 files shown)


eminwux:
**PR #676 Review — `feat(build): kuke build --platform multi-platform output (phase 2c)`**

LGTM — `--platform` parses/normalizes/dedups targets at parse time and wires the comma-joined set to BuildKit's dockerui `platform` attr, following the exact shim→engine forwarding precedent of `--push`/`--cache-to`; new fields are all used, commits are signed, and the two unchecked test-plan items are host-gated (`kuke init`) and documented.

Issue #671, opened 2026-05-22
runtime: containerd container records vanish after a clean cell start


eminwux:
## Context

After `kuke start cell sbsh-pm-0`, the daemon logged a clean create+start for both containers, with tasks running:

```
created container id=default_default_sbsh-pm-0_root ... image=registry.eminwux.com/busybox:latest
started root container id=default_default_sbsh-pm-0_root ... pid=857162
created container id=default_default_sbsh-pm-0_work ... image=registry.eminwux.com/claude:latest
started container id=default_default_sbsh-pm-0_work
container state determined via TaskStatus container=root taskStatus=running internalState=1
container state determined via TaskStatus container=work taskStatus=running internalState=1
```

Yet a subsequent `ctr -n default.kukeon.io c ls` (and `t ls`) is **empty** — the containerd container records are gone, while the cell metadata under `/opt/kukeon/data/default/default/default/sbsh-pm-0/metadata.json` persists. The cell therefore keeps appearing in `kuke get co` with phantom rows.

This is the upstream cause of the phantom-container symptom; #670 is the reporting bug that masks it (an absent containerd record renders as `Stopped` instead of an honest "gone" state).

Likely mechanism (unconfirmed): the payload exits immediately (e.g. `busybox:latest` with no long-running command) and the containerd container is reaped, leaving the spec on disk with no backing container. The exact teardown/reaper path has not been pinned.

## Proposal

Investigate and resolve the divergence between persisted cell metadata and containerd state:
1. Determine what deletes the containerd container record after task exit (reaper? start path? cleanup-on-exit?).
2. Decide the intended contract: should a stopped/exited container retain its containerd record (so STATE is faithful), or is removal intended and the metadata/listing must reflect it?
3. Implement whichever side of the contract is wrong.

## Acceptance criteria
- [ ] Root cause of the disappearing containerd records is identified and documented in the issue/PR.
- [ ] After a cell's payload exits, `kuke get co` and `ctr c ls` agree on whether the container exists.
- [ ] Regression test (or smoke-test step) covering start → payload-exit → state-query for a short-lived payload like `busybox`.

## Notes
- Depends on / paired with #670 (reporting fix). #670 stops the phantom `Stopped`; this issue addresses why the records vanish.
- Daemon and `--no-daemon` hit the same containerd via the bind-mounted socket, so this is not a daemon-routing/parity issue.
[...]

PR #678, opened 2026-05-22
fix: report absent containers as NotCreated instead of Stopped

### LINKED ISSUE #670: cli: kuke get container reports Stopped for containers absent from containerd
## Context

`kuke get co` lists a cell's containers from on-disk metadata (the desired-state specs under `/opt/kukeon/data/.../metadata.json`) and computes the STATE column live per container. Observed: the table shows `root`/`work` as `Stopped` even though `ctr -n default.kukeon.io c ls` shows **no such containers exist** in containerd, and the daemon itself logs `"container does not exist in containerd"`.

Root cause: `internal/controller/runner/container_state.go:154-160` treats "absent from containerd" as `Stopped`:

```go
if !containerExists {
    r.logger.InfoContext(... "container does not exist in containerd" ...)
    return intmodel.ContainerStateStopped, nil
}
```

The internal state enum (`internal/modelhub/container.go:224-230`) is `Pending / Ready / Stopped / Failed / Unknown` — there is no value distinguishing "no containerd record at all" from "exists but stopped". Container state is deliberately not persisted (`internal/controller/get_container.go:116-120`), so the column is always computed at query time and every absent container collapses to `Stopped`.

This runs identically on the daemon and `--no-daemon` paths since both share `GetContainerState`.

Operator impact: a cell whose containers were reaped/lost is indistinguishable from a normally-stopped cell. This masks the real disappearance (see linked follow-up).

## Proposal

Make `GetContainerState` distinguish absent from stopped. Options (dev's call):
- introduce a distinct state value (e.g. `ContainerStateNotCreated`) and map it to a distinct STATE string, or
- reuse `Unknown`/`Pending` for the absent case with a clear display string, and/or
- reconcile the listing so absent specs render an honest "not realized" state.

Pick one; the load-bearing requirement is that an absent containerd record must not display as `Stopped`.

## Acceptance criteria
- [ ] When a container has no containerd record, `kuke get co` shows a STATE that is visibly distinct from a genuinely stopped-but-present container.
- [ ] Behaviour is identical on the daemon and `--no-daemon` paths (both go through `GetContainerState`).
- [ ] Unit test in `internal/controller/runner/container_state_test.go` covering the absent-container branch asserts the new state (not `Stopped`).
- [ ] Display-string mapping updated in `cmd/kuke/get/container/container.go` (`containerStateToString`) for any new enum value.

## Notes
- Emitting site: `internal/controller/runner/container_state.go:154-160`.
- State enum: `internal/modelhub/container.go:224-230`; display mapping in `cmd/kuke/get/container/container.go`.
- Investigation follow-up: #671 (why the containerd records vanish in the first place). This issue is the reporting fix that stops masking it.

eminwux:
## Summary
- `kuke get co` collapsed two distinct conditions into `Stopped`: a container whose containerd record exists but whose task is gone (a clean stop / post-reboot, #543) **and** a container with no containerd record at all (never realized, or reaped/lost — #671). An operator could not tell a cell whose containers vanished from a normally-stopped cell.
- The emitting site `internal/controller/runner/container_state.go:154` (the `!containerExists` branch) now returns a new `ContainerStateNotCreated` instead of `Stopped`. The task-gone branch below it still returns `Stopped`, so the two are now visibly distinct in the STATE column ("NotCreated" vs "Stopped").
- Behaviour is identical on the daemon and `--no-daemon` paths since both go through `GetContainerState`.

## Design notes (for reviewer)
- **Enum ordinal lockstep.** `intmodel.ContainerState` and `v1beta1.ContainerState` are converted by a direct `int` cast (`internal/apischeme/scheme.go:434,497`), so the new value is appended **last** in both enums (after `Unknown`) to avoid renumbering existing ordinals. `assignContainerStateInt`'s range check upper bound was bumped to `ContainerStateNotCreated`, and `String()`/`parseContainerState`/`StateNotCreatedStr` were added so the daemon's JSON round-trip carries the distinct label rather than collapsing to `Unknown`.
- **Reconciler behavior preserved.** `refresh.go` deliberately relies on absent→`Stopped` for lifecycle decisions (auto-delete latch #543, wind-down #301). All three switches that classify `GetContainerState`'s result (`rootContainerStillRunning`, `deriveCellStateFromNonRootContainerStatuses`, `deriveCellStateFromRootContainer`) now treat `NotCreated` exactly like `Stopped`, so cell-state derivation and auto-delete/wind-down behavior are unchanged — only the per-container reported STATE differs. Two stale comments referencing the old absent→Stopped mapping were updated.

## Test plan
- [x] `make test` (root module unit suite + kukebuild module) — pass
- [x] `GOWORK=off go build ./...` / `go vet ./...` on affected packages — pass
- [x] `internal/controller/runner/container_state_test.go`: the absent-branch test now asserts `NotCreated` (was `Stopped`); renamed `TestGetContainerState_ContainerAbsentReturnsNotCreated`
- [x] `internal/controller/runner/refresh_test.go`: added `absent_workload_counts_as_terminal_like_stopped` locking in that a `NotCreated` workload still winds the cell to `Stopped`
- [ ] `make dev-init` daemon-parity smoke — not run: this change is in query-time container-state derivation, not the daemon's `/opt/kukeon` read/bootstrap path the parity check guards; cell-state derivation outcomes are unchanged (NotCreated treated as Stopped).

Closes #670

### DIFF
--- a/cmd/kuke/get/container/container.go
+++ b/cmd/kuke/get/container/container.go
@@ -414,6 +414,8 @@ func containerStateToString(state v1beta1.ContainerState) string {
 		return "Pausing"
 	case v1beta1.ContainerStateFailed:
 		return "Failed"
+	case v1beta1.ContainerStateNotCreated:
+		return "NotCreated"
 	default:
 		return "Unknown"
 	}
--- a/internal/controller/runner/container_state.go
+++ b/internal/controller/runner/container_state.go
@@ -152,12 +152,17 @@ func (r *Exec) GetContainerState(cell intmodel.Cell, containerID string) (intmod
 	}
 
 	if !containerExists {
-		// Container doesn't exist in containerd
+		// No containerd record at all — distinct from Stopped (a record that
+		// exists but whose task is gone, handled below). Reporting must not
+		// collapse the two: an absent record means the container was never
+		// realized or was reaped/lost, which an operator needs to tell apart
+		// from a normal clean stop (#670). The reconciler treats NotCreated
+		// exactly like Stopped for its lifecycle decisions — see refresh.go.
 		r.logger.InfoContext(r.ctx, "container does not exist in containerd",
 			"container", containerID,
 			"containerdID", containerdID,
 			"namespace", namespace)
-		return intmodel.ContainerStateStopped, nil
+		return intmodel.ContainerStateNotCreated, nil
 	}
 
 	// Get container state using TaskStatus from ctr package
--- a/internal/controller/runner/container_state_test.go
+++ b/internal/controller/runner/container_state_test.go
@@ -92,13 +92,15 @@ func TestGetContainerState_TaskNotFound_ReturnsStopped(t *testing.T) {
 	}
 }
 
-// TestGetContainerState_ContainerGoneAlsoStopped sanity-checks the existing
-// pre-fix path: a container that's been removed from containerd entirely
-// (operator ran `ctr -n <ns> containers rm` between boots) reads back as
-// Stopped via the "container does not exist" branch. The Layer 2 fix added
-// in #543 must not regress this case — both paths land at Stopped so the
-// cell-level derivation reaches a terminal state either way.
-func TestGetContainerState_ContainerGoneAlsoStopped(t *testing.T) {
+// TestGetContainerState_ContainerAbsentReturnsNotCreated locks down the #670
+// reporting fix: a container with no containerd record at all (never realized,
+// or reaped/lost — operator ran `ctr -n <ns> containers rm`, or the records
+// vanished per #671) must read back as NotCreated, NOT Stopped. Collapsing the
+// two makes a cell whose containers were lost indistinguishable from a normally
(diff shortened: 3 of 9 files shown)


eminwux:
**PR #678 Review — `fix: report absent containers as NotCreated instead of Stopped`**

LGTM — splits absent-record from task-gone into a distinct `NotCreated` state at `container_state.go:165`, with reconciler behavior preserved (all three derivation switches in `refresh.go` treat `NotCreated` as terminal alongside `Stopped`). Verified the ordinal lockstep that the blind int-cast in `scheme.go:434,497` depends on: both enums append `NotCreated` last in identical order, and every consumer (`String()`, `parseContainerState`, `assignContainerStateInt` bound, `containerStateToString`) is updated — no missed switch.

The `make dev-init` daemon-parity smoke is the only unchecked box; the author's reasoning is sound (this is query-time derivation, not the daemon's `/opt/kukeon` read/bootstrap path, and cell-state outcomes are unchanged) — non-blocking, but worth a quick run pre-merge if a host is handy.

PR #627, opened 2026-05-20
feat(container): add repos[] with kuketty pre-Serve clone (phase 1a)

eminwux:
## Summary

Phase 1a of #423, rebased onto **phase 0 (#645)**. Adds a first-class `containers[].repos[]` declaration so a container can depend on git repositories without the hand-rolled `if [ ! -d $DIR/.git ]; then git clone …; fi` blocks crew templates duplicate in `onInit` scripts.

**This PR was rescoped after #645 merged.** PM converted the prior revision (sidecar + status-file design) to draft and re-sequenced the work; per the [draft-conversion note](https://github.com/eminwux/kukeon/issues/617) this revision **drops** the `internal/reposwire` sidecar and the `repos-status.json` write-back, and rebases the carry-forward pieces onto phase 0's `ContainerDoc` mount.

### What carried forward from the prior revision
- `ContainerSpec.Repos` / `ContainerStatus.Repos` schema (+ `v1beta1` & `internal/modelhub` mirrors), all four apischeme spec conversions and both status conversions, apply-parser validation (`name`/`target`-absolute/`url` required), and `errdefs` sentinels.
- The clone-if-absent engine in `cmd/kuketty/repos.go`.

### What changed for phase 1a
- **No sidecar doc.** kuketty reads `repos[]` straight from `doc.Spec.Repos` — the kukeon-native `ContainerDoc` it already mounts since #645 (`cmd/kuketty/main.go` calls `processRepos(ctx, doc.Spec.Repos, logger)` after `buildTerminalSpec`, before `sbshserver.Serve`). `internal/reposwire`, the conditional `repos.json` RO bind, and the `daemon-side` renderer changes are gone.
- **No status file.** The `/run/kukeon/tty/repos-status.json` write + daemon read-back are dropped; per-repo status reporting into `ContainerStatus.Repos` rides the `GetSetupStatus` RPC in **phase 1b (#642)**. `ContainerStatus.Repos` lands as schema only here (populated in 1b).
- The field rides phase 0's existing serialization: `writeKukettyDoc` → `BuildContainerSpecExternalFromInternal` (now converts `Repos`) → mounted `metadata.json` → kuketty. **No new daemon code** beyond the conversion helper.

### Design notes (please push back)
[...]

PR #680, opened 2026-05-22
fix: diff Spec.Repos and Spec.Secrets in diffContainerSpec

### LINKED ISSUE #647: apply/diff: diffContainerSpec ignores Spec.Repos (and Spec.Secrets)
Surfaced during review of #627 (phase 1a of #423).

`diffContainerSpec` at `internal/controller/apply/diff.go:534` enumerates container spec fields field-by-field (Image, Command, Args, Env, Ports, Volumes, Privileged, User, ReadOnlyRootFilesystem, Capabilities, SecurityOpts, Tmpfs, Resources) to classify drift as breaking / compatible / none. It does **not** compare `.Spec.Repos` (added in #627) — nor `.Spec.Secrets` (pre-existing).

Effect: a `kuke apply` that only edits `repos[]` (or `secrets[]`) registers no change, so no update/recreate is triggered; the new declaration is realized only on the next container start.

#627 follows the existing `Secrets` precedent, so this was non-blocking for that PR. Filing at the precedent's scope: either both fields should be diffed, or the omission should be documented as intentional (e.g. secrets/repos are re-resolved every reconcile regardless).

Suggested fix: add `Repos` and `Secrets` comparisons alongside the `Tmpfs`/`Volumes` branches (with `reposEqual` / `secretsEqual` helpers mirroring `tmpfsEqual`), or a comment in `diffContainerSpec` explaining why they are deliberately excluded.

eminwux:
## Summary
- `diffContainerSpec` enumerated container spec fields one-by-one but never compared `.Spec.Repos` (added in #627) or `.Spec.Secrets` (pre-existing), so a `kuke apply` editing only `repos[]` or `secrets[]` registered no drift — the new declaration was realized lazily on the next container start instead of triggering an update/recreate.
- Took the precedent's "diff both fields" option from the issue (not the "document as intentional" alternative): added `secrets` and `repos` compatible-change branches alongside `tmpfs`/`volumes`, plus `reposEqual` / `secretsEqual` helpers mirroring `tmpfsEqual`/`volumeMountsEqual`.
- `ContainerRepo` is a flat comparable struct (direct `!=`); `ContainerSecret` carries a `*ContainerSecretRef` pointer, so `secretsEqual` compares scalar fields and dereferences `SecretRef` via a nil-safe `secretRefEqual`.

## Test plan
- [x] `make test` (full CI target: `go test $(go list ./... | grep -v /e2e)` + `cd cmd/kukebuild && go test ./...`) — passes
- [x] `GOWORK=off go build ./...` / `go vet ./internal/controller/apply/` — clean
- [x] New tests: `TestDiffContainer_ReposChange`, `TestDiffContainer_SecretsChange` (incl. `SecretRef` pointer-value comparison), `TestDiffContainer_ReposSecretsNoChange`

Closes #647

### DIFF
--- a/internal/controller/apply/diff.go
+++ b/internal/controller/apply/diff.go
@@ -649,6 +649,24 @@ func diffContainerSpec(desired, actual *intmodel.ContainerSpec, rootContainer bo
 		result.Details["resources"] = "resource limits changed"
 	}
 
+	if !secretsEqual(desired.Secrets, actual.Secrets) {
+		result.HasChanges = true
+		if result.ChangeType == ChangeTypeNone {
+			result.ChangeType = ChangeTypeCompatible
+		}
+		result.ChangedFields = append(result.ChangedFields, "secrets")
+		result.Details["secrets"] = "secrets changed"
+	}
+
+	if !reposEqual(desired.Repos, actual.Repos) {
+		result.HasChanges = true
+		if result.ChangeType == ChangeTypeNone {
+			result.ChangeType = ChangeTypeCompatible
+		}
+		result.ChangedFields = append(result.ChangedFields, "repos")
+		result.Details["repos"] = "repos changed"
+	}
+
 	return result
 }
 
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #680 Review — `fix: diff Spec.Repos and Spec.Secrets in diffContainerSpec`**

LGTM — adds `repos`/`secrets` compatible-change branches to `diffContainerSpec` mirroring the `tmpfs`/`volumes` precedent; verified the equality helpers cover every field of `ContainerSecret`/`ContainerRepo`/`ContainerSecretRef`, so the previously-missed drift is now detected with no field gaps.

Issue #684, opened 2026-05-22
runner: remove lazy ctrClient init race in ensureClientConnected


eminwux:
## Context
`ensureClientConnected` (`internal/controller/runner/runner.go:276-281`) does an unguarded nil-check-then-assign of `r.ctrClient`; `NewRunner` leaves the field nil. Two concurrent first-use goroutines (e.g. the first RPC handler and the first reconcile tick) can each observe nil, each construct a client, and the second assignment overwrites the first — a data race on the pointer plus a possible duplicate containerd connection.

## Proposal
Initialize `ctrClient` eagerly in `NewRunner`, or guard the lazy init with `sync.Once` / a mutex.

## Acceptance criteria
- [ ] Concurrent first-use cannot construct two clients.
- [ ] A `-race` test on concurrent first-connect passes.

## Notes
Same #303 (5e76b30) concurrency class as the unguarded cgroups-map cache (filed separately); narrow startup window. Dev may reasonably bundle the two fixes in one PR.

eminwux:
Related: #681 — same #303 shared-client concurrency class (unguarded `cgroups` map). The two fixes can land in one PR if dev prefers.

PR #686, opened 2026-05-22
fix: guard cgroups manager cache with a mutex

### LINKED ISSUE #681: ctr: guard cgroups manager cache with a mutex
## Context
The shared `ctr.client` caches cgroup2 managers in an unsynchronized map (`internal/ctr/client.go:38`). The daemon serves each RPC connection in its own goroutine (`internal/daemon/server.go:236`) plus a background reconcile loop (`internal/daemon/server.go:167`), all sharing one client. The sibling `containers` / `tasks` caches are guarded by `containersMu` / `tasksMu`; the `cgroups` map has no mutex.

## Proposal
Add a mutex for the `cgroups` map and guard all three accessors (`storeManager` / `dropManager` / `managerFor`, `internal/ctr/cgroups.go:549-577`) consistently with the existing container/task caches.

## Acceptance criteria
- [ ] All `client.cgroups` reads/writes are synchronized on every path.
- [ ] A `-race` test exercises concurrent store/drop/load of the cgroups cache and passes.
- [ ] No remaining unguarded access to `client.cgroups`.

## Notes
Same bug class as #303 (5e76b30, "remove shared mutable namespace from ctr client"), on the cache field that refactor missed. Concurrent CreateCell/StartCell (cgroup provision) vs DeleteCell/purge (`DeleteCgroup`) vs reconcile (`LoadCgroup`) produce `fatal error: concurrent map writes`, panicking the daemon. Same #303 concurrency class as the lazy-`ctrClient`-init race (filed separately); dev may bundle the two fixes.

eminwux:
## Summary
- The shared `ctr.client` cached cgroup2 managers in an unsynchronized map (`internal/ctr/client.go:38`). The daemon serves each RPC connection in its own goroutine plus a background reconcile loop, all sharing one client. The sibling `containers`/`tasks` caches are guarded by `containersMu`/`tasksMu`; the `cgroups` map had no mutex, so concurrent CreateCell/StartCell vs DeleteCell/purge vs reconcile could produce `fatal error: concurrent map writes` and panic the daemon.
- Add `cgroupsMu sync.RWMutex` and guard all three accessors (`storeManager`/`dropManager`/`managerFor`) consistently with the existing container/task caches: write-lock the store/drop mutations; `managerFor` takes the read lock for the cache-hit check and releases it before the (filesystem-touching) `cgroup2.LoadManager` miss path, mirroring `loadContainer` in `cache.go`.

## Test plan
- [x] `make test` — full CI target, passes (ctr package included).
- [x] `go build ./...`, `go vet ./internal/ctr/...` — clean.
- [x] New `TestCgroupsCacheConcurrency` (plus `TestStoreManagerInitializesMap`/`TestDropManager`/`TestDropManagerNilMap`) drives concurrent store/drop/load of the cgroups cache; passes under `make test`. It mirrors the sibling `TestContainerCacheConcurrency`/`TestTaskCacheConcurrency` and is structured for the race detector (load path pinned to a pre-populated cache-hit group so it never falls through to `cgroup2.LoadManager`).
- [ ] `go test -race` — could not run locally: `-race` requires cgo and no C compiler (`gcc`/`clang`/`cc`) is present on the dev host. The repo's CI target (`make test`, `.github/workflows/test.yaml`) does not pass `-race`, so this is not a CI-target gap; the test still exercises the concurrent paths under plain `go test`.

## Notes
- Same #303 concurrency class (5e76b30) as the lazy-`ctrClient`-init race in #684; this PR fixes only the cgroups cache field that refactor missed. #684 is left as its own change (the issue noted bundling was optional).

Closes #681

### DIFF
--- a/internal/ctr/cgroups.go
+++ b/internal/ctr/cgroups.go
@@ -547,6 +547,8 @@ func (c *client) effectiveMountpoint(mountpoint string) string {
 }
 
 func (c *client) storeManager(group string, manager *cgroup2.Manager) {
+	c.cgroupsMu.Lock()
+	defer c.cgroupsMu.Unlock()
 	if c.cgroups == nil {
 		c.cgroups = make(map[string]*cgroup2.Manager)
 	}
@@ -554,6 +556,8 @@ func (c *client) storeManager(group string, manager *cgroup2.Manager) {
 }
 
 func (c *client) dropManager(group string) {
+	c.cgroupsMu.Lock()
+	defer c.cgroupsMu.Unlock()
 	if c.cgroups == nil {
 		return
 	}
@@ -564,7 +568,10 @@ func (c *client) managerFor(group, mountpoint string) (*cgroup2.Manager, error) 
 	if err := validateGroupPath(group); err != nil {
 		return nil, err
 	}
-	if manager, ok := c.cgroups[group]; ok {
+	c.cgroupsMu.RLock()
+	manager, ok := c.cgroups[group]
+	c.cgroupsMu.RUnlock()
+	if ok {
 		return manager, nil
 	}
 	mp := c.effectiveMountpoint(mountpoint)
--- a/internal/ctr/cgroups_internal_test.go
+++ b/internal/ctr/cgroups_internal_test.go
@@ -1,0 +1,104 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #686 Review — `fix: guard cgroups manager cache with a mutex`**

LGTM — adds `cgroupsMu` and guards all three cgroups-cache accessors exactly like the sibling `containers`/`tasks` caches (`managerFor` reads under RLock then stores misses via the locked `storeManager`), closing the `concurrent map writes` race with no behavior change. `-race` is host-gated (no cgo compiler) and not a CI-target gap.

PR #688, opened 2026-05-22
fix: start tasks in RecreateCell before stamping Ready

### LINKED ISSUE #682: runner: RecreateCell stamps Ready without starting tasks
## Context
`RecreateCell` (`internal/controller/runner/recreate_cell.go`) deletes and re-creates a cell's containers via `createCellContainers`, then calls `markCellReady` and persists `Ready` (`recreate_cell.go:173-183`). But `createCellContainers` only creates container *records* — it does not start tasks or run CNI ADD (`internal/controller/runner/provision.go:1346`; the comment at `provision.go:1446` confirms "CNI ADD runs during StartCell"). The only caller, `internal/controller/apply/reconcile.go:373`, returns immediately without a `StartCell` (contrast the re-materialize branch at `reconcile.go:357`, which does call `StartCell`).

## Proposal
After recreating the containers, the cell must reach the same running state a fresh create -> start produces *before* `Ready` is stamped. Implementation (call `StartCell` vs. restructure) is dev's choice. Also correct the inline comment at `recreate_cell.go:114-118`, which wrongly claims `createCellContainers` "re-runs CNI ADD".

## Acceptance criteria
- [ ] After `kuke apply -f` of a root image/command/args change, the cell's task is actually running with a network attachment.
- [ ] `Ready` is never persisted on created-but-not-started containers.
- [ ] A test covers apply-of-root-spec-change ending in a running cell.

## Notes
Produces the same "phantom attach" divergence #657 (1503273) addressed — metadata says `Ready` but no task is running. Complementary to the run/attach task-liveness guard (filed separately): that guard is the safety net; this is one producer of the diverged state.

eminwux:
## Summary

- `RecreateCell` (the apply root-container-change handler reached via `apply/reconcile.go`'s `diff.RootContainerChanged` branch) called `createCellContainers` then stamped `Ready` and persisted — but `createCellContainers` only creates container *records*; it does not start tasks or run CNI ADD ("CNI ADD runs during StartCell", `provision.go:1446`). The cell was reported `Ready` with no task running and no network attachment — the same "phantom attach" divergence #657 addressed.
- Fix routes the recreated cell through `StartCell` (mirroring the create branch at `reconcile.go:331` and the re-materialize branch at `reconcile.go:357`, both of which call `StartCell`). The recreated container records are persisted first so `StartCell`'s `GetCell` reads the new root spec + freshly assigned containerd IDs; `StartCell` then starts the tasks, runs CNI ADD against the deterministic root ID, and stamps + persists `Ready` only once the cell is actually up.
- Removed the premature `markCellReady` from `RecreateCell` so `Ready` is never persisted over created-but-not-started containers (AC #2).
- Corrected the inline comment at `recreate_cell.go` that wrongly attributed the CNI re-ADD to `createCellContainers`; the re-ADD happens in the trailing `StartCell` call.

Implementation note (dev's choice per the issue): the fix lives inside `RecreateCell` rather than in the `reconcile.go` caller, keeping `RecreateCell` a complete teardown→recreate→start operation symmetric with `StartCell` itself. The `createCellContainers` → persist → `StartCell` shape matches the existing create-branch pattern (`CreateCell` creates records, `StartCell` recreates and starts them), so the brief double-create of the root is consistent with current behavior, not new churn.

## Test plan
- [x] `make test` (full CI target: `go test $(go list ./... | grep -v /e2e)` + `cd cmd/kukebuild && go test ./...`) — passed
- [x] `go vet ./internal/controller/...` — passed
- [x] New `internal/controller/runner/recreate_cell_test.go`: `TestRecreateCell_StartsTasksBeforeReady` (AC #1/#3 — apply-of-root-spec-change drives the root task start and ends `Ready`) and `TestRecreateCell_DoesNotStampReadyWhenStartFails` (AC #2 — a failed task start surfaces an error and the cell is not left `Ready`). Both fail against the pre-fix code (`startContainerCalls == 0`; `Ready` stamped with nil error).
- [ ] `make dev-init` daemon-parity smoke — not run: no containerd daemon is running in this environment (`/run/containerd/containerd.sock` absent) and the host lacks the image-build path. This smoke guards `kuke init` realm/namespace/cgroup bootstrap + bind-mount parity, which this change does not touch — it alters only the apply recreate-cell lifecycle path, with no change to the build path, `/opt/kukeon` layout, or daemon bootstrap.

Closes #682

### DIFF
--- a/internal/controller/runner/recreate_cell.go
+++ b/internal/controller/runner/recreate_cell.go
@@ -111,11 +111,12 @@ func (r *Exec) RecreateCell(desired intmodel.Cell) (intmodel.Cell, error) {
 	}
 
 	// Release the root container's CNI/IPAM reservation before deleting it.
-	// createCellContainers below rebuilds the root under this same deterministic
-	// containerd ID and re-runs CNI ADD; without releasing first, host-local
-	// IPAM rejects the re-ADD as a duplicate allocation (issue #630). releaseCNI
-	// runs before the stop/delete so the netns is still valid for CNI DEL where
-	// applicable; purgeCNI scrubs the residual allocation file afterward.
+	// createCellContainers below rebuilds the root container record under this
+	// same deterministic containerd ID, and the StartCell call at the end of
+	// RecreateCell re-runs CNI ADD against it; without releasing first,
+	// host-local IPAM rejects the re-ADD as a duplicate allocation (issue #630).
+	// releaseCNI runs before the stop/delete so the netns is still valid for CNI
+	// DEL where applicable; purgeCNI scrubs the residual allocation file afterward.
 	cellName := strings.TrimSpace(existing.Metadata.Name)
 	cniConfigPath, _ := r.resolveSpaceCNIConfigPath(realmName, spaceName)
 	networkName := r.resolveRootCNINetworkName(realmName, spaceName)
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #688 Review — `fix: start tasks in RecreateCell before stamping Ready`**

LGTM — routes the recreated cell through `StartCell` (which starts tasks, runs CNI ADD, and stamps `Ready` only once up) instead of stamping `Ready` over created-but-not-started container records; mirrors the existing `CreateCell`→`StartCell` create branch, failure flips the cell to `Failed` rather than leaving a phantom `Ready`, and both behaviors are covered by new unit tests.

PR #652, opened 2026-05-21
feat(build): add kuke build + kukebuild binary embedding BuildKit (phase 1)

eminwux:
## Summary

Phase 1 of the kukebuild series (#522): a native image build path that removes the docker / nerdctl + buildkitd dependency from kukeon's "self-contained runtime" pitch.

- **`cmd/kukebuild/`** — a standalone binary that embeds BuildKit as a library. It drives BuildKit's Dockerfile frontend through an **in-process Controller** (served over an in-memory gRPC pipe, driven by a BuildKit client — no long-running buildkitd) backed by a **containerd worker** pointed at the host socket, and writes the resulting OCI image straight into the target realm's containerd namespace (`<realm>.kukeon.io`).
- **`cmd/kuke/build/`** — a thin `kuke build` CLI shim that locates `kukebuild` on PATH and `syscall.Exec`'s it. Like kuketty, kukebuild is *not* argv[0]-dispatched from `kuke`, so BuildKit's transitive moby/runc/grpc closure stays entirely out of the `kuke` + `kukeond` import sets (verified below).
- Makefile `kukebuild` target, `BINS` membership, `make clean`, and per-arch `release-build`.

### Non-obvious implementation calls (reviewer please push back)

- **BuildKit pin: `v0.25.2`.** It is the latest BuildKit release whose `go` directive (`go 1.24.0`) is satisfied by kukeon's pinned toolchain (`go 1.24.5`); v0.26+ require `go 1.25.x`, which would also force bumping the `golang:1.24-bookworm` builder in the repo `Dockerfile` (out of scope). v0.25.2 pins `containerd/v2 v2.1.4` ≤ kukeon's `v2.2.0`, so MVS keeps containerd at v2.2.0 — **no containerd downgrade**, verified by `go.mod` diff. The BuildKit-as-library Go API is explicitly unstable (per the issue notes); the in-process Controller wiring in `build.go` mirrors buildkitd's own `newController`/`newWorkerController` so a future bump is a localized refactor.
- **genproto resolution.** Embedding BuildKit surfaced the classic `google.golang.org/genproto` monolith-vs-split ambiguous-import error. Resolved by pinning the monolith to `v0.0.0-20240814211410` (a post-split-cleanup version with a `go 1.21` directive) rather than `@latest` (which carries a `go 1.25` directive) — keeping the repo `go` directive at `1.24.5`.
[...]

PR #691, opened 2026-05-22
refactor(kukebuild): gate runBuild on a testable requireRoot helper

### LINKED ISSUE #656: kukebuild: route the root gate through shared.RequireRoot / ErrMustRunAsRoot

eminwux:
## Summary

- Replaces the inline `os.Geteuid() != 0` root check in `runBuild` (`cmd/kukebuild/build.go`) with a `requireRoot()` helper backed by a `geteuid` package-var seam + `SetGeteuidForTesting`, mirroring `cmd/kuke/shared.RequireRoot`.
- Adds a kukebuild-local `ErrMustRunAsRoot` sentinel so the non-root gate is classifiable via `errors.Is`, and unit tests (`root_test.go`) that drive both the root and non-root branches without forking under a different UID — the testability the inline form lacked.

## Reinterpreted scope (premise rot — please audit)

Issue #656 was filed during review of **#652**, when `cmd/kukebuild` lived **inside the root module**, so its suggested fix — call `cmd/kuke/shared.RequireRoot` or at least return `errdefs.ErrMustRunAsRoot` — was directly importable.

**#662 later split `cmd/kukebuild` into its own Go module** (`cmd/kukebuild/go.mod`, go1.25.5) with a deliberately-incompatible BuildKit dependency closure vs. the root module (go1.24.5). The `go.work` header documents that the two graphs must never be unioned. Importing *either* `cmd/kuke/shared` *or* `internal/errdefs` from kukebuild now requires a cross-module `require`+`replace` of the root module, dragging its closure into kukebuild's — reintroducing the exact coupling #662 removed. The literal fix is therefore infeasible post-split.

The issue's *intent* — a testable root gate that reuses a classifiable sentinel — is still satisfiable **within** the kukebuild module, which is what this PR does: a local `geteuid` seam + local `ErrMustRunAsRoot` sentinel mirroring the shared pattern, with no cross-module import. Cross-module `errors.Is` classification against the root `errdefs.ErrMustRunAsRoot` was never reachable anyway (kukebuild is invoked as a separate binary by `kuke build`). Reviewer: push back if the local-sentinel redeclaration oversteps vs. simply keeping the inline `errors.New`.

## Test plan

- [x] `cd cmd/kukebuild && GOWORK=off go build ./...`
- [x] `cd cmd/kukebuild && GOWORK=off go vet ./...`
- [x] `cd cmd/kukebuild && GOWORK=off go test ./...` (new `root_test.go` passes)
- [x] `make test` (full CI gate — both modules green)
- [ ] golangci-lint pre-commit hook — structurally cannot analyze the go1.25 kukebuild module (v2.6.2's own module declares go1.24, so it errors `Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.5)`). Not a CI gate: `.github/workflows/test.yaml` runs only `make test` + `make e2e`, no lint step. All other pre-commit hooks (gofmt, addlicense, go-mod-tidy, gitlint) pass.
- [ ] `make dev-init` smoke — N/A: behavior-preserving refactor of the build binary's root gate; touches neither the daemon, `kuke init`, nor `/opt/kukeon`.

Closes #656

### DIFF
--- a/cmd/kukebuild/build.go
+++ b/cmd/kukebuild/build.go
@@ -18,7 +18,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -105,14 +104,8 @@ const containerdDialTimeout = 60 * time.Second
 // resolveNamespaceSuffix). progressW receives the human-readable build
 // progress.
 func runBuild(ctx context.Context, cfg *buildConfig, progressW io.Writer) error {
-	// kukebuild writes directly into containerd's content store; the
-	// standalone containerd socket is root-only on a stock host. Fail fast
-	// under non-root rather than letting containerd surface an opaque EACCES
-	// several phases in. Same posture as `kuke image load --no-daemon`.
-	if os.Geteuid() != 0 {
-		return errors.New(
-			"kukebuild writes to root-owned containerd state — re-run as root (e.g. via `sudo kuke build`)",
-		)
+	if err := requireRoot(); err != nil {
+		return err
 	}
 
 	if fi, err := os.Stat(cfg.contextDir); err != nil {
--- a/cmd/kukebuild/root.go
+++ b/cmd/kukebuild/root.go
@@ -1,0 +1,62 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"errors"
+	"fmt"
+	"os"
+)
+
+// ErrMustRunAsRoot mirrors the root module's errdefs.ErrMustRunAsRoot
+// sentinel. kukebuild is a separate Go module (#662) whose BuildKit
+// closure is deliberately incompatible with the root module's graph, so
+// the shared sentinel can't be imported across the boundary without
+// re-coupling the two — it's redeclared here so kukebuild callers can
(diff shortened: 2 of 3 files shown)


eminwux:
Re-running failed CI jobs once (`make e2e` in the **Test** workflow: https://github.com/eminwux/kukeon/actions/runs/26290797329). Will revisit when they finish; if they fail again the PR will be marked `ci-stalled`.

eminwux:
**PR #691 Review — `refactor(kukebuild): gate runBuild on a testable requireRoot helper`**

LGTM — behavior-preserving extraction of the non-root gate into a `requireRoot()` helper with a `geteuid` test seam and a local `ErrMustRunAsRoot` sentinel; both branches are unit-tested and `errors.Is` classification holds. The local sentinel redeclaration is the right call given the #662 module split — importing `internal/errdefs` would drag the root module's incompatible BuildKit closure back into kukebuild, and cross-binary `errors.Is` against the root sentinel was never reachable anyway.

PR #667, opened 2026-05-21
feat(container): report repo setup status over GetSetupStatus RPC into ContainerStatus (phase 1b)

eminwux:
## Summary

Phase 1b of `containers[].repos[]` (#642, split from #617). Phase 1a (#627) clones/fetches a container's declared repos in kuketty's pre-Serve step but only **logs** the outcome. This wires the locked design: kuketty reports per-repo outcome to `kukeond` over a **custom RPC verb on its existing sbsh control socket** (the same socket `kuke attach` dials), and `kukeond` pulls it post-Serve into `ContainerStatus.Repos`. `ContainerStatus` is the single source of truth — there is no status file in the container.

- **kuketty (producer):** `processRepos` now returns the per-repo outcome (`cloned`/`fetched`/`failed`, with commit/error) instead of logging only; a `GetSetupStatus` verb is registered on the sbsh control server via the new `server.WithHandlers` hook and serves that snapshot.
- **kukeond (consumer):** a stdlib `net/rpc/jsonrpc` client (`pullSetupStatus`) dials the container's kuketty socket and calls `Setup.GetSetupStatus`; `populateCellContainerStatuses` invokes it best-effort for Attachable + Ready containers with `repos[]` and writes `ContainerStatus.Repos`. Any dial/call error leaves `Repos` empty so a still-starting or wedged kuketty never stalls `kuke get`.
- **`kuke get`:** `GetContainer` carries the populated `Repos` slice through, so `kuke get container <name> -o yaml/json` surfaces it.

## ✅ MERGE-BLOCK RESOLVED — re-pinned to tagged sbsh v0.12.0

This depended on the custom-RPC-handler-registration API (`server.WithHandlers`) from **eminwux/sbsh#291** (landed via sbsh PR #294, commit `a101efe`). sbsh has now cut **`v0.12.0`**, which contains that commit (`a101efe` is an ancestor of the `v0.12.0` tag — verified via `gh api .../compare/a101efe...v0.12.0` → `ahead`).

`go.mod` has been re-pinned from the temporary pseudo-version (`v0.11.5-0.20260521084435-a101efe9914c`) to the tagged release **`github.com/eminwux/sbsh v0.12.0`** (commit `build: re-pin sbsh to tagged v0.12.0 (drop pseudo-version)`), per the review note's "same pattern as the v0.11.4 bump." The pseudo-version is gone; this is now the intended merge state. `go mod tidy` pulled in one transitive indirect dep (`github.com/hinshun/vt10x`).

[...]

PR #692, opened 2026-05-22
fix: anchor kuketty terminal metadata.json inside the tty bind mount

### LINKED ISSUE #672: attach: tty dir ownership doesn't cover the terminals/<id> metadata subtree
## Context

The in-cell terminal manager (the `github.com/eminwux/sbsh` runtime, embedded via the `kuketty` wrapper) writes per-terminal metadata under `/run/kukeon/tty/terminals/<id>/metadata.json` using an atomic temp-file + rename (`.meta-*.tmp`). On terminal close it fails with permission denied:

```
WARN  "failed to update metadata on close" id=58ac0c22 err=write /run/kukeon/tty/terminals/58ac0c22/metadata.json: open /run/kukeon/tty/terminals/58ac0c22/.meta-3372523117.tmp: permission denied
ERROR "failed to update terminal state" id=58ac0c22 err=...permission denied
```

(Emitting code lives in sbsh v0.11.1: `internal/terminal/terminalrunner/metadata.go:38,80,94,117` and `lifecycle.go:199`. Filed there too — see Notes.)

kukeon owns the `/run/kukeon/tty` directory tree and sets up its ownership in `internal/controller/runner/attachable.go`:
- `attachableBuildOpts` (≈ lines 109-140) creates the per-container `tty/` leaf dir and its parent, chmods them, and chowns them `root:kukeonGID`.
- `attachablePostCreateChown` (≈ lines 485-513) re-chowns the per-container **`tty/` leaf dir** (and the kuketty metadata file) to the container's *resolved process uid*, so a non-root image USER can create the kuketty socket inside `tty/`.

But the sbsh runtime writes its terminal metadata to a sibling subtree — `<runPath>/terminals/<id>/` — which `attachablePostCreateChown` does **not** chown. sbsh creates that dir itself with `os.MkdirAll(dir, 0o700)` (single-owner), so when create and close run under different uids (the privilege transition between daemon setup and the in-cell process), the close-time write is denied.

Impact: terminal state is never updated on close; stranded/incorrect terminal metadata. Functional attach still works, so this is observability/cleanup, not a hard break.

## Proposal

Ensure kukeon's tty ownership setup covers the path the embedded sbsh runtime actually writes terminal metadata to, across both create and close — i.e. the `terminals/<id>/` subtree must be writable by the same resolved container uid that `attachablePostCreateChown` already grants the `tty/` leaf. Confirm where sbsh's `runPath` resolves inside the cell and extend the chown/mode setup to match, or coordinate the dir creation so sbsh inherits a writable, correctly-owned parent.

## Acceptance criteria
- [ ] After a cell's terminal closes, no `permission denied` writing `/run/kukeon/tty/terminals/<id>/metadata.json`; terminal state updates successfully on close.
- [ ] Works for a non-root image USER (the case `attachablePostCreateChown` exists to support).
- [ ] Test/smoke step: attach to a cell, exit the shell, confirm the terminal metadata reflects the closed state with no permission errors in the daemon/kuketty logs.

## Notes
- Paired upstream issue (sbsh close-write robustness): eminwux/sbsh#345
- kukeon anchors: `internal/controller/runner/attachable.go` `attachableBuildOpts` (≈109-140) and `attachablePostCreateChown` (≈485-513).
- Separate from #670 / #671 (container-state reporting) — this is the tty/attach metadata path.

eminwux:
## Summary
- kuketty now explicitly sets `TerminalSpec.MetadataDir = attachableTTYDir` (`/run/kukeon/tty`) in `cmd/kuketty/spec.go`, so sbsh writes terminal `metadata.json` into the per-container tty bind mount — the directory the daemon owns and `attachablePostCreateChown` re-owns to the resolved container uid — instead of the legacy single-owner `RunPath/terminals/<id>/` (mode 0700) subtree whose ownership denied the close-time write the issue reports.
- Added `TestBuildTerminalSpec_MetadataDirAnchored` pinning the invariant across default, in-bind-mount, and out-of-bind-mount `Tty.LogFile` values.

## Premise rot / reinterpreted scope (please audit)
The issue (#672) was filed against sbsh **v0.11.1**, where the runner unconditionally wrote `RunPath/terminals/<id>/metadata.json` via `MkdirAll(..., 0o700)` — the path named in the issue's permission-denied logs. Current `main` already depends on sbsh **v0.12.0** (bumped in #667, the second-most-recent commit on `main`), which added `TerminalSpec.MetadataDir`: when all three of `LogFile`/`CaptureFile`/`SocketFile` are caller-supplied, sbsh's builder derives `MetadataDir = filepath.Dir(LogFile)` and skips the `terminals/<id>/` subdir (`internal/profile/profile.go:305-337`, `pkg/api/terminal.go:97-106`). kuketty's `buildTerminalSpec` already supplies all three (`spec.go:94,101,139`), so on `main` metadata.json no longer lands in the un-chowned `terminals/<id>/` subtree for the default log path.

So the issue's **literal mechanism** (kukeon must chown `terminals/<id>/`) has rotted — that subtree is dead on `main`, and a chown of it would be a no-op. Per dev CLAUDE.md's premise-rot guidance I proceeded with the **salvageable intent** (terminal metadata persists on close, in the kukeon-owned/host-visible/chowned dir, for a non-root image USER) rather than stalling:

- The implicit `dirname(LogFile)` derivation leaves a residual gap: when an operator pins `Tty.LogFile` outside the bind mount (covered today by `TestBuildTerminalSpec_LogFileOverride`'s `/var/log/sbsh-debug.log` case), `MetadataDir` follows the log out of `/run/kukeon/tty`, dragging metadata.json out of the kukeon-owned, host-visible dir.
- Setting `MetadataDir` explicitly closes that gap and is robust to a future refactor dropping one of the `With*` artifact options (which would otherwise empty `MetadataDir` and resurrect the `terminals/<id>/` layout). For the default config the value is byte-identical to the implicit derivation — **no behavior change** there.

sbsh exposes no `WithMetadataDir` builder option (checked v0.12.0 `pkg/builder/terminal.go`), so the anchor is applied by mutating the returned `*TerminalSpec` before it is handed to `sbshserver.New` (`cmd/kuketty/main.go:122,153`).

## Test plan
- [x] `make test` (full suite incl. `cmd/kukebuild`) — green, 92 pkg ok
- [x] `GOWORK=off go build ./...`, `go vet ./cmd/kuketty/...` — clean
- [x] `go test ./cmd/kuketty/...` — incl. the new `TestBuildTerminalSpec_MetadataDirAnchored`
- [ ] Live attach→exit→inspect smoke (AC #3's manual step: attach to a cell, exit the shell, confirm terminal metadata reflects the closed state with no permission errors) — not run: needs a live daemon + `kuke attach` plumbing + a non-root-USER image, an environmental e2e tier not available in this dev sandbox. The transform-level invariant is covered by the new unit test; defer the host smoke to a reviewer environment.

Closes #672

### DIFF
--- a/cmd/kuketty/spec.go
+++ b/cmd/kuketty/spec.go
@@ -155,6 +155,26 @@ func buildTerminalSpec(
 	if err != nil {
 		return nil, fmt.Errorf("build kuketty terminal spec: %w", err)
 	}
+
+	// Anchor sbsh's metadata.json inside the per-container tty bind mount
+	// (issue #672). sbsh writes metadata.json via an atomic
+	// create-temp-then-rename that needs write+exec on the holding directory at
+	// every point in the terminal lifecycle — create AND close. The legacy
+	// layout (RunPath/terminals/<id>/, mode 0700, owned by the creating uid)
+	// denied the close-time write whenever the closing uid differed from the
+	// creating uid. We pin MetadataDir to attachableTTYDir — the directory the
+	// daemon bind-mounts and attachablePostCreateChown re-owns to the resolved
+	// container uid — so the holding dir is always the kukeon-owned, host-
+	// visible tty dir. Setting it explicitly (rather than leaning on sbsh's
+	// implicit "MetadataDir = dirname(LogFile) when all three artifact paths are
+	// caller-supplied" derivation) keeps metadata.json in the bind mount even
+	// when an operator pins Tty.LogFile outside it, and survives a future
+	// refactor that drops one of the WithSocketFile/WithCaptureFile/WithLogFile
+	// options (which would otherwise empty MetadataDir and resurrect the
+	// terminals/<id>/ layout). No basename collision: kuketty's own rendered doc
+	// is kuketty-metadata.json; sbsh writes metadata.json.
+	terminalSpec.MetadataDir = attachableTTYDir
+
 	return terminalSpec, nil
 }
 
--- a/cmd/kuketty/spec_test.go
+++ b/cmd/kuketty/spec_test.go
@@ -245,6 +245,34 @@ func TestBuildTerminalSpec_LogFileOverride(t *testing.T) {
 	}
 }
 
+// TestBuildTerminalSpec_MetadataDirAnchored locks #672: metadata.json is always
+// written into the per-container tty bind mount (attachableTTYDir), the dir the
+// daemon owns and attachablePostCreateChown re-owns to the resolved container
+// uid — never the legacy single-owner RunPath/terminals/<id>/ subtree whose
+// 0700/creating-uid ownership denied the close-time write. The anchor must hold
+// regardless of where the operator points Tty.LogFile: leaning on sbsh's
+// implicit "MetadataDir = dirname(LogFile)" derivation would let an
+// out-of-bind-mount log path (the "override outside bind mount" case) drag
+// metadata.json out of the kukeon-owned, host-visible dir.
+func TestBuildTerminalSpec_MetadataDirAnchored(t *testing.T) {
+	cases := []struct {
+		name string
+		tty  *v1beta1.ContainerTty
+	}{
+		{"no Tty: daemon default log", nil},
+		{"log inside bind mount", &v1beta1.ContainerTty{LogFile: attachableTTYDir + "/custom.log"}},
+		{"log outside bind mount", &v1beta1.ContainerTty{LogFile: "/var/log/sbsh-debug.log"}},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			out := buildSpec(t, v1beta1.ContainerSpec{ID: "c1", Command: "/bin/sh", KukeonGroupGID: 986, Tty: tc.tty})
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #692 Review — `fix: anchor kuketty terminal metadata.json inside the tty bind mount`**

LGTM — pins `TerminalSpec.MetadataDir = attachableTTYDir` after `BuildTerminalSpec` returns; verified the fix actually takes effect: sbsh's `dirname(LogFile)` derivation runs only inside the builder (guarded on `MetadataDir == ""`), the mutated spec flows unmodified into `sbshserver.New` (`main.go:122→153`), and at runtime `resolveMetadataDir` returns any non-empty override verbatim (`internal/terminal/terminalrunner/metadata.go:133`). The new test exercises the real `buildTerminalSpec` path and its out-of-bind-mount case locks the residual gap. Premise-rot reinterpretation is well-reasoned and the explicit set is the only mechanism (sbsh v0.12.0 exposes no `WithMetadataDir` option).

Defer pre-merge: the live attach→exit→inspect smoke (AC #3) is host-gated and unrunnable on the agent sandbox — worth a manual pass before merge.

PR #695, opened 2026-05-22
fix: serialize lazy ctrClient init with sync.Once

### LINKED ISSUE #684: runner: remove lazy ctrClient init race in ensureClientConnected

eminwux:
## Summary
- `ensureClientConnected` did an unguarded nil-check-then-assign of `r.ctrClient` (`internal/controller/runner/runner.go`). Two concurrent first-use goroutines (e.g. the first RPC handler and the first reconcile tick) could each observe nil, each construct a client, and the second assignment overwrite the first — a data race on the pointer plus a duplicate containerd connection.
- Fix guards the lazy construction with a `sync.Once`, the lower-blast-radius option the issue offers alongside eager init. Eager init in `NewRunner` was rejected: `TestMarkCellFailed_PreStampsBeforeKillCell` builds `&Exec{}` directly with no `ctrClient` and relies on lazy init creating a real client that errors at connect (no socket); removing the nil-check would nil-panic it.
- The nil-check is kept *inside* the `Once.Do` so a test-injected fake (`&Exec{ctrClient: fake}`, used by `delete_cell_test.go` et al.) is not clobbered on first use; `Connect()` stays outside the Once so the reconnect-on-broken-connection path still runs every call.
- **Review fix-up (`dbde4e4`):** the `Once` only serialized wrapper construction — the two first-use goroutines then both called `Connect()` concurrently on the shared `*client`, racing on `c.cClient` (`internal/ctr/client.go`) with no lock and each dialing containerd (the duplicate-connection harm #684 names). Added a `cClientMu sync.Mutex` to `ctr.client` guarding the read-modify-write of `c.cClient` in `Connect()`/`Close()`. Extracted `closeLocked()` (the un-locked teardown body) so `Connect()`'s internal reconnect/verify-failure paths reuse it without re-entrant locking, while public `Close()` takes the lock then delegates. Post-connect reads of `c.cClient` (image.go, container.go, …) run after the connection is established and are sequenced behind it, so they stay outside the lock — the fix is scoped to the construction/teardown race the reviewer flagged.

## Test plan
- [x] `make test` (CI target) — passes (main module + `cmd/kukebuild`).
- [x] `go build ./internal/ctr/... ./internal/controller/runner/...` / `go vet` — clean.
- [x] `TestEnsureClientConnected_ConcurrentFirstUseSingleClient` — 32 goroutines hit first-use concurrently; asserts a single, stable client pointer (every goroutine observes the same one). Runner context carries a 100ms deadline so the dial against a nonexistent socket fails fast.
- [x] `TestEnsureClientConnected_PreservesInjectedClient` — 16 concurrent calls; asserts an injected fake survives (not overwritten) and `Connect` is called once per call.
- [ ] `go test -race` — not run on the dev host: no C compiler (`gcc`/`cc`/`clang` absent, re-verified this session), and `-race` requires cgo. CI's `make test` target does not use `-race` either. With the fix-up, the `c.cClient` read-modify-write in `Connect()`/`Close()` is now fully mutex-guarded, so `TestEnsureClientConnected_ConcurrentFirstUseSingleClient` is race-free by construction on a cgo-enabled host; verified functionally under plain `go test`.

Note: pre-commit `golangci-lint` hook panics on this host (go1.24 binary type-checking the go1.25 module via the `go.work` union — environmental, not a finding); all other hooks (go-fmt, go-mod-tidy, addlicense, whitespace) pass.

Closes #684

### DIFF
--- a/internal/controller/runner/runner.go
+++ b/internal/controller/runner/runner.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"io"
 	"log/slog"
+	"sync"
 	"time"
 
 	"github.com/eminwux/kukeon/internal/cni"
@@ -168,6 +169,13 @@ type Exec struct {
 	opts   Options
 
 	ctrClient ctr.Client
+	// ctrClientOnce serializes the lazy construction of ctrClient so two
+	// concurrent first-use goroutines (e.g. the first RPC handler and the
+	// first reconcile tick) cannot each build a client and overwrite the
+	// other's assignment — a data race on the pointer plus a duplicate
+	// containerd connection (issue #684). Connect() itself stays outside the
+	// Once so the reconnect-on-broken-connection path still runs every call.
+	ctrClientOnce sync.Once
 
 	cniConf *cni.Conf
 
@@ -274,8 +282,13 @@ func (r *Exec) Close() error {
 // ensureClientConnected ensures the containerd client is initialized and connected.
 // It creates a new client if needed, and reconnects if the connection was closed.
 func (r *Exec) ensureClientConnected() error {
-	if r.ctrClient == nil {
-		r.ctrClient = ctr.NewClient(r.ctx, r.logger, r.opts.ContainerdSocket)
-	}
+	r.ctrClientOnce.Do(func() {
+		// Guard the nil check so a test-injected fake (constructed via
+		// &Exec{ctrClient: fake}) survives — only a runner that started with
+		// no client builds the real one here.
+		if r.ctrClient == nil {
+			r.ctrClient = ctr.NewClient(r.ctx, r.logger, r.opts.ContainerdSocket)
+		}
+	})
 	return r.ctrClient.Connect()
 }
--- a/internal/controller/runner/runner_test.go
+++ b/internal/controller/runner/runner_test.go
@@ -1,0 +1,129 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//nolint:testpackage // exercises *Exec.ensureClientConnected against in-package state
+package runner
+
+import (
+	"context"
+	"io"
+	"log/slog"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/eminwux/kukeon/internal/ctr"
+)
+
+// TestEnsureClientConnected_ConcurrentFirstUseSingleClient pins issue #684:
+// two concurrent first-use goroutines must not each construct a ctr.Client and
+// overwrite the other's assignment to r.ctrClient. The sync.Once guard
+// serializes construction, so every goroutine observes the same client
+// pointer. Run under `go test -race` to also catch the unguarded
+// nil-check-then-assign data race the fix removes.
+//
+// The socket points at a nonexistent path and the runner context carries a
+// short deadline, so the constructed client's Connect fails fast without a
+// real containerd; the error is irrelevant here — the assertion is on the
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #695 Review — `fix: serialize lazy ctrClient init with sync.Once`**

The `sync.Once` correctly removes the `r.ctrClient` wrapper-construction race and the duplicate-wrapper, and keeping the nil-check *inside* `Once.Do` correctly preserves an injected fake. The two tests are well-constructed. One blocking item: the fix doesn't fully close the concurrent-first-use race #684 names.

- **Residual race / duplicate connection** `internal/ctr/client.go:174,195`: `Once` serializes wrapper construction, but the two first-use goroutines the PR describes both then call `Connect()` concurrently on the shared `*client`. `c.cClient` is read at `:174` and written at `:195` with no lock (the `client` struct at `client.go:37` has mutexes for cgroups/containers/tasks but none for `cClient`), so concurrent first use still data-races on `c.cClient`, and each goroutine seeing `nil` calls `containerd.New` — leaking N-1 connections, the exact "duplicate containerd connection" harm #684 names. `TestEnsureClientConnected_ConcurrentFirstUseSingleClient` drives this with a real client, so under `-race` (the test's documented validation) it reports this race rather than running clean. Guard `c.cClient` with a mutex in `ctr.client`, or serialize the first `Connect()`.

Changes requested on the item above — the rest LGTMs once the concurrent-`Connect` path is made single-connection/race-free.

eminwux:
Addressed in dbde4e4:
- **Residual race / duplicate connection (`internal/ctr/client.go`)** — confirmed: the `sync.Once` serialized only the wrapper construction; both first-use goroutines then called `Connect()` concurrently on the shared `*client`, racing on `c.cClient` (read at the connected-check, written at the `containerd.New` assign) with no lock, each dialing a fresh connection. Added a `cClientMu sync.Mutex` guarding the read-modify-write of `c.cClient` across `Connect()`/`Close()`. Extracted an un-locked `closeLocked()` so `Connect()`'s reconnect and verify-failure paths reuse the teardown while already holding the lock (no re-entrant deadlock); public `Close()` takes the lock then delegates. First use is now single-connection and race-free.

Scoping note: post-connect reads of `c.cClient` (image.go, container.go, namespaces.go, …) run after the connection is established and are sequenced behind it, so they're left outside the lock — the fix targets the construction/teardown race #684 and the review name, not a package-wide `c.cClient` RWMutex (that'd be scope creep).

`make test` green; `go test -race` still deferred — no C compiler on the dev host (re-verified), and the guard makes the concurrent-first-use test race-free by construction.

eminwux:
Re-running failed CI jobs once (`make e2e`). Will revisit when they finish; if they fail again the PR will be marked `ci-stalled`.

eminwux:
**PR #695 Review — `fix: serialize lazy ctrClient init with sync.Once`**

LGTM — the two-layer serialization is correct and tightly scoped to the #684 first-use race. The `sync.Once` in `runner.go` guards wrapper construction (nil-check kept inside so test-injected fakes survive), and the `cClientMu` in `ctr.client` guards the `c.cClient` read-modify-write in `Connect()`/`Close()`. I verified the lock discipline: `verifyConnection()` only reads `c.cClient` and never locks, the sole `.Close()` inside `client.go` is `c.cClient.Close()` (the containerd client, not the wrapper), and `Connect()`'s reconnect path goes through `closeLocked()` — so no re-entrant acquisition of the non-reentrant mutex. `Connect()` correctly stays outside the `Once` so reconnect still runs every call. Tests are well-constructed; all commits verified.

One deferred verification (non-blocking): `go test -race` is unchecked because the dev host lacks a C compiler (cgo). Worth running once pre-merge on a cgo-enabled host — the concurrency tests (`TestEnsureClientConnected_ConcurrentFirstUseSingleClient`) only assert race-freedom by construction without it.

PR #669, opened 2026-05-21
feat(secret): add ContainerSecret.secretRef source + resolution path (phase 3c)

eminwux:
## Summary
- Adds a third `ContainerSecret` source, `secretRef`, pointing at a daemon-managed `kind: Secret` (#619) by name + scope, alongside the existing `fromFile`/`fromEnv`.
- Container start resolves `secretRef` → reads bytes from the #619 storage layout via `fs.SecretPath`, then feeds the existing `resolvedSecrets` env/mount dispatch (no new injection machinery).
- Apply-time validation: exactly one of `fromFile`/`fromEnv`/`secretRef`; for `secretRef`, name + realm required and deeper scope coordinates require every shallower one.
- **AC5 delete-blocking guard (added in review fix-up `bbc4169`):** `kuke delete secret` now refuses while any persisted cell's container references the secret via `secretRef`, with an error naming the referencing cells (`ErrSecretInUse`). Matching is by resolved `fs.SecretPath`, so a reference and the target are "the same secret" exactly when they resolve to the same on-disk file. The stale `unsafeDeleteWarning` help text and its test were replaced with the new guard contract.

## Implementation notes (for reviewer)
- **AC storage-path divergence:** the issue's AC cites `/opt/kukeon/<scope>/secrets/<name>`, but the actual #619 layout is `<RunPath>/data/<scope>/secrets/<name>`. Reused `fs.SecretPath` (the #619 helper) rather than hardcoding the AC's literal path, so resolution tracks the real layout. RunPath is threaded into the spec builder via a new `WithSecretRunPath` option, injected once in `daemonDefaultBuildOpts`.
- **AC5 placement (resolved):** the issue's ordering note assigns the delete-blocking AC to "whichever lands second" between #622/#3b and this. Phase 3b (#664) has merged, so per the issue's own framing it lands here — implemented in `internal/controller/secret.go`'s `DeleteSecret` (the guard scans `runner.ListCells` for matching `secretRef`s). The earlier deferral note was wrong and is withdrawn.
- **"Live" scope:** the guard blocks on any cell whose persisted spec references the secret (not just running cells) — deleting the bytes would break a stopped-but-declared cell on its next start. Mirrors the `DeleteRealm` dependency guard.

## Test plan
- [x] `make test` (full CI target, `GOWORK=off`; 92 packages green incl. new guard tests in `internal/controller/secret_test.go`)
- [x] `go vet ./...` (`GOWORK=off`)
- [x] `go build ./...` (`GOWORK=off`)
[...]

PR #698, opened 2026-05-22
fix: reject path-traversal in secret name and scope coordinates

### LINKED ISSUE #673: secret: sanitize secretRef/Secret name + scope coords against path traversal
Surfaced during review of #669 (phase 3c, `ContainerSecret.secretRef`).

`secretRef.name` and its scope coordinates (`realm`/`space`/`stack`/`cell`) are validated only for non-empty + coordinate contiguity in `validateSecretRef` (`internal/apply/parser/parser.go:303`). They then flow unsanitized into `fs.SecretPath` → `filepath.Join` → `os.ReadFile` as root in `readSecretRefValue` (`internal/ctr/secrets.go:847`). A crafted name such as `../../../etc/shadow` would `filepath.Join`-clean out of the secrets tree, so the daemon reads an arbitrary host file and injects it into a container env var / mount.

Not a privilege boundary in the current model (root-owned daemon socket, operator-authored manifests), which is why #669 was not blocked on it. But it's a latent risk and the same gap exists on the `kind: Secret` **write** path — `validateSecretScope` (`internal/apply/parser/parser.go:447`, issue #619) is equally permissive on names/coords.

**Suggested fix:** add a shared scope/name validator (reject `..`, path separators, empty-after-clean) and apply it at *both* `validateSecretRef` and `validateSecretScope` so the whole secret subsystem is covered, rather than patching only the secretRef site.

eminwux:
## Summary
- `validateSecretRef` and `validateSecretScope` (`internal/apply/parser/parser.go`) validated secret names and scope coordinates only for non-empty + coordinate contiguity. The values then flow unsanitized into `fs.SecretPath` → `filepath.Join` (`internal/util/fs/metadata.go:271`) and on into `os.ReadFile` as root via `readSecretRefValue` (`internal/ctr/secrets.go:190`). A crafted name or coordinate such as `../../../etc/shadow` would `filepath.Join`-clean out of the secrets tree, letting the daemon read an arbitrary host file into a container env var / mount.
- Adds a shared `validateSecretSegment` helper that rejects any non-empty name or coordinate containing a `/`/`\`/`filepath.Separator`, a NUL byte, or that is exactly `.`/`..`, and applies it at **both** validators so the whole secret subsystem (`kind: Secret` write path and `secretRef` read path) is covered — per the issue's suggested fix.
- New `errdefs.ErrSecretCoordUnsafe` sentinel so the rejection is classifiable via `errors.Is`, matching the existing `errdefs` convention for the secret subsystem.

## Notes
- Not a privilege boundary in the current model (root-owned daemon socket, operator-authored manifests), which is why #669 was not blocked on it — this closes the latent risk as defense-in-depth.
- Empty coordinates remain legal (a missing coordinate means "not scoped at that level"); the segment guard only fires on non-empty values, and the existing realm-required / contiguity checks run first.

## Test plan
- [x] `make test` (full CI gate — both modules green, exit 0, zero failures).
- [x] `go build ./...` / `go vet ./internal/apply/parser/... ./internal/errdefs/...` — clean.
- [x] Extended `TestValidateDocument_Secret` and `TestValidateDocument_Container_SecretRefValidation` with traversal cases (`..`, embedded `/`, dot-dot coordinate) on both the `kind: Secret` and `secretRef` paths; assert `ErrSecretCoordUnsafe`.
- [ ] golangci-lint pre-commit hook — environmental panic on this host (`package requires newer Go version go1.25 (application built with go1.24)`); not a CI gate (`.github/workflows` runs `make test` + `make e2e`). All other hooks (gofmt, go-mod-tidy, addlicense, whitespace) pass.
- [ ] `make dev-init` smoke — N/A: apply-time parser validation only; touches neither the daemon, the build path, nor `/opt/kukeon`.

Closes #673

### DIFF
--- a/internal/apply/parser/parser.go
+++ b/internal/apply/parser/parser.go
@@ -438,6 +438,28 @@ func ValidateDocument(doc *Document) *ValidationError {
 	return nil
 }
 
+// validateSecretSegment rejects a single secret name or scope coordinate that
+// would escape the secrets tree once fs.SecretPath filepath.Join's it into a
+// host path: a "/" or "\" separator, a "." or ".." element, or a NUL byte.
+// Empty values are the caller's concern (a missing coordinate is legal; an
+// empty name is rejected separately), so this only guards non-empty segments.
+// Issue #673.
+func validateSecretSegment(value string) error {
+	if value == "" {
+		return nil
+	}
+	if value == "." || value == ".." {
+		return errdefs.ErrSecretCoordUnsafe
+	}
+	if strings.ContainsRune(value, 0) ||
+		strings.ContainsRune(value, '/') ||
+		strings.ContainsRune(value, '\\') ||
+		strings.ContainsRune(value, filepath.Separator) {
+		return errdefs.ErrSecretCoordUnsafe
+	}
+	return nil
+}
+
 // validateSecretScope enforces the `kind: Secret` scope-coordinate contract:
 // metadata.realm is always required, and a deeper coordinate may only be set
 // when every shallower one is also set (a cell-scoped secret must name its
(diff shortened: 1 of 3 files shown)


eminwux:
Re-running failed CI jobs once (`make e2e`). Will revisit when they finish; if they fail again the PR will be marked `ci-stalled`.

eminwux:
**PR #698 Review — `fix: reject path-traversal in secret name and scope coordinates`**

LGTM — adds a shared `validateSecretSegment` guard at both the `kind: Secret` and `secretRef` validators, rejecting `/`/`\`/NUL/`.`/`..` before any name or coordinate reaches `fs.SecretPath`, with `%w`-wrapped errors and traversal-case tests on both paths.

PR #699, opened 2026-05-22
fix: gate run/attach on root-container task liveness, not record existence

### LINKED ISSUE #683: run/attach: gate phantom attach on task liveness, not record existence
## Context
The #657 (1503273) phantom-attach guard at `cmd/kuke/run/run.go:317` keys on `pre.RootContainerExists` -> `ExistsCellRootContainer` -> `ExistsContainer` (`internal/controller/runner/exists.go:64`), which is containerd container-*record* existence. Records survive a host reboot (`internal/controller/runner/refresh.go` documents "container records survive, tasks gone"), so the guard passes even when the backing task is dead, and `kuke run` attaches to a dead socket. `kuke attach` (`cmd/kuke/attach/attach.go:124`) has no guard at all — it calls `AttachContainer` directly.

## Proposal
Introduce one task-liveness guard (task status == Running) and apply it on both the run and attach paths.

## Acceptance criteria
- [ ] Against a cell whose metadata says `Ready` but whose task is gone (post-reboot), both `kuke run` and `kuke attach` refuse with the diverged-state error instead of attaching.
- [ ] A single shared guard backs both entry points.
- [ ] Tests cover the task-dead case on both `kuke run` and `kuke attach`.

## Notes
Strengthens #657 and extends it to the attach path. The reboot window is enabled by there being no eager reconcile-on-startup (`internal/daemon/server.go:170` waits a full reconcile interval before the first pass). Complementary to the RecreateCell fix (filed separately), which removes one producer of the diverged state this guard defends against.

eminwux:
## Summary
- The #654 phantom-attach guard at `cmd/kuke/run/run.go` keyed on `RootContainerExists`, which is containerd container-*record* existence. Records survive a host/daemon restart while their tasks do not, so the guard passed even when the backing task was dead and `kuke run` attached to a dead socket. `kuke attach` had no guard at all.
- Gate on task liveness instead: a new `RootContainerTaskRunning` signal on `GetCellResult` reports whether the root container's containerd task is actually Running (`GetContainerState == ContainerStateReady`). A single shared guard (`shared.GuardCellTaskLiveness`) backs both `kuke run`'s Ready short-circuit and `kuke attach`, refusing with the same diverged-state error + `kuke delete cell` pointer.
- Scoped to Ready (a legitimately Stopped cell has no live task by design and is handled by run's start path). Per the issue proposal, only a Running task counts as live — a record-present-but-task-gone, NotCreated, or transient-Unknown root all refuse.

## Design notes
- `controller.GetCell` now queries the root container's live task state only when `RootContainerExists` is true; a cell whose spec carries no root container is treated as live (matching the reconciler's `deriveCellStateFromRootContainer`).
- `kuke attach` previously issued no `GetCell`; it now does one before `AttachContainer` to evaluate the guard. The guard returns nil for non-Ready cells, so the attach error surface for Stopped/missing cells is unchanged.

## Test plan
- [x] `make test` (root module + kukebuild module) — pass
- [x] `go build ./cmd/... ./internal/... ./pkg/...` — pass
- [x] `go vet` on changed packages — pass
- [x] New: `TestRun_ExistingCell_RecordedReady_TaskGone_Refuses` (run refuses when record present but task gone)
- [x] New: `TestAttach_RecordedReady_TaskGone_Refuses` (attach refuses, never reaches AttachContainer / sbsh loop)
- [x] New controller cases: `RootContainerTaskRunning` true (Running task) / false (record present, task Stopped) / error path
- [ ] Local smoke test (`make dev-init`) — not run: this change is in the CLI guard / controller read path, touches no build path, daemon serve path, or `/opt/kukeon` layout, so the daemon-parity regression guard does not apply.

Closes #683

### DIFF
--- a/cmd/kuke/attach/attach.go
+++ b/cmd/kuke/attach/attach.go
@@ -120,6 +120,19 @@ func runAttach(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	// Divergence guard (#683): refuse to attach to a cell whose metadata records
+	// it Ready but whose root-container task is gone from containerd (post-reboot
+	// divergence). Without this, attach hands back a socket backed by a dead task
+	// and the sbsh loop fails with `connection refused`. Shares the run path's
+	// guard so both entry points refuse identically.
+	cellGet, err := client.GetCell(cmd.Context(), buildCellDoc(cell, realm, space, stack))
+	if err != nil {
+		return err
+	}
+	if err := kukeshared.GuardCellTaskLiveness(cellGet, cell); err != nil {
+		return err
+	}
+
 	doc := buildContainerDoc(container, realm, space, stack, cell)
 	result, err := client.AttachContainer(cmd.Context(), doc)
 	if err != nil {
@@ -160,6 +173,25 @@ func resolveRun(cmd *cobra.Command) runFn {
 		return mock
 	}
 	return attach.Run
+}
+
+// buildCellDoc assembles the lookup CellDoc the divergence guard queries via
+// GetCell. Only the identity coordinates are needed — GetCell resolves the
+// persisted spec and the live root-task status from them.
+func buildCellDoc(cell, realm, space, stack string) v1beta1.CellDoc {
+	return v1beta1.CellDoc{
+		APIVersion: v1beta1.APIVersionV1Beta1,
+		Kind:       v1beta1.KindCell,
+		Metadata: v1beta1.CellMetadata{
+			Name:   cell,
+			Labels: make(map[string]string),
+		},
+		Spec: v1beta1.CellSpec{
+			RealmID: realm,
+			SpaceID: space,
+			StackID: stack,
+		},
+	}
 }
 
 func buildContainerDoc(name, realm, space, stack, cell string) v1beta1.ContainerDoc {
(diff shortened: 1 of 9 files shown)


eminwux:
**PR #699 Review — `fix: gate run/attach on root-container task liveness, not record existence`**

LGTM — gating on `RootContainerTaskRunning` (root task actually Running) instead of record existence correctly closes the #683 post-reboot divergence: records outlive tasks, so the old `RootContainerExists` guard passed onto a dead socket. The single shared `GuardCellTaskLiveness` backs both `kuke run`'s Ready short-circuit and `kuke attach` identically; the #654 record-gone case still refuses (record absent ⇒ `RootContainerTaskRunning` stays false). Both `GetCellResult` mirrors carry the new field, the controller only queries task state when the record exists, and tests exercise run-refuses / attach-refuses / controller true-false-error. No secret leaks, no dead symbols, commits verified.

One non-blocking nit:

- **Wording** `cmd/kuke/shared/cell_liveness.go:50`: the diverged-state error says "its containers are gone from containerd," but in the #683 case the container *record* survives — only the backing task is gone. Remedy (`kuke delete cell` + rerun) is identical and correct, so this is purely cosmetic; "its root-container task is gone" would read more precisely now that the guard covers both cases.

Smoke-test omission is justified — the change is a read-path/CLI-guard addition that touches no build path, daemon serve path, or `/opt/kukeon` layout, so the daemon-parity regression guard doesn't apply.

PR #701, opened 2026-05-22
ci: split kukebuild tests into a distinct CI step

### LINKED ISSUE #696: ci: surface kukebuild test results as a distinct CI signal
## Context

kukeon's CI `test` job (`.github/workflows/test.yaml`) runs `make test`, whose recipe is two sequential commands:

```make
test:
	go test $(shell go list ./... | grep -v /e2e)
	cd cmd/kukebuild && go test ./...
```

The second line tests the standalone `cmd/kukebuild` module (deliberately split from the root module — see `go.work`'s GOWORK=off contract). Both run inside one `make test` step, so the CI summary shows a single pass/fail for the pair. When the run is red, you can't tell from the job status alone whether the root module or the kukebuild module failed — you have to open the log. Minor, but it costs a click on every failure and obscures which component regressed.

## Proposal

Give kukebuild its own visible CI signal. Lowest-friction option: split the kukebuild line into its own workflow step (e.g. a `make test-kukebuild` target invoked as a separate step in the existing `test` job), so a failure names the component in the step status. A dedicated parallel job is also acceptable but heavier (extra checkout + Go setup). Keep `GOWORK=off` semantics intact (the Makefile already exports it).

## Acceptance criteria

- [ ] A red kukebuild test surfaces as a distinctly-named failed step/job in the CI summary, separate from the root-module tests.
- [ ] Root-module and kukebuild tests still both run on PRs/pushes touching `cmd/kukebuild/**` (the `**/*.go`, `**/go.mod`, `**/go.sum` path triggers already cover this).
- [ ] `GOWORK=off` resolution is preserved for the kukebuild step (no workspace-union ambiguous-import regression).

## Notes

Non-blocking nit — surfaced while measuring kukebuild test coverage (66.0% of statements, `GOWORK=off go test -cover ./...`). Related but distinct from #693 (golangci-lint cannot analyze the go1.25 kukebuild module). No behavior change to `make test` locally is required, only CI-visible granularity.

eminwux:
## Summary
- The CI `test` job ran `make test` as a single step, whose recipe runs both the root-module unit tests and the standalone `cmd/kukebuild` module tests sequentially. A red run gave one combined status, so you couldn't tell from the job summary which module regressed without opening the log.
- Split the Makefile `test` recipe into `test-root` + `test-kukebuild` sub-targets (with `test: test-root test-kukebuild`), so `make test` still runs both locally — no local behavior change — while CI can invoke each independently.
- Replaced the single "Run unit tests" workflow step with two distinctly-named steps: "Run root-module tests" (`make test-root`) and "Run kukebuild module tests" (`make test-kukebuild`). A failure now names the component in the step status.
- `GOWORK=off` is preserved by the Makefile's existing top-level `export GOWORK := off`, which reaches every recipe — no workspace-union ambiguous-import regression.

## Test plan
- [x] `make test-root` — root-module unit tests pass (e2e excluded)
- [x] `make test-kukebuild` — `ok github.com/eminwux/kukeon/cmd/kukebuild`
- [x] `GOWORK=off go vet ./...` (e2e excluded) — clean
- [x] `python3 -c "import yaml; yaml.safe_load(...)"` — workflow YAML parses
- [ ] `make e2e` — not run; unaffected by this change (e2e job and `test-e2e` target untouched) and requires root + docker + standalone containerd not available here

Closes #696

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -53,8 +53,16 @@ jobs:
           go-version-file: go.mod
           cache: true
 
-      - name: Run unit tests
-        run: make test
+      # Split per Go module so a red run names the failing component in the
+      # step status (issue #696): root-module unit tests vs. the standalone
+      # cmd/kukebuild module. Both still run on every PR/push; GOWORK=off is
+      # preserved by the Makefile's top-level export, so neither step re-couples
+      # the workspace-union import graph.
+      - name: Run root-module tests
+        run: make test-root
+
+      - name: Run kukebuild module tests
+        run: make test-kukebuild
 
   # `make e2e` needs three host prerequisites — root (sudo -E), a docker
   # daemon for `docker build` of the local kukeond image, and a standalone
--- a/Makefile
+++ b/Makefile
@@ -136,8 +136,16 @@ kill:
 # tests (excluding the e2e suite) and the kukebuild module's tests. `go list
 # ./...` from the root lists only root-module packages — the nested kukebuild
 # module is excluded — so its tests are run explicitly from its module dir.
-test:
+# Split into per-module sub-targets so CI can run each as its own named step
+# and a red run names the failing component (issue #696); `make test` keeps
+# running both locally. GOWORK=off (exported above) reaches every sub-recipe.
+.PHONY: test test-root test-kukebuild
+test: test-root test-kukebuild
+
+test-root:
 	go test $(shell go list ./... | grep -v /e2e)
+
+test-kukebuild:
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #701 Review — `ci: split kukebuild tests into a distinct CI step`**

LGTM — splits `make test` into faithful `test-root`/`test-kukebuild` sub-targets (`.PHONY` declared, `GOWORK=off` still reaches each via the top-level export) and mirrors them as two named CI steps, so a red run names the failing module; no behavior change locally.

PR #707, opened 2026-05-22
test: prevent e2e parallel namespace collisions via process-wide name entropy

### LINKED ISSUE #704: e2e: unique-name helpers collide (8-bit entropy) → parallel tests share a namespace and tear down each other's pulls
## Context

CI hit a spurious failure in `TestDeleteF_Container_VerifyAllResourcesDeleted`:

```
Cell "e-ce-f4": failed
  Error: failed to create cell: failed to create root container: failed to pull image registry.eminwux.com/busybox:latest: lease does not exist: not found
```

This is **not** a registry-availability or production lease-management bug. The pull got far enough to start writing content under a lease, then the lease was deleted out from under it — the tell is the error names a *missing lease*, not a network/auth/manifest failure.

Root cause is a test-harness name collision. All e2e tests run `t.Parallel()` and share the single host containerd; isolation rests entirely on each test getting a unique realm → unique containerd namespace. But the "unique" name helpers collapse `time.Now().UnixNano()` to its low byte:

```go
// e2e/e2e_kuke_realm_test.go:34
hexID := fmt.Sprintf("%02x", timestamp&0xFF)   // only 256 possible values: e-r-00 .. e-r-ff
return fmt.Sprintf("e-r-%s", hexID)
```

The same `& 0xFF` pattern is in all five generators:
- `generateUniqueRealmName` — `e2e/e2e_kuke_realm_test.go:34`
- `generateUniqueSpaceName` — `e2e/e2e_test.go:357`
- `generateUniqueStackName` — `e2e/e2e_test.go:454`
- `generateUniqueCellName` — `e2e/e2e_test.go:558`
- `generateUniqueContainerName` — `e2e/e2e_test.go:566`

With many tests firing in parallel, two collide on `e-r-XX` routinely (the failing cell `e-ce-f4` is exactly this pattern). A realm name maps straight to a containerd namespace (`e-r-XX.kukeon.io`), so colliding tests share state. The race:

1. Test A applies `cell.yaml` → `pullImage` creates a lease and starts pulling busybox into `e-r-f4.kukeon.io` (`internal/ctr/image.go:106`, `:150`).
2. Test B reaches `t.Cleanup` → `cleanupRealm` → daemon purge path runs `drainLeases`, which lists **every** lease in `e-r-f4.kukeon.io` and deletes them with `leases.SynchronousDelete` (`internal/ctr/namespaces.go:250-261`) — including Test A's in-flight pull lease.
3. Test A's `Pull` fails: `lease does not exist: not found`.

Confirm by re-running with `-p 1` (or the single test in isolation): the collision and the flake disappear.

## Proposal

Give the e2e name generators enough entropy that parallel tests in the same `go test` process cannot share a realm/space/stack/cell/container name — **without** lengthening the names enough to trip the path-length limits the harness already fights.

Constraints that bound the fix (do not regress these):
- Realm/space/stack/cell names feed the containerd namespace (`<...>.kukeon.io`) and the on-disk metadata path under `/opt/kukeon/data/<ns>/...`. Long names risk metadata-path and namespace-label-length errors.
- Run-path sockets already overflow `SUN_PATH` — `getRandomRunPath` (`e2e/e2e_test.go:128`) deliberately drops to a short `/tmp/ke-` prefix to stay under the limit. Any name growth here compounds that budget.
- So the fix is *more entropy, not more characters*: a process-wide atomic counter (monotonic, collision-free within a run) rendered in a compact base (e.g. base36) is the natural shape. Optionally seed it from a couple of `crypto/rand` bytes per process so a namespace left behind by a crashed prior run can't collide with a fresh `e-r-01`. Avoid folding in `t.Name()` or full nanosecond strings — that reintroduces the length problem.

## Acceptance criteria

- [ ] All five generators (`generateUniqueRealmName` / `Space` / `Stack` / `Cell` / `Container`) produce names that are unique across concurrent parallel tests within a single `go test` run (no `& 0xFF` truncation).
- [ ] Generated names stay short — realm/space/stack/cell names no longer than the current `e-r-XX` family by more than a few characters, and demonstrably within containerd namespace-label and `/opt/kukeon/data` metadata-path limits.
- [ ] Realm names remain valid containerd namespace labels (charset/format unchanged: `e-r-<id>` shape, lowercase, no `/` `:` etc.).
- [ ] The full e2e suite passes at the default parallelism (not just `-p 1`) across repeated runs without `lease does not exist` / shared-namespace teardown failures.
- [ ] (If a per-process random seed is added) the seed source and counter are documented inline so the cross-run-leftover rationale isn't lost.

## Notes

- Production code (`internal/ctr/image.go` lease handling, `internal/ctr/namespaces.go` `drainLeases`) is correct — it is being sabotaged by a sibling test sharing its namespace. No production change needed; this is test-infra only.
- Diagnosed from CI failure on `main` (post #697). Single reviewable PR.

eminwux:
## Summary
- The five e2e unique-name generators truncated `time.Now().UnixNano()` to its low byte (`& 0xFF`), giving only 256 possible values per kind. Under `t.Parallel()` on a shared host containerd, two tests routinely collided on the same `e-r-XX` realm name → same containerd namespace → one test's cleanup (`drainLeases` + `SynchronousDelete`) deleted a sibling's in-flight pull lease, surfacing as the spurious `lease does not exist: not found` flake.
- Replaced the per-kind truncation with a single process-wide monotonic `atomic.Uint64` counter shared by all five generators, so no two names allocated within one `go test` run can collide regardless of timing.
- Seeded the counter from a per-process `crypto/rand` 16-bit base (with a `UnixNano & 0xFFFF` fallback if the read fails) so a realm/namespace left behind by a crashed prior run won't collide with a fresh run's first-allocated name. Both the seed source and counter are documented inline.
- IDs are rendered in base36 (lowercase `0-9a-z`) to keep names valid containerd namespace labels and short — `e-r-<id>` is typically 3-4 id chars, only a couple longer than the old `e-r-XX`, staying within the namespace-label and `/opt/kukeon/data` metadata-path length budgets the harness already fights.

## Test plan
- [x] `make test` (test-root + test-kukebuild) — passes; the CI target excludes `/e2e`.
- [x] `go vet ./e2e/` and e2e test-binary compile (`go test -run xxxNONExxx ./e2e/`) — clean.
- [ ] Full e2e suite at default parallelism across repeated runs (AC bullet 4) — requires docker build + standalone containerd + daemon bring-up, not safely hostable here; deferred to CI/reviewer environment.

Closes #704

### DIFF
--- a/e2e/e2e_kuke_realm_test.go
+++ b/e2e/e2e_kuke_realm_test.go
@@ -23,7 +23,6 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/util/fs"
@@ -33,9 +32,7 @@ import (
 // generateUniqueRealmName generates a unique realm name for tests.
 func generateUniqueRealmName(t *testing.T) string {
 	t.Helper()
-	timestamp := time.Now().UnixNano()
-	hexID := fmt.Sprintf("%02x", timestamp&0xFF)
-	return fmt.Sprintf("e-r-%s", hexID)
+	return fmt.Sprintf("e-r-%s", uniqueNameID())
 }
 
 // verifyRealmMetadataExists verifies realm metadata file exists.
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -18,13 +18,17 @@ package e2e_test
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -37,6 +41,31 @@ const (
 	kuke = "kuke"
 	ctr  = "ctr"
 )
+
+var (
+	// nameSeed is a per-process random base for unique test-resource names. Seeding
+	// from crypto/rand ensures a realm/namespace left behind by a crashed prior run
+	// cannot collide with a fresh run's first-allocated name (e.g. e-r-1).
+	nameSeed = func() uint64 {
+		var b [2]byte
+		if _, err := rand.Read(b[:]); err != nil {
+			return uint64(time.Now().UnixNano() & 0xFFFF)
+		}
+		return uint64(binary.BigEndian.Uint16(b[:]))
+	}()
+	// nameCounter is a process-wide monotonic counter shared by every generator, so
+	// no two names allocated within a single `go test` run can collide regardless of
+	// how closely in time they are requested.
+	nameCounter atomic.Uint64
+)
+
+// uniqueNameID returns a short, collision-free identifier for test-resource names.
+// The seed+counter sum is rendered in base36 (lowercase 0-9a-z) to stay a valid
+// containerd namespace label while keeping names within the namespace-label and
+// /opt/kukeon/data metadata-path length limits the harness already fights.
+func uniqueNameID() string {
+	return strconv.FormatUint(nameSeed+nameCounter.Add(1), 36)
+}
 
 // runReturningBinary runs the provided binary with args, fails the test on non-zero exit or empty output.
 // If the binary file does not exist, the test is skipped.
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #707 Review — `test: prevent e2e parallel namespace collisions via process-wide name entropy`**

LGTM — replaces the 256-value `UnixNano & 0xFF` truncation in all five e2e name generators with a shared process-wide monotonic `atomic.Uint64`, which fully eliminates intra-run collisions; the crypto/rand 16-bit seed randomizes the start offset against leftover namespaces, and base36 keeps names valid containerd labels. Test-only, all five generators converted (no stragglers), commits verified.

Non-blocking — the one deferred test-plan item (full e2e at default parallelism, AC bullet 4) is host-gated on docker + standalone containerd + daemon bring-up; please run it in CI/your environment pre-merge to confirm the flake is gone.

PR #708, opened 2026-05-22
fix: resolve teardown CNI purge network name deterministically

### LINKED ISSUE #685: runner: root CNI purge skipped on teardown when space metadata is gone
## Context
On stop/kill/delete, the file-level CNI purge safety net is gated on `networkName != ""` (`internal/controller/runner/stop.go:278`, `kill.go:255`, `delete_cell.go:179`). `networkName` is resolved from `r.GetSpace(...)` then `getSpaceNetworkName(...)`, and the `GetSpace` error is swallowed (`stop.go:210-213`: `if spaceErr == nil`). If space metadata is gone/corrupt while a cell root still holds an IPAM reservation, `networkName` stays empty and the purge net is skipped. The earlier `detachRootContainerFromNetwork` still runs (it uses `cniConfigPath`, not `networkName`), so the leak only manifests when **both** `GetSpace` fails **and** the clean CNI DEL fails (e.g. invalid netns / conflist also gone) — hence the narrow trigger.

## Proposal
Resolve the purge target deterministically (container ID / cell spec) so the file-level CNI purge runs on teardown regardless of whether `GetSpace` succeeds.

## Acceptance criteria
- [ ] Teardown purges the root CNI/IPAM reservation even when space metadata is unavailable.
- [ ] Test: stop/kill/delete with the space metadata removed leaves no stranded `/var/lib/cni/networks/<net>/<ip>` file.

## Notes
Teardown-side residue of the same class fixed on the re-ADD side by #640 (bc8d244) and #650 (3904f26). Trigger verified by code-trace at `stop.go:278`; the conditional (DEL-also-fails) repro is worth confirming during implementation.

eminwux:
## Summary
- On stop/kill/delete, the file-level CNI purge safety net (`purgeCNIForContainer`) was gated on a non-empty `networkName`, but `networkName` was resolved via `GetSpace` → `getSpaceNetworkName` with the `GetSpace` error swallowed. When space metadata was gone or corrupt while a cell root still held an IPAM reservation, `networkName` stayed empty and the purge was skipped — leaking `/var/lib/cni/networks/<net>/<ip>`.
- The network name is a pure function of `(realm, space)` (`naming.BuildSpaceNetworkName` → `"<realm>-<space>"`), so the `GetSpace` round-trip is unnecessary for teardown. Added `buildRootCNINetworkName(realm, space)` (helpers.go) that resolves the name deterministically, and switched the three teardown sites (`stop.go`, `kill.go`, `delete_cell.go`) to it.
- Left `resolveRootCNINetworkName` (the re-ADD helper used by `start.go`/`recreate_cell.go`) untouched: that path intentionally returns `""` for a vanished space so attach is skipped, and `TestResolveRootCNINetworkName_EmptyWhenSpaceMissing` pins that contract. The teardown semantics are the opposite (purge regardless), hence a separate helper rather than a behavior change to the shared one.

## Scope note for the reviewer
AC #2 asks for a "stop/kill/delete with space metadata removed leaves no stranded `/var/lib/cni/networks/<net>/<ip>` file" test. `cni.CNINetworksDir` is a hardcoded `const` (`/var/lib/cni/networks`) — a root-owned system path — so a unit test cannot safely create/assert IPAM files there without touching real host state. The regression is precisely at the resolution boundary (empty `networkName` gated the purge out), so the added test pins it there: with no space metadata on disk, `buildRootCNINetworkName` returns the deterministic name where `resolveRootCNINetworkName` returns `""`. End-to-end on-disk verification belongs in the integration/e2e tier where the real CNI dir exists.

## Test plan
- [x] `make test` (full CI target: test-root + test-kukebuild) — pass
- [x] `GOWORK=off go build ./...` — pass
- [x] `GOWORK=off go vet ./internal/controller/runner/` — pass
- [x] `go test ./internal/controller/runner/ -run CNINetworkName -v` — both resolution-boundary tests pass
- [x] `pre-commit run --files <changed>` (golangci-lint skipped — go1.24 hook binary panics analyzing the go1.25 module via go.work, a known environmental issue; `go vet` covers the same surface)

Closes #685

### DIFF
--- a/internal/controller/runner/delete_cell.go
+++ b/internal/controller/runner/delete_cell.go
@@ -81,20 +81,10 @@ func (r *Exec) DeleteCell(cell intmodel.Cell) error {
 		cellID = internalCell.Metadata.Name
 	}
 
-	// Get space for network name
-	lookupSpace := intmodel.Space{
-		Metadata: intmodel.SpaceMetadata{
-			Name: cellSpaceName,
-		},
-		Spec: intmodel.SpaceSpec{
-			RealmName: internalCell.Spec.RealmName,
-		},
-	}
-	space, spaceErr := r.GetSpace(lookupSpace)
-	var networkName string
-	if spaceErr == nil {
-		networkName, _ = r.getSpaceNetworkName(space)
-	}
+	// Resolve the network name deterministically from realm+space so the
+	// post-delete CNI/IPAM purge below runs even when space metadata is gone or
+	// corrupt (issue #685). The name is a pure function of (realm, space).
+	networkName := r.buildRootCNINetworkName(internalCell.Spec.RealmName, cellSpaceName)
 
 	// Aggregate container-deletion failures rather than swallow them. Issue
 	// #371: a single warn-and-continue on a non-NotFound DeleteContainer
--- a/internal/controller/runner/helpers.go
+++ b/internal/controller/runner/helpers.go
@@ -473,6 +473,23 @@ func (r *Exec) resolveRootCNINetworkName(realmName, spaceName string) string {
 	return networkName
 }
 
+// buildRootCNINetworkName derives the CNI network name for a cell's space
+// deterministically from the realm and space names, without consulting space
+// metadata. Teardown paths (stop/kill/delete) use this for the post-delete
+// IPAM-file purge safety net (purgeCNIForContainer) so the purge runs even when
+// space metadata is gone or corrupt — the network name is a pure function of
+// (realm, space) per naming.BuildSpaceNetworkName, so the GetSpace round-trip
+// resolveRootCNINetworkName performs on the re-ADD side is unnecessary here and
+// its failure must not silently skip the purge (issue #685). Returns "" only
+// when realm/space are empty, which the teardown callers already guard against.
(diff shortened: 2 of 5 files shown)


eminwux:
**PR #708 Review — `fix: resolve teardown CNI purge network name deterministically`**

Correct fix for #685. Since `getSpaceNetworkName` ultimately just calls `naming.BuildSpaceNetworkName(realm, space)` with no per-space override, the new deterministic `buildRootCNINetworkName` yields an identical name to the old GetSpace path — but without gating the IPAM purge on a round-trip that fails exactly when space metadata is gone. All three teardown sites switched, the re-ADD helper left intact, and the regression test pins the resolution boundary. No secrets, no dead symbols, commits verified, CI green.

One non-blocking item:

- **Stale docstring** `internal/controller/runner/helpers.go:454-459`: `resolveRootCNINetworkName`'s comment still says it's "for use in the post-delete IPAM-file purge safety net... that stop.go and kill.go run" — but after this PR that job is `buildRootCNINetworkName`'s, and `resolveRootCNINetworkName` is now exclusively the re-ADD helper for `start.go`/`recreate_cell.go`. Retarget the comment so a future debugger isn't pointed at the attach helper for a teardown leak.

LGTM — the docstring is non-blocking.

PR #702, opened 2026-05-22
feat: add kind CellBlueprint with daemon storage and kuke run -b

eminwux:
## Summary

Phase 4a-i of the L1↔L2 templating epic (#423). Introduces a **new** `kind: CellBlueprint` — a daemon-stored, scopable, parametrized cell template — alongside the existing client-side `CellProfile` (this is an addition, not a rename). A blueprint is written to daemon storage with `kuke apply` and instantiated with `kuke run -b`.

- **Schema** (`pkg/api/model/v1beta1/cellblueprint.go`): `CellBlueprintDoc` registered as `KindCellBlueprint` in consts. Scopable at realm/space/stack only (never cell — a template scoped to one cell is nonsensical); realm always required. Declares two fill channels: scalar `${KEY}` parameters (reusing `CellProfileParameter`) and **structural slot declarations** — `containers[].repos[]` (the existing `ContainerRepo` shape, but url is *not* required: a urless repo is a slot a CellConfig fills) and `containers[].secrets[]` (`BlueprintSecretSlot`: the consumption side — env/file mode — only). Slot declarations only; the Config-side fill machinery lands in #624.
- **Daemon storage** (`internal/controller/runner/blueprint.go`): persists to `<run-path>/<scope>/blueprints/<name>`, root-owned and **world-readable** (0o755 dir / 0o644 file) — contrasted with the root-only 0o700/0o600 secrets path, since a blueprint carries only template references, no credential bytes. Reuses the #619/#660 atomic temp-file+rename primitive.
- **`kuke apply -f blueprint.yaml`** writes to daemon storage over RPC (`ReconcileBlueprint`, with the same scope-reachability gate as `ReconcileSecret`).
- **`kuke run -b <blueprint> --param K=V`** resolves the blueprint from daemon storage via a new `GetBlueprint` RPC, substitutes scalar parameters, and materializes a fresh `<prefix>-<6hex>` cell carrying a `kukeon.io/blueprint=<name>` back-reference label. `-b` joins the `-f`/`-p` mutual-exclusivity + one-required flag group.
- **`kuke run -p`** now emits a one-line deprecation notice pointing at `-b`/`-c`; CellProfile/`loadFromProfile` behavior otherwise unchanged.

### Design decisions (please push back)

[...]

Issue #620, opened 2026-05-19
blueprint: introduce CellBlueprint + store + kuke run -b (phase 4.1.1)


eminwux:
## Context

Phase 4a-i of #423 (first slice of phase 4). Depends on #619 (daemon-stored-resource storage primitive).

`CellBlueprint` is a **new daemon-stored kind**, introduced *alongside* today's client-side `CellProfile` — **not a rename**. `kuke run -p` + `kind: CellProfile` + the client-side `loadFromProfile` path stay completely untouched by this issue. The new `kuke run -b` verb runs a daemon-stored Blueprint. This makes phase 4 additive: operators migrate from `-p` to `-b`/`-c` at their own pace, and the legacy `-p` path is removed only in the final slice (#626, phase 4d).

Today `CellProfile` is client-side only — the daemon never sees it; `kuke run -p` materialises a `CellDoc` locally and proceeds along the `-f` path (`cmd/kuke/run/run.go:386-419`, `loadFromProfile`). Templates can't be applied, listed, deleted, or scoped server-side. `CellBlueprint` fixes that without disturbing the existing path.

## Proposal

Introduce `kind: CellBlueprint` — a daemon-stored, scopable parametrised template, peer to other daemon-stored kinds.

```yaml
apiVersion: v1beta1
kind: CellBlueprint
metadata:
  name: dev
  realm: kuke-system            # scope: realm | space | stack (not cell)
spec:
  parameters:
    - { name: PROJECT_DIR, required: true }
  cell:
    containers:
      - id: work
        image: ${REGISTRY}/claude:latest
        repos:
          - { name: project, target: /home/claude/${PROJECT_DIR}, required: true }
        secrets:
          - { name: anthropic-token, mode: env, envName: ANTHROPIC_AUTH_TOKEN, required: true }
```

Storage under `/opt/kukeon/<scope>/blueprints/<name>` — root-owned, world-readable (no secret data; references only). Reuses the daemon-stored-resource primitive landed in #619. The schema reuses today's `CellProfileParameter` shape (`pkg/api/model/v1beta1/cellprofile.go:64`) for `parameters[]`.

### Two fill channels (declarations here; Config-side fill in #624)

1. **Scalar parameters** — today's `${KEY}` substitution. Declared as `parameters: [{name, default?, required}]`.
[...]

PR #716, opened 2026-05-22
chore(cellprofile): remove dead exported CloneNode and ParamRefRE

### LINKED ISSUE #705: cellprofile: remove dead exported CloneNode and ParamRefRE
Surfaced during review of #702.

`internal/cellprofile/params.go` exports two functions that have zero callers anywhere in the tree:

- `CloneNode` (`internal/cellprofile/params.go:336`)
- `ParamRefRE` (`internal/cellprofile/params.go:344`)

Both were added in #702 with doc comments stating they are "Exported for the CellBlueprint materialize path (issue #620)" / "Exposed so the blueprint resolver can scan a template body," but the blueprint resolver (`internal/cellblueprint/cellblueprint.go`) uses only `cellprofile.SubstituteScalars` — it neither clones nodes nor pre-scans the body for `${KEY}` references.

Suggested fix: delete both exported wrappers (the underlying private `cloneNode` / `paramRefRE` stay). Non-blocking — internal package, not consumer-facing API.

eminwux:
## Summary
- `CloneNode` and `ParamRefRE` in `internal/cellprofile/params.go` were added in #702 with doc comments claiming they serve the CellBlueprint materialize path (#620), but the blueprint resolver (`internal/cellblueprint/cellblueprint.go`) uses only `cellprofile.SubstituteScalars` — it never clones nodes nor pre-scans a body for `${KEY}` refs.
- A tree-wide grep confirms zero callers of either exported wrapper anywhere (incl. tests, e2e, docs). The underlying private `cloneNode`/`paramRefRE` stay — both are still live within the package.
- Note: the original stall comment on #705 predated #702's merge; #702 has since merged (`3ebb9cd`) so the symbols now exist on `main` and the deletion is a real diff. The `stalled` label has been cleared; issue carries only `followup-nits`.

## Test plan
- [x] `make test` (test-root + test-kukebuild) — pass
- [x] `go vet ./internal/cellprofile/` — clean
- [x] tree-wide grep for `CloneNode`/`ParamRefRE` callers — none outside the deleted definitions

Closes #705

### DIFF
--- a/internal/cellprofile/params.go
+++ b/internal/cellprofile/params.go
@@ -331,20 +331,6 @@ func SubstituteScalars(node *yaml.Node, values map[string]string) {
 	substituteScalars(node, values)
 }
 
-// CloneNode returns a deep copy of n so substitutions don't mutate the input.
-// Exported for the CellBlueprint materialize path (issue #620).
-func CloneNode(n *yaml.Node) *yaml.Node {
-	return cloneNode(n)
-}
-
-// ParamRefRE returns the compiled `${KEY}` reference matcher shared by the
-// profile and blueprint substitution paths. Exposed so the blueprint resolver
-// can scan a template body for the parameter shape without re-deriving the
-// regex. Issue #620.
-func ParamRefRE() *regexp.Regexp {
-	return paramRefRE
-}
-
 // ParseParamArgs validates a slice of `KEY=VALUE` strings (from --param) and
 // returns the equivalent map. Empty values are allowed (`KEY=` → "");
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #716 Review — `chore(cellprofile): remove dead exported CloneNode and ParamRefRE`**

LGTM — deletion-only; both exported wrappers have zero tree-wide callers, the private `cloneNode`/`paramRefRE` they wrapped stay live in the package, and the `regexp` import remains used, so the build holds.

PR #721, opened 2026-05-22
fix: guard ctr reader sites against cClient reconnect reassignment

### LINKED ISSUE #709: ctr: guard remaining c.cClient reads against the reconnect-path reassignment
Surfaced during review of #695 (which serialized the `Connect()`/`Close()` write of `c.cClient` behind `cClientMu`).

That PR correctly closed the #684 first-use construction race, but the fix only guards the *writer*. The reconnect path inside `Connect()` still does `c.cClient = nil` (via `closeLocked()`) followed by `c.cClient, err = containerd.New(...)` under the lock, while the many *reader* sites read `c.cClient` with no lock:

- `internal/ctr/cache.go:53`
- `internal/ctr/container.go:331,388,411,433`
- `internal/ctr/image.go:95,105,150,169,197,224,251`
- `internal/ctr/namespaces.go:52,77,84,120,143,210,217`
- `internal/ctr/spec.go:123`

A reconnect (triggered when `verifyConnection()` fails on a broken connection) concurrent with an in-flight operation could read a nil or torn pointer. This is pre-existing and narrow (reconnect is rare), hence non-blocking for #695, but it is a genuine latent data race that `go test -race` would flag on a cgo-enabled host.

Suggested fix: take `cClientMu` (read lock if converted to `sync.RWMutex`) around the reader sites, or snapshot `c.cClient` into a local under the lock at the top of each operation.

eminwux:
## Summary
- #695 serialized the *writer* of `c.cClient` behind `cClientMu` (closing the #684 first-use race), but left every reader site dereferencing `c.cClient` with no lock. A `Connect()` reconnect (`closeLocked()` sets `c.cClient = nil`, then `containerd.New(...)` reassigns it, both under the lock) concurrent with an in-flight operation could read a nil or torn pointer.
- Adds a `conn()` helper on `*client` that snapshots `c.cClient` under `cClientMu` and returns it. The lock is held only for the pointer read, so the reader path is never serialized against an in-flight RPC, and a concurrent reconnect that swaps the pointer cannot tear an operation already dereferencing the prior one.
- Converts every reader site enumerated in #709 to take a snapshot via `conn()`: `cache.go`, `container.go` (×4), `image.go` (×7), `namespaces.go` (×7, including the two `if c.cClient == nil { Connect() }` ensure-connected sites), and `spec.go`.

### Design call (the one #709 left open)
The issue offered "snapshot under the lock" *or* "convert `cClientMu` to `sync.RWMutex` and hold RLock for the whole operation". I chose the **snapshot-with-plain-Mutex** approach because it is strictly lower-contention: holding an RLock for an operation's full duration would block a reconnect (writer) until every in-flight reader returned, whereas a snapshot holds the lock only for the pointer copy and lets the in-flight op proceed on its own stable reference. No field-type change, minimal lock hold. The reviewer can push back if RWMutex is preferred for other reasons.

`conn()` returns nil if the client has never connected — same precondition the prior direct `c.cClient` reads carried, so nil-safety is unchanged at sites that don't gate on `Connect()` first.

## Test plan
- [x] `GOWORK=off go vet ./internal/ctr/...`
- [x] `make test` (full CI target: `test-root` + `test-kukebuild`) — green, `internal/ctr` passes
- [ ] `go test -race` on the reconnect/reader concurrency — not run: the dev host has no cgo toolchain (`which gcc` empty) and `-race` requires it. `make test` (the named CI gate per `.github/workflows/test.yaml`) does **not** run `-race`, so this is a deferred verification, not a CI-target gap. Note also there is no existing concurrency test exercising `Connect()` against the readers; the fix is correct-by-construction (snapshot taken under the same mutex that guards the writer).

Closes #709

### DIFF
--- a/internal/ctr/cache.go
+++ b/internal/ctr/cache.go
@@ -50,7 +50,7 @@ func (c *client) loadContainer(namespace, id string) (containerd.Container, erro
 	c.containersMu.RUnlock()
 
 	nsCtx := c.namespaceCtx(namespace)
-	container, err := c.cClient.LoadContainer(nsCtx, id)
+	container, err := c.conn().LoadContainer(nsCtx, id)
 	if err != nil {
 		// Only wrap "not found" errors with ErrContainerNotFound
 		// Other errors (connection failures, permission errors, etc.) should be returned as-is
--- a/internal/ctr/client.go
+++ b/internal/ctr/client.go
@@ -224,6 +224,20 @@ func (c *client) Close() error {
 	return c.closeLocked()
 }
 
+// conn returns the current containerd client snapshotted under cClientMu, so a
+// reader observes a consistent pointer even while Connect()'s reconnect path is
+// concurrently swapping c.cClient (issue #709 — #695 guarded only the writer).
+// The lock is held just for the pointer read; the snapshot is then used for the
+// whole operation, so the reader path is never serialized against an in-flight
+// RPC and a concurrent reconnect cannot tear an operation already dereferencing
+// the prior pointer. Returns nil if the client has not connected yet — callers
+// that have not already gated on Connect() must nil-check.
+func (c *client) conn() *containerd.Client {
+	c.cClientMu.Lock()
+	defer c.cClientMu.Unlock()
+	return c.cClient
+}
+
 // closeLocked tears down c.cClient. The caller must hold c.cClientMu — it is
 // invoked both by the public Close() (which takes the lock) and by Connect()'s
 // reconnect path (which already holds the lock), so the teardown stays guarded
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #721 Review — `fix: guard ctr reader sites against cClient reconnect reassignment`**

LGTM — snapshots `c.cClient` under `cClientMu` via `conn()` and routes every lock-free reader site through it, closing the #709 reconnect-vs-read race. Verified the systemic conversion is complete: `git grep c.cClient` on the head leaves only the writer path (`Connect`/`closeLocked`/`verifyConnection`), all lock-held — `verifyConnection` is called solely from `Connect()` under the lock, so leaving it direct is correct (converting would deadlock the non-reentrant mutex). Snapshot-per-operation and ensure-connected (`conn()`→nil-check→`Connect()`→re-`conn()`) sites are sound. Commits verified, CI green.

Deferred for you pre-merge: `go test -race` on the reconnect/reader path — host-gated (no cgo toolchain on the agent host), not run here. Non-blocking per the host-gated carve-out.

PR #722, opened 2026-05-22
feat: serialize cell lifecycle ops with a per-cell lifecycle lock

### LINKED ISSUE #714: runner: add per-cell lifecycle lock spanning containerd+CNI side effects
## Context

The runner `Exec` struct (`internal/controller/runner/runner.go:180-211`) has no per-cell lock — it carries only `ctrClientOnce sync.Once` (runner.go:192) plus the per-instance mutex inside `*cni.SubnetAllocator` (which arbitrates space-subnet allocation, not cell lifecycle).

The daemon runs the reconcile ticker on its own goroutine (`go s.runReconcileLoop`, `internal/daemon/server.go:167`; loop body `server.go:170-208`) and serves each gRPC call on its own goroutine. The server's `mu sync.Mutex` (`server.go:81`) guards only listener/closed state (used at server.go:132-134, 246-248, 265-273) — not cell mutations.

Therefore `ReconcileCell` and the gRPC `StartCell`/`StopCell`/`KillCell`/`DeleteCell`/`RecreateCell` handlers do **not** serialize against each other on the same cell. They can run containerd delete + CNI teardown concurrently against the same root container ID and IPAM file, which can scrub a freshly-allocated reservation, double-delete a container, or diverge metadata from containerd state.

The metadata flock + generation-token CAS protect only the on-disk document, not the containerd/CNI mutations. #686 (cgroup cache mutex) and #695 (ctrClient `sync.Once`) are point mutexes on internal caches, **not** a lifecycle lock — they do not close this window.

## Proposal

Add an in-process keyed mutex (keyed by realm/space/stack/cell) on `Exec`, held across the containerd+CNI side-effect span of every lifecycle op including `ReconcileCell`.

## Acceptance criteria
- [ ] A concurrent `StartCell` + `ReconcileCell` on the same cell cannot interleave teardown; a test drives both and asserts a single teardown.
- [ ] Lock is keyed per-cell (no global serialization of unrelated cells).

## Notes
- This is the structural root cause behind the fix-chasing-fix churn in this subsystem; it reduces the severity of the residual leak paths tracked in the sibling CNI/IPAM issues filed alongside this one.
- Found via subsystem audit at HEAD ed3c723; verified against origin/main 6f1a7cc before filing (Exec struct, server.go mutex scope, reconcile goroutine all confirmed). Re-verify line numbers before acting — the repo moves fast.

eminwux:
## Summary
- Adds a per-cell keyed mutex (`cellLockManager`, keyed by realm/space/stack/cell) on `Exec` and holds it across the containerd+CNI side-effect span of every cell lifecycle op: `StartCell`, `StopCell`, `KillCell`, `DeleteCell`, `RecreateCell`, `CreateCell`, `PurgeCell`, and `ReconcileCell`. This closes the window (issue #714) where a gRPC handler and the reconcile-loop tick could run teardown concurrently against the same root container ID and IPAM file. The metadata flock + generation-token CAS guard only the on-disk document, not the containerd/CNI mutations — this lock covers the mutation span.
- Keys are independent, so unrelated cells never serialize against one another (per-cell, not global).
- The lock is non-reentrant, but several lifecycle ops nest others (`RecreateCell`→`StopCell`/`StartCell`, `StartCell` failure-cleanup→`KillCell`, `ReconcileCell` wind-down/auto-delete→`KillCell`/`DeleteCell`, `PurgeCell`→`DeleteCell`). To avoid self-deadlock, each nested-callee public method (`Start`/`Stop`/`Kill`/`Delete`) is split into a thin locking wrapper plus an unexported `*Locked` body; the outermost op acquires the lock once and nested calls route to the `*Locked` variants. The four top-level-only ops (`Create`/`Recreate`/`Reconcile`/`Purge`) acquire the lock at entry and call the `*Locked` variants internally.
- The lock manager is ref-counted: an entry is dropped from the map once its last holder releases, so the map cannot grow unbounded over the daemon's lifetime. It is lazily initialized via `sync.Once` so `*Exec` values built directly in tests (not via `NewRunner`) share the same serialization without extra fixture wiring.

## Implementation notes (for reviewer)
- **Scope beyond the AC's named handlers:** the AC enumerates `StartCell`/`StopCell`/`KillCell`/`DeleteCell`/`RecreateCell` + `ReconcileCell`. I also locked `CreateCell` and `PurgeCell` because both touch containerd/CNI on the same cell and would otherwise race the locked ops — consistent with the AC's "every lifecycle op" intent. Push back if you'd rather scope them out.
- The negative test was verified by hand: with the `StartCell` lock disabled, `TestCellLifecycleLock_StartCellWaitsForReconcile` fails (StartCell returns while ReconcileCell holds the lock); with it restored, it passes.

## Test plan
- [x] `make test` (full CI target; 93 packages green, GOWORK=off)
- [x] `go build ./...` and `go vet ./internal/controller/runner/...`
- [x] New unit tests on the keyed mutex (`TestCellLockManager_*`, `TestCellLockKey_*`): same-key mutual exclusion, different-key concurrency, idle-entry reclaim, identity-folding.
- [x] New concurrency tests on the wired methods: `TestCellLifecycleLock_StartCellWaitsForReconcile` (same-cell StartCell + ReconcileCell cannot interleave — AC #1) and `TestCellLifecycleLock_DifferentCellsRunConcurrently` (distinct cells run concurrently — AC #2).
- [x] Negative verification: same-cell test fails with the lock removed, passes with it restored.
- [x] `make dev-init` smoke (nested mode): full re-bootstrap green, attach smoke clean, parent attach plumbing intact, and the daemon-parity guard (`kuke get realms` vs `--no-daemon`) shows identical Ready output for both realms.

Closes #714

### DIFF
--- a/internal/controller/runner/cell_lock.go
+++ b/internal/controller/runner/cell_lock.go
@@ -1,0 +1,103 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package runner
+
+import (
+	"strings"
+	"sync"
+
+	intmodel "github.com/eminwux/kukeon/internal/modelhub"
+)
+
+// cellLockManager hands out a per-cell mutex keyed by the cell's
+// realm/space/stack/name identity. It serializes the containerd+CNI
+// side-effect span of the lifecycle ops (Start/Stop/Kill/Delete/Recreate/
+// Create/Purge/Reconcile) against each other on the same cell, closing the
+// window where a gRPC handler and the reconcile-loop tick could run teardown
+// concurrently against the same root container ID and IPAM file (issue #714).
+//
+// The metadata flock + generation-token CAS guard only the on-disk document,
+// not the containerd/CNI mutations; this lock covers the mutation span. Keys
+// are independent, so unrelated cells never serialize against one another.
+type cellLockManager struct {
+	mu    sync.Mutex
+	locks map[string]*cellLockEntry
+}
+
+// cellLockEntry is the per-key mutex plus a reference count of goroutines
+// currently holding or waiting on it. The entry is dropped from the map once
+// the count returns to zero so the map cannot grow unbounded over the daemon's
+// lifetime.
+type cellLockEntry struct {
+	mu      sync.Mutex
+	waiters int
+}
+
+func newCellLockManager() *cellLockManager {
+	return &cellLockManager{locks: make(map[string]*cellLockEntry)}
+}
+
+// lock acquires the per-key mutex and returns a release func. The release func
+// unlocks the mutex and drops the entry when no other goroutine is waiting on
+// it. Call the returned func exactly once (typically via defer).
+func (m *cellLockManager) lock(key string) func() {
+	m.mu.Lock()
+	e := m.locks[key]
+	if e == nil {
+		e = &cellLockEntry{}
+		m.locks[key] = e
+	}
+	e.waiters++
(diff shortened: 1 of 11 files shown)


eminwux:
**PR #722 Review — `feat: serialize cell lifecycle ops with a per-cell lifecycle lock`**

LGTM — per-cell refcounted keyed mutex closes the #714 teardown race; deadlock-safe because every public wrapper acquires once and all internal/nested calls (incl. `markCellFailed`→`killCellLocked`, only reachable under the lock) route to the `*Locked` bodies, with no remaining internal callers of the public wrappers.

PR #724, opened 2026-05-22
fix: issue best-effort CNI DEL when root task is already dead

### LINKED ISSUE #715: runner: detachRootContainerFromNetwork skips CNI DEL when root task is dead
## Context

`detachRootContainerFromNetwork` (`internal/controller/runner/helpers.go:497-557`) issues the in-netns CNI DEL only inside nested guards: container exists (`GetContainer` err == nil) → task exists (`Task()` err == nil) → `rootPID > 0` (gate at helpers.go:505-507). If the root task is already dead (OOM, crash, host reboot left the record but no task), the whole block is a silent no-op and the clean CNI DEL release never runs.

#708 made the post-delete IP-file purge run deterministically (`stop.go:203`, `kill.go:184` now resolve the network name via `buildRootCNINetworkName`), which catches the leaked IPAM *file*. But the proper in-netns CNI DEL release still never executes on the dead-task path — only the file-scrub safety net does.

## Proposal
- On a dead/absent task, fall back to a best-effort CNI DEL (empty/derived netns) before relying on the IP-file purge; **or** document that the file purge is the authoritative release and remove the dead in-netns branch so the contract is explicit.

## Acceptance criteria
- [ ] A cell whose root task has exited still has its CNI/IPAM reservation fully released on stop/kill (test against a temp host-local store).

## Notes
- Residual of the same class #708 addressed (#708 fixed the file purge; this is the clean-DEL gap that remains).
- Concurrency exposure of this path is tracked by the per-cell lifecycle-lock issue filed alongside this one.
- Found via subsystem audit at HEAD ed3c723; verified against origin/main 6f1a7cc (the PID>0 gate is confirmed at helpers.go:505-507). Re-verify line numbers before acting.

eminwux:
## Summary
- `detachRootContainerFromNetwork` (`internal/controller/runner/helpers.go`) previously nested the in-netns CNI DEL behind container-exists → task-exists → `rootPID > 0` guards, so when the root task was already dead (OOM, crash, host reboot left the container record but no task) the whole block was a silent no-op and the clean plugin-chain release never ran — only the post-delete IP-file purge safety net did (the clean-DEL residual #708 left behind).
- Restructured so the CNI DEL is issued **unconditionally**, with an empty netns on the dead/absent-task path. `host-local` IPAM DEL keys off the container ID and tolerates a missing netns, and chained plugins are required by the CNI spec to complete DEL without error when their resources are already gone — so a best-effort empty-netns DEL releases the reservation rather than leaking it.
- Extracted the netns resolution into `rootContainerNetnsPath`, which returns `""` for the dead/absent-task cases — a testable seam.

**Design choice (issue offered two options):** took Option A (best-effort CNI DEL on the dead path) over Option B (document the file purge as authoritative and remove the dead branch). Option A is the only one that satisfies the AC's "fully released" requirement: the file-scrub safety net only removes the IPAM allocation file, whereas the plugin-chain DEL also releases any chained-plugin state (portmap/firewall/etc.) and is the proper release path. Option B would have been a behavior reduction.

## Test plan
- [x] `make test` (full CI target: `test-root` + `test-kukebuild`) — pass.
- [x] `go vet ./internal/controller/runner/ ./internal/cni/` — pass.
- [x] New unit test `TestRootContainerNetnsPath_EmptyOnDeadOrAbsentTask` — covers both the absent-record (`GetContainer` error) and dead-task (`Task()` reports gone) branches resolving to `""`.
- [x] New test `TestDetachRootContainerFromNetwork_DeadTaskReleasesReservation` — **the AC against a temp host-local store**: seeds a reservation in a temp store dir, runs detach on the dead-task path (netns `""`), and asserts a self-provided fake `host-local`-style CNI plugin releases the reservation on DEL. Uses a fake plugin script (not real `/opt/cni/bin`) so the test is CI-safe and deterministic, matching the precedent in `internal/cni/container_test.go` that loaded-config plugin runs are otherwise integration-only. Confirmed it fails without the fix (DEL is never issued on the dead path).
- [x] `make dev-init` smoke test — daemon-parity check identical for `kuke get realms` and `--no-daemon`; the attach/purge smoke (which exercises CNI teardown/release paths) completed cleanly.

Closes #715

### DIFF
--- a/internal/controller/runner/detach_network_test.go
+++ b/internal/controller/runner/detach_network_test.go
@@ -1,0 +1,164 @@
+//go:build !integration
+
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//nolint:testpackage // exercises *Exec.detachRootContainerFromNetwork and its netns helper (unexported)
+package runner
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"testing"
+
+	containerd "github.com/containerd/containerd/v2/client"
+	"github.com/containerd/containerd/v2/pkg/cio"
+	"github.com/eminwux/kukeon/internal/cni"
+	"github.com/eminwux/kukeon/internal/errdefs"
+)
+
+// deadTaskContainer is a near-empty containerd.Container whose Task() reports
+// the post-reboot / OOM signature: the container record survives but its task
+// is gone. It models the exact dead-task path issue #715 cares about — distinct
+// from an absent container record (GetContainer error), which deleteCellFakeClient
+// already produces. Only Task() carries behavior; every other method returns a
+// zero value because rootContainerNetnsPath never calls them.
+type deadTaskContainer struct {
+	containerd.Container
+}
+
+func (c deadTaskContainer) Task(context.Context, cio.Attach) (containerd.Task, error) {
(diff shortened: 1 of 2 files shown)


eminwux:
Re-running failed CI jobs once (`make e2e` in the Test workflow: https://github.com/eminwux/kukeon/actions/runs/26301925951). Will revisit when they finish; if they fail again the PR will be marked `ci-stalled`.

eminwux:
**PR #724 Review — `fix: issue best-effort CNI DEL when root task is already dead`**

LGTM — issues an unconditional best-effort CNI DEL (empty netns on the dead/absent-task path) so the plugin-chain release runs instead of relying solely on the file-purge safety net; live path is unchanged, `cniConf` is non-nil at all four call sites, and the new test fails without the fix.

PR #728, opened 2026-05-22
test: add direct tests for purgeCNIForContainer and containsExactContainerID

### LINKED ISSUE #719: runner: add direct tests for purgeCNIForContainer and containsExactContainerID
## Context

`purgeCNIForContainer` (`internal/controller/runner/helpers.go:41`) and `containsExactContainerID` (`helpers.go:165`) are the safety-net functions every CNI/IPAM leak fix in this subsystem relies on, but they have no direct tests:

- `containsExactContainerID` has **zero** `*_test.go` references.
- `purgeCNIForContainer` is named only in test comments (`helpers_test.go:103,116`) and dereferenced indirectly via DeleteCell tests where it deliberately no-ops (`delete_cell_test.go:200` — empty network/cache dir).

## Proposal
- Table test for `purgeCNIForContainer` against a temp dir laid out like `/var/lib/cni/networks/<net>/` with a realistic IP file (`<id>\neth0`), a `last_reserved_ip.0`, and a same-prefix sibling; assert only the correct file is removed.
- Direct tests for `containsExactContainerID` covering both host-local content formats and the shared-prefix false-positive case.

## Acceptance criteria
- [ ] Both functions have direct unit tests; the prefix false-positive case is asserted.

## Notes
- The original brief listed this as "Blocked by ISSUE 1 (build broken)". That blocker is **dropped**: the build break was an artifact of building through the `go.work` workspace; `GOWORK=off go build/vet ./...` (what CI and make use) both pass, so these tests can run today.
- Found via subsystem audit at HEAD ed3c723; verified against origin/main 6f1a7cc (test absence + helper line numbers confirmed). Re-verify line numbers before acting.

eminwux:
## Summary
- `containsExactContainerID` (`internal/controller/runner/helpers.go:166`) and the host-local IPAM scan in `purgeCNIForContainer` (`helpers.go:42`) are the safety net every CNI/IPAM leak fix in this subsystem relies on, yet had no direct tests — `containsExactContainerID` had zero `*_test.go` references and `purgeCNIForContainer` was only exercised indirectly via DeleteCell tests where it deliberately no-ops.
- Added `TestContainsExactContainerID`: a table test asserting exact-trim match, host-local two-line (`<id>\neth0`) match, structured-JSON match, and the load-bearing **shared-prefix false-positive** cases (sibling `cid-1000` and suffix `cid-100extra` must not match `cid-100`), plus the `last_reserved_ip` content that holds an IP not a container ID.
- Added `TestPurgeCNIForContainer_RemovesOnlyMatchingIPAMFile`: lays out a temp `networks/<net>/` dir with a matching IP file (`<id>\neth0`), a same-prefix sibling, and a `last_reserved_ip.0`, then asserts only the matching file is removed — the sibling and bookkeeping file survive.

## Non-obvious implementation call
- The IPAM scan keys off `cni.CNINetworksDir`, which was a `const` hardcoded to `/var/lib/cni/networks` with no test seam. I converted it to a package-level `var` (default value unchanged) so the test can redirect the scan at a temp dir; production never reassigns it. All four call sites read it at runtime via `filepath.Join`, so there are no const-expression dependents and behavior is identical. Flagged here so the reviewer can push back if a different seam is preferred.

## Test plan
- [x] `make test` (CI target: `test-root` + `test-kukebuild`) — exit 0, 93 `ok`, no failures
- [x] `GOWORK=off go build ./...` — clean
- [x] `GOWORK=off go vet ./internal/controller/runner/ ./internal/cni/` — clean
- [x] `pre-commit run --files <changed>` — go fmt, go-mod-tidy, addlicense, etc. all pass
- [ ] `golangci-lint` pre-commit hook — panics with `package requires newer Go version go1.25 (application built with go1.24)`; the dev host's golangci-lint binary is built against go1.24 and can't analyze the go1.25 module via go.work. Environmental, not a finding; CI runs a compatible toolchain.

Closes #719

### DIFF
--- a/internal/cni/types.go
+++ b/internal/cni/types.go
@@ -104,4 +104,6 @@ const (
 
 // CNINetworksDir is the standard directory where CNI stores network state and IPAM allocations.
 // This is the default location used by CNI plugins for host-local IPAM.
-const CNINetworksDir = "/var/lib/cni/networks"
+// It is a var rather than a const so tests can redirect the host-local IPAM
+// scan in purgeCNIForContainer at a temp dir; production never reassigns it.
+var CNINetworksDir = "/var/lib/cni/networks"
--- a/internal/controller/runner/helpers_test.go
+++ b/internal/controller/runner/helpers_test.go
@@ -20,11 +20,18 @@
 package runner
 
 import (
+	"context"
 	"errors"
 	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"sort"
 	"testing"
 	"time"
 
+	"github.com/eminwux/kukeon/internal/cni"
 	"github.com/eminwux/kukeon/internal/errdefs"
 	intmodel "github.com/eminwux/kukeon/internal/modelhub"
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #728 Review — `test: add direct tests for purgeCNIForContainer and containsExactContainerID`**

Solid coverage for the IPAM-purge safety net. Both tests assert against the real production logic: I traced `containsExactContainerID`'s regex `(^|[^\w])` + escaped-ID + `([^\w]|$)` against all seven cases (shared-prefix `cid-1000`/`cid-100extra` correctly miss, JSON/host-local/exact correctly hit), and `purgeCNIForContainer`'s ReadDir scan removes only the matching IP file while the sibling and `last_reserved_ip.0` survive. No leaks, all commits verified.

One non-blocking note (you flagged the seam yourself):

- **Seam choice** `internal/cni/types.go:109`: the `const`→`var` `CNINetworksDir` is fine and behavior-preserving, but its three sibling CNI dirs (`CniConfigDir`/`CniBinDir`/`CniCacheDir`) already route through the `cni.Conf` DI struct — threading a `NetworksDir` field there instead would unify the contract and avoid the package-global being a latent hazard if a future test adds `t.Parallel()`. Pre-existing split (the const was already package-level), so not introduced here; leaving as-is is acceptable.

The unchecked `golangci-lint` item is a documented host toolchain mismatch (go1.25 module vs go1.24 binary) and CI's `make test`/`make e2e` are green — non-blocking.

LGTM — test-only PR, minimal documented seam, correct assertions.

PR #729, opened 2026-05-22
fix: fail-fast in release-build loops to stop masking build errors

### LINKED ISSUE #713: build: release-build for-loop masks failed go build, poisoning the kukebuild cache
## Repro

On a host with low disk headroom (or any transient `go build` failure), run the kukebuild image build that the dev-init smoke drives:

```bash
make dev-init        # → kuke build -t kukeon-local:dev --realm kuke-system .
```

The Dockerfile `builder` stage runs `RUN make release-build KUKEON_VERSION=v0.0.0-dev OS=linux ARCHS=amd64`. The recipe (`Makefile:88`) is a nested `for` loop whose commands are joined with `;`:

```make
release-build:
	for OS in $(OS); do \
		for ARCH in $(ARCHS); do \
			... go build -a ... -o kuke-$$OS-$$ARCH ./cmd; \
			ln -sf kuke-$$OS-$$ARCH kukeond-$$OS-$$ARCH; \
			... go build -a ... -o kuketty-$$OS-$$ARCH ./cmd/kuketty; \
		done \
	done
```

## Observed

When the first `go build ./cmd` fails (here: `compile: writing output: ... no space left on device`), the `;` separators let the loop continue to `ln -sf` and the next `go build`, and the recipe's exit status is that of the **last** command in the sequence — not the failed one. The RUN step exits 0 ("`#16 ... DONE 28.1s`") despite never producing `kuke-linux-amd64`. BuildKit then caches that layer as successful, and the failure surfaces several steps later as a misleading:

```
[stage-2 4/9] COPY --from=builder /workspace/kuke-linux-amd64 /bin/kuke:
ERROR: ... "/workspace/kuke-linux-amd64": not found
kukebuild: build "kukeon-local:dev": failed to solve: failed to compute cache key: ... not found
```

The poisoned layer persists in the per-namespace cache (`/var/lib/kukebuild/<ns>`), so subsequent builds re-hit the cached "successful" RUN and keep failing at COPY until the cache is manually busted — turning a one-line disk/build error into a multi-cycle debugging detour.

## Root cause

`release-build` (`Makefile:88`) does not fail-fast. Make runs the whole `for` loop in one `/bin/sh -c`, and the `;` chaining means a non-final `go build` failure neither stops the loop nor fails the recipe. A masked build failure is then frozen into the BuildKit cache as a success.

## Suggested fix

Make the recipe fail on any failed step. Either chain the per-arch commands with `&&` instead of `;`, or prepend `set -e;` to the loop body so a failed `go build` (or `ln`) aborts the recipe with a non-zero status. That converts the confusing downstream `COPY ... not found` into a clear `go build` failure at the RUN step and prevents BuildKit from caching an incomplete builder layer. (`release-build-kukebuild` at `Makefile:115` should be audited for the same pattern.)

eminwux:
## Summary
- `release-build` (`Makefile:88`) and `release-build-kukebuild` (`Makefile:115`) run their per-OS/per-ARCH `for` loops in a single `/bin/sh -c` with commands chained by `;`. A non-final `go build` failure neither stopped the loop nor failed the recipe — the recipe's exit status was that of the *last* command, so the `RUN make release-build` step in the Dockerfile `builder` stage exited 0 despite producing no binary. BuildKit then cached that incomplete layer, surfacing the failure later as a misleading `COPY ... /workspace/kuke-linux-amd64: not found` that persisted across builds until the cache was manually busted.
- Prepended `set -e;` to both recipe bodies so any failed `go build`/`ln` aborts the recipe with a non-zero status, turning the masked failure into a clear, immediate `go build` error at the RUN step and preventing BuildKit from caching an incomplete builder layer.
- Fixed both recipes flagged in the issue (`release-build` and the audited `release-build-kukebuild`).

## Test plan
- [x] `make test` — passes (root module + kukebuild module).
- [x] `make release-build` happy path — exits 0, all six artifacts (`kuke`/`kukeond`/`kuketty` × `linux/amd64`,`linux/arm64`) produced.
- [x] Fail-fast verified via a standalone Makefile reproduction: with `set -e;`, a forced `false` in the loop body aborts at the first failure (`make ... Error 1`, exit 2) and the loop does not continue; without it the loop ran to completion and exited 0.

Closes #713

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,7 @@ kukebuild:
 
 release-build:
 	# Build for all OS and ARCH combinations
+	set -e; \
 	for OS in $(OS); do \
 		for ARCH in $(ARCHS); do \
 			GO111MODULE=on CGO_ENABLED=0 GOOS=$$OS GOARCH=$$ARCH \
@@ -113,6 +114,7 @@ release-build:
 # never needs the go-1.25 BuildKit closure.
 .PHONY: release-build-kukebuild
 release-build-kukebuild:
+	set -e; \
 	for OS in $(OS); do \
 		for ARCH in $(ARCHS); do \
 			cd $(CURDIR)/cmd/kukebuild && \


eminwux:
**PR #729 Review — `fix: fail-fast in release-build loops to stop masking build errors`**

LGTM — both recipes run the whole `for` loop as one shell line, so Make only checks the last command's status; prepending `set -e;` correctly aborts on any intermediate `go build`/`ln` failure and stops BuildKit from caching an incomplete builder layer.

PR #730, opened 2026-05-22
fix: deterministic CNI network name on RecreateCell + start teardown purge

eminwux:
## Summary
- Two teardown purge sites were still resolving the CNI network name via the swallowing `resolveRootCNINetworkName`, which returns `""` (gating out the IPAM-file purge) whenever `GetSpace`/`getSpaceNetworkName` fails — leaking the reservation when space metadata is gone or corrupt. #708 fixed the equivalent swallow on stop/kill/delete; these two were missed.
- `recreate_cell.go:140` (release-before-recreate teardown) and `start.go:472` (container-exists release-before-delete branch, #630) now use the deterministic `buildRootCNINetworkName(realm, space)`, matching `stop.go` / `kill.go` / `delete_cell.go`. The network name is a pure function of (realm, space), so the purge runs even when space metadata is unreadable (issue #685 rationale).

## start.go:442/450 decision (the explicit call the issue asked for)
- **Kept on `resolveRootCNINetworkName`.** That site is the *create-fresh* re-ADD safety net (`ErrContainerNotFound` branch), not a teardown — no existing container to release; it pre-emptively scrubs a *potential* stale reservation before CNI ADD. It runs at cell start, where the space is expected present, a different risk profile than teardown (which can run after a cascade space deletion). This matches the `buildRootCNINetworkName` docstring's documented teardown-vs-re-ADD boundary and the issue author's note that metadata "is expected present" there.
- **Finding for reviewer (push back if you disagree):** verified that `networkName` at both `start.go:450` and `:472` feeds *only* `purgeCNIForContainer` — the actual CNI ADD at `start.go:677` resolves its network from `cniConfigPath`, independent of this string. So `start.go:450`'s purge would, in principle, suffer the same silent-skip-on-missing-metadata behavior. I left it on the resolver to keep the diff scoped to the AC's two teardown sites and to preserve `resolveRootCNINetworkName`'s sole remaining production caller rather than orphaning the function in a bug-fix PR. A reviewer who wants the create-fresh scrub deterministic too can say so and it's a one-line follow-up.

## Test plan
- [x] `make test` (test-root + test-kukebuild) — pass
- [x] `make kuke` build — pass
- [x] `go vet ./internal/controller/runner/` — clean
- [x] `go test ./internal/controller/runner/` — pass (both resolver unit tests in `helpers_test.go` still valid; both functions retained)
[...]

PR #732, opened 2026-05-22
fix: deterministic CNI scrub network name on create-fresh path

### LINKED ISSUE #731: runner: make create-fresh CNI scrub deterministic at start.go:450
Surfaced during review of #730 (the author explicitly flagged it as a deferred scope decision).

`internal/controller/runner/start.go:450` — the create-fresh re-ADD branch (`ErrContainerNotFound`) — still resolves the purge network name via the swallowing `resolveRootCNINetworkName`:

```go
networkName := r.resolveRootCNINetworkName(realmID, spaceID)
purgeStaleRootContainerCNI(networkName, func() {
    _ = r.purgeCNIForContainer(containerID, "", networkName)
})
```

`networkName` here feeds **only** `purgeCNIForContainer`, not the actual CNI ADD (which resolves from `cniConfigPath` at `start.go:677`). So when space metadata is gone or corrupt, the resolver returns `""` and the stale-reservation scrub is silently skipped — the same silent-skip behavior #730 just fixed at the two teardown sites (#685 rationale). In the corrupt-metadata + stale-reservation case this can leave a residual host-local IPAM allocation that gets the re-ADD rejected as a duplicate (#630).

#730 deliberately scoped to the two teardown sites named in #717 and kept this on the resolver because the create-fresh path runs at cell start where space metadata is expected present — a lower-risk profile. The one-line follow-up the author offered:

```go
networkName := r.buildRootCNINetworkName(realmID, spaceID)
```

Single site. Filing because it's a latent correctness risk (stale-reservation leak), not a cosmetic nit.

eminwux:
## Summary
- The create-fresh re-ADD branch (`ErrContainerNotFound`) at `internal/controller/runner/start.go` resolved the stale-reservation scrub's network name via the swallowing `resolveRootCNINetworkName`, which returns `""` when space metadata is gone or corrupt — silently skipping the scrub. That can leave a residual host-local IPAM allocation that gets the re-ADD rejected as a duplicate (#630).
- Swapped that single site to `buildRootCNINetworkName`, which derives the name deterministically from realm+space via `naming.BuildSpaceNetworkName`. This matches the stop/kill/delete teardown paths #730 fixed for the same #685 silent-skip class. The actual CNI ADD still resolves from `cniConfigPath` and is unaffected.
- Updated the two now-stale doc comments (the `purgeStaleRootContainerCNI` godoc and the `TestPurgeStaleRootContainerCNI_CreateFreshSafetyNet` comment) that named the old resolver as the source of the empty-name no-op.

`resolveRootCNINetworkName` is intentionally left in place — it retains dedicated unit coverage (`helpers_test.go`) and removing it would be an unrelated cleanup outside this nit's scope.

## Test plan
- [x] `make test` (test-root + test-kukebuild) — all packages pass
- [x] `GOWORK=off go build ./...` / `go vet ./internal/controller/runner/...`
- [x] `go test ./internal/controller/runner/...`
- [ ] `make dev-init` smoke — not run: this is a one-line CNI network-name resolver swap on the corrupt-metadata create-fresh edge path (covered by the unit tests above); dev-init's regression guard is realm/bind-mount parity, which this change does not touch, and the affected path requires a corrupt-space + stale-reservation state not reproducible on a clean re-bootstrap.

Closes #731

### DIFF
--- a/internal/controller/runner/start.go
+++ b/internal/controller/runner/start.go
@@ -164,7 +164,7 @@ func teardownRootContainerCNI(releaseCNI func(), deleteContainer func() error, p
 // before CNI ADD on the ErrContainerNotFound branch, giving the create-fresh
 // path the same second line of defence the teardown branch already has.
 //
-// Best-effort and a no-op when networkName is "" (resolveRootCNINetworkName
+// Best-effort and a no-op when networkName is "" (buildRootCNINetworkName
 // couldn't derive a name), so a clean start with no stale reservation is
 // unaffected. Decoupled from the concrete CNI client — same rationale as
 // teardownRootContainerCNI — so the run/no-op decision is unit-testable.
@@ -447,7 +447,10 @@ func (r *Exec) startCellLocked(cell intmodel.Cell) (_ intmodel.Cell, retErr erro
 			// reservation keyed to this deterministic root-container ID. Scrub
 			// it before CNI ADD so the re-ADD isn't rejected as a duplicate
 			// allocation. Best-effort and a no-op when no stale file exists.
-			networkName := r.resolveRootCNINetworkName(realmID, spaceID)
+			// Resolve the network name deterministically from realm+space so
+			// the scrub runs even when space metadata is gone or corrupt
+			// (issue #685), matching the stop/kill/delete teardown paths.
+			networkName := r.buildRootCNINetworkName(realmID, spaceID)
 			purgeStaleRootContainerCNI(networkName, func() {
 				_ = r.purgeCNIForContainer(containerID, "", networkName)
 			})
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #732 Review — `fix: deterministic CNI scrub network name on create-fresh path`**

LGTM — swaps the create-fresh re-ADD scrub from the `GetSpace`-dependent `resolveRootCNINetworkName` (silently `""` on corrupt metadata → skipped scrub, #630) to the deterministic `buildRootCNINetworkName`, matching the #730 teardown precedent. Inputs (`realmID`/`spaceID`) come from the validated cell spec, and the empty case stays a safe no-op. Commits signed; CI green.

One non-blocking follow-up (filed separately, not gating):

- **Dead code** `internal/controller/runner/helpers.go:462`: `resolveRootCNINetworkName` now has zero production callers (this was its last one) and its godoc at line 455 still names `start.go`/`recreate_cell.go` callers that no longer exist. Author deliberately deferred removal — fine to clean up later.

Issue #737, opened 2026-05-23
tty: render-time gate omitting done create stages + edited-stage re-run (phase C2)


eminwux:
## Context

Phase C1 (#690) makes `ContainerStatus.Stages` durable across stop/start with content-hash keying. This phase consumes that durable state: `writeKukettyDoc` omits already-done `create` stages from the next boot's `ContainerDoc` so they do not re-run, and an *edited* stage (hash mismatch) re-renders to re-run.

`writeKukettyDoc` (`internal/controller/runner/attachable.go:197`) is the per-container `ContainerDoc` renderer that replaced `writeKukettyMetadata` in phase 0 (#641). It currently emits every declared `create` stage. The done-set must flow from controller state through `attachableBuildOpts` (4 call sites: `start.go:842`, `start.go:1108`, `provision.go:1560`, `provision.go:2024`) into the renderer.

The interaction with the #423/#625 phase-4.3 divergent-spec / reconcile path also lands here: when reconcile drives the container back to a prior spec, the done-set must continue to gate correctly (hashes carried in `Stages` either match a stage in the rolled-back spec — skip — or don't — re-run).

## Proposal

- **Render gate:** `writeKukettyDoc` reads the controller-provided done-set (`map[hash]struct{}` derived from `State == "done"` entries in `ContainerStatus.Stages`) and omits matching `create` stages from the rendered doc.
- **Plumbing:** add the done-set to `attachableBuildOpts`' signature and thread it through the 4 call sites. Sourced from the live `ContainerStatus.Stages` on the cell being rendered.
- **Edit re-run:** an edited stage produces a new content hash, no done-record matches, the renderer includes it, kuketty's pre-Serve executor re-runs it, and phase B's populate stamps the new hash.
- **#625 reconcile interaction:** documented in the design notes and exercised by a unit test that flips spec back to a prior version and confirms the done-set still gates correctly.

## Acceptance criteria

- [ ] On a subsequent boot, `writeKukettyDoc` omits `create` stages whose hash appears in the done-set; the rendered `ContainerDoc` is the render-time evidence the gate fires.
- [ ] An edited `create` stage (new hash) re-renders and re-runs on the next boot.
- [ ] The #625 reconcile path is documented and exercised by a unit test that rolls back spec and confirms hash-keyed gating still applies.
- [ ] Unit tests: gate-on-restart skip, edited-stage re-run, reconcile-path round-trip.

## Dependencies

[...]

Issue #423, opened 2026-05-12
planning: absorb crew layers — repos, git, Secret, Blueprint/Config


eminwux:
## Context

`eminwux/crew` is the host-side bootstrap that turns a kukeon `CellProfile` into a working "crew member" cell. Inspection on 2026-05-12 shows what crew provides today vs. what kukeon models:

| Layer | Mechanism (crew) | kukeon today |
|---|---|---|
| Fleet matrix (projects × roles) | `crew.conf` + `install.sh` | not modeled |
| Profile rendering | `envsubst` over `templates/*.yaml.template` | `spec.parameters[]` + `--param` (scalar substitution only) |
| Container images | Dockerfiles + `build-images.sh` → `kuke image load` | external (correct) |
| Host state tree | `~/.crew/{claude,claude.json,config,secrets.env}` | not modeled |
| Secret injection | `render-profile.sh` (envsubst `${TOKEN}` → pipe to `kuke apply -f -`) | `ContainerSecret{fromFile,fromEnv,mountPath}` exists (`pkg/api/model/v1beta1/container.go:191`, runtime `internal/ctr/secrets.go`) — purely host-path references; daemon never stores bytes |
| Git identity & signing | 13× `GIT_AUTHOR_* / GIT_COMMITTER_* / GIT_CONFIG_*` env vars baked per template | not modeled |
| Repo cloning | `onInit` shell scripts: `git clone $PROJECT_REPO`, etc. | not modeled — every template duplicates the same script |
| Claude credentials | bind-mount `~/.crew/claude/.credentials.json` (writable — refresh tokens) | bind mounts work as-is |
| `gh` credentials | bind-mount `~/.crew/config/gh/hosts.yml` | bind mounts work as-is |
| Crew-member identity (project, role, image, state-tree, profile) | encoded via directory layout + `crew.conf` | not modeled |

Several layers are duplicated across every template and become first-class primitives under this umbrella — collapsing crew templates from ~80 to ~30 lines and making lifecycle observable in `Status`. The largest single payoff is the new kind hierarchy (`CellBlueprint` + `CellConfig` + `Cell`) that supersedes today's client-side-only `CellProfile` — added **alongside** `CellProfile` (additive, with `kuke run -b`/`-c`), with `-p` + `CellProfile` removed only in the final slice. It depends on the smaller absorptions and on the daemon-stored-resource architecture introduced in phase 3.

### Where new fields go: Container, not Cell

[...]

PR #739, opened 2026-05-23
feat(apischeme): reject runOn: create when no persistent writable mount

### LINKED ISSUE #738: apischeme: reject/warn runOn: create when container has no persistent writable mount (phase C3)
## Context

A `runOn: create` stage writes side effects (`npm ci`, DB seed, bootstrap) into the container's filesystem. If the container has no persistent writable mount (all-tmpfs, readonly-rootfs), those side effects evaporate on the next restart while `ContainerStatus.Stages` continues to report `State == "done"` — the run-once gate (phase C2, #737) then skips a stage whose effect no longer exists.

`TtyStage` has no `target` field (a stage is an opaque `Script`), so apischeme cannot statically reason about each stage's individual write target. The reasoning has to be at the *container* scope: if the container has no persistent writable mount, every `runOn: create` stage is at risk.

This phase settles the AC-6 design decision deferred from #635: reject (or warn) `runOn: create` when the container has no persistent writable mount.

## Proposal

- **Validation rule:** apischeme rejects (or downgrades to warn-only per the settled gate) a `ContainerSpec` that declares `runOn: create` stages when the container's mount set is all-tmpfs or its rootfs is readonly and no persistent volume mount is declared.
- **Recommended gate:** reject by default — the run-once contract is too easily broken silently. Add a `--force-create-on-ephemeral` opt-in (or equivalent spec-level annotation) if operators have a legitimate ephemeral-bootstrap use case.
- **Alternative gate:** if reviewer prefers, downgrade to warn-only — `kuke apply` prints a warning but the spec is admitted.

## Acceptance criteria

- [ ] `runOn: create` on an all-tmpfs container is rejected (or warned, per the settled gate).
- [ ] `runOn: create` on a readonly-rootfs container with no persistent volume mount is rejected (or warned).
- [ ] `runOn: create` on a container with at least one persistent writable mount (host bind, named volume, `pvc:`) is accepted unchanged.
- [ ] Unit tests cover all-tmpfs, readonly-rootfs, mixed-mount, and persistent-only cases.

## Dependencies

- Relates to phase C1 (#690) and phase C2 (#737): the validation closes the silent-skip footgun the run-once gate would otherwise have on ephemeral targets. Lands independently of either.

## Notes

- Part of the #423 crew-absorption series (`blueprint`).
- Sliced out of the original phase-C scope (#690) per dev's resize-issue flag.

eminwux:
## Summary
- Phase C3 of the run-once tty-stage series (#423 umbrella). Adds an apischeme validation pass that rejects a `ContainerSpec` declaring any `runOn: create` `TtyStage` when the container has no persistent writable mount, closing the silent-skip footgun the run-once gate (phase C2, #737) would otherwise have on ephemeral targets.
- A mount counts as "persistent writable" when it is a `VolumeMount` whose `Kind` is anything other than `tmpfs` (`bind`, plus the empty back-compat default, plus any future persistent kind such as a future `pvc:`) and is not declared `ReadOnly`. `ContainerSpec.Tmpfs` entries are explicitly ephemeral and never count. `ReadOnlyRootFilesystem` alone is not what gates the decision — only the absence of a persistent writable `VolumeMount`.
- Chose the **reject-by-default** gate per the issue's recommended option. The proposal's alternative `--force-create-on-ephemeral` opt-in is **not** added in this PR — no AC checkbox covers it, and shipping the opt-in adds CLI/spec surface that is best deferred until operators have a concrete ephemeral-bootstrap use case. Easy to layer on later as a separate spec-level annotation or CLI flag without changing the reject default.
- The new check is wired into the existing `validateContainerTty` after the per-stage `runOn` enum validation, so it fires on both the `ContainerDoc` and the nested-in-`CellDoc` paths without a second call-site touch. `CellBlueprintDoc` stores the spec as opaque YAML, so it is unaffected (validation happens when the materialized doc enters the normal `Normalize*` lane).

## Test plan
- [x] `make test` (project CI gate — `test-root` + `test-kukebuild`) — both green.
- [x] `make test-root` — verbose re-run to confirm every package in the apischeme dependency closure still passes after the new validation lands.
- [x] New unit-test matrix `TestContainerCreateStagePersistenceValidation` covers all four AC cases plus the cell-doc embedding path and the no-create-stage exemption: all-tmpfs (rejected), readonly-rootfs without volumes (rejected), no-volumes writable-rootfs (rejected — the writable container layer is also ephemeral under kukeon's recreate model), readonly-rootfs with a bind (accepted), mixed mounts tmpfs+bind (accepted), persistent-only (accepted), empty-`Kind` back-compat default (accepted as bind), readonly-bind-only (rejected — not writable), and start-only stages (unaffected).
- [x] Updated the existing `TestContainerTtyRoundTripV1Beta1` and `TestContainerTtyStageRunOnEnumValidation` so their `runOn: create` cases carry a bind mount — the round-trip and enum-validation tests still exercise only what they advertise, and would otherwise have been collateral failures.
- [ ] `make e2e` and the `make dev-init` smoke — not run; the change is pure declarative validation in the apischeme layer with no daemon, runtime, or build-path touch.

## Out of scope (filed as follow-ups)
- User-facing docs (`docs/site/manifests/container.md`) — the `runOn: create` reference text and YAML example will be updated to mention the persistent-writable-mount requirement. Repo convention (e.g. PRs #706, #687) keeps `docs(manifests):` updates in their own PR under the `documentation` label; filing as a docs follow-up alongside this PR.
- The optional `--force-create-on-ephemeral` opt-in mentioned in the issue's Proposal section (intentionally deferred — see PR summary).

Closes #738

### DIFF
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -689,7 +689,58 @@ func validateContainerTty(spec ext.ContainerSpec) error {
 			return fmt.Errorf("container %q: onInit[%d]: %w", spec.ID, i, err)
 		}
 	}
+	if err := validateContainerCreateStagePersistence(spec); err != nil {
+		return err
+	}
 	return nil
+}
+
+// validateContainerCreateStagePersistence enforces that a container declaring
+// runOn: create stages has at least one persistent writable mount. Without one,
+// the side effects of create stages (npm ci, DB seed, bootstrap) evaporate when
+// the container's writable layer is dropped on the next recreate while
+// ContainerStatus.Stages still reports State == "done" — the run-once gate
+// (phase C2, #737) would then silently skip a stage whose effect no longer
+// exists. TtyStage has no per-stage target field, so the reasoning is
+// container-scoped: any runOn: create stage on a container with no persistent
+// writable mount is at risk. Issue #738.
+//
+// A mount is treated as a persistent writable target when it is a VolumeMount
+// whose Kind is anything other than tmpfs (so VolumeKindBind, including the
+// empty back-compat default, plus any future persistent kind like a future
+// pvc:) and is not declared ReadOnly. ReadOnlyRootFilesystem alone does not
+// affect the decision — the test is solely the presence of at least one
+// persistent writable VolumeMount. ContainerSpec.Tmpfs entries are explicitly
+// ephemeral and never count.
+func validateContainerCreateStagePersistence(spec ext.ContainerSpec) error {
+	if spec.Tty == nil {
+		return nil
+	}
+	hasCreate := false
+	for _, s := range spec.Tty.OnInit {
+		if s.RunOn == ext.RunOnCreate {
+			hasCreate = true
+			break
+		}
+	}
+	if !hasCreate {
+		return nil
+	}
+	for _, v := range spec.Volumes {
+		if v.Kind == ext.VolumeKindTmpfs {
+			continue
+		}
+		if v.ReadOnly {
+			continue
+		}
+		return nil
+	}
+	return fmt.Errorf(
+		"container %q: tty.onInit has runOn: %q stages but the container has no persistent writable mount; "+
+			"declare a volumes: entry of kind bind covering the stage's write target so its side effects "+
+			"survive container recreate, or move the work into a runOn: %q stage",
+		spec.ID, ext.RunOnCreate, ext.RunOnStart,
+	)
 }
 
 // validateTtyLogLevel accepts the empty string (the daemon defaults to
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #739 Review — `feat(apischeme): reject runOn: create when no persistent writable mount`**

LGTM — pure declarative validation wired into the existing `validateContainerTty` lane so it covers both `NormalizeContainer` and `NormalizeCell` paths (verified), the test matrix walks every AC checkbox plus the empty-`Kind` back-compat default and the cell-doc embedding path, and the smoke-skip is justified (no daemon, build-path, or `/opt/kukeon` touch).

Issue #644, opened 2026-05-21
config: add kuke get/delete config verbs (phase 4.2.2)


eminwux:
## Context

Phase 4b-ii of #423. Split from #624 (phase 4b-i) per the single-PR sizing rule, mirroring the phase-3 split (#619 schema+apply / #622 get/delete) and the phase-4a split (#620 / 4a-ii).

#624 ships `kind: CellConfig` + daemon store + `kuke apply` + slot-fill validation + stable-name + back-ref label. This issue adds the read and delete verbs.

## Proposal

CLI verbs against the daemon-stored CellConfig resource from #624:

- `kuke get configs [--scope realm=<r>|space=<s>|stack=<t>] [-o json]` — list, optionally scope-filtered.
- `kuke get config <name> [--scope ...] [-o json]` — show one (blueprint ref, values, slot fills).
- `kuke delete config <name> [--scope ...]` — remove from daemon storage.

Deleting a Config does **not** delete the cell it materialised (that's `kuke delete cell`); the interaction is the notice AC below.

## Acceptance criteria

- [ ] `kuke get configs` lists in current scope; `--scope` flags filter by realm/space/stack.
- [ ] `kuke get config <name>` shows a single Config (blueprint ref, values, repo/secret slot fills).
- [ ] `kuke delete config <name>` removes the file at `/opt/kukeon/<scope>/configs/<name>`.
- [ ] `kuke delete config <name>` emits a one-line notice (not a refusal) when a live cell carries the back-reference label to it, pointing at `kuke delete cell <name>`.
- [ ] Verbose / quiet / `-o json` output consistent with existing `kuke get` / `delete` verbs.

## Notes

- Phase 4b-ii of #423. Depends on **#624 (phase 4b-i)**.
- Sibling-precedent: Secret get/delete (#622), Blueprint get/delete (4a-ii).
- Config scope is realm/space/stack (not cell) — matches #624.

Issue #624, opened 2026-05-19
config: add CellConfig + slot-fill validation + identity (phase 4.2.1)


eminwux:
## Context

Phase 4b-i of #423 (slice of phase 4). Depends on #620 (`CellBlueprint` + daemon store) and #619 (storage primitive).

#620 introduces `kind: CellBlueprint` — the parametrised template. This issue introduces `kind: CellConfig` — the resource that **owns** a cell's identity. The load-bearing distinction (why `CellConfig` is a peer kind, not a `--values-from` flag) is in the umbrella's "Identity and lifecycle" section: one CellConfig → at most one live cell with a deterministic name within scope.

This issue *defines and stores* a Config — schema, daemon storage, `kuke apply`, slot-fill validation against the referenced Blueprint, the stable-name derivation, and the back-reference label. The runtime **state machine** that drives materialise/start/attach/refuse moves to **#625 (phase 4c)** with the `kuke run -c` verb it serves — a vertical cut, so neither issue is logic-without-a-caller.

## Proposal

New top-level kind:

```yaml
apiVersion: v1beta1
kind: CellConfig
metadata:
  name: kukeon-dev
  realm: kuke-system
  space: default
spec:
  blueprint: { name: dev, realm: kuke-system }
  values:
    PROJECT_DIR: kukeon
  repos:
    project:
      url: git@github.com:eminwux/kukeon.git
      branch: main
  secrets:
    anthropic-token:
      secretRef: { name: anthropic, realm: kuke-system }
```

Storage under `/opt/kukeon/<scope>/configs/<name>` — root-owned, world-readable (references only). Reuses the storage primitive from #619 and the structural-slot pattern from #620.

### Identity (defined here; driven by `run -c` in #625)

- **Stable name.** A materialised cell's name derives deterministically from `CellConfig.metadata.name` — this issue settles whether it's `<config.metadata.name>` verbatim or `<config.metadata.name>-<short-hash-of-values>`, and ships the derivation function.
- **One Config → at most one live cell** within scope. The materialised cell carries a back-reference label to the Config (analog of today's `cellprofile.LabelProfile`).
- The no-cell / running / stopped / error **state machine** that uses this identity to decide materialise / attach / start / refuse is **#625** (it ships with the `kuke run -c` CLI that drives it).

### L1↔L2 slot fill

[...]

PR #744, opened 2026-05-23
refactor(runner): unify reserved-subdir exclusion across subtree walkers

### LINKED ISSUE #734: chore: blueprint/secret subtree walkers don't exclude all reserved resource subdirs
## Context

Surfaced while implementing #644 (`kuke get/delete config`). Each scope level under `<RunPath>/data/<realm>/...` mixes child-scope subdirectories (spaces, stacks, cells) with reserved resource subdirectories: `secrets/` (#619), `blueprints/` (#620), and `configs/` (#624). The subtree-list walkers enumerate child scopes by reading the scope dir and must skip the reserved subdirs so none is mistaken for a phantom space/stack.

The three walkers are inconsistent about which reserved subdirs they exclude:

- `internal/controller/runner/config.go` `configChildScopeNames` (added in #644) — excludes **all three** (`secrets`, `blueprints`, `configs`). Correct.
- `internal/controller/runner/blueprint.go` `blueprintChildScopeNames` — excludes `secrets` + `blueprints`, **misses `configs`** (predates #624's `configs/` subdir).
- `internal/controller/runner/secret.go` `childScopeNames` (~line 226) — excludes `secrets` only, **misses `blueprints` AND `configs`**.

### Why it's benign today (but worth fixing)

The phantom descent currently finds nothing: e.g. `kuke get blueprints` descending into a `configs/` dir as a fake space then reads `<realm>/configs/blueprints/`, which doesn't exist, yielding no blueprints. The reserved dirs hold files (or a differently-named resource subdir), not the walker's own resource subdir, so no wrong rows leak. But it's an asymmetric, fragile invariant: a future layout change (e.g. a reserved resource that itself nests a `blueprints/` or `secrets/` subdir) would surface phantom entries, and the inconsistency is a latent trap for the next reader.

## Proposal

Make the three walkers share one reserved-subdir exclusion set (all of `consts.Kukeon{Secrets,Blueprints,Configs}Subdir`), ideally via a single shared helper rather than three near-identical `childScopeNames` variants.

## Acceptance criteria

- [ ] `blueprintChildScopeNames` excludes `configs/` (and `secrets/`, `blueprints/`).
- [ ] secret `childScopeNames` excludes `blueprints/` and `configs/` (and `secrets/`).
- [ ] Ideally the three walkers share one exclusion helper so a future reserved subdir is added in exactly one place.
- [ ] Regression tests: a sibling `configs/` (and `blueprints/`/`secrets/`) reserved subdir is never surfaced as a child scope by `kuke get secrets` / `kuke get blueprints`. (#644 already covers this for `kuke get configs` via `TestListConfigs_IgnoresReservedSubdirs`.)

## Out of scope

- Any change to the on-disk layout itself.

## References

- `internal/controller/runner/blueprint.go` — `blueprintChildScopeNames`.
- `internal/controller/runner/secret.go` — `childScopeNames` (~line 226).
- `internal/controller/runner/config.go` — `configChildScopeNames` (the correct reference impl, from #644).
- `internal/consts/consts.go` — `KukeonSecretsSubdir` / `KukeonBlueprintsSubdir` / `KukeonConfigsSubdir`.

eminwux:
## Summary

- Collapses the three near-duplicate child-scope walkers (`childScopeNames` in `secret.go`, `blueprintChildScopeNames` in `blueprint.go`, `configChildScopeNames` in `config.go`) onto a single shared `(r *Exec) childScopeNames(dir, want)` in a new `internal/controller/runner/scope_children.go`, with one `reservedScopeSubdirs` set as the source of truth for the per-scope reserved resource subdirectories to skip.
- Fixes the inconsistency #734 named: the blueprint walker missed `configs/` (predated #624's configs/ subdir); the secret walker missed both `blueprints/` and `configs/`. Both now exclude all three.
- Refactor only — behavior previously was benign in steady state (a phantom descent finds nothing since the reserved dirs hold files, not the walker's own resource subdir), but the inconsistency was a latent trap for a future reserved subdir or a layout change that nested matching subdirs. After this, a fourth reserved resource is added in exactly one place: `reservedScopeSubdirs`.

## Test plan

- [x] `make test` — full repo suite, including the new + broadened regression tests:
  - **new** `TestListSecrets_IgnoresReservedSubdirs` — seeds a realm secret, blueprint, and config; asserts `ListSecrets("")` yields only the secret (the blueprints/ + configs/ subdirs were previously fall-through to a phantom space).
  - **broadened** `TestListBlueprints_IgnoresReservedSubdirs` — now also seeds a realm config so the configs/ exclusion is pinned by `ListBlueprints` (was the case the old `blueprintChildScopeNames` missed).
  - **unchanged** `TestListConfigs_IgnoresReservedSubdirs` — the original reference case from #644, still passing.
- [x] `pre-commit run --files <changed>` — all hooks pass except `golangci-lint`, which panics with the known go1.24-binary-vs-go1.25-module incompatibility documented in agent memory; environmental, not a finding (`go vet` covers the equivalent checks and is green).
- [ ] `make dev-init` smoke test — not run; this is a pure internal refactor of subtree-list walkers (no build path, daemon, or `/opt/kukeon` touch), so CLAUDE.md's smoke-when-touching-daemon-or-build clause does not apply.

## AC mapping

- [x] `blueprintChildScopeNames` excludes `configs/` — folded into the shared helper.
- [x] secret `childScopeNames` excludes `blueprints/` and `configs/` — folded into the shared helper.
- [x] The three walkers share one exclusion helper (`reservedScopeSubdirs` + `(r *Exec) childScopeNames`).
- [x] Regression tests for `kuke get secrets` / `kuke get blueprints` against sibling reserved subdirs.

## Files

- `internal/controller/runner/scope_children.go` — **new**; `reservedScopeSubdirs` map and the shared `(r *Exec) childScopeNames(dir, want)` method.
- `internal/controller/runner/secret.go` — deletes the local `childScopeNames`; drops the now-unused `consts` import. Call sites in `collectSecretSubtree` resolve to the shared method by name.
- `internal/controller/runner/blueprint.go` — deletes `blueprintChildScopeNames`; updates the two call sites in `collectBlueprintSubtree` to `r.childScopeNames`; drops the now-unused `consts` import.
- `internal/controller/runner/config.go` — deletes `configChildScopeNames`; updates the two call sites in `collectConfigSubtree` to `r.childScopeNames`; drops the now-unused `consts` import.
- `internal/controller/runner/secret_test.go` — adds `TestListSecrets_IgnoresReservedSubdirs`.
- `internal/controller/runner/blueprint_test.go` — broadens `TestListBlueprints_IgnoresReservedSubdirs` to also seed a realm config (pinning the configs/ exclusion the old helper missed).

Closes #734

### DIFF
--- a/internal/controller/runner/blueprint.go
+++ b/internal/controller/runner/blueprint.go
@@ -23,7 +23,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/errdefs"
 	intmodel "github.com/eminwux/kukeon/internal/modelhub"
 	"github.com/eminwux/kukeon/internal/util/fs"
@@ -142,7 +141,7 @@ func (r *Exec) collectBlueprintSubtree(out *[]intmodel.CellBlueprint, realm, spa
 		}
 	}
 
-	spaces, err := r.blueprintChildScopeNames(fs.RealmMetadataDir(r.opts.RunPath, realm), space)
+	spaces, err := r.childScopeNames(fs.RealmMetadataDir(r.opts.RunPath, realm), space)
 	if err != nil {
 		return err
 	}
@@ -153,7 +152,7 @@ func (r *Exec) collectBlueprintSubtree(out *[]intmodel.CellBlueprint, realm, spa
 			}
 		}
 
-		stacks, stErr := r.blueprintChildScopeNames(fs.SpaceMetadataDir(r.opts.RunPath, realm, sp), stack)
+		stacks, stErr := r.childScopeNames(fs.SpaceMetadataDir(r.opts.RunPath, realm, sp), stack)
 		if stErr != nil {
 			return stErr
 		}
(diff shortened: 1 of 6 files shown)


eminwux:
**PR #744 Review — `refactor(runner): unify reserved-subdir exclusion across subtree walkers`**

LGTM — folds three drifted child-scope helpers into one shared `childScopeNames` keyed off `reservedScopeSubdirs`, fixes the secret/blueprint walkers that missed `configs/` (and `blueprints/`) per #734, and pins the broadened exclusions with `TestListSecrets_IgnoresReservedSubdirs` + the broadened `TestListBlueprints_IgnoresReservedSubdirs`. Behavior preserved on the `want`-filter branch (matches prior helpers), old helper names fully removed, `consts` imports cleanly dropped, signature verified.

PR #746, opened 2026-05-23
chore: scope pre-commit golangci-lint to root module via GOWORK=off

### LINKED ISSUE #693: chore: pre-commit golangci-lint cannot analyze the go1.25 kukebuild module
## Context

Surfaced while implementing #656 (a kukebuild refactor). The `golangci-lint` pre-commit hook silently fails to lint the `cmd/kukebuild` module, so every kukebuild contributor hits a confusing panic at commit time and the module ships locally unlinted.

## Problem

`.pre-commit-config.yaml:38-42` pins `golangci-lint` at `v2.6.2`. That release's own module declares `go 1.24`, so the binary refuses any module targeting a newer Go — and `cmd/kukebuild/go.mod` is `go 1.25.5` (the deliberate module split from #662). The hook errors with:

```
panic: package requires newer Go version go1.25 (application built with go1.24)
```

or, with a freshly-installed binary:

```
can't load config: the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.5)
```

Rebuilding/reinstalling golangci-lint v2.6.2 under go1.25 does **not** help — the go1.24 floor is baked into that release's own `go.mod`, not the build toolchain.

Net effect: a contributor touching `cmd/kukebuild` must `SKIP=golangci-lint git commit` to get past the hook, and the kukebuild module receives no lint coverage. (CI is unaffected — `.github/workflows/test.yaml` runs only `make test` + `make e2e`, no lint step — so this is purely local dev-loop friction, not a CI gate.)

## Proposal

Pick one:

1. **Bump golangci-lint** to a release whose module targets `go >= 1.25` so the hook can analyze both modules. Verify it still lints the root (`go 1.24.5`) module too.
2. **Scope the hook to the root module** and add a second hook entry (or a `local` hook) that runs golangci-lint from inside `cmd/kukebuild/` with a go1.25-capable binary, so kukebuild keeps lint coverage.
3. **Explicitly exclude `cmd/kukebuild/`** from the golangci-lint hook with an inline comment pointing at the module-split rationale, accepting no kukebuild lint coverage until option 1 is viable — at minimum this removes the confusing panic.

## Acceptance criteria

- [ ] `pre-commit run golangci-lint --files cmd/kukebuild/build.go` exits cleanly (lints or cleanly skips — no panic / version-floor error).
- [ ] The chosen approach is documented inline in `.pre-commit-config.yaml` referencing the #662 module split.
- [ ] Root-module lint coverage is unchanged.

## Out of scope

- Any change to CI (`.github/workflows/test.yaml` runs no lint step today; this issue does not propose adding one).

## References

- Surfaced during #656 / PR #691.
- Module split: #662 (`cmd/kukebuild/go.mod`, `go 1.25.5`) and #655 (`go.work`).
- Hook pin: `.pre-commit-config.yaml:38-42`.

eminwux:
## Summary

- `.pre-commit-config.yaml`'s golangci-lint hook (pinned at v2.6.2, built with go 1.24) was crashing the type checker with `package requires newer Go version go1.25 (application built with go1.24)` whenever it was invoked on any change while `go.work` was enabled. The workspace package loader was pulling `cmd/kukebuild` (go 1.25.5, #662 module split) into analysis even though `./...` from the root module doesn't list those packages.
- Closest to the issue's **option 3** (exclude `cmd/kukebuild/`). The mechanism is "disable workspace mode" (`env GOWORK=off`) rather than enumerating root packages, because explicit package enumeration still panics — the workspace loader pulls kukebuild in transitively even when only root patterns are passed. `GOWORK=off` is the smallest change that actually stops the analyzer walking into the go 1.25 module.
- Inline comment in the hook references the #662 module split and points at #693 for the upgrade path (bump golangci-lint to a release built with go 1.25+, then drop `GOWORK=off` and add a second hook for kukebuild).
- Root-module lint behavior is unchanged: same binary, same args, same `./...`, same `.golangci.yml` config — only the workspace mode bit flips.
- The `entry:` override drifts from upstream's `golangci-lint run --new-from-rev HEAD --fix` if that string ever changes; the comment block makes the divergence obvious to the next reader.
- This is dev-loop friction only: `.github/workflows/test.yaml`'s `paths:` filter excludes `.pre-commit-config.yaml`, so this PR will not trigger CI.

## Test plan

- [x] `pre-commit run golangci-lint --files cmd/kukebuild/build.go` exits cleanly (AC1) — was a multi-thousand-line panic stack, now `Passed` in ~5s.
- [x] `pre-commit run golangci-lint --files cmd/kuke/apply/apply.go` still `Passed` (root-module lint unchanged, AC3).
- [x] `pre-commit run golangci-lint --all-files` `Passed` (no new findings surfaced anywhere).
- [x] Approach documented inline in `.pre-commit-config.yaml` referencing #662 (AC2).
- [ ] `make test` / `make e2e` not run — `.pre-commit-config.yaml` is not on CI's `paths:` trigger list and the diff touches zero Go code; the four pre-commit checks above cover the AC.

Closes #693

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,6 +39,14 @@ repos:
     rev: v2.6.2
     hooks:
       - id: golangci-lint
+        # cmd/kukebuild lives in its own go 1.25 module (#662 module split).
+        # v2.6.2's binary is built with go 1.24; with go.work enabled, the
+        # workspace package loader pulls kukebuild into analysis and the
+        # type checker panics ("package requires newer Go version go1.25").
+        # Disable workspace mode so the hook only sees the root module
+        # (go 1.24.5). cmd/kukebuild stays unlinted by pre-commit until
+        # golangci-lint is bumped to a release built with go 1.25+ (#693).
+        entry: env GOWORK=off golangci-lint run --new-from-rev HEAD --fix
         args: ["./..."]
 
   - repo: https://github.com/pre-commit/mirrors-prettier


eminwux:
**PR #746 Review — `chore: scope pre-commit golangci-lint to root module via GOWORK=off`**

LGTM — minimal dev-loop fix: `entry:` override faithfully reproduces upstream v2.6.2's default (`golangci-lint run --new-from-rev HEAD --fix`) with `env GOWORK=off` prepended, comment block points the next reader at #662 (cause) and #693 (unwind path).

PR #750, opened 2026-05-23
docs(cli): cover kuke run -b/-c and get/delete blueprint+config

eminwux:
## Summary
Best-effort documentation update applied from issue #748.
Mode: best-effort — the issue body identified files and sections but did not provide paste-ready Before/After wording, so wording was authored to match the issue's Proposal and Acceptance criteria.

## Files changed
- `docs/site/cli/kuke-run.md` — synopsis, intro, flags table, new `## Blueprints and Configs` section (with the `-c` identity state machine subsection), three new examples; `## Profiles` heading carries a deprecation note pointing to `-b`/`-c`.
- `docs/site/cli/kuke-get.md` — resources line extended with `blueprint`/`config` (plus plurals and `bp`/`cfg` aliases); hierarchy-flags table gains rows for `get blueprint` and `get config`; two new examples.
- `docs/site/cli/kuke-delete.md` — added `### kuke delete blueprint` and `### kuke delete config` subsections after `### kuke delete container`, including the no-cascade note and the "delete config does not delete the cell" semantic that the per-resource Go source emits as a runtime notice; two new examples.

## Editorial choices
- `-b`/`-c` flags table rows phrased to mirror the `Use:` and flag-help strings in `cmd/kuke/run/run.go`, so the doc tracks the source's authoritative wording.
- The `-c` identity state machine surfaced as a four-row markdown table (no cell / live+running / live+stopped / live+error) rather than prose — easier to skim and maps 1:1 to the four cases the issue enumerates.
- The `## Profiles` section header got a one-line deprecation block (block quote) pointing to `-b`/`-c`; section body left intact per the issue's Notes block ("once #626 lands, a follow-up doc sweep can strip the deprecation note and remove the section — that cleanup is **not** part of this issue").
- Forward-links to `../manifests/blueprint.md` and `../manifests/config.md` — those pages landed on `main` in #749 during this branch's life (the companion issue the body referenced as "filed separately"), so the link is direct rather than deferred. After the rebase the references resolve cleanly.
- `kuke-delete.md`: the synopsis usage block was left unchanged per the issue ("the `<resource> <name>` shape as-is; the new subcommands fit the existing pattern").

## Acceptance criteria check
[...]

PR #742, opened 2026-05-23
feat: add kuke run -c + identity state machine (phase 4.3)

eminwux:
## Summary

- Adds `kuke run -c <config>` as the daemon-stored, idempotent counterpart to `-b`: the at-most-one live cell per Config that the CellConfig contract promises (#624, #423 phase 4c).
- New `internal/cellconfig.Materialize(cfg, bp)` resolves the referenced Blueprint with the Config's scalar values, applies the Config's repo / secret slot fills, stamps the deterministic `StableName(cfg.Metadata.Name)`, and attaches the `kukeon.io/config=<name>` back-reference label.
- Wires `-c` into the run mutex group alongside `-f`/`-p`/`-b`, with `--param`/`--param-file`/`--name` rejected on `-c` (a Config carries its own values and a stable identity).
- Drops the `(#625)` markers from `-p`'s deprecation hint and the `cellblueprint.MaterializeWithName` structural-slot error now that the verb is shipping.

## Identity state machine

After materialisation, `runRun` calls `client.GetCell(cellDoc)` against the deterministic name:

- **No live cell** → `CreateCell` + attach (no-op when `-d`).
- **Live cell, Ready** → no-op summary, attach.
- **Live cell, Stopped** → `StartCell` + attach.
- **Live cell, Pending / Failed / Unknown** → refuse with `kuke delete cell <name>` pointer.
- **Live cell whose spec diverges from `Materialize(cfg, bp)`** → emit the AC's one-line warning to stderr and attach to live state; do **not** refuse or reconcile.

The Ready/Stopped/error transitions reuse `runExistingCell` so the `-f` and `-c` paths share the same liveness-guard (#683) and CNI re-entry (#630) semantics. The `-c` branch differs only at the divergence check: `-f` refuses with a `kuke apply -f` pointer; `-c` warns and continues (per the AC).

## Non-obvious calls (reviewer push-back welcome)

- **Lookup scope flags.** The issue body sketches `--scope realm=<r>` but the AC's binding constraints don't require that syntax. `-c` reuses the existing `--realm`/`--space`/`--stack` flags (consistent with `-b`'s lookup-scope semantics via `explicitScope`); the AC's "joins the run mutual-exclusivity / one-required flag group" wording is satisfied.
[...]

PR #772, opened 2026-05-23
feat(apply): add kuke apply -b/-c for Blueprint/Config reconciliation

### LINKED ISSUE #752: apply: support kuke apply -b/-c for daemon-stored Blueprint/Config reconciliation
## Context

`kuke apply` today is `-f`-only (`cmd/kuke/apply/apply.go` accepts only the `--file` flag and reads a YAML doc from disk or stdin). `kuke run` already supports daemon-stored kinds via `-b <blueprint>` and `-c <config>` (PRs #702, #742). The asymmetry leaves operators with no `apply` shape for the daemon-stored kinds — divergent live state vs. a stored Blueprint/Config can only be reconciled today by exporting the Blueprint to YAML and re-applying with `-f`, or by manual `kuke delete cell` + `kuke run -c` (which loses overlay).

The `-c` warn-and-attach behavior on divergent specs (`cmd/kuke/run/run.go:310`, pinned by `run_test.go:3032`) currently emits a "use `kuke apply -c <name>` to reconcile" pointer that doesn't yet correspond to a real verb. This issue lands that verb.

This is the substrate the divergent-reject behavior change on `kuke run -b/-c --name` (separate issue) and AgentRun's `onDivergentSpec: sync` mode (kukeon#503) both delegate to.

## Proposal

Extend `kuke apply` to accept daemon-stored kinds by ref:

```
kuke apply -b <blueprint> [--name <cell>] [--param K=V] [--param-file <file>] [--realm R] [--space S] [--stack S]
kuke apply -c <config>    [--name <cell>] [--realm R] [--space S] [--stack S]
```

Flag handling mirrors `kuke run`:
- `-b`: accepts `--name`, `--param`, `--param-file`. Without `--name`, materializes a fresh `<prefix>-<6hex>` name (same convention as `kuke run -b`).
- `-c`: accepts `--name` only on the named-cell path. Rejects `--param` / `--param-file` (edit the Config's `spec.values` instead — same as `kuke run -c`).
- `-f`: unchanged.
- The three sources are mutually exclusive (cobra mutex), matching `run`.

### Reconciliation semantics

For the cell named (per `--name`, `StableName(config)`, or `<prefix>-<6hex>`):

| State | `apply` behavior |
|---|---|
| Missing | Create. |
| Exists, spec matches | No-op. |
| Exists, spec differs | **Stop → update → start**. No attach. |

No attach loop. `apply` is pure reconciliation; the operator runs `kuke attach <cell>` separately if they want a session. This matches `kuke apply -f`'s current shape (also no attach).

### Spec comparison

Compare the materialized `CellDoc` (post-Blueprint resolution + Config slot-fill) against the live `CellDoc` snapshot the daemon stores per cell. Diff at the resolved-spec level, not the source-Blueprint level — equivalent specs from different routes still compare equal.

### Error paths

- Blueprint/Config not found in scope → `errdefs.ErrBlueprintNotFound` / `errdefs.ErrConfigNotFound`, as `run` already returns.
- `--param` required by the Blueprint and not supplied → existing `cellblueprint.Resolve` error.
- Cell exists but belongs to a different lineage (label mismatch) → refuse with a clear error pointing at the existing cell's source. Avoids "the cron silently took over Alice's hand-built cell."

## Acceptance criteria

- [ ] `kuke apply -b <bp> --name myCell --param K=V` against a missing cell creates it; against an identical live cell is a no-op; against a divergent live cell stops, updates, starts; surfaces the diff in the daemon log.
- [ ] `kuke apply -c <config>` (no `--name`) reconciles the `StableName(config)` cell with the same semantics.
- [ ] `kuke apply -c <config> --name myCell` reconciles the named cell (named-cell mode); rejects `--param` / `--param-file` with the existing error shape.
- [ ] Lineage-label mismatch is refused with an error naming the existing cell's source.
- [ ] No attach loop — `apply` returns once the cell is reconciled.
- [ ] `cmd/kuke/apply/apply_test.go` covers the four cell-state transitions (missing / matches / differs / lineage-mismatch) for both `-b` and `-c`.

## Notes

- Substrate for kukeon#503 (AgentRun's `onDivergentSpec: sync` mode) and for the upcoming `kuke run -b/-c --name` divergent-reject behavior change.
- The "use `kuke apply -c <name>` to reconcile" error string in `cmd/kuke/run/run.go:310` becomes a real pointer once this lands.
- Out of scope: making `apply -f` accept Blueprint/Config kinds inline in the YAML body (already supported via existing kind-router; no change here). This issue is specifically about the `-b`/`-c` *ref* shape on `apply`.

eminwux:
## Summary

- Extends `kuke apply` to accept daemon-stored Blueprint (`-b`) and Config (`-c`) refs, the apply analogues of `kuke run -b/-c`. The materialised CellDoc is round-tripped through the daemon's existing `ApplyDocuments` so the per-cell reconciliation (create on missing, no-op on match, daemon-side stop/update/start on divergence) reuses the path `kuke apply -f` already drives.
- Adds the lineage-mismatch refusal the issue calls out: before applying, `assertCellLineage` reads the live cell and refuses when its `kukeon.io/blueprint` / `kukeon.io/config` label does not match the named source (hand-built cells or sibling-source cells are surfaced explicitly so the cron never silently overwrites Alice's work).
- Mirrors `kuke run`'s flag grammar: `-f` / `-b` / `-c` are mutually exclusive and exactly one is required; `--name`, `--param`, `--param-file` are rejected with `-f`; `--param`, `--param-file` are rejected with `-c` (Config owns its values); `--realm`, `--space`, `--stack` constrain the Blueprint/Config lookup.
- No attach loop: `apply` is pure reconciliation. Matches `apply -f`'s existing shape.

## Notes for the reviewer

- The `-f` path is behaviourally unchanged. The cobra-error wording for "no source" shifts from `required flag` to `at least one of the flags in the group [file blueprint config] is required` (the `MarkFlagsOneRequired` form replaces `MarkFlagRequired("file")`); the existing test for that case was updated to match.
- I duplicated the Blueprint/Config loaders that already live in `cmd/kuke/run/run.go` (`loadFromBlueprint`, `loadFromConfig`, `buildParamMap`, `pickRealm` / `explicitScope`, `resolveCellLocation`) rather than extracting them into a shared package — the run-side helpers are unexported and lifting them into a new shared package felt like scope creep against the issue's "extend `kuke apply`" framing. The duplication is small (~90 lines) and the next phase that touches `kuke run -b/-c --name` divergent-reject behaviour (issue body's notes) is a natural moment to factor them out.
- Reconciliation work is delegated to `client.ApplyDocuments` on the marshalled CellDoc, so the daemon-side `ReconcileCell` / `UpdateCell` path owns the per-container Stop → recreate → Start sequence already exercised by `apply -f`. The CLI side only adds the lineage gate (which the daemon doesn't enforce).
- `docs/site/cli/kuke-apply.md` does not yet mention `-b`/`-c`. Following the pattern of PR #750 (docs for `run -b/-c` landed separately after the feature in #742), I've left the doc update for a follow-up PR; happy to bundle if preferred.

## Test plan

- [x] `GOWORK=off go test ./cmd/kuke/apply/...` (covers the seven new -b/-c cases plus the four existing -f cases)
- [x] `make test` (full root + kukebuild suite; all green)
- [x] `pre-commit run --files cmd/kuke/apply/apply.go cmd/kuke/apply/apply_test.go`
- [ ] `make dev-init` smoke not run — change is CLI-only (no daemon, build path, or `/opt/kukeon` reads touched), so the regression guard the smoke catches does not apply here.

Closes #752

### DIFF
--- a/cmd/kuke/apply/apply.go
+++ b/cmd/kuke/apply/apply.go
@@ -17,75 +17,465 @@
 package apply
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
+	"os"
+	"strings"
 
+	"github.com/eminwux/kukeon/cmd/config"
 	kukshared "github.com/eminwux/kukeon/cmd/kuke/shared"
+	"github.com/eminwux/kukeon/internal/cellblueprint"
+	"github.com/eminwux/kukeon/internal/cellconfig"
+	"github.com/eminwux/kukeon/internal/cellprofile"
 	"github.com/eminwux/kukeon/internal/errdefs"
 	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
 )
 
 // MockControllerKey is used to inject a mock kukeonv1.Client via context in tests.
 type MockControllerKey struct{}
 
+const (
+	outputFormatJSON = "json"
+	outputFormatYAML = "yaml"
+)
+
+// NewApplyCmd builds the `kuke apply` cobra command. `-f` reads a multi-document
+// YAML stream from disk or stdin (unchanged). `-b` resolves a daemon-stored
+// CellBlueprint by name + scope, substitutes `--param` values, and reconciles
+// the resulting cell against the live state. `-c` resolves a daemon-stored
+// CellConfig by name + scope and reconciles its materialised cell. The three
+// sources are mutually exclusive; exactly one is required.
 func NewApplyCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:           "apply -f <file>",
-		Short:         "Apply resource definitions from YAML file",
-		Long:          "Apply resource definitions from a YAML file or stdin. Supports multi-document YAML separated by '---'.",
+		Use:   "apply (-f <file> | -b <blueprint> | -c <config>)",
+		Short: "Apply resource definitions from YAML or reconcile a daemon-stored Blueprint/Config",
+		Long: "Apply resource definitions from a YAML file or stdin (-f), reconcile a " +
+			"daemon-stored CellBlueprint by ref (-b), or reconcile a daemon-stored " +
+			"CellConfig by ref (-c). The -b/-c forms materialise a CellDoc the same way " +
+			"`kuke run -b/-c` does and reconcile it against the live cell: a missing cell " +
+			"is created and started; an identical live cell is a no-op; a divergent live " +
+			"cell is stopped, updated, and started (no attach). A live cell whose " +
+			"lineage label (kukeon.io/blueprint or kukeon.io/config) does not match the " +
+			"-b/-c source is refused, so apply never silently takes over an unrelated cell.",
 		Args:          cobra.NoArgs,
 		SilenceUsage:  true,
 		SilenceErrors: false,
-		RunE: func(cmd *cobra.Command, _ []string) error {
-			file, err := cmd.Flags().GetString("file")
-			if err != nil {
-				return err
-			}
-			if file == "" {
-				return errors.New("file flag is required (use -f <file> or -f - for stdin)")
-			}
+		RunE:          runApply,
+	}
 
-			output, err := cmd.Flags().GetString("output")
-			if err != nil {
-				return err
-			}
+	cmd.Flags().StringP("file", "f", "", "File to read YAML from (use - for stdin); mutually exclusive with -b/-c")
+	cmd.Flags().StringP("output", "o", "", "Output format: json, yaml (default: human-readable)")
 
-			reader, cleanup, err := kukshared.ReadFileOrStdin(file)
-			if err != nil {
-				return err
-			}
-			defer func() { _ = cleanup() }()
+	cmd.Flags().StringP("blueprint", "b", "",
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #772 Review — `feat(apply): add kuke apply -b/-c for Blueprint/Config reconciliation`**

LGTM. The lineage gate (`assertCellLineage`) is the right shape — hand-built and sibling-source cells are refused with a specific error rather than silently overwritten, missing-cell skips correctly. Materialisation reuses `client.ApplyDocuments` so the daemon-side reconciler owns the per-container Stop → recreate → Start sequence. Tests cover the 3×2 matrix (missing/matching/divergent × -b/-c) plus lookup-miss and lineage-refusal.

One non-blocking observation:

- **Scope-resolution asymmetry with `kuke run`** `cmd/kuke/apply/apply.go:405-422`: `pickRealm` / `explicitScope` here only read the cobra flag, while `cmd/kuke/run/run.go:776-784` fall back to `os.LookupEnv(kv.Key)` (KUKE_RUN_SPACE/_STACK) and `viper.GetString` / `config.KUKE_RUN_REALM`. Operators who rely on env-var scope for `kuke run -b/-c` will find `kuke apply -b/-c` ignores the same env. The planned helper extraction (per your PR description) is the natural moment to unify — flagging so it isn't lost.

Filing the `docs/site/cli/kuke-apply.md` gap as a followup-nit per the PR #750 precedent you cited.

PR #802, opened 2026-05-23
feat(run): reject divergent live cell on -b --name and -c paths

eminwux:
## Summary

- `kuke run -c <config>` against a divergent live cell now refuses to attach with a `kuke apply -c <config>` reconcile pointer instead of warning and silently attaching to the diverged state. Brings `-c` into line with the `-f` discipline that already keeps `run` a pure read/materialize verb.
- `kuke run -b <bp> --name <cell>` adds the same divergent-spec gate against the pinned name, pointing at `kuke apply -b <bp> --name <cell>`. The generated-name `-b` path is untouched because each invocation materialises a fresh `<prefix>-<6hex>` cell — no collision possible.
- Error format mirrors the existing `-f` rejection text: names the cell, names the source (CellConfig / CellBlueprint), enumerates the diverging field(s), and hands back the exact reconcile invocation. `-c`'s pointer omits `--name` because the cell name *is* the Config's stable name (`apply -c` recomputes it).
- The `-f` / `-p` divergent path continues to point at `kuke apply -f` (deprecated `-p` doesn't get the new behaviour per the issue's explicit carve-out).
- Updates `cmd/kuke/run/run.go`'s long help text and `docs/site/cli/kuke-run.md`'s identity-state-machine section to describe the new reject behaviour.

## Test plan

- [x] `GOWORK=off go vet ./cmd/kuke/run/...`
- [x] `make test` (full repo + `cmd/kukebuild` module)
- [x] `pre-commit run --files cmd/kuke/run/run.go cmd/kuke/run/run_test.go docs/site/cli/kuke-run.md` — go-fmt, golangci-lint, prettier, addlicense all green
- [x] Renamed test `TestRun_FromConfig_DivergentSpec_RefusesAndPointsToApply` asserts the new reject path (was `_WarnsAndAttaches`) — checks substring of cell name, source name, diverging field, `refusing to attach`, `kuke apply -c prod`, `reconcile`; asserts `CreateCell=0` + `StartCell=0`.
- [x] New `TestRun_FromBlueprint_NamedDivergent_RefusesAndPointsToApply` covers the `-b --name` divergent case with the same substring shape (pointer is `kuke apply -b web --name pinned`).
- [x] Existing `-c` happy-path tests (`LiveReadyCell_AttachesWithoutCreate`, `LiveStoppedCell_StartsThenAttaches`, `LiveFailedCell_RefusesWithDeletePointer`) still pass — matching-spec attach is unchanged.
[...]

Issue #753, opened 2026-05-23
run: reject divergent live cell on -b --name and -c, point at apply -b/-c


eminwux:
## Context

`kuke run -f file.yaml` against a live cell whose spec has diverged from the file refuses the run and points the operator at `kuke apply -f` to reconcile (`cmd/kuke/run/run.go:75`). This makes `run` a pure read/materialize verb and routes destructive updates through `apply`.

The `-b` and `-c` paths don't yet have this discipline:

- **`kuke run -c <config>`** on a divergent live cell *warns and attaches to live state*, no update (`cmd/kuke/run/run.go:310`, pinned by `run_test.go:3032`). The warning string already points at `kuke apply -c <name>` to reconcile — but until #752 lands, that pointer doesn't correspond to a real verb, and even after, the path still silently attaches to the diverged state instead of refusing.
- **`kuke run -b <bp> --name myCell`** has no explicit divergent-spec check. `MaterializeWithName` (`internal/cellblueprint/cellblueprint.go:160`) stamps the name and returns; the downstream create-or-attach machinery handles the collision without diffing.

Both should reject divergent named cells the same way `-f` does. The reconciliation path is `kuke apply -b/-c --name myCell` (#752).

### `kuke run -b/-c` on a divergent live named cell → reject

For `kuke run -c <config>` and `kuke run -b <bp> --name myCell`, when the named cell exists with a spec that differs from the materialized doc:

```
live cell "myCell" spec differs from CellConfig "dev-claude" — refusing to attach;
use `kuke apply -c dev-claude --name myCell` to reconcile (stops, updates, starts)
```

Error format mirrors the existing `-f` rejection text. `-b` variant points at `kuke apply -b <bp> --name <cell>` instead.

### Matching specs → attach as today

When the spec matches, behavior is unchanged: idempotent attach (`-c` StableName path), or attach to the named cell (`-b --name` path). Only divergence triggers the reject.

### Generated-name path is unaffected

`kuke run -b <bp>` (no `--name`) and `kuke run -c <config> --generate-name` (separate issue) materialize fresh `<prefix>-<6hex>` cells — no name collision possible, no divergent-spec check needed.

### Test fixture update

`cmd/kuke/run/run_test.go:3032` currently pins the warn-and-attach behavior:

```go
want := "live cell spec differs from CellConfig prod; attaching to live state — kuke apply -c prod to reconcile"
```

Updates to assert the reject + non-attach + reconcile pointer pattern instead.

[...]

PR #805, opened 2026-05-23
feat(cli): unify kuke apply -b/-c scope resolution with kuke run via shared helpers

### LINKED ISSUE #776: cli(apply): make kuke apply -b/-c scope resolution symmetrical with kuke run -b/-c (env vars)
Surfaced during review of #772.

`kuke apply -b/-c` resolves `--realm`/`--space`/`--stack` strictly from the cobra flags, ignoring the env-var fallbacks that `kuke run -b/-c` honors. Operators who set `KUKE_RUN_REALM` / `KUKE_RUN_SPACE` / `KUKE_RUN_STACK` (or the equivalent viper defaults from `config.KUKE_RUN_*`) for their `kuke run` workflow will find `kuke apply -b/-c` does not pick up the same scope and falls back to `realm=default, space="", stack=""`.

**Apply side** (`cmd/kuke/apply/apply.go:405-422`):

```go
func pickRealm(cmd *cobra.Command) string {
    if v, _ := cmd.Flags().GetString("realm"); strings.TrimSpace(v) != "" {
        return strings.TrimSpace(v)
    }
    return "default"
}

func explicitScope(cmd *cobra.Command, flagName string) string {
    if !cmd.Flags().Changed(flagName) {
        return ""
    }
    v, _ := cmd.Flags().GetString(flagName)
    return strings.TrimSpace(v)
}
```

**Run side** (`cmd/kuke/run/run.go:763-795`):

```go
func resolveCellLocation(doc *v1beta1.CellDoc) {
    doc.Spec.RealmID = pickLocation(doc.Spec.RealmID, &config.KUKE_RUN_REALM)
    doc.Spec.SpaceID = pickLocation(doc.Spec.SpaceID, &config.KUKE_RUN_SPACE)
    doc.Spec.StackID = pickLocation(doc.Spec.StackID, &config.KUKE_RUN_STACK)
}

func explicitScope(cmd *cobra.Command, flagName string, kv *config.Var) string {
    if cmd.Flags().Changed(flagName) {
        v, _ := cmd.Flags().GetString(flagName)
        return strings.TrimSpace(v)
    }
    if v, ok := os.LookupEnv(kv.Key); ok {
        return strings.TrimSpace(v)
    }
    return ""
}

func pickLocation(fromDoc string, kv *config.Var) string {
    if v := strings.TrimSpace(fromDoc); v != "" {
        return v
    }
    if v := strings.TrimSpace(viper.GetString(kv.ViperKey)); v != "" {
        return v
    }
    return strings.TrimSpace(kv.ValueOrDefault())
}
```

**Suggested fix.** The PR #772 author flagged the duplication explicitly in the PR description and named "the next phase that touches `kuke run -b/-c --name` divergent-reject behaviour" as the natural moment to extract the helpers into a shared package. That extraction is the right place to unify scope resolution as well — both commands should share one `pickRealm` / `explicitScope` / `resolveCellLocation` set, reading from cobra flag → env var (`KUKE_RUN_*` or a renamed `KUKE_*` umbrella) → viper default → hard "default" for realm, in that precedence. Either pull the run-side `pickLocation` / `explicitScope(cmd, name, kv)` shape into the new shared package, or define a CLI-wide scope-resolution helper that both subcommands consume.

If a config-var rename is desired so the env keys aren't `kuke run`-flavored when called from `kuke apply`, that's a separate decision (and a backwards-compat consideration) but worth noting at the extraction moment.

Acceptance:

- `KUKE_RUN_REALM=foo kuke apply -b web` resolves the Blueprint from realm `foo`, matching the behavior of `KUKE_RUN_REALM=foo kuke run -b web`.
- Same for `KUKE_RUN_SPACE` / `KUKE_RUN_STACK`.
- Test coverage for both subcommands lives in one place after the helper extraction.

eminwux:
## Summary

- `kuke apply -b/-c` now honors `KUKE_RUN_REALM` / `KUKE_RUN_SPACE` / `KUKE_RUN_STACK` env-var fallbacks, matching `kuke run -b/-c`. The pre-fix apply path read scope strictly from cobra flags, so an operator who pinned `KUKE_RUN_REALM=foo` for their `kuke run` workflow found `kuke apply -b web` silently looked up in `realm=default` instead.
- Extracts the scope-resolution helpers into `cmd/kuke/shared/scope.go` (`PickLookupRealm`, `ExplicitScope`) and points both `kuke apply -b/-c` and `kuke run -b/-c` at them. One precedence chain, one source of truth, one test surface — addresses the AC's "test coverage for both subcommands lives in one place after the helper extraction."
- Precedence: `--realm`/`--space`/`--stack` flag → `KUKE_RUN_*` env var → `"default"` (realm) / `""` (space, stack — `""` keeps realm-scoped Blueprints/Configs findable). Both helpers bypass viper deliberately; the inline godoc explains why (the existing run-side `explicitScope` comment surfaced the same `SetDefault`-trips-`IsSet` reason).
- `run.go`'s local `explicitScope` is deleted (replaced by the shared helper). `run.go`'s local `pickLocation` stays because the **fill** path on the materialised cell doc still wants the viper-backed session pin — the lookup path is the only one that needs to bypass viper.
- `apply.go`'s `resolveCellLocation` now also picks up the env-var fallbacks via the shared helpers, so the materialised cell's `Spec.RealmID`/`SpaceID`/`StackID` matches the lookup scope when the doc itself left them empty.

### Scope notes

- The issue's "Suggested fix" floated waiting on the divergent-reject phase (#753 → PR #802) before extracting helpers. That PR is still in flight; this change extracts the narrow lookup helpers now so the env-var symmetry isn't blocked on it. The divergent-reject phase can consume the same `cmd/kuke/shared` namespace when it lands — no helper-shape collision.
- Did not rename `KUKE_RUN_*` to a CLI-wide umbrella (`KUKE_*` or similar). The issue body flagged it as a separate decision with backwards-compat implications and explicitly out of scope here. Operators continue to set `KUKE_RUN_REALM` and both subcommands consume it.

## Test plan

- [x] `make test` — full repo + `cmd/kukebuild` module pass.
- [x] `GOWORK=off go build ./...` — clean compile.
- [x] `pre-commit run --files <changed>` — go fmt / golangci-lint / addlicense all pass. golangci-lint rewrote `scope_test.go` to wrap long table-driven test cases; re-staged and re-ran clean.
- [x] New `cmd/kuke/shared/scope_test.go` covers `PickLookupRealm` and `ExplicitScope` precedence (no-flag/no-env → default; env only; flag only; flag wins over env; whitespace trim).
- [x] New `cmd/kuke/apply/apply_test.go` regressions: `TestApply_FromBlueprint_EnvVarScopeFallback` and `TestApply_FromConfig_EnvVarScopeFallback` verify `KUKE_RUN_REALM`/`SPACE`/`STACK` feed the BP / Config lookup; `TestApply_FromBlueprint_FlagWinsOverEnvVar` pins the flag-over-env precedence.
- [x] Existing `cmd/kuke/run` tests still pass — the run-side lookup wiring swap is behavior-preserving (the new shared helper returns the same flag → env → default chain the prior `pickLocation("", &kv)` + `explicitScope` pair produced).
- [ ] `make dev-init` smoke — not run; this change is scoped to a CLI-side helper extraction (cmd/kuke only), with no build path, daemon, or `/opt/kukeon` touch points. The `kuke status` parity guard would only be re-confirming an untouched code path.

Closes #776

### DIFF
--- a/cmd/kuke/apply/apply.go
+++ b/cmd/kuke/apply/apply.go
@@ -286,9 +286,9 @@ func loadFromBlueprint(cmd *cobra.Command, client kukeonv1.Client, flags applyFl
 	lookup := v1beta1.CellBlueprintDoc{
 		Metadata: v1beta1.CellBlueprintMetadata{
 			Name:  flags.blueprintName,
-			Realm: pickRealm(cmd),
-			Space: explicitScope(cmd, "space"),
-			Stack: explicitScope(cmd, "stack"),
+			Realm: kukshared.PickLookupRealm(cmd, &config.KUKE_RUN_REALM),
+			Space: kukshared.ExplicitScope(cmd, "space", &config.KUKE_RUN_SPACE),
+			Stack: kukshared.ExplicitScope(cmd, "stack", &config.KUKE_RUN_STACK),
 		},
 	}
 
@@ -317,9 +317,9 @@ func loadFromConfig(cmd *cobra.Command, client kukeonv1.Client, flags applyFlags
 		Kind:       v1beta1.KindCellConfig,
 		Metadata: v1beta1.CellConfigMetadata{
 			Name:  flags.configName,
-			Realm: pickRealm(cmd),
-			Space: explicitScope(cmd, "space"),
-			Stack: explicitScope(cmd, "stack"),
+			Realm: kukshared.PickLookupRealm(cmd, &config.KUKE_RUN_REALM),
+			Space: kukshared.ExplicitScope(cmd, "space", &config.KUKE_RUN_SPACE),
+			Stack: kukshared.ExplicitScope(cmd, "stack", &config.KUKE_RUN_STACK),
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


eminwux:
**PR #805 Review — `feat(cli): unify kuke apply -b/-c scope resolution with kuke run via shared helpers`**

LGTM — narrow, well-scoped extraction: `kuke apply -b/-c` now honors `KUKE_RUN_REALM/SPACE/STACK` via the same `cmd/kuke/shared` helpers `kuke run` uses, deleted the duplicated locals, kept `pickLocation` for the materialised-cell fill path with a godoc that explains why, and the test surface covers both the helpers in isolation and the apply integration (flag-only / env-only / flag-wins-over-env). `make dev-init` carve-out is appropriate — pure CLI helper extraction with no daemon or `/opt/kukeon` touch points.

PR #832, opened 2026-05-23
refactor(naming): extract shared <prefix>-<6hex> random-suffix helper

### LINKED ISSUE #808: cellconfig/cellblueprint/cellprofile: extract shared <prefix>-<6hex> random-suffix helper
Non-blocking nit surfaced during review of #803.

After #803, three packages now carry the same 3-byte / `crypto/rand` / `hex.EncodeToString` machinery for the `<prefix>-<6hex>` cell-name suffix shared by `-b`/`-p`/`-c -g`:

- `internal/cellblueprint/cellblueprint.go:45` (`const nameSuffixBytes = 3`) and `:278` (`func randomHexSuffix() (string, error)`)
- `internal/cellprofile/cellprofile.go:45` (`const nameSuffixBytes = 3`) and `:278` (`func randomHexSuffix() (string, error)`)
- `internal/cellconfig/materialize.go:34` (`const nameSuffixBytes = 3`) and `:284` (`func GenerateName(configName string) (string, error)` — inlines the same `rand.Read` + `hex.EncodeToString` body)

Three independent copies of the same entropy contract: change the suffix width, you have to remember to change it in all three; same for swapping the encoding or the RNG source.

Suggested fix: extract a small helper to e.g. `internal/util/naming` (or wherever fits the layering) exposing `RandomHexSuffix(bytes int) (string, error)` or `MakeSuffixed(prefix string) (string, error)`, and have the three sites call it. The cellblueprint and cellprofile private `randomHexSuffix` helpers become thin call-throughs; the exported `cellconfig.GenerateName` becomes a thin wrapper too. PR #803's `nameSuffixBytes = 3` stays the canonical width.

#803 followed precedent correctly (the prior two copies pre-dated it); the cleanup is at the codebase scope, not at PR #803's scope.

eminwux:
## Summary

- Three packages — `internal/cellblueprint`, `internal/cellprofile`, `internal/cellconfig` — each carried an independent copy of the `3-byte / crypto/rand / hex.EncodeToString` machinery for the `<prefix>-<6hex>` cell-name suffix used by `kuke run -b`, `-p`, and `-c -g`. Three places to change the suffix width, the encoding, or the RNG source.
- Adds `internal/util/naming.RandomHexSuffix(n int) (string, error)` and `naming.DefaultCellNameSuffixBytes = 3`; rewires the three call sites to use it. The canonical width and the per-site error-wrapping context are unchanged — each caller still wraps with its own document-flavored message ("blueprint %q", "profile %q", "config %q").
- The private `randomHexSuffix` helpers in cellblueprint/cellprofile were only used in a single call site each, so they're inlined to the call rather than kept as thin re-export wrappers — fewer hops to read.

## Test plan

- [x] `GOWORK=off go test ./internal/util/naming/ ./internal/cellblueprint/ ./internal/cellprofile/ ./internal/cellconfig/` (new `randomsuffix_test.go`: length+alphabet table, 256-sample uniqueness, canonical-width pin)
- [x] `make test`
- [x] `pre-commit run --files <staged>` (golangci-lint auto-applied a `range samples` rewrite + one Errorf line wrap in the new test; both re-staged)

Closes #808

### DIFF
--- a/internal/cellblueprint/cellblueprint.go
+++ b/internal/cellblueprint/cellblueprint.go
@@ -25,13 +25,12 @@
 package cellblueprint
 
 import (
-	"crypto/rand"
-	"encoding/hex"
 	"fmt"
 	"strings"
 
 	"github.com/eminwux/kukeon/internal/cellprofile"
 	"github.com/eminwux/kukeon/internal/errdefs"
+	"github.com/eminwux/kukeon/internal/util/naming"
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 	"gopkg.in/yaml.v3"
 )
@@ -41,10 +40,6 @@ import (
 // every cell produced by `kuke run -b` so operators can list all instances
 // with `kuke get cells -l kukeon.io/blueprint=<name>`.
 const LabelBlueprint = "kukeon.io/blueprint"
-
-// nameSuffixBytes is the entropy width for the `<prefix>-<6hex>` suffix —
-// 3 bytes → 6 lowercase hex chars, matching cellprofile.
-const nameSuffixBytes = 3
 
 // Resolve substitutes `${KEY}` scalar parameters in the blueprint body against
 // the resolution order cliParams[k] > parameters[k].default > lookupEnv(k),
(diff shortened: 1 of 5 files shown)


eminwux:
**PR #832 Review — `refactor(naming): extract shared <prefix>-<6hex> random-suffix helper`**

LGTM — three duplicate `crypto/rand`+`hex.EncodeToString` helpers consolidated into one `internal/util/naming` package with no behavior change; canonical 3-byte width pinned by a dedicated test, and the per-site error wrapping is preserved verbatim.

PR #830, opened 2026-05-23
feat(daemon): detect OutOfSync state for Config-lineage cells in reconciler loop

eminwux:
## Summary

- Foundation phase of #819's umbrella: surface a persistent OutOfSync flag in cell status by re-deriving the would-be cell from the daemon-stored Config + Blueprint on every reconcile tick and diffing against the live spec. Phase 2 (`kuke restart cell` #821), phase 3 (SYNC column in `kuke get cell` #822), and phase 4 (hard-remove `kuke apply -b/-c` #823) all depend on this status field landing.
- Adds three fields to `CellStatus` (both `internal/modelhub` and `pkg/api/model/v1beta1`, round-tripped in `apischeme`): `OutOfSync bool`, `OutOfSyncReason string`, `OutOfSyncError string`. The third field is a distinct surface for undecidable diffs (referenced Blueprint missing, materialization error) so a downstream `kuke restart cell --on-out-of-sync` can route a routine drift differently from an error state per the AC.
- Detection runs as a per-cell post-pass in `controller.ReconcileCells` after the existing lifecycle `runner.ReconcileCell` returns, gated on the `kukeon.io/config` lineage label so non-Config cells are skipped (`-b`-lineage and hand-built cells are out of v1 scope per the umbrella). Reuses the same `cellconfig.Materialize` + `apply.DiffCell` pipeline `kuke apply -c`'s CLI-side check uses today — verified against `cmd/kuke/apply/apply.go:358` and `internal/controller/apply/diff.go:263` during the sanity-check pass.
- No-write fast-path on the persistence step: a Synced fleet pays at most one Config + one Blueprint read per tick (no metadata write); a divergent cell whose status already records the same reason also skips the write. Covers the AC's "performance: extra Config + Blueprint lookup per Config-lineage cell per reconcile tick, negligible at O(10) cells" note.

## Design call surfaced for reviewer pushback

The AC pinned the new tests to `internal/controller/reconcile_test.go`, but the codebase convention is one `reconcile_*_test.go` per reconcile pass (see `reconcile_config_test.go`, `reconcile_blueprint_test.go`, `reconcile_secret_test.go`). I added the 5 AC scenarios to a new `reconcile_outofsync_test.go` co-located with the new helper, plus a sixth no-write fast-path test guarding the diff-the-status side of the persistence predicate. Happy to inline them into `reconcile_test.go` if the reviewer prefers literal AC adherence over the per-pass convention.

## Test plan

- [x] `make test` — full suite green
[...]

PR #838, opened 2026-05-23
feat(cli): add kuke restart cell with implicit reconcile on OutOfSync (step 2)

### LINKED ISSUE #821: feat(cli): add kuke restart cell <name> with implicit reconcile on OutOfSync (step 2)
## Context

Child of #819. With OutOfSync detection in place (sibling child issue, foundation phase), operators need a verb that:

1. Bounces a cell (stop + start with the same spec) on demand — today there's no atomic restart; operators do `kuke stop && kuke start`
2. Picks up the reconcile path when the cell is OutOfSync — replaces today's `kuke apply -c` workflow

A single `kuke restart cell` verb covers both. Semantics: always restart; if OutOfSync, the "start" step uses the freshly-materialised spec instead of the old one (so the restart is also a reconcile).

## Proposal

Add `kuke restart cell <name>` (mirroring `kuke start cell` / `kuke stop cell` shape):

- Positional `<name>`; scope flags `--realm`/`--space`/`--stack`
- **On a Synced cell**: stop the running task, start with the same spec — vanilla restart
- **On an OutOfSync cell**: stop the running task, update the cell spec to the current Config materialisation, start with the new spec — implicit reconcile
- **On a Stopped cell**: equivalent to `kuke start cell <name>` (already-stopped restart = start)
- **On an error/partial-state cell**: refuse with a useful error pointing at `kuke delete cell <name>` (same shape as `kuke run -f`'s error/partial handling)
- **Severs any active attach session** — operator typed restart, predictable consequence
- No `--detach`/`--attach` flags — restart is a lifecycle verb, not an attach-flow verb

## Acceptance criteria

- [ ] New subcommand `cmd/kuke/restart/cell/cell.go` registered under a new parent `cmd/kuke/restart/restart.go` (mirror `cmd/kuke/start/` structure)
- [ ] `kuke restart cell <name>` restarts the named cell in the resolved scope
- [ ] Synced cell restart: stop + start with the same spec; preserves all spec fields
- [ ] OutOfSync cell restart: stop + update (re-materialise from Config) + start with the new spec; implicit reconcile
- [ ] Stopped cell restart: equivalent to `kuke start cell <name>`
- [ ] Error/partial-state cell restart: refuses with a useful error
- [ ] Any active attach session on the cell is severed when restart fires
- [ ] Scope flags `--realm`/`--space`/`--stack` work consistent with `kuke start cell`
- [ ] Shell completion: cell names in scope
- [ ] Help text describes the dual mode (vanilla restart + implicit reconcile on OutOfSync)
- [ ] Tests cover: Synced restart preserves spec; OutOfSync restart picks up the new spec; Stopped restart starts; error-state cell refused; attach severed

## Notes

- Umbrella: #819
- **Depends on**: the OutOfSync detection issue (sibling child under #819) — restart's reconcile-on-OutOfSync behavior needs the status field to read from
- Pattern reference: `cmd/kuke/start/cell/cell.go` for the verb structure (scope flags, completion, MockControllerKey for tests)
- **Open question for implementation**: the atomic stop + update + start path may need a new daemon RPC, or can be composed at CLI level from existing stop / update-spec / start RPCs. Verify which is cleaner during implementation; if a new RPC, scope might warrant a small predecessor issue, but the user-facing acceptance criteria don't change either way

eminwux:
## Summary

- New verb `kuke restart cell <name>` mirroring `kuke start cell` / `kuke stop cell`'s surface (positional name, `--realm/--space/--stack` scope flags, completion).
- Dual-mode behaviour:
  - **Synced Ready cell** → vanilla `StopCell + StartCell` with the same daemon-stored spec.
  - **OutOfSync Ready cell** → re-materialise from the lineage Config (`kukeon.io/config=<name>` label) via `GetConfig + GetBlueprint + cellconfig.Materialize`, route through `ApplyDocuments` to reconcile the spec, then explicitly `StopCell + StartCell` unless the reconcile already bounced every container (root-container recreate or from-Stopped rematerialize).
  - **Stopped cell** → equivalent to `kuke start cell <name>` (no reconcile, per AC).
  - **Pending/Failed/Unknown cell** → refuse with a `kuke delete cell <name>` pointer (same pattern as `cmd/kuke/run/run.go:505-516`).
- Implements the umbrella's step 2 (Closes #821), builds on the #830 OutOfSync detection foundation shipped in step 1.

## Design notes

- **CLI-level composition vs. new RPC.** The issue's open implementation question called out two options for the OutOfSync atomic stop+update+start: a new daemon RPC, or composing at CLI level over the existing primitives. I went with the CLI-level path (`GetConfig + GetBlueprint + Materialize + ApplyDocuments`) because no wire-format addition is needed. The AC's user-facing contract holds either way (per the issue's note).
- **Reconcile-no-bounce gap (addressed in fix-up 8ef685a).** The daemon's `ApplyDocuments` only bounces containers when image/command/args change (`internal/controller/runner/update_cell.go:containerSpecChanged`) — pure env-var or metadata-only divergence is patched in place by `UpdateCell` with no container bounce. To preserve `kuke restart`'s contract ("all running containers bounce"), the CLI now follows the `ApplyDocuments` call with an explicit `StopCell + StartCell` unless the reconcile already bounced every container (`reconcileBouncedAll` detects the full-bounce signals: `Action == "created"`, `"root container recreated"`, `"runtime stopped: containers re-materialized"`). The rare double-bounce in the already-bounced-everything case is acceptable; under-bouncing silently breaks the contract.
- **`OutOfSyncError` and "lineage Config deleted" reasons.** Neither case can drive a reconcile (divergence is undecidable or the source is gone). Both fall through to a vanilla restart with a one-line stderr notice so the cell's runtime still bounces as the operator asked and the operator sees the cause. Same fall-through for an OutOfSync cell that somehow lost its lineage label.
- **Attach severing** is a side effect of the daemon's existing `StopCell` behaviour (the container task is killed, which closes any attach session). The CLI does not need to do anything extra; both the vanilla and reconcile branches go through that stop step.
- **`ClientFromCmd` (not `DaemonClientFromCmd`)** matches the sibling lifecycle verbs `start`/`stop`; `local.Client` implements all the methods used here (`GetCell`, `StopCell`, `StartCell`, `GetConfig`, `GetBlueprint`, `ApplyDocuments`).
- The new help text describes the dual mode so operators see the contract without reading code (per AC: "Help text describes the dual mode").

## Test plan

- [x] `make test` — full root + kukebuild module test target passes (cached after first run).
- [x] `GOWORK=off go vet ./...` clean on root module.
- [x] `pre-commit run --files cmd/kuke/restart/cell/cell.go cmd/kuke/restart/cell/cell_test.go` clean (gofmt, golangci-lint, addlicense).
- [x] `./kuke restart --help` / `./kuke restart cell --help` render the dual-mode help text and the expected flags.
- [x] `cmd/kuke/restart/cell/cell_test.go` covers: Synced restart preserves spec (no `ApplyDocuments` call); OutOfSync restart reconciles then explicitly bounces (asserts `ApplyDocuments` *and* `StopCell + StartCell` called); **env-var-only OutOfSync divergence reconciles then explicitly bounces** (reviewer-requested coverage of the no-bounce gap — Apply returns `container "main" updated: env`, asserts the follow-up `StopCell + StartCell`); **root-container-recreate OutOfSync skips the redundant bounce** (Apply returns `root container recreated`, asserts `stopCalls == 0` and `startCalls == 0`); Stopped restart calls `StartCell` only; Pending/Failed cells refused with `kuke delete cell` pointer; missing cell returns `ErrCellNotFound`; missing realm refused; OutOfSync-without-lineage-label, OutOfSyncError, and deleted-Config edge cases each fall through to vanilla restart with a stderr notice; `ApplyDocuments` failed-row surfaces as error; `StopCell` error skips `StartCell`.
- [x] Attach-severing is exercised by the `StopCell`-was-called assertion in the Synced and reconcile-fallback paths (daemon owns the actual sever step).
- [ ] `make dev-init` host smoke — not run; this change is purely client-side (CLI verb composing existing RPCs), no daemon code or `/opt/kukeon` reads/writes change.

Closes #821

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -488,6 +488,14 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_LOG_FOLLOW = DefineKV("KUKE_LOG_FOLLOW", "kuke/log/follow", "false")
 
+	// Restart command variables
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_RESTART_CELL_REALM = DefineKV("KUKE_RESTART_CELL_REALM", "kuke/restart/cell/realm", "default")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_RESTART_CELL_SPACE = DefineKV("KUKE_RESTART_CELL_SPACE", "kuke/restart/cell/space", "default")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKE_RESTART_CELL_STACK = DefineKV("KUKE_RESTART_CELL_STACK", "kuke/restart/cell/stack", "default")
+
 	// Stop command variables
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_STOP_CELL_REALM = DefineKV("KUKE_STOP_CELL_REALM", "kuke/stop/cell/realm", "default")
--- a/cmd/kuke/kuke.go
+++ b/cmd/kuke/kuke.go
@@ -39,6 +39,7 @@ import (
 	logcmd "github.com/eminwux/kukeon/cmd/kuke/log"
 	purgecmd "github.com/eminwux/kukeon/cmd/kuke/purge"
 	refreshcmd "github.com/eminwux/kukeon/cmd/kuke/refresh"
+	restartcmd "github.com/eminwux/kukeon/cmd/kuke/restart"
 	runcmd "github.com/eminwux/kukeon/cmd/kuke/run"
 	startcmd "github.com/eminwux/kukeon/cmd/kuke/start"
 	statuscmd "github.com/eminwux/kukeon/cmd/kuke/status"
@@ -156,6 +157,7 @@ func SetupKukeCmd(rootCmd *cobra.Command) error {
 	rootCmd.AddCommand(killcmd.NewKillCmd())
 	rootCmd.AddCommand(purgecmd.NewPurgeCmd())
 	rootCmd.AddCommand(refreshcmd.NewRefreshCmd())
+	rootCmd.AddCommand(restartcmd.NewRestartCmd())
 	rootCmd.AddCommand(runcmd.NewRunCmd())
 	rootCmd.AddCommand(attachcmd.NewAttachCmd())
 	rootCmd.AddCommand(logcmd.NewLogCmd())
--- a/cmd/kuke/kuke_test.go
+++ b/cmd/kuke/kuke_test.go
@@ -60,7 +60,7 @@ func TestNewKukeCmd(t *testing.T) {
 		t.Errorf("Short mismatch: got %q, want %q", cmd.Short, "Kukeon is a tool for managing Kukeon entities")
 	}
 
-	expectedSubcommands := []string{"init", "create", "get", "delete", "start", "stop", "kill", "purge", "version"}
+	expectedSubcommands := []string{"init", "create", "get", "delete", "start", "stop", "restart", "kill", "purge", "version"}
 	for _, subcmdName := range expectedSubcommands {
 		subcmd := cmd.Commands()
 		found := false
@@ -165,7 +165,7 @@ func TestSetupKukeCmd(t *testing.T) {
 		t.Fatalf("setupKukeCmd() error = %v, want nil", err)
 	}
 
-	expectedSubcommands := []string{"init", "create", "get", "delete", "start", "stop", "kill", "purge", "version"}
+	expectedSubcommands := []string{"init", "create", "get", "delete", "start", "stop", "restart", "kill", "purge", "version"}
 	commands := rootCmd.Commands()
 	commandMap := make(map[string]bool)
 	for _, cmd := range commands {
--- a/cmd/kuke/restart/cell/cell.go
+++ b/cmd/kuke/restart/cell/cell.go
@@ -42,11 +42,17 @@ type MockControllerKey struct{}
 // error/partial-state cell is refused with a `kuke delete cell` pointer.
 //
 // The reconcile branch composes at CLI level over GetConfig + GetBlueprint +
-// ApplyDocuments (which the daemon already wires to stop+update+start on a
-// divergent live cell) rather than introducing a new daemon-side RPC — keeps
-// this change scoped to one CLI verb and avoids a wire-format addition. The
-// AC's user-facing contract holds either way (per the issue's open-question
-// note).
+// ApplyDocuments rather than introducing a new daemon-side RPC — keeps this
+// change scoped to one CLI verb and avoids a wire-format addition. The daemon's
+// ApplyDocuments only bounces containers when image/command/args change (see
+// internal/controller/runner/update_cell.go:containerSpecChanged) — pure
+// env-var or metadata-only divergence is patched in place. To preserve the
+// "restart" verb's contract ("all running containers bounce"), the CLI follows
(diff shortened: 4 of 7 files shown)


eminwux:
**PR #838 Review — `feat(cli): add kuke restart cell with implicit reconcile on OutOfSync (step 2)`**

Solid step-2 wiring — dual-mode branching, scope-flag parity with sibling lifecycle verbs, and the OutOfSync-without-lineage / OutOfSyncError / deleted-Config edge cases all fall through to vanilla restart with a stderr notice. Tests cover the CLI composition comprehensively.

One blocking item:

- **Stated-invariant gap (reconcile-no-bounce)** `cmd/kuke/restart/cell/cell.go:122-128` + `:264`: the help text claims *"the restart is also a reconcile"* and design notes claim the daemon *"wires stop+update+start on a divergent live cell"*, but the daemon's `ReconcileCell` (`internal/controller/apply/reconcile.go:392-396`) routes non-root non-image divergence through `UpdateCell` (`internal/controller/runner/update_cell.go:168-207`), which only stops+restarts containers when image/cmd/args changed — pure env-var or metadata-only divergence is patched in place with no bounce. User sees `Restarted cell "X" ... (reconciled from config "Y")` while the running containers were not actually restarted. Two viable fixes: (a) update the help/output text to disclaim the in-place case, or (b) after `ApplyDocuments` returns and the response indicates no bounce occurred, also call `StopCell`+`StartCell` to honor the restart contract (with a test that exercises an env-var-only OutOfSync divergence).

LGTM once the reconcile-no-bounce contract gap is addressed — everything else is in good shape.

eminwux:
Addressed in 8ef685a:

- **Reconcile-no-bounce gap.** Verified the reviewer's reading: `internal/controller/runner/update_cell.go:282-286` (`containerSpecChanged`) only returns `true` for image/command/args, so the daemon's `ApplyDocuments → UpdateCell` path leaves env-var / metadata-only OutOfSync divergence patched in place with no container bounce. Took option (b) from the review — the CLI now follows `ApplyDocuments` with an explicit `StopCell + StartCell` unless the reconcile already bounced everything (`reconcileBouncedAll` detects the full-bounce signals: `Action == "created"`, `Changes` contains `"root container recreated"`, or `Changes` contains `"runtime stopped: containers re-materialized"`). The "restart" verb's contract — running containers always bounce — now holds across every divergence class.
- **Test coverage.** Added the reviewer-requested env-var-only OutOfSync divergence test (`Apply` returns `container "main" updated: env`, asserts the follow-up `StopCell + StartCell`), updated the existing `outofsync ready cell` test to assert the new bounce, and added a complementary `root-container recreate` test that asserts the bounce is *skipped* when the reconcile already bounced everything (so `reconcileBouncedAll`'s gate is exercised in both directions).
- **PR body / help text / godoc.** Updated the `restartReady` godoc and the PR body design notes to accurately describe the daemon's no-bounce-on-metadata-only behaviour and the CLI's explicit follow-up bounce. The user-facing help text (`Long`) already says "the restart is also a reconcile", which remains accurate — restart now always bounces, plus reconciles on OutOfSync.

Rebased onto current `origin/main` (`1631aad`). `make test` clean post-rebase; `pre-commit` clean.

eminwux:
**PR #838 Review — `feat(cli): add kuke restart cell with implicit reconcile on OutOfSync (step 2)`**

LGTM — dual-mode contract is correct, the daemon-signal strings in `reconcileBouncedAll` match `internal/controller/apply/reconcile.go` exactly, sibling start/stop precedent followed for `ClientFromCmd`, all five `CellState` values covered, errdefs used throughout, and the env-var-only divergence + root-container-recreate paths both have explicit test coverage that pins the "bounce all or skip redundant bounce" decision.

PR #845, opened 2026-05-24
refactor(apply): remove kuke apply -b/-c — apply becomes -f only (#823)

eminwux:
## Summary

Step 4 of #819 — the final slice. Hard-removes `kuke apply -b/-c` (daemon-side reconcile-by-ref) so `kuke apply` is `apply -f <file>` only. With `kuke restart cell <name>` (#821) providing the reconcile path for Config-lineage workflows and OutOfSync surfaced in `kuke get cell` (#822), the apply-side reconcile-by-ref forms no longer have a unique role. No deprecation cycle per umbrella scope (no production kukeon yet).

### What ships

- **Flag definitions removed** from `cmd/kuke/apply/apply.go`: `-b/--blueprint`, `-c/--config`, `--name`, `--param`, `--param-file`, `--realm`, `--space`, `--stack`. The `-f/--file` and `-o/--output` flags remain. `cobra.MarkFlagsMutuallyExclusive` / `MarkFlagsOneRequired` calls drop — there is now exactly one source.
- **Code paths deleted**: `runApplyRef`, `loadCellRef`, `loadFromBlueprint`, `loadFromConfig`, `buildParamMap`, `resolveCellLocation`, `assertCellLineage`, `lineageExpectation`, `lineageMismatchError`. `runApply` now routes unconditionally to `runApplyFile`. `applyFlags` collapses to `{file, output}`.
- **Imports trimmed**: `cmd/config`, `internal/cellblueprint`, `internal/cellconfig`, `internal/cellprofile`, `pkg/api/model/v1beta1`, `gopkg.in/yaml.v3`, `os`, `strings`, `context` — all dropped from `apply.go`.
- **`Use:` / `Long` rewritten**: `Use: "apply -f <file>"`, `Short` / `Long` describe `-f` only with a one-line nod to `kuke restart cell` for Config reconciliation.
- **Shell completion registrations removed**: the `cmd.RegisterFlagCompletionFunc` calls for `realm`/`space`/`stack`/`blueprint`/`config` all came in alongside `-b/-c` and go away with them.
[...]

Issue #835, opened 2026-05-23
cli(run): add --reuse — start a healthy-Stopped clone (overlay preserved), fall back to --clone (step 3)


eminwux:
## Context

Child of #840 (step 3). [`--clone`](https://github.com/eminwux/kukeon/issues/839) introduces the pool: each `--clone` forks the source Config into a counter-named clone (`<source>-0`, `-1`, …), creates a cell, and persists the clone Config so it's addressable by name later. Once clones exist with cells in healthy-Stopped state, the operator (or cron) wants to **restart one of them rather than fork a new one** — so the cell's overlay filesystem (project repo clone, `.claude.json`, any per-cell state) is reused.

That's `--reuse`: pick a healthy-Stopped clone, start its cell (preserving the containerd overlay), fall back to `--clone` only if no idle clone is available.

Driver use case: cron-driven skill execution ([crew #54](https://github.com/eminwux/crew/issues/54)). Each cron tick fires `kuke run kukeon-dev-pick-issue --reuse --env LABEL=bug -d`. First tick lands on an empty pool → falls back to `--clone` to seed `kukeon-dev-pick-issue-0` (initial repo clone happens here). Subsequent ticks start `-0`'s stopped cell — the repo is already cloned, `tty.onInit`'s `git pull` is a cheap delta, claude runs the skill, cell exits → back to Stopped. Overlapping ticks (`-0` still Running) fall back to `--clone` → `-1`. Pool grows to peak concurrent ticks and stays there.

**State preservation is the headline win.** Without `--reuse`, every tick re-clones the project repo into a fresh cell (tens of MB + several seconds of wall-clock per tick for a repo like kukeon). With `--reuse`, the clone happens once per pool member.

## Proposal

Add `--reuse` to `kuke run` on the CellConfig path. Same scope as `--new` and `--clone`: valid only with the `<config>` positional (or `-c <config>`).

**Pick algorithm:**

1. Query CellConfigs in the target realm carrying `metadata.annotations.kukeon.io/source-config: <source-name>` (set by `--clone` per [#839](https://github.com/eminwux/kukeon/issues/839)).
2. For each clone Config, examine the state of its cell (the cell with `metadata.name = <clone-config-name>`).
3. Filter to **healthy-Stopped** — the cell exists, is in Stopped state, and not in a crashed/errored sub-state. (If the daemon's cell-state taxonomy doesn't expose a clean "healthy" filter yet, ship with "Stopped at all" and refine in a follow-up.)
[...]

PR #846, opened 2026-05-24
feat(run): add --clone — fork CellConfig with gap-fill counter naming (step 2)

### LINKED ISSUE #839: cli(run): add --clone — fork CellConfig with counter naming (interactive multi-instance + cron pool seed) (step 2)
## Context

Child of #840 (step 2). The CellConfig path on `kuke run` provides stable single-cell identity (`kuke run <cfg>` is idempotent against one cell named after the Config) and an opt-out for ephemeral cells via [`-g/--generate-name`](https://github.com/eminwux/kukeon/pull/803) (renamed to `--new` in [#833](https://github.com/eminwux/kukeon/issues/833)). Both paths assume a *single* persistent identity per source Config — and `--new` doesn't create a Config artifact at all.

Missing: a "fork the Config" operation for the use cases where the operator wants **multiple persistent runnable units** from one source Config, each with its own stable identity and per-cell state. Two driver use cases:

1. **Interactive multi-instance.** Run multiple persistent dev cells from `kukeon-dev` — `kukeon-dev-0`, `kukeon-dev-1`, `kukeon-dev-2` — each its own attachable shell with its own state. The operator switches between them, leaves work-in-progress in each.

2. **Cron pool seeding.** A cron-driven skill ([crew #54](https://github.com/eminwux/crew/issues/54)) wants to amortize the project repo clone across ticks. The pool is built from clone Configs; [`--reuse`](https://github.com/eminwux/kukeon/issues/835) starts a stopped clone (preserving the containerd overlay so the repo is already there), and falls back to `--clone` on empty pool.

K8s analog: `--clone` is to a CellConfig what `kubectl scale statefulset --replicas` is to a StatefulSet — each new replica gets a stable counter-suffixed identity, and its overlay state persists across restarts. A cloned `kukeon-dev-0` is a first-class CellConfig that `kuke run kukeon-dev-0` resolves directly.

## Proposal

Add `--clone` to `kuke run` on the CellConfig path. Same scope as `--new` and `--reuse`: valid only with the `<config>` positional (or `-c <config>`).

**`--clone` semantics:**

1. Clones the source Config into a new persistent CellConfig in the daemon's config store.
2. The clone's `metadata.name` follows gap-fill counter: `<source>-<N>` where N is the lowest integer ≥ 0 not currently occupied by a clone of this source in the target realm.
3. The clone carries `metadata.annotations.kukeon.io/source-config: <source-name>` so [`--reuse`](https://github.com/eminwux/kukeon/issues/835) can query the pool and `kuke get cellconfigs` can filter clones by source.
4. The clone's `spec` is a deep copy of the source's `spec` — independent identity, independently editable via `kuke apply -f`.
5. After cloning, a cell is started from the clone (cell name = clone Config name, following kukeon's stable-identity rule). Operator attaches by default.

**`--clone --name X`** is the explicit-name path: clone Config named X (create-or-fail if Config X already exists). Same lineage annotation; same cell-from-clone behavior.

**N allocation is atomic** under concurrent `--clone` invocations against the same source — two parallel invocations produce distinct N's, not race on the same slot.

**Examples:**

- Empty state, `kuke run kukeon-dev --clone` → clone Config `kukeon-dev-0` + cell `kukeon-dev-0`. Attach.
- Then `kuke run kukeon-dev --clone` → clone Config `kukeon-dev-1` + cell `kukeon-dev-1`.
- `kuke delete cellconfig kukeon-dev-0`, then `kuke run kukeon-dev --clone` → clone Config `kukeon-dev-0` reused (gap-fill).
- `kuke run kukeon-dev --clone --name debug` → clone Config `debug` + cell `debug`. Fails if Config `debug` exists.
- Later, `kuke run kukeon-dev-1` → resolves the clone Config, restarts the cell idempotently (today's stable-identity path).

## Acceptance criteria

- [ ] `--clone` flag accepted on `kuke run`. Valid only with the `<config>` positional (or `-c <config>`); rejected with `-f`, `-p`, `-b`.
- [ ] Mutex set: `--clone` mutually exclusive with `--new`, `--reuse`, `--rm`. Combinable with `--name`.
- [ ] **Default-name path:** `kuke run <src> --clone` allocates `<src>-<N>` where N is the lowest unused integer ≥ 0 with `kukeon.io/source-config=<src>` annotation in the target realm. Atomic under concurrent invocations.
- [ ] **Named create-or-fail path:** `kuke run <src> --clone --name X` creates a clone Config named `X`. If a CellConfig named `X` already exists in the target realm, fail with `cellconfig "X" already exists; --clone --name requires the name to be free`. No counter fallback.
- [ ] Clone Config persists in the daemon's config store after the cell exits — explicit `kuke delete cellconfig <clone-name>` is required to remove it.
- [ ] Clone Config carries `metadata.annotations.kukeon.io/source-config: <source-name>`. `kuke get cellconfigs` lists clones; a future filter (`--annotation` or equivalent) lists clones of one source.
- [ ] Clone Config's `spec` is independent: editing the source Config does **not** propagate to existing clones.
- [ ] Cell started from clone follows kukeon's stable-identity rule: cell name = clone Config name. `kuke run <clone-name>` (later, no `--clone`) resolves the clone Config and attaches/restarts its cell idempotently.
- [ ] Concurrent `--clone` invocations against the same source: each picks a distinct N (atomic claim). Test: 10 parallel `--clone`s expect N=0…9 distinct.
- [ ] `--clone --name X` racing another `--clone --name X` claim losing the race: fails with the existing collision error.
- [ ] Tests cover: serial counter allocation (0,1,2), gap-fill on `kuke delete cellconfig`, concurrent allocation, `--clone --name` create-or-fail success, `--clone --name` collision, cross-realm isolation, spec independence (source edit doesn't affect clones).
- [ ] Help text positions the three-flag set: `--new` (ephemeral cell, no Config artifact, hex name), `--clone` (persistent clone Config + cell, counter-named, for multi-instance + cron pool seeding), `--reuse` ([#835](https://github.com/eminwux/kukeon/issues/835)) (start a stopped clone, fork via `--clone` if pool empty).
- [ ] Docs in `docs/site/cli/kuke-run.md` and `docs/cli-use-cases.md`. The §"Create a single cell from YAML" / "Conflicting `kuke run`" sections cover the three-flag set and their mutex relationships.

## Out of scope

- `--clone` on the `-b/--blueprint` path. Blueprints aren't identities; Configs are.
- `--clone` on `-f/--file`. The file *is* the spec — `kuke apply -f` is the path.
- Automatic clone-Config GC. Operator owns the lifetime via `kuke delete cellconfig`.
- A `--clone-of` field on `CellConfig.spec` (typed source pointer). The annotation is sufficient for v1.
- Source-config edits propagating to existing clones. Each clone is independent post-fork.

## Notes

- **Why `--clone --rm` is rejected:** `--rm` removes the cell on exit. With `--clone`, that leaves a clone Config whose cell is gone — operationally messy (operator created a persistent slot, immediately disposed of its cell). If you want a cell-only ephemeral run with a specific name, use `--new --name X --rm` (no Config artifact). If you want a persistent slot but no cell yet, `kuke apply -f <clone-spec>` creates the Config without running a cell. `--clone --rm` would only buy you "create the slot AND a cell, throw the cell away now" — more cleanly expressed as `kuke apply -f`. Rejection keeps the flag's intent crisp.
- **Driver use cases:**
  - Interactive multi-instance: `kuke run kukeon-dev --clone` → `kukeon-dev-0`, repeat for more parallel shells. Detach and re-attach via `kuke run kukeon-dev-1`.
  - Cron pool seeding ([crew #54](https://github.com/eminwux/crew/issues/54)): first cron tick lands on empty pool; [`--reuse`](https://github.com/eminwux/kukeon/issues/835) falls back to `--clone` to seed `kukeon-dev-pick-issue-0`. Subsequent ticks `--reuse` it (overlay preserved, repo already cloned).
- **Counter is keyed on the source-config annotation**, not on cell names — clone Configs are the identity, cells follow. Matches the StatefulSet analogy (replicas are identity, pods follow).
- **Forward-compat with [AgentRun #503](https://github.com/eminwux/kukeon/issues/503) / [CronAgentRun #755](https://github.com/eminwux/kukeon/issues/755):** when those kinds land, an `AgentRun` references a source CellConfig and the runner internally does a `--clone`-equivalent to materialize the run's identity. `--clone` is the CLI surface for the same operation in the meantime.

eminwux:
## Summary

- Add \`--clone\` to \`kuke run\` on the \`<config>\` positional path. Forks the source CellConfig into a new persistent CellConfig and starts a cell from the clone — the foundation for interactive multi-instance (\`kukeon-dev-0\`, \`kukeon-dev-1\`, …) and cron pool seeding (\`--reuse\`, #835 planned).
- **Default name path:** \`kuke run <src> --clone\` allocates \`<src>-<N>\` where N is the lowest unused integer ≥ 0 among clones of the source in the target realm (gap-fill counter). Atomic under concurrent invocations.
- **Named create-or-fail path:** \`kuke run <src> --clone --name X\` creates clone CellConfig \`X\` and fails with the AC-pinned message \`cellconfig "X" already exists; --clone --name requires the name to be free\` on collision.
- **Lineage marker:** the clone carries \`metadata.annotations.kukeon.io/source-config: <source>\` so the counter loop and future \`kuke get cellconfigs\` filters can enumerate clones of a source.
- **Spec independence:** the clone's \`spec\` is a deep copy of the source's — editing the source post-clone never propagates to existing clones. Maps and pointer fields (\`CellConfigSecretFill.SecretRef\`) are independently copied.
- **Atomic write path:** the AC's concurrent-allocation guarantee depends on the daemon write being create-only. New runner method \`WriteConfigIfAbsent\` uses \`os.Link\` (POSIX's atomic "create destination, fail if exists" primitive) on the temp file after writing; rename — which the existing \`WriteConfig\` uses — overwrites silently and cannot serve the AC. New controller method \`CreateConfig\` runs the same scope / blueprint / slot-fill validation as \`ReconcileConfig\` but persists via the atomic path. New \`KukeonV1.CreateConfig\` RPC surfaces it to the CLI.
- **Mutex set:** \`--clone\` is mutually exclusive with \`--new\` (different operations — ephemeral cell vs. persistent clone Config) and \`--rm\` (orphans the clone Config). Combinable with \`--name\`. Rejected with \`-f\`/\`-p\`/\`-b\`.
- **\`Annotations\` on \`CellConfigMetadata\`** (external + propagation via the existing YAML-document round-trip; the internal carrier only lifts identity fields for path resolution, so the daemon never interprets annotations). The annotation is carried into the next reader's view of the Config (\`kuke get config -o yaml\`).
- **Errmap:** \`ErrConfigExists\` and \`ErrConfigNotFound\` mapped to wire-Kind strings so \`errors.Is\` survives the JSON-RPC roundtrip — the named-collision error message would otherwise come back as the bare sentinel.
- **Docs:** \`docs/site/cli/kuke-run.md\` flag table + identity-shapes table (7 rows now: bare, \`--name\`, \`--new\`, \`--new --name\`, \`--new --rm\`, \`--clone\`, \`--clone --name\`) + mutex matrix + examples. \`docs/cli-use-cases.md\` "Run a one-off cell from YAML" sequence list + invariant paragraph.

## Test plan

- [x] \`make test\` — full suite green on both the root and \`cmd/kukebuild\` modules.
- [x] \`make dev-init\` — daemon-parity smoke + nested kuke-attach smoke pass on the dev host with the new daemon binary. (The full re-run hit an unrelated CNI bridge-state hiccup on the dev-init-attach subtest; the daemon-parity tail produced the expected \`default\` / \`kuke-system\` Ready output both times.)
- [x] Hands-on smoke: \`kuke apply\` source CellConfig + 3× \`kuke run src --clone -d\` against a real daemon. Verified:
  - Serial counter: \`src-0\`, \`src-1\`, \`src-2\` allocated in order.
  - Gap-fill: after deleting \`src-1\`, the next \`--clone\` reuses \`src-1\`.
  - Annotation: \`kuke get config src-3 -o yaml\` shows \`metadata.annotations.kukeon.io/source-config: src\`.
  - Named create-or-fail: \`--clone --name custom\` succeeds first time, second invocation surfaces \`Error: cellconfig "custom" already exists; --clone --name requires the name to be free\`.
- [x] \`GOWORK=off go vet ./...\` clean.
- [x] \`pre-commit run --files <touched>\` — golangci-lint + prettier + go fmt + go-mod-tidy + addlicense all pass (golangci-lint rewrote two for-range-N loops in \`cmd/kuke/run/clone.go\` to the Go 1.22+ \`for range N\` shape; intentional, no behaviour change).
- [x] Unit tests cover every AC checkbox:
  - \`TestRun_FromConfig_Clone_DefaultName_AllocatesCounterZero\` — empty pool → \`<src>-0\` + annotation + cell name.
  - \`TestRun_FromConfig_Clone_DefaultName_GapFillSkipsExistingClones\` — \`src-0\` + \`src-2\` taken → next is \`src-1\`.
  - \`TestRun_FromConfig_Clone_DefaultName_IgnoresUnannotatedSameShapedName\` — manual \`prod-3\` (no annotation) doesn't reserve the counter slot.
  - \`TestRun_FromConfig_Clone_NamedExplicit_HappyPath\` — \`--clone --name debug\` single-shot CreateConfig path.
  - \`TestRun_FromConfig_Clone_NamedExplicit_CollisionFails\` — pins the AC's exact error wording.
  - \`TestRun_FromConfig_Clone_DefaultName_RetriesOnRace\` — first CreateConfig races and loses; loop retries with next-free N.
  - \`TestRun_FromConfig_Clone_Concurrent_DistinctNs\` — 10 sequential clones against a shared "taken" set (the daemon's \`os.Link\` is what enforces atomicity in production; the test pins that the client loop composes the right candidates under a shared lock) emit N=0…9 distinct.
  - \`TestRun_FromConfig_Clone_SpecIndependence\` — mutating \`source.Spec.Values/Repos/Secrets\` post-clone doesn't propagate to the captured clone.
  - \`TestRun_FromConfig_Clone_CrossRealmIsolation\` — clones in \`other-realm\` don't bump \`cfg-realm\`'s counter.
  - \`TestRun_FromConfig_Clone_MutexWithNew\`, \`TestRun_FromConfig_Clone_MutexWithRm\`, \`TestRun_Clone_OnlyValidWithConfig\` — flag-parsing rejection paths.
  - Runner-level: \`TestWriteConfigIfAbsent_CreatesAndRejectsCollision\`, \`TestWriteConfigIfAbsent_LeavesNoTempFiles\`.
  - Controller-level: \`TestCreateConfig_HappyPathWritesAtomicCreateOnly\`, \`TestCreateConfig_CollisionPropagatesErrConfigExists\`, \`TestCreateConfig_RejectsMissingScope\`.
- [ ] \`make e2e\` — not run locally (CI gates this; the dev host has no docker daemon and the e2e suite needs one alongside containerd).

## Design notes

- **Atomic write primitive:** \`os.Link\` is the portable atomic create-or-fail on POSIX. \`renameat2(RENAME_NOREPLACE)\` would also work on Linux but ties us to a syscall package; \`O_EXCL\` doesn't compose with the existing temp-file-then-rename idiom (the temp file is the inode; we want the destination to be the atomic gate, not the temp). The write-to-temp + link-to-destination pattern keeps reader safety (no torn writes) and adds the create-only gate without a new syscall.
- **Why \`CreateConfig\` not \`ApplyDocuments\`:** \`kuke apply -f\` is write-through by contract — re-applying a CellConfig of the same name updates the on-disk document. The clone path needs the opposite: a write that *fails* on collision so the counter loop can retry. \`CreateConfig\` is the create-only sibling of \`ReconcileConfig\` sharing scope / blueprint / slot validation; the only delta is the atomic-write call site.
- **Counter scan:** \`ListConfigs\` returns metadata only, so the CLI \`GetConfig\`s each \`<src>-<digits>\` candidate to verify the source-config annotation. The annotation is the source of truth (a manually-named \`<src>-3\` Config doesn't count as a clone), and a clone's name slot in the broader name set still gets honoured so the counter never proposes a candidate that would collide with an unrelated Config. In practice the pool is small (< 10 entries) so the per-candidate \`GetConfig\` round-trip is cheap.
- **Stale \`kuke apply -c <config>\` pointers in pre-existing docs:** \`docs/site/cli/kuke-run.md\` (lines 92, 113) and \`docs/cli-use-cases.md\` (lines 284, 286) still reference the divergence-reconcile pointer that PR #845 removed. These predate my changes and I left them alone per CLAUDE.md "no unrelated cleanups". Surfacing here for follow-up — a single chore PR can sweep them once the \`kuke apply -c\` → \`kuke apply -f\` doc migration starts.

Closes #839


eminwux:
**PR #846 Review — `feat(run): add --clone — fork CellConfig with gap-fill counter naming (step 2)`**

LGTM — gap-fill counter loop + atomic `os.Link` create-or-fail enforces the AC's concurrent-allocation guarantee, deep-copy of `CellConfigSpec` (including the `SecretRef` pointer rewrap) keeps clones independent of post-clone source mutations, and the mutex / positional-only rejection paths cover every shape the AC names. CLAUDE.md daemon-parity smoke + the AC's hands-on flow both documented as run.

_(First-pass review routed via `/recheck-pr`.)_

Issue #840, opened 2026-05-24
planning: kuke run CellConfig identity flags — --new, --clone, --reuse + --env runtime knob


eminwux:
## Context

The CellConfig path on `kuke run` evolved organically: `kuke run <cfg>` for stable-identity attach, `-g/--generate-name` ([#803](https://github.com/eminwux/kukeon/pull/803)) for ephemeral one-shot cells. Two operator needs surface this isn't enough:

1. **Multi-instance from one Config** — run multiple persistent dev cells from `kukeon-dev` (`kukeon-dev-0/1/2`), each attachable and independently editable.
2. **Cron-driven skill execution with state preservation** — schedule `/pick-issue`-style ticks against per-(project, role, skill) CellConfigs ([crew #54](https://github.com/eminwux/crew/issues/54)) without re-cloning the project repo on every tick.

Today the only opt-out from stable identity is `-g`, which generates a hex-named cell with no Config artifact — fine for one-shot ephemeral, but it can't seed a pool that subsequent ticks can reuse (the cell has no addressable Config to come back to).

This umbrella introduces three new identity flags (`--new`, `--clone`, `--reuse`) and one orthogonal runtime knob (`--env`) on `kuke run` — together they cover the three use cases above plus the existing stable-identity path, with K8s-style mental models:

| Flag | Operation | Cell name | Config side-effect | K8s analog |
|---|---|---|---|---|
| (no flag) | Idempotent start/attach | `<cfg>` | None (existing) | — |
| `--new` ([#833](https://github.com/eminwux/kukeon/issues/833)) | Ephemeral cell | hex or `--name` | None | unmanaged pod (`kubectl run`) |
| `--clone` ([#839](https://github.com/eminwux/kukeon/issues/839)) | Fork Config + cell | counter or `--name` | Persistent clone Config | StatefulSet scaling |
| `--reuse` ([#835](https://github.com/eminwux/kukeon/issues/835)) | Start stopped clone (overlay preserved); fork on empty pool | clone's stable name | Reuses existing, or forks via `--clone` | (slot pool with overlay reuse — no direct analog) |
| `--env` ([#834](https://github.com/eminwux/kukeon/issues/834)) | Runtime env injection (orthogonal to identity) | n/a | None | env var injection at start |

**Mutex set** (the three identity flags are pairwise mutex; `--name` and `--rm` combine selectively):

| Combination | Status |
|---|---|
[...]

PR #849, opened 2026-05-24
feat(run): add --env KEY=VALUE for runtime container env injection (step 4)

### LINKED ISSUE #834: cli(run): add --env KEY=VALUE for runtime container env injection (step 4)
## Context

Child of #840 (step 4). Today's `kuke run` knobs are render-time or identity-time:

- `--param KEY=VALUE` / `--param-file <file>` — blueprint parameter substitution at render time. Rejected with the `<config>` positional (`cmd/kuke/run/run.go:326`): a Config's identity is locked.
- `--name <override>` — cell name override. Rejected with `<config>` positional (`cmd/kuke/run/run.go:312`).
- `--new` ([#833](https://github.com/eminwux/kukeon/issues/833)) — ephemeral cell with hex name (or `--name`), no Config artifact.
- `--clone` ([#839](https://github.com/eminwux/kukeon/issues/839)) — persistent clone Config + counter-named cell.
- `--reuse` ([#835](https://github.com/eminwux/kukeon/issues/835)) — start a stopped clone (overlay preserved).

None of these inject *runtime container env*. For cron-driven skill execution against a stable CellConfig (the [crew #54](https://github.com/eminwux/crew/issues/54) use case), the operator wants the *work filter* (e.g. `LABEL=bug` vs `LABEL=enhancement`) to vary per crontab line against the same `kukeon-dev-pick-issue` Config — without relaxing the Config's identity contract (which `--param` correctly forbids).

The natural mechanism is a `--env KEY=VALUE` flag that injects extra env into the cell's attachable container at start time. `--env` is conceptually distinct from `--param`: `--param` substitutes into the rendered manifest (changing the *cell spec*), `--env` adds to the running container's environment (not changing the spec, just the runtime). Keeping them distinct preserves CellConfig identity while exposing a per-invocation knob.

## Proposal

Add `--env KEY=VALUE` to `kuke run`. Repeatable (`--env LABEL=bug --env PRIORITY=A`). Format `KEY=VALUE`; empty value (`KEY=`) allowed; missing `=` rejected with a clear error.

Scope: env entries are merged into the **attachable container's** env list at cell start time (the container `kuke run` attaches to and runs onInit in — for crew's role blueprints, the `work` container). Other containers (e.g. `root`) are not touched. If a use case for broader injection surfaces, file separately.

Override semantics: `--env LABEL=bug` whose key collides with an existing entry in `spec.cell.containers[<attachable>].env` overrides the spec value. Documented in help text; tests cover the precedence.

Source-path compatibility: works with `<config>` positional (the cron driver), `-b/--blueprint`, `-p/--profile`, and `-f/--file`. No source-specific rejection. Compatible with all three identity flags (`--new`, `--clone`, `--reuse`).

Divergent-spec interaction: the divergent-spec check at `cmd/kuke/run/run.go` (the one that refuses to attach to a Ready cell whose spec doesn't match) must **exclude** `--env`-injected keys from the comparison set — otherwise every `--env` invocation against a Ready cell would trip the check. With `--reuse` ([#835](https://github.com/eminwux/kukeon/issues/835)) the cell is started (not attached-as-Ready), so the divergent check runs against the cell's stored spec; the carve-out must apply there too.

Reuse interaction: with `--reuse`, `--env` re-injects per cron tick. The cell's stored spec env is unchanged across restarts; only the runtime container env carries the per-invocation `--env`. This is the load-bearing combination for the cron use case — same cell across ticks, different `LABEL` per tick.

Typical cron line after this lands + #833 + #839 + #835 (the [crew #54](https://github.com/eminwux/crew/issues/54) driver use case):

```cron
# Cron: pool reuse with overlay preservation (recommended for skills that clone the project repo)
*/10 * * * * kuke run kukeon-dev-pick-issue --env LABEL=priority:A --reuse -d
*/30 * * * * kuke run kukeon-dev-pick-issue --env LABEL=priority:B --reuse -d
*/10 * * * * kuke run kukeon-pm-pick-doc    --env LABEL=documentation --reuse -d

# One-shot: ephemeral per tick (no state preservation)
*/10 * * * * kuke run kukeon-pr-reviewer-review-pr --env LABEL=ready-for-review --new --rm -d
```

## Acceptance criteria

- [ ] `kuke run --env KEY=VALUE` flag accepted, repeatable. Format `KEY=VALUE`; empty value allowed (`KEY=`); missing `=` rejected with `--env requires KEY=VALUE (got: <input>)`.
- [ ] Multiple `--env` occurrences merge; identical key occurring twice with different values is rejected with a clear error (don't silently take "last wins"; explicit is better).
- [ ] Env entries injected into the cell's attachable container's env at cell start time, in addition to `spec.cell.containers[<attachable>].env`.
- [ ] A `--env` key that collides with a key already in the attachable container's spec env overrides the spec value. Help text documents this.
- [ ] Works with all source paths: `<config>` positional, `-b`, `-p`, `-f`. Cross-source tests cover each.
- [ ] Works in combination with `--reuse` ([#835](https://github.com/eminwux/kukeon/issues/835)): `--env` re-injects per invocation against the restarted cell. Cell's stored spec env unchanged across restarts.
- [ ] Works in combination with `--new`, `--clone` ([#833](https://github.com/eminwux/kukeon/issues/833), [#839](https://github.com/eminwux/kukeon/issues/839)): `--env` injects on the freshly created cell.
- [ ] Divergent-spec check (the Ready short-circuit's spec comparison around `cmd/kuke/run/run.go:486`) excludes `--env`-injected keys from the comparison. Test: `kuke run <cfg>` then `kuke run <cfg> --env LABEL=bug` against the same Ready cell does not trip the divergent check.
- [ ] Injected env is available to `tty.onInit` scripts (the gate step inherits them). Test: an onInit script that echoes `$LABEL` sees the value supplied by `--env LABEL=bug`.
- [ ] Help text positions the flag: *"--env injects runtime container env (per-invocation knob); for render-time spec substitution use --param (blueprint path only)."*
- [ ] Docs updated in `docs/site/cli/kuke-run.md` and `docs/cli-use-cases.md`. The latter's §"Conflicting `kuke run`" mentions that `--env` is the per-invocation surface against a stable CellConfig.

## Out of scope

- `--env-file <file>` (file form). Defer until a real use case surfaces.
- Persisting `--env` into the CellConfig's spec. `--env` is per-invocation only — that's the whole point.
- Broader injection scope (all containers, init containers, etc.). Attachable-only for v1.
- Env templating (interpolating `$OTHER_VAR` inside the supplied value). Pass the value verbatim; let the operator's shell do the work upstream.
- A typed `env:` field on `CellConfig.spec` that would let the Config itself enumerate runtime-overridable env keys. Premature; `--env` is operator-facing only.

## Notes

- Driver use case: [crew #54](https://github.com/eminwux/crew/issues/54) — per-(project, role, skill) CellConfigs whose GATE_CHECK and PROMPT are baked into the Config but whose `$LABEL` is supplied per cron tick via `--env`.
- Co-evolves with three sibling flags on the CellConfig path: `--new` ([#833](https://github.com/eminwux/kukeon/issues/833), ephemeral cell), `--clone` ([#839](https://github.com/eminwux/kukeon/issues/839), persistent clone), `--reuse` ([#835](https://github.com/eminwux/kukeon/issues/835), start-stopped-clone). `--env` is orthogonal to which identity flag is used; combine freely.
- Forward-compat with [AgentRun #503](https://github.com/eminwux/kukeon/issues/503) / [CronAgentRun #755](https://github.com/eminwux/kukeon/issues/755): `--env` maps cleanly onto a typed `AgentRunSpec.env` (or similar) when those kinds land. The CLI surface added here is the lightweight operator-facing equivalent in the meantime.
- Note for implementers: crew blueprint's onInit scripts use `${VAR}` substitution that's *kukeon-side* (render-time, for declared blueprint parameters). To read `--env`-injected vars in onInit scripts without colliding with kukeon's substitution, the script should reference them with plain `$LABEL` (or quoted `"$LABEL"`) rather than `${LABEL}` if kukeon's substitution is strict on undeclared `${VAR}`. Worth confirming the substitution semantics during implementation and documenting the recommended pattern in the help text.

eminwux:
## Summary

Adds `--env KEY=VALUE` to `kuke run` — the **per-invocation runtime knob** for injecting extra env into the cell's attachable container at start time, without changing the cell's stored spec (step 4 of #840's CellConfig identity epic).

- **Repeatable StringArray flag.** Format `KEY=VALUE`; empty value (`KEY=`) is allowed; missing `=` is rejected at parse time (`--env requires KEY=VALUE, got "..."`). The same KEY twice with different values is rejected (`--env LABEL specified twice with different values ("bug", "enh"); pick one`) rather than silently last-wins. Same KEY same VALUE is deduped (cron lines that drift over time stay harmless).
- **Server-side merge into the attachable container.** The runner picks the target with the same precedence as `kuke attach` (`tty.default` > first `attachable: true` non-root) and merges runtime entries into the OCI process `Env` — surviving spec entries first, runtime entries appended, runtime wins on KEY collision. Other containers are untouched.
- **Transport-only on `CellSpec.RuntimeEnv`.** New `v1beta1.CellSpec.RuntimeEnv []string` with `yaml:"-"` on the author surface. The `ConvertCellDocToInternal` direction (CLI → daemon) preserves it so the runner sees the operator's intent; `BuildCellExternalFromInternal` (daemon → external) deliberately drops it — the same builder produces both metadata.json writes and RPC reply payloads, so dropping keeps the persisted spec clean **and** keeps the divergent-spec check on a subsequent `kuke run <config>` (no `--env`) from tripping on prior injections. Each invocation re-supplies its own RuntimeEnv.
- **All source paths + all identity flags.** Wired through `-f`/`-p`/`-b`/`<config>` and `--new`/`--clone`/`--reuse`. `--reuse` re-injects per invocation against the restarted cell via the StartCell doc — overlay preserved, env refreshed (cron-driver use case from #840).
- **`viper.BindPFlag` intentionally omitted** for `--env` because `StringArray` doesn't round-trip cleanly through viper's `GetStringSlice`. `KUKE_RUN_ENV` is declared in `cmd/config/env.go` for parity but not bound. (Documented in the flag's call site.)

### Non-obvious calls
- **JSON tag on `RuntimeEnv` kept, drop-on-outbound done in apischeme.** Using `json:"-"` would also strip the field from JSON-RPC traffic between CLI and daemon, which is the one direction that must preserve it. The cleanest place to break the symmetry is in `BuildCellExternalFromInternal`, where every disk-write AND every RPC reply already flow through. Unit-tested in `TestRuntimeEnv_InboundPreserved_OutboundDropped`.
- **Disk format is JSON, not YAML.** Initial draft relied on `yaml:"-"` to strip the field on persistence; an end-to-end smoke against the live daemon (`metadata.json` carried `runtimeEnv` after a `--env` invocation) caught it. Fixed by the apischeme drop. The `yaml:"-"` tag is retained for the author surface (a `kuke apply -f manifest.yaml` should still refuse to read `runtimeEnv:` from a hand-edited file).
- **`--reuse` needed an extra parameter on `pickAndStartReusableClone`.** The helper builds a fresh `cellLookup` doc internally; the operator's `--env` couldn't ride on cfgRes.Config — it lives at the cell level, not the CellConfig level. Threaded `runtimeEnv []string` through.

## Test plan

- [x] `make test` — all package suites pass (unit + integration).
- [x] `GOWORK=off go test ./cmd/kuke/run/ -run TestRun_EnvFlag -v` — 12 new integration tests cover -f / -b / `<config>` / `--new` / `--clone` / `--reuse` / `--reuse` empty-pool fallback / Stopped-cell StartCell path / format errors at CLI / divergent-spec check ignoring `--env`.
- [x] `GOWORK=off go test ./internal/controller/runner/ -run TestResolveAttachable -v` and `TestMergeRuntimeEnv` — 11 helper tests cover the attachable-container precedence (Tty.Default > first attachable > skip-root > empty) and the merge contract (distinct keys, runtime-wins-on-collision, no-op on empty runtime/spec, non-attachable unchanged, no input mutation).
- [x] `GOWORK=off go test ./internal/apischeme/ -run TestRuntimeEnv -v` — the inbound/outbound asymmetry on `CellSpec.RuntimeEnv`.
- [x] `GOWORK=off go test ./cmd/kuke/run/ -run TestParseEnvArgs -v` — 9 parser tests cover happy path, empty input, empty value accepted, missing `=` rejected, empty KEY rejected, duplicate KEY different VALUE rejected, duplicate KEY same VALUE deduped, value with embedded `=`, KEY whitespace trimmed.
- [x] `make dev-init` — the daemon-parity smoke from the project CLAUDE.md ran clean before the post-hoc apischeme persistence fix; the fix is contained in `apischeme/scheme.go` and unit-tested. A second `make dev-init` re-run was attempted but blocked by transient registry DNS failures (`dial tcp: lookup registry-1.docker.io ... i/o timeout`) on the host — environmental, not a code issue, and the daemon-parity contract is unchanged by the persistence fix (the fix narrows what gets serialized into metadata.json after a `--env` invocation; it does not affect `kuke init` itself).
- [x] **End-to-end smoke against the running daemon:** `sudo kuke run -f /tmp/env-smoke.yaml --env LABEL=bug --env PRIORITY=A -d` then `sudo ctr -n default.kukeon.io container info default_default_env-smoke_work` — confirmed the OCI process env carries `SPEC_KEY=spec-value` (preserved), `LABEL=bug` (overrode spec's `LABEL=spec-label`), and `PRIORITY=A` (appended).
- [x] `pre-commit run --files <changed>` — passes (prettier auto-formatted the kuke-run.md docs row; re-run clean).

Closes #834


eminwux:
**PR #849 Review — `feat(run): add --env KEY=VALUE for runtime container env injection (step 4)`**

LGTM — clean threading of the new `RuntimeEnv` field from CLI through the `apischeme` inbound-preserve / outbound-drop asymmetry into the runner's OCI build merge, correctly scoped to the attachable container with override-on-key. Parser rules (missing `=`, dup-KEY-diff-value, empty value, KEY whitespace trim) are well-pinned, the `TestRuntimeEnv_InboundPreserved_OutboundDropped` test cleanly nails the persistence contract, and the end-to-end smoke against the live daemon was actually executed.

Pre-merge: re-run `make dev-init` once registry DNS recovers per the test plan note — the apischeme outbound-drop only narrows what gets serialized so the daemon-parity contract is unchanged, but the project CLAUDE.md asks for the smoke whenever the daemon read path is touched.

Issue #599, opened 2026-05-18
kuketty: always-on debug log at tty/kuketty.log + configurable Tty.LogLevel


eminwux:
## Context

Today the kuketty wrapper's own slog output is **opt-in and operator-invisible by default**:

- `cmd/kuketty/main.go:194-203` (`openTerminalLogger`) returns a `slog.New(slog.NewTextHandler(io.Discard, nil))` whenever `doc.Spec.LogFile` is empty.
- `internal/controller/runner/attachable.go:243-284` actively zeros sbsh's derived default (`terminalSpec.LogFile = ""`, `LogFileMode = 0`, `LogFileGID = nil`) when `Tty.LogFile` is unset.
- The log level is hardcoded `"info"` at `cmd/kuketty/main.go:198`: `sbshlogging.NewFileLogger(logfile, "info")` — no spec field controls it.

This conflates two distinct concerns:

1. **Workload capture** — the bytes the workload writes to its TTY. Correctly opt-in, exposed at `tty/capture` (sbsh capture file).
2. **kuketty's own process log** — debug/error output from the wrapper itself (config parse, socket bind, JSON-RPC accept errors, lifecycle). This is the operator's primary diagnostic when an attach session misbehaves; it should always be available, not gated on the operator pre-populating a path in the container spec.

#289 phase 3 deliberately made `Tty.LogFile` opt-in. That was the right call for the workload-capture concern but is wrong for the kuketty-process log.

## Proposal

- kuketty always writes its slog output to a fixed in-container path under the per-container TTY dir: **`tty/kuketty.log`** (peer to `tty/capture`, same bind-mount visibility, same permission model — kukeon-group GID + `0640`-style mode).
- Add a **`Tty.LogLevel`** field (string enum: `debug` / `info` / `warn` / `error`, default `info`) plumbed end-to-end:
  - `pkg/api/model/v1beta1/container.go` (`ContainerTty` struct)
  - `internal/modelhub/container.go` (intermediate model)
  - `internal/controller/runner/attachable.go` (rendered into the kuketty `TerminalDoc`)
  - `cmd/kuketty/main.go` (passed through to `sbshlogging.NewFileLogger`)
- The current `Tty.LogFile` field is retired (or repurposed) since the path is now daemon-controlled — operators don't pick where it lands, same way they don't pick where `tty/capture` lands.
- Workload capture (`tty/capture`) is unchanged.

## Acceptance criteria

[...]

PR #856, opened 2026-05-24
fix(attachable): chown stale tty children on restart so non-root kuketty can re-open them

### LINKED ISSUE #850: attachable: postCreateChown doesn't cover pre-existing tty-dir files; kuketty exits silently when run as non-root USER after a restart
## Context

`(*Exec).attachablePostCreateChown` (`internal/controller/runner/attachable.go:484`) chowns only two paths to the resolved container UID before kuketty starts:

- the tty leaf directory at line 511, and
- the per-container `kuketty-metadata.json` file at line 519.

On a fresh provision that's enough — kuketty creates `kuketty.log`, `capture`, and per-terminal subtrees as the container UID itself. On a **restart** of an existing cell whose tty dir already contains those files from a prior run, the stale inodes stay owned by root (the daemon uid that wrote them when the chown gap was introduced).

When the work image's resolved USER is non-root (e.g. `claude:latest` → UID 1000), kuketty inside the container then fails at `openTerminalLogger` (`cmd/kuketty/main.go:132`). `sbsh/pkg/logging.NewFileLogger` opens the existing `kuketty.log` with `O_RDWR` (no `O_CREATE`, per the v0.11.1 contract), the existing file is `root:kukeon 0640`, and the open returns EACCES. kuketty exits with code 70 *before writing a single byte to its own log* — so `kuketty.log`'s mtime stays frozen at the prior run's timestamp and there is nothing to grep for.

The freshly bound socket inode (created by `claimSocketListener` earlier in `cmd/kuketty/main.go:127`) survives kuketty's exit with no listener attached. The reconciler observes work=Stopped, wind-down kills the cell, and `kuke attach` dials a dead socket → `connection refused`.

### Repro on `kukeon-pm-0` (post-reboot, `claude:latest` image)

```
$ kuke run -p kukeon-pm --name kukeon-pm-0
Cell "kukeon-pm-0" (realm "default", space "default", stack "default")
  - metadata: already existed
  - cgroup: already existed
  - root container: missing
  - containers: started

$ sudo ls -la /opt/kukeon/data/.../kukeon-pm-0/work/tty/
-rw-r----- 1 root kukeon 8366725 May 24 18:34 kuketty.log    # still root-owned after restart
srwxr-xr-x 1 inwx kukeon       0 May 24 19:59 socket          # freshly bound by kuketty, but kuketty already exited
```

Manual workaround that confirms the diagnosis:

```
$ sudo chown 1000:kukeon /opt/kukeon/data/.../kukeon-pm-0/work/tty/{kuketty.log,capture,metadata.json}
$ kuke run -p kukeon-pm --name kukeon-pm-0
# attach now succeeds
```

This is a silent regression that landed with #599 (the always-on `kuketty.log` at `tty/kuketty.log`): the chown coverage was not extended to the new file. Distinct from #672 — that fix targeted sbsh's per-terminal `terminals/<id>/` subtree, not the tty leaf's own log/capture files.

### Why it bites attachable cells specifically

The work container running the user workload picks up the image's USER (1000 for `claude:latest`). Cells whose work image runs as root (e.g. `kukeon-dev-0` with `claude-root:latest`) are unaffected — the stale root-owned files are still openable by their root-UID kuketty.

## Proposal

Extend `attachablePostCreateChown` to cover the files kuketty will open at startup, not just the tty leaf dir and the metadata file. Implementer's choice between:

- **chown the stale files** to the resolved container UID (preserves capture/log history across restarts, which #599 explicitly wanted), or
- **delete the stale files** at start (simpler, kuketty re-creates them fresh under its own uid; loses prior capture/log).

Both approaches close the operator-facing symptom. The current set of inodes that need either treatment: `kuketty.log`, `capture`, `capture.*.gz`, `metadata.json`, `.meta-*.tmp`, plus the `terminals/<id>/` subtree #672 already addressed for the metadata path.

Either approach should no-op on `ENOENT` (mirrors the metadata chown branch at line 521).

## Acceptance criteria

- [ ] After `kuke run` / `kuke start cell` on an existing-Stopped cell with a non-root image USER and a stale `kuketty.log` from a prior run, kuketty starts cleanly: `kuketty.log` mtime advances within the start window, the host-side socket gets a listener, and `kuke attach` succeeds.
- [ ] No regression on root-USER cells (e.g. `claude-root:latest`-style cells): the new step is a no-op when the stale files are already correctly owned.
- [ ] Test pre-seeds the tty dir with `root:kukeon 0640 kuketty.log` and `root:kukeon 0640 capture`, starts a cell whose container resolves to UID 1000, and asserts (a) the freshly bound socket has a listener and (b) `kuke attach` succeeds.

## Notes

- Code anchor: `internal/controller/runner/attachable.go:484 attachablePostCreateChown` (existing chown coverage at lines 511 and 519).
- Introducing change: #599 (always-on `kuketty.log` at `tty/kuketty.log`).
- Related closed chown work: #258, #260, #369, #672.

eminwux:
## Summary
- `attachablePostCreateChown` only chowned the tty leaf dir and the per-container `kuketty-metadata.json`. The stale `kuketty.log` / `capture` / sbsh `metadata.json` / `.meta-*.tmp` inodes left under `tty/` from a prior run stayed root-owned, so on a restart of a cell whose work-image USER resolved non-root (e.g. `claude:latest` → 1000) the in-container kuketty hit EACCES at `openTerminalLogger` and exited (code 70) before claiming the socket listener. The freshly bound socket then had no listener and `kuke attach` got `connection refused`. Issue #850.
- Extend the post-create chown to walk every child under the tty leaf dir and reset its owner to the resolved container uid (kept the chown direction over the delete option to preserve the capture/log history #599 explicitly wanted across restarts). `filepath.Walk` + `os.Lchown` so a stray symlink (none today, defense-in-depth) is chowned at the link; per-entry ENOENT tolerated to absorb a concurrent unlink race. Recurses into subdirectories too, so any pre-#672 `terminals/<id>/` subtree inherited from an older daemon also gets covered in the same pass. A fresh provision sees an empty dir and the walk is a no-op — zero overhead for the hot path.
- Coverage: three new unit tests in `attachable_test.go` exercise the restart layout (seeded `kuketty.log`, `capture`, `capture.001.gz`, `metadata.json`, `.meta-*.tmp`, plus a pre-#672 `terminals/<id>/` subdir), the empty-dir no-op, and ENOENT tolerance. The restart-layout test does an end-to-end uid assertion via `syscall.Stat_t.Uid` when running as root (chown observable) and falls back to walk-completes-cleanly when running unprivileged (self-chown no-op). Pass either way.

## Test plan
- [x] `GOWORK=off go test ./internal/controller/runner/` — full runner package green, including all three new `TestChownAttachableStaleChildren_*` cases.
- [x] `GOWORK=off go vet ./...` clean.
- [x] `make test` — full CI target green (both `test-root` and `test-kukebuild` sub-targets).
- [x] `make dev-init` daemon-parity regression guard tail: `default` + `kuke-system` both `Ready`, matching the CLAUDE.md expected output for `kuke get realms` and `kuke get realms --no-daemon`. This is the regression check CLAUDE.md mandates for daemon-touching changes — my diff lives in `internal/controller/runner/attachable.go` which the daemon reads, so the parity check is the load-bearing smoke.
- [ ] `make dev-init` attach-smoke tail (the post-parity `kuke attach` PTY smoke against a busybox cell) — fails on `fix/attachable-chown-stale-tty-files` with `rendered metadata at .../kuketty-metadata.json missing apiVersion v1beta1`, but **reproduces identically on `origin/main` (3130818) under an isolated worktree** — pre-existing project bug, unrelated to this diff. The smoke creates the cell, the kuketty socket binds (the script's 20s wait passes), then the script's `grep '"apiVersion": "v1beta1"'` against the rendered metadata fails. The renderer (`writeKukettyDoc`) is untouched by this PR; my change lives strictly inside `attachablePostCreateChown` and operates on `ttyDir`'s children, never on `containerDir/kuketty-metadata.json` (the file the grep checks, which lives one level up). Filing the pre-existing failure separately via `/reflect`'s verified-bug lane after this PR lands.

Closes #850

### DIFF
--- a/internal/controller/runner/attachable.go
+++ b/internal/controller/runner/attachable.go
@@ -527,7 +527,63 @@ func (r *Exec) attachablePostCreateChown(namespace string, spec intmodel.Contain
 	default:
 		return fmt.Errorf("chown %q to (uid=%d, gid=%d): %w", metadataPath, uid, gid, chownErr)
 	}
+	// Restart-path complement to the ttyDir chown above (#850). The leaf
+	// dir chown covers a fresh provision (MkdirAll just made the dir
+	// empty), but on a restart the dir already holds stale files from the
+	// prior run — kuketty.log (#599), sbsh's capture/metadata.json/.meta-
+	// *.tmp anchored to attachableTTYDir (#672), and any pre-#672
+	// terminals/<id>/ subtree — all root-owned because the daemon wrote
+	// them. When the work image's resolved USER is non-root (e.g. claude:
+	// latest → 1000), kuketty inside the container then hits EACCES at
+	// openTerminalLogger trying to O_RDWR-reopen the stale root-owned
+	// kuketty.log and exits with code 70 before claiming the socket
+	// listener; `kuke attach` later dials the freshly-bound but
+	// listenerless socket and gets "connection refused". Walking and
+	// chowning every leaf child to the resolved container uid covers the
+	// full set of kuketty/sbsh-written inodes without enumerating
+	// basenames, so a future addition under tty/ inherits the same
+	// restart-safety automatically.
+	if err = chownAttachableStaleChildren(ttyDir, int(uid), gid); err != nil {
+		return fmt.Errorf("chown stale children under %q: %w", ttyDir, err)
+	}
 	return nil
+}
+
+// chownAttachableStaleChildren resets the owner of every pre-existing
+// file and directory under root to (uid, gid). It is the restart-path
+// complement to the top-level ttyDir chown in attachablePostCreateChown:
+// on a fresh provision the dir is empty (MkdirAll just produced it) and
+// the walk is a no-op; on a restart it covers every kuketty/sbsh-written
+// leaf inherited from the prior run (#850).
+//
+// The walk uses Lchown so a stray symlink (none today, defense-in-depth)
+// is chowned at the link, never its target. Per-entry ENOENT is tolerated
+// (a concurrent unlink during the walk is a benign race — the inode is
+// already gone, the chown is moot). Subdirectories are walked too: any
+// pre-#672 terminals/<id>/ subtree left over from an older daemon gets
+// covered in the same pass.
+//
+// The root dir itself is intentionally skipped — the caller chowned it
+// explicitly with its own error surface, and re-chowning is wasteful.
+func chownAttachableStaleChildren(root string, uid, gid int) error {
+	return filepath.Walk(root, func(path string, _ os.FileInfo, err error) error {
+		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				return nil
+			}
+			return err
+		}
+		if path == root {
+			return nil
+		}
+		if chownErr := os.Lchown(path, uid, gid); chownErr != nil {
+			if errors.Is(chownErr, os.ErrNotExist) {
+				return nil
+			}
+			return fmt.Errorf("chown %q to (uid=%d, gid=%d): %w", path, uid, gid, chownErr)
+		}
+		return nil
+	})
 }
 
 // attachableSocketSymlinkDirMode is the mode applied to <RunPath>/s when
--- a/internal/controller/runner/attachable_test.go
+++ b/internal/controller/runner/attachable_test.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"testing"
 
 	"github.com/eminwux/kukeon/internal/consts"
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #856 Review — `fix(attachable): chown stale tty children on restart so non-root kuketty can re-open them`**

LGTM — focused #850 fix that walks the tty leaf dir and Lchowns every stale kuketty/sbsh inode to the resolved container uid, with ENOENT-tolerant race handling and root-skip on the helper; tests cover restart layout, fresh-provision no-op, and the ENOENT path; CI green and the daemon-parity smoke is in the test plan.

PR #858, opened 2026-05-25
fix(attach): extend liveness guard to refuse non-Ready cell state

### LINKED ISSUE #852: attach: extend liveness guard to refuse non-Ready cell state (currently #683 only refuses on Ready)
## Context

`GuardCellTaskLiveness` (`cmd/kuke/shared/cell_liveness.go:42`, added in #683) short-circuits when the cell's persisted state is not `Ready`:

```go
if get.Cell.Status.State != v1beta1.CellStateReady {
    return nil   // passes through for Stopped / NotCreated / Failed / Unknown
}
```

The intent (per #683's description) was to defend the post-reboot case where on-disk state still reads `Ready` but the task is gone. But the reconciler's wind-down (`refresh.go:741 windDownCell`) actively flips a Ready cell to Stopped once its non-root work container exits — which is the steady state both after the chown failure (related issue) and after any clean exit of an attachable shell. In that Stopped state, `kuke attach` then:

1. passes `GuardCellTaskLiveness` (state is Stopped, not Ready);
2. calls `client.AttachContainer` (`internal/client/local/local.go:1199`), which only validates that the spec is Attachable and returns the host socket path — no task-liveness check;
3. dials the orphan socket inode → `connection refused`.

The actionable error #683 promised the operator — "delete it with `kuke delete cell` before re-running" — never reaches them on the path that's actually most common after a reboot.

The orphan inode itself is a second, distinct concern: kuketty's `claimSocketListener` (`cmd/kuketty/main.go:259`) removes-then-binds, so a kuketty that does run cleans up its own predecessor's inode. But when kuketty never runs again (the cell is wound down, the daemon is restarted, the host reboots), the file persists across the host, and the dead inode is the trigger that makes the missing guard visible to the operator.

### Observed end-to-end

```
$ kuke attach kukeon-pm-0          # cell in Stopped state
```

Expected:

```
$ kuke attach kukeon-pm-0
Error: cell "kukeon-pm-0" is in state Stopped (no work container is running); start it first with `kuke start cell kukeon-pm-0`
```

## Proposal

Three changes, all small:

1. **Generalize the liveness guard** to refuse on any non-Ready cell state (Stopped, NotCreated, Failed, Unknown), each with a state-appropriate recovery pointer:
   - `Stopped` → "start it first with `kuke start cell <name>`"
   - `Failed` / `Unknown` → keep the existing "delete it with `kuke delete cell <name>` before re-running" message
   - `NotCreated` → "create it first with `kuke run -p <profile> --name <name>`"

   Both `kuke run`'s Ready-short-circuit and `kuke attach` should call the generalized form.

2. **Unlink the host-side attach socket inode on the way down**, on every clean teardown — `KillCell`, `stopCell`, `windDownCell`. Defense-in-depth complement to (1) for cases where state lookup races a wind-down tick or where the on-disk state is itself stale. The next `kuke attach` then gets ENOENT instead of `connection refused`, which surfaces a cleaner error class regardless of where the divergence came from.

3. **Add a server-side task-liveness check inside `AttachContainer`** at `internal/client/local/local.go:1199`. Today it validates only that the spec exists and is Attachable, then returns the host socket path with no task-liveness gate — so any consumer that bypasses the CLI's `GuardCellTaskLiveness` (RPC clients, scripts, future API consumers, in-process callers in `kuke run`'s detach-first path) still receives a path backed by nothing. The check should mirror the client-side guard's predicates (task is Running) and return a typed error the wire/local-client paths can both surface. This is the third layer of the defense-in-depth — client guard catches the common case (1), socket cleanup catches the race (2), server check catches everything else.

The three changes are scoped to one issue because they fix the same operator-facing UX from three angles. Implementer is free to land them as one PR or split if the testing surfaces want it.

## Acceptance criteria

- [ ] `kuke attach <stopped-cell>` exits non-zero with a message that names the state and the recovery command, instead of dialing the socket.
- [ ] `kuke attach <not-created-cell>` (cell missing in containerd) exits non-zero with the "create it first" pointer.
- [ ] `kuke attach <ready-but-task-dead-cell>` continues to refuse with the existing #683 diverged-state message (no regression on that path).
- [ ] After a clean `kuke stop cell` or after the reconciler's wind-down completes, the host-side attach socket inode is removed.
- [ ] `AttachContainer` (server-side) returns a typed task-liveness error when the cell's work task is not Running, independent of which client called it. A direct RPC call against a non-Running cell does not return a usable socket path.
- [ ] Tests cover the four states (Ready+live, Ready+dead-task, Stopped, Failed) and the post-teardown socket cleanup, exercising both the client-side guard and the server-side `AttachContainer` check.

## Notes

- Code anchors: `cmd/kuke/shared/cell_liveness.go:42 GuardCellTaskLiveness`; `cmd/kuke/attach/attach.go:131`; `cmd/kuke/run/run.go` (Ready short-circuit); `internal/client/local/local.go:1199 AttachContainer`. Socket-cleanup anchor: `internal/controller/runner/kill.go` (and the wind-down path through `refresh.go:741`).
- Extends #683 from "task-dead Ready" to "task-dead in any state" + adds the symmetric inode cleanup.

eminwux:
## Summary

- Generalize `GuardCellTaskLiveness` (`cmd/kuke/shared/cell_liveness.go`) to refuse on every non-Ready cell state — `NotCreated` (`!MetadataExists`), `Stopped`, `Pending`, `Failed`, `Unknown` — each with a state-appropriate recovery pointer (`kuke run` for NotCreated, `kuke start cell` for Stopped, `kuke delete cell` for Failed/Unknown/Pending). The `Ready` + `RootContainerTaskRunning` happy path is unchanged, and the `Ready` + task-dead branch keeps #683's "diverged" wording verbatim.
- Reorder `kuke attach` (`cmd/kuke/attach/attach.go`) to call the cell-liveness guard **before** `PickContainer`, so a missing cell surfaces "create it first" instead of falling out of the picker with `ErrAttachNoCandidate`.
- Unlink the host-side attach socket inode **and** the SUN_PATH-safe symlink on every clean teardown — `KillCell`, `StopCell`, and (transitively) the reconciler's `windDownCell` — via a new `removeAttachableSocketRuntimeArtifacts` helper next to the existing `removeAttachableSocketSymlink`. Defense-in-depth so any consumer that slips past the guards gets `ENOENT` instead of `connection refused` from a dead inode.
- Add a server-side task-liveness check inside `(*local.Client).AttachContainer` (`internal/client/local/local.go`) — refuses with a new typed `errdefs.ErrAttachTaskNotRunning` (wired through `kukeonv1/errmap.go` so it survives the wire round-trip) when the target container's task is not `Running`. Catches RPC clients, scripts, and in-process callers that bypass the CLI guard.
- Tests cover the four cell states (Ready+live, Ready+dead-task, Stopped, Failed, plus NotCreated/Unknown/Pending for completeness) on both the shared guard and the `kuke attach` CLI plumbing, the server-side `guardAttachableTaskLiveness` predicate (Ready passes; Stopped/Failed/Unknown/Pending/NotCreated all return `ErrAttachTaskNotRunning`), and the post-teardown socket cleanup (both inodes removed; idempotent on repeat; short-circuits on non-Attachable specs).

## Test plan

- [x] \`make test\` (full repo + kukebuild) — all packages pass.
- [x] \`GOWORK=off go vet ./...\` — clean.
- [x] Targeted: \`go test ./cmd/kuke/shared/... ./cmd/kuke/attach/... ./internal/client/local/... ./internal/controller/runner/...\` — all new cases pass.
- [ ] \`make dev-init\` smoke not run — the change is contained to the attach guard / teardown cleanup / server-side liveness check and does not touch the daemon bootstrap path, the build path, or anything under \`/opt/kukeon\` provisioning. Reviewer can re-run the project's smoke on a host with containerd available if they want the daemon-parity tail re-checked.
- [ ] Manual repro of the issue's end-to-end \`kuke attach <stopped-cell>\` scenario not run on this host (no live cell in a wound-down state to attach to). The behavior is covered by the new \`TestAttach_Stopped_PointsAtStart\` unit case driving the same code path.

## Notes

- AC #3 specifies the server-side check should "mirror the client-side guard's predicates (task is Running)". The client-side guard checks the **cell's root** task liveness as a proxy for the cell being alive; the server-side check operates at the **attachable container's own** task level — that's the more precise predicate at the per-`AttachContainer` boundary (the kuketty process that binds the socket runs inside that specific container, not the cell's root). Documented in `local.go`'s `guardAttachableTaskLiveness` doc.
- The unlink helper removes both the symlink (`<RunPath>/s/<hash>`) and the deep socket inode (`<ContainerTTYDir>/socket`) — the AC mentions "host-side attach socket inode" specifically; removing the symlink alone would give ENOENT at `connect(2)` time, but removing the inode too keeps the cell's bind-mount source clean for the next kuketty boot (kuketty's `claimSocketListener` removes-then-binds either way, so unlinking ahead of the next start is safe).
- `Pending` is folded into the "delete it" branch alongside Failed/Unknown. The reconciler labels `Pending` transiently mid-creation; from the attach caller's perspective at attach time it means the cell is not yet usable, and there's no clean "wait for it" pointer that fits — the operator deletes and re-runs.

Closes #852

### DIFF
--- a/cmd/kuke/attach/attach.go
+++ b/cmd/kuke/attach/attach.go
@@ -110,6 +110,20 @@ func runAttach(cmd *cobra.Command, args []string) error {
 	}
 	defer func() { _ = client.Close() }()
 
+	// Cell-level liveness guard (#683, #852): refuse to attach to a cell
+	// whose state is anything other than Ready+task-live. Runs before the
+	// container picker so a missing cell surfaces a "create it first"
+	// pointer instead of falling out of ListContainers with
+	// ErrAttachNoCandidate, and so a Stopped / Failed cell short-circuits
+	// before the attach loop dials a now-orphan socket inode.
+	cellGet, err := client.GetCell(cmd.Context(), buildCellDoc(cell, realm, space, stack))
+	if err != nil {
+		return err
+	}
+	if err := kukeshared.GuardCellTaskLiveness(cellGet, cell); err != nil {
+		return err
+	}
+
 	if container == "" {
 		container, err = kukeshared.PickContainer(cmd.Context(), client, realm, space, stack, cell,
 			func(spec v1beta1.ContainerSpec) bool {
@@ -118,19 +132,6 @@ func runAttach(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
-	}
-
-	// Divergence guard (#683): refuse to attach to a cell whose metadata records
-	// it Ready but whose root-container task is gone from containerd (post-reboot
-	// divergence). Without this, attach hands back a socket backed by a dead task
-	// and the sbsh loop fails with `connection refused`. Shares the run path's
-	// guard so both entry points refuse identically.
-	cellGet, err := client.GetCell(cmd.Context(), buildCellDoc(cell, realm, space, stack))
-	if err != nil {
-		return err
-	}
-	if err := kukeshared.GuardCellTaskLiveness(cellGet, cell); err != nil {
-		return err
 	}
 
 	doc := buildContainerDoc(container, realm, space, stack, cell)
--- a/cmd/kuke/attach/attach_test.go
+++ b/cmd/kuke/attach/attach_test.go
@@ -597,3 +597,151 @@ func TestAttach_MissingCellArg(t *testing.T) {
 		t.Errorf("error %q does not mention arg-count requirement", err.Error())
 	}
 }
+
+// TestAttach_NotCreated_PointsAtRun covers #852: the cell-level liveness
+// guard fires before ListContainers, so a cell that has never been
+// created surfaces the operator-actionable "create it first" pointer
+// instead of falling out of the container picker with the unhelpful
+// ErrAttachNoCandidate. Also verifies the guard short-circuits before
+// reaching AttachContainer.
+func TestAttach_NotCreated_PointsAtRun(t *testing.T) {
+	t.Cleanup(viper.Reset)
+
+	listCalls := 0
+	attachCalls := 0
+	fc := &fakeClient{
+		getCellFn: func(_ v1beta1.CellDoc) (kukeonv1.GetCellResult, error) {
+			return kukeonv1.GetCellResult{MetadataExists: false}, nil
+		},
+		listContainersFn: func(_, _, _, _ string) ([]v1beta1.ContainerSpec, error) {
+			listCalls++
+			return nil, nil
+		},
+		attachContainerFn: func(_ v1beta1.ContainerDoc) (kukeonv1.AttachContainerResult, error) {
+			attachCalls++
+			return kukeonv1.AttachContainerResult{}, nil
+		},
+	}
+	run := &runCapture{}
+	cmd := newCmdWithCtx(t, fc, run)
+	cmd.SetArgs([]string{
+		"--realm", "r1", "--space", "s1", "--stack", "st1", "kukeon-pm-0",
+	})
+
+	err := cmd.Execute()
+	if err == nil {
+		t.Fatal("Execute returned nil on missing cell, want NotCreated error")
+	}
+	if got := err.Error(); !strings.Contains(got, "does not exist") {
+		t.Errorf("error %q missing %q wording", got, "does not exist")
+	}
+	if got := err.Error(); !strings.Contains(got, "kuke run") {
+		t.Errorf("error %q missing `kuke run` recovery pointer", got)
+	}
+	if listCalls != 0 {
+		t.Errorf("ListContainers called %d times, want 0 (guard must fire first)", listCalls)
+	}
+	if attachCalls != 0 {
+		t.Errorf("AttachContainer called %d times, want 0", attachCalls)
+	}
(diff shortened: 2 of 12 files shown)


eminwux:
**PR #858 Review — `fix(attach): extend liveness guard to refuse non-Ready cell state`**

LGTM — generalizes the Ready+task-dead guard from #683 into a full state switch with per-state operator pointers, layers a server-side `ErrAttachTaskNotRunning` for RPC bypass, and unlinks both socket inodes on every clean teardown so post-guard slip-throughs land on `ENOENT` instead of `connection refused`. `#683`'s "diverged" wording is preserved verbatim, the run path stays scoped to the Ready branch (broader switch is `kuke attach`-only as documented), `windDownCell → killCellLocked` carries the new cleanup transitively, all five terminal states have unit + CLI coverage, and the new error sentinel routes through `errmap.go` so wire-round-trip `errors.Is` keeps working. The two unchecked test-plan items are documented host-gated (`make dev-init` needs containerd; live wound-down repro needs a real cell) — operator can run those pre-merge.

PR #859, opened 2026-05-25
fix(start): probe task liveness before marking cell Ready

### LINKED ISSUE #851: StartCell marks cell Ready without verifying the work task survives the start; wind-down then races the caller's attach
## Context

`(*Exec).StartCell` (`internal/controller/runner/start.go:297`) calls `markCellReady(&internalCell)` at line 950, immediately after the per-container `ctrClient.StartContainer` loop returns. There is no post-start liveness check: the function trusts that "containerd accepted task creation" means the task is, and will remain, running.

When the work container's process exits within milliseconds — e.g. kuketty fails to open its log file (see the related chown issue), an `onInit` script exits the shell, runc returns success but the workload immediately faults — `StartCell` still returns success and the controller stamps `state: Ready` on disk.

The next reconciler tick (default interval is a few seconds) runs `deriveCellStateFromNonRootContainerStatuses` (`refresh.go:943`), sees the work container as Stopped, flips the cell back to Stopped, and `shouldWindDownCell` (`refresh.go:854`) fires `KillCell` — tearing down the root container, removing the containerd records, and leaving the orphaned attach socket file the operator's `kuke run` was about to dial.

From the operator's perspective:

```
Cell "kukeon-pm-0" (realm "default", space "default", stack "default")
  - metadata: already existed
  - cgroup: already existed
  - root container: missing
  - containers: started                                   # StartCell-success print
                                                          # the cell is already gone
```

`kuke run` prints `containers: started` between the start and the attach, then attach fails. #654 fixed the *records-missing-from-containerd* divergence so `containers: already existed` no longer lies on that path, but it did not address this *records-present-but-task-exited-immediately* path.

The asymmetry today: `markCellFailed` already exists (`start.go:237`) with a fully wired transition through containerd-touching failures during the start. What's missing is using it for the "container started then immediately exited" outcome that lands between `StartContainer` returning success and the reconciler's first tick.

## Proposal

Before `markCellReady`, perform a brief liveness check against every just-started non-root container — long enough to catch a fast-exit but short enough that a healthy start sees no observable latency. On any non-root container in a non-Running state, return a typed error (e.g. `ErrCellWindDownImmediate`) so the existing defer at `start.go:315` routes through `markCellFailed("StartCellFailed", …)` — the same path `provisionStarted=true` already enables for the runc-level failures earlier in the function.

The CLI surface would change from a misleading `containers: started` + `connection refused` to a clear failure like:

```
Cell "kukeon-pm-0" (realm "default", space "default", stack "default")
  - ...
  - containers: failed (work exited immediately, code=70)
Error: cell kukeon-pm-0 started but work container exited within startup grace; run `kuke log container kukeon-pm-0/work` for details
```

Grace-window duration, polling interval, and failure-message wording are implementer's choice — what matters is that a healthy start doesn't slow down and a fast-exiting container can no longer masquerade as Ready.

## Acceptance criteria

- [ ] When a non-root container exits within a bounded post-start grace window, `StartCell` returns a non-nil error, the cell transitions to `Failed` via the existing `markCellFailed` path, and the CLI prints non-zero with a pointer to `kuke log container`.
- [ ] Clean starts: a cell whose work task survives the grace window persists `state: Ready` and the CLI prints `containers: started` — no behavior change on the happy path.
- [ ] Tests cover: (a) work container exits with non-zero code immediately after StartContainer; (b) work container survives the grace window; (c) root container exits immediately (currently unhandled — the derivation loop only checks non-root); (d) the same liveness check covers `(*Exec).StartContainer` (single-container variant at `runner/start.go:968`), not only `StartCell` — both are operator-reachable verbs (`kuke start cell` and `kuke start container`) and share the same silent-success failure mode.
- [ ] No measurable startup latency added to the healthy path (grace window short, polling cheap).

## Notes

- Code anchor: `internal/controller/runner/start.go:950 markCellReady` and the surrounding non-root start loop. `markCellFailed` already exists at `start.go:237`.
- Closely related closed work: #654 (records-missing-from-containerd case on the same CLI path), #683 (post-start attach guard for Ready+task-dead).

eminwux:
## Summary

- StartCell / StartContainer used to call `markCellReady` immediately after `ctrClient.StartContainer` returned. A task that exits within milliseconds — kuketty failing to open its log file, an `onInit` script returning non-zero, runc returning success but the entrypoint immediately faulting — would still land on disk as `state: Ready`. The next reconciler tick (every few seconds) derives `Stopped` from the dead workload, `shouldWindDownCell` fires `KillCell`, and the operator's in-flight `kuke run --attach` races the teardown — printing `containers: started` and then dialing a socket that no longer exists.
- Add a bounded post-start liveness probe (300 ms grace window, 50 ms polls) that calls `ctrClient.TaskStatus` against the just-started root + non-root containers. The first probe that observes a task not in `containerd.Running` (or whose `TaskStatus` call errors) fails fast with a wrapped `errdefs.ErrCellWindDownImmediate`; the existing `provisionStarted` defer in `startCellLocked` / `StartContainer` then routes through `markCellFailed("StartCellFailed", retErr)` / `markCellFailed("StartContainerFailed", retErr)` — the same code path runc-level startup failures already use, so the cell lands at `Failed` instead of the misleading `Ready → Stopped → reaped` cycle.
- Cover the AC's root-container exit branch (item (c)) too: the existing `deriveCellStateFromNonRootContainerStatuses` reconciler derivation only walks non-root containers, so a dead root would otherwise slip past the wind-down guard. The new probe is the only place that catches it before `markCellReady`.

## Test plan

- [x] `make test` (test-root + test-kukebuild) — all green
- [x] `make dev-init` smoke — daemon parity check + nested `kuke attach` smoke both pass; expected two-realm tail observed
- [x] `pre-commit run --files <changed>` — passes
- [x] New unit coverage in `internal/controller/runner/start_test.go`:
  - `TestFirstNonLiveContainer` — table-driven per-probe decision matrix (AC items (a), (c), defensive empty-ContainerdID, Root: true skip, first-stopped-wins ordering, statusFn errors)
  - `TestVerifyCellTasksLiveAfterStart` — wraps the per-probe decision with the time-bounded poll loop: healthy start observes all-Running through the window (AC (b)), fast-exit returns inside the window with a wrapped `ErrCellWindDownImmediate` (AC (a)), root exit inside the window also flagged (AC (c)), `graceWindow=0` disables the guard
  - `TestVerifyCellTasksLiveAfterStart_SingleContainerVariant` — pins AC (d): the same probe shape covers `StartContainer`'s single-container slice + root pair
  - `TestNonRootContainerSpecsFromCell` — the adapter that strips `Root: true` entries before the probe
- [x] `TestRecreateCell_StartsTasksBeforeReady` (issue #682 regression guard) updated: the test's fake `ctr.Client` now tracks which container IDs have been "started" so its `TaskStatus` returns `Running` for the freshly-started root (so the new probe lets `markCellReady` stamp) while still returning zero-value for the pre-existing root (so the `cellTasksAllRunningFn` idempotency guard at `start.go:414` does *not* short-circuit and the destructive teardown-and-recreate path the AC pins still runs)

## Implementation notes for the reviewer

- `cellStartLivenessGraceWindow = 300ms` / `cellStartLivenessPollInterval = 50ms` are package-private constants (call-site immutable). The probe pays the full grace window on the happy path — bounded and the proposal says "implementer's choice"; 300 ms is small enough that it falls well inside the existing reconcile interval and the e2e suite's startup variance.
- `verifyCellTasksLiveAfterStart` is decoupled from real time (`nowFn` / `sleepFn`) and from `*Exec`'s concrete containerd client (`statusFn` closure), matching the shape of `cellTasksAllRunningFn` / `teardownRootContainerCNI` so both the fail-fast and observe-through-window branches are exercised without a live runtime.
- `firstNonLiveContainer` treats a `statusFn` error the same as a non-Running status — a freshly-started task whose record vanished mid-check is exactly the wind-down-then-race condition the guard is defending against. Also treats an empty `ContainerdID` on a non-root spec as not-live (defensive; matches `cellTasksAllRunningFn`'s treatment of the same case).
- The `StartContainer` (single-container variant) wiring deliberately probes the just-started container *and* the root: a root that died between the cell coming up and this `StartContainer` call would otherwise stamp `Ready` despite a dead root. Operator-reachable as `kuke start container`.
- Error message names the cell, the offending container (or `root container "<containerdID>"` for the auto-default-root case), the observed status / probe error, and points at `kuke log <cell> [--container <id>]`.

Closes #851

### DIFF
--- a/internal/controller/runner/recreate_cell_test.go
+++ b/internal/controller/runner/recreate_cell_test.go
@@ -151,10 +151,28 @@ func TestRecreateCell_StartsTasksBeforeReady(t *testing.T) {
 	)
 
 	var startContainerCalls int
+	started := map[string]bool{}
 	fake := &recreateCellFakeClient{
-		deleteCellFakeClient: &deleteCellFakeClient{},
-		startContainerFn: func(_ string, _ ctr.ContainerSpec, _ ctr.TaskSpec) (containerd.Task, error) {
+		deleteCellFakeClient: &deleteCellFakeClient{
+			// Track per-container task liveness across two StartCell guards:
+			// (1) cellTasksAllRunningFn's idempotency skip (issue #149) — must
+			// NOT short-circuit here, because the existing cell's root has
+			// never been "started" in this fake, so its TaskStatus returns
+			// zero-value (not Running); the test then exercises the
+			// destructive teardown-and-recreate path the AC pins.
+			// (2) verifyCellTasksLiveAfterStart's post-start liveness probe
+			// (issue #851) — once startContainerFn runs, the new root task
+			// should observe as Running so the probe lets markCellReady stamp.
+			taskStatusFn: func(_, id string) (containerd.Status, error) {
+				if started[id] {
+					return containerd.Status{Status: containerd.Running}, nil
+				}
+				return containerd.Status{}, nil
+			},
+		},
+		startContainerFn: func(_ string, spec ctr.ContainerSpec, _ ctr.TaskSpec) (containerd.Task, error) {
 			startContainerCalls++
+			started[spec.ID] = true
 			return recreateCellTask{pid: 4242}, nil
 		},
 	}
--- a/internal/controller/runner/start.go
+++ b/internal/controller/runner/start.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"time"
 
 	containerd "github.com/containerd/containerd/v2/client"
 	"github.com/containerd/containerd/v2/pkg/namespaces"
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #859 Review — `fix(start): probe task liveness before marking cell Ready`**

LGTM — the bounded post-start probe (root-first, then non-root, fail-fast on `containerd.Running` mismatch or `TaskStatus` error) plugs the AC (c) root-exit gap that `deriveCellStateFromNonRootContainerStatuses` couldn't see, the wrapped `ErrCellWindDownImmediate` flows cleanly through the existing `provisionStarted` defer into `markCellFailed`, both call sites (`StartCell` + `StartContainer` single-container variant) are wired symmetrically, the `nowFn`/`sleepFn`/`statusFn` decoupling lets the table tests pin all four AC branches without a live runtime, and the `TestRecreateCell_StartsTasksBeforeReady` update threads the new probe past the pre-existing `cellTasksAllRunningFn` idempotency guard without short-circuiting the destructive recreate path it was originally pinning.

Issue #854, opened 2026-05-24
post-reboot recovery: attachable cells with non-root USER end up unreachable after host reboot (umbrella)


eminwux:
## Context

After a host reboot, cells whose attachable container runs as a non-root image USER end up in a stuck state: `kuke get cell` reports `state: Stopped` with `cgroupReady: true`, `kuke run` reports `containers: started` but immediately fails with `connection refused` on the kuketty socket. The same sequence reproduces on every Stopped cell whose work image is `claude:latest` (non-root); does **not** reproduce on `claude-root:latest` (root) cells.

Investigation against `kukeon-pm-0` traced the failure to five small gaps, each in a different layer:

1. `attachablePostCreateChown` doesn't re-chown pre-existing log/capture inodes in the tty dir, so kuketty's `openTerminalLogger` fails on the stale root-owned `kuketty.log`, and kuketty exits before binding the socket → silent failure with no log entry to grep for. → **#850**
2. `StartCell` calls `markCellReady` immediately after the runner's `StartContainer` returns, without checking the task survives more than a few ms. The wind-down reconciler then races the operator and reaps the cell. → **#851**
3. `GuardCellTaskLiveness` (the #683 fix) only fires for state == Ready. After the wind-down flips the cell to Stopped, the guard does nothing, and `kuke attach` dials the orphan socket inode and surfaces `connection refused` — a UX cliff with no actionable error. → **#852**
4. The background reconciler never updates `Status.CgroupReady`. After a reboot the cgroup tmpfs is wiped, but every cell that was Ready before still reports `cgroupReady: true` indefinitely, hiding the divergence from `kuke get cell -o yaml` and the `kuke status` check. → **#853**
5. The background reconciler also never *heals* a wiped cgroup — `ensureCellCgroup` runs only at `CreateCell` time, so cells with surviving metadata but no cgroup stay broken until the operator runs `kuke start cell <name>` per cell. → **#855**

Each is small in isolation; together they produce the silent-failure UX captured in the investigation transcript.

### Repro signature (this session)

```
$ kuke run -p kukeon-pm --name kukeon-pm-0
Cell "kukeon-pm-0" (realm "default", space "default", stack "default")
  - metadata: already existed
  - cgroup: already existed
  - root container: missing
  - container "root": already existed
  - container "work": already existed
  - containers: started
Error: error attaching client: ping failed: dial unix /opt/kukeon/s/97504a301eb3a71f: connect: connection refused
```

[...]

PR #861, opened 2026-05-25
fix(reconcile): write Status.CgroupReady from ExistsCgroup result

### LINKED ISSUE #853: reconcile: Status.CgroupReady never updated by the background loop; persists true after a reboot wipes the cgroup
## Context

The background reconcile loop (`(*Exec).ReconcileCell` at `internal/controller/runner/refresh.go:632`) calls `r.ExistsCgroup(cell)` at line 663 but only uses the **error** return to decide whether to derive cell state or park at Unknown. It never writes the **boolean** result to `cell.Status.CgroupReady`. The flag is monotonic-write-once: set to `true` by `ensureCellCgroup` at create time, and never falsified by any subsequent reconcile tick.

`(*Exec).RefreshCell` (the on-demand `kuke refresh` path, `refresh.go:356`) does keep the flag in sync — see lines 369–374 and the diff check at line 434. So the bug is precisely that the **background** reconciler does not close the same loop that the **on-demand** refresh already does.

After a host reboot the cgroup tmpfs is wiped. The reconciler quietly re-derives cell state (post-#543 fix), but `cgroupReady: true` stays stamped on every cell that was Ready before the reboot. Verified directly on the host:

```
$ ls /sys/fs/cgroup/kukeon/default/default/default/kukeon-pm-1/
ls: cannot access ...: No such file or directory

$ sudo cat /opt/kukeon/data/.../kukeon-pm-1/metadata.json \
    | python3 -c "import sys,json;print(json.load(sys.stdin)['status']['cgroupReady'])"
True

$ kuke get cell kukeon-pm-1 -o yaml | grep cgroupReady
  cgroupReady: true
```

Same divergence reproduced on 13 of 19 cells in a single host across one reboot.

`kuke status` consults `cgroupReady` as part of the state-consistency check; the current bug means that check is silently bypassed for any cell whose cgroup was wiped by a reboot but never re-derived through the on-demand `RefreshCell` path.

## Proposal

In `ReconcileCell`, take the boolean from `r.ExistsCgroup(cell)` too, set `cell.Status.CgroupReady = cgroupExists` (false on the call-error path is fine — `RefreshCell` does this implicitly via its struct default at `refresh.go:361-364`), and include the diff in the `updated` boolean at `refresh.go:706-718` so the metadata write is triggered.

One-function change, ~5 lines, mirrors the on-demand `RefreshCell` exactly.

## Acceptance criteria

- [ ] After a host reboot, a single background reconcile tick falsifies `Status.CgroupReady` on every cell whose `/sys/fs/cgroup/.../<cell>` directory is gone.
- [ ] After `kuke start cell` (or any path that re-materializes the cgroup), the next reconcile tick re-sets `Status.CgroupReady: true`.
- [ ] No regression on `RefreshCell`'s on-demand path (which already does this correctly).
- [ ] Test asserts the round trip: seed a cell with `cgroupReady: true` + no cgroup on disk, run one reconcile tick, assert the metadata write flips the flag false.

## Notes

- Code anchors: `internal/controller/runner/refresh.go:632 ReconcileCell` (cgroup check at line 663, missing write); `refresh.go:369-374 RefreshCell` (the parallel that already does it); the diff predicate at `refresh.go:706-718`.
- Closely related closed work: #543 (added the post-reboot derivation fallback that this completes).

eminwux:
## Summary
- The background reconcile loop (`ReconcileCell` at `internal/controller/runner/refresh.go:632`) called `r.ExistsCgroup(cell)` but only used the **error** return, never the **boolean**. `Status.CgroupReady` was therefore monotonic-write-once: set true at create time by `ensureCellCgroup` and never falsified. After a host reboot wipes the tmpfs-backed `/sys/fs/cgroup/.../<cell>` tree, every cell that was Ready before the reboot kept `cgroupReady: true` stamped on it forever, silently bypassing `kuke status`'s state-consistency check.
- Fix takes the boolean from `ExistsCgroup` too and writes `cell.Status.CgroupReady = cgroupErr == nil && cgroupExists`, mirroring exactly what the on-demand `RefreshCell` path already does at `refresh.go:369-374`. Also extends the `updated` diff predicate at `refresh.go:706-718` to include the `CgroupReady` flip so the metadata write fires.
- `RefreshCell`'s on-demand path is untouched (it was already correct).

## Test plan
- [x] `make test` — full project test suite green, including the new `TestReconcileCell_PostReboot_FalsifiesCgroupReady` and the existing `TestReconcileCell_PostReboot_TransitionsToStopped` (#543's complement).
- [x] New test seeds a cell with `cgroupReady: true` + no cgroup on disk, runs one `ReconcileCell` tick, asserts both the returned cell and the persisted metadata file flip `CgroupReady` to false — covering AC#4's round-trip.
- [ ] AC#1 / AC#2 host-level smoke (reboot the host, observe a single reconcile tick falsifies `cgroupReady` across surviving cells; then `kuke start cell` re-materializes the cgroup and the next tick re-sets `cgroupReady: true`) — deferred to a host with a real kernel reboot cycle. The unit test exercises the same code paths via the fake `ctr.Client`; in-package coverage gates the regression.

## Notes
- One-function change, ~9 lines + a new test. Mirrors the on-demand `RefreshCell` path exactly per the issue proposal.
- Part of the post-reboot recovery umbrella `epic:reboot` (#854), completing the loop #543 started (#543 added the post-reboot state-derivation fallback; #853 closes the matching probe-readiness write).

Closes #853

### DIFF
--- a/internal/controller/runner/container_state_test.go
+++ b/internal/controller/runner/container_state_test.go
@@ -280,6 +280,7 @@ func seedPostRebootCell(
 		Status: v1beta1.CellStatus{
 			State:         v1beta1.CellStateReady,
 			ReadyObserved: true,
+			CgroupReady:   true,
 		},
 	}
 	path := fs.CellMetadataPath(r.opts.RunPath, realm, space, stack, cellName)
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #861 Review — `fix(reconcile): write Status.CgroupReady from ExistsCgroup result`**

LGTM — the fix takes the `cgroupExists` bool that the background reconciler was already throwing away and writes it through to `cell.Status.CgroupReady` with the same `err==nil && exists` semantics the on-demand `RefreshCell` path already used (refresh.go:368-373), and the `updated` diff predicate is extended in the same hunk so the post-reboot true → false flip actually reaches the metadata file. New `TestReconcileCell_PostReboot_FalsifiesCgroupReady` round-trips through `GetCell` to verify the persisted side too, which is exactly the user-visible surface the bug exposed.

Deferred per the PR body: the AC#1/AC#2 real-reboot host smoke is the one to run pre-merge once a host with a kernel reboot cycle is available — the unit test covers the code path via the fake `ctr.Client`, but the end-to-end "reboot wipes /sys/fs/cgroup, single reconcile tick falsifies cgroupReady across surviving cells" is the user-facing claim.

Issue #862, opened 2026-05-25
chore(claude.md): document post-reboot cgroup heal contract in smoke-test section


eminwux:
## Context

#855 lands the post-reboot cgroup heal in the background reconciler: when a cell's persisted `Status.ReadyObserved` latch is true and the cgroup check returns absent, `ReconcileCell` re-runs `ensureCellCgroup` to re-create the cgroup directory and re-assert subtree controllers. Cells that never reached Ready (a half-CreateCell that crashed mid-way) are not promoted by the re-ensure.

#855's AC #1 asks for the chosen behavior to be documented in the project `CLAUDE.md`'s smoke-test contract. Dev's role rule forbids `<project>/CLAUDE.md` edits in a code-fix PR (it travels as its own project issue + PR), so this issue carries the deferred doc work.

## Proposal

Add a "Post-reboot cgroup recovery" subsection under the existing local-smoke-test block in `CLAUDE.md` that:

- Names the reconcile-loop heal contract (re-ensure when cgroup absent AND ReadyObserved is true) and the half-CreateCell exclusion (re-ensure skipped when ReadyObserved is false, so an in-flight CreateCell finishes its own ensureCellCgroup path under the per-cell lock).
- Mentions that subtree controllers are re-asserted on the same pass (the second half of the post-#314/#328 contract).
- States that `kuke status` after a reboot shows zero FAIL rows on the cell-cgroup row for cells whose metadata exists — once the next reconcile tick completes — regardless of whether the reconciler healed them or the operator did via `kuke start cell <name>`.
- Notes that the `make dev-init` smoke does **not** exercise reboot recovery (per umbrella #854's Notes); reboot recovery is exercised by the unit tests in `internal/controller/runner/reconcile_heal_test.go`.

Verbatim proposed wording for the new subsection (sized to slot under \"### Post-init: \`kuke status\`\"):

> ### Post-reboot cgroup recovery
>
> A host reboot wipes the cgroup tmpfs while cell metadata under \`/opt/kukeon/data/...\` survives. The background reconciler (`(*Exec).ReconcileCell`) re-runs `ensureCellCgroup` on the next tick for any cell whose cgroup is absent AND whose persisted `Status.ReadyObserved` is true — re-creating the cgroup directory and re-asserting subtree controllers in the same pass.
>
[...]

PR #863, opened 2026-05-25
fix(reconcile): heal wiped cell cgroup on the ReadyObserved latch

### LINKED ISSUE #855: reconcile: missing cell cgroup is not re-provisioned by the background loop; reboot leaves cells with no cgroup until operator runs kuke start cell
## Context

`ensureCellCgroup` (`internal/controller/runner/provision.go:1264`) is only called from `CreateCell` (`internal/controller/runner/create_cell.go:151`). The background reconciler (`(*Exec).ReconcileCell` at `refresh.go:632`) calls `r.ExistsCgroup(cell)` at line 663 only to decide between deriving cell state and parking at Unknown — it never tries to re-create the missing cgroup.

After a host reboot, the cgroup tmpfs is wiped. Cell metadata under `/opt/kukeon/data/...` survives. The reconciler then sees the cells, correctly derives `state: Stopped` from `populateCellContainerStatuses` (the post-#543 fallback), but leaves the cgroup absent. The only path that brings cell cgroups back is `kuke start cell <name>` — which works because runc honors the OCI `linux.cgroupsPath` and creates the dir during task launch.

Verified on a host post-reboot (13 of 19 cells affected):

```
$ ls /sys/fs/cgroup/kukeon/default/default/default/ | grep -E 'kukeon-pm|crew|meta|sbsh'
kukeon-pm-0
kukeon-pm-1
kukeon-pm-2
kukeon-pm-ff6e0f
# 13 other cells (crew-dev-0/1, crew-pr-0, kukeon-dev-1/2, kukeon-pr-0/1/2,
# meta-0..3, sbsh-pm-0) have no cgroup directory at all

$ kuke get cells | grep -E 'crew-dev-0|sbsh-pm-0'
crew-dev-0  default  default  default  Stopped  -  /kukeon/default/default/default/crew-dev-0
sbsh-pm-0   default  default  default  Stopped  -  /kukeon/default/default/default/sbsh-pm-0
```

The operator-facing impact is two-fold:

1. **`kuke status` divergence after #853 lands** — once `Status.CgroupReady` is honestly tracked, every cell whose cgroup the reboot wiped shows as a divergence row. With this issue unfixed, the operator must run `kuke start cell <name>` per cell to clear them.
2. **Silent reliance on the StartCell-implicit recreation** — any in-process workflow (`kuke run --reuse` against a clone whose cgroup is gone, `kuke apply -c <cfg>` over a divergent cell, an in-flight clone-and-start) silently depends on runc-via-StartCell to re-materialise the cgroup directory rather than seeing a fresh, properly-controllered cgroup state.

For cells with subtree-controller customisations (#314, #328), the reconciler's lack of re-ensure also means `Status.SubtreeControllers` is not re-asserted after a reboot — drift from the spec is invisible until the next start.

`bootstrapCell` (`internal/controller/bootstrap.go:621`) does re-ensure cgroups for cells that were bootstrap-managed, but that path is `kuke init`-only — it does not run on a normal daemon restart, and it does not cover user-created cells.

## Proposal

Have the background reconciler re-create the cell cgroup when its check reports absent. Open design points the implementer can choose between:

- **Always re-ensure** when `cgroupExists == false` and cell metadata is present. Idempotent, simple, slight cost per tick (one syscall per missing cgroup). Symmetric with `RefreshCell`'s view that "cgroup state is derived, not authoritative."
- **Re-ensure only when `Status.ReadyObserved == true`** — i.e. only for cells that were Ready before the reboot. Skips cells that never reached Ready (a half-CreateCell that crashed mid-way). The #850–#853 set already use `readyObserved` as the "cell has ever been Ready" latch.
- **Don't auto-re-ensure; surface the divergence** in `kuke status` and refuse `kuke run` / `kuke start cell` until the operator explicitly opts in via a `kuke refresh --reprovision` (or similar) verb. Punts to the operator but leaves no implicit re-ensure cycle.

Whichever option is taken should also extend to `Status.SubtreeControllers` re-assertion (the `enableCellControllers` call already in `ensureCellCgroup` at `provision.go:1326`) — that is the second half of the post-#314/#328 contract.

## Acceptance criteria

- [ ] After a host reboot, the background reconcile loop either (a) re-creates the cell cgroup directory and re-asserts subtree controllers, or (b) surfaces the divergence to `kuke status` with an actionable recovery command. The chosen behaviour is documented in `CLAUDE.md`'s smoke-test contract.
- [ ] `kuke status` after a reboot shows zero `FAIL` rows on the cell-cgroup row for cells whose metadata exists, regardless of whether the reconciler healed them or the operator did.
- [ ] Tests cover: cell with metadata + no cgroup → run one reconcile tick → assert chosen behaviour. Tests cover subtree controllers being re-asserted on healed cgroups.
- [ ] No regression on cells that never reached Ready (the half-CreateCell case) — they don't get accidentally promoted by the re-ensure.

## Notes

- Code anchors: `internal/controller/runner/refresh.go:632 ReconcileCell` (cgroup check at line 663 — boolean result is currently discarded); `internal/controller/runner/provision.go:1264 ensureCellCgroup` (the existing re-ensure path); `internal/controller/runner/provision.go:1326 enableCellControllers` (subtree controllers).
- Closely related closed work: #314 + #328 (subtree controller delegation), #543 (post-reboot state derivation).
- Companion of **#853** — that one fixes `cgroupReady` to honestly *report* divergence; this one decides whether to *heal* it.

eminwux:
## Summary

- Background reconciler now re-runs `ensureCellCgroup` when `ExistsCgroup` reports absent **and** the persisted `Status.ReadyObserved` latch is true, re-creating the wiped cgroup directory and re-asserting subtree controllers on the same pass.
- Closes the half-CreateCell gap explicitly: a crashed CreateCell that never reached Ready (latch still false on disk) is **not** promoted by the re-ensure — the in-flight CreateCell will finish its own `ensureCellCgroup` path under the per-cell lock, and the reconciler stepping in would race that flow.
- No behaviour change for cells whose cgroup is already present, or for cells with no metadata on disk.

This is option (b) from the issue's Proposal section — "re-ensure only when `Status.ReadyObserved == true`" — chosen because the AC's #4 ("no regression on cells that never reached Ready — the half-CreateCell case") directly names the latch as the discriminator.

## Implementation notes

- The reconcile loop's existing `ExistsCgroup` call discarded its boolean result. Captured it and gated the heal on `(cgroupErr == nil && !cgroupExists && originalStatus.ReadyObserved)`. On heal failure the warning is logged and the tick continues (best-effort, next pass retries); heal success swaps the in-memory cell for the one `ensureCellCgroup` returns so the downstream derivation and persist read `Status.CgroupPath` / `Status.SubtreeControllers` set by the heal.
- `ensureCellCgroup` already re-asserts subtree controllers via `enableCellControllers` on every ensure-pass (the post-#314/#328 contract); the heal path inherits that for free, closing the AC's third bullet.
- **Post-rebase fix-up (b99ccb1):** rebased onto current `main` to pick up #861 (now `0bf04aa`), which stamps `Status.CgroupReady = cgroupErr == nil && cgroupExists` immediately above this heal block. `ensureCellCgroup` writes `CgroupPath` / `SubtreeControllers` but not `CgroupReady`, so without an explicit write the heal would have shipped `cgroupReady:false` in the same tick the cgroup was re-created (next pass would re-probe and correct, but the false-then-true flip on disk is the kind of stale-state surface #861 just closed). The heal-success branch now sets `cell.Status.CgroupReady = true` and `TestReconcileCell_HealsWipedCgroupWhenReadyObserved` pins the invariant.
- The `deleteCellFakeClient` shared fake grew three overridable hooks (`newCgroupFn`, `ensureSubtreeControllersFn`, `enableCellSubtreeFn`) so the heal tests can observe `NewCgroup` + `EnableCellSubtreeControllers` calls without rewriting the fake.

## Note on AC #1's CLAUDE.md doc

The AC asks for the chosen behaviour to be documented in `CLAUDE.md`'s smoke-test contract. The dev role rule forbids `<project>/CLAUDE.md` edits in a code-fix PR (it travels as its own project issue + PR), so the doc work is deferred to #862 with the verbatim proposed wording in the issue body. Note that `make dev-init` does not exercise reboot recovery (per umbrella #854's Notes); the new `internal/controller/runner/reconcile_heal_test.go` regression guards cover both the heal-positive and half-CreateCell-negative paths.

## Test plan

- [x] `make test` (full project CI test target) — passes after rebase + fix-up, no regressions.
- [x] `make dev-init` (project smoke) — daemon-parity check shows both realms Ready with matching `/kukeon/<realm>` cgroup rows; attach smoke against `kuketty`-wrapped cell succeeds.
- [x] `go test ./internal/controller/runner/... -run 'TestReconcileCell_HealsWipedCgroupWhenReadyObserved|TestReconcileCell_DoesNotHealHalfCreateCell|TestReconcileCell_PostReboot'` — heal-positive (now pinning `CgroupReady=true` post-heal), half-CreateCell-negative, and post-reboot derivation guards pass.
- [ ] Live host-reboot smoke (reboot host, observe cgroups re-created on next reconcile tick) — not exercised by `make dev-init` per umbrella #854's Notes; covered by the unit tests above.

Closes #855

### DIFF
--- a/internal/controller/runner/delete_cell_test.go
+++ b/internal/controller/runner/delete_cell_test.go
@@ -45,14 +45,19 @@ import (
 // adding a new code path through ctr.Client doesn't require revisiting this
 // fake — the controller-level error assertions catch unintended behavior.
 // Also reused by container_state / reconcile tests that need ExistsContainer
-// + TaskStatus hooks (the post-reboot reproduction in #543).
+// + TaskStatus hooks (the post-reboot reproduction in #543), and by the
+// reconcile-heal tests (#855) that need NewCgroup +
+// EnsureSubtreeControllers hooks to observe the re-ensure pass.
 type deleteCellFakeClient struct {
-	deleteContainerFn func(namespace, id string, opts ctr.ContainerDeleteOptions) error
-	stopContainerFn   func(namespace, id string, opts ctr.StopContainerOptions) (*containerd.ExitStatus, error)
-	listContainersFn  func(namespace string, filters ...string) ([]containerd.Container, error)
-	existsContainerFn func(namespace, id string) (bool, error)
-	taskStatusFn      func(namespace, id string) (containerd.Status, error)
-	loadCgroupFn      func(group, mountpoint string) (*cgroup2.Manager, error)
+	deleteContainerFn          func(namespace, id string, opts ctr.ContainerDeleteOptions) error
+	stopContainerFn            func(namespace, id string, opts ctr.StopContainerOptions) (*containerd.ExitStatus, error)
+	listContainersFn           func(namespace string, filters ...string) ([]containerd.Container, error)
+	existsContainerFn          func(namespace, id string) (bool, error)
+	taskStatusFn               func(namespace, id string) (containerd.Status, error)
+	loadCgroupFn               func(group, mountpoint string) (*cgroup2.Manager, error)
+	newCgroupFn                func(spec ctr.CgroupSpec) (*cgroup2.Manager, error)
+	ensureSubtreeControllersFn func(group, mountpoint string, controllers []string) ([]string, error)
+	enableCellSubtreeFn        func(group, mountpoint string, controllers []string) ([]string, error)
 }
 
 func (c *deleteCellFakeClient) Connect() error { return nil }
@@ -72,7 +77,10 @@ func (c *deleteCellFakeClient) CleanupNamespaceResources(string, string) error {
 func (c *deleteCellFakeClient) GetCgroupMountpoint() string               { return "" }
 func (c *deleteCellFakeClient) GetCurrentCgroupPath() (string, error)     { return "", nil }
 func (c *deleteCellFakeClient) CgroupPath(string, string) (string, error) { return "", nil }
-func (c *deleteCellFakeClient) NewCgroup(ctr.CgroupSpec) (*cgroup2.Manager, error) {
+func (c *deleteCellFakeClient) NewCgroup(spec ctr.CgroupSpec) (*cgroup2.Manager, error) {
+	if c.newCgroupFn != nil {
+		return c.newCgroupFn(spec)
+	}
 	//nolint:nilnil // cgroup2.Manager has unexported fields; the test path discards the value
 	return nil, nil
 }
@@ -85,11 +93,21 @@ func (c *deleteCellFakeClient) LoadCgroup(group, mountpoint string) (*cgroup2.Ma
 	return nil, nil
 }
 func (c *deleteCellFakeClient) DeleteCgroup(string, string) error { return nil }
-func (c *deleteCellFakeClient) EnsureSubtreeControllers(string, string, []string) ([]string, error) {
+func (c *deleteCellFakeClient) EnsureSubtreeControllers(
+	group, mountpoint string, controllers []string,
+) ([]string, error) {
+	if c.ensureSubtreeControllersFn != nil {
+		return c.ensureSubtreeControllersFn(group, mountpoint, controllers)
+	}
 	return nil, nil
 }
 
-func (c *deleteCellFakeClient) EnableCellSubtreeControllers(string, string, []string) ([]string, error) {
+func (c *deleteCellFakeClient) EnableCellSubtreeControllers(
+	group, mountpoint string, controllers []string,
+) ([]string, error) {
+	if c.enableCellSubtreeFn != nil {
+		return c.enableCellSubtreeFn(group, mountpoint, controllers)
+	}
 	return nil, nil
 }
 
--- a/internal/controller/runner/reconcile_heal_test.go
+++ b/internal/controller/runner/reconcile_heal_test.go
@@ -126,6 +126,14 @@ func TestReconcileCell_HealsWipedCgroupWhenReadyObserved(t *testing.T) {
 	if !updatedCell.Status.ReadyObserved {
 		t.Errorf("Status.ReadyObserved = false, want true (the latch must survive the heal)")
 	}
+	// Pair with the #861 probe: the heal must ship cgroupReady:true in
+	// the same tick. Pre-fix the probe stamped false (cgroup absent)
+	// and ensureCellCgroup never touched the field, so the persisted
+	// snapshot carried cgroupReady:false for a cell whose cgroup is in
+	// fact present — exactly the stale-state surface #861 just closed.
+	if !updatedCell.Status.CgroupReady {
+		t.Errorf("Status.CgroupReady = false, want true after heal (heal must pair with the #861 probe so cgroupReady:true ships in the same tick, not next-pass)")
+	}
 }
 
 // TestReconcileCell_DoesNotHealHalfCreateCell is the regression guard for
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #863 Review — `fix(reconcile): heal wiped cell cgroup on the ReadyObserved latch`**

Heal logic is correctly gated on the persisted `ReadyObserved` latch, exercises both halves of the post-#314/#328 contract via `ensureCellCgroup`, and the test pair covers heal-positive + half-CreateCell-negative cleanly. One blocking interaction with the just-merged #861 surfaces on rebase against current main.

**One blocking item:**

- **Stale `CgroupReady` after heal** `internal/controller/runner/refresh.go:683-696`: PR #861 (now on main, post-c3eb7d0) added `cell.Status.CgroupReady = cgroupErr == nil && cgroupExists` immediately above this heal block. After auto-merge the sequence becomes: (1) `cell.Status.CgroupReady = false` (cgroup absent on probe), (2) `cell = healedCell` — but `ensureCellCgroup` only writes `Status.CgroupPath` / `Status.SubtreeControllers`, not `CgroupReady`, so the false value survives. The `updated` predicate then persists `cgroupReady: false` for a cell whose cgroup is in fact present, which is exactly the kind of stale-state surface #861 just closed (the next reconcile tick re-probes and corrects, so the window is one tick — but it still ships a false-then-true flip through the metadata file). Fix on rebase: after `cell = healedCell`, also set `cell.Status.CgroupReady = true`; add the assertion to `TestReconcileCell_HealsWipedCgroupWhenReadyObserved` so the invariant is pinned. (Verified by running `git merge --no-commit pr-863` onto `origin/main` and inspecting refresh.go:683-696 in the merged tree.)

Non-blocking nit (will file as followup once this lands):
- **Dead fake hook** `internal/controller/runner/delete_cell_test.go:59`: `ensureSubtreeControllersFn` is declared and wired but no test sets it — the heal path goes through `EnableCellSubtreeControllers`, not `EnsureSubtreeControllers`. Drop the field + the wiring at delete_cell_test.go:96-104 or add a test that exercises it.

LGTM once the rebase + `CgroupReady = true` patch lands; the dead fake hook is non-blocking.

eminwux:
Addressed in b99ccb1 (after rebase onto current main, picking up #861 / `0bf04aa`):

- **Stale `CgroupReady` after heal** — heal-success branch now sets `cell.Status.CgroupReady = true` immediately after `cell = healedCell`, pairing the heal with the same truth #861's probe stamps so a healed cell ships `cgroupReady:true` in the same tick (no more false-then-true flip through metadata). `TestReconcileCell_HealsWipedCgroupWhenReadyObserved` got the matching assertion so the invariant is pinned.

Deferred: dead-fake-hook nit (`ensureSubtreeControllersFn` at `delete_cell_test.go:59`) — left as a separate follow-up per your review note; not bundling into this fix-up.

eminwux:
**PR #863 Review — `fix(reconcile): heal wiped cell cgroup on the ReadyObserved latch`** (post-rebase + commit b99ccb1)

LGTM — rebased on main cleanly and the heal block now sets `cell.Status.CgroupReady = true` after `cell = healedCell` at refresh.go:704-708, with the matching `Status.CgroupReady` assertion added to `TestReconcileCell_HealsWipedCgroupWhenReadyObserved` at reconcile_heal_test.go:115-122. The same-tick `cgroupReady:true` ship pairs cleanly with #861's probe contract; both blockers from the prior review are addressed.

Issue #868, opened 2026-05-25
kuke daemon reset + kuke init does not refresh kukeond cell when image tag is reused


eminwux:
## Context

Filed as the Layer 2 follow-up of #857 (the dev-init.sh image-digest guard PR is the Layer 1 fail-loud diagnostic; this issue is the underlying code defect that diagnostic surfaces).

When the operator iterates on the daemon source tree and re-runs `make dev-init`, the sequence is:

1. `kuke daemon reset` — supposed to stop + delete the kukeond cell + clear `/run/kukeon/kukeond.{sock,pid}`.
2. `kuke build -t kukeon-local:dev --realm kuke-system .` — overwrites the `:dev` tag in the kuke-system realm's containerd namespace with a fresh manifest.
3. `kuke init --kukeond-image docker.io/library/kukeon-local:dev` — supposed to (re)create the kukeond cell against the just-built image and bring the daemon up.

On the failing path observed in #857, the running daemon process PID does not change across step 1, and the `kuke init` step prints `cell "kukeond": already existed` / `cell containers: started`, meaning the existing kukeond cell record was reused and the new image was never propagated to the running container. The running daemon binary stays on the pre-rebuild commit; downstream behavior (e.g. the kuketty doc renderer change in #641) silently regresses.

Both candidate root-cause sites in #857's diagnostic table are plausible and not yet pinpointed:

- **`kuke daemon reset` is not destructive enough.** The user observed `kuke daemon reset` ran without error yet the daemon PID did not change. `cmd/kuke/daemon/reset/reset.go:140` calls `client.DeleteCell(cmd.Context(), doc)` and returns `errdefs.ErrControllerNoChange` if `MetadataDeleted` is false — so either the on-disk metadata *was* deleted but the running container task survived (`StopCell` did not actually terminate the kukeond process inside its cell), or the cell record itself was not removed despite the success indication.
- **`kuke init`'s `bootstrapCell` (`internal/controller/bootstrap.go:621`) does not compare images on the existing-cell path.** When `section.CellMetadataExistsPre` is true, the path is `EnsureCell` + `StartCell` — neither call looks at the desired `image` (passed via `kukeondCellDoc(image, …)`) and reconciles it against the persisted cell's container spec. If a prior reset left the kukeond cell record around with a stale image ref *and* a stale running container, init silently re-uses both.

[...]

PR #869, opened 2026-05-25
fix(dev-init): fail loud when running kukeond container is anchored on a stale image

### LINKED ISSUE #857: dev-init: attach-smoke fails with "missing apiVersion v1beta1" when running daemon predates #641 (kuke daemon reset does not replace stale daemon binary)
## Repro

On a host where the running \`kukeond\` was built before #641/#645 (the change that switched the per-container metadata renderer from sbsh TerminalDoc to kukeon ContainerDoc — commit 3d59273, 2026-05-21 17:55), bring the worktree to current \`main\` (verified at commit 3130818 today, 2026-05-25) and run:

\`\`\`bash
make dev-init
\`\`\`

The first phases pass (build, daemon reset, image stage, daemon-parity check on \`kuke get realms\`). The script then enters the \`==> kuke attach smoke against a kuketty-wrapped cell\` section, creates a \`cattach\` cell with an attachable busybox container, waits 20 s for the kuketty socket to bind (which succeeds — the script does not trip the "kuketty socket not bound" timeout), and then fails the next check:

\`\`\`
rendered metadata at /opt/kukeon/data/dev-init-attach/ds/dks/cattach/work/kuketty-metadata.json missing apiVersion v1beta1
\`\`\`

## Observed

Inspecting \`kuketty-metadata.json\` on this host before the script's auto-purge fires (re-created the smoke cell manually post-failure to capture the on-disk content):

\`\`\`json
{
  "apiVersion": "sbsh/v1beta1",
  "kind": "Terminal",
  "metadata": {
    "name": "c",
    "labels": {
      "kukeon.io/cell": "cattach",
      ...
    }
  },
  "spec": {
    "id": "023063f1",
    "kind": 0,
    "name": "gallant_amonhen",
    ...
    "runPath": "/run/kukeon/tty",
    "captureFile": "/run/kukeon/tty/capture",
    "logFile": "",
    "socketIO": "/run/kukeon/tty/socket",
    ...
  }
}
\`\`\`

The on-disk file is a pre-#641 sbsh \`TerminalDoc\` (\`apiVersion: sbsh/v1beta1\`, \`kind: Terminal\`), not the post-#641 kukeon \`ContainerDoc\` (\`apiVersion: v1beta1\`, \`kind: Container\`) the daemon's current \`writeKukettyDoc\` (\`internal/controller/runner/attachable.go:214 writeKukettyDoc\`) renders. The script's grep at \`scripts/dev-init.sh:311\` is correct against current-code expectations — the file content is the divergence.

The running daemon (\`ps -o pid,lstart\` on the \`/bin/kukeond serve\` process) was started 2026-05-21 06:51:15, ~11 hours **before** #641 merged. The kukeond cell on disk (\`/opt/kukeon/data/kuke-system/kukeon/kukeon/kukeond/\`) was created at the same timestamp; \`make dev-init\`'s "Reset prior kukeond cell" phase (\`scripts/dev-init.sh:192 sudo ./kuke daemon reset\`) ran today, but the daemon process PID did not change across the run and the running daemon's binary is still the pre-#641 build. The \`kuke build\` step did produce a fresh \`docker.io/library/kukeon-local:dev\` image into the kuke-system realm's containerd namespace, but \`kuke init --kukeond-image\` apparently kept the existing kukeond cell in place (consistent with the script's own log output:

\`\`\`
- cell containers: started
\`\`\`

— \`started\` covers "already running," not "replaced with new image").

Confirmed cross-branch: the failure reproduces identically on \`origin/main\` (3130818) under an isolated worktree on the same host, so it's not specific to any in-flight PR.

## Root cause

1. **\`kuke daemon reset\` is not destructive enough to force a binary swap on the next \`kuke init\`.** The CLAUDE.md description ("stops + deletes the prior \`kukeond\` cell and clears \`/run/kukeon/kukeond.{sock,pid}\`") suggests the cell should be deleted, but the script's subsequent \`kuke init\` log line "cell "kukeond": already existed" indicates the cell record survived. If reset is supposed to delete it, that's a reset bug. \`internal/controller/runner/\` (the reset path) and \`cmd/kuke/daemon/reset/reset.go\` are the candidates.
2. **\`kuke init --kukeond-image\` is idempotent against the running cell** and treats an already-Ready kukeond as up-to-date even when \`--kukeond-image\` resolves to a digest the running container does not have. The fix would be to compare the running container's image digest against \`--kukeond-image\` and replace on mismatch. \`internal/controller/runner/\` (the init/provision path) and \`cmd/kuke/init/init.go\` are the candidates.
3. **\`make dev-init\` flow gap** — the script trusts \`kuke daemon reset\` + \`kuke init\` to converge on the fresh image and never verifies the running daemon's version against the just-built image. \`scripts/dev-init.sh\` could add a post-init "running kukeond image digest must match \`\${KUKEOND_IMAGE_REF}\`" check that fails-loud before the smoke step, regardless of whether the underlying cause is in reset or init.

The user-facing failure under each candidate is identical, so all three would be fixed by adding the post-init image-digest check at the bare minimum, and the root cause then surfaces as either "reset didn't delete the cell" or "init didn't replace the image."

## Suggested fix

Two-layer fix:

1. \`scripts/dev-init.sh\` — after \`kuke init\`, query the running kukeond container's image digest (e.g. \`sudo ctr -n kuke-system.kukeon.io container info kukeon_kukeon_kukeond_kukeond | jq -r '.Image'\`) and compare against the just-staged \`\${KUKEOND_IMAGE_REF}\` digest from \`sudo ctr -n kuke-system.kukeon.io images ls\`. Fail-loud with "running kukeond image \`<old>\` does not match staged \`<new>\` — try \`kuke daemon reset --purge-system\` to force a clean reprovision" before the smoke phase runs. This catches the gap regardless of where the underlying bug is.
2. \`internal/controller/runner/\` or \`cmd/kuke/daemon/reset/reset.go\` — investigate whether the reset path is actually deleting the kukeond cell record (the \`--purge-system\` flag suggests it currently keeps user-realm data; the \`kukeond\` cell itself should be unconditionally deleted on reset). If reset is correct, then \`kuke init\` needs a "image-digest mismatch → recreate" branch on the existing-cell path.

Memory hint: the [Init/bootstrap self-sufficiency gaps are bugs](../../feedback_self_sufficiency_gaps_are_bugs.md) feedback applies — the operator should not need to manually \`kuke uninstall\` + reinstall after upgrading the source tree across a daemon-affecting change. \`make dev-init\` is documented as "the canonical re-bootstrap loop" (project CLAUDE.md) and the user contract is "re-running on a healthy host produces a clean re-bootstrap."

eminwux:
## Summary
- Adds a post-`kuke init` snapshot-chain-digest check between init and the attach smoke so a stale-running-daemon scenario (#857) fails loud at the place the drift was introduced rather than as an opaque schema mismatch in the kuketty rendered metadata.
- The check compares the running kukeond container's snapshot `Parent` (set at container-create time) against the just-built image's chain digest, extracted via `manifest → config blob → containerd.io/gc.ref.snapshot.overlayfs label`. A byte-precise drift signal that catches "same tag, new digest" — which a name-only comparison cannot.
- Implements Layer 1 of #857's suggested two-layer fix. Layer 2 (the underlying `kuke daemon reset` + `kuke init` refresh defect) is filed as #868 and explicitly out of scope here.

## Implementation note (AC reinterpretation)

#857's "Suggested fix" Layer 1 asks for `ctr container info … | jq -r '.Image'` compared to the staged image's digest from `ctr images ls`. Read literally, that compares an image **name** (the `.Image` field is a ref string set at container-create time, e.g. `docker.io/library/kukeon-local:dev`) against a manifest **digest** — type-incompatible, and unchanged when `kuke build` overwrites the same tag with a new manifest. Posted a heads-up comment on #857 (https://github.com/eminwux/kukeon/issues/857#issuecomment-4534165283) before coding; this PR ships the spirit of Layer 1 (digest-vs-digest drift detection) via the snapshot chain comparison.

The implementation stays grep/sed-only per `scripts/install.sh`'s "no jq dependency" convention.

## Layer 2 deferral

#868 covers the underlying defect (reset+init don't refresh the running daemon when the image tag is reused). It needs a focused investigation cycle into both candidate sites (`cmd/kuke/daemon/reset/reset.go:140` for the reset path's containerd task-termination semantics; `internal/controller/bootstrap.go:621`'s `bootstrapCell` for the existing-cell image-drift branch) that the parent issue acknowledges is open ("Root cause: Not yet pinpointed"). Bundling that investigation into this PR would either oversize the diff or guess at the root cause; #868 keeps the lineage auditable.

This Layer 1 guard remains useful as defense-in-depth after #868 lands — a fresh class of refresh regression would surface here even on a `bootstrapCell` that's been audited.

## Test plan
- [x] `make test` (full project test target — `test-root` + `test-kukebuild`) — all green.
- [x] `GOWORK=off go vet ./...` — clean (per repo's go.work toolchain note).
- [x] `pre-commit run --files scripts/dev-init.sh` — trailing whitespace, EOF, mixed line endings all pass.
- [x] Manual end-to-end smoke of the chain-digest extraction on this host's actual containerd state: `CTR_CHAIN`, `MANIFEST_DIGEST`, `CONFIG_DIGEST`, `IMG_CHAIN` all resolved correctly; healthy path produced MATCH.
- [ ] `make dev-init` end-to-end against the #857 stale-daemon repro — requires a host where the running kukeond predates a daemon-affecting commit, which this host does not currently reproduce. The chain-digest extraction is verified above; the fail-loud branch is exercised by the equality check on the same data.

Closes #857

### DIFF
--- a/scripts/dev-init.sh
+++ b/scripts/dev-init.sh
@@ -207,6 +207,80 @@ sudo --preserve-env=KUKEON_HOST ./kuke build --build-arg VERSION=v0.0.0-dev \
 step "Run kuke init with --kukeond-image ${KUKEOND_IMAGE_REF}"
 sudo --preserve-env=KUKEON_HOST,KUKEOND_SOCKET ./kuke init --kukeond-image "${KUKEOND_IMAGE_REF}"
 
+# Fail-loud when the running kukeond container is anchored on a stale
+# image (issue #857). The bug captured there is that `kuke daemon reset`
+# + `kuke init --kukeond-image` does not refresh the kukeond cell when
+# the image tag is reused (init's bootstrapCell takes the existing-cell
+# path on a leftover record and never compares the desired image
+# against the running container). The user-visible failure surfaces
+# downstream as an opaque pre-#641 schema mismatch in the attach smoke;
+# this check turns that into a fail-loud at the place the divergence
+# was introduced.
+#
+# Compare the *snapshot chain digest*, not the image ref name: a
+# container's `.Image` field is the ref string set at create time
+# (e.g. `docker.io/library/kukeon-local:dev`) and is unchanged when
+# `kuke build` overwrites the same tag with a new manifest. The
+# canonical drift signal is the chain digest stored on the container's
+# snapshot at unpack time, derived from the image config blob's
+# `containerd.io/gc.ref.snapshot.overlayfs` label. The two should be
+# byte-identical on a healthy `make dev-init`; a mismatch means the
+# running container is anchored on an older image than the one we just
+# built.
+#
+# Stay grep/sed-only — scripts in this repo intentionally avoid a jq
+# dependency (see scripts/install.sh's "no jq dependency" guarantee).
+step "Verify running kukeond container is anchored on the just-built image"
+KUKEOND_NS="kuke-system.kukeon.io"
+KUKEOND_CTR_ID="kukeon_kukeon_kukeond_kukeond"
+
+ctr_chain="$(sudo ctr -n "${KUKEOND_NS}" snapshots info "${KUKEOND_CTR_ID}" 2>/dev/null \
+    | grep -oE '"Parent":[[:space:]]*"sha256:[a-f0-9]+"' \
+    | sed -E 's/.*"(sha256:[a-f0-9]+)"/\1/')"
+if [ -z "${ctr_chain}" ]; then
+    printf 'unable to resolve snapshot chain for running kukeond container %s/%s\n' \
+        "${KUKEOND_NS}" "${KUKEOND_CTR_ID}" >&2
+    exit 1
+fi
+
+manifest_digest="$(sudo ctr -n "${KUKEOND_NS}" images ls 2>/dev/null \
+    | awk -v r="${KUKEOND_IMAGE_REF}" 'NR>1 && $1==r {print $3; exit}')"
+if [ -z "${manifest_digest}" ]; then
+    printf 'image %s not found in containerd namespace %s after kuke build\n' \
+        "${KUKEOND_IMAGE_REF}" "${KUKEOND_NS}" >&2
+    exit 1
+fi
+
+config_digest="$(sudo ctr -n "${KUKEOND_NS}" content get "${manifest_digest}" 2>/dev/null \
+    | grep -oE '"digest":[[:space:]]*"sha256:[a-f0-9]+"' \
+    | head -1 \
+    | sed -E 's/.*"(sha256:[a-f0-9]+)"/\1/')"
+if [ -z "${config_digest}" ]; then
+    printf 'unable to resolve config digest from manifest %s\n' "${manifest_digest}" >&2
+    exit 1
+fi
+
+img_chain="$(sudo ctr -n "${KUKEOND_NS}" content ls 2>/dev/null \
+    | awk -v d="${config_digest}" '$1==d' \
+    | grep -oE 'containerd\.io/gc\.ref\.snapshot\.overlayfs=sha256:[a-f0-9]+' \
+    | head -1 \
+    | cut -d= -f2)"
+if [ -z "${img_chain}" ]; then
+    printf 'unable to resolve image chain digest from config blob %s\n' "${config_digest}" >&2
+    exit 1
+fi
+
+if [ "${ctr_chain}" != "${img_chain}" ]; then
+    printf 'running kukeond container is anchored on a stale image (#857)\n' >&2
+    printf '  running container snapshot chain: %s\n' "${ctr_chain}" >&2
+    printf '  just-built image chain:           %s\n' "${img_chain}" >&2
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #869 Review — `fix(dev-init): fail loud when running kukeond container is anchored on a stale image`**

LGTM — read-only, well-commented Layer-1 guard around #857, AC reinterpretation properly socialized via the upstream comment and #868 deferral. Snapshot-key assumption (`kukeon_kukeon_kukeond_kukeond` = container ID, no `SnapshotKey` override on the bootstrap path) matches `internal/ctr/container.go:351`; chain-digest extraction is algorithmically sound for the canonical overlayfs flow. The #857 stale-daemon repro that exercises the fail-loud branch is host-gated and documented in the test plan — please run it on a host that reproduces #857 pre-merge.

Issue #824, opened 2026-05-23
cli: kuke get image[s] — promote image listing to first-class get kind (replaces kuke image get)


eminwux:
## Context

`kuke image get` (added by #211) lists images in a realm's containerd namespace, but it sits under the `kuke image` verb group alongside `load` / `delete` rather than under the `kuke get <kind>` family. With `kukebuild` landed — images are now built straight into a realm's namespace by `kuke build` with no docker / `ctr` hop — images are a first-class realm-scoped artifact like spaces, stacks, cells, and containers. The asymmetry shows up in two places:

- `kuke get` enumerates `realm`, `space`, `stack`, `cell`, `container`, `secret`, `blueprint`, `config` — `image` is the one realm-scoped kind missing.
- Cross-realm listing (the default for `kuke get cells` / `stacks` / etc.) requires running `kuke image get` twice today, once per realm.

This issue promotes image listing to `kuke get image[s]` with cross-realm-by-default semantics matching the rest of the `get` family, and retires `kuke image get` in the same PR. The non-list image verbs (`kuke image load`, `kuke image delete`) stay where they are — they have no parallel under `get`.

## Proposal

In `cmd/kuke/get/`:

- Add `cmd/kuke/get/image/image.go` exposing `NewImageCmd()`, registered in `cmd/kuke/get/get.go` alongside the other leaves.
  - `Use: \"image [ref]\"`, aliases `images`, `img`.
  - Default table (no positional, no `--realm`) lists **across every realm**, columns `NAME REALM SIZE AGE`.
  - Positional `<ref>` describes a single image (yaml by default, json with `-o json`).
  - `--realm <r>` narrows listing to one realm; `REALM` column is then redundant but kept for grep-ability.
  - `-o wide` keeps `CREATED` and adds `DIGEST` after `AGE` (parallel to the wide-column treatment in #601–#605).
  - `-o yaml` / `-o json` as today.

In `cmd/kuke/image/`:

- Delete `cmd/kuke/image/get.go` + `get_test.go`. The `kuke image` parent now exposes only `load` and `delete`.
- Update `cmd/kuke/image/image.go` `Short` / help text accordingly.

In `docs/cli-use-cases.md`:

- Replace the three `kuke image get …` lines (around lines 222–223, 229, 232) with `kuke get image …` equivalents, and add `kuke get images` (cross-realm) to the listing block around lines 151–158.
- Note the cross-realm default in the behavioral invariants list.

## Acceptance criteria

[...]

PR #871, opened 2026-05-25
refactor(get/shared): pipe PrintYAML/PrintJSON through cmd.OutOrStdout

### LINKED ISSUE #829: chore(get/shared): pipe PrintYAML/PrintJSON through cmd.OutOrStdout for testability
## Context

`cmd/kuke/get/shared/shared.go:88-99` defines `PrintYAML` and `PrintJSON` as:

```go
func PrintYAML(doc interface{}) error {
    encoder := yaml.NewEncoder(os.Stdout)
    ...
}
func PrintJSON(doc interface{}) error {
    encoder := json.NewEncoder(os.Stdout)
    ...
}
```

Both write to `os.Stdout` directly. That breaks the cobra `cmd.SetOut(buf)` test-capture pattern — yaml/json output bypasses the buffer the test wired up, so a leaf's structured-output tests cannot assert on actual emitted bytes.

`cmd/kuke/get/container/container.go:39-54` works around this by exposing package-level printer-function vars (`YAMLPrinter`, `JSONPrinter`, `TablePrinter`) so tests can reassign them. The pattern works but triggers two strict-lint findings on every new file that adopts it:

- `gochecknoglobals` on the three vars (3 issues per leaf).
- `reassign` on every override site in `_test.go` (2+ issues per test entry-point).

#824 hit this during the new `kuke get image` leaf: I sidestepped both by inlining `encodeYAML(cmd, doc)` / `encodeJSON(cmd, doc)` that write to `cmd.OutOrStdout()`, but that means each new `get` leaf either re-derives the same helpers or carries the lint debt the container leaf already accepts. The sibling-series issues #601–#605 will hit the same fork in the road.

## Proposal

Refactor the shared printers to take an `io.Writer` (or a `*cobra.Command` and call `cmd.OutOrStdout()` internally), then update every existing caller. The migration is mechanical — every call site is in `cmd/kuke/get/<kind>/<kind>.go`.

Option A — accept `io.Writer`:

```go
func PrintYAML(w io.Writer, doc any) error { ... }
func PrintJSON(w io.Writer, doc any) error { ... }
```

Callers pass `cmd.OutOrStdout()`. `PrintTable` already takes a `*cobra.Command` and uses `cmd.Println`, so it stays as is.

Option B — accept `*cobra.Command`:

```go
func PrintYAML(cmd *cobra.Command, doc any) error { ... }
func PrintJSON(cmd *cobra.Command, doc any) error { ... }
```

Symmetric with `PrintTable`. Slightly more cobra-coupled but matches the existing PrintTable shape and is what every caller actually has on hand.

Either approach unlocks straight `cmd.SetOut(buf)` capture in tests, lets `cmd/kuke/get/container/` retire its `YAMLPrinter`/`JSONPrinter`/`TablePrinter` globals + override pattern (3 + 4 = 7 lint findings the package currently accepts), and removes the inline helpers from `cmd/kuke/get/image/image.go` introduced in #824.

## Acceptance criteria

- [ ] `cmd/kuke/get/shared/shared.go`: `PrintYAML` / `PrintJSON` take a writer (or `*cobra.Command`) and stop writing to `os.Stdout`.
- [ ] All existing `kuke get <kind>` callers updated to pass through `cmd.OutOrStdout()` (or the `*cobra.Command`).
- [ ] `cmd/kuke/get/container/`: drop `YAMLPrinter` / `JSONPrinter` / `TablePrinter` package vars and the `runContainerCmdWithDeps` injection seam — call the shared printers directly. Tests use `cmd.SetOut(buf)` capture.
- [ ] `cmd/kuke/get/image/image.go`: drop the inline `encodeYAML` / `encodeJSON` helpers — call the shared printers directly. Tests already use `cmd.SetOut(buf)` and need no changes beyond replacing the helper calls.
- [ ] `golangci-lint` clean on the touched files — no new `gochecknoglobals` / `reassign` findings.
- [ ] `make test` green.

## Out of scope

- The `PrintTable` signature is already `(*cobra.Command, []string, [][]string)`; no change there.
- Issues #601–#605 (the AGE / -o wide infra) are tracked separately; this refactor is independent and can land first or after.

## References

- Surfaced during #824 implementation (#828) — see PR body for the workaround that motivated this filing.
- `cmd/kuke/get/shared/shared.go:88-99` — current `PrintYAML` / `PrintJSON`.
- `cmd/kuke/get/container/container.go:39-54` — existing package-global override pattern that triggers the lint debt.
- `cmd/kuke/get/image/image.go` — inline `encodeYAML` / `encodeJSON` workaround introduced by #824.

eminwux:
## Summary

- Refactor `cmd/kuke/get/shared/shared.go`'s `PrintYAML` / `PrintJSON` to take `*cobra.Command` and write to `cmd.OutOrStdout()` instead of `os.Stdout`. Symmetric with the existing `PrintTable(cmd, …)` shape — went with Option B (cobra-coupled) per the issue body's two-option offer, since every caller already has `cmd` on hand and the symmetry keeps the call sites tight.
- Thread `cmd` through each leaf's single-doc `print<Kind>(…)` helper (`printRealm`, `printStack`, `printSpace`, `printCell`, `printSecret`, `printBlueprint`, `printConfig`, `printContainer`). The list-form `print<Kind>s(…)` helpers already took `cmd`.
- `cmd/kuke/get/container/`: drop the `printObjectFunc` / `tablePrinterFunc` types, the `YAMLPrinter` / `JSONPrinter` / `TablePrinter` package vars, and the `runContainerCmdWithDeps` injection seam. `printContainer` / `printContainersWithState` now call `shared.PrintYAML` / `PrintJSON` / `PrintTable` directly. Container tests already use `cmd.SetOut(buf)` capture and need no further changes.
- `cmd/kuke/get/image/image.go`: drop the inline `encodeYAML` / `encodeJSON` helpers (the #824 workaround) and call the shared printers directly. The `encoding/json` and `gopkg.in/yaml.v3` imports become unused and are dropped.
- `cmd/kuke/get/shared/shared_test.go`: replace the `os.Pipe` capture in `TestPrintYAML` / `TestPrintJSON` with the same `newOutputCommand()` + `cmd.SetOut(buf)` helper `TestPrintTable` already uses.
- `cmd/kuke/get/cell/cell_test.go`: drop the `os.Pipe` workaround in `TestNewCellCmd_SyncColumn`'s yaml subtests — yaml output now lands in `buf` alongside the table output, so the `wantStatus` substrings assert against the same buffer. The stale comment claiming `PrintYAML` writes to `os.Stdout` goes with it.

## Test plan

- [x] `make test` — full project test target green (root + `cmd/kukebuild`).
- [x] `GOWORK=off go build ./...` — clean compile.
- [x] `GOWORK=off go vet ./...` — no findings.
- [x] `GOWORK=off go test ./cmd/kuke/get/...` — every leaf (`shared`, `realm`, `space`, `stack`, `cell`, `container`, `image`, `secret`, `blueprint`, `config`) passes; no regressions from the threaded-`cmd` plumbing or the dropped override seam.
- [x] `pre-commit run --files <12 changed>` — trailing-whitespace, EOF, `go fmt`, `go-mod-tidy`, golangci-lint (`--new-from-rev HEAD --fix`), addlicense all green. Pre-commit's lint config is exactly the gating check the AC asks for — no new `gochecknoglobals` / `reassign` findings.
- [ ] `make dev-init` smoke — not run. Diff is pure CLI-side presentation logic; doesn't touch the daemon, the build path, or anything under `/opt/kukeon`, so the regression guard isn't materially exercised.

Closes #829

### DIFF
--- a/cmd/kuke/get/blueprint/blueprint.go
+++ b/cmd/kuke/get/blueprint/blueprint.go
@@ -103,7 +103,7 @@ func NewBlueprintCmd() *cobra.Command {
 				if !result.MetadataExists {
 					return fmt.Errorf("blueprint %q not found", name)
 				}
-				return printBlueprint(&result.Blueprint, outputFormat)
+				return printBlueprint(cmd, &result.Blueprint, outputFormat)
 			}
 
 			blueprints, err := client.ListBlueprints(cmd.Context(), realm, space, stack)
@@ -142,23 +142,23 @@ func resolveClient(cmd *cobra.Command) (kukeonv1.Client, error) {
 	return kukeshared.ClientFromCmd(cmd)
 }
 
-func printBlueprint(blueprint *v1beta1.CellBlueprintDoc, format shared.OutputFormat) error {
+func printBlueprint(cmd *cobra.Command, blueprint *v1beta1.CellBlueprintDoc, format shared.OutputFormat) error {
 	switch format {
 	case shared.OutputFormatJSON:
-		return shared.PrintJSON(blueprint)
+		return shared.PrintJSON(cmd, blueprint)
 	case shared.OutputFormatYAML, shared.OutputFormatTable:
-		return shared.PrintYAML(blueprint)
+		return shared.PrintYAML(cmd, blueprint)
 	default:
-		return shared.PrintYAML(blueprint)
+		return shared.PrintYAML(cmd, blueprint)
 	}
 }
 
 func printBlueprints(cmd *cobra.Command, blueprints []v1beta1.CellBlueprintDoc, format shared.OutputFormat) error {
 	switch format {
 	case shared.OutputFormatYAML:
-		return shared.PrintYAML(blueprints)
+		return shared.PrintYAML(cmd, blueprints)
 	case shared.OutputFormatJSON:
-		return shared.PrintJSON(blueprints)
+		return shared.PrintJSON(cmd, blueprints)
 	case shared.OutputFormatTable:
 		if len(blueprints) == 0 {
 			cmd.Println("No blueprints found.")
@@ -178,7 +178,7 @@ func printBlueprints(cmd *cobra.Command, blueprints []v1beta1.CellBlueprintDoc, 
 		shared.PrintTable(cmd, headers, rows)
 		return nil
 	default:
-		return shared.PrintYAML(blueprints)
+		return shared.PrintYAML(cmd, blueprints)
 	}
 }
 
--- a/cmd/kuke/get/cell/cell.go
+++ b/cmd/kuke/get/cell/cell.go
@@ -129,7 +129,7 @@ outOfSync / outOfSyncReason / outOfSyncError status fields.`,
 				if !result.MetadataExists {
 					return fmt.Errorf("cell %q not found in stack %q/%q/%q", name, realm, space, stack)
 				}
-				return printCell(&result.Cell, outputFormat)
+				return printCell(cmd, &result.Cell, outputFormat)
 			}
 
 			cells, err := client.ListCells(cmd.Context(), realm, space, stack)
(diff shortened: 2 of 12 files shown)


eminwux:
**PR #871 Review — `refactor(get/shared): pipe PrintYAML/PrintJSON through cmd.OutOrStdout`**

LGTM — symmetric with the existing `PrintTable(cmd, …)` shape, drops the container override seam and the `image.go` `encodeYAML`/`encodeJSON` #824 workaround, simplifies `cell_test.go` / `shared_test.go` capture, and Explore confirms zero remaining references to the removed symbols (`YAMLPrinter`/`JSONPrinter`/`TablePrinter`/`runContainerCmdWithDeps`/`printObjectFunc`/`tablePrinterFunc`/`encodeYAML`/`encodeJSON`) with all 50 call sites on the new two-arg signature. `make dev-init` deferral is justified — pure CLI-side presentation, doesn't touch the daemon or `/opt/kukeon`.

PR #848, opened 2026-05-24
feat(run): add --reuse — restart a healthy-Stopped clone, preserving overlay (step 3)

eminwux:
## Summary

- New `kuke run <config> --reuse` (#835, step 3 of the run-identity epic) picks the lowest-N healthy-Stopped clone of the source CellConfig (annotated by `--clone`) and starts that cell **in place** via the existing `client.StartCell` — the daemon creates a new task in the existing containerd snapshot, so the overlay filesystem (project repo clone, `.claude.json`, any per-cell state) survives the stop/start transition. State preservation across cron ticks is the headline win.
- On an empty pool — first tick, post-reboot, or every clone already Running — `--reuse` transparently falls through to `--clone`'s code path (atomic gap-fill counter allocation, new clone Config + cell). The operator never sees a "pool empty" error.
- Concurrent `--reuse` invocations against the same source pick distinct cells: the daemon's `StartCell` rejects a duplicate Start on a Ready cell with `must first be stopped`, and the loser transparently advances to the next pool candidate. Running clones are invisible to the pool query; Pending / Failed / Unknown sub-states are excluded.
- Mutex set per AC: `--reuse` ↔ `--new`, `--clone`, `--name`, `--rm`. Rejected on `-f`/`-p`/`-b` (only the `<config>` positional reaches the daemon-stored Config pool).
- Refactored `loadCellDoc` to return a `loadResult` struct (`Doc`, `PreStarted`, `StartResult`) so the `--reuse`-claim path can signal "I already StartCell'd this cell" and runRun can skip its own GetCell → existing-cell branch and emit the matching-spec + Started rollup directly. Non-config sources return `PreStarted: false` and walk the standard create-or-attach flow unchanged.

## Why I chose this shape

- **Atomic claim via `StartCell` rather than a new "claim" RPC.** The existing `controller.StartCell` already refuses a Start on a Ready or running cell ("must first be stopped"); reusing that guard means no new daemon-side primitive and no protocol change. `pickAndStartReusableClone` recognises the race-error substring and advances rather than surfacing it. Documented in `reuse.go` `isStartCellRace`.
- **Falls back to `--clone` on empty pool, not a separate code path.** `loadFromConfig` calls `cloneCellConfig` (the existing #839 helper) when `pickAndStartReusableClone` returns `errReusePoolEmpty`. Same gap-fill atomic-claim machinery, same annotation lineage, no duplication.
[...]

PR #872, opened 2026-05-25
fix(stop/kill): preserve containerd container record + overlay snapshot

### LINKED ISSUE #867: stop/kill cell delete the containerd container record (and snapshot) — overlay wiped on every stop/start; violates --reuse contract
## Context

`kuke stop cell`, `kuke kill cell`, and `kuke delete cell` are documented as three escalating verbs distinguished by what they tear down (`docs/cli-use-cases.md:374`):

| Verb     | Documented semantics                                                                 |
| -------- | ------------------------------------------------------------------------------------ |
| `stop`   | Graceful SIGTERM to the cell's containers; leaves metadata in `Stopped` state.       |
| `kill`   | Immediate SIGKILL of containerd tasks; leaves metadata in `Stopped` state.           |
| `delete` | Removes metadata; refuses if the cell still has running containers.                  |

Standard containerd-tier semantics agree with this split: `ctr task kill` kills the task and leaves the container record; `ctr container delete` removes the record + snapshot.

The current implementation does not honor that split. Both `StopCell` and `KillCell` follow their task-shutdown step with `DeleteContainer{SnapshotCleanup: true}`, which removes the containerd container record **and the overlay snapshot** — exactly the work `DeleteCell` is supposed to own exclusively.

Code anchors:

- `internal/controller/runner/kill.go:160-163` (workload containers, in `killCellLocked`)
- `internal/controller/runner/kill.go:228-230` (root container, in `killCellLocked`)
- `internal/controller/runner/kill.go:433-435` (single-container `KillContainer`)
- `internal/controller/runner/stop.go:179-180` (workload containers, in `stopCellLocked`)
- `internal/controller/runner/stop.go:250-251` (root container, in `stopCellLocked`)
- `internal/controller/runner/stop.go:464-465` (single-container `StopContainer`)

All six call sites pass `SnapshotCleanup: true`, so the writable overlay layer is wiped along with the container record.

`DeleteCell` (`internal/controller/runner/delete_cell.go:33`) already owns the full teardown via `stopAndDeleteContainer` (lines 136, 147, 170), plus cgroup removal and metadata-file removal. Stop/kill are duplicating its containerd-tier work without doing the metadata removal.

## Why this matters

1. **Contradicts the `--reuse` feature.** #835 / #848 (`feat(run): add --reuse — restart a healthy-Stopped clone, preserving overlay`) explicitly says it "preserves the containerd overlay filesystem across the stop/start transition." For that to be true, `stop cell` must leave the container record + snapshot in place. Today it doesn't, so `--reuse` is silently restarting against a freshly-created overlay every cycle, not a reused one. The feature's headline guarantee is currently violated.

2. **Forces unnecessary state loss on every restart.** Anything written to the writable overlay layer between cell starts is wiped. The only way to keep state across a `stop`/`start` today is to bind-mount it from the host, which not every operator workflow does (and shouldn't have to). Concretely: a `kuke kill cell` + `kuke start cell` on a working cell wipes anything in `/home/<user>/` that wasn't a bind mount — including repo clones an `onInit` script just produced.

3. **Contradicts standard containerd semantics.** An operator familiar with `ctr task kill` vs `ctr container delete` will be surprised to find `kuke kill cell` did both. The mental model the docs promise is the standard one; the code diverges silently.

4. **Loses a useful debugging tier.** With the documented split, an operator can `kuke kill cell <name>` and then poke at the on-disk overlay layer before `kuke start cell <name>` brings it back. Today that workflow is impossible — the kill flattened the overlay.

## Observed end-to-end

```
$ kuke run -p kukeon-pm --name kukeon-pm-0
[cell created, image pulled, overlay populated by onInit's git clones]

$ kuke kill cell kukeon-pm-0
[expected: task killed, container record + overlay preserved]
[actual:   task killed, container record gone, overlay gone]

$ kuke start cell kukeon-pm-0
[expected: task starts against the preserved overlay; cloned repos still there]
[actual:   CreateContainer rebuilds the container; overlay is empty; onInit re-clones from scratch]
```

The user-observable symptom is "my cell restarts without my repo clones / installed packages / written files." The architectural symptom is "stop/kill are doing delete's work."

## Proposal

Move all `DeleteContainer` calls out of `kill.go` and `stop.go`. `delete_cell.go` is already the sole caller of `stopAndDeleteContainer` for the documented full teardown — keep that as the only path that touches the containerd container record.

Specifically:

1. **`stopCellLocked` / `StopContainer` (stop.go):** drop the `DeleteContainer{SnapshotCleanup: true}` calls at `stop.go:179-180`, `stop.go:250-251`, `stop.go:464-465`. The `StopContainer` task-shutdown step stays; the CNI detach on the root container stays; the host-side attach-socket inode cleanup stays; the `internalCell.Status.State = CellStateStopped` stamp stays.

2. **`killCellLocked` / `KillContainer` (kill.go):** symmetrically, drop `DeleteContainer{SnapshotCleanup: true}` at `kill.go:160-163`, `kill.go:228-230`, `kill.go:433-435`. The `killContainerTask` step stays; rest of the cleanup stays.

3. **`delete_cell.go`:** unchanged — already does the right thing via `stopAndDeleteContainer`.

4. **Verify `StartCell` / `StartContainer` resume against existing container records.** They already tolerate `container 'root': already existed` / `container 'work': already existed` on the CLI surface (visible in `kuke run`'s reuse print path), so the resume should already work; the fix should add explicit test coverage that a `stop → start` cycle reuses the same containerd container ID and the same snapshot.

5. **Verify `--reuse` (#848) actually preserves the overlay** post-fix. The end-to-end test should write a marker file into the work container's writable layer pre-stop, `kuke stop cell`, `kuke start cell` (or `--reuse`), and assert the marker file is still there.

Implementer is free to scope the test coverage as they see fit, but the marker-file round-trip is the headline behavioral check this issue is asking for.

## Acceptance criteria

- [ ] `kuke stop cell <name>` followed by `sudo ctr -n <ns> containers ls` shows the cell's container records still present. The task is gone (`tasks ls` empty for the cell); the container is not.
- [ ] `kuke kill cell <name>` has the same containerd-tier postcondition as stop (records present, task gone).
- [ ] A file written into the work container's writable overlay before `kuke stop cell` survives `kuke start cell` (i.e., the snapshot is preserved across the stop/start transition).
- [ ] `StartCell`'s existing-container branch (`internal/controller/runner/start.go:651-705`) no longer calls `DeleteContainer{SnapshotCleanup: true}` on the happy path. The destructive recreate moves to the explicit mutation verbs (`RecreateCell` / `UpdateCell`), which already own it; the marker-file AC above will not pass without this change, because today `StartCell` wipes the snapshot on every restart when the root container record exists.
- [ ] At container-create / container-recreate time, every kukeon-managed containerd container record carries a `kukeon.io/spec-hash` label whose value is a deterministic hash of the CellSpec fields baked into the OCI spec (image, command, args, env, mounts, security context, resources — i.e., the domain `apply.DiffCell` classifies as "requires containerd recreate"). Stamped per-container by every create path: `start.go` (root, fresh-record branch), `create_container.go` (children), and the recreate paths in `recreate_cell.go` / `update_cell.go`. A unit test pins the hash domain to `DiffCell`'s breaking-field set so the two cannot silently drift apart.
- [ ] `StartCell`'s existing-container branch (post the AC above that removes the destructive recreate) recomputes the spec-hash from the on-disk CellSpec and compares against the record's `kukeon.io/spec-hash` label before deciding what to do. Match → start the task on the existing record (overlay preserved). Mismatch → refuse with an actionable error naming the diverged field and pointing the operator to `kuke apply -f <cell.yaml>` to reconcile. Mismatch is only reachable via crash-mid-apply or an unsupported hand-edit of `metadata.yaml`; the supported `kuke apply -f` flow re-stamps the hash as part of the same `RecreateCell` / `UpdateCell` transaction that rewrites containerd state, so steady-state restarts always match. (Post-#823, `kuke apply -f` is the sole producer of CellSpec mutation, so the supported mutation surface has exactly one entry point.)
- [ ] `kuke delete cell <name>` continues to remove the containerd container records + snapshots + cgroup + kukeon metadata file (no regression on the full-teardown verb).
- [ ] `--reuse` (#848) restarts a Stopped clone against the same overlay snapshot it had pre-stop; an end-to-end test asserts the marker-file invariant above.
- [ ] Unit/integration tests cover the new stop/kill semantics (container record present after stop; snapshot preserved across stop/start; delete still wipes everything).

## Notes

- Related closed work: #835 / #848 (`--reuse` overlay-preservation contract — currently silently broken because of this bug).
- Code anchors: `kill.go:160-163, 228-230, 433-435`; `stop.go:179-180, 250-251, 464-465`; `delete_cell.go:33 DeleteCell` (already correct).
- Doc reference: `docs/cli-use-cases.md:374` (the verb-semantics table this issue restores conformance to).
- Surfaced while diagnosing a separate `kukeon-pm-0` un-attachable state: the recovery instruction "`kuke kill cell` then `kuke start cell`" was costing the operator their entire overlay (cloned repos, installed packages) when it shouldn't have. That separate un-attachable bug is its own follow-on and not in scope here.

eminwux:
## Summary

- `kuke stop cell` / `kuke kill cell` / `kuke stop container` / `kuke kill container` no longer tear down the containerd container record (and overlay snapshot) — that work is left to the `delete` verb, restoring the documented stop/kill/delete split (`docs/cli-use-cases.md:374`). The headline operator-visible consequence is that a `stop → start` cycle now preserves anything written to the writable overlay (cloned repos from an `onInit`, installed packages, marker files), which is the invariant `--reuse` (#835 / #848) silently failed.
- Adds a `kukeon.io/spec-hash` label stamped on every kukeon-managed containerd container record at create time. The hash is a SHA-256 over the CellSpec field set `apply.DiffCell` classifies as "requires containerd recreate" (image, command, args — both root via `rootContainerSpecChanged` and non-root via `update_cell.go::containerSpecChanged`). A unit test pins the hash domain to `DiffCell`'s breaking-field set so the two cannot silently drift apart (issue AC #5 / #6).
- `StartCell` / `StartContainer` no longer wipe the existing containerd record on the happy path. Instead, they read the `kukeon.io/spec-hash` label, compare it to the freshly-computed hash, and:
  - **match** → drop any stale task and resume on the preserved snapshot;
  - **mismatch** → refuse with `errdefs.ErrCellSpecHashDrift` naming both hashes and pointing the operator at `kuke apply -f` (the supported mutation surface re-stamps the label inside the same `RecreateCell` / `UpdateCell` transaction that rewrites containerd state);
  - **legacy record with no label** → treated as match for backward compat so the first restart after this upgrade does not refuse a healthy cell.

## Hash-domain interpretation (reviewer pushback welcome)

The issue body lists "image, command, args, env, mounts, security context, resources" as the hash payload but also says the domain must equal what `apply.DiffCell` classifies as "requires containerd recreate". Those two are inconsistent — `DiffCell` classifies env / mounts / security / resources as Compatible (the apply layer handles them via in-place `UpdateCell`, not recreate). I read the AC's "pin the hash domain to `DiffCell`'s breaking-field set" as binding and went with the narrow domain (image, command, args). A wider hash would force `ErrCellSpecHashDrift` on every `kuke apply -f` that tweaked a Compatible field, defeating `UpdateCell` entirely. The pinning test catches a future widening of `DiffCell` so the two stay in lock-step. Happy to switch to the wider domain if the reviewer prefers the literal AC text.

## AC carve-out: child containers on the StartCell hot path

Issue AC #4 cites the root-container code window (`start.go:651-705`) for the destructive recreate removal. The same destructive `DeleteContainer{SnapshotCleanup: true}` pattern existed for child containers in the StartCell loop (`start.go:1015-1019`) and in single-container `StartContainer` (`start.go:1322-1326`); both are extended with the same spec-hash guard. Otherwise a `stop cell` + `start cell` would preserve the root overlay but wipe child overlays, which would not satisfy the marker-file behavioral check the issue is asking for.

## Test plan

- [x] `GOWORK=off make test` — full repo unit + integration suite, all green
- [x] New `internal/controller/runner/spec_hash_test.go`: determinism, nil-vs-empty args parity, image/command/args change the hash, compatible fields don't, **pinning to `apply.DiffCell` breaking-field set** (AC #5)
- [x] New `internal/controller/runner/stop_kill_test.go`: `StopCell` / `KillCell` / `StopContainer` no longer call `DeleteContainer` (issue ACs #1, #2)
- [x] New `internal/controller/runner/spec_hash_guard_test.go`: `reuseOrRefuseExistingChildContainer` happy / mismatch / legacy-label / absent-record matrix; the diverged-hash error wraps `ErrCellSpecHashDrift` and surfaces both hashes + `kuke apply -f`
- [x] e2e suite compiles (`go test -run XXX ./e2e/`)
- [ ] `make dev-init` (containerd-on-host) two-realm parity check — **deferred to reviewer's environment**: the dev host has no `kuke daemon` running and `make dev-init` requires a privileged containerd + cgroup-v2 host, which the agent sandbox cannot provide safely. The marker-file `stop → write → start` invariant (`echo X > /work/marker; kuke stop cell; kuke start cell; cat /work/marker`) is the headline behavioral check this PR enables and should be exercised on a real cell before merge.
- [ ] e2e suite run — **deferred to reviewer's environment**: same containerd-host dependency as `make dev-init`

Closes #867

### DIFF
--- a/internal/controller/runner/kill.go
+++ b/internal/controller/runner/kill.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/eminwux/kukeon/internal/ctr"
 	"github.com/eminwux/kukeon/internal/errdefs"
 	intmodel "github.com/eminwux/kukeon/internal/modelhub"
 )
@@ -157,28 +156,11 @@ func (r *Exec) killCellLocked(cell intmodel.Cell) (intmodel.Cell, error) {
 			fields...,
 		)
 
-		// Delete container after killing
-		err = r.ctrClient.DeleteContainer(internalRealm.Spec.Namespace, containerID, ctr.ContainerDeleteOptions{
-			SnapshotCleanup: true,
-		})
-		if err != nil {
-			// Log warning but don't fail - container might already be deleted
-			fields = appendCellLogFields([]any{"id", containerID}, cellID, cellName)
-			fields = append(fields, "space", spaceID, "realm", realmID, "err", fmt.Sprintf("%v", err))
-			r.logger.WarnContext(
-				r.ctx,
-				"failed to delete container, continuing",
-				fields...,
-			)
-		} else {
-			fields = appendCellLogFields([]any{"id", containerID}, cellID, cellName)
-			fields = append(fields, "space", spaceID, "realm", realmID)
-			r.logger.InfoContext(
-				r.ctx,
-				"deleted container",
-				fields...,
-			)
-		}
+		// Issue #867: leave the containerd container record (and snapshot)
+		// in place. kill is the SIGKILL-escalated cousin of stop; the
+		// documented stop/kill/delete split (docs/cli-use-cases.md) only
+		// distinguishes the two on signal escalation, not on snapshot
+		// teardown. delete still owns record + snapshot teardown.
 	}
 
 	// Kill root container last and detach from CNI
@@ -224,28 +206,10 @@ func (r *Exec) killCellLocked(cell intmodel.Cell) (intmodel.Cell, error) {
 		)
 	}
 
-	// Delete root container after killing
-	err = r.ctrClient.DeleteContainer(internalRealm.Spec.Namespace, rootContainerID, ctr.ContainerDeleteOptions{
-		SnapshotCleanup: true,
-	})
-	if err != nil {
-		// Log warning but don't fail - container might already be deleted
-		fields := appendCellLogFields([]any{"id", rootContainerID}, cellID, cellName)
-		fields = append(fields, "space", spaceID, "realm", realmID, "err", fmt.Sprintf("%v", err))
-		r.logger.WarnContext(
-			r.ctx,
-			"failed to delete root container, continuing",
-			fields...,
-		)
-	} else {
-		fields := appendCellLogFields([]any{"id", rootContainerID}, cellID, cellName)
-		fields = append(fields, "space", spaceID, "realm", realmID)
-		r.logger.InfoContext(
-			r.ctx,
-			"deleted root container",
-			fields...,
-		)
-	}
+	// Issue #867: leave the root container record + snapshot in place — see
+	// the matching comment in the workload-container loop above. CNI cleanup
+	// below still runs so the IPAM reservation is released; without that, the
+	// next StartCell's CNI ADD would be rejected as a duplicate allocation.
 
 	// Always run comprehensive CNI cleanup after container deletion as a safety net
 	// This ensures IPAM allocations are cleaned up even if CNI DEL succeeded or failed
@@ -429,38 +393,9 @@ func (r *Exec) KillContainer(cell intmodel.Cell, containerID string) error {
 		fields...,
 	)
 
-	// Delete container after killing
-	err = r.ctrClient.DeleteContainer(internalRealm.Spec.Namespace, containerdID, ctr.ContainerDeleteOptions{
-		SnapshotCleanup: true,
-	})
-	if err != nil {
-		// Log warning but don't fail - container might already be deleted
-		fields = appendCellLogFields([]any{"id", containerdID}, cellID, cellName)
-		fields = append(
-			fields,
-			"space",
-			spaceName,
-			"realm",
-			realmName,
-			"containerName",
-			containerID,
-			"err",
-			fmt.Sprintf("%v", err),
-		)
-		r.logger.WarnContext(
-			r.ctx,
-			"failed to delete container, continuing",
-			fields...,
-		)
-	} else {
-		fields = appendCellLogFields([]any{"id", containerdID}, cellID, cellName)
-		fields = append(fields, "space", spaceName, "realm", realmName, "containerName", containerID)
-		r.logger.InfoContext(
-			r.ctx,
-			"deleted container",
-			fields...,
-		)
-	}
-
+	// Issue #867: leave the containerd container record (and snapshot) in
+	// place. The cell-wide cleanup comments in killCellLocked above carry
+	// the full rationale; the single-container `kuke kill container` verb
+	// mirrors the cell-wide kill verb's contract.
 	return nil
 }
--- a/internal/controller/runner/provision.go
+++ b/internal/controller/runner/provision.go
@@ -1510,7 +1510,7 @@ func (r *Exec) createCellContainers(cell *intmodel.Cell) (containerd.Container, 
 				cell.Spec.Containers[i] = containerSpec
 			}
 
-			rootLabels := buildRootContainerLabels(*cell)
+			rootLabels := stampSpecHashOnLabels(buildRootContainerLabels(*cell), containerSpec)
 			ctrContainerSpec := ctr.BuildRootContainerSpec(containerSpec, rootLabels, r.daemonDefaultBuildOpts()...)
 
 			createdContainer, createErr = r.ctrClient.CreateContainer(namespace, ctrContainerSpec, creds)
@@ -1574,6 +1574,9 @@ func (r *Exec) createCellContainers(cell *intmodel.Cell) (containerd.Container, 
 				return nil, fmt.Errorf("failed to prepare attachable container %s: %w", containerdID, attachErr)
 			}
 			buildOpts := append(r.daemonDefaultBuildOpts(), attachOpts...)
+			buildOpts = append(buildOpts, ctr.WithExtraLabels(map[string]string{
+				SpecHashLabelKey: ComputeContainerSpecHash(containerSpec),
+			}))
 			// Merge `kuke run --env` runtime env into the attachable
 			// container's spec env (issue #834). Returns containerSpec
 			// unchanged for non-attachable containers or when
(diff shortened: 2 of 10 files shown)


eminwux:
**PR #872 Review — `fix(stop/kill): preserve containerd container record + overlay snapshot`**

LGTM — restores the documented stop/kill/delete split, adds the spec-hash guard the issue's AC asked for, and pins the hash domain to `apply.DiffCell`'s breaking field set with a regression test (`spec_hash_test.go::TestSpecHashDomainPinsToDiffCellBreakingFields`) so the two cannot drift. The AC-carve-out for child-container windows in `start.go:1015` / `start.go:1322` is the right call — without extending the same guard to the StartCell child-loop and the single-container `StartContainer`, a `stop cell` + `start cell` would preserve the root overlay but wipe child overlays, defeating the marker-file behavioral check. The narrow hash domain (image/command/args) over the literal AC list is well-reasoned and the pinning test catches future widening. Namespace-path staleness across stop/start cycles is handled by the existing `applySpecOpts` clear-and-reapply mechanism in `internal/ctr/spec.go:104-108`, so the reuse path is safe. The two deferred `make dev-init` / e2e items are appropriately gated on the containerd-host environment the agent sandbox cannot provide; the marker-file invariant should be exercised on a real cell pre-merge.

PR #873, opened 2026-05-25
fix(init): recreate kukeond cell on image drift and verify post-stop reset

### LINKED ISSUE #868: kuke daemon reset + kuke init does not refresh kukeond cell when image tag is reused

eminwux:
## Summary

- `bootstrapCell` now compares the persisted kukeond root container image against the desired `--kukeond-image` and routes through a `RecreateCell` (stop + delete + rebuild root container) branch on drift instead of the idempotent `EnsureCell + StartCell` path that would silently restart the stale image. The init output renders `cell "kukeond": recreated (image drift: <old> → <new>)` so the operator sees the rebuild rather than the misleading "already existed" line.
- `lifecycle.StopPhase` now runs a post-stop verification after `StopCell` returns `Stopped=true`: a follow-up `GetCell` confirms no container task remains `Ready`. If one does (the runner's per-container `StopContainer` errors are logged-and-continue rather than aggregated into the `StopCell` result, so a soft-fail surfaces here), `KillCell` is escalated and the cell is re-checked. A surviving task even after kill surfaces a hard error so the caller (`kuke daemon reset`) refuses to proceed to `DeleteCell` against a live shim, addressing #868's reset-path AC.

## Scope notes

- **Drift detection is ref-string-based, not digest-based.** A persisted image of `docker.io/library/kukeon-local:dev` matched against a desired `docker.io/library/kukeon-local:dev@<different-sha>` (the canonical `make dev-init` rebuild-the-same-tag case) does **not** trip the recreate branch — the strings match. PR #869 (Layer 1) is the chain-digest defense for this scenario at the script layer. The init-side digest-aware comparison would need a runner method to inspect the running container's resolved image digest and is left to a follow-up; the reset-path fix below is what unsticks #857's `make dev-init` repro in practice.
- The reset-path verification reuses the existing `KillCell` escalation — no new RPC surface; only the call flow inside `StopPhase` changed. The escalation notice prints `"force-killed after stop returned success but task survived"` so the operator distinguishes it from the existing `"force-killed after <timeout> grace period expired"` path.
- The post-stop verification is shared across `kuke daemon stop`/`restart`/`reset` (all of which call `StopPhase`); the corresponding table-driven tests were updated to make their fake `getCellFn` stateful (Ready on the pre-stop branch-picker call, Stopped on the post-stop verify) so the existing assertions on the success path still hold.

## Test plan

- [x] `make test` — full suite passes locally (cached after the first run).
- [x] `go vet ./...` clean.
- [x] `make dev-init` — full bootstrap loop passes including the daemon-parity check, the chain-digest guard (`running kukeond container chain ... matches just-built`), the kuke attach smoke against a kuketty-wrapped cell, and the nested-mode parent attach plumbing preservation check.
- [x] Direct repro of the same-image idempotent path (`sudo ./kuke init --kukeond-image docker.io/library/kukeon-local:dev` on a warm host): renders `cell "kukeond": already existed`, no spurious recreate.
- [x] New unit tests: `TestBootstrapCell_ImageDriftRecreatesCell`, `TestBootstrapCell_SameImageEnsuresInPlace`, `TestBootstrapCell_NoPriorMetadataCreatesFresh`, `TestStopPhase_GracefulSuccessButTaskSurvivesEscalates`, `TestStopPhase_PostKillVerificationStillRunningErrors`, `TestDaemonReset_StopSucceedsButTaskSurvivesEscalatesToKill`, `TestDaemonReset_StopSucceedsAndKillCannotClearSurvivor`.

Closes #868

### DIFF
--- a/cmd/kuke/daemon/internal/lifecycle/lifecycle.go
+++ b/cmd/kuke/daemon/internal/lifecycle/lifecycle.go
@@ -165,7 +165,18 @@ func StopPhase(
 			"kukeond stopped (cell %q in realm %q)\n",
 			consts.KukeSystemCellName, consts.KukeSystemRealmName,
 		)
-		return nil
+		// StopCell can return Stopped=true while a container task survives —
+		// the runner's per-container StopContainer errors are logged-and-
+		// continue rather than aggregated into the StopCell result (see
+		// internal/controller/runner/stop.go), so a containerd shim that
+		// rejects SIGTERM or a stop call that returns before the task exits
+		// surfaces here as "successful stop" with a still-Ready container.
+		// `kuke daemon reset` would then proceed to DeleteCell against a
+		// surviving kukeond task, leaving the operator's daemon binary
+		// frozen on the prior build across a re-bootstrap. Issue #868.
+		// Verify the post-stop state and escalate to KillCell if a task
+		// remains.
+		return verifyStoppedOrEscalate(cmd, client, doc)
 	case <-time.After(timeout):
 		cancel()
 		killRes, killErr := client.KillCell(cmd.Context(), doc)
@@ -184,4 +195,54 @@ func StopPhase(
 		)
 		return nil
 	}
+}
+
+// verifyStoppedOrEscalate re-reads the cell after a successful StopCell, and
+// if a container task is still Ready, drives the SIGKILL escalation that
+// would otherwise have fired on grace-period timeout. The post-kill GetCell
+// is the second gate: if the task remains Ready even after KillCell claimed
+// to terminate it, the caller must surface a hard error rather than allow
+// the subsequent DeleteCell to race against a live shim.
+func verifyStoppedOrEscalate(
+	cmd *cobra.Command,
+	client kukeonv1.Client,
+	doc v1beta1.CellDoc,
+) error {
+	getRes, getErr := client.GetCell(cmd.Context(), doc)
+	if getErr != nil {
+		return fmt.Errorf("re-inspect kukeond cell after stop: %w", getErr)
+	}
+	if !getRes.MetadataExists || !IsCellRunning(getRes.Cell) {
+		return nil
+	}
+
+	killRes, killErr := client.KillCell(cmd.Context(), doc)
+	if killErr != nil {
+		return fmt.Errorf(
+			"escalate to kill kukeond cell after stop returned success but task survived: %w",
+			killErr,
+		)
+	}
+	if !killRes.Killed {
+		return fmt.Errorf(
+			"escalate to kill kukeond cell after stop returned success but task survived: %w",
+			errdefs.ErrControllerNoChange,
+		)
+	}
+
+	verifyRes, verifyErr := client.GetCell(cmd.Context(), doc)
+	if verifyErr != nil {
+		return fmt.Errorf("re-inspect kukeond cell after escalated kill: %w", verifyErr)
+	}
+	if verifyRes.MetadataExists && IsCellRunning(verifyRes.Cell) {
+		return fmt.Errorf(
+			"kukeond cell %q in realm %q still reports a running container after stop and escalated kill",
+			consts.KukeSystemCellName, consts.KukeSystemRealmName,
+		)
+	}
+	cmd.Printf(
+		"kukeond force-killed after stop returned success but task survived (cell %q in realm %q)\n",
+		consts.KukeSystemCellName, consts.KukeSystemRealmName,
+	)
+	return nil
 }
--- a/cmd/kuke/daemon/internal/lifecycle/lifecycle_test.go
+++ b/cmd/kuke/daemon/internal/lifecycle/lifecycle_test.go
@@ -92,10 +92,108 @@ func TestStopPhase_GracefulSuccess(t *testing.T) {
 			t.Fatal("KillCell must not run when StopCell succeeds within the grace period")
 			return kukeonv1.KillCellResult{}, nil
 		},
+		// Post-stop verification (issue #868) re-reads the cell to confirm
+		// the task is actually gone; returning Stopped clears the gate so
+		// the escalation path is not taken.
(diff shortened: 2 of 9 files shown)


eminwux:
**PR #873 Review — `fix(init): recreate kukeond cell on image drift and verify post-stop reset`**

LGTM — the drift branch correctly suppresses the post-switch `StartCell` (since `RecreateCell` runs it internally), `verifyStoppedOrEscalate` wraps the no-change sentinel with `%w`, and the test matrix covers all six paths (drift / same-image / no-prior-metadata on init, plus graceful-success / survive-stop / survive-kill on reset).

Issue #605, opened 2026-05-18
cli: kuke get container — trim default to add RESTARTS+AGE, add -o wide with IMAGE+EXIT


eminwux:
## Context

Per-entity column changes for `kuke get container` in the 5-issue series that redefines what each `kuke get <kind>` table shows. Cross-cutting infrastructure (the `wide` output value, the AGE helper, the rule that `CGROUP` only appears in `-o yaml`/`-o json`) lands in #601; this issue is scoped to container's table composition only. Container does not have `--show-controllers` today, so no flag removal is needed here.

Container is the densest entity (~30 spec fields) and motivates this series most strongly. The new default imports kubectl's `RESTARTS` convention while preserving the full hierarchy (parallel to the other four kinds). `IMAGE` moves to `-o wide` — without it in wide, image is invisible in any table view.

## Proposal

In `cmd/kuke/get/container/container.go`:

- Default table becomes `NAME REALM SPACE STACK CELL STATE RESTARTS AGE` (8 cols). Drops `ROOT` and `IMAGE` from default.
  - `NAME` continues to substitute `root` for the displayed ID when `Spec.Root == true` (existing `containerDisplayName` behavior — unchanged).
  - `RESTARTS` = `Status.RestartCount`.
- `kuke get container -o wide` becomes `NAME REALM SPACE STACK CELL STATE RESTARTS AGE IMAGE EXIT` (10 cols).
  - `IMAGE` = `Spec.Image`.
  - `EXIT` = `ExitCode/ExitSignal` rendered as `<code>/<signal>` when either is non-zero/non-empty; `-` otherwise. Most meaningful on `Stopped`/`Failed` states.

## Acceptance criteria

- [ ] `kuke get container` default prints `NAME REALM SPACE STACK CELL STATE RESTARTS AGE` (8 cols).
- [ ] `kuke get container -o wide` prints `NAME REALM SPACE STACK CELL STATE RESTARTS AGE IMAGE EXIT` (10 cols).
- [ ] `NAME` renders `root` for `Spec.Root == true` containers (existing behavior preserved).
- [ ] `RESTARTS` renders `Status.RestartCount` as a plain integer.
- [ ] `EXIT` renders `<code>/<signal>` when either field is non-zero/non-empty; `-` otherwise.
- [ ] `CGROUP`, `ROOT` (as a column), and `IMAGE` (as a default column) do not appear in the default table.
[...]

Issue #604, opened 2026-05-18
cli: kuke get cell — trim default + -o wide with CONTAINERS (ready/total) and BRIDGE


eminwux:
## Context

Per-entity column changes for `kuke get cell` in the 5-issue series that redefines what each `kuke get <kind>` table shows. Cross-cutting infrastructure (the `wide` output value, the AGE helper, removal of `--show-controllers` and `CONTROLLERS` column, the rule that `CGROUP` only appears in `-o yaml`/`-o json`) lands in #601; this issue is scoped to cell's table composition only.

Cell is the first entity with meaningful runtime state — it owns the network bridge, has actual lifecycle (`Stopped` is a real state), and hosts containers. The wide view surfaces the two highest-signal cell-only fields.

## Proposal

In `cmd/kuke/get/cell/cell.go`:

- Default table becomes `NAME REALM SPACE STACK STATE AGE` (6 cols). Drops `CGROUP`.
- `kuke get cell -o wide` becomes `NAME REALM SPACE STACK STATE AGE CONTAINERS BRIDGE` (8 cols).
  - `CONTAINERS` = `ready/total` count from `Status.Containers` (e.g. `1/3`). Ready means `ContainerStatus.State == ContainerStateReady`. Empty container list renders as `0/0`.
  - `BRIDGE` = `Status.Network.BridgeName` (canonical `k-{8hex}` form). Renders `-` when empty.

## Acceptance criteria

- [ ] `kuke get cell` default prints `NAME REALM SPACE STACK STATE AGE` (6 cols).
- [ ] `kuke get cell -o wide` prints `NAME REALM SPACE STACK STATE AGE CONTAINERS BRIDGE` (8 cols).
- [ ] `CONTAINERS` renders `ready/total` correctly: counts entries in `Status.Containers` with `State == ContainerStateReady` for the numerator, total length for the denominator.
- [ ] `BRIDGE` renders the `Status.Network.BridgeName` value or `-` when empty.
- [ ] `CGROUP` does not appear in any cell table output.
- [ ] `cmd/kuke/get/cell/cell_test.go` updated for the new columns, including ready/total edge cases (0/0, 0/1, 1/1, 2/3).

## Notes

- Depends on #601 (shared `OutputFormatWide`, AGE helper, `--show-controllers` removal). Land #601 first.
[...]

Issue #603, opened 2026-05-18
cli: kuke get stack — trim default to NAME REALM SPACE STATE AGE


eminwux:
## Context

Per-entity column changes for `kuke get stack` in the 5-issue series that redefines what each `kuke get <kind>` table shows. Cross-cutting infrastructure (the `wide` output value, the AGE helper, removal of `--show-controllers` and `CONTROLLERS` column, the rule that `CGROUP` only appears in `-o yaml`/`-o json`) lands in #601; this issue is scoped to stack's table composition only.

Stack carries no spec field beyond hierarchy pointers, so `-o wide` has nothing meaningful to add over the default. `-o wide` is still accepted as an output value (for symmetry across all five kinds) — it just renders the same 5 columns.

## Proposal

In `cmd/kuke/get/stack/stack.go`:

- Default table becomes `NAME REALM SPACE STATE AGE` (5 cols). Drops `CGROUP`.
- `kuke get stack -o wide` accepts `wide` as a valid output value but renders the same 5 columns as default.

Update doc snippets that show the stack table:
- `docs/site/getting-started.md` line 89 (`NAME REALM SPACE STATE CGROUP`) → new 5-col shape.

## Acceptance criteria

- [ ] `kuke get stack` default prints `NAME REALM SPACE STATE AGE` (5 cols).
- [ ] `kuke get stack -o wide` is accepted and prints the same 5 columns (no error, no extra columns).
- [ ] `CGROUP` does not appear in any stack table output.
- [ ] `docs/site/getting-started.md` stack-table snippet updated.
- [ ] `cmd/kuke/get/stack/stack_test.go` updated for the new columns.

## Notes

- Depends on #601 (shared `OutputFormatWide`, AGE helper, `--show-controllers` removal). Land #601 first.
- Stack's lack of distinct wide columns is deliberate — adding `UPDATED`/`REASON` or a derived child-cell count would be noise at scan time. Operators who need stack-level investigation use `-o yaml`.

Issue #602, opened 2026-05-18
cli: kuke get space — trim default to NAME REALM STATE AGE, add -o wide with EGRESS NET-DEFAULTS


eminwux:
## Context

Per-entity column changes for `kuke get space` in the 5-issue series that redefines what each `kuke get <kind>` table shows. Cross-cutting infrastructure (the `wide` output value, the AGE helper, removal of `--show-controllers` and `CONTROLLERS` column, the rule that `CGROUP` only appears in `-o yaml`/`-o json`) lands in #601; this issue is scoped to space's table composition only.

## Proposal

In `cmd/kuke/get/space/space.go`:

- Default table becomes `NAME REALM STATE AGE` (4 cols). Drops `CGROUP`.
- `kuke get space -o wide` becomes `NAME REALM STATE AGE EGRESS NET-DEFAULTS` (6 cols).
  - `EGRESS` = `Spec.Network.Egress.Default` (`allow`/`deny`/`-` when unset). Renders `-` when `Spec.Network` or `Spec.Network.Egress` is nil.
  - `NET-DEFAULTS` = `yes`/`-` boolean of whether `Spec.Defaults != nil` (tells operators whether containers in this space inherit isolation policy).

Update doc snippets that show the space table:
- `docs/site/getting-started.md` line 85 (`NAME REALM STATE CGROUP`) → new 4-col shape.

## Acceptance criteria

- [ ] `kuke get space` default prints `NAME REALM STATE AGE` (4 cols).
- [ ] `kuke get space -o wide` prints `NAME REALM STATE AGE EGRESS NET-DEFAULTS` (6 cols).
- [ ] `EGRESS` renders `allow`/`deny`/`-` correctly across the nil-cases for `Spec.Network` and `Spec.Network.Egress`.
- [ ] `NET-DEFAULTS` renders `yes`/`-`.
- [ ] `CGROUP` does not appear in any space table output.
- [ ] `docs/site/getting-started.md` space-table snippet updated.
- [ ] `cmd/kuke/get/space/space_test.go` updated for the new columns.

## Notes

- Depends on #601 (shared `OutputFormatWide`, AGE helper, `--show-controllers` removal). Land #601 first.
- The `--show-controllers` flag removal for space happens in #601, not here.

PR #881, opened 2026-05-25
feat(get): shared OutputFormatWide + AGE helper, retire --show-controllers/CGROUP

### LINKED ISSUE #827: cli: kuke get — shared infra + AGE + realm cols (step 1)
## Context

Foundation step under the #601 umbrella (`epic:get`). Carries the shared infrastructure plus realm-specific columns; the sibling per-kind steps (#602, #603, #604, #605) assume the `OutputFormatWide` enum value and AGE helper introduced here.

See #601 for the cross-cutting motivation (overloaded default tables, missing AGE, asymmetric across kinds, no label-selector, `kuke get image` not a first-class get kind).

## Proposal

**Shared infrastructure (in `cmd/kuke/get/shared/shared.go`):**

- Add `OutputFormatWide` (`"wide"`) as a fourth value of the `OutputFormat` enum, accepted by `ParseOutputFormat` alongside `yaml`/`json`/`table`. Default (no `-o` flag) remains `table`.
- Add a small AGE helper that renders `time.Time` as a kubectl-style duration (e.g. `5d`, `2h`, `30s`) from `Status.CreatedAt`. Empty/zero time renders as `-`.
- Remove `FormatControllers` once all four entity commands stop using it.

**Cross-cutting cleanup (applies to `cmd/kuke/get/realm/realm.go`, `space/space.go`, `stack/stack.go`, `cell/cell.go`):**

- Remove the `--show-controllers` flag registration.
- Remove the `CONTROLLERS` column logic from each command's table builder.
- Container has no `--show-controllers` today; no change there.

**Realm-specific (in `cmd/kuke/get/realm/realm.go`):**

- Default table becomes `NAME STATE AGE` (3 cols). Drops `NAMESPACE` and `CGROUP`.
- `kuke get realm -o wide` becomes `NAME STATE AGE NAMESPACE` (4 cols). `CGROUP` never appears in any table; only in `-o yaml`/`-o json`.

**Regression-guard + docs updates landing in this PR:**

- `CLAUDE.md` "Local smoke test" section: update the expected `kuke get realms` output (currently `NAME NAMESPACE STATE CGROUP` with a `-` separator row) to the new 3-column shape.
- `docs/cli-use-cases.md` line 157 (`kuke get realms --show-controllers`) and the surrounding paragraph: remove.
- `docs/site/install/install-linux.md`, `docs/site/guides/local-dev.md`, `docs/site/cli/kuke-get.md`, `docs/site/getting-started.md`: update the inline realm-table snippets.

## Acceptance criteria

- [ ] `cmd/kuke/get/shared/shared.go` defines `OutputFormatWide` and `ParseOutputFormat` accepts `"wide"`.
- [ ] AGE helper lives in `cmd/kuke/get/shared/` and renders `5d`/`2h`/`30s`/`-` per kubectl convention.
- [ ] `--show-controllers` flag is removed from realm, space, stack, and cell commands.
- [ ] `CONTROLLERS` column is removed from realm, space, stack, and cell table output.
- [ ] `FormatControllers` helper is removed from `cmd/kuke/get/shared/shared.go`.
- [ ] `kuke get realm` default prints `NAME STATE AGE` (3 cols).
- [ ] `kuke get realm -o wide` prints `NAME STATE AGE NAMESPACE` (4 cols).
- [ ] `CGROUP` does not appear in any table output for any entity (verified by reading the four sibling command files — they should not reference `CgroupPath` in their table builders).
- [ ] `CLAUDE.md` regression-guard snippet shows the new 3-col realm table; `make dev-init` smoke test passes against the new shape.
- [ ] Doc snippets in `docs/site/install/install-linux.md`, `docs/site/guides/local-dev.md`, `docs/site/cli/kuke-get.md`, `docs/site/getting-started.md` reflect the new realm columns.
- [ ] `docs/cli-use-cases.md` removes the `--show-controllers` example.
- [ ] Existing realm/space/stack/cell tests under `cmd/kuke/get/*_test.go` updated for the new columns.

## Dependencies

Depends on: none
Blocks: #602, #603, #604, #605, #614, #824

## Notes

- Cross-cutting decisions captured here apply uniformly across the sibling steps; the cross-cutting context is not re-explained in each sibling body.
- `--no-daemon` flag is untouched; #223 retires it separately.
- The `--show-controllers` removal supersedes #328's user-facing surface; the underlying `Status.SubtreeControllers` field stays in the model for `-o yaml`/`-o json`.
- Migrated wholesale from the original #601 body (now the umbrella). No design changes from the prior framing — only the relocation out of the umbrella per `epic:get`'s split.

eminwux:
## Summary

Foundation step under the `epic:get` umbrella (#601). Lands the shared infrastructure the per-kind sibling steps (#602–#605) consume, plus the cross-cutting cleanup that retires `--show-controllers` and the `CGROUP`/`CONTROLLERS` table columns across the four hierarchy kinds.

- `shared.OutputFormatWide` (`"wide"`) is now the fourth `OutputFormat` enum value accepted by `ParseOutputFormat`; default with no `-o` flag remains `table`.
- `shared.RenderAge(time.Time, time.Time)` renders kubectl-style coarse durations (`5d`, `2h`, `30s`, `-` for zero) for the AGE column the siblings will adopt; `now` is injected for deterministic tests.
- `shared.FormatControllers` is gone — no callers remain after the cross-cutting cleanup.
- `kuke get realm` default table now `NAME STATE AGE`; `-o wide` appends `NAMESPACE`. `CGROUP`/`CONTROLLERS` are no longer rendered in any table for any of the four kinds (still visible via `-o yaml`/`-o json` through `Status.CgroupPath` / `Status.SubtreeControllers`).
- `--show-controllers` removed from realm, space, stack, and cell commands; container never had it.
- Space/stack/cell tables only drop CGROUP in this PR — sibling per-kind issues (#602, #603, #604) introduce their AGE column and the per-kind wide-only fields (`EGRESS`/`NET-DEFAULTS`, `CONTAINERS`/`BRIDGE`, etc.); accepting `-o wide` is wired here so the enum is usable from the kuke get plane immediately.
- `-o wide` is also accepted on the four leaf gets (`container`, `blueprint`, `secret`, `config`) — output is currently identical to `-o table`, sibling per-kind issues add their wide-only columns. Help strings and `CompleteOutputFormat` shell completion enumerate `wide`.
- Docs updated: CLAUDE.md daemon-parity tail, `docs/cli-use-cases.md`, `docs/site/install/install-linux.md`, `docs/site/guides/local-dev.md`, `docs/site/cli/kuke-get.md`, `docs/site/getting-started.md`, plus the realm/space tables in the top-level `README.md` (not enumerated in the AC but visibly stale otherwise).

## Notes for the reviewer

- The AC item "`CGROUP` does not appear in any table output for any entity (verified by reading the four sibling command files — they should not reference `CgroupPath` in their table builders)" is read as a hard end-state for step 1: this PR drops `CGROUP` from all four hierarchy kinds rather than waiting for the per-kind siblings to do it. The Proposal's per-entity-realm bullet ("`CGROUP` never appears in any table; only in `-o yaml`/`-o json`") confirms the broader rule. Sibling steps then layer their own AGE / wide columns onto each.
- `cell.go` and `image.go` keep their local `resolveOutput` shim that pre-translates `wide → table+bool`; once `shared.OutputFormatWide` is in the enum they could switch to consuming it directly, but that refactor is out of scope here and naturally lands with their per-kind sibling issues.
- The AGE rendering for realm tables uses `time.Now()` inside `printRealms`; the shared `RenderAge` helper accepts an injected `now` for tests, but the production call site reads wall-clock — there is no controller-injected clock to thread through `cmd/kuke/get` today. Sibling per-kind tables follow the same call shape.

## Test plan

- [x] `make test-root` — full root-module test suite (`go test ./...`)
- [x] `make test-kukebuild` — kukebuild module test suite
- [x] `pre-commit run --files $(git diff --name-only HEAD)` — all hooks pass (trailing-whitespace, end-of-file, go-fmt, go-mod-tidy, golangci-lint, prettier, addlicense)
- [x] `grep -rn 'show-controllers'` across the tree — only intentional history references (cli-use-cases.md retirement note) and negative-presence assertions in the four test files
- [ ] `make dev-init` end-to-end smoke + daemon-parity check — not run in this dev environment (no `kukeon-dev-root` parent host; the script's nested-mode probe would attempt to publish to `/run/kukeon-dev/` and is best exercised on a clean host). Daemon-parity expectation in CLAUDE.md updated to the new 3-col shape; the in-process test coverage in `cmd/kuke/get/realm` pins both default and `-o wide` column sets.

Closes #827

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -50,13 +50,13 @@ make dev-init
 The daemon-parity tail of the output (the regression guard) must read:
 
 ```
-NAME         NAMESPACE              STATE  CGROUP
------------  ---------------------  -----  -------------------
-default      default.kukeon.io      Ready  /kukeon/default
-kuke-system  kuke-system.kukeon.io  Ready  /kukeon/kuke-system
+NAME         STATE  AGE
+-----------  -----  ---
+default      Ready  <age>
+kuke-system  Ready  <age>
 ```
 
-Both `kuke get realms` and `kuke get realms --no-daemon` must produce that output. If only the `--no-daemon` list is populated, the daemon's view of `/opt/kukeon` diverged from the in-process controller — that's the bind-mount regression this check catches.
+Both `kuke get realms` and `kuke get realms --no-daemon` must produce the same shape (same columns, same rows, same `STATE`); `AGE` may differ by ms between the two invocations but renders at second granularity, so a single re-run normally produces byte-identical output. `NAMESPACE` and `CGROUP` no longer appear in the default table (epic:get retires them); use `kuke get realms -o wide` to append `NAMESPACE`, or `-o yaml`/`-o json` to surface `cgroupPath`. If only the `--no-daemon` list is populated, the daemon's view of `/opt/kukeon` diverged from the in-process controller — that's the bind-mount regression this check catches.
 
 **If your change touches anything the daemon reads, this check must be in the PR's test plan.** Reviewer agent will flag PRs that miss it.
 
--- a/README.md
+++ b/README.md
@@ -139,16 +139,16 @@ Common workflows for working with realms, spaces, stacks, and cells.
 
 ```bash
 $ sudo kuke get realms
-NAME           NAMESPACE         STATE    CGROUP
-default        kukeon-default    Running  /kukeon/default
-kukeon-system  kukeon-system     Running  /kukeon/kukeon-system
+NAME           STATE    AGE
+default        Ready    <age>
+kukeon-system  Ready    <age>
 
 $ sudo kuke get spaces
-NAME     REALM    STATE    CGROUP
-default  default  Running  /kukeon/default/default
+NAME     REALM    STATE
+default  default  Ready
 ```
 
-Add `-o yaml` or `-o json` for full resource details.
+Add `-o wide` for the per-kind extra columns (realm's wide appends `NAMESPACE`), or `-o yaml` / `-o json` for full resource details (including `cgroupPath`).
 
 ### Run a hello-world cell
 
--- a/cmd/config/autocomplete.go
+++ b/cmd/config/autocomplete.go
@@ -726,10 +726,10 @@ func CompleteProfileNames(_ *cobra.Command, _ []string, toComplete string) ([]st
 	return names, cobra.ShellCompDirectiveNoFileComp
 }
 
-// CompleteOutputFormat provides shell completion for output format values (yaml, json, table).
+// CompleteOutputFormat provides shell completion for output format values (yaml, json, table, wide).
 // This function can be used for flag completion in commands that accept output format flags.
 func CompleteOutputFormat(_ *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-	formats := []string{"yaml", "json", "table"}
+	formats := []string{"yaml", "json", "table", "wide"}
 
 	names := make([]string, 0, len(formats))
 	for _, format := range formats {
--- a/cmd/config/autocomplete_test.go
+++ b/cmd/config/autocomplete_test.go
@@ -936,7 +936,7 @@ func TestCompleteOutputFormat(t *testing.T) {
 		{
 			name:       "success with all formats",
 			toComplete: "",
-			wantNames:  []string{"yaml", "json", "table"},
+			wantNames:  []string{"yaml", "json", "table", "wide"},
 		},
 		{
 			name:       "success with prefix filter 'y'",
@@ -954,6 +954,11 @@ func TestCompleteOutputFormat(t *testing.T) {
 			wantNames:  []string{"table"},
 		},
 		{
+			name:       "success with prefix filter 'w'",
+			toComplete: "w",
+			wantNames:  []string{"wide"},
+		},
+		{
 			name:       "success with prefix filter 'ya'",
 			toComplete: "ya",
 			wantNames:  []string{"yaml"},
@@ -964,6 +969,11 @@ func TestCompleteOutputFormat(t *testing.T) {
 			wantNames:  []string{"table"},
 		},
 		{
+			name:       "success with prefix filter 'wi'",
+			toComplete: "wi",
+			wantNames:  []string{"wide"},
+		},
+		{
 			name:       "success with no matches",
 			toComplete: "x",
 			wantNames:  []string{},
@@ -971,7 +981,7 @@ func TestCompleteOutputFormat(t *testing.T) {
 		{
 			name:       "works without logger in context",
 			toComplete: "",
-			wantNames:  []string{"yaml", "json", "table"},
+			wantNames:  []string{"yaml", "json", "table", "wide"},
 			noLogger:   true,
 		},
 	}
--- a/cmd/kuke/get/blueprint/blueprint.go
+++ b/cmd/kuke/get/blueprint/blueprint.go
@@ -121,7 +121,7 @@ func NewBlueprintCmd() *cobra.Command {
 	cmd.Flags().String("stack", "", "Filter blueprints by stack name")
 	_ = viper.BindPFlag(config.KUKE_GET_BLUEPRINT_STACK.ViperKey, cmd.Flags().Lookup("stack"))
 	cmd.Flags().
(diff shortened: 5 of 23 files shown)


eminwux:
**PR #881 Review — `feat(get): shared OutputFormatWide + AGE helper, retire --show-controllers/CGROUP`**

Foundation step is clean — `OutputFormatWide` is wired into the parser, `RenderAge` covers every coarse-unit promotion (the table-driven test pins each boundary), and the CGROUP/CONTROLLERS retirement is consistent across realm/space/stack/cell tables, tests, and docs. Two leaf-coverage gaps need closing before this can ship without regressing the kinds it doesn't touch.

Blocking:

- **Silent `-o wide` → YAML regression** `cmd/kuke/get/container/container.go:332`, `cmd/kuke/get/blueprint/blueprint.go:162`, `cmd/kuke/get/secret/secret.go:166`, `cmd/kuke/get/config/config.go:162`: these four leaves call `shared.ParseOutputFormat` directly, but their `print*s` switch arms don't include `OutputFormatWide`, so `-o wide` falls through `default:` and silently emits YAML. Before this PR the parser rejected `wide` with an explicit error; after, the user gets the wrong output silently. `docs/cli-use-cases.md:163` adds the invariant *\"wide accepted by every `kuke get <kind>` for symmetry\"* — that doesn't hold. Fix: extend each leaf's `OutputFormatTable` case to `case shared.OutputFormatTable, shared.OutputFormatWide:` (matching realm/space/stack) and update the four `--output` help strings (still `\"(yaml, json, table)\"` at `container.go:62`, `blueprint.go:124`, `secret.go:127`, `config.go:124`).
- **Shell completion stale** `cmd/config/autocomplete.go:732`: `CompleteOutputFormat` still hardcodes `formats := []string{\"yaml\", \"json\", \"table\"}` (docstring at line 729 also stale). Every `get` subcommand registers this for `-o` tab-completion, so operators won't discover `wide` interactively even though the help text, parser, and reference docs all advertise it. Add `\"wide\"` to the slice, refresh the docstring, and add a `wide` case to `TestCompleteOutputFormat` (`cmd/config/autocomplete_test.go:929`).

Non-blocking:

- `printRealms` reads `time.Now()` directly — author called this out in the notes; fine for the foundation step.
- `make dev-init` smoke is host-gated and deferred per the test plan — recommend running it pre-merge to confirm the new daemon-parity tail renders byte-identically (the CLAUDE.md update reads correctly, but only a real run on the daemon path can confirm).

LGTM once `-o wide` no longer silently degrades on the four uncovered leaves and `CompleteOutputFormat` enumerates `wide`.

eminwux:
Addressed in f652536:
- `-o wide` regression on the four leaves (`container`, `blueprint`, `secret`, `config`): extended each plural printer's switch arm to `case shared.OutputFormatTable, shared.OutputFormatWide:` (matching realm/space/stack); updated each `--output` help string to enumerate `wide`. Output for `-o wide` is currently identical to `-o table` on these four kinds — sibling per-kind issues will add the wide-only columns. Singular printers fall through to YAML on `wide`, matching `printRealm`'s pattern.
- `CompleteOutputFormat` enumerates `wide`: added to the `formats` slice in `cmd/config/autocomplete.go:732`, docstring at line 729 refreshed, and `TestCompleteOutputFormat` extended with two `wide` prefix cases plus updated the all-formats and no-logger expectations.

`make test` (root + kukebuild), `go vet`, and `pre-commit` clean post-rebase onto current `main` (bdea276).

Non-blocking items in the review (`printRealms` wall-clock; `make dev-init` host-gated smoke) carried over from the original PR body unchanged.

eminwux:
**PR #881 Re-review — fix commit `f652536`**

LGTM — both blockers point-fixed exactly as suggested: `CompleteOutputFormat` enumerates `wide` (with `w`/`wi` prefix-filter cases added to `TestCompleteOutputFormat`), and `container/blueprint/secret/config` each extend their `OutputFormatTable` switch arm to `OutputFormatTable, OutputFormatWide` plus the help-text refresh. No scope creep beyond the prescribed fix. The base also moved (clean rebase onto current main), CI green on the new head.

PR #906, opened 2026-05-26
fix(uninstall): reclaim per-namespace BuildKit cache to prevent stale snapshot/content references

### LINKED ISSUE #904: uninstall: /var/lib/kukebuild/<ns>/ is not torn down; stale cache.db strands next kuke build with "parent snapshot does not exist" or "content digest ... not found"
## Summary

`kuke uninstall` purges every realm's containerd namespace and tears down
`/run/kukeon/` + the run path, but it is **unaware of `/var/lib/kukebuild/`**
— BuildKit's per-namespace state directory (`cache.db`, `history.db`,
snapshot metadata) that `kukebuild` keeps at `/var/lib/kukebuild/<ns>/` for
every realm `kuke build` has ever targeted (default base `/var/lib/kukebuild`,
see `cmd/kukebuild/main.go:211` and `resolveBuildRoot` at
`cmd/kukebuild/build.go:217`).

That cache references containerd by **two** kinds of ID: snapshot IDs and
content-store digests. When the backing objects go away (either by
`uninstall`'s own realm purge, or by an operator hand-cleaning containerd
after a partial uninstall), the BuildKit cache is left pointing at IDs
that no longer exist. Depending on which side drifts first, the next
`kuke build` into that realm fails with one of two errors — both
symptoms of the same root cause:

- **Snapshot drift** (cached layer's parent snapshot no longer in the snapshotter):
  ```
  failed to prepare <id-A> as <id-B>: parent snapshot <id-A> does not exist: not found
  ```
- **Content drift** (cached layer references a blob no longer in the content store):
  ```
  failed to extract layer sha256:<layer>: failed to get reader from content store:
  content digest sha256:<blob>: not found
  ```

## Repro (two realms, same underlying cause)

1. `kuke uninstall` → realm-purge step errored while deleting snapshots in
   `kuke-system.kukeon.io`. The earlier (successful) per-realm passes had
   already torn down content/snapshots in `default.kukeon.io`.
2. I cleaned the residual containerd state in `kuke-system.kukeon.io` by
   hand (leases + snapshots) and re-ran `make dev-init`. Both realms came
   back as fresh, empty containerd namespaces.
3. **Surface error #1 (snapshot variant) — kuke-system realm:**
   `kuke build -t kukeon-local:dev --realm kuke-system .` (driven by
   `make dev-init`) failed during `[builder 6/10] COPY cmd/ cmd/` with
   `parent snapshot <id> does not exist: not found`, repeatedly, even
   after `kuke daemon reset` and another round of containerd lease
   cleanup. Fix: `sudo rm -rf /var/lib/kukebuild/kuke-system.kukeon.io`,
   then `make dev-init` ran clean.
4. **Surface error #2 (content variant) — default realm:** later,
   `kuke build -t registry.eminwux.com/claude:latest .` (into the
   `default` realm) failed during the final unpack step with
   `failed to extract layer sha256:219a998c60509502b47b97f1158067d5dd62640d2d689560d32cfd5594f6bc40: failed to get reader from content store: content digest sha256:5b4d6ff92fc4e14e911b7753c954fac965d48c40fe1075758d284148ccace970: not found`.
   Same fix: `sudo rm -rf /var/lib/kukebuild/default.kukeon.io`.

The second failure is the more general case — the operator did
**nothing** by hand in the `default` namespace; the only thing that
disturbed it was `kuke uninstall`'s own (successful) realm purge. So
this is reproducible on a **clean** uninstall, not just the half-cleaned
edge case.

## Why uninstall doesn't catch this

- `cmd/kuke/uninstall/uninstall.go` and `internal/controller/` never
  reference `/var/lib/kukebuild` (grep is empty). The `uninstallLongDesc`
  enumerates the dirs it removes (`/run/kukeon/`, run path, user/group);
  the BuildKit state dir is not in that list.
- This is a problem on **both** paths:
  - **Clean uninstall**: the purge succeeds, snapshots/content are gone,
    but `/var/lib/kukebuild/<ns>/cache.db` still holds layer→snapshot and
    layer→content mappings. A subsequent `kuke init` + `kuke build` into
    the recreated namespace will hit one of the two errors above the
    moment a cached entry is requested.
  - **Half-cleaned uninstall** (the kuke-system case above): `uninstall`
    prints `skipped (realm purge failed): residual containerd namespace
    prevented teardown` and points operators at manual cleanup in
    `docs/cli-use-cases.md` — but that doc/remediation does not mention
    `/var/lib/kukebuild/<ns>/`, so operators don't know to wipe it.

## Expected

For every realm `uninstall` successfully purges, the matching
`/var/lib/kukebuild/<realm-namespace>/` directory should be removed in
the same pass, with a row in the uninstall report, e.g.:

```
- build cache /var/lib/kukebuild/kuke-system.kukeon.io: removed
- build cache /var/lib/kukebuild/default.kukeon.io: removed
```

On the half-cleaned path (`skipped (realm purge failed)`), the
remediation hint should also mention wiping `/var/lib/kukebuild/<ns>/`
after the operator clears the residual containerd namespace.

The base dir `/var/lib/kukebuild/` itself should only be removed if
empty after all per-namespace subdirs are gone (other realms may still
want their cache; only an unscoped uninstall should `rmdir` the base).

## Suggested fix sketch

1. In `controller.Uninstall`, after a realm's namespace is dropped, also
   `os.RemoveAll(filepath.Join("/var/lib/kukebuild", namespace))` and
   record the outcome on `UninstallReport` so `printReport` can render it.
2. Resolve the base path the same way `kukebuild` does (honor an
   override if one is configured) rather than hard-coding
   `/var/lib/kukebuild` in two places — extract a shared constant if
   that's the cleanest seam given the `cmd/kukebuild` separate-module
   boundary.
3. Update `uninstallLongDesc` and the half-cleaned-host remediation
   section of `docs/cli-use-cases.md` to mention the BuildKit cache dir.

## Workaround

```
sudo rm -rf /var/lib/kukebuild/<realm-namespace>
```

after `kuke uninstall` (or before re-running `kuke build` on a freshly
reinstalled host) — once per realm you intend to build into again.

eminwux:
## Summary

- Wires `kuke uninstall` to reclaim `/var/lib/kukebuild/<namespace>/` for every realm whose containerd namespace was successfully removed in the same pass. The BuildKit cache references containerd by snapshot ID and content digest, so a surviving cache after a namespace drop strands the next `kuke build` with `parent snapshot does not exist` or `content digest ... not found` (issue #904).
- Adds a shared `consts.KukebuildBaseDir` constant (mirrors `defaultBuildRoot` in `cmd/kukebuild/main.go`; the two cannot be a single source of truth because `cmd/kukebuild` is a separate Go module — cross-references added on both sides).
- Adds per-namespace `BuildCachePurgeOutcome` rows and a trailing `build cache <base>: removed | kept (cousin instance caches present)` row to the uninstall report. Base dir is rmdir'd only when empty after the per-namespace sweep, so a scoped uninstall (`--server-configuration`) leaves cousin instances' caches under sibling subdirs intact.
- Updates `uninstallLongDesc` and the half-cleaned-host remediation in `docs/cli-use-cases.md` so operators recovering a partial uninstall know to wipe the matching `/var/lib/kukebuild/<namespace>/` (the partial-cleanup path is the operator's; the follow-up `kuke uninstall` only sweeps caches it successfully purges in *its* pass).

The cache reclaim is intentionally **not** gated on `CleanupSkipped`: a namespace that was actually removed in a partial uninstall should still have its cache reclaimed even when a sibling realm's purge failed. A realm whose namespace survived contributes no row so its cache stays on disk for the follow-up recovery to reason about.

Closes #904

## Test plan

- [x] `make test` — full root + `cmd/kukebuild` module suites pass.
- [x] `go vet ./...` on both modules.
- [x] New `internal/controller` tests cover: reclaim path (per-namespace dirs removed + base rmdir'd when empty), gated path (namespace-survive → cache survives → base not rmdir'd), absent-base idempotency, and the empty-`BuildCacheBaseDir` → `consts.KukebuildBaseDir` default fallback.
- [x] New `cmd/kuke/uninstall` renderer tests cover: three-row "removed" output, "kept (cousin instance caches present)" on a scoped uninstall, and quiet no-row output on a never-built host.
- [ ] **Not run on this CI:** `make dev-init` smoke. The change is uninstall-only — it does not touch the build path, the daemon serve path, or the runtime container state — so the `kuke get realms` daemon-parity guard at the tail of `make dev-init` is unaffected. The full operator round-trip (`kuke uninstall` → `kuke init` → `kuke build`) is what exercises the fix end-to-end and is the natural smoke for the next operator-driven dev loop.

### DIFF
--- a/cmd/kuke/uninstall/uninstall.go
+++ b/cmd/kuke/uninstall/uninstall.go
@@ -70,7 +70,14 @@ This is the global counterpart to "kuke purge" (which is per-resource). It:
      a Restart=on-failure unit cannot respawn kukeond mid-uninstall.
   2. Purges every realm with --cascade (drains spaces, stacks, cells,
      containers and their containerd tasks/containers, and deletes the
-     containerd namespaces created by kukeon).
+     containerd namespaces created by kukeon). For every realm whose
+     namespace was actually removed, also reclaims the matching per-
+     namespace BuildKit cache directory at /var/lib/kukebuild/<namespace>/
+     (the cache references containerd by snapshot ID and content digest, so
+     leaving it behind strands the next "kuke build" with "parent snapshot
+     does not exist" or "content digest ... not found"). After the sweep,
+     /var/lib/kukebuild/ itself is rmdir'd if empty — cousin instances'
+     caches under sibling subdirs are left in place.
   3. Removes /run/kukeon/ recursively.
   4. Removes the configured run path (default /opt/kukeon) recursively.
   5. Removes the kukeon system user and group (no-op if absent).
@@ -81,7 +88,9 @@ mounts on disk would strand the next "kuke init" with stale containerd state.
 The report flags every dir/account row as "skipped (realm purge failed)" so
 the half-cleaned host is visible without scrolling to the trailing error.
 Resolve the realm-purge failure (often: stop the live daemon, remove the
-residual containerd namespace by hand) and re-run the command.
+residual containerd namespace by hand, wipe the matching
+/var/lib/kukebuild/<namespace>/ left behind by the partial pass) and re-run
+the command.
 
 The /usr/local/bin/kuke binary and the kukeond symlink are NOT removed —
 uninstalling runtime state is not the same as uninstalling the binary.
@@ -268,6 +277,7 @@ func printReport(cmd *cobra.Command, report controller.UninstallReport) {
 			fmt.Fprintf(out, "  - realm %q (namespace %q): purged\n", r.Name, r.Namespace)
 		}
 	}
+	renderBuildCaches(out, report)
 	if report.CleanupSkipped {
 		fmt.Fprintf(out, "  - %s: %s\n", report.SocketDir, outcomeSkipped)
 		fmt.Fprintf(out, "  - %s: %s\n", report.RunPath, outcomeSkipped)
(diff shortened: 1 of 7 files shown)


eminwux:
**PR #906 Review — `fix(uninstall): reclaim per-namespace BuildKit cache to prevent stale snapshot/content references`**

Targeted fix for issue #904. The gating logic — per-namespace reclaim runs only on `NamespaceRemoved=true`, base rmdir uses plain `os.Remove` so a scoped uninstall preserves cousin caches, and the cache sweep is intentionally ungated on the half-cleaned-host flag — matches the PR description, and the test coverage proves each invariant. Dual-module constant sync (`consts.KukebuildBaseDir` ↔ `cmd/kukebuild/main.go defaultBuildRoot`) gets explicit cross-references on both sides. Commit signed, CodeQL green, dead-code cross-check clean (every new exported symbol has a reader, both helpers have callers).

A few small items, none blocking:

- **`TestUninstall_BuildCacheBaseDefaultsToConst` exercises the host filesystem** `internal/controller/uninstall_test.go:791-810`: leaving `BuildCacheBaseDir` empty means the controller's default-fallback runs `rmdirIfEmpty(\"/var/lib/kukebuild\")` against the real host path. The realm-purge stub returns `NamespaceRemoved=false` so no per-namespace reclaim fires, but the base rmdir runs unconditionally after the realm loop. On a CI runner that happens to have an empty `/var/lib/kukebuild`, this test would actually rmdir it. The test comment acknowledges the trade-off (\"Pointing at a host path is undesirable in tests\"). Non-blocking — the assertion shape is right and rmdir is plain (not RemoveAll) so blast radius is small, but a future test could thread the default through a hook (e.g. an internal package-level var set under `t.Cleanup`) so the production-default path stays observable without poking the host.
- **Defensive `case c.Existed:` branch in `renderBuildCaches`** `cmd/kuke/uninstall/uninstall.go:60-65`: per `reclaimNamespaceBuildCache`'s contract (`Removed=true` or `Err!=nil`, no other combination), this branch is genuinely unreachable on the success path. The inline comment acknowledges it's defensive against regressions — fine to keep, but it skirts the \"no defensive code for impossible scenarios\" guidance. Cosmetic.
- **BSD-aliased ENOTEMPTY=EEXIST check in `rmdirIfEmpty`** `internal/controller/uninstall.go:505`: on Linux `syscall.ENOTEMPTY` (39) and `syscall.EEXIST` (17) are distinct constants, and this is Linux-only code (containerd path). The double-check is harmless but the comment overstates the cross-platform risk. Cosmetic.

Test plan is well-reasoned — the `make dev-init` carve-out is correct (this PR is uninstall-only, no daemon/build-path changes, the daemon-parity smoke at the tail would not exercise the cache reclaim). Renderer + controller unit coverage spans the four state branches (reclaim, gate-on-namespace-survive, absent-base, default-fallback) plus the scoped-uninstall \"kept (cousin instance caches present)\" path.

LGTM.

PR #907, opened 2026-05-26
fix(del): surface cell subdir residue and os.RemoveAll errors on stack/cell delete

### LINKED ISSUE #905: controller: del stack --cascade can claim success while leaving cell metadata dirs on disk
## Summary

`kuke del stack <name> --cascade` reported success, but the stack's metadata
directory was still on disk afterwards with leftover cell subdirectories
inside it. `kuke get stacks` correctly showed the stack as gone, but the
stack dir under `/opt/kukeon/data/<realm>/<space>/<stack>/` had to be
`rm -rf`'d by hand before `kuke create stack` with the same name would
succeed.

The user invoked it as `kuke del stack default --cascade`.

## Repro

1. Create a stack with one or more cells in it.
2. `kuke del stack <name> --cascade`.
3. `kuke get stacks` no longer lists the stack — cascade reported success
   (exit 0).
4. Inspect `/opt/kukeon/data/<realm>/<space>/<stack>/` on disk: the
   directory still exists and contains cell subdirectories.
5. Re-creating the stack with the same name fails until the directory is
   manually removed.

## Expected

Cascade delete should either:

- fully remove the stack directory and all child cell dirs from disk, or
- fail loudly with a non-zero exit and a clear error if any cleanup step
  (cell teardown, cgroup removal, dir removal) does not complete, so the
  operator does not discover residue by hand.

## Actual

The stack's `kukeon.yaml` is removed (so the stack disappears from
listings) but the stack directory and its cell children remain on disk
while the CLI reports success.

## Suspect code paths (root cause unconfirmed — two candidates that can
combine)

Two places interact to produce this; the exact trigger for this incident
wasn't captured at debug log level, so the cause is recorded as a
candidate shortlist rather than a single root cause. Either path warrants
the same fix shape (surface the error instead of swallowing it).

**1. `os.RemoveAll` error swallowed in `runner.DeleteStack`** —
`internal/controller/runner/delete_stack.go:131`:

```go
metadataRunPath := fs.StackMetadataDir(...)
_ = os.RemoveAll(metadataRunPath)
return nil
```

The stack's own `kukeon.yaml` deletion error IS checked just above (line
120), but the subsequent recursive cleanup of the stack directory is
best-effort — any failure (busy file, EACCES on a child, residual bind
mount inside a cell subdir) is discarded and `DeleteStack` returns nil.
The same `_ = os.RemoveAll(...)` pattern exists in
`internal/controller/runner/delete_cell.go:226`.

**2. `ListCells` silently skips cells with unreadable metadata** —
`internal/controller/runner/get.go` around line 437:

```go
cell, readErr := r.readCellInternal(metadataPath)
if readErr != nil {
    r.logger.DebugContext(r.ctx, "skipping cell metadata file",
        "path", metadataPath, "error", readErr)
    continue
}
```

The cascade loop in `internal/controller/delete_stack.go:110-118`
iterates exactly the slice returned by `ListCells`. A cell directory
that exists on disk but whose `kukeon.yaml` is missing/corrupt
(interrupted create, partial earlier teardown, etc.) is invisible to
cascade — its containers/cgroups are never torn down by the per-cell
path, and only the swallowed `os.RemoveAll` above would have removed
the directory itself.

These together permit: cascade sees 0 (or fewer than actual) cells,
deletes the stack's `kukeon.yaml`, then quietly fails to fully remove
the directory tree, and exits 0.

## Suggested fix sketch

- Surface `os.RemoveAll` errors from `runner.DeleteStack` and
  `runner.DeleteCell` instead of `_ =`-ing them. Wrap them under
  `errdefs.ErrDeleteStack` / `errdefs.ErrDeleteCell` so the CLI exits
  non-zero with a clear message naming the path that couldn't be
  removed.
- In cascade mode, before declaring the stack deleted, re-read the
  stack metadata directory and refuse to delete the stack's
  `kukeon.yaml` if cell subdirectories remain that cascade did not just
  handle — closes the `ListCells`-skipped-an-unreadable-cell gap
  regardless of which container/cgroup state those orphans hold.
- Optionally, treat an unreadable cell directory as a hard error in
  cascade mode (rather than silently skipping it), perhaps gated on
  whether the operator passed `--force`.

## Workaround

```
sudo rm -rf /opt/kukeon/data/<realm>/<space>/<stack>
```

after the apparently-successful cascade. Then `kuke create stack ...`
with the same name succeeds.

## Notes

- A deterministic repro recipe wasn't captured here; the user noticed
  residue after their normal teardown sequence. Reproducing on demand
  would benefit from running with `kukeond` at debug log level so the
  `skipping cell metadata file` debug line (path #2 above) is visible,
  and from observing whether `os.RemoveAll` is hitting an actual error
  (path #1 above). Recording either trigger is not a prerequisite for
  the fix — both paths warrant the loud-failure change.

eminwux:
## Summary

- `kuke del stack <name> --cascade` could exit 0 while leaving the stack metadata
  dir on disk with surviving cell subdirectories — the operator then had to
  `rm -rf /opt/kukeon/data/<realm>/<space>/<stack>/` by hand before
  `kuke create stack` with the same name would succeed. Both contributor paths
  named in the issue are addressed.
- `runner.DeleteStack` now refuses to remove the stack metadata file when a
  cell-shaped subdirectory survives under the stack metadata dir. ListCells
  silently skips cells with unreadable/missing `metadata.json` (`get.go:437`),
  so the cascade loop in `controller.deleteStackCascade` could complete with 0
  handled cells while a partial cell subdir remained. The new pre-check scans
  for entries that aren't `metadata.json` / `.lock` / `secrets` / `blueprints`
  / `configs`, refuses with an `ErrDeleteStack`-wrapped error naming the
  residue, and leaves both the stack metadata and the residue in place so a
  re-run after manual repair completes cleanly.
- `runner.DeleteStack` and `runner.DeleteCell` no longer swallow
  `os.RemoveAll` errors via the `_ = os.RemoveAll(...)` pattern. Failures are
  wrapped under `errdefs.ErrDeleteStack` / `errdefs.ErrDeleteCell` so the CLI
  exits non-zero with a message naming the path that could not be removed.
- Three regression tests cover the residue refusal (named cell subdir
  preserved, error wraps `ErrDeleteStack`), the scope-subdir allowlist
  (`secrets`/`blueprints`/`configs` do not trigger refusal), and the happy
  path (stack dir fully removed on success — direct guard against the
  reported symptom).

Scope discipline: `purge_stack.go` and `delete_space.go` carry the same
`_ = os.RemoveAll(...)` pattern but were out of the issue's named scope and
are left for a follow-up.

Closes #905

## Test plan

- [x] `make test-root` — all packages green, including the new
      `TestDeleteStack_*` tests in `internal/controller/runner/`.
- [x] `make test-kukebuild` — green.
- [x] `GOWORK=off go build ./...` and `go vet ./...` — clean.
- [ ] `make dev-init` smoke (the daemon-parity guard) — not exercised; this
      change is to the runner's delete paths and does not touch the daemon
      build, the bootstrap, or anything `kukeond` reads at startup.

### DIFF
--- a/internal/controller/runner/delete_cell.go
+++ b/internal/controller/runner/delete_cell.go
@@ -215,7 +215,13 @@ func (r *Exec) deleteCellLocked(cell intmodel.Cell) error {
 		return fmt.Errorf("%w: failed to delete cell metadata: %w", errdefs.ErrDeleteCell, err)
 	}
 
-	// Remove metadata directory completely
+	// Remove metadata directory completely. metadata.DeleteMetadata above
+	// removes the dir when empty (only the stack's own metadata.json + lock
+	// were there), but the cell dir also owns peer artifacts — etc-hosts,
+	// etc-hostname, attachable socket directories — that the metadata
+	// sweeper doesn't reach. Surface any failure so the half-deleted state
+	// (kukeon.yaml gone, cell dir surviving) reported in issue #905 cannot
+	// be silently swallowed.
 	metadataRunPath := fs.CellMetadataDir(
 		r.opts.RunPath,
 		internalCell.Spec.RealmName,
@@ -223,7 +229,9 @@ func (r *Exec) deleteCellLocked(cell intmodel.Cell) error {
 		cellStackName,
 		internalCell.Metadata.Name,
 	)
-	_ = os.RemoveAll(metadataRunPath)
+	if err = os.RemoveAll(metadataRunPath); err != nil {
+		return fmt.Errorf("%w: failed to remove cell directory %s: %w", errdefs.ErrDeleteCell, metadataRunPath, err)
+	}
 
 	return nil
 }
--- a/internal/controller/runner/delete_stack.go
+++ b/internal/controller/runner/delete_stack.go
@@ -20,8 +20,10 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 
+	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/ctr"
 	"github.com/eminwux/kukeon/internal/errdefs"
 	"github.com/eminwux/kukeon/internal/metadata"
@@ -110,6 +112,32 @@ func (r *Exec) DeleteStack(stack intmodel.Stack) error {
 		// Continue with metadata deletion
 	}
 
+	// Refuse to remove the stack while cell subdirectories survive under the
+	// stack metadata dir. ListCells silently skips a cell whose metadata.json
+	// is missing or unreadable (controller/runner/get.go), so the cascade
+	// loop in controller.deleteStackCascade can complete with zero handled
+	// cells while a partial cell subdir remains on disk. Removing the
+	// stack's own metadata here would orphan that subdir behind a stack the
+	// operator can no longer list — the failure surfaced as residue under
+	// /opt/kukeon/data/<realm>/<space>/<stack>/ after `kuke del stack
+	// --cascade` exited 0. Issue #905.
+	metadataRunPath := fs.StackMetadataDir(
+		r.opts.RunPath,
+		internalStack.Spec.RealmName,
+		internalStack.Spec.SpaceName,
+		internalStack.Metadata.Name,
+	)
+	residue, residueErr := scanStackDirCellResidue(metadataRunPath)
+	if residueErr != nil {
+		return fmt.Errorf("%w: scan stack dir %q: %w", errdefs.ErrDeleteStack, metadataRunPath, residueErr)
+	}
+	if len(residue) > 0 {
+		return fmt.Errorf(
+			"%w: cell subdirectory residue under %s — cascade did not remove %s; inspect and remove by hand before re-running",
+			errdefs.ErrDeleteStack, metadataRunPath, strings.Join(residue, ", "),
+		)
+	}
+
 	// Delete stack metadata file
 	metadataFilePath := fs.StackMetadataPath(
 		r.opts.RunPath,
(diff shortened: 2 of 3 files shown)


eminwux:
## Review

LGTM. The fix correctly addresses the root cause described in #905: `scanStackDirCellResidue` runs **before** `metadata.DeleteMetadata`, so the refusal is atomic — the stack's own `metadata.json` and the residue both survive on disk for operator inspection (locked in by `TestDeleteStack_RefusesOnCellSubdirResidue`). The residue scan is intentionally stricter than `ListCells` (it flags any non-allowlisted entry including files, while `ListCells` only iterates `entry.IsDir()` and silently skips unreadable metadata at `internal/controller/runner/get.go:437` — exactly the swallow path #905 hit). The allowlist matches the constants in `internal/consts/consts.go` (`KukeonMetadataFile`, `KukeonSecretsSubdir`, `KukeonBlueprintsSubdir`, `KukeonConfigsSubdir`) and `metadata.LockFileSuffix`, and `TestDeleteStack_AllowsKnownScopeSubdirs` pins that stack-scoped scope subdirs don't trigger the refusal. The `os.RemoveAll` error surfacing on both stack and cell paths replaces the `_ = os.RemoveAll(...)` silent swallow with `errdefs.ErrDeleteStack` / `errdefs.ErrDeleteCell` wraps, so the CLI now exits non-zero on the failure path that originally let #905 ship as exit 0. Single signed commit, license headers present, test file colocated with source.

The `purge_stack.go` / `delete_space.go` carve-out is called out in the PR body — agreed that's a separate follow-up; the issue's named scope is satisfied.

### Non-blocking nits (documentation only)

1. **`internal/controller/runner/delete_cell.go:218-224`** — the comment says *"metadata.DeleteMetadata above removes the dir when empty (only the **stack's own** metadata.json + lock were there)"* but this is `delete_cell.go`, so the dir under discussion is the cell metadata dir and the items removed are the cell's own metadata.json + lock. Reads as a copy-paste slip from the `delete_stack.go` neighbor.

2. **`internal/controller/runner/delete_cell.go:223` and `internal/controller/runner/delete_stack.go:154`** — both comments reference *"kukeon.yaml gone"*. There is no `kukeon.yaml` anywhere in the codebase (grep is empty); the metadata file is `metadata.json` (`consts.KukeonMetadataFile`). Likely a holdover from an older naming. Replacing with `metadata.json gone` matches the actual on-disk symptom and keeps a future maintainer from grepping for a file that doesn't exist.

### Test plan check

PR-reported: `make test-root`, `make test-kukebuild`, `go build ./...`, `go vet ./...` all green. `make dev-init` smoke skipped with a documented justification (no daemon, bootstrap, or `/opt/kukeon` reader path touched) — matches CLAUDE.md's "If your change touches anything the daemon reads, this check must be in the PR's test plan" gate; this PR doesn't, so the carve-out is appropriate.

PR #911, opened 2026-05-26
fix(attachable): unblock kuke run when host/image kukeon GID drift + reuse-path tty bind-mount chown is missed

eminwux:
## Summary

`kuke run kukeon-pm` (and any cell with an attachable work container that
runs as a non-root image USER) fails inside the startup grace window with
`cell wound down immediately after start: ... container "work" exited within
startup grace window (task status=stopped)`. No output is captured: the
container's `tty/` dir under `/opt/kukeon/data/.../work/tty/` is empty —
not even `kuketty.log` — because kuketty exits before `openTerminalLogger`
opens it.

The wound-down container is kuketty failing `net.Listen("unix", "/run/kukeon/tty/socket")`
with EACCES on the kukeon-group-owned tty bind-mount. Two interlocking
causes:

1. **Host vs. image `kukeon` group GID drift.** `kuke init`'s
   `groupadd --system kukeon` lets the host pick a GID (989 here);
   `crew/images/crew-base-user/Dockerfile` hardcodes `groupadd -g 988
   kukeon`. The daemon creates the tty bind-mount `root:kukeon (989)`
   mode 02750, but `oci.WithUser("claude")` resolves the workload's
   AdditionalGids from the image's /etc/group, giving the process
   `additionalGids: [1000, 988]`. The bind-mount has no world bits, so
   the in-container process can't traverse it.

2. **`attachablePostCreateChown` skipped on the reuse-existing-child
   start path.** `attachableBuildOpts` chmods + chowns the tty dir back
   to `root:kukeon` on every call (it has to, in case the prior boot left
   it loose). `attachablePostCreateChown` then resets the owner to the
   container's resolved uid so the workload can write the socket /
   capture / log siblings. But the chown call was nested under
   `if !reuseExistingChild` in both StartCell (start.go) and
   StartContainer — so reuse-path starts left the dir root-owned even
   when the work container ran as uid 1000.

## Fix

- `ctr.WithKukeonGroupGID(gid)` BuildOption + SpecOpts appending the
  host's kukeon GID to `Process.User.AdditionalGids` (deduped, zero-GID
  is a no-op). Wired through `daemonDefaultBuildOpts` from
  `r.opts.KukeonGroupGID`, so the host GID flows into every container's
  process user regardless of how the image numbered (or omitted) its
  own kukeon group.
- Lift `attachablePostCreateChown` out of the `if !reuseExistingChild`
  block in StartCell and StartContainer. The chown is idempotent and
  the container exists on the reuse path by definition, so the call is
  safe on both paths.
- Unit tests in `internal/ctr/spec_test.go` pin the AdditionalGids
  contract: appended-on-miss, deduped-on-match, zero-is-no-op,
  mirrored for `BuildRootContainerSpec`.

## Test plan

- [x] `go test ./internal/ctr/ ./internal/controller/runner/` — green
  (70s + 0.9s).
- [x] `make dev-init` — bootstraps clean; daemon-parity tail prints
  both realms Ready.
- [x] `kuke delete cell kukeon-pm` + `kuke run kukeon-pm` —
  cell transitions to Ready (was `Failed` / `StartCellFailed` before);
  `tty/` dir is `inwx:kukeon 02750` with `socket`, `kuketty.log`,
  `capture`, `metadata.json` all present.
- [x] `sudo ctr -n default.kukeon.io container info default_default_kukeon-pm_work`
  reports `user: {'uid': 1000, 'gid': 1000, 'additionalGids': [1000, 988, 989]}`
  — image's 988 stays, host's 989 is appended.
- [ ] Reviewer: confirm the chown lift is safe on the recreate path
  where `attachableBuildOpts` was already side-effecting the dir
  before `CreateContainerFromSpec` — it is, because the chown reads
  the resolved uid out of containerd's OCI spec which is now persisted.

### DIFF
--- a/internal/controller/runner/runner.go
+++ b/internal/controller/runner/runner.go
@@ -329,6 +329,11 @@ func (r *Exec) daemonDefaultBuildOpts() []ctr.BuildOption {
 	if r.opts.RunPath != "" {
 		opts = append(opts, ctr.WithSecretRunPath(r.opts.RunPath))
 	}
+	if r.opts.KukeonGroupGID > 0 {
+		//nolint:gosec // KukeonGroupGID is a real GID (uint16 territory); the
+		// runtime-spec field is uint32 so the conversion is widening.
+		opts = append(opts, ctr.WithKukeonGroupGID(uint32(r.opts.KukeonGroupGID)))
+	}
 	return opts
 }
 
--- a/internal/controller/runner/start.go
+++ b/internal/controller/runner/start.go
@@ -1117,18 +1117,26 @@ func (r *Exec) startCellLocked(cell intmodel.Cell) (_ intmodel.Cell, retErr erro
 				"created container",
 				fields...,
 			)
-
-			if err = r.attachablePostCreateChown(namespace, containerSpec); err != nil {
-				return intmodel.Cell{}, fmt.Errorf(
-					"failed to chown attachable tty dir for %s: %w", ctrContainerID, err,
-				)
-			}
 		}
 		// On the reuse path the BuildOptions attachOpts returned aren't
 		// applied (the existing OCI spec already carries the attachable
 		// wrapping); the call is still made above for its side effects —
 		// kuketty binary staging, per-container metadata render, ttyDir
 		// chmod — which must run before the task starts.
+		//
+		// The post-create chown also runs on both paths: attachableBuildOpts
+		// resets ttyDir to root:kukeon mode 02750 every call, so without a
+		// matching chown-to-container-uid here the in-container process
+		// (claude:1000) cannot create the kuketty socket/log inside the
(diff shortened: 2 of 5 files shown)


eminwux:
**PR #911 Review — `fix(attachable): unblock kuke run when host/image kukeon GID drift + reuse-path tty bind-mount chown is missed`**

LGTM — the host-GID injection (`WithKukeonGroupGID` ordered after `securitySpecOpts` so `oci.WithUser` runs first; deduped append; zero-no-op; widening uint16-territory `int → uint32`) and the `attachablePostCreateChown` lift in both `startCellLocked` and `StartContainer` cleanly close the EACCES reuse-path window without touching the provision-side paths (both already pair the chown with the create unconditionally). Tests pin the AdditionalGids contract for both `BuildContainerSpec` and `BuildRootContainerSpec`; single commit is signed/verified.

PR #913, opened 2026-05-27
deps: bump sbsh to v0.12.1 + e2e regression for stopped→run EACCES race

### LINKED ISSUE #361: terminal: bind-then-chmod race in OpenSocketCtrl briefly exposes 0o600 socket — group attach hits EACCES

### LINKED ISSUE #912: deps+test: bump sbsh once sbsh#361 lands (closes kuke run StartCell→attach EACCES race) and add e2e coverage
## Context

Operator-side repro: a non-root user in the `kukeon` group running
`kuke run <stopped-cell>` intermittently fails with

```
Error: error attaching client: ping failed: dial unix /opt/kukeon/s/<id>: connect: permission denied
```

The summary line in the same output prints `containers: started` (not
`not started`), confirming the failure only happens on the
`Stopped → StartCell → attach` branch in `cmd/kuke/run/run.go:runExistingCell`.
Already-Ready cells (no-op branch) never race.

Root cause is upstream in sbsh: `OpenSocketCtrl` binds the control
socket with `lc.Listen(...)` (mode `0o666 & ~umask`, effectively
`srw-------` under the daemon's `0o077` umask) and only then calls
`applySocketPerms(socketMode=0o660, socketGID=kukeon)`. Between bind and
chmod the inode accepts `connect(2)` but is owner-only, so kukeon-group
operators dialing inside that window hit EACCES.

`kuke run` lands the dial in that window because `StartCell` returns
after the 300ms post-start liveness probe (containerd reports the work
task `Running`), which is satisfied before sbsh's intra-container chmod
syscall has landed.

PR #911 (host/image kukeon GID drift + reuse-path bind-mount chown) fixed
the adjacent EACCES that prevented kuketty from binding the socket at
all. It does **not** close the listen/chmod window.

Tracked upstream: **https://github.com/eminwux/sbsh/issues/361**

## Proposal

Two-step landing, gated on the upstream fix:

1. **Bump `github.com/eminwux/sbsh`** from the current `v0.12.0` to whatever
   release ships the sbsh#361 fix (likely v0.12.1 / v0.13.0). Refresh
   `go.sum` via `go mod tidy`. No public-API churn expected — the
   suggested upstream fix is internal to `OpenSocketCtrl`.

2. **Add an e2e regression test** in `e2e/e2e_kuke_attach_test.go` (peer
   to `TestKuke_AttachDetach_KeepsTaskRunning`) that exercises the exact
   path that races today:

   - Create an attachable cell as the kukeon daemon (root).
   - `kuke stop cell <name>` to drop it into `Stopped`.
   - `kuke run <name>` as a **non-root** test user that is a member of
     the `kukeon` group. The harness already runs the daemon as root
     and has the group plumbed — see `harness_daemon_test.go`.
   - Assert: exit 0, no `permission denied` on stderr, the summary
     prints `containers: started`, and a follow-up read on the work
     container's tty/socket inode shows `0o660 root:kukeon`.
   - Run the test in a tight loop (≥20 iterations) to catch the race
     even if the window narrows in future sbsh changes.

   A pure unit test on the kukeon side isn't meaningful — the race is
   inside the in-container kuketty process. The e2e bar is what locks
   the fix down.

## Acceptance criteria

- [ ] `go.mod` pins `github.com/eminwux/sbsh` at the release that ships
      eminwux/sbsh#361.
- [ ] `go.sum` refreshed via `go mod tidy`; `go build ./...` and
      `go test ./...` green.
- [ ] `e2e/e2e_kuke_attach_test.go` grows a `TestKuke_RunFromStopped_NonRootKukeonGroup_NoEACCES`
      (or similarly named) test that exercises the
      `Stopped → kuke run → attach` path as a kukeon-group non-root user
      and loops ≥20 iterations.
- [ ] `make dev-init` smoke test passes after the bump.
- [ ] PR description references this issue and eminwux/sbsh#361 so the
      upstream→downstream chain is searchable.

## Out of scope

- Adding a CLI-side EACCES retry in `cmd/kuke/run/attach.go`. The race
  is fixable upstream and the band-aid would mask the real bug; do it
  here only if the upstream fix slips for an extended window.
- Capture/log file perms (sbsh#200 territory).

## Notes

- Block on the upstream landing — don't bump a half-fix.

eminwux:
## Summary

Closes #912.

- Bumps `github.com/eminwux/sbsh` from `v0.12.0` to `v0.12.1`. The upstream release ships eminwux/sbsh#362 (closes eminwux/sbsh#361), which closes the bind-then-chmod EACCES window in `OpenSocketCtrl` by setting the umask around `net.ListenConfig.Listen` so the socket inode is born at the configured mode rather than `0o666 & ~daemonUmask`. With the kukeon per-container TTY dir already `2750` `root:kukeon`, the setgid bit propagates the gid to the bound inode at bind time, so the dial as a kukeon-group non-root caller sees `0o660 root:kukeon` from the first packet — no follow-up chown race. `go.sum` refreshed via `go mod tidy`.

- Adds `TestKuke_RunFromStopped_NonRootKukeonGroup_NoEACCES` in `e2e/e2e_kuke_attach_test.go` (peer of `TestKuke_AttachDetach_KeepsTaskRunning`). The test exercises the exact path that raced under v0.12.0 — a kukeon-group non-root caller invoking `kuke run` against a Stopped cell — and loops 20 iterations to catch the window even after the upstream fix narrowed it. Each iteration drops the cell into Stopped, launches `kuke run` via a real PTY with `SysProcAttr.Credential` swapping the child to `nobody:kukeon`, waits a 1s attach grace, then sends `Ctrl+]Ctrl+]` and asserts exit 0, no `permission denied` substring on the captured output, and the `containers: started` summary line. A final stat on the work container's symlink confirms steady-state mode `0o660 root:kukeon`. The test skips cleanly when run as non-root, when the host has no `kukeon` group (i.e. `kuke init` has not run), or when no `nobody` user is present.

### Divergence from issue body

The issue described the harness as already having "the group plumbed — see `harness_daemon_test.go`". It did not: `startKukeondDaemon` defaulted to `socketModeRootOnly` because no `--socket-gid` was passed, and `startPTY` did not accept a `*syscall.Credential`. Both gaps are filled here in two new helpers in `e2e_kuke_attach_test.go`:

- `startKukeondDaemonWithSocketGID(t, runPath, gid)` mirrors `startKukeondDaemon` but adds `--socket-gid` so the daemon listener inode lands at `0o660 :kukeon`. The per-test SUN_PATH-safe sockDir is also chmoded `0o755` so the unprivileged caller can traverse to the socket basename.
- `startPTYWithCredential(t, env, cred, command, args...)` mirrors `startPTY` but sets `cmd.SysProcAttr.Credential` so the PTY child exec's as `uid=nobody`, `gid=nobody`, `supplementary={kukeon}`. `creack/pty.Start` preserves pre-set `SysProcAttr` fields, so the credential survives the fork; the PTY master/slave pair is opened by the root test process and inherited as the child's stdio, side-stepping the slave-open DAC check.

The test also relaxes `runPath` and `<runPath>/s/` to `0o755` after the daemon creates them so the unprivileged caller can traverse the path chain to the symlink — the per-test `MkdirTemp` defaults to `0o700`, which would block `nobody` before the EACCES race could surface.

### Upstream→downstream chain

- Upstream issue: eminwux/sbsh#361 (closed)
- Upstream fix: eminwux/sbsh#362 (Option A: `runtime.LockOSThread` + `syscall.Umask` around `Listen`)
- Upstream release: eminwux/sbsh `v0.12.1`
- Downstream bump (this PR): `go.mod` `v0.12.0 → v0.12.1`
- Downstream regression test (this PR): `TestKuke_RunFromStopped_NonRootKukeonGroup_NoEACCES`

### Out of scope (per issue)

- No CLI-side EACCES retry in `cmd/kuke/run/attach.go` — the race is fixed at the bind site upstream; a retry would mask the real bug.

## Test plan

- [x] `GOWORK=off go build ./...` — clean
- [x] `GOWORK=off go test -c -o /dev/null ./e2e/...` — e2e package compiles
- [x] `make test` — full unit suite passes (test-root + test-kukebuild)
- [x] `make dev-init` smoke — daemon-parity tail shows identical `default` / `kuke-system` Ready rows, attach smoke + nested-mode plumbing checks all green
- [ ] `make e2e` with kukeon group + nobody user present — runs the new `TestKuke_RunFromStopped_NonRootKukeonGroup_NoEACCES` 20-iteration loop. The agent dev environment cannot exercise this (the e2e harness needs a built kukeon image plus `kuke init`-provisioned host); the test is written to skip cleanly outside that environment, so the e2e job on a CI runner with the standard host provisioning will be the first run of the new test.

### DIFF
--- a/e2e/e2e_kuke_attach_test.go
+++ b/e2e/e2e_kuke_attach_test.go
@@ -17,13 +17,19 @@
 package e2e_test
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
+	"os/user"
 	"path/filepath"
+	"strconv"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
+
+	"github.com/creack/pty"
 
 	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/util/fs"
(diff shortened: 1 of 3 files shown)


eminwux:
## Review

LGTM. The dep bump (`go.mod`/`go.sum`) is a one-liner to `eminwux/sbsh v0.12.1`, which closes the upstream `OpenSocketCtrl` bind-then-chmod EACCES window (eminwux/sbsh#361 → #362) — the umask wrap around `net.ListenConfig.Listen` lands the socket inode at the configured mode from birth, and the existing kukeon per-container TTY dir setgid (`2750 root:kukeon`) propagates the gid to bound inodes, so a kukeon-group non-root dialer sees `0o660 root:kukeon` from the first packet. The downstream regression test (`TestKuke_RunFromStopped_NonRootKukeonGroup_NoEACCES` at `e2e/e2e_kuke_attach_test.go:497`) exercises the exact path that raced under v0.12.0 — kukeon-group non-root caller invoking `kuke run` against a Stopped cell — and the 20-iteration loop is the right shape to catch a future re-narrowing of the window. The two new harness helpers (`startKukeondDaemonWithSocketGID` at line 681 and `startPTYWithCredential` at line 797) cleanly fill the gaps the issue body assumed were already in place (`--socket-gid` not threaded through `startKukeondDaemon`; `startPTY` accepting no `*syscall.Credential`), and `creack/pty.Start`'s preservation of pre-set `SysProcAttr` is correctly relied on for the privilege drop. The runPath / `<runPath>/s/` chmod-0o755 dance for unprivileged path traversal is justified (per-test `MkdirTemp` defaults to `0o700`, which would ENOENT-out before the EACCES window could surface). The final stat against `socketPath` (lines 645–668) pins `0o660 root:kukeon` as steady-state, which is the regression backstop if the iterations ever happen to race favourably. Single signed commit; license header carried over from the existing test file.

### Non-blocking nits

1. **`e2e/e2e_kuke_attach_test.go:455` — `runFromStoppedAttachGrace = 1 * time.Second` is half the peer test's `attachConnectGrace = 2 * time.Second` (line 47).** The peer's docstring justifies its 2s on "the bytes can race pkg/attach's raw-mode setup and land before its filter is wired up, in which case the detach sequence is forwarded verbatim instead of triggering a clean exit." The new test's docstring justifies 1s on EACCES-window timing, which is about the dial, not about the filter being wired. Both grace concerns apply here — the test does both a dial (EACCES check) AND a detach keystroke (filter race check). If 1s is genuinely tight on a loaded CI runner, the failure mode is the keystroke being echoed verbatim and the 20s exit timeout firing on `permission denied`-free output — which the assertions would attribute to a hang rather than to the filter race, making the failure harder to read. CI is green now, so this is speculative — but either reusing `attachConnectGrace` or extending the docstring to address the filter-wire race would future-proof the test.

2. **`startKukeondDaemonWithSocketGID` and `startPTYWithCredential` duplicate ~70 + ~30 lines each from their non-credential cousins.** The PR docstring explicitly justifies this ("a single-purpose helper is easier to read at the test site") and I agree the current shape reads better than a functional-options-laden parent. Flagging only as a forward-looking observation: a third privilege-drop scenario landing in this file would tip the trade-off toward extracting the common skeleton.

### Test plan check

PR-reported: `go build ./...`, `go vet ./...`, `make test`, `make dev-init` smoke all green (the CLAUDE.md `make dev-init` gate applies because the sbsh bump affects the daemon's listener bind path; the author ran it ✓). `make e2e` is the unchecked box, with a documented justification (the agent dev env can't host-provision the `nobody`/`kukeon` preconditions). CI's `make e2e` job is green, which is the first real run of the new test — confirms the assertions pass on the standard provisioning. Acceptable.

PR #917, opened 2026-05-27
fix(dev-init): sudo the reset gate + recreate kukeond on image digest drift (#915)

### LINKED ISSUE #915: dev-init: "already existed" stale-image trap survives #857 — reset gate is permission-blocked AND kuke init does not recreate cell on image change
## Repro

On the current `main` (commit `64c03e8`, today 2026-05-27), with a previously-bootstrapped host (i.e. `make dev-init` has been run at least once successfully, so `/opt/kukeon/data/kuke-system/...` exists and `/opt/kukeon` is owned `root:kukeon` mode `0o2750` after the post-init permission fix), and as a user **not** in the `kukeon` group, run:

```bash
make dev-init
```

The script fails at the post-#857 verifier:

```
==> Verify running kukeond container is anchored on the just-built image
running kukeond container is anchored on a stale image (#857)
  running container snapshot chain: sha256:700a21b4fa4c2…
  just-built image chain:           sha256:23bbde593a4b0…
  manifest sha256:1e7ca876c71bf… did not propagate to the kukeond cell —
  `kuke daemon reset` + `kuke init --kukeond-image` did not refresh the container.
  Try `sudo ./kuke daemon reset --purge-system` then re-run `make dev-init`.
```

## Observed

Two distinct defects are at play — both still present after #857 was closed. The verifier #857 added correctly catches the stale-image symptom but neither of the underlying causes was fixed.

### Defect 1 — `scripts/dev-init.sh:286` reset gate is permission-blocked

```bash
if [ -d "${KUKEOND_CELL_DIR}" ]; then
    step "Reset prior kukeond cell"
    sudo --preserve-env="${PRESERVE_ENV_ADMIN}" ./kuke daemon reset \
        "${SERVER_CONFIG_FLAGS[@]}"
else
    step "No prior kukeond cell at ${KUKEOND_CELL_DIR}; skipping reset"
fi
```

`KUKEOND_CELL_DIR=/opt/kukeon/data/kuke-system/kukeon/kukeon/kukeond`. The `[ -d ... ]` test is run as the unprivileged invoking user, **not** under `sudo`. Every adjacent step (`kuke daemon reset`, `kuke init`, `kuke build`) shells out via `sudo --preserve-env=…`, so the test is the only un-sudo'd filesystem touch in this block.

After the first successful `kuke init`, the post-init permission fix sets `/opt/kukeon` to `drwxr-s--- root:kukeon` mode `0o2750` (see the init output line `"/opt/kukeon" (recursive): chown root:kukeon mode 0o2750 (dirs) / 0o0640 (files)`). Any invoking user not in the `kukeon` group can no longer traverse the tree, so `[ -d "${KUKEOND_CELL_DIR}" ]` returns false even though the directory exists. Confirmed on this host:

```
$ id
uid=1000(inwx) gid=1000(inwx) groups=…,27(sudo),… (no kukeon group)
$ ls -ld /opt/kukeon /opt/kukeon/data
ls: cannot access '/opt/kukeon/data': Permission denied
drwxr-s--- 5 root kukeon 4096 May 26 17:53 /opt/kukeon
$ [ -d /opt/kukeon/data/kuke-system/kukeon/kukeon/kukeond ] && echo y || echo n
n
$ sudo test -d /opt/kukeon/data/kuke-system/kukeon/kukeon/kukeond && echo y
y
```

So the gate takes the else branch and prints `==> No prior kukeond cell at …; skipping reset`, even though `pgrep -a containerd` shows the live shims:

```
1262 containerd-shim-runc-v2 -namespace kuke-system.kukeon.io -id kukeon_kukeon_kukeond_root  …
1304 containerd-shim-runc-v2 -namespace kuke-system.kukeon.io -id kukeon_kukeon_kukeond_kukeond …
```

Consequence: on every `make dev-init` run after the first, the `kuke daemon reset` step is silently skipped for any user not in the `kukeon` group. The first-pass `kuke init` block (line 280) also reports `cell "kukeond": already existed` / `cell containers: started`, and the second-pass init (line 305) does the same — the running container is never replaced.

### Defect 2 — `kuke init --kukeond-image` keeps the existing cell anchored to the old image

This is the un-fixed candidate root cause #2 from the body of #857:

> **`kuke init --kukeond-image` is idempotent against the running cell** and treats an already-Ready kukeond as up-to-date even when `--kukeond-image` resolves to a digest the running container does not have.

#857 closed after only the layer-1 fix landed (the post-init snapshot-chain verifier at `scripts/dev-init.sh:332-384`). The layer-2 fix — "init's existing-cell path must compare image digest and recreate on mismatch" — never landed. The empirical evidence is in the second init's output from the failing run above:

```
- cell containers: started
```

`started` means "already running," not "replaced with new image" — the same wording #857 called out as the divergence signal. Even when the reset path *does* fire (i.e. the user is in the `kukeon` group, so Defect 1 doesn't bite), `kuke daemon reset` leaves enough state behind that the subsequent `kuke init` takes the existing-cell branch and never compares the requested `--kukeond-image` against the running container's snapshot chain. The only working recovery is `kuke daemon reset --purge-system`, which wipes `/opt/kukeon/data/kuke-system` and forces init's create branch.

Defects 1 and 2 compound: Defect 1 is the user-facing entrypoint (silently skips reset on a hardened-perms host), Defect 2 is the contract gap that means even the non-skipped path doesn't refresh the container. Closing only one of them leaves `make dev-init` broken for a non-trivial fraction of contributors.

## Root cause

**Defect 1:** `scripts/dev-init.sh:286` uses unprivileged `[ -d … ]` to gate a sudo'd operation, but every directory it can target is locked behind `root:kukeon` `0o2750` after the first init. The same pattern is **not** present at line 272 (`if [ ! -d "${SYSTEM_REALM_DIR}" ]`), but that one only governs the first-time-bootstrap branch, and a fresh host has `/opt/kukeon` absent so the traversal isn't blocked. The bug is structural to line 286 specifically.

**Defect 2:** Init's bootstrap-cell path (presumably under `internal/controller/runner/` or `cmd/kuke/init/`) does not compare the requested `--kukeond-image` digest against the running container's image when the cell record is present, so the "already existed" path never triggers a recreate-on-image-change.

## Suggested fix

Two-layer fix matching the two defects:

1. **`scripts/dev-init.sh:286` — sudo the gate.** Smallest possible change: `if sudo test -d "${KUKEOND_CELL_DIR}"; then` (or hoist into a helper for symmetry with the `SYSTEM_REALM_DIR` probe at line 272 if that one also needs hardening — it doesn't bite today because the first-pass init creates the dir before the perm fix runs, but it's the same anti-pattern). This is a single-line change with no behavioral surprise: every other access to `${KUKEOND_CELL_DIR}` in the file is already sudo'd.

2. **`kuke init` existing-cell path — recreate on image mismatch.** When the cell record is present and the running container's image digest differs from the resolved `--kukeond-image` digest, init should either (a) recreate the container in place, or (b) fail-loud with a "image digest mismatch — run `kuke daemon reset --purge-system`" instead of silently logging `started`. (a) is the contract `make dev-init` documents ("re-running on a healthy host produces a clean re-bootstrap"); (b) at minimum stops masking the bug. The verifier #857 added already encodes the correct check (snapshot-chain comparison via `ctr snapshots info` + `ctr content get`), so most of the logic for (b) can be lifted from `scripts/dev-init.sh:332-384` directly into the init path.

The [Init/bootstrap self-sufficiency gaps are bugs](../../) feedback applies — `make dev-init` is documented as the canonical re-bootstrap loop, and any flow where "re-running on a healthy host" silently produces a broken state (Defect 1) or a half-refreshed daemon (Defect 2) is a bug, not an operator-skill issue.

## Notes

- Closes the gap left by #857: that issue's layer-1 fix (verifier) shipped; layer-2 (init image-mismatch recreate) did not. This issue tracks both layer-2 *and* the new Defect 1 surface that wasn't visible in #857's repro environment.
- Reproduces on commit `64c03e8` (today's `main` tip).
- Workaround until fixed: `sudo ./kuke daemon reset --purge-system` then re-run `make dev-init`. The verifier message already documents this.

eminwux:
## Summary

Two defects exposed in #915 — both leave \`make dev-init\` shipping a stale daemon container after a re-run.

**Defect 1 — \`scripts/dev-init.sh\` reset gate is permission-blocked.** The \`[ -d \"\${KUKEOND_CELL_DIR}\" ]\` test ran unprivileged, but after the first \`kuke init\` the post-init permission fix sets \`/opt/kukeon\` to \`root:kukeon 0o2750\`. Any contributor not in the \`kukeon\` group can no longer traverse the tree, the test returns false, and every adjacent \`sudo --preserve-env=... ./kuke daemon reset\` call is silently skipped. Switched to \`sudo test -d\` to match every other access in the script. The sibling \`[ ! -d \"\${SYSTEM_REALM_DIR}\" ]\` probe at line 272 is the same anti-pattern but doesn't bite today because the first-pass init is idempotent on a pre-bootstrapped host — left alone per the issue's lower-blast-radius default.

**Defect 2 — \`bootstrapCell\` only compared image *ref strings*.** When \`kuke build\` re-points the same tag (\`kukeon-local:dev\`) at a new manifest, the persisted \`ContainerSpec.Image\` string still matches the desired \`--kukeond-image\` value. The existing #868 drift check returned equal, the existing-cell \`EnsureCell+StartCell\` path ran, and the daemon container kept its stale anchored snapshot. Extended \`bootstrapCell\`'s drift detection: when the ref strings match *and* the root container exists, query containerd for (a) the chainID the image at the ref would unpack to today and (b) the chainID the container's snapshot is anchored on. If they differ the existing \`RecreateCell\` path runs — the verifier #857 added (snapshot-chain comparison at \`scripts/dev-init.sh:332-384\`) already encoded the equivalent check, but only post-init; this lifts the same signal into \`kuke init\` so the swap happens before the verifier fires. Probe failures (containerd hiccup, image absent, snapshot key missing) fall through with a logged WARN so init is never gated on a transient inspect failure.

The two-layer fix matches the issue's suggested-fix structure. \`CellRecreatedDueToImageDrift\` is reused for both ref-drift and digest-drift; \`printCellActions\` now branches on \`CellPriorImage == image\` to render \"image digest drift on \<ref\>\" instead of the misleading \"\<ref\> → \<ref\>\".

New ctr/runner surface: \`ImageChainID(namespace, ref)\` returns the chainID via \`image.RootFS\` + \`identity.ChainID\`; \`ContainerRootChainID(namespace, containerID)\` returns the snapshot's Parent via \`container.Info\` + \`snapshotter.Stat\`. Five \`ctr.Client\` fakes in the runner test package got the same null stubs.

## Test plan

- [x] \`make test-root\` — all packages pass, including the four new \`TestBootstrapCell_Digest*\` / \`TestBootstrapCell_MatchingDigest*\` cases and the two new \`TestPrintCellActions\` subtests.
- [x] \`make test-kukebuild\` — passes.
- [x] \`pre-commit run --files \$(git diff --name-only HEAD)\` — go-fmt, go-mod-tidy, golangci-lint all green.
- [x] \`make dev-init\` end-to-end on this host (user \`inwx\` not in \`kukeon\` group, \`/opt/kukeon\` 0o2750). The reset gate fired (\`cell \"kukeond\": created\` instead of \"already existed\"), kukeond came up at \`unix:///run/kukeon/kukeond.sock\`, the post-init snapshot-chain verifier matched, and the daemon-parity tail printed identical \`-o wide\` / \`--no-daemon\` output. The downstream \`kuke attach\` smoke (\`cattach\`) fails on this host with an unrelated \`parent snapshot ... not found\` against \`registry.eminwux.com/busybox:latest\` — pre-existing environmental issue, not regressed by this change.
- [x] Reproduced defect 1 conditions: unprivileged \`[ -d /opt/kukeon/data/kuke-system/kukeon/kukeon/kukeond ]\` → false; \`sudo test -d ...\` → true. Confirmed the fix takes the reset branch.

Closes #915

### DIFF
--- a/cmd/kuke/init/init.go
+++ b/cmd/kuke/init/init.go
@@ -740,11 +740,24 @@ func printCellActions(cmd *cobra.Command, section controller.CellSection, image 
 		// Image drift on a surviving cell record (issue #868). Surfacing the
 		// prior → desired transition is what tells the operator the rebuild
 		// actually picked up their `kuke build` output instead of silently
-		// restarting the stale container.
-		cmd.Println(fmt.Sprintf(
-			"    - cell %q: recreated (image drift: %s → %s)",
-			section.CellName, section.CellPriorImage, image,
-		))
+		// restarting the stale container. When the prior and desired refs
+		// match byte-for-byte but bootstrapCell still routed through
+		// RecreateCell, the trigger is a content digest swap on the same
+		// ref (issue #915 defect 2: `kuke build` re-pointed the same tag
+		// at fresh layers). The ref-string transition would render as
+		// "kukeon-local:dev → kukeon-local:dev" which says nothing; the
+		// digest-drift label is the readable form.
+		if section.CellPriorImage == image {
+			cmd.Println(fmt.Sprintf(
+				"    - cell %q: recreated (image digest drift on %s)",
+				section.CellName, image,
+			))
+		} else {
+			cmd.Println(fmt.Sprintf(
+				"    - cell %q: recreated (image drift: %s → %s)",
+				section.CellName, section.CellPriorImage, image,
+			))
+		}
 	case section.CellCreated:
 		cmd.Println(fmt.Sprintf("    - cell %q: created (image %s)", section.CellName, image))
 	default:
--- a/cmd/kuke/init/init_test.go
+++ b/cmd/kuke/init/init_test.go
@@ -500,6 +500,58 @@ func TestPrintCellActions(t *testing.T) {
 			section:  controller.CellSection{},
 			expected: []string{"- cell: not provisioned"},
 		},
+		{
+			// Ref-string drift renders "prior → desired" — the existing
+			// #868 path. The transition is informative because the two
+			// strings differ.
+			name: "recreated-image-ref-drift",
+			section: controller.CellSection{
+				CellName:                     "kukeond",
+				CellMetadataExistsPre:        true,
+				CellMetadataExistsPost:       true,
+				CellCgroupExistsPre:          true,
+				CellCgroupExistsPost:         true,
+				CellRootContainerExistsPre:   true,
+				CellRootContainerExistsPost:  true,
+				CellRootContainerCreated:     true,
+				CellStartedPre:               true,
+				CellStartedPost:              true,
+				CellRecreatedDueToImageDrift: true,
+				CellPriorImage:               "ghcr.io/eminwux/kukeon:v0.9.0",
+			},
+			image: "ghcr.io/eminwux/kukeon:v1.0.0",
+			expected: []string{
+				"- cell \"kukeond\": recreated (image drift: ghcr.io/eminwux/kukeon:v0.9.0 → ghcr.io/eminwux/kukeon:v1.0.0)",
+				"- cell root container: created",
+			},
+		},
+		{
+			// Digest drift on the same ref: prior image string equals
+			// desired so "prior → desired" would render the same string
+			// twice — useless to the operator. The dedicated "image
+			// digest drift on <ref>" message is the readable form for
+			// the #915 defect-2 path.
+			name: "recreated-image-digest-drift",
+			section: controller.CellSection{
+				CellName:                     "kukeond",
+				CellMetadataExistsPre:        true,
+				CellMetadataExistsPost:       true,
+				CellCgroupExistsPre:          true,
+				CellCgroupExistsPost:         true,
+				CellRootContainerExistsPre:   true,
+				CellRootContainerExistsPost:  true,
+				CellRootContainerCreated:     true,
+				CellStartedPre:               true,
+				CellStartedPost:              true,
+				CellRecreatedDueToImageDrift: true,
+				CellPriorImage:               "docker.io/library/kukeon-local:dev",
+			},
+			image: "docker.io/library/kukeon-local:dev",
+			expected: []string{
+				"- cell \"kukeond\": recreated (image digest drift on docker.io/library/kukeon-local:dev)",
+				"- cell root container: created",
+			},
+		},
 	}
 
 	for _, tc := range testCases {
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/creack/pty v1.1.24
 	github.com/eminwux/sbsh v0.12.1
 	github.com/opencontainers/go-digest v1.0.0
+	github.com/opencontainers/image-spec v1.1.1
 	github.com/opencontainers/runtime-spec v1.2.1
 	github.com/spf13/cobra v1.10.1
 	github.com/spf13/viper v1.21.0
@@ -56,7 +57,6 @@ require (
 	github.com/moby/sys/userns v0.1.0 // indirect
 	github.com/onsi/ginkgo/v2 v2.25.1 // indirect
 	github.com/onsi/gomega v1.38.1 // indirect
-	github.com/opencontainers/image-spec v1.1.1 // indirect
 	github.com/opencontainers/selinux v1.13.0 // indirect
 	github.com/pelletier/go-toml/v2 v2.2.4 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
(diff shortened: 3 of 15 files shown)


eminwux:
**PR #917 Review — `fix(dev-init): sudo the reset gate + recreate kukeond on image digest drift (#915)`**

LGTM — two-layer drift detection (ref-string + chainID) with a documented soft-fail fallback on probe error, four-branch test coverage (`Digest`/`Matching`/`ProbeFailure`/`RootAbsent`), and a `sudo test -d` gate that matches every other access pattern in `dev-init.sh`. The deferred sibling fix at `scripts/dev-init.sh:272` is acknowledged in the comment block and the smaller blast radius is the right scope call.

Issue #635, opened 2026-05-20
tty: add TtyStage.runOn field + create-stage executor routing (phase A)


eminwux:
## Context

`containers[].tty.onInit[]` stages forward to sbsh's `Stages.OnInit` (`internal/controller/runner/attachable.go`, via `ttyStagesToExecSteps` -> `WithOnInit`) and run **in the wrapped shell on every boot** — i.e. every `kuke start`, since the daemon re-renders the kuketty metadata per start. There is no per-stage timing control today.

That makes non-idempotent setup unusable as `onInit`: a one-time `npm ci`, DB seed, or bootstrap re-runs on every restart of a long-lived (stopped/started) cell. Operators have no declarative way to say "run this once when the cell is created, not on every start."

Persistence already supports run-once: `kuke stop` -> `kuke start` reuses the same container + snapshot (`internal/ctr/container.go:60-116`); only `kuke delete` cleans the snapshot. So a create-time effect survives a restart — what's missing is a way to *not re-run the step* on subsequent boots. This phase lays the field + routing + executor foundation; the run-once render gate itself lands in phase C (#690).

## Proposal

Add a per-stage `runOn` knob to `TtyStage` (the struct was built for exactly this — see `pkg/api/model/v1beta1/container.go:193-197`, "future stage knobs ... runOn"):

```go
type TtyStage struct {
    Script string
    RunOn  string // "" / "start" -> today's behavior (every boot); "create" -> run-once per cell instance
}
```

Routing by `runOn`:

- **`start`** (default, absent): unchanged — mapped to sbsh `Stages.OnInit`, runs in-shell every boot. Byte-for-byte identical to today.
- **`create`**: routed into kuketty's **pre-Serve executor** (the slot phase 1a / #617 introduced for repo clone, `cmd/kuketty/repos.go`) — runs to completion as a separate step, **not** handed to sbsh. Failure policy follows phase-1a's required-failure contract (a failed required `create` stage exits before `Serve`).

This phase adds the `ContainerStatus.Stages` **schema only** — exactly as #617 added the `RepoStatus` schema that #642 later populated. Reporting outcomes through the RPC channel and the run-once render gate are deferred (see Out-of-scope follow-ups).

## Acceptance criteria

- [ ] `TtyStage.RunOn` field + json/yaml tags + v1beta1<->internal conversion; empty/`start`/`create` accepted, others rejected at apply time (`validateContainerTty`).
[...]

Issue #617, opened 2026-05-19
phase 1a: add containers[].repos[] with kuketty pre-Serve clone


eminwux:
## Context

Phase 1a of #423, building on **phase 0 (#641)**. Crew templates duplicate the same `if [ ! -d $DIR/.git ]; then git clone ...; fi` block in every `onInit` script, and the resulting clone state is invisible to `kuke get`. Kukeon has no first-class concept of a repo a container depends on.

Today `ContainerSpec` carries `Volumes`, `Secrets`, `Env`, `WorkingDir`, but no repos; setup runs in `onInit` shell scripts with no daemon-observable status. Phase 0 makes kuketty read the kukeon-native `ContainerDoc` and build the `TerminalSpec` itself, so `repos[]` can be a plain `ContainerSpec` field kuketty reads directly — **no sidecar doc**.

## Proposal

Add `repos[]` to `ContainerSpec`, next to `Volumes` and `Secrets`:

```yaml
spec:
  containers:
    - id: work
      image: ${REGISTRY}/claude:latest
      repos:
        - name: project
          target: /home/claude/${PROJECT_DIR}
          branch: main
          url: ${PROJECT_REPO}     # scalar-param substitution (slot fill is phase 4 / #423)
          required: false          # default; true => fatal pre-Serve failure
```

### Execution — kuketty pre-Serve, clone-if-absent

kuketty grows a **pre-Serve step** (run before `sbshserver.Serve`) that reads `repos[]` straight from the mounted `ContainerDoc.Spec` (phase 0):

1. **clone-if-absent**: if `target/.git` exists, `git fetch` + ff-only (or skip); else `git clone url` into `target` on `branch`. Uses the container's `~/.ssh/`, `~/.gitconfig`, `GIT_SSH_COMMAND` — the same identity `onInit` scripts use today. Because the writable rootfs persists across `kuke stop`/`start` (`internal/ctr/container.go:60-116`; only `kuke delete` cleans the snapshot, `:489`), a restart never re-clones — the artifact is its own idempotency check.
2. `required: false` (default) -> a clone/fetch failure does not block startup. `required: true` -> kuketty exits non-zero **before** `sbshserver.Serve`, so the daemon observes the task as Failed and the container goes to `Failed` with a clear reason. No mystery stderr buried in attach output.

### Out of this issue

[...]

PR #918, opened 2026-05-27
fix(kuketty): apply socket mode+GID at bind to close pre-Serve EACCES window

### LINKED ISSUE #916: kuketty: control socket perms deferred until sbshserver.Serve — kuke run EACCES
## Context

Operator-side repro: a non-root user in the `kukeon` group runs `kuke run <Stopped-cell>` and gets

```
- containers: started
Error: error attaching client: ping failed: dial unix /opt/kukeon/s/<id>: connect: permission denied
```

This is **distinct from #912 / sbsh#361** (closed by #913 / sbsh v0.12.1). #912 closed the listen-vs-`applySocketPerms` window inside sbsh's `OpenSocketCtrl`. This issue is a separate, kuketty-side deferral that sbsh v0.12.1 does not cover — kuketty does not go through `OpenSocketCtrl`.

## Root cause

`cmd/kuketty/main.go` `run()` order:

```
127  claimSocketListener     ← bare net.Listen("unix", path), no umask/chmod
144  processRepos            ← #617 — git clones, slow or fails on bad DNS
158  processStages           ← #635 — runOn:create stages
168  sbshserver.New
172  srv.Serve               ← bringUp → UseListener → applySocketPerms (chmod 0o660 + chown :kukeon)
```

`claimSocketListener` (`cmd/kuketty/main.go:259`) is just `os.Remove` + `net.Listen` — no umask manipulation, no chmod, no chown. Linux `bind(2)` for `AF_UNIX` creates the inode at `0o777 & ~umask`. The container's root umask is `0o022`, so the socket is born `srwxr-xr-x` — group bits `r-x`, no `w`. `connect(2)` requires write on the socket inode for the matching class, so a `kukeon`-group operator hits EACCES.

The kukeon group ownership is **not** the bug — the per-container `tty/` dir is SGID (`drwxr-s---` 2750 root:kukeon, see `attachableTTYDirInitialPerms`), so the socket inherits group=kukeon from the directory. The mode is what's wrong.

The chmod-to-0o660 + chown :kukeon lives in sbsh's `UseListener` → `applySocketPerms` (`sbsh@v0.12.1/internal/terminal/terminalrunner/terminal_runner_exec.go:241`, `sockets.go:106`), which only fires from `sbshserver.Serve` → `bringUp`. By then `processRepos` and `processStages` have already run inside kuketty.

`kuke run`'s daemon-side `StartCell` returns "containers: started" as soon as containerd marks the workload task `Running` — i.e. while kuketty is still in `claimSocketListener` / `processRepos` / `processStages`, well before `sbshserver.Serve`. The host client dials the `/opt/kukeon/s/<id>` symlink immediately and loses to the deferred chmod.

## Repro (dev-init container, no DNS)

The agent dev container has no DNS to GitHub. With the kukeon-dev-root cell declaring two `required=true` repos, each `git clone` blocks ~20s on `Could not resolve hostname github.com`. After both repos fail, kuketty exits non-zero from `processRepos` (required=true) and the container task dies — `sbshserver.New` is **never reached** so `applySocketPerms` is **never called**. The socket stays at `0o755` for its entire lifetime.

Sampling the socket mode while `kuke run kukeon-dev-root` is blocked:

```
t=1s   socket mode = 755
t=3s   socket mode = 755
t=8s   socket mode = 755
t=15s  socket mode = 755
t=25s  socket mode = 755
t=45s  socket mode = 755
t=65s  socket mode = (absent)   ← kuketty exited, bind-mount dirent gone
```

kuketty.log shows two `repo resolution failed` WARNs ~20s apart, then the process exits. EACCES is 100% reproducible in this configuration.

On a host where DNS resolves and clones succeed, this manifests as an intermittent race over a shorter window (clone duration + stage duration) rather than a 100% failure — but still racy by the same mechanism.

## Why sbsh v0.12.1 does not cover this

sbsh v0.12.1 closed the bind→chmod window inside `OpenSocketCtrl` via `listenUnixWithMode` (`runtime.LockOSThread` + temporary `unix.Umask(^mode & 0o777)` around `lc.Listen`). kuketty does **not** call `OpenSocketCtrl` — it pre-binds via `claimSocketListener` and hands the listener to sbsh via `UseListener`. sbsh applies perms in `UseListener`, but only when called from `sbshserver.Serve` → `bringUp`. Both branches are individually correct now, but kuketty's pre-Serve work between `claimSocketListener` and `sbshserver.New(...).Serve(...)` reopens the same EACCES window for the full duration of `processRepos` + `processStages` — and indefinitely if either of those fails.

## Fix sketch

Apply mode + GID to the socket inode inside `claimSocketListener`, immediately after `net.Listen`, using the `socketMode` and `socketGID` already settled on `spec` by `buildTerminalSpec` (`cmd/kuketty/spec.go:129-138`). Mirror the recipe sbsh exposes in `sbsh@v0.12.1/internal/terminal/terminalrunner/sockets.go`:

1. Bind under a temporary umask of `^mode & 0o777` (with `runtime.LockOSThread`) so the inode is born at the configured mode and there is no window at all.
2. Belt-and-braces `os.Chmod(path, mode)` + `os.Chown(path, -1, gid)` immediately after.

sbsh's later `UseListener` → `applySocketPerms` becomes an idempotent re-chmod, which is harmless and consistent with the doc comment on `UseListener` ("Spec.SocketMode and Spec.SocketGID are applied to the inode resolved from ln.Addr() so the on-disk permissions match what the OpenSocketCtrl path produces").

This closes:
- the deferred-chmod race after "containers: started" → host dial,
- the never-recovers case where kuketty exits in `processRepos`/`processStages` before `sbshserver.Serve` runs.

## Test plan

- [ ] Unit test in `cmd/kuketty/` that calls the (refactored) socket-claim helper and asserts the on-disk socket mode is `0o660` and group is the configured GID immediately after return — before any sbsh involvement.
- [ ] e2e: extend `e2e/e2e_kuke_attach_test.go`'s `runFromStopped*` cases with a variant where the cell declares a `required=true` repo with an unresolvable hostname. `kuke run` must EACCES at the dial today and pass after the fix.
- [ ] `make dev-init` smoke unchanged.

eminwux:
## Summary

- `cmd/kuketty/main.go` `claimSocketListener` now applies the configured `SocketMode` + `SocketGID` at bind time (umask-controlled `net.Listen` plus belt-and-braces `os.Chmod` / `os.Chown`), so the per-container kuketty control socket is born at the operator-visible mode and group instead of deferring those perms to `sbshserver.Serve → bringUp → applySocketPerms` later in the boot path.
- Closes the EACCES window a kukeon-group non-root operator hit when `kuke run`'s daemon-side `StartCell` returned "containers: started" while kuketty was still inside the pre-Serve work — `processRepos` (#617) or `processStages` (#635) — and the host client dialed the socket before sbsh's chmod landed. With a `required=true` repo that fails DNS, kuketty exits in `processRepos` and `sbshserver.New` is never reached, so the pre-fix code left the socket at `0o755` for its entire lifetime; with healthy DNS it was the same race but over a shorter window. Distinct from #912 / sbsh#361 (closed by #913 / sbsh v0.12.1) — that fix lives inside sbsh's `OpenSocketCtrl`, which kuketty does not use (kuketty pre-binds and hands the listener to sbsh via `UseListener`).
- Mirrors sbsh's `listenUnixWithMode` recipe (`runtime.LockOSThread` + process-wide `listenerUmaskMu` around a temporary `unix.Umask(^mode & 0o777)`), so the late `applySocketPerms` inside `Serve` becomes an idempotent re-chmod — both paths land at the same on-disk perms.

Closes #916

## Test plan

- [x] `cmd/kuketty/` unit tests — `TestClaimSocketListener_AppliesModeAndGID` asserts `0o660` + configured GID on the inode immediately after the helper returns (before any sbsh involvement); `TestClaimSocketListener_NilGIDLeavesGroupUnchanged` covers the no-kukeon-group fallback (`0o600` owner-only); `TestClaimSocketListener_BindsAndUnlinksStale` was updated for the new signature and continues to lock the unlink-then-listen sequence.
- [x] `make test-root` — `go test ./...` minus `/e2e` — green.
- [x] `make dev-init` smoke — clean end-to-end after a `--purge-system` reset (a prior local run had accumulated stale `dev-init-attach.kukeon.io` containerd snapshot state from repeated `dev-init` invocations, unrelated to this change — removing the namespace via `ctr namespaces remove` and re-running produces a clean pass, including the kuketty-wrapped `kuke attach` smoke which exercises this fix in the real OCI path).
- [x] `go vet` on the changed packages (`./cmd/kuketty/...`, `./e2e/...`) — clean.
- [ ] `go vet ./...` repo-wide — pre-existing failure in `github.com/containerd/cgroups/v2@v2.0.0-20221109034041` (third-party `pids.Limit` type mismatch) reproduces on `origin/main` HEAD without this change applied; not introduced here.
- [x] e2e — new `TestKuke_RunFromStopped_RequiredRepoUnresolvable_NoEACCES` covers the AC test plan: a cell fixture (`attachable-cell-required-repo-unresolvable.yaml`) declares a `required=true` repo with an unresolvable hostname (`.invalid` per RFC 6761), `applyAttachableCell` auto-starts the workload, and the test stats the per-container socket the moment it appears (inside the pre-Serve window, *before* sbsh's `Serve` would have run `applySocketPerms`) and asserts mode `0o660` + group=kukeon. The test compiles cleanly (`go test -run NoTestMatchesAtAll ./e2e/...`); the full e2e suite was not driven locally since it needs the `kukeon-local:e2e` docker image (`make test-e2e` runs the full bring-up).

### DIFF
--- a/cmd/kuke/apply/testdata/attachable-cell-required-repo-unresolvable.yaml
+++ b/cmd/kuke/apply/testdata/attachable-cell-required-repo-unresolvable.yaml
@@ -1,0 +1,33 @@
+apiVersion: v1beta1
+kind: Cell
+metadata:
+  name: test-cell
+spec:
+  id: test-cell
+  realmId: test-realm
+  spaceId: test-space
+  stackId: test-stack
+  containers:
+    - id: root
+      root: true
+      image: registry.eminwux.com/busybox:latest
+      command: sleep
+      args:
+        - "3600"
+    - id: work
+      image: registry.eminwux.com/busybox:latest
+      command: sleep
+      args:
+        - "3600"
+      attachable: true
+      # Required repo with an unresolvable hostname (RFC 6761 reserves
+      # .invalid for tests). kuketty's processRepos will fail and the
+      # workload task will go to Failed — but the per-container kuketty
+      # control socket is born inside claimSocketListener *before*
+      # processRepos runs. Issue #916 locks that the socket lands at
+      # 0o660 :kukeon at bind time, not deferred to sbshserver.Serve.
+      repos:
+        - name: unresolvable
+          target: /tmp/never-cloned
+          url: http://kukeon-test.invalid/never-exists.git
+          required: true
--- a/cmd/kuketty/main.go
+++ b/cmd/kuketty/main.go
@@ -40,11 +40,14 @@ import (
 	"net"
 	"os"
 	"os/signal"
+	"runtime"
+	"sync"
 	"syscall"
 
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 	sbshlogging "github.com/eminwux/sbsh/pkg/logging"
 	sbshserver "github.com/eminwux/sbsh/pkg/terminal/server"
+	"golang.org/x/sys/unix"
 )
 
 // defaultConfigPath is the fixed in-container path of the kukeond-rendered
@@ -124,7 +127,7 @@ func run(args []string) error {
 		return err
 	}
 
-	listener, err := claimSocketListener(spec.SocketFile)
+	listener, err := claimSocketListener(ctx, spec.SocketFile, spec.SocketMode, spec.SocketGID)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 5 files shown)


eminwux:
**PR #918 Review — `fix(kuketty): apply socket mode+GID at bind to close pre-Serve EACCES window`**

LGTM — closes the pre-Serve EACCES window by binding under a temporary umask + belt-and-braces chmod/chown, mirrors sbsh@v0.12.1's `listenUnixWithMode` recipe, and the new unit + e2e tests pin the on-disk shape (0o660 :kukeon) at the bind site so the post-Serve `applySocketPerms` is now an idempotent re-chmod.

PR #922, opened 2026-05-28
fix(reconcile): scope-narrow lineage Config lookup so realm-scoped Configs bound to stack-placed cells resolve

### LINKED ISSUE #921: fix(reconcile): scope-narrowing lookup for lineage Config finds realm-scoped configs bound to stack-placed cells
## Summary

`lookupLineageConfig` (daemon reconciler) and `materialiseFromConfig` (CLI `kuke restart cell`) both use the **cell's** realm/space/stack to look up its lineage Config. When a Config is bound at a shallower scope than the cell it materialized, the lookup misses the Config entirely.

## Root cause

`kuke run <config>` uses `ExplicitScope` for the Config lookup — space and stack are empty unless the operator sets them explicitly — so a plain `kuke run kukeon-dev-root-0` finds the Config at realm scope (`/opt/kukeon/data/default/configs/kukeon-dev-root-0`). `resolveCellLocation` then fills the session-default space/stack (`"default"/"default"`) onto the materialized cell before creating it.

The `kukeon.io/config` lineage label records only the Config **name**, not its scope. When the daemon reconciler or the restart CLI later does `GetConfig(realm, space="default", stack="default")`, it resolves to `/opt/kukeon/data/default/default/default/configs/kukeon-dev-root-0` — which does not exist — and treats the Config as deleted.

## Symptoms

- `kuke get cell <name>` shows `outOfSync: true`, `outOfSyncReason: lineage Config deleted` even though the Config still exists.
- `kuke restart cell <name>` prints:
  ```
  notice: cell "<name>" is OutOfSync but reconcile materialisation failed (config not found …); bouncing without reconcile
  ```
  and falls back to a plain stop+start, never reconciling from the Config.
- The OutOfSync state is permanent — every reconciler tick re-detects the "deleted" Config and re-sets the flag.

## Affected paths

| File | Function |
|------|----------|
| `internal/controller/reconcile_outofsync.go` | `lookupLineageConfig` |
| `cmd/kuke/restart/cell/cell.go` | `materialiseFromConfig` |

Both functions pass the cell's full scope coordinates (realm+space+stack) directly to `GetConfig`, instead of probing progressively shallower scopes.

## Fix

Scope-narrowing in both lookup sites: probe full scope → space-only → realm-only, use first hit. Skip duplicate probes when the cell's space or stack is already empty.

This mirrors the `kuke run` side: `loadFromConfig` already uses `ExplicitScope` (no space/stack unless explicit), so a realm-scoped Config is always found there. The reconciler and restart CLI need the same widening.

## Reproduction

```bash
# Config at realm scope, no space/stack
kuke get config kukeon-dev-root-0 --realm default          # found ✓
kuke get config kukeon-dev-root-0 --realm default --space default --stack default  # not found ✗

# Cell at realm=default / space=default / stack=default carries kukeon.io/config=kukeon-dev-root-0
kuke get cell kukeon-dev-root-0 --realm default --space default --stack default -o yaml \
  | grep -E 'outOfSync|outOfSyncReason'
# outOfSync: true
# outOfSyncReason: lineage Config deleted   ← wrong: Config exists, just at realm scope

kuke restart cell kukeon-dev-root-0
# notice: cell "kukeon-dev-root-0" is OutOfSync but reconcile materialisation failed
#         (config not found …); bouncing without reconcile
```

eminwux:
## Summary

- `lookupLineageConfig` (daemon reconciler) and `materialiseFromConfig` (`kuke restart cell` CLI) both used the cell's full `realm/space/stack` to resolve its `kukeon.io/config` lineage Config. The label records only the Config **name**, not its scope, so a `kuke run kukeon-dev-root-0` against a realm-scoped Config materialized a cell at `default/default/default` and every later lookup at full scope missed — the reconciler persisted a permanent `outOfSyncReason: lineage Config deleted` and `kuke restart cell <name>` printed `reconcile materialisation failed (config not found …)`, falling back to a bounce-only path.
- Both lookups now probe progressively shallower scopes — full → space-only → realm-only — and use the first hit, mirroring the `kuke run` side that already does this via `cmd/kuke/shared.ExplicitScope`. Probes are deduplicated when the cell's space or stack is already empty.

## Test plan

- [x] `make test` (full root + kukebuild test suite)
- [x] `make dev-init` (rebuild + re-bootstrap; daemon parity tail prints identical `default` / `kuke-system` rows; `kuke attach` smoke green)
- [x] New regression test `TestReconcileCells_OutOfSync_RealmScopedConfigFoundForStackPlacedCell` pins the daemon-side probe walk (full → space-only → realm-only) and the corresponding non-`lineage Config deleted` verdict
- [x] New regression test `TestReconcileCells_OutOfSync_RealmScopedCellSkipsRedundantProbes` pins the dedupe rule (no narrowing when the cell already sits at realm scope)
- [x] New regression test `TestRestartCell_RealmScopedConfigFoundForStackPlacedCell` pins the same probe walk + reconcile path on the CLI side

Closes #921

### DIFF
--- a/cmd/kuke/restart/cell/cell.go
+++ b/cmd/kuke/restart/cell/cell.go
@@ -319,23 +319,19 @@ func restartStopped(
 
 // materialiseFromConfig re-runs the CellConfig materialisation pipeline
 // against the cell's lineage Config: GetConfig + GetBlueprint +
-// cellconfig.Materialize. The cell's stored realm/space/stack supply the
-// lookup scope — a Config materialises a cell into the scope it lives in,
-// matching the daemon's reconciler (internal/controller/reconcile_outofsync.go).
+// cellconfig.Materialize. The cell's stored realm/space/stack name the
+// lookup's deepest scope, but the `kukeon.io/config` lineage label records
+// only the Config name — a `kuke run <config>` against a realm-scoped
+// Config materializes a cell at realm=default / space=default /
+// stack=default (resolveCellLocation fills the session defaults), so the
+// lookup must probe progressively shallower scopes (full → space-only →
+// realm-only) and use the first hit. Mirrors the daemon's reconciler
+// (internal/controller/reconcile_outofsync.go lookupLineageConfig).
 func materialiseFromConfig(
 	ctx context.Context, client kukeonv1.Client,
 	cellDoc v1beta1.CellDoc, configName string,
 ) (v1beta1.CellDoc, error) {
-	cfgRes, err := client.GetConfig(ctx, v1beta1.CellConfigDoc{
-		APIVersion: v1beta1.APIVersionV1Beta1,
-		Kind:       v1beta1.KindCellConfig,
-		Metadata: v1beta1.CellConfigMetadata{
-			Name:  configName,
-			Realm: cellDoc.Spec.RealmID,
-			Space: cellDoc.Spec.SpaceID,
-			Stack: cellDoc.Spec.StackID,
-		},
-	})
+	cfgRes, err := lookupLineageConfigCLI(ctx, client, cellDoc, configName)
 	if err != nil {
 		return v1beta1.CellDoc{}, err
 	}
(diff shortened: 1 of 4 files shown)


eminwux:
**PR #922 Review — `fix(reconcile): scope-narrow lineage Config lookup so realm-scoped Configs bound to stack-placed cells resolve`**

LGTM — both `lookupLineageConfig` (daemon) and `lookupLineageConfigCLI` (CLI) probe full → space-only → realm-only with correct dedupe and stop-on-real-error semantics, the three regression tests pin both the probe order and the dedupe rule, and the test plan covers `make test` + `make dev-init`.

PR #924, opened 2026-05-28
fix(run): scope-narrow Config lookup so kuke run <config> finds Configs at deeper scopes

### LINKED ISSUE #923: kuke run <config>: config not found when Config is stored at a deeper scope than realm-only
## Summary

`kuke run <config>` (and `kuke run <clone-name>`) fails with:

```
Error: config not found (config "<name>" in scope realm="default" space="" stack="")
```

when the target Config is stored at `space="default"` / `stack="default"` and the user does not pass `--space`/`--stack` explicitly.

## Root cause

`loadFromConfig` (`cmd/kuke/run/run.go`) builds its lookup scope using `ExplicitScope`, which intentionally returns `""` for space/stack when neither the flag nor the env var is set — to avoid hiding realm-scoped Configs. The single `client.GetConfig` call at that one scope is the whole lookup:

```go
lookup := v1beta1.CellConfigDoc{
    Metadata: v1beta1.CellConfigMetadata{
        Realm: kukshared.PickLookupRealm(...),   // "default"
        Space: kukshared.ExplicitScope(...),      // "" if flag/env not set
        Stack: kukshared.ExplicitScope(...),      // "" if flag/env not set
    },
}
cfgRes, err := client.GetConfig(cmd.Context(), lookup)
```

A Config stored at `space="default"` / `stack="default"` (e.g. a clone Config created from a source whose cell was placed there by `resolveCellLocation`) is invisible to this lookup.

## Reproduction

```bash
# Source Config lives at space="default"/stack="default"
kuke run kukeon-dev-root --space default --stack default  # creates cell, clone Config at that scope
kuke run kukeon-dev-root-0                               # fails: config not found (space="")
kuke run kukeon-dev-root-0 --space default --stack default  # works (workaround)
```

## Expected behaviour

`kuke run <config>` without `--space`/`--stack` finds the Config at whatever scope it is stored, mirroring the behaviour already established in:

- `materialiseFromConfig` (`cmd/kuke/restart/cell/cell.go`) — probes full→space-only→realm-only
- `lookupLineageConfig` (`internal/controller/reconcile_outofsync.go`) — same probe pattern

## Fix

Replace the single `client.GetConfig` call in `loadFromConfig` with the same three-probe loop used in `materialiseFromConfig`:

1. `(pickLocation space, pickLocation stack)` — viper-backed full scope (matches where `resolveCellLocation` would place the cell)
2. `(pickLocation space, "")` — space-only
3. `("", "")` — realm-only

`pickLocation` is already available in the same file and handles the flag→env→default fallback correctly. Using it (rather than `ExplicitScope`) for the starting probe scope is safe here because the loop — not `IsSet` — is what determines whether a Config is actually present.

## Workaround

Pass `--space default --stack default` (or whatever scope the Config lives in) explicitly.

eminwux:
## Summary
- `kuke run <config>` (and `kuke run <clone-name>`) now walks progressively shallower scopes — full (realm/space/stack) → space-only (realm/space) → realm-only (realm) — when resolving the daemon-stored CellConfig, mirroring the daemon reconciler (`internal/controller/reconcile_outofsync.go` `lookupLineageConfig`) and the CLI restart path (`cmd/kuke/restart/cell/cell.go` `lookupLineageConfigCLI`, landed in #922). The starting scope comes from `pickLocation` (viper-backed, defaults to `default`) instead of `ExplicitScope`, so a Config stored at `default/default/default` is found by a bare `kuke run <config>` without forcing `--space default --stack default`.
- Probes deduplicate when space/stack are already empty, and a real RPC error short-circuits the walk so the operator sees the underlying failure rather than a misleading realm-only not-found.
- The issue body cites `materialiseFromConfig` / `lookupLineageConfig` as already implementing this three-probe walk. That premise had rotted between issue filing and pick-up (those functions previously did single-scope lookups). #922 landed the walk in both before this PR, so the cited reference matches reality at merge time.

## Test plan
- [x] `make test-root` (existing run-package tests still pass; new tests cover all three probe positions, RPC short-circuit, and full-scope error message)
- [x] `make test-kukebuild`
- [x] `pre-commit run --files cmd/kuke/run/run.go cmd/kuke/run/run_test.go`
- [x] `make kuke` (binary builds; `./kuke run --help` still renders)
- [ ] `make dev-init` — not run; the change is CLI-only (no daemon-read or build-path edits), so the project CLAUDE.md's smoke gate ("change touches the build path, the daemon, or anything under `/opt/kukeon`") does not apply.

Closes #923

### DIFF
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -26,6 +26,7 @@ package run
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #924 Review — `fix(run): scope-narrow Config lookup so kuke run <config> finds Configs at deeper scopes`**

LGTM — three-probe walk in `lookupConfigWithFallback` matches the daemon-side `lookupLineageConfig` and CLI-restart `lookupLineageConfigCLI` precedents (same probe order, same dedup, same RPC short-circuit), tests pin all three probe positions plus the short-circuit, signed commit, dev-init carve-out is justified (CLI-only path).

PR #925, opened 2026-05-28
fix(dev-init): remove dev-init-attach namespace after purge to make attach-smoke idempotent

### LINKED ISSUE #919: dev-init: attach-smoke fails on re-run with "parent snapshot does not exist" — teardown_attach_smoke_state leaves cross-run snapshot lineage stranded
## Repro

Tree: \`origin/main\` (commit \`64c03e8\` at the time of writing) with a single-host
kukeon, containerd at \`/run/containerd/containerd.sock\`, the kukeon group
present, and the kukeond daemon already running healthily.

\`\`\`bash
# First run — clean baseline (kuke-system already provisioned, daemon healthy).
make dev-init               # exits 0; attach-smoke "kuke attach exited cleanly"

# Second run, no intervening operator action.
make dev-init               # exits 2; attach-smoke fails at `kuke apply`
\`\`\`

The second run reproduces 100% on this host. \`\`scripts/dev-init.sh\`\`'s
\`\`teardown_attach_smoke_state\`\` (l.435–443) successfully purges the four
\`\`dev-init-attach\`\` resources via \`\`kuke purge --cascade\`\`, the run re-creates
the realm/space/stack, and the cell-create step fails inside the
\`\`dev-init-attach.kukeon.io\`\` containerd namespace.

## Observed

Tail of the failing \`\`make dev-init\`\` run, lines from \`\`==> kuke attach
smoke against a kuketty-wrapped cell\`\` onward (script step
\`\`step \"kuke attach smoke against a kuketty-wrapped cell\"\`\`,
\`\`scripts/dev-init.sh:423\`\`):

\`\`\`
==> kuke attach smoke against a kuketty-wrapped cell
[ teardown purges the previous run's dev-init-attach realm — succeed ]
Realm \"dev-init-attach\" (namespace \"dev-init-attach.kukeon.io\")
  - metadata: created
  - containerd namespace: created
  - cgroup: created
Space \"ds\" ... created
Stack \"dks\" ... created
Cell \"cattach\": failed
  Error: failed to create cell: failed to create root container: failed to create container: parent snapshot sha256:7f74ca72855668453fc4207c1e5c50bb5653e72bb9939eb1d5512c001624e927 does not exist: not found
\`\`\`

Cross-namespace snapshot inspection right before the failing run:

\`\`\`
$ sudo ctr -n dev-init-attach.kukeon.io snapshots ls
KEY PARENT KIND
$ sudo ctr -n dev-init-attach.kukeon.io images ls
REF TYPE DIGEST SIZE PLATFORMS LABELS
$ sudo ctr -n default.kukeon.io snapshots ls
KEY                                                                     PARENT KIND
sha256:7f74ca72855668453fc4207c1e5c50bb5653e72bb9939eb1d5512c001624e927        Committed
\`\`\`

The exact snapshot the failing \`kuke apply\` looks for in
\`dev-init-attach.kukeon.io\` (the namespace its realm just freshly created)
exists in \`default.kukeon.io\`. \`kuke purge realm --cascade\` removed the
namespace's metadata but did not remove the namespace itself — running
\`sudo ctr namespaces remove dev-init-attach.kukeon.io\` and re-invoking
\`make dev-init\` makes the smoke pass again (verified — the run logged
\`Cell \"cattach\": created\` + \`kuke attach exited cleanly\`).

## Root cause

Not yet pinpointed. Candidates for the next picker to consider:

1. **\`teardown_attach_smoke_state\` doesn't remove the containerd namespace.**
   \`scripts/dev-init.sh:435\` only calls \`kuke purge ... --cascade\` for cell,
   stack, space, realm. The cascade clears resource metadata, but the
   \`dev-init-attach.kukeon.io\` containerd namespace survives empty between
   runs (\`ctr namespaces ls\` lists it after teardown). The next run's
   \`kuke create realm\` then \"creates\" the namespace as already-present
   rather than fresh, and \`kuke apply\`'s image-pull-then-unpack lands in a
   namespace whose snapshotter state is in an unexpected partial shape.
2. **Image-pull-then-unpack interaction with cross-namespace
   content-addressed snapshot IDs.** Adjacent root-cause family as #904
   (\`/var/lib/kukebuild/<ns>/\` stale \`cache.db\` strands the next
   \`kuke build\` with the same \`parent snapshot ... does not exist\`
   surface error). #904 was scoped to \`kuke build\`; this surfaces in the
   \`kuke apply\` → \`StartCell\` → \`prepare(snapshot)\` path instead, but the
   shape of the error matches.

## Suggested fix

Probably the smallest viable fix: extend \`teardown_attach_smoke_state\` in
\`scripts/dev-init.sh\` to also remove the \`dev-init-attach.kukeon.io\`
containerd namespace after the \`kuke purge\` chain. Either via
\`sudo ctr namespaces remove dev-init-attach.kukeon.io 2>/dev/null || true\`
or via a daemon-side \`kuke purge realm --cascade --remove-namespace\`-style
verb if the project wants the cleanup observable through the daemon
rather than going around it. Wider verification needed: confirm the
namespace-remove alone makes \`make dev-init\` idempotent across N runs,
or whether the kukebuild / BuildKit cache (#904's lane) also needs a
sister cleanup for the attach-smoke's image lineage to be fully reset.

The fix is small but the contract is load-bearing —
\`CLAUDE.md\`'s "Local smoke test" section claims
"\`scripts/dev-init.sh\` ... is idempotent — re-running on a healthy host
produces a clean re-bootstrap." This bug breaks that claim in the
attach-smoke phase specifically.

## References

- Surfaced during dev work on #916 (kuketty socket perms at bind) — the
  smoke ran clean on first invocation and exercised the fix in the real
  OCI attach path; the second invocation surfaced this independent bug.
- Adjacent lane: #904 (closed) — same \`parent snapshot does not exist\`
  family for the \`kuke build\` path, addressed by tearing down
  \`/var/lib/kukebuild/<ns>/\`. The fix shape there is the analogue of
  what \`teardown_attach_smoke_state\` likely needs here.

eminwux:
## Summary

- `scripts/dev-init.sh`'s `teardown_attach_smoke_state` purges the four `dev-init-attach` metadata resources via `kuke purge ... --cascade`, but the `dev-init-attach.kukeon.io` containerd namespace lingers empty after the purge chain. The next `make dev-init` then re-creates the realm against the surviving namespace, and `kuke apply`'s image-pull→snapshot-unpack fails with `parent snapshot ... does not exist` — `make dev-init` is no longer idempotent across runs, breaking the contract `CLAUDE.md`'s "Local smoke test" pins.
- Fix: at the tail of the purge chain, also `sudo ctr namespaces remove "${ATTACH_SMOKE_NS}"` to backstop whatever residual containerd state survives `runner.PurgeRealm`'s `DeleteNamespace` call. Idempotent (`2>/dev/null || true`) so a clean-state pre-emptive teardown is a silent no-op. `ATTACH_SMOKE_NS` is derived from `KUKEOND_NS` so the dev-profile suffix (`dev-init-attach.dev.kukeon.io`) is picked up automatically alongside the default (`dev-init-attach.kukeon.io`).
- No companion `/var/lib/kukebuild/<ns>/` wipe (the #904 sister lane): the attach smoke uses `kuke apply`, never `kuke build`, so the kukebuild BuildKit cache is never populated for this realm — verified locally (`sudo ls /var/lib/kukebuild/` contains only `kuke-system.kukeon.io`).

## Test plan

- [x] `make test` (root + kukebuild): all packages pass.
- [x] `GOWORK=off go build ./...`: clean compile.
- [x] `GOWORK=off go vet ./...` (excluding e2e per Makefile pattern): clean.
- [x] `bash -n scripts/dev-init.sh`: syntax OK.
- [x] `make dev-init` run #1 (clean state): attach-smoke succeeds, `kuke attach exited cleanly`, the daemon-parity tail still reads `default Ready / kuke-system Ready`.
- [x] `make dev-init` run #2 (no operator intervention between runs): attach-smoke succeeds again — the bug body's `Cell "cattach": failed` / `parent snapshot ... does not exist` repro from `origin/main` does **not** surface with this patch in place.
- [x] Post-run-2 namespace listing: `dev-init-attach.kukeon.io` is absent from `ctr namespaces ls` (the EXIT trap's `ctr namespaces remove` is doing the work).
- [ ] `make e2e`: not run — exercises a different daemon-bringing path that does not cover the dev-init attach-smoke teardown and would not catch this regression.

Closes #919

### DIFF
--- a/scripts/dev-init.sh
+++ b/scripts/dev-init.sh
@@ -439,6 +439,11 @@ ATTACH_SMOKE_CONTAINER="work"
 ATTACH_SMOKE_BASE="${METADATA_ROOT}/${ATTACH_SMOKE_REALM}/${ATTACH_SMOKE_SPACE}/${ATTACH_SMOKE_STACK}/${ATTACH_SMOKE_CELL}/${ATTACH_SMOKE_CONTAINER}"
 ATTACH_SMOKE_SOCKET="${ATTACH_SMOKE_BASE}/tty/socket"
 ATTACH_SMOKE_METADATA="${ATTACH_SMOKE_BASE}/kuketty-metadata.json"
+# Pin to the same suffix the daemon resolves for the dev-init-attach realm
+# (defaults: kukeon.io; KUKEON_PROFILE=dev: dev.kukeon.io). KUKEOND_NS
+# encodes the active suffix as `kuke-system.<suffix>`; strip the
+# `kuke-system.` prefix to derive the attach-smoke realm's namespace.
+ATTACH_SMOKE_NS="${ATTACH_SMOKE_REALM}.${KUKEOND_NS#kuke-system.}"
 ATTACH_SMOKE_TMP="$(mktemp -d)"
 
 teardown_attach_smoke_state() {
@@ -449,6 +454,17 @@ teardown_attach_smoke_state() {
         --realm "${ATTACH_SMOKE_REALM}" --space "${ATTACH_SMOKE_SPACE}" 2>/dev/null || true
     sudo --preserve-env="${PRESERVE_ENV_WORKLOAD}" ./kuke purge space "${ATTACH_SMOKE_SPACE}" --realm "${ATTACH_SMOKE_REALM}" 2>/dev/null || true
     sudo --preserve-env="${PRESERVE_ENV_WORKLOAD}" ./kuke purge realm "${ATTACH_SMOKE_REALM}" 2>/dev/null || true
+    # Backstop the kuke purge chain: `kuke purge realm` does call
+    # DeleteNamespace, but issue #919 surfaces a cross-run trap where the
+    # `dev-init-attach.kukeon.io` containerd namespace lingers empty after
+    # the purge and strands the next run's `kuke apply` image-pull→
+    # snapshot-unpack with "parent snapshot ... does not exist". Removing
+    # the namespace by hand makes the teardown idempotent across N runs.
+    # The attach smoke never targets this realm with `kuke build`, so no
+    # `/var/lib/kukebuild/<ns>/` cache exists to wipe alongside it
+    # (the #904 lane only applies to realms that have been a `kuke build`
+    # target).
+    sudo ctr namespaces remove "${ATTACH_SMOKE_NS}" 2>/dev/null || true
 }
 
 cleanup_attach_smoke() {


eminwux:
**PR #925 Review — `fix(dev-init): remove dev-init-attach namespace after purge to make attach-smoke idempotent`**

LGTM — small, idempotent shell-side backstop that closes the cross-run trap in `teardown_attach_smoke_state`; the `${KUKEOND_NS#kuke-system.}` derivation correctly tracks both default and `KUKEON_PROFILE=dev` suffixes, and the inline comment proactively explains the asymmetry with the #904 `kukebuild` cache lane.

Issue #926, opened 2026-05-28
kuke run: attach ping deadline-exceeds when kuketty Serve() lags StartCell (readiness-handshake gap)


eminwux:
Carved out of #920 — the second of the two bugs that issue filed together. #920 ships the Bug 1 fix (secrets-divergence false positive) as a small scoped PR; this issue tracks Bug 2 separately because the design space is bigger (daemon-side handshake vs. client-side retry) and the failure class is unrelated to the divergence guard #920 fixes.

## Repro

Tree: `origin/main` (commit `8679fb4` at the time #920 was filed) with a healthy daemon and the two default realms. `kuke run` against an existing CellConfig:

```bash
inwx@kukeon-dev:~$ kuke run meta
Cell \"meta\" (realm \"default\", space \"default\", stack \"default\")
  - metadata: created
  - cgroup: created
  - root container: created
  - container \"root\": created
  - container \"work\": created
  - containers: started
Error: error attaching client: ping failed: context deadline exceeded
```

Intermittent — the cell created and started cleanly but the post-\`StartCell\` attach client could not get a PONG within 3 s.

## Observed

The \`error attaching client: ping failed: context deadline exceeded\` surface comes from \`github.com/eminwux/sbsh@v0.12.1/internal/client/clientrunner/io.go:42\` \`(*Exec).dialTerminalCtrlSocket\`:

\`\`\`go
ctx, cancel := context.WithTimeout(sr.ctx, 3*time.Second)
defer cancel()

ping := api.PingMessage{Message: \"PING\"}
var pong api.PingMessage
if err := sr.terminalClient.Ping(ctx, &ping, &pong); err != nil {
    return fmt.Errorf(\"ping failed: %w\", err)
}
\`\`\`

The 3 s budget is from the moment kuke dials the per-container kuketty control socket. The CLI flow is:

  1. \`kuke run\` → \`client.CreateCell\` → \`client.StartCell\` (daemon reports \`containers: started\`).
  2. Immediately after the start prints, the CLI calls \`client.AttachContainer\` → the daemon returns the \`HostSocketPath\` for the kuketty control socket inside the container.
  3. \`attach.Run\` dials that unix socket and ping-then-attaches.

[...]

PR #927, opened 2026-05-28
fix(run): deref *ContainerSecretRef in containerSecretsEqual to suppress false-positive secrets divergence

### LINKED ISSUE #920: kuke run: secrets-divergence false positive on every re-run of `secretRef:`-using cells (pointer-identity compare in containerSecretsEqual); attach ping deadline-exceeds when kuketty Serve() lags StartCell
Two independently reproducible bugs on the `kuke run <existing-cell>` →
attach path. Filing as one because the operator hits them back-to-back on
a single repro and both block the same flow ("get me re-attached to my
cell without `kuke restart`").

## Repro

Tree: `origin/main` (commit `8679fb4` at the time of writing) with a
healthy daemon, the two default realms provisioned (`default`,
`kuke-system`), and a `CellConfig` whose blueprint resolves a
`secretRef:`-source secret into one of its containers. Concretely the
agent author's local `meta` CellConfig:

```yaml
# /opt/kukeon/data/default/configs/meta
apiVersion: v1beta1
kind: CellConfig
metadata:
  name: meta
  realm: default
spec:
  blueprint:
    name: meta
    realm: default
  ...
  secrets:
    claude-code-oauth-token:
      secretRef:
        name: claude-code-oauth-token
        realm: default
```

Whose referenced blueprint declares the container-side consumer:

```yaml
# /opt/kukeon/data/default/blueprints/meta — container "work"
- id: work
  ...
  secrets:
    - name: claude-code-oauth-token
      envName: CLAUDE_CODE_OAUTH_TOKEN
```

Two consecutive `kuke run` invocations against this `CellConfig`, no
intervening operator action:

```bash
inwx@kukeon-dev:~$ kuke run meta
Cell "meta" (realm "default", space "default", stack "default")
  - metadata: created
  - cgroup: created
  - root container: created
  - container "root": created
  - container "work": created
  - containers: started
Error: error attaching client: ping failed: context deadline exceeded

inwx@kukeon-dev:~$ kuke run meta
Error: live cell "meta" spec differs from CellConfig "meta" \
  (spec.containers["work"].secrets) — refusing to attach; \
  use `kuke restart cell meta` to reconcile (stops, updates, starts)
```

Bug 1 (the second invocation) reproduces 100% on this host for any
`CellConfig` that funnels a `secretRef:`-source secret into a container.
Bug 2 (the first invocation) is intermittent — the cell created and
started cleanly but the post-`StartCell` attach client could not get a
PONG within 3 s.

### Bug 1 — secrets-divergence false positive (deterministic)

`cmd/kuke/run/run.go:1358` `containerSecretsEqual` compares
`v1beta1.ContainerSecret` slices with a struct-equality loop:

```go
func containerSecretsEqual(a, b []v1beta1.ContainerSecret) bool {
    if len(a) != len(b) { return false }
    for i := range a {
        if a[i] != b[i] {            // struct == ; SecretRef compared by pointer
            return false
        }
    }
    return true
}
```

`pkg/api/model/v1beta1/container.go:253` `ContainerSecret` holds a
`SecretRef *ContainerSecretRef`. Go's `==` on a struct with a pointer
field compares the pointer by identity, not by value. The round-trip
through `internal/apischeme/scheme.go:944` `secretRefToInternal` and
`:976` `secretRefToExternal` allocates a *fresh* `*ContainerSecretRef`
in each direction:

```go
return &intmodel.ContainerSecretRef{ Name: in.Name, Realm: in.Realm, ... }
// and the inverse:
return &ext.ContainerSecretRef{ Name: in.Name, Realm: in.Realm, ... }
```

So the YAML-decoded side and the daemon-persisted side carry semantically
equal but address-distinct pointers → `a[i] != b[i]` returns `true` →
`divergedContainerFields` (`cmd/kuke/run/run.go:1344-1346`) emits
`secrets` → `rejectDivergentNamedCell` (`:665-669`) refuses with the
`kuke restart cell <name>` pointer. Every second `kuke run` of an
unchanged `secretRef:`-using cell trips this.

The existing test
`TestRun_ExistingCell_DivergingContainerSecrets_RefusesAndPointsToApply`
(`cmd/kuke/run/run_test.go:1042-1101`) only exercises the `FromFile`
source, where `ContainerSecret` is pointer-free and struct `==` works
correctly. There is no positive-side test pinning "two cells with
equal-by-value `SecretRef` must not register as drift", so the bug
escaped review.

### Bug 2 — attach ping context-deadline-exceeded (intermittent)

The `error attaching client: ping failed: context deadline exceeded`
surface comes from
`github.com/eminwux/sbsh@v0.12.1/internal/client/clientrunner/io.go:42`
`(*Exec).dialTerminalCtrlSocket`:

```go
ctx, cancel := context.WithTimeout(sr.ctx, 3*time.Second)
defer cancel()

ping := api.PingMessage{Message: "PING"}
var pong api.PingMessage
if err := sr.terminalClient.Ping(ctx, &ping, &pong); err != nil {
    return fmt.Errorf("ping failed: %w", err)
}
```

The 3 s budget is from the moment kuke dials the per-container kuketty
control socket. The CLI flow is:

  1. `kuke run` → `client.CreateCell` → `client.StartCell` (daemon
     reports `containers: started`).
  2. Immediately after the start prints, the CLI calls
     `client.AttachContainer` → the daemon returns the
     `HostSocketPath` for the kuketty control socket inside the
     container.
  3. `attach.Run` dials that unix socket and ping-then-attaches.

Between step 1's "containers: started" and step 3's first read of the
PONG, kuketty (the process *inside* the container) has to traverse: bind
the control socket (closed by #916/#918), enter the RPC server's
`Serve()` accept loop, and answer the incoming `Ping`. The unix
`connect(2)` from the CLI side succeeds as soon as the bind is in place
(the kernel queues the connect into the listening socket's backlog),
which is exactly the window #918 closed for the EACCES surface — but it
does *not* close the readiness window for `Serve()`: a connection sitting
in the backlog with no `Accept()` to drain it yields no protocol reply,
and the 3 s ping deadline fires.

Distinct from #912 / sbsh#361 (the StartCell→attach EACCES race the
kuketty bind-side fix #918 closes — that one surfaced as EACCES on
`connect(2)`/`bind`, not as a 3 s ping deadline mid-RPC). This is a
"socket present, accept loop not yet running" race, not a "socket file
permissions not yet right" race.

## Root cause

**Bug 1.** Struct equality on `ContainerSecret` is unsafe — the type
holds a pointer field (`SecretRef`) and the apischeme round-trip
guarantees fresh allocations on both sides of the comparison. The diff
helper must compare `SecretRef` by deref'd value, not by `==` on the
struct.

**Bug 2.** The daemon's `StartCell` return is "kuketty process spawned,
control socket bound", not "kuketty has entered Serve() and is
answering Ping". The CLI's attach immediately after a successful
`StartCell` is racing kuketty's `Serve()` boot, which `dialTerminalCtrl
Socket`'s 3 s timeout can lose on a slow host. Need either a daemon-side
readiness handshake before `StartCell` returns (server in `Serve()` ⇒
self-ping or sentinel file), or a client-side retry/backoff in
`runAttachLoop` that distinguishes the 3 s ping-deadline class from a
hard-error class.

## Suggested fix

**Bug 1** (small, scoped, write-one-test).
Replace the struct-`==` loop in `cmd/kuke/run/run.go` `containerSecretsEqual`
with field-by-field comparison plus a deref'd `*ContainerSecretRef` check:

```go
func containerSecretsEqual(a, b []v1beta1.ContainerSecret) bool {
    if len(a) != len(b) { return false }
    for i := range a {
        if a[i].Name != b[i].Name ||
            a[i].FromFile != b[i].FromFile ||
            a[i].FromEnv != b[i].FromEnv ||
            a[i].MountPath != b[i].MountPath ||
            !secretRefEqual(a[i].SecretRef, b[i].SecretRef) {
            return false
        }
    }
    return true
}

func secretRefEqual(a, b *v1beta1.ContainerSecretRef) bool {
    if a == nil || b == nil { return a == b }
    return *a == *b   // ContainerSecretRef is all scalars
}
```

Add a sibling to
`TestRun_ExistingCell_DivergingContainerSecrets_RefusesAndPointsToApply`
in `cmd/kuke/run/run_test.go` that puts an equal-by-value
`SecretRef: &v1beta1.ContainerSecretRef{Name: "x", Realm: "default"}`
on both the existing cell *and* the freshly-decoded YAML, and asserts
that `kuke run` does *not* refuse — the regression guard the current
test set lacks.

**Bug 2** (bigger; carve out its own PR).
Pick one of:

  - Daemon side: after `StartCell` brings up kuketty, the daemon
    self-pings the per-container control socket until either PONG or a
    bounded deadline, and only then returns "containers: started" to
    the CLI. Pushes the readiness wait inside the same RPC the CLI is
    already blocked on, so no client-side retry is needed and the
    existing 3 s ping budget is enough.
  - Client side: in `cmd/kuke/run/attach.go` `runAttachLoop`,
    distinguish the sbsh "ping failed: context deadline exceeded" class
    via `errors.Is` (an `errdefs.ErrAttachPingTimeout`-shaped sentinel
    promoted from sbsh) and retry the dial-then-ping with a few hundred
    ms of backoff up to a small budget (e.g. 10 s total). Less invasive
    but pushes the race onto operator-visible wall-clock.

The daemon-side option is the one that matches the existing contract
("`containers: started` means ready to attach"); the client-side
fallback is a smaller change but weakens that contract.

## References

- #912 / sbsh#361 — StartCell→attach EACCES race (bind-side perms);
  closed by sbsh@v0.12.1. Same surface area, different failure class:
  there the connect failed; here the connect succeeds but PONG never
  arrives.
- #916 / #918 — kuketty control socket perms applied at bind, closing
  the pre-Serve EACCES window. Bug 2 is the readiness-handshake gap
  that survives the perms fix.
- `cmd/kuke/run/run.go:1297-1351` `divergedContainerFields` /
  `containerSecretsEqual` — site of Bug 1.
- `internal/apischeme/scheme.go:925-987` — fresh-allocation round-trip
  through `convertSecretsToInternal` / `buildSecretsExternalFromInternal`
  that makes the pointer-identity compare unsafe.
- `github.com/eminwux/sbsh@v0.12.1/internal/client/clientrunner/io.go:42`
  — the 3 s ping budget Bug 2 loses against.

eminwux:
## Summary
- `containerSecretsEqual` in `cmd/kuke/run/run.go` compared `[]ContainerSecret` elements with struct `==`. `ContainerSecret` carries a `*ContainerSecretRef` pointer field, so struct `==` compares the pointer by identity. The apischeme round-trip (`secretRefToInternal` / `secretRefToExternal`) allocates a fresh `*ContainerSecretRef` on each conversion, so YAML-decoded and daemon-persisted sides were always address-distinct even when value-equal — every re-`kuke run` of a `secretRef:`-using cell tripped the divergence guard and printed `kuke restart cell …` (#920 Bug 1, deterministic repro).
- Fix: keep the comment trail explaining *why* struct `==` is unsafe here, and replace the inner compare with a field-by-field check plus a new `secretRefEqual` helper that deref's the pointer when both sides are non-nil and treats matching nil-ness as equal. `ContainerSecretRef` is all scalar strings, so a direct `==` on the dereferenced value is safe.
- Adds `TestRun_ExistingCell_EqualSecretRefByValue_DoesNotRefuse` as a sibling to the existing `…DivergingContainerSecrets_RefusesAndPointsToApply` — puts an equal-by-value `SecretRef` on both the YAML side and the GetCell stub side, and asserts `kuke run` does *not* refuse. Pre-fix this test fails with the exact `cell "my-cell" exists with diverging spec (spec.containers["work"].secrets)` message the operator sees in the issue's repro; the existing Diverging test only exercised `FromFile` (pointer-free), which is how the bug escaped review.
- Bug 2 from #920 (attach ping context-deadline-exceeded from the post-`StartCell` Serve()-readiness race) is **deliberately not included** — the issue body itself defers it ("Bug 2 (bigger; carve out its own PR)") and the design space (daemon-side handshake vs. client-side retry budget) is independent. Filed as #926 so the deferral remains visible after #920 auto-closes on this merge.

## Test plan
- [x] `GOWORK=off go test ./cmd/kuke/run/ -run 'TestRun_ExistingCell_(DivergingContainerSecrets|EqualSecretRefByValue)' -v` — both pass on this branch; the new test fails (with the issue's exact divergence error) when the fix is stashed.
- [x] `make test-root` — full root-module suite, all packages green.
- [x] `GOWORK=off go vet ./cmd/kuke/run/...` — clean.
- [x] `pre-commit run --files cmd/kuke/run/run.go cmd/kuke/run/run_test.go` — all hooks pass (gofmt, addlicense, golangci-lint scoped on the changed files).
- [ ] `make dev-init` smoke (the `kuke get realms` daemon-parity tail in CLAUDE.md §Local smoke test) — not run; the diff is scoped to `cmd/kuke/run`'s in-memory diff helper, touches no daemon code, no build-path code, and nothing under `/opt/kukeon`. The unit test exercises the exact code path the operator's repro trips on.
- [ ] `make e2e` — not run; e2e exercises live containerd + kuketty paths, which this CLI-side diff does not touch. The repro path (`kuke run` re-tick against a Ready cell) is unit-covered.

Closes #920

### DIFF
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -1353,18 +1353,35 @@ func divergedContainerFields(actual, desired v1beta1.ContainerSpec) []string {
 // containerSecretsEqual compares two ContainerSecret slices field-by-field.
 // nil and empty are treated as equal so YAML that omits secrets does not
 // register as drift against on-disk metadata that persisted it as nil.
-// ContainerSecret carries only scalar string fields, so a direct == on each
-// element is enough.
+// ContainerSecret carries a *ContainerSecretRef pointer; struct == on it
+// would compare that pointer by identity, and the apischeme round-trip
+// allocates a fresh *ContainerSecretRef on each conversion, so YAML-decoded
+// and daemon-persisted sides are always address-distinct even when
+// value-equal. Issue #920.
 func containerSecretsEqual(a, b []v1beta1.ContainerSecret) bool {
 	if len(a) != len(b) {
 		return false
 	}
 	for i := range a {
-		if a[i] != b[i] {
+		if a[i].Name != b[i].Name ||
+			a[i].FromFile != b[i].FromFile ||
+			a[i].FromEnv != b[i].FromEnv ||
+			a[i].MountPath != b[i].MountPath ||
+			!secretRefEqual(a[i].SecretRef, b[i].SecretRef) {
 			return false
 		}
 	}
 	return true
+}
+
+// secretRefEqual compares two *ContainerSecretRef by deref'd value, treating
+// matching nil-ness as equal. ContainerSecretRef is all scalar strings, so a
+// direct == on the dereferenced struct is safe.
+func secretRefEqual(a, b *v1beta1.ContainerSecretRef) bool {
+	if a == nil || b == nil {
+		return a == b
+	}
+	return *a == *b
 }
 
 // containerTtysEqual compares two ContainerTty pointers. Empty values on
--- a/cmd/kuke/run/run_test.go
+++ b/cmd/kuke/run/run_test.go
@@ -1100,6 +1100,106 @@ func TestRun_ExistingCell_DivergingContainerSecrets_RefusesAndPointsToApply(t *t
 	}
 }
 
+// TestRun_ExistingCell_EqualSecretRefByValue_DoesNotRefuse pins the
+// regression from #920: containerSecretsEqual used to compare
+// []ContainerSecret element-by-element with struct ==, which on a struct
+// holding a *ContainerSecretRef pointer field compares the pointer by
+// identity. The apischeme round-trip allocates a fresh *ContainerSecretRef
+// on each conversion, so the YAML-decoded side and the daemon-persisted
+// side are always address-distinct even when value-equal — every
+// re-`kuke run` of a secretRef:-using cell tripped the divergence guard
+// and printed the `kuke restart cell ...` pointer.
+//
+// The sibling Diverging... test above only exercises FromFile (pointer-free
+// fields), so the pointer-identity bug escaped review. This test puts a
+// SecretRef equal by value on both sides and asserts `kuke run` does not
+// refuse.
+func TestRun_ExistingCell_EqualSecretRefByValue_DoesNotRefuse(t *testing.T) {
+	t.Cleanup(viper.Reset)
+
+	const yamlWithSecretRef = `apiVersion: v1beta1
+kind: Cell
+metadata:
+  name: my-cell
+spec:
+  id: my-cell
+  realmId: my-realm
+  spaceId: my-space
+  stackId: my-stack
+  containers:
+    - id: root
+      root: true
+      image: registry.eminwux.com/busybox:latest
+      command: sleep
+      args:
+        - "3600"
+    - id: work
+      image: registry.eminwux.com/busybox:latest
+      command: sleep
+      args:
+        - "3600"
+      secrets:
+        - name: claude-code-oauth-token
+          secretRef:
+            name: claude-code-oauth-token
+            realm: default
+`
+
+	existing := v1beta1.CellDoc{
+		Metadata: v1beta1.CellMetadata{Name: "my-cell"},
+		Spec: v1beta1.CellSpec{
+			RealmID: "my-realm",
+			SpaceID: "my-space",
+			StackID: "my-stack",
+			Containers: []v1beta1.ContainerSpec{
+				{
+					ID:      "root",
+					Root:    true,
+					Image:   "registry.eminwux.com/busybox:latest",
+					Command: "sleep",
+					Args:    []string{"3600"},
+				},
+				{
+					ID:      "work",
+					Image:   "registry.eminwux.com/busybox:latest",
+					Command: "sleep",
+					Args:    []string{"3600"},
+					Secrets: []v1beta1.ContainerSecret{
+						{
+							Name: "claude-code-oauth-token",
+							SecretRef: &v1beta1.ContainerSecretRef{
+								Name:  "claude-code-oauth-token",
+								Realm: "default",
+							},
+						},
+					},
+				},
+			},
+		},
+		Status: v1beta1.CellStatus{State: v1beta1.CellStateReady},
+	}
+	fc := &fakeClient{
+		getCellFn: func(_ v1beta1.CellDoc) (kukeonv1.GetCellResult, error) {
+			return kukeonv1.GetCellResult{
+				Cell:                     existing,
+				MetadataExists:           true,
+				CgroupExists:             true,
+				RootContainerExists:      true,
+				RootContainerTaskRunning: true,
+			}, nil
+		},
+	}
+	cmd, _ := newCmd(t, fc)
+	cmd.SetArgs([]string{"-f", writeTempYAML(t, yamlWithSecretRef), "-d"})
+
+	if err := cmd.Execute(); err != nil {
+		t.Fatalf("Execute returned %v, want nil (equal-by-value SecretRef must not register as drift)", err)
+	}
+	if fc.createCalls != 0 {
+		t.Errorf("CreateCell calls=%d want 0 (Ready cell with matching spec is a no-op)", fc.createCalls)
+	}
+}
+
 // TestRun_ExistingCell_DivergingContainerTty_RefusesAndPointsToApply
 // exercises the tty branch of divergedContainerFields. ContainerTty is
 // user-authored shell-UX config the daemon persists verbatim and never


eminwux:
**PR #927 Review — `fix(run): deref *ContainerSecretRef in containerSecretsEqual to suppress false-positive secrets divergence`**

LGTM — pointer-identity diagnosis is correct (`apischeme.secretRefToInternal`/`secretRefToExternal` allocate fresh pointers on round-trip), `secretRefEqual` follows the existing `containerTtysEqual` nil-check-then-deref precedent, `ContainerSecretRef` is all scalar strings so `*a == *b` after nil-check is safe, and the new test exercises the exact pointer-bearing path the sibling Diverging test missed.

PR #930, opened 2026-05-28
test(e2e): reuse attachConnectGrace for runFromStopped scenario

### LINKED ISSUE #914: e2e: clarify runFromStoppedAttachGrace vs attachConnectGrace divergence
Surfaced during review of #913.

`e2e/e2e_kuke_attach_test.go:455` introduces `runFromStoppedAttachGrace = 1 * time.Second` for the new `TestKuke_RunFromStopped_NonRootKukeonGroup_NoEACCES`. The peer test `TestKuke_AttachDetach_KeepsTaskRunning` already has `attachConnectGrace = 2 * time.Second` at `e2e/e2e_kuke_attach_test.go:47` serving an overlapping purpose (let pkg/attach's raw-mode setup finish before the detach keystroke arrives, so the keystroke isn't echoed verbatim and the test doesn't hang on the exit timeout).

Both grace concerns apply to the new test — it does both a dial (the EACCES window check) AND a detach keystroke (the filter-wire race). The new constant's docstring justifies 1s on dial-timing grounds only; it does not address why the filter-wire race the peer's 2s targets is not a concern here.

CI is green now, so this is speculative — but the failure mode of an under-grace would be the keystroke being forwarded verbatim, the 20s exit timeout firing on `permission denied`-free output, and the test attributing the hang to a non-existent regression. Across 20 iterations the surface area for that timing miss multiplies.

**Suggested fix:** either reuse `attachConnectGrace` (`= 2 * time.Second`) — the dial check tolerates a longer grace, and the test's 20-iteration cost is still under the e2e budget at 2s/iter — or extend `runFromStoppedAttachGrace`'s docstring to explicitly explain why the filter-wire race the peer's 2s targets doesn't apply in this scenario.

Non-blocking; #913 shipped as-is.

eminwux:
## Summary

`TestKuke_RunFromStopped_NonRootKukeonGroup_NoEACCES` defined its own `runFromStoppedAttachGrace = 1 * time.Second` to consume the EACCES window before sending the detach keystroke, but the scenario also depends on the in-process pkg/attach loop entering raw mode and wiring its keystroke filter — the *exact* race the peer `attachConnectGrace = 2 * time.Second` (used by `TestKuke_AttachDetach_KeepsTaskRunning`) was sized for. A 1s grace risks the detach byte arriving before the filter is live: the byte is forwarded verbatim, the workload never sees a detach signal, and the test would hang to the 20s exit timeout on a CI runner with no `permission denied` in output — attributing the failure to a non-existent regression.

Consolidate onto the existing `attachConnectGrace`. The constant's docstring ("complete the unix-socket handshake") already covers the dial-side concern (the EACCES window), so the longer 2s grace covers both purposes. At 20 iterations the added 1s/iter (20s total) stays well inside the e2e suite's 60s per-test budget on a healthy runner.

The use-site comment now spells out the dual-purpose so a future reader sees why this scenario reaches for the peer's grace.

Surfaced during review of #913.

## Test plan

- [x] `go vet ./e2e/` clean
- [x] `go test -c ./e2e/` compiles (e2e test binary builds)
- [ ] Full e2e suite — `make dev-init` smoke not run for this comment + sleep-duration tweak; deferred to PR-time CI

Closes #914

### DIFF
--- a/e2e/e2e_kuke_attach_test.go
+++ b/e2e/e2e_kuke_attach_test.go
@@ -443,16 +443,6 @@ func getCellIDNoDaemon(t *testing.T, runPath, realmName, spaceName, stackName, c
 // 60s budget on a healthy runner.
 const runFromStoppedRaceIterations = 20
 
-// runFromStoppedAttachGrace is the wait between launching `kuke run`
-// (which races StartCell → attach internally) and sending the detach
-// keystroke. The attach phase enters the in-process pkg/attach loop
-// within tens of milliseconds on a healthy host; this grace is set so
-// the EACCES window, if it re-opens, is consumed by the dial before the
-// detach byte arrives. A v0.12.0 race would surface as a non-zero exit
-// with `permission denied` on stderr inside this window; a v0.12.1
-// healthy attach loop swallows the keystroke and exits cleanly.
-const runFromStoppedAttachGrace = 1 * time.Second
-
 // runFromStoppedExitTimeout caps the wait for a single `kuke run`
 // invocation to return after its detach keystroke. The healthy path
 // (post-v0.12.1) returns within a second; the cap is generous so a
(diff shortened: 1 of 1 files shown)


eminwux:
## Review: test(e2e): reuse attachConnectGrace for runFromStopped scenario

**Verdict: LGTM**

### Correctness

The consolidation is correct. `runFromStoppedAttachGrace = 1 * time.Second` was under-sized: the `runFromStopped` scenario calls `kuke run`, which enters the in-process `pkg/attach` loop — meaning the test must wait not only for the EACCES window (the v0.12.0 dial race) but also for raw-mode entry and filter-wire setup before sending the detach byte. `attachConnectGrace = 2 * time.Second` was already explicitly sized for exactly that: "the in-process pkg/attach loop to enter raw mode and complete the unix-socket handshake on a busy CI runner." Using 1s here was a latent flakiness risk; the PR summary's analysis (detach byte arriving before the filter is live → byte forwarded verbatim → test hangs to 20s exit timeout with no `permission denied` signal) is sound.

The per-iteration cost increase (1s → 2s, 20 iterations = 20s added) stays comfortably inside the 60s per-test budget.

### Dead-symbol check

`runFromStoppedAttachGrace` is fully gone from the PR branch — no definition, no call-sites, no references anywhere in the repo. `attachConnectGrace` is confirmed-active at five use-sites (lines 141, 250, 275, 606 + the new comment at 600).

### Commit signature

Single commit — verified ✓.

### Test plan

`make e2e` passed in CI. `make dev-init` smoke is correctly deferred: no daemon code was touched, so a live re-bootstrap would not add signal. The test-plan checklist accurately reflects this.

---

### Non-blocking nit (followup)

The `attachConnectGrace` docstring (line 43–47) says "between starting `kuke attach` and sending the detach sequence" — now that the constant is also used in a `kuke run` scenario, a one-line addition ("also reused by scenarios where `kuke run` enters the in-process attach loop") would help a future reader who scans only the constant definition. The use-site comment added here is thorough, so this is purely additive.

Issue #933, opened 2026-05-28
fix(run): retry attach on EACCES so stale pre-fix socket does not permanently block kuke run from stopped


eminwux:
## Summary

`kuke run <stopped-cell>` fails with `connect: permission denied` when a stale tty socket from a kuketty binary predating sbsh#361 is present on the host.

## Steps to reproduce

1. A cell that was last started with a kuketty binary older than sbsh v0.12.1 — socket at `<RunPath>/data/.../tty/socket` has mode `0o640` (group read, no write).
2. Update `kuketty` (e.g. via `make dev-init`).
3. `kuke run <that-cell>` while the cell is stopped.

**Expected:** cell starts and attach succeeds (kuketty replaces the stale socket at `0o660`).
**Actual:**
```
Error: error attaching client: ping failed: dial unix /opt/kukeon/s/<hash>: connect: permission denied
```

## Root cause

`kuke run` calls `StartCell`, waits the 300 ms post-start liveness probe, then immediately dials the symlink path. At that point kuketty (new binary, inside the container) is still in its init path and has not yet called `os.Remove` + `listenUnixWithMode` to replace the stale socket. The stale inode is at mode `0o640` — group has read only, no write — so `connect(2)` returns `EACCES`.

`runWithPingRetry` (`cmd/kuke/run/attach.go`) only retries on `context.DeadlineExceeded` (the sbsh 3 s ping-window timeout addressed by #928). **`EACCES` propagates as a hard failure immediately.**

sbsh#361 (shipped in v0.12.1, already in `go.mod` and `/opt/kukeon/bin/kuketty`) closed the *new-socket* race — sockets born after the fix have the correct mode from `bind(2)` time. It does not close the *stale-socket* race: the window between `StartCell` returning and kuketty calling `os.Remove` on the stale inode.

## Fix

Extend `runWithPingRetry` / `isAttachPingTimeout` in `cmd/kuke/run/attach.go` to also treat `EACCES` / `syscall.EACCES` as a retryable condition (same budget and backoff as the ping-deadline retry). Once kuketty unlinks the stale socket and binds a new `0o660` one, the retry succeeds. After the stale socket has been replaced on a cell's first post-upgrade start, the race window collapses to the negligible Remove→Listen gap.

An `ENOENT` retry (brief gap between `os.Remove` and `listenUnixWithMode`) may also be worth adding for defense-in-depth, though that window is sub-millisecond in practice.

## Observed environment

[...]

PR #936, opened 2026-05-29
fix(run): re-apply attach socket mode in no-op StartCell so old-kuketty wrong-mode socket is corrected

### LINKED ISSUE #935: fix(run): kuke run EACCES on permanently-wrong-mode socket when old kuketty stays running across dev-init
## Summary

`kuke run <cell>` fails permanently with `connect: permission denied` when the cell's kuketty process is still running from before a `make dev-init` that updated the binary. The EACCES-retry fix from #933 does not help because the socket is the **active, listening** socket of the old kuketty — its mode is permanently wrong and never changes during the retry window.

## Steps to reproduce

1. A cell (`kukeon-pm-0`) is running with an old kuketty binary. Its control socket at `<RunPath>/data/.../tty/socket` has mode `0o640` (group read, no write) — the pre-sbsh#361 mode.
2. Run `make dev-init` to update `/opt/kukeon/bin/kuketty` without restarting the cell.
3. `kuke run <that-cell>` while the cell is still running (workload never exited, kuketty still up).

**Expected:** attach succeeds (or fails with a clear diagnostic that the cell needs to be restarted).
**Actual:**
```
Error: attach stale socket: kuketty has not yet unlinked the pre-fix tty socket: error attaching client: ping failed: dial unix /opt/kukeon/s/<hash>: connect: permission denied
```
The EACCES retry budget (10 s) exhausts. The socket mode never changes.

## Root cause

`StartCell` in `internal/controller/runner/start.go:605` detects all container tasks already running and returns early:
```
"cell tasks already running, StartCell is a no-op"
```
This path skips `attachableBuildOpts` and `attachablePostCreateChown` entirely — so `chownAttachableStaleChildren` never runs and the socket permissions are never corrected.

The old kuketty process bound the socket with mode `0o640` (group read-only). `connect(2)` on a Unix socket requires write permission on the inode. A kukeon-group member on the host gets `EACCES`.

The EACCES-retry fix from #933 addressed the **stale-socket-during-new-kuketty-startup** race: a wrong-mode socket that is about to be replaced. It does not help when the socket is the **live, permanent listener** of a still-running process — retrying EACCES 10 s on a socket whose permissions never change accomplishes nothing.

## Observed environment

- kukeon: `main` @ `fd3518c`
- kuketty binary updated at 22:56 via `make dev-init`; old kuketty PID 249905 running since 16:59
- Socket: `srw-r----- 1 root kukeon 0 May 28 16:59 socket` (inode confirmed listening via `/proc/net/unix` inside the container)
- All container tasks `RUNNING` in containerd → `StartCell` no-op → no chown → no-op on socket mode

## Fix direction

Two complementary approaches:

**1. Re-apply socket permissions in the no-op StartCell path.**
When `StartCell` returns early because tasks are already running, the daemon should still verify (and if necessary correct) the attach socket's mode and group. A lightweight `os.Chmod` + `os.Chown` on the live socket inode by root is safe — it is idempotent when the mode is already correct and closes the wrong-mode window without restarting the workload.

**2. Detect "permanently wrong" EACCES in the retry loop.**
The EACCES retry in `cmd/kuke/run/attach.go` (`runWithPingRetry`) could distinguish a transient EACCES (socket about to be replaced by a starting kuketty) from a persistent one (socket is actively listened to and will never change). After a small number of retries with continued EACCES, the loop could surface a clearer error ("cell socket has wrong permissions; try `kuke restart cell` or contact the operator") rather than exhausting the budget silently.

## Related

- #933 — EACCES retry fix (handles transient stale-socket-during-startup race, not this persistent case)
- #912 — downstream tracking for sbsh#361 bump + e2e regression
- sbsh#361 — upstream bind/chmod fix (v0.12.1)

eminwux:
## Summary

- `kuke run <cell>` failed permanently with `connect: permission denied` when the cell's kuketty process was still running from before a `make dev-init` that updated the binary (#935). The old kuketty (pre-sbsh#361) bound its control socket with mode `0o640` (group-read only); `connect(2)` needs group-**write**, so a kukeon-group member on the host got EACCES forever. The #933 EACCES-retry never helped — the live listener's inode mode never changes during the retry window.
- Root cause: `StartCell` short-circuits when all container tasks are already running (`start.go:591`, "cell tasks already running, StartCell is a no-op") and that path skips `attachableBuildOpts` + `attachablePostCreateChown` entirely, so the wrong-mode socket is never corrected.
- Fix (issue's **approach #1** — the root-cause fix): in the idempotent no-op path, re-assert the live socket inode's mode (`0o660`, mirroring `ctr.AttachableSocketMode`) and the kukeon group on every Attachable container, via a new `reapplyAttachableSocketPerms`. A root-owned `os.Chmod` + `os.Chown` on the live inode is idempotent when the mode is already correct (a post-#361 kuketty already bound `0o660`) and never restarts the workload. This makes `kuke run` succeed against the still-running cell, satisfying the issue's "**Expected:** attach succeeds" terminal state.

### Note on the issue's two approaches

The issue lists two complementary approaches. I implemented **#1** (correct the socket in the daemon's no-op path) as the complete, root-cause fix — it makes attach actually succeed, which is the primary Expected outcome. I deliberately **deferred #2** (a client-side heuristic in `runWithPingRetry` to surface a clearer error after N persistent-EACCES retries):

- #1 resolves the reported failure entirely, so #2's clearer-error fallback is unreachable in the reported scenario.
- #2 is in direct tension with #933: the client cannot distinguish a *transient* EACCES (a starting kuketty about to re-bind the socket — must keep retrying for the full budget) from a *permanent* one (a still-running old kuketty) without a host-side heuristic. Cutting the EACCES budget short to surface an error would regress #933's transient case.

Documenting the call here so the reviewer can push back if #2 should also land (it would be a separable follow-up, not a blocker for the reported bug).

### Best-effort contract

`reapplyAttachableSocketPerms` is best-effort: a per-container chmod/chown failure is logged and skipped rather than failing the no-op StartCell, because the cell is already healthy and running and a chmod miss must not regress the idempotent-skip contract. A socket kuketty has not yet bound (ENOENT) is a benign no-op — the normal startup chown path sets the mode when the listener appears.

## Test plan

- [x] `go test ./internal/controller/runner/ -run ReapplyAttachableSocketPerms` — 4 new tests pass (corrects wrong mode → `0o660`; idempotent on already-correct `0o660`; skips non-Attachable; tolerant of missing socket).
- [x] Full CI `test-root` target (`go test $(go list ./... | grep -v /e2e)`) — exit 0, 0 failures.
- [x] `test-kukebuild` target (`cd cmd/kukebuild && go test ./...`) — exit 0.
- [x] `go build ./...` + `go vet ./internal/controller/runner/` — clean (run with `GOWORK=off GOTOOLCHAIN=go1.24.5` to match CI's go.mod-pinned toolchain; the agent dev host's go.work forces go 1.25.5, under which the vendored `containerd/cgroups` dep fails to compile — environmental, unrelated to this change).
- [x] `pre-commit` (incl. `golangci-lint --new-from-rev HEAD`) — **0 issues**. Note: the lint hook was OOM-killed (SIGKILL/-9, no diagnostics) on first runs because a concurrent agent session held a 4.8G RAM-backed tmpfs dir on this shared host; re-running with `GOGC=20 GOMEMLIMIT=450MiB GOMAXPROCS=1` fit it in available memory and it passed clean. The commit's own hook invocation passed with the same budget.
- [ ] `make dev-init` end-to-end smoke — not run: requires a destructive daemon re-bootstrap (`kuke daemon reset`) on the shared dev host, which would disrupt other concurrent agent sessions' running cells. The change only adds an idempotent chmod/chown in a daemon code path fully covered by the new unit tests; the daemon-parity tail is unaffected.

Closes #935

### DIFF
--- a/internal/controller/runner/attachable.go
+++ b/internal/controller/runner/attachable.go
@@ -586,6 +586,69 @@ func chownAttachableStaleChildren(root string, uid, gid int) error {
 	})
 }
 
+// attachableSocketReapplyMode is the mode the idempotent StartCell no-op
+// path (#935) re-asserts on a live per-container attach socket inode. It
+// mirrors ctr.AttachableSocketMode ("0660"): rw for the owning container
+// uid + rw for the kukeon group, no world. connect(2) on a Unix socket
+// requires write permission on the inode, so a socket an old kuketty
+// (pre-sbsh#361) bound with mode 0o640 (group-read only) is permanently
+// undialable by a kukeon-group member on the host. Re-chmod'ing the live
+// inode closes that window without bouncing the workload.
+const attachableSocketReapplyMode os.FileMode = 0o660
+
+// reapplyAttachableSocketPerms re-asserts the mode and group of every live
+// per-container attach socket on the StartCell idempotent no-op path (#935).
+//
+// When StartCell short-circuits because all container tasks are already
+// running, it skips the destructive recreate loop and with it
+// attachableBuildOpts + attachablePostCreateChown — so a socket an old
+// kuketty bound with the wrong mode (0o640 group-read only, the pre-sbsh#361
+// mode) is never corrected. A kukeon-group member on the host then dials
+// that live, listening socket and gets EACCES forever: the inode's mode
+// never changes during the attach retry window because nothing rewrites it
+// (the EACCES-retry fix from #933 only helps a socket about to be replaced
+// by a starting kuketty, not the permanent listener of a still-running one).
+//
+// The fix is a root-owned os.Chmod + os.Chown on the live socket inode:
+// idempotent when the mode is already correct (a post-#361 kuketty already
+// bound 0o660), and safe because it never restarts the workload. It is
+// best-effort — a per-container failure is logged and skipped rather than
+// failing the no-op StartCell, because the cell is already healthy and
+// running and a chmod miss must not regress the idempotent-skip contract.
+// A socket kuketty has not bound yet (ENOENT) is a benign no-op: the normal
+// startup chown path will set the mode when the listener appears.
+func (r *Exec) reapplyAttachableSocketPerms(cell intmodel.Cell) {
+	for i := range cell.Spec.Containers {
+		spec := cell.Spec.Containers[i]
+		if !spec.Attachable {
+			continue
+		}
+		socketPath := fs.ContainerSocketPath(
+			r.opts.RunPath,
+			spec.RealmName, spec.SpaceName, spec.StackName, spec.CellName, spec.ID,
+		)
+		if chmodErr := os.Chmod(socketPath, attachableSocketReapplyMode); chmodErr != nil {
+			if errors.Is(chmodErr, os.ErrNotExist) {
+				continue
+			}
+			r.logger.WarnContext(r.ctx,
+				"failed to re-chmod attachable socket on idempotent StartCell skip",
+				"socket", socketPath, "error", chmodErr)
+			continue
+		}
+		// Leave the owning uid untouched (-1); only re-assert the kukeon
+		// group when one is configured, mirroring attachablePostCreateChown.
+		if gid := r.opts.KukeonGroupGID; gid > 0 {
+			if chownErr := os.Chown(socketPath, -1, gid); chownErr != nil &&
+				!errors.Is(chownErr, os.ErrNotExist) {
+				r.logger.WarnContext(r.ctx,
+					"failed to re-chown attachable socket group on idempotent StartCell skip",
+					"socket", socketPath, "gid", gid, "error", chownErr)
+			}
+		}
+	}
+}
+
 // attachableSocketSymlinkDirMode is the mode applied to <RunPath>/s when
 // the runner stages it on first Attachable provision. 0o755 mirrors the
 // /opt/kukeon root layout: world-traversable so any process can resolve a
--- a/internal/controller/runner/attachable_test.go
+++ b/internal/controller/runner/attachable_test.go
@@ -891,6 +891,137 @@ func TestRemoveAttachableSocketRuntimeArtifacts_SkipsNonAttachable(t *testing.T)
 	}
 }
 
+// seedAttachableSocket stages a plain file standing in for the bound
+// kuketty socket inode at the deep ContainerSocketPath, created with mode.
+// A regular file is sufficient: os.Chmod / os.Stat are inode-mode
+// operations and do not distinguish a socket from a file, the same way the
+// teardown tests above stand in a regular file for the bound socket.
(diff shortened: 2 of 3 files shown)


eminwux:
## Review

**Verdict: LGTM** — unconditional.

### Correctness

The root-cause analysis is sound. `startCellLocked`'s idempotent-skip path (`cellTasksAllRunningFn` short-circuit at `start.go:591`) entirely bypasses the destructive recreate loop, including `attachablePostCreateChown` — so a socket an old kuketty (pre-sbsh#361) bound at mode `0o640` (group-read only) is never corrected during the retry window. `connect(2)` on a Unix socket inode requires **write** permission; the EACCES is permanent against a live listener, exactly as described in #933's analysis.

`reapplyAttachableSocketPerms` fixes this correctly:

- `os.Chmod(socketPath, 0o660)` — always runs, mode correction is unconditional on the `Attachable` path.
- `os.Chown(socketPath, -1, gid)` — gated on `gid > 0`, which is correct: with no kukeon group configured the socket group was already set to root by `attachablePostCreateChown`, and without a kukeon group there are no group-member callers to care about. The mode fix still fires.
- Best-effort pattern (log + continue on EACCES/other errors) is consistent with the no-op StartCell contract — a chmod miss must not flip a healthy, running cell to Failed.
- `attachableSocketReapplyMode = 0o660` correctly mirrors `ctr.AttachableSocketMode = "0660"`.
- `internalCell` at the call site is the disk-loaded cell (`r.GetCell`), so `cell.Spec.Containers` is fully populated.

### Dead code

All three new symbols confirmed-used:
- `reapplyAttachableSocketPerms`: call site at `start.go`, 4 test call sites.
- `attachableSocketReapplyMode`: used in `reapplyAttachableSocketPerms` body, 2 assertion sites in tests.
- `seedAttachableSocket`: 3 test call sites.

### Tests

Four unit tests cover the four contract invariants — wrong-mode corrected, idempotent on correct mode, non-`Attachable` skipped, missing socket tolerated. Test coverage is solid for this code path.

### `make dev-init` flag (per CLAUDE.md)

CLAUDE.md requires this check in the test plan for any change touching daemon code. The PR documents the skip explicitly ("destructive daemon re-bootstrap… would disrupt other concurrent agent sessions") with a clean rationale. The fix is a single idempotent chmod/chown in one daemon code path, fully covered by the new unit tests, with no effect on the daemon-parity tail. The explicit documented skip is acceptable for this change; noting it here per reviewer convention.

### Approach #2 deferral

The tension with #933 is correctly identified: a client-side heuristic that cuts the EACCES budget short to surface a clearer error would regress transient-EACCES retry from #933, since the client cannot distinguish permanent from transient without host-side introspection. Deferral is the right call — approach #1 is the complete fix and #2 is unreachable in the reported scenario.

### Commit

Single commit `bae19b6`, verified signature. ✓

PR #952, opened 2026-05-29
refactor(lifecycle): extract kukeond ready-check helpers and relocate package

### LINKED ISSUE #899: daemon: extract kukeond ready-check helpers into lifecycle/
Surfaced during review of #898.

`cmd/kuke/daemon/recreate/recreate.go:266-300` duplicates `waitForKukeondReady`, `pingKukeond`, and the `kukeondReadyTimeout`/`kukeondReadyTick` constants from `cmd/kuke/init/init.go:480-520`. PR #898's drift-prevention story (the factored `ProvisionKukeondCell`) covers cell-creation but the wait helpers are now a parallel drift surface — a future change to one will not propagate to the other.

**Suggested fix:** extract the helpers + constants into `cmd/kuke/daemon/internal/lifecycle/` (where `StopPhase`, `IsCellRunning`, `ResolveSocketPath` etc. already live) and have both call sites use the shared version. The lifecycle package's whole point is preventing exactly this kind of "the three verbs cannot drift on what 'graceful' means" drift, per its package doc.

**Sites to update:**
- `cmd/kuke/init/init.go:480-520` — replace with call to lifecycle helper
- `cmd/kuke/daemon/recreate/recreate.go:266-300` — same

Non-blocking; PR #898 shipped as-is.

eminwux:
## Summary

- Extracts `waitForKukeondReady` / `pingKukeond` and the `kukeondReadyTimeout` / `kukeondReadyTick` constants — previously duplicated in `cmd/kuke/init/init.go` and `cmd/kuke/daemon/recreate/recreate.go` — into a single `cmd/kuke/internal/lifecycle/ready.go` exporting `WaitForKukeondReady`, `PingKukeond`, `KukeondReadyTimeout`, `KukeondReadyTick`.
- Relocates the lifecycle package from `cmd/kuke/daemon/internal/lifecycle/` → `cmd/kuke/internal/lifecycle/` so it's reachable from `cmd/kuke/init/` (Go's `internal/` rule blocked the literal AC path; see the heads-up on the issue). All ~14 existing daemon-lifecycle callers updated to the new import path; package name and exported surface unchanged.
- The two prior copies had drifted: `init.go`'s `pingKukeond` did an explicit raw `net.Dial` before the kukeonv1 Ping while `recreate.go`'s was Ping-only. The consolidated implementation keeps the dial-first behavior (the safer of the two, per init.go's `// dial AND ping` comment that guards against the socket-bound-but-handler-not-installed window) so both call sites now match init's prior semantics. Mentioned this divergence here because the issue body presupposed byte-identical duplicates — flagging so the reviewer can audit which semantics each call site inherits.

Closes #899

## Test plan

- [x] `GOWORK=off go build ./cmd/kuke/... ./cmd/kukeond/...`
- [x] `GOWORK=off go test ./...` (full suite green; affected packages: `cmd/kuke/internal/lifecycle`, `cmd/kuke/init`, `cmd/kuke/daemon/{recreate,reset,restart,start,stop,kill,logs}`)
- [ ] `make dev-init` regression-guard tail (deferred — requires standalone containerd at `/run/containerd/containerd.sock` which is not available in this dev environment; refactor preserves call-site semantics — production code paths and arguments at the two callers are equivalent to the pre-refactor inlined copies)


eminwux:
## Review

**Verdict: LGTM ✓**

### Summary

Pure refactor that correctly resolves a Go `internal/` visibility problem: `cmd/kuke/init/` couldn't reach `cmd/kuke/daemon/internal/lifecycle/` across the tree boundary, so the package is relocated to `cmd/kuke/internal/lifecycle/` and the two drifted inline copies are collapsed into a single exported surface.

### Correctness

- **Package relocation is complete.** `cmd/kuke/daemon/internal/lifecycle/` is fully gone on the PR branch; all ~14 import paths updated. No stale reference to the old path survives (`git grep` clean).
- **Semantic divergence resolved correctly.** The PR author flagged that the two prior `pingKukeond` copies had drifted: `init.go` did dial → Ping; `recreate.go` did Ping-only. The consolidated `PingKukeond` keeps the dial-first shape (the safer of the two — guards the window where the socket file exists but the gRPC handler isn't installed yet, exactly per the `// dial AND ping` comment in the old `init.go`). `recreate.go` now inherits the stricter check, which is the correct direction.
- **Call sites.** Both `lifecycle.WaitForKukeondReady` call sites pass `lifecycle.KukeondReadyTimeout` (not a hardcoded literal), so the shared-timeout invariant the PR aims for is honoured.
- **License header.** `ready.go` carries the Apache 2.0 + SPDX header. ✓
- **Cross-PR interaction.** Five commits landed on `main` since this PR's base (`706fe9b`): four docs + `fix(daemon): re-assert FORWARD admission chain`. No file overlap with the changed `cmd/kuke/...` paths; test-merge is clean.
- **CI.** All five checks green; single verified commit.

### Non-blocking nit

`ready.go` introduces two new exported functions (`WaitForKukeondReady`, `PingKukeond`) and two constants, but there is no `ready_test.go`. The existing `lifecycle_test.go` (renamed from the old path) covers `lifecycle.go` only. The functions existed inline before without dedicated unit tests, so this isn't a regression — but a `ready_test.go` with a mock-socket harness (similar to the pattern in `lifecycle_test.go`) would complete the colocated-test convention from CLAUDE.md. Worth a follow-up.

### Test plan note

The `make dev-init` item is marked deferred in the test plan. The change is confined to the CLI-side polling path (`cmd/kuke/...`) and does not touch daemon code, the build path, or anything under `/opt/kukeon`, so the hard requirement in CLAUDE.md doesn't apply here. Acknowledged.

---

Ready to merge.

Issue #953, opened 2026-05-30
[umbrella] Reconcile Space network state (CNI + FORWARD admission + egress) in the daemon loop


eminwux:
## Summary

A host reboot took every cell in a space offline: `kuke run <cell>` failed with `attach ping timed out: kuketty control socket bound but not yet serving`. Root cause was **lost host firewall state** — the `KUKEON-FORWARD` admission chain (and the FORWARD jump) is installed **only once, at `kuke init`** (`cmd/kuke/init/init.go:322`), lives in the kernel's iptables, and is wiped by a reboot. With Docker present (`-P FORWARD DROP`), cell egress was silently dropped → in-cell DNS to `1.1.1.1` timed out → kuketty's boot-time required-repo `git fetch` stalled ~20s → past the client's 10s attach retry budget (`cmd/kuke/run/attach.go`) → run failed; kuketty then exited with no client and the cell tore down.

Confirmed on the affected host: `KUKEON-FORWARD` and `KUKEON-EGRESS` both absent; manually re-adding a FORWARD ACCEPT for the bridge restored DNS (rc 0) and `kuke run` attached.

This umbrella tracks making Space **network state** (CNI bridge + host FORWARD admission + per-space egress) a **continuously reconciled, daemon-owned** desired-state derived from `space.Spec.Network`, instead of one-shot imperative install — and doing so in a way that coexists with Docker and is fail-closed for egress.

## Three coupled gaps

| # | Gap | Evidence |
|---|-----|----------|
| **A** | Firewall state is installed **once, imperatively, at `kuke init`** and never re-asserted. The daemon loop reconciles only cells (`reconcileFn = core.ReconcileCells`, `internal/daemon/server.go:99`); there is **no periodic space reconcile**. | After reboot both kukeon chains were gone and nothing re-installed them. |
| **B** | kukeon inserts its jump **directly into the builtin `FORWARD` chain** (`ensureForwardJump`, `internal/firewall/forward.go:272`), which it co-owns with Docker. | Docker sets `-P FORWARD DROP` and re-asserts `DOCKER-USER`/`DOCKER-FORWARD` on every daemon restart and `docker network create`, racing kukeon's positional insert. |
[...]

PR #954, opened 2026-05-30
fix(daemon): re-assert FORWARD admission chain on Serve startup

eminwux:
## Problem

A host reboot flushes the kernel's iptables, stripping the `KUKEON-FORWARD` admission chain and its `FORWARD` jump. That chain is installed in **exactly one place** today — `kuke init` (`cmd/kuke/init/init.go:322`), the only caller of `firewall.Install`. Nothing re-asserts it on daemon startup or in the reconcile loop.

On a host whose `FORWARD` default policy is `DROP` (anywhere Docker is installed — Docker sets `-P FORWARD DROP`), the result after a reboot is that **every cell loses egress**:

1. No `FORWARD` admission → cell bridge traffic is dropped → in-cell DNS (`1.1.1.1`) times out.
2. kuketty's boot-time required-repo `git fetch` stalls ~20s on the hung DNS.
3. That blows past the client's 10s attach retry budget (`cmd/kuke/run/attach.go`) → `kuke run <cell>` fails with `attach ping timed out: kuketty control socket bound but not yet serving`.
4. With no client attached, kuketty exits and the cell tears down.

Diagnosed live: after a reboot both `KUKEON-FORWARD` and `KUKEON-EGRESS` were absent; manually re-adding a `FORWARD` ACCEPT for the bridge restored in-cell DNS (rc 0) and `kuke run` attached.

## Fix

Re-assert the admission chain once on **daemon `Serve` startup**. The daemon restarts on its own after a reboot, so this self-heals the host without a re-run of `kuke init`.

- **Idempotent** — `firewall.Install` does `-C` before every `-I/-A` (`internal/firewall/forward.go:154`), so on a healthy host it produces no rule churn.
- **Best-effort** — a missing `iptables` binary is a WARN-and-skip (`IsIptablesAvailable` guard), and a transient `iptables` error is logged and swallowed so it never blocks the daemon from serving.
- **Works from the daemon** — kukeond already manipulates host iptables in host scope (it adds the CNI `nat/POSTROUTING` MASQUERADE during cell start; #96), so a startup `Install` lands on the host.
- Wired via a `forwardAdmissionFn` seam mirroring the existing `reconcileFn` pattern, so the test exercises the startup hook without touching iptables.

## Scope

Intentionally limited to host-global **admission** — the reboot case that takes cells offline. Per-space **egress** re-assertion (currently fail-open) and Docker-churn ordering (anchoring from `DOCKER-USER`) are the larger refactor tracked in the #953 umbrella; this is its P1 down-payment.

## Test plan

- [x] `GOWORK=off go test ./internal/daemon/ ./internal/firewall/` — pass (daemon builds against `go.mod`, not the kuketty `go.work`)
- [x] `gofmt` / `go vet` clean
- [x] New `TestServeReassertsForwardAdmissionOnStartup` asserts the hook runs on `Serve` and that a hook error does **not** abort startup (best-effort contract)
- [x] `make dev-init` — ran end-to-end (daemon startup code change). Daemon-parity tail byte-identical across `kuke get realms` and `--no-daemon` (`default`/`kuke-system` both `Ready`); the PTY `kuke attach` smoke against the kuketty-wrapped cell exited cleanly and the nested parent-attach plumbing stayed intact. This exercises the new `Serve`-startup `firewall.Install` re-assert on a healthy host (idempotent `-C`-before-`-I/-A`, no rule churn).

Refs #953

### DIFF
--- a/internal/daemon/server.go
+++ b/internal/daemon/server.go
@@ -153,15 +153,15 @@ func (s *Server) Serve() error {
 	}
 
 	s.logger.InfoContext(s.ctx, "kukeond listening", "socket", s.opts.SocketPath)
-	if s.forwardAdmissionFn != nil {
-		if err := s.forwardAdmissionFn(); err != nil {
-			// Best-effort: a flushed/unavailable FORWARD admission chain must
-			// not block the daemon from serving. The next reboot self-heals
-			// here; durable per-space egress + Docker-churn handling is #953.
-			s.logger.WarnContext(s.ctx,
-				"forward admission re-assert failed on startup; continuing",
-				"error", err)
-		}
+	// NewServer always wires forwardAdmissionFn; call it unguarded to match the
+	// reconcileFn pattern (runReconcileOnce dereferences s.reconcileFn directly).
+	if err := s.forwardAdmissionFn(); err != nil {
+		// Best-effort: a flushed/unavailable FORWARD admission chain must
+		// not block the daemon from serving. The next reboot self-heals
+		// here; durable per-space egress + Docker-churn handling is #953.
+		s.logger.WarnContext(s.ctx,
+			"forward admission re-assert failed on startup; continuing",
+			"error", err)
 	}
 	s.startReconcileLoop()
 	s.acceptLoop(listener)
--- a/internal/daemon/server.go
+++ b/internal/daemon/server.go
@@ -34,6 +34,7 @@ import (
 
 	"github.com/eminwux/kukeon/internal/client/local"
 	"github.com/eminwux/kukeon/internal/controller"
+	"github.com/eminwux/kukeon/internal/firewall"
 	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
 )
 
@@ -72,6 +73,11 @@ type Server struct {
 	// it before Serve so they exercise the ticker without a real controller.
 	reconcileFn func() (controller.ReconcileResult, error)
 
+	// forwardAdmissionFn re-asserts the host FORWARD admission chain once on
+	// Serve startup. Defaults to reassertForwardAdmission; tests overwrite it
+	// before Serve so they exercise the startup hook without touching iptables.
+	forwardAdmissionFn func() error
+
 	// stopCh is closed by Stop to terminate the reconcile loop independently
 	// of s.ctx; loopWG lets Stop block until the loop exits, so core.Close
 	// never races an in-flight reconcile pass.
@@ -97,6 +103,7 @@ func NewServer(ctx context.Context, logger *slog.Logger, opts Options) *Server {
 		stopCh: make(chan struct{}),
 	}
 	srv.reconcileFn = srv.core.ReconcileCells
+	srv.forwardAdmissionFn = srv.reassertForwardAdmission
 	return srv
 }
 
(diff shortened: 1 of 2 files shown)


eminwux:
## Review

**fix(daemon): re-assert FORWARD admission chain on Serve startup**

The fix is correct and the diagnosis is solid. A host reboot flushes the kernel's iptables; the daemon restarts on its own but `firewall.Install` was only wired into `kuke init`, so the `KUKEON-FORWARD` chain was never re-asserted. On any host where Docker (or anything else) has set `FORWARD DROP`, every cell loses egress silently until the operator re-runs `kuke init`. Wiring the re-assert into `Serve` is exactly the right fix — the daemon is the thing that restarts.

The implementation is clean:
- The `forwardAdmissionFn` seam mirrors the existing `reconcileFn` pattern exactly, keeping the new code testable without touching the host's iptables.
- Best-effort contract (WARN-log + continue on error) is correct for startup — a missing or transiently broken iptables binary must not prevent the daemon from serving.
- `firewall.Install` is idempotent (`-C` before every `-I/-A`), so re-running on a healthy host produces no rule churn.
- The test covers both contracts: the hook fires on startup AND an error from it does not abort `Serve`.

---

### Blocking

**Missing `make dev-init` in the test plan.**

`CLAUDE.md` is explicit:

> When a change touches the build path, the daemon, or anything under `/opt/kukeon`, run this end-to-end before opening a PR.
> **If your change touches anything the daemon reads, this check must be in the PR's test plan. Reviewer agent will flag PRs that miss it.**

This PR modifies daemon startup code (`internal/daemon/server.go`). The test plan lists only unit tests (`go test ./internal/daemon/ ./internal/firewall/`) and `go vet`. Please add a `make dev-init` entry (with a note that you ran it) to the test plan checklist before re-requesting review.

---

### Nit (non-blocking)

`Serve` guards the call with `if s.forwardAdmissionFn != nil`, but `NewServer` always assigns it to `reassertForwardAdmission`, and there is no code path that sets it to nil after construction. The existing `reconcileFn` field has no such nil guard in its own call site (`runReconcileOnce` calls `s.reconcileFn()` directly). The defensive check is harmless but creates a minor inconsistency — worth a one-liner comment if you intend nil as a supported "disable the hook" signal, or just drop the guard to match the `reconcileFn` pattern.

---

**Verdict: changes-requested** — one blocking item above. Happy to approve once `make dev-init` is in the test plan.

eminwux:
Addressed in b6621ed (+ test-plan update):

**Blocking — `make dev-init` test plan entry.** Ran `make dev-init` end-to-end and added it to the test plan checklist. The daemon-parity tail was byte-identical across `kuke get realms` and `--no-daemon` (`default`/`kuke-system` both `Ready`), the PTY `kuke attach` smoke against the kuketty-wrapped cell exited cleanly, and the nested parent-attach plumbing stayed intact. Since this PR's change is the `Serve`-startup `firewall.Install` re-assert, the run also confirmed the idempotent (`-C`-before-`-I/-A`) path produces no rule churn on a healthy host.

**Nit — nil guard on `forwardAdmissionFn`.** Dropped the `if s.forwardAdmissionFn != nil` guard so `Serve` dereferences it directly, matching the sibling `reconcileFn` seam (`runReconcileOnce` calls `s.reconcileFn()` unguarded). Added a one-liner noting the `NewServer`-always-wires invariant. `TestServeReassertsForwardAdmissionOnStartup` still passes.

`gofmt` / `go vet` clean; `GOWORK=off go test ./internal/daemon/ ./internal/firewall/` passes.

eminwux:
## Review (follow-up)

Both items from the prior review are addressed:

**Blocking → resolved.** `make dev-init` is now in the test plan with a clear note confirming it was run (idempotent `-C`-before-`-I/-A` on a healthy host, daemon-parity tail byte-identical, PTY attach smoke clean).

**Nit → resolved.** The spurious nil guard on `forwardAdmissionFn` is gone; the call site now reads identically to the `reconcileFn` pattern, and the short comment explains why the unguarded call is safe (`NewServer` always wires the field).

No further changes needed.

**LGTM** — marking ready to merge.

PR #956, opened 2026-05-30
test(lifecycle): cover WaitForKukeondReady and PingKukeond

### LINKED ISSUE #955: followup-nit(lifecycle): add ready_test.go for WaitForKukeondReady / PingKukeond
**Origin:** PR #952 review

`cmd/kuke/internal/lifecycle/ready.go` exports `WaitForKukeondReady`, `PingKukeond`, `KukeondReadyTimeout`, and `KukeondReadyTick`, but has no colocated `ready_test.go`.

The existing `lifecycle_test.go` covers `lifecycle.go` only. A `ready_test.go` with a mock-socket harness would complete the colocated-test convention from CLAUDE.md for the new public API surface.

eminwux:
## Summary

- Adds `cmd/kuke/internal/lifecycle/ready_test.go` to close the colocated-test gap CLAUDE.md prescribes for the public surface added in #952 — `WaitForKukeondReady`, `PingKukeond`, `KukeondReadyTimeout`, `KukeondReadyTick`.
- Mock-socket harness mirrors the pattern already used in `internal/controller/runner/setup_status_client_test.go`: a stub service registers `Ping` under `kukeonv1.ServiceName` and serves over `jsonrpc.NewServerCodec`, so the production `NewUnixClient` round-trips against the exact wire protocol the daemon ships.
- Pins both gates of `PingKukeond` separately: missing socket → `dial:` prefix; accept-and-close listener (the listener-bound-but-handler-not-installed window that motivates the helper) → `ping:` prefix.
- `WaitForKukeondReady` cases cover first-tick success, the realistic "becomes ready partway through polling" race, deadline-expiry with the last attempt error preserved, and `context.Canceled` propagation. Constants are pinned at 30s / 200ms so silent drift breaks the test, not the bring-up trade-off.

## Test plan

- [x] `GOWORK=off go test ./cmd/kuke/internal/lifecycle/ -v` — all 8 new cases pass, full lifecycle suite green
- [x] `GOWORK=off go vet ./cmd/kuke/internal/lifecycle/` — clean

Closes #955

### DIFF
--- a/cmd/kuke/internal/lifecycle/ready_test.go
+++ b/cmd/kuke/internal/lifecycle/ready_test.go
@@ -1,0 +1,264 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package lifecycle_test
+
+import (
+	"context"
+	"errors"
+	"net"
+	"net/rpc"
+	"net/rpc/jsonrpc"
+	"path/filepath"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/eminwux/kukeon/cmd/kuke/internal/lifecycle"
+	"github.com/eminwux/kukeon/pkg/api/kukeonv1"
+)
+
+// stubPingService mirrors the daemon's KukeonV1Service.Ping handler: a
+// net/rpc receiver registered under kukeonv1.ServiceName whose Ping method
+// returns OK. Served over the stdlib jsonrpc codec so PingKukeond's
+// production client speaks the same wire protocol the daemon serves over,
+// without standing up a full controller.
+type stubPingService struct {
+	calls atomic.Int32
+}
+
+func (s *stubPingService) Ping(_ *kukeonv1.PingArgs, reply *kukeonv1.PingReply) error {
+	s.calls.Add(1)
+	reply.OK = true
+	reply.Version = "test"
+	return nil
+}
+
+// servePingStub registers svc under kukeonv1.ServiceName on a unix socket
+// inside a temp dir and serves connections with the stdlib jsonrpc server
+// codec until the test ends. Returns the socket path the client should dial.
+func servePingStub(t *testing.T, svc *stubPingService) string {
+	t.Helper()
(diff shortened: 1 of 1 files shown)


eminwux:
## Review: PR #956 — test(lifecycle): cover WaitForKukeondReady and PingKukeond

**Verdict: LGTM ✅**

### Summary

Adds `cmd/kuke/internal/lifecycle/ready_test.go` — 8 test cases covering the full exported surface of `ready.go` (`WaitForKukeondReady`, `PingKukeond`, `KukeondReadyTimeout`, `KukeondReadyTick`). Closes the colocated-test gap CLAUDE.md requires for the public surface introduced in #952.

### Correctness

- **Mock-socket harness** (`servePingStub` / `serveAcceptAndClose`) correctly mirrors the production wire protocol — the stub registers under `kukeonv1.ServiceName` and serves via `jsonrpc.NewServerCodec`, so `PingKukeond`'s production `NewUnixClient` round-trips against the exact codec the daemon ships. ✓
- **`TestPingKukeond_DialError`** correctly pins the `"dial:"` prefix from `PingKukeond`'s first gate (`d.DialContext` on a missing socket). ✓
- **`TestPingKukeond_PingError`** correctly exercises the accept-and-close window — the raw dial succeeds (listener is bound), but the JSON-RPC Ping hits EOF, surfacing the `"ping:"` prefix from the second gate. The two-connection design in `PingKukeond` (raw dial then RPC client) is precisely what makes this path reachable, and the test exercises both hops. ✓
- **`TestWaitForKukeondReady_ContextCanceled`** correctly traces through the production select: when `ctx` is already canceled on entry, `attemptCtx` inherits the cancellation so `PingKukeond` returns quickly, and the `select { case <-ctx.Done(): return ctx.Err() }` arm fires, returning `context.Canceled` unwrapped — matching the `errors.Is` assertion. ✓
- **`TestWaitForKukeondReady_BecomesReady`** uses `2 * KukeondReadyTick` (400ms) as the server-start delay within a 2s budget, ensuring at least one failed Ping attempt before the server is ready. The `wg.Wait()` after `WaitForKukeondReady` returns correctly synchronises teardown of the inner accept goroutine via `t.Cleanup(ln.Close)`. ✓
- **`TestKukeondReadyConstants`** pins both tunables at their designed values — 30s / 200ms — so a silent constant drift breaks the test, not the bring-up trade-off. ✓

### Conventions

- Apache 2.0 + SPDX header present. ✓
- `package lifecycle_test` (black-box package) — correct for an `_test.go` file testing the public API. ✓
- Tests colocated with source (`ready.go` + `ready_test.go`). ✓
- No new error sentinels; no new `errors.New` in test code. ✓
- Signed commit (892d19b3d8aa, GPG verified). ✓

### Cross-PR check

One commit landed on `main` since this PR branched (`53a41eb` — removes CellProfile / `-p` flag). No overlap with `cmd/kuke/internal/lifecycle/`; no semantic interaction. ✓

### Non-blocking nit

`TestWaitForKukeondReady_Timeout` uses `lifecycle.KukeondReadyTimeout` (30s) as the upper bound guard for a 100ms-budget test whose realistic elapsed time is ~200ms (one tick). The guard catches only a full-hang scenario; a tighter bound like `5 * lifecycle.KukeondReadyTick` (1s) would fail faster on any real regression. Worth tightening in a follow-up but not a blocker.

Issue #981, opened 2026-05-31
followup-nit(ctr): tighten test fatalf message in TestBuildRootContainerSpec_PauseBindMount


eminwux:
Followup nit from PR #980 review.

**File:** `internal/ctr/container_utils_test.go:192`
**Current:** `t.Fatalf("no mount for kukepause binary found")`
**Suggested:** `t.Fatalf("no kukepause bind mount found")`

The current wording reads awkwardly ("for … found"). Pre-existing from the `/pause` era, carried forward unchanged by the PR. A one-word tweak removes the dangling "for".

Low priority — no functional impact.

eminwux:
Triaged for implementation, but the premise has **rotted** — the targeted wording no longer exists on current `main`, so there is no edit to make.

Verified against up-to-date `origin/main` (`f6f5955`, the merge commit of #980):

- `grep -rn 'no mount for kukepause binary found'` returns nothing in the working tree or on `origin/main`. The substring `no mount for` does not appear anywhere in the repo.
- `internal/ctr/container_utils_test.go:192` on current `main` is the `func TestBuildRootContainerSpec(t *testing.T) {` declaration — not a `t.Fatalf`.
- The assertion that fires when the kukepause bind mount is absent is now at `internal/ctr/container_utils_test.go:182`:
  ```go
  t.Fatalf("no %q bind mount in spec.Mounts = %+v",
      ctr.RootContainerPauseBinaryTarget, spec.Mounts)
  ```
  This already reads cleanly ("no `<target>` bind mount in spec.Mounts …") with no dangling "for".

Despite the nit's note that #980 "carried [the awkward wording] forward unchanged," #980's diff actually rewrote these `/pause`-era messages into the `kukepause` `%q bind mount` form (visible in `git show f6f5955 -- internal/ctr/container_utils_test.go`). The dangling-"for" phrasing this nit targets is already gone.

Recommend closing as already-resolved / superseded by #980. Releasing the `in-progress` lock; no PR will follow.

PR #982, opened 2026-05-31
feat(kukepause): emit ready line to stderr at startup

eminwux:
**Opened in error — do not merge; do not review.** This PR adds an unrequested kukepause startup-stderr-logging change that does not correspond to any open issue. It was drafted off a misread of issue #981 before that issue's body was read. Issue #981 is an unrelated test-message nit whose premise has since rotted. The `Closes` reference has been removed intentionally so this PR is no longer linked to any issue. Awaiting maintainer to close (agent's `gh pr close` was not permitted in this session).

### DIFF
--- a/cmd/kukepause/main.go
+++ b/cmd/kukepause/main.go
@@ -36,6 +36,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"os/signal"
 	"syscall"
@@ -52,6 +53,14 @@ func main() {
 	// drains every exited child per wake, so a single slot is sufficient.
 	chld := make(chan os.Signal, 1)
 	signal.Notify(chld, syscall.SIGCHLD)
+
+	// Positive startup signal on stderr. With both handlers now armed, this line
+	// doubles as proof main() was reached: an operator debugging a root container
+	// that won't stay up can distinguish "kukepause started and parked" from
+	// "the bind-mounted binary never executed" (wrong path, ENOENT, exec format
+	// error), which silence alone cannot. stderr keeps stdout empty; the stable
+	// "kukepause: ready" substring is the grep anchor.
+	fmt.Fprintf(os.Stderr, "kukepause: ready (pid %d); SIGTERM→exit, SIGCHLD→reap; parking\n", os.Getpid())
 
 	for {
 		select {


eminwux:
Closing as opened-in-error: this PR adds an unrequested kukepause startup-stderr-logging change that does not correspond to issue #981 or any open issue. It was drafted off a misread of the issue before the body was read. Issue #981 is an unrelated test-message nit (and its premise has since rotted). No part of this branch should be merged.

eminwux:
## Review

The actual code change (one `fmt.Fprintf(os.Stderr, …)` between the second `signal.Notify` and the `for { select }` loop) is clean: correct position, correct format spec (`%d` against `os.Getpid()`'s `int`), stderr-only so stdout stays empty per the acceptance criteria, and the comment explains the WHY (distinguishing silent-but-parked from never-executed) without overshooting. The `fmt` import is the only new dependency and the idle-block mechanism is genuinely unchanged. Per-commit signature check is green (1 commit, `verified: true`).

### Blocking

1. **`Closes #981` does not match the work in this PR.** Issue #981 is a `followup-nit` about renaming a `t.Fatalf` message in `internal/ctr/container_utils_test.go:192` (`"no mount for kukepause binary found"` → `"no kukepause bind mount found"`). This PR touches `cmd/kukepause/main.go` only and does not modify that test file at all (the t.Fatalf message there reads `"no %q bind mount in spec.Mounts = %+v"` on both base and head, unchanged by this PR). Merging as-is will auto-close #981 without addressing it. Either drop `Closes #981` from the description (the PR still merges, #981 stays open for a dedicated fix), or fold the one-line test message rename into this PR so the `Closes` is accurate. The former is the cleaner split — this PR is scoped to the diagnostic emit, the nit is scoped to a test message.

### Nits (non-blocking)

None.

### Test plan

The `gofmt`/`build`/`vet`/`go test ./cmd/kukepause/` quartet plus manual stdout-is-0-bytes + stderr substring check is the right scope for a 9-line additive stderr emit. The PR's explicit note that the `make dev-init` daemon-parity tail is unaffected is correct — the smoke checks realm visibility (`kuke get realms -o wide` vs. `--no-daemon`), which a PID-1 stderr line in root containers cannot perturb.

PR #989, opened 2026-05-31
fix(run): enrich divergent-spec error with actual+desired values

### LINKED ISSUE #984: run: divergent-spec rejection still fires after a successful kuke restart cell
## Repro

1. `kuke apply -f mycell-config.yaml`, `kuke run mycell-config` → cell starts.
2. Edit `mycell-config.yaml` (e.g. bump the container image), `kuke apply -f mycell-config.yaml`.
3. Wait for `OutOfSync` to appear in `kuke get cell mycell` (or trigger a reconciler tick).
4. `kuke restart cell mycell` → returns OK; `OutOfSync` clears in `kuke get cell mycell`.
5. `kuke run mycell-config` → **still** rejects with the divergent-spec error and points back to `kuke restart cell mycell`.

Operator-reported on a live host; not yet reproduced in CI.

## Observed

`kuke run <config>` exits non-zero with the `rejectDivergentNamedCell` error (\"spec differs at containers[…]\", \"refusing to attach\") immediately after a successful `kuke restart cell` that, per the OutOfSync reconcile path, **just** wrote a freshly-materialized spec to disk via `ApplyDocuments`.

## Root cause

Not yet pinpointed. The comparison in `cmd/kuke/run/run.go:591` (`divergedFields(pre.Cell.Spec, cellDoc.Spec)`) compares **on-disk Cell.Spec** (from `GetCell`) against a **freshly re-materialized Cell.Spec** (from the materialization pipeline). It does **not** read `Status.OutOfSync`, so it's resilient to flag staleness — meaning the bug isn't a stale flag.

Two plausible candidates the next picker should investigate first:

1. **`cellconfig.Materialize` is non-deterministic between calls** — generated names, ordering, embedded timestamps, or anything else that changes between successive materializations would cause the spec `restart` wrote to disk to not match the spec `run` re-computes on its next call.
2. **`ApplyDocuments`'s persist path normalizes the spec differently than `Materialize` produces** — i.e. `disk(after_ApplyDocuments) ≠ Materialize(call_again)`, even though restart wrote `Materialize(call_one)` through `ApplyDocuments`.

The normalization exclusions in `divergedFields` (`cmd/kuke/run/run.go:1257-1303`) already strip a few runner-synthesized fields, but if a field flips between Materialize calls outside that allowlist, the check fires false-positive.

## Suggested fix

- Add a debug surface that prints the exact diverging fields — either `kuke status cell <name> --explain-divergence` or extend the `rejectDivergentNamedCell` error message to list field paths and the two values. Without this, the next picker can't tell which of the two candidates is firing.
- Add a unit test in `cmd/kuke/run/` (or `internal/controller/`) asserting: `Materialize → ApplyDocuments → GetCell → Materialize → divergedFields(a, b) == nil`. This is the regression guard.
- Once the gap is identified, tighten either `cellconfig.Materialize` (make deterministic) or `divergedFields`'s exclusion set so the comparator agrees with the persistence pipeline.

Cross-links: surfaced alongside #983 (stop+start vs restart asymmetry — once `start` also re-materializes, the same flake will surface from the stop+start path). Related to #753 (introduced refuse-on-divergence in `run`).

eminwux:
## Summary

Closes #984.

`kuke run <config>` rejecting with the divergent-spec error immediately
after a successful `kuke restart cell` (which just wrote a
freshly-materialised spec to disk via the OutOfSync reconcile path) was
diagnostically opaque: the operator saw `spec differs at containers[…]` with
no way to tell which side of `divergedFields(pre.Cell.Spec, cellDoc.Spec)`
normalised differently.

This PR ships the two non-conditional items from the issue's "Suggested fix"
list — debug surface + regression guard — so the gap is observable the next
time it fires and a future PR can target it precisely.

- `divergedFields` / `divergedContainerFields` now embed actual+desired
  values into each returned entry (`spec.containers["main"].image
  (actual="registry.example.com/web:v1", desired="registry.example.com/web:v2")`).
  Path prefix preserved verbatim so existing `strings.Contains` test
  assertions still match.
- New `TestDivergedFields_MaterializePersistMaterializeIsIdempotent` in
  `cmd/kuke/run/` asserts the invariant: Materialize → daemon-persist
  simulation (yaml round-trip + apischeme convert + parent-ref / Space-defaults
  / ContainerdID mutations + JSON marshal/unmarshal + GetCell readback) →
  Materialize → `divergedFields == nil`. Passes on both a minimal cell and a
  rich one (env, ports, volumes, networks, git, repos, secrets, tty); the
  simulation does not reproduce the operator-reported false positive, which
  is consistent with the issue's "not yet reproduced in CI" note. Test is a
  regression guard against the persistence pipeline drifting from the
  comparator on future changes.

The third suggested fix (tighten Materialize determinism or
divergedContainerFields' exclusion set "once the gap is identified") is
deferred — the gap is still not identifiable from the current simulation,
and the enriched error from this PR is what enables the next live-host
encounter to pinpoint it.

## Test plan

- [x] `go build ./... && go vet ./...`
- [x] `go test ./...` (full suite green)
- [x] `go test ./cmd/kuke/run/` — new regression test passes; existing
  divergent-spec tests (`TestRun_FromConfig_DivergentSpec_RefusesAndPointsToApply`,
  `TestRun_FromBlueprint_NamedDivergent_RefusesAndPointsToApply`, the
  per-field secrets/tty/work tests) still match on path prefix
- [ ] `make dev-init` smoke — not applicable; change is CLI-side error
  formatting only, no daemon or `/opt/kukeon` touchpoints

### DIFF
--- a/cmd/kuke/run/divergence_idempotency_test.go
+++ b/cmd/kuke/run/divergence_idempotency_test.go
@@ -1,0 +1,277 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package run
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/eminwux/kukeon/internal/apischeme"
+	"github.com/eminwux/kukeon/internal/cellconfig"
+	intmodel "github.com/eminwux/kukeon/internal/modelhub"
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
+	"gopkg.in/yaml.v3"
+)
+
+// TestDivergedFields_MaterializePersistMaterializeIsIdempotent guards the
+// invariant that `kuke restart cell` writes a spec to disk that — when
+// re-materialised by the next `kuke run <config>` — compares clean against
+// the persisted version (issue #984). The flow being asserted:
+//
+//	Materialize(cfg, bp) → ApplyDocuments(persist) → GetCell → Materialize(cfg, bp)
+//	→ divergedFields(persisted.Spec, freshly-materialised.Spec) == nil
+//
+// The daemon-side ApplyDocuments + GetCell pair is simulated here via the
+// apischeme + JSON pipeline plus the mutations UpdateCell applies (parent
+// refs, ContainerdID, Space-defaults merge). A real daemon test would need a
+// containerd socket and a temp filesystem; the simulation captures the
+// normalisations divergedContainerFields is sensitive to (the Space-defaults
+// fields are deliberately excluded from that comparator, but the simulation
+// applies them anyway to verify the exclusion stays correct under realistic
+// inputs).
+//
+// If this test trips, divergedFields and the persistence pipeline have
+// drifted apart. Tighten either side per the issue's "Suggested fix" — make
+// Materialize deterministic, or grow divergedFields' exclusion set to match
+// a new normalisation introduced on the persist side.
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #989 Review — `fix(run): enrich divergent-spec error with actual+desired values`**

LGTM — error rendering enrichment only (path prefix preserved verbatim, existing `strings.Contains` assertions still match), plus a regression test locking the Materialize ↔ daemon-persist ↔ Materialize idempotency. The test honestly notes it does not reproduce the operator-reported false positive yet; the deferred third "Suggested fix" depends on a live-host pinpoint, which the enriched error now enables.

PR #999, opened 2026-06-01
fix(apply): diff cell-level Tty, AutoDelete, NestedCgroupRuntime fields

eminwux:
## Summary

- `DiffCell` previously only compared `Metadata.Name`, `Metadata.Labels`, `Spec.RealmName`, `Spec.SpaceName`, `Spec.StackName` at the cell level. Three user-authored persisted `CellSpec` fields had no comparator branch: `Tty`, `AutoDelete`, `NestedCgroupRuntime`. Edits to those fields were persisted to cell metadata but `kuke apply` reported "no changes" and never re-stamped the runtime.
- New branches placed between the labels check and the root-container handling:
  - `Spec.Tty` → Compatible (`spec.tty` in `ChangedFields`). Runner re-resolves the attach target lazily.
  - `Spec.AutoDelete` → Compatible (`spec.autoDelete` in `ChangedFields`). Reaper consults the persisted field on the next tick; emitting the diff entry surfaces the change in the apply-layer report.
  - `Spec.NestedCgroupRuntime` → Breaking (`spec.nestedCgroupRuntime` in `BreakingChanges`). Flipping the flag re-runs `EnableCellAllSubtreeControllers` (#318) and recomputes the in-container `/sys/fs/cgroup` mount per `BuildContainerSpec`; in-place toggle is not supported, RecreateCell is the safe path.
- New helper `cellTtyEqual` lives alongside the existing equality helpers (style match with `capabilitiesEqual` / `resourcesEqual`). Nil pointer is treated as equal to a zero-value `CellTty` so a same-file re-apply with one side nil and the other an explicitly-empty block does not register as drift.
- `OutOfSyncReason` on Config-lineage cells now picks up these field names automatically: `outOfSyncSpecDifferReason` / `collectDiffFieldPaths` in `internal/controller/reconcile_outofsync.go` aggregate over `ChangedFields` + `BreakingChanges`, so the new branches surface in the persisted reason string without any reconciler-side change.

## Test plan

- [x] `GOWORK=off go test ./internal/controller/apply/ -run TestDiffCell -v` — all DiffCell tests pass, including the four new ones (`TestDiffCell_Tty_CompatibleChange`, `TestDiffCell_Tty_NilEqualsZero`, `TestDiffCell_AutoDelete_CompatibleChange`, `TestDiffCell_NestedCgroupRuntime_BreakingChange`).
- [x] `GOWORK=off make test` — both root and `cmd/kukebuild` modules clean.
[...]

Issue #992, opened 2026-06-01
reconcile: DiffCell ignores CellSpec.Tty, AutoDelete, NestedCgroupRuntime; edits silently lost on apply


eminwux:
## Context

\`internal/controller/apply/diff.go:269-450\` (\`DiffCell\`) compares cell-level fields before descending into the per-container loop. The cell-level comparator set is:

- \`Metadata.Name\` — Breaking
- \`Metadata.Labels\` — Compatible (via \`userLabelsEqual\`, filtering controller-managed keys per #437)
- \`Spec.RealmName\`, \`Spec.SpaceName\`, \`Spec.StackName\` — Breaking

That's it. The \`CellSpec\` type (\`internal/modelhub/cell.go:36-64\`) carries three additional user-authored persisted fields with **no comparator branch**:

| Field | Line in \`cell.go\` | Semantics |
|-------|--------------------|-----------|
| \`Tty\` | 42 (\`*CellTty\`) | Cell-default TTY (\`CellTty.Default\`) — picks which container's TTY \`kuke attach\` lands on |
| \`AutoDelete\` | 47 | Mirrors v1beta1; reaper deletes the cell once all children stop |
| \`NestedCgroupRuntime\` | 52 | Mirrors v1beta1; delegates the full host-available subtree-controller set (#314, #318, #328) |

(\`RuntimeEnv\` at line 63 is acknowledged not-persisted — the v1beta1 field carries \`yaml:\"-\"\` and the disk-read paths return empty \`RuntimeEnv\`. It's correctly excluded from diff. \`RootContainerID\` at line 41 is a synthesized identifier, not user-authored.)

## Concrete consequences

- **\`Tty.Default\`**: An operator editing the cell-default TTY pointer (so \`kuke attach\` lands on a different container) will see the edit lost on \`kuke apply\`. Cell metadata persists the new value but \`DiffCell\` reports no change, so UpdateCell/RecreateCell never re-stamp anything.
- **\`AutoDelete\`**: Toggling \`autoDelete: true\` after the fact does not surface as a diff. The reaper reads the persisted spec and may eventually behave correctly on the next reconcile pass that consults \`cell.Spec.AutoDelete\` directly, but the apply layer does not flag the change for the operator — the \"no changes\" verdict is misleading.
[...]

PR #1001, opened 2026-06-01
fix(reconcile): widen root-container diff to all spec fields

### LINKED ISSUE #990: reconcile: root-container diff only compares image/command/args; every other root spec field is silently ignored
## Context

`internal/controller/apply/diff.go:929-942` routes the root container through a dedicated comparator:

```go
func rootContainerSpecChanged(desired, actual *intmodel.ContainerSpec) bool {
    return desired.Image != actual.Image ||
        desired.Command != actual.Command ||
        !slicesEqual(desired.Args, actual.Args)
}
```

This is the *only* comparator that runs against the root container. `DiffCell`'s per-container loop (`diff.go:376-388`) filters root containers out (`if !container.Root && container.ID != ""`), so the per-field diff in `diffContainerSpec` — which covers env, ports, volumes, privileged, user, readOnlyRootFilesystem, capabilities, securityOpts, tmpfs, resources, secrets, repos — never executes on the root container.

The `ContainerSpec` type (`internal/modelhub/container.go:32-102`) carries ~25 user-authored fields. The root container's diff path checks 3 of them.

## Concrete consequences

An operator editing `spec.containers[<root>]` on a Config-lineage cell and running `kuke restart cell` (or the next OutOfSync reconcile tick) will get a clean "in sync" result on every change *except* image/command/args. Silently lost edits include, among others:

- `Env` — environment variable changes on the root container
- `Volumes` — bind/tmpfs mount changes
- `Privileged`, `User`, `ReadOnlyRootFilesystem`, `Capabilities`, `SecurityOpts` — security posture changes
- `Resources` — cgroup limits
- `Secrets` — credential references

Non-root containers do detect these via `diffContainerSpec` (`diff.go:534-670`). Root containers don't.

This is separately impactful from the per-field gaps in `diffContainerSpec` itself (filed as a sibling issue) because closing the per-field gaps there does **not** fix the root container — the root container needs its own routing decision (call `diffContainerSpec` with `rootContainer=true` so image/command/args still classify as Breaking, but every other field gets compared and routed through Breaking/Compatible appropriately).

## Proposal

In `DiffCell`, after handling root-container presence add/remove (`diff.go:353-379`), call `diffContainerSpec(desiredRoot, actualRoot, true)` and merge its result into the cell diff with appropriate `ChangeType` classification.

For each currently-Compatible field on non-root containers, decide its classification on a root container:

- **Definitely Breaking on root** (cannot in-place update because the root container's OCI spec is baked into cell-namespace setup): `WorkingDir`, host-namespace toggles, `Privileged`, `User`, `ReadOnlyRootFilesystem`, `Capabilities`, `Tmpfs`, `Resources` cgroup-shape changes.
- **Probably Compatible on root** (kuketty-side or post-start re-evaluable): `Env`, `Secrets` (env mode), `Repos`.

The split needs a design decision — the current code has none, treating "image/command/args" as a proxy for "all breaking root changes" without enumerating the actual breaking set.

## Acceptance criteria

- [ ] `rootContainerSpecChanged` is either removed or expanded to cover every non-runtime-only `ContainerSpec` field.
- [ ] `DiffCell` detects spec drift on `Env`, `Volumes`, `Privileged`, `User`, `ReadOnlyRootFilesystem`, `Capabilities`, `SecurityOpts`, `Resources`, `Secrets`, `Repos`, `Tmpfs` for the root container.
- [ ] Each newly-detected root-container field has an explicit Breaking-vs-Compatible classification documented inline near the diff site.
- [ ] `OutOfSyncReason` populates with the changed field name when a Config-lineage cell's root container drifts from the materialized desired spec.
- [ ] Test coverage in `internal/controller/apply/diff_test.go` for at least one field per classification bucket (one new Breaking root field, one new Compatible root field).

## Out of scope

- Runtime-only fields stamped by the runner (`CellCgroupPath`, `EtcHostsPath`, `EtcHostnamePath`) — those are correctly never persisted and never user-authored, so they're outside the diff contract.
- The companion gap in `diffContainerSpec` itself (missing `WorkingDir`, host namespaces, networking) — filed as a separate sibling issue so the two can be sequenced independently.
- The cell-level `Spec.Tty` / `Spec.AutoDelete` / `Spec.NestedCgroupRuntime` gaps — filed separately.

## References

- `internal/controller/apply/diff.go:269-450` — `DiffCell` cell-level + container-loop
- `internal/controller/apply/diff.go:534-670` — `diffContainerSpec` (the comparator root containers never reach)
- `internal/controller/apply/diff.go:929-942` — `rootContainerSpecChanged` (current 3-field check)
- `internal/modelhub/container.go:32-102` — `ContainerSpec` field inventory
- `internal/controller/runner/spec_hash.go:45-50` — spec-hash sibling guard, also limited to image/command/args
- Related landed work: #485 (in-place update for non-root image/command/args), #820/#867 (OutOfSync foundation, spec-hash labeling)
- Epic: epic:reconcile

eminwux:
## Summary

- Closes #990. Before this, `DiffCell`'s root-container path routed through a 3-field `rootContainerSpecChanged(image, command, args)` comparator and silently dropped every other user-authored edit — Env, Volumes, Privileged, User, ReadOnlyRootFilesystem, Capabilities, SecurityOpts, Tmpfs, Resources, Secrets, Repos. Operators editing `spec.containers[<root>]` on a Config-lineage cell got a clean "in sync" verdict on every change except image/command/args.
- Replaces the 3-field comparator with `diffContainerSpec(rootContainer=true)`, threading per-field Breaking-vs-Compatible classification through the same comparator non-root containers use. Each new field call site carries an inline classification comment (AC #3).
- Qualifies each root-container divergence with a `rootContainer.<field>` prefix on `BreakingChanges` / `ChangedFields`, so `OutOfSyncReason` populates with the changed field name on a Config-lineage cell whose root drifts from materialized spec (AC #4) and distinguishes root drift from same-named drift on a child container (`containers[<id>].<field>`).
- Widens `containerSpecHashPayload` to include the new Breaking-on-root fields (`privileged`, `user`, `readOnlyRootFilesystem`, `capabilities`, `tmpfs`, `resources`). `TestSpecHashDomainPinsToDiffCellBreakingFields` is the regression guard that keeps the spec-hash domain in lockstep with the apply layer (issue #867); the test's own comment anticipates this update.
- Tightens the reconciler gate at `internal/controller/apply/reconcile.go` so `RecreateCell` only fires on `RootContainerChanged && ChangeType == Breaking` — matching the spec-hash test's existing gate. Without the AND, a Compatible-on-root edit (env, ports, volumes, securityOpts, secrets, repos) toggles `RootContainerChanged=true` for diff-readout purposes but would have triggered a full kill-and-recreate, contradicting the PR's stated Compatible-on-root semantics. New `TestReconcileCell_RootEnvCompatible_RoutesToUpdate` pins env-on-root to the UpdateCell branch; `TestReconcileCell_RootImageBreaking_RoutesToRecreate` pins image-on-root to RecreateCell so the AND doesn't regress the Breaking path.

### Design call: Breaking vs Compatible on root

The issue body proposes a split but explicitly notes "the split needs a design decision". The lower-blast-radius classification I shipped:

- **Breaking on root** (cannot in-place update — baked into the cell's OCI runtime spec at StartCell): `image`, `command`, `args`, `privileged`, `user`, `readOnlyRootFilesystem`, `capabilities`, `tmpfs`, `resources`.
- **Compatible on root** (re-evaluated at start or rebuilt from spec): `env`, `ports`, `volumes`, `securityOpts`, `secrets`, `repos`.

The issue's proposal listed `volumes` and `securityOpts` only in the "silently lost edits" set without explicitly classifying them; this PR keeps them Compatible on root for now (drift is at least detected, recreate cost avoided). If runner-side UpdateCell gaps surface on `volumes`/`securityOpts` for root containers, a follow-up can widen the Breaking-on-root set without further changes to the apply layer.

### Out of scope (per the issue body)

- Runtime-only fields stamped by the runner (`CellCgroupPath`, `EtcHostsPath`, `EtcHostnamePath`).
- Companion `diffContainerSpec` gaps for `WorkingDir` / host namespaces / networking (sibling issue).
- Cell-level `Spec.Tty` / `Spec.AutoDelete` / `Spec.NestedCgroupRuntime` (PR #999, issue #992).

## Test plan

- [x] `GOWORK=off GOTOOLCHAIN=go1.24.5 go test $(go list ./... | grep -v /e2e)` — full project test target, all packages pass.
- [x] `GOWORK=off GOTOOLCHAIN=go1.24.5 go vet ./...` — clean.
- [x] `pre-commit run --files <changed>` — go-fmt, go-mod-tidy, golangci-lint, addlicense all pass.
- [x] New `TestDiffCell_RootContainerPrivileged_Breaking` (Breaking-on-root field), `TestDiffCell_RootContainerEnv_Compatible` (Compatible-on-root field), `TestDiffCell_RootContainer_NoDriftOnEqualSpec` (no false-positive regression).
- [x] Updated `TestSpecHashDomainPinsToDiffCellBreakingFields` and `TestComputeContainerSpecHash_{BreakingFieldsChangeHash,CompatibleFieldsDoNotChangeHash}` to reflect the widened Breaking-on-root domain.
- [x] New `TestReconcileCell_RootEnvCompatible_RoutesToUpdate` (asserts RecreateCell is *not* called for env-on-root, UpdateCell *is*) and `TestReconcileCell_RootImageBreaking_RoutesToRecreate` (asserts the Breaking path stays on RecreateCell).
- [ ] `make dev-init` smoke — not executed; no daemon code path touched (this is a pure apply-layer + spec-hash change), no `/opt/kukeon` or runner-build path change. The unit tests above pin the behavior end-to-end against the persisted-status consumer (`reconcileCellOutOfSync` via the existing OutOfSync test suite — all 8 still pass).

Closes #990

### DIFF
--- a/internal/controller/apply/diff.go
+++ b/internal/controller/apply/diff.go
@@ -340,15 +340,45 @@ func DiffCell(desired, actual intmodel.Cell) CellDiffResult {
 	desiredRoot := findRootContainer(desired.Spec.Containers)
 	actualRoot := findRootContainer(actual.Spec.Containers)
 
-	// Check if root container spec changed (breaking)
+	// Diff the root container spec field-by-field. Routing through
+	// `diffContainerSpec(rootContainer=true)` closes issue #990: prior to
+	// this, the root path checked only image/command/args via a dedicated
+	// 3-field comparator and silently dropped every other user-authored
+	// edit (Env, Volumes, Privileged, User, ReadOnlyRootFilesystem,
+	// Capabilities, SecurityOpts, Tmpfs, Resources, Secrets, Repos).
+	// `diffContainerSpec` carries per-field Breaking-vs-Compatible
+	// classification keyed off the `rootContainer` flag — see the inline
+	// classification comments at each field call site.
 	switch {
 	case desiredRoot != nil && actualRoot != nil:
-		if rootContainerSpecChanged(desiredRoot, actualRoot) {
+		rootDiff := diffContainerSpec(desiredRoot, actualRoot, true)
+		if rootDiff.HasChanges {
 			result.HasChanges = true
-			result.ChangeType = ChangeTypeBreaking
 			result.RootContainerChanged = true
-			result.BreakingChanges = append(result.BreakingChanges, "spec.rootContainer")
-			result.RootContainerDetails["rootContainer"] = "root container spec changed (image, command, or args)"
+			if rootDiff.ChangeType == ChangeTypeBreaking {
+				result.ChangeType = ChangeTypeBreaking
+			} else if result.ChangeType == ChangeTypeNone {
+				result.ChangeType = rootDiff.ChangeType
+			}
+			// Qualify each diverged field with the `rootContainer.` prefix
+			// so the cell-level readout and `OutOfSyncReason` distinguish
+			// root-container drift from same-named drift on a child
+			// container (the child path uses `containers[<id>].<field>`).
+			for _, field := range rootDiff.BreakingChanges {
+				result.BreakingChanges = append(
+					result.BreakingChanges,
+					"rootContainer."+field,
+				)
+			}
+			for _, field := range rootDiff.ChangedFields {
+				result.ChangedFields = append(
+					result.ChangedFields,
+					"rootContainer."+field,
+				)
+			}
+			for k, v := range rootDiff.Details {
+				result.RootContainerDetails[k] = v
+			}
 		}
 	case desiredRoot != nil && actualRoot == nil:
 		// Root container added
@@ -405,8 +435,9 @@ func DiffCell(desired, actual intmodel.Cell) CellDiffResult {
 			// are classified as in-place updateable (Compatible) here so
 			// UpdateCell stops, removes, recreates, and starts the affected
 			// child container instead of refusing the diff. The root
-			// container's image/command/args bypass diffContainerSpec via
-			// rootContainerSpecChanged above and stay on the recreate path.
+			// container's image/command/args are classified as Breaking by
+			// the same comparator (called from the root branch above with
+			// `rootContainer=true`) and stay on the RecreateCell path.
 			containerDiff := diffContainerSpec(desiredContainer, actualContainer, false)
 			if containerDiff.HasChanges {
 				result.HasChanges = true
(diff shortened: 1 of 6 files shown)


eminwux:
**PR #1001 Review — `fix(reconcile): widen root-container diff to all spec fields`**

Solid issue-#990 fix on the DiffCell side — the per-field comparator + inline Breaking-vs-Compatible classification is the right shape, and the spec-hash widening is correctly pinned by `TestSpecHashDomainPinsToDiffCellBreakingFields`. One blocker on the reconcile-pipeline interaction:

- **Correctness** `internal/controller/apply/reconcile.go:373`: `RootContainerChanged` is now set for Compatible-on-root edits (env, ports, volumes, securityOpts, secrets, repos), but `ReconcileCell` still gates `RecreateCell` unconditionally on that flag — so `kuke apply -f` of an env edit on the root container now triggers a full kill-and-recreate. This contradicts both the PR's stated Compatible-on-root semantics ("re-evaluated at start or rebuilt from spec") and the `TestDiffCell_RootContainerEnv_Compatible` docstring ("not forced through RecreateCell"). The companion test at `internal/controller/runner/spec_hash_test.go:252` already uses the correct gate (`diff.RootContainerChanged && diff.ChangeType == apply.ChangeTypeBreaking`); apply the same AND to the reconciler check. The existing `diff_test.go` coverage only pins the diff result, not the reconcile branch decision — add a ReconcileCell test that asserts `RecreateCell` is *not* called for an env-on-root edit (and routes through `UpdateCell` instead).

LGTM once the reconciler gate distinguishes Breaking from Compatible on root and the new test covers it.

eminwux:
Addressed in 8eb0800:

- **Correctness — `internal/controller/apply/reconcile.go:373`**: tightened the gate to `diff.RootContainerChanged && diff.ChangeType == apply.ChangeTypeBreaking`, matching the existing pattern at `internal/controller/runner/spec_hash_test.go:252`. Compatible-on-root edits (env, ports, volumes, securityOpts, secrets, repos) now fall through to `UpdateCell` rather than triggering a full kill-and-recreate. Added an inline comment naming the spec-hash companion gate so the lockstep is auditable.
- **New test — `TestReconcileCell_RootEnvCompatible_RoutesToUpdate`** (`internal/controller/apply/reconcile_test.go`): asserts an env-on-root edit routes through `UpdateCell` and *not* `RecreateCell`, via a minimal interface-embedding fake runner that nil-derefs on any unexpected method call (the fake's strictness is intentional — it pins the negative assertion structurally, not just by checking a flag).
- **New test — `TestReconcileCell_RootImageBreaking_RoutesToRecreate`**: positive counterpart, pins the Breaking path so the new AND gate doesn't regress image/command/args/privileged/etc.

Full test target (`go test $(go list ./... | grep -v /e2e)`), `go vet`, and `pre-commit` all clean post-rebase.

eminwux:
**PR #1001 Review — `fix(reconcile): widen root-container diff to all spec fields`**

The widened root-container diff cleanly closes #990: per-field Breaking-vs-Compatible classification on the root path, qualified `rootContainer.<field>` surfacing, and the reconciler-gate AND keep RecreateCell honest. Test coverage on both the diff side and the reconcile-routing side is solid; the spec-hash domain is updated in lockstep with the apply-layer Breaking-on-root set.

A few non-blocking nits:

- **Redundant OutOfSyncReason entry** `internal/controller/reconcile_outofsync.go:262`: `collectDiffFieldPaths` still adds `"spec.rootContainer"` unconditionally when `RootContainerChanged` is true, but the new field-level entries (`rootContainer.privileged`, etc.) already populate `BreakingChanges`/`ChangedFields`. Result: `OutOfSyncReason = "spec differs at rootContainer.privileged, spec.rootContainer"` — trailing marker is now noise. Dropping that `add("spec.rootContainer")` line works (the root-added branch in `DiffCell` already appends `spec.rootContainer` to `BreakingChanges` directly, so it still surfaces in the add case).
- **Dangling doc comment** `internal/controller/apply/diff.go:561-563`: the block above `diffContainerSpec` ends with the sentence-fragment "which needs a per-field diff branch..." — its antecedent was the removed `//nolint:funlen` line. The trailing `//` plus blank line also detaches the doc from the function. Drop the orphan `//`/blank pair and rephrase or delete the "which needs..." sentence.

LGTM — both are non-blocking.

PR #1004, opened 2026-06-01
refactor(api): rename CellProfileParameter to CellBlueprintParameter

### LINKED ISSUE #986: api: rename CellProfileParameter to CellBlueprintParameter
## Context

`pkg/api/model/v1beta1/cellblueprint.go:82` still defines the parameter struct as `CellProfileParameter`, even though the `CellProfile` *kind* was removed in #626 and the struct is now blueprint-only. The struct is the parameter type for `CellBlueprintSpec.Parameters` (`cellblueprint.go:68`) — so every reader who hits the Blueprint type sees a `CellProfile`-prefixed field type for a kind that no longer exists.

The mismatch is already acknowledged in the type comment (`cellblueprint.go:78-81`):

> "The name keeps its 'CellProfile' prefix from before #626 removed the CellProfile kind — the struct is now blueprint-only but renaming it would break sibling projects (sbsh, sbcrew, …) that import this package. The type's role moved; the wire identifier stayed put."

This is a permanent name-vs-role tax on every future reader of the Blueprint type. The JSON/YAML wire shape (`{name, description, default, required}`) has no kind discriminator and is referenced positionally as `parameters[]` under `CellBlueprintSpec` — so the rename is Go-side only and does not change the on-wire schema.

In-repo footprint (`grep CellProfileParameter` across `*.go`): 17 references in 8 files —
- `pkg/api/model/v1beta1/cellblueprint.go` (declaration + 1 prose mention)
- `internal/cellblueprint/cellblueprint.go`, `_test.go`
- `internal/cellconfig/materialize_test.go`
- `cmd/kuke/run/run_test.go`
- `cmd/kuke/create/cell/cell_test.go`
- `cmd/kuke/create/config/config.go`, `config_test.go`

## Proposal

Rename the Go identifier `CellProfileParameter` → `CellBlueprintParameter` in `pkg/api/model/v1beta1/cellblueprint.go` and update every in-repo call site (controller, materializer, parser, tests, `cmd/kuke/create/config/config.go:writeValues`).

Two migration options for the sibling-project (sbsh, sbcrew) breakage concern called out in the type comment — pick one in the PR:

- **(a) One-release deprecated alias.** Keep `type CellProfileParameter = CellBlueprintParameter` as a `// Deprecated:` alias for one release so sbsh/sbcrew can rename on their own cadence, then drop the alias.
- **(b) Same-cycle coordinated rename.** Land the rename without an alias; coordinate the corresponding rename in sbsh and sbcrew in the same release cycle.

Drop the now-stale "name predates #626" sentence from `cellblueprint.go:78-81` once the rename lands. Drop the matching prose reference at `cellblueprint.go:27` too.

## Acceptance criteria

- [ ] `CellProfileParameter` no longer appears in kukeon source — except, if option (a) is chosen, as a `// Deprecated:` type alias with a migration pointer in the comment.
- [ ] `CellBlueprintParameter` is the canonical type name for the parameter struct, exported from `pkg/api/model/v1beta1`.
- [ ] All in-repo call sites compile and `go test ./...` passes against the renamed type.
- [ ] The Blueprint type comment at `cellblueprint.go:78-81` no longer carries the name-vs-role apology; the prose reference at `cellblueprint.go:27` is updated to `CellBlueprintParameter`.
- [ ] The migration choice (alias-bridge vs. same-cycle coordinated rename) is documented in the PR body, and — if (a) — the deprecation comment names the removal release.
- [ ] On-wire YAML/JSON for `CellBlueprintSpec.parameters[]` is unchanged (no kind discriminator on the struct; positionally referenced).

## Notes

- The `refactor` label suggested in the originating conversation does not exist on this repo (`gh label list`); filed with `go` (Go code change) + `priority:C` (cosmetic clarity, no behavior change).
- The "breaking" classification only applies under option (b), and only to external importers — there are no in-repo callers that would not be renamed by the same PR.
- Surfaced during a design-rationale review of the `CellBlueprint` / `CellConfig` / `Cell` type triplet; not a runtime defect.

eminwux:
## Summary

- Rename the Go identifier `CellProfileParameter` → `CellBlueprintParameter` in `pkg/api/model/v1beta1/cellblueprint.go`, drop the `// The name keeps its 'CellProfile' prefix` apology from the type comment, and update the prose reference at line 27.
- Update every in-repo call site (controller, materializer, parser, tests, `cmd/kuke/create/config/config.go:writeValues`) — 9 files total, including `cmd/kuke/run/divergence_idempotency_test.go` which the issue body did not enumerate (count drifted from 8 → 9 since filing; intent unchanged).
- **Migration choice: option (a) one-release deprecated alias.** Keep `type CellProfileParameter = CellBlueprintParameter` with a `// Deprecated:` doc comment in `pkg/api/model/v1beta1/cellblueprint.go` so sibling projects (sbsh, sbcrew, …) can rename on their own cadence. The deprecation comment names **v0.7.0** as the removal release (latest tag is v0.5.0, so the alias survives the v0.6.0 release window and is dropped before v0.7.0). Lower-blast-radius default per CLAUDE.md §Default mode.
- On-wire YAML/JSON for `CellBlueprintSpec.parameters[]` is unchanged: there is no kind discriminator on the struct, the field is referenced positionally, and the alias is a Go-side type identity (same wire shape, same JSON tags).

## Test plan

- [x] `GOTOOLCHAIN=go1.24.5 make test-root` — all packages green (root module's go.mod pins go 1.24.5; the host's Go 1.25.5 trips a pre-existing pre-existing `containerd/cgroups/v2` typing break on `main` itself, recoverable via the go.mod-required toolchain per CLAUDE.md §When the project's CI test target's toolchain isn't installed; CI uses `go-version-file: go.mod` so the green run reproduces in CI).
- [x] `GOTOOLCHAIN=go1.24.5 GOWORK=off go vet ./...` — clean.
- [x] `grep -rn 'CellProfileParameter' .` — only two matches, both in the deprecated-alias declaration in `pkg/api/model/v1beta1/cellblueprint.go` (doc comment + `type` line).
- [x] `gofmt -l pkg/api/model/v1beta1/cellblueprint.go internal/...` on touched files — clean.
- [ ] `make dev-init` smoke — not exercised; the rename is Go-source-identity-only with no wire-format, daemon, or `/opt/kukeon` impact. The two-realm tail wouldn't surface anything this PR could break.

Closes #986

### DIFF
--- a/cmd/kuke/create/cell/cell_test.go
+++ b/cmd/kuke/create/cell/cell_test.go
@@ -614,7 +614,7 @@ func blueprintDoc() v1beta1.CellBlueprintDoc {
 		},
 		Spec: v1beta1.CellBlueprintSpec{
 			Prefix:     "web",
-			Parameters: []v1beta1.CellProfileParameter{{Name: "TAG", Default: &def}},
+			Parameters: []v1beta1.CellBlueprintParameter{{Name: "TAG", Default: &def}},
 			Cell: v1beta1.BlueprintCellSpec{
 				Containers: []v1beta1.BlueprintContainer{
 					{ID: "main", Image: "registry.example.com/web:${TAG}", Attachable: true},
--- a/cmd/kuke/create/config/config.go
+++ b/cmd/kuke/create/config/config.go
@@ -206,7 +206,7 @@ func emitConfigYAML(out io.Writer, configName, realm, space, stack string, bp *v
 	return err
 }
 
-func writeValues(b *strings.Builder, params []v1beta1.CellProfileParameter) {
+func writeValues(b *strings.Builder, params []v1beta1.CellBlueprintParameter) {
 	if len(params) == 0 {
 		b.WriteString("  values: {}\n")
 		return
--- a/cmd/kuke/create/config/config_test.go
+++ b/cmd/kuke/create/config/config_test.go
@@ -177,7 +177,7 @@ func TestNewConfigCmd_EmitsParsableYAML(t *testing.T) {
 			blueprint: v1beta1.CellBlueprintDoc{
 				Metadata: v1beta1.CellBlueprintMetadata{Name: "bp", Realm: "r1"},
 				Spec: v1beta1.CellBlueprintSpec{
-					Parameters: []v1beta1.CellProfileParameter{
+					Parameters: []v1beta1.CellBlueprintParameter{
 						{Name: "TOKEN", Required: true},
 					},
 					Cell: v1beta1.BlueprintCellSpec{
@@ -192,7 +192,7 @@ func TestNewConfigCmd_EmitsParsableYAML(t *testing.T) {
 			blueprint: v1beta1.CellBlueprintDoc{
 				Metadata: v1beta1.CellBlueprintMetadata{Name: "bp", Realm: "r1"},
 				Spec: v1beta1.CellBlueprintSpec{
-					Parameters: []v1beta1.CellProfileParameter{
+					Parameters: []v1beta1.CellBlueprintParameter{
 						{Name: "PORT", Default: ptr("5432")},
 						{Name: "ENV", Default: ptr("production")},
 						{Name: "TAGS", Default: ptr("a,b c")},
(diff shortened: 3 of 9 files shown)


eminwux:
**PR #1004 Review — `refactor(api): rename CellProfileParameter to CellBlueprintParameter`**

LGTM — mechanical rename with type-alias shim (`type CellProfileParameter = CellBlueprintParameter`) so sibling projects (sbsh, sbcrew) can migrate on their own cadence; wire shape unchanged, deprecation clearly marked for removal in v0.7.0.

PR #1000, opened 2026-06-01
fix(apply): diff WorkingDir, host-ns toggles, networks, RestartPolicy, Attachable, Tty, Git on container spec

eminwux:
## Summary

- Extends `diffContainerSpec` (`internal/controller/apply/diff.go`) with comparator branches for the 11 persisted, user-authored fields enumerated in the issue: `WorkingDir`, `Networks`, `NetworksAliases`, `HostNetwork`, `HostPID`, `HostCgroup`, `CNIConfigPath`, `RestartPolicy`, `Attachable`, `Tty`, `Git`. Without these, `kuke apply -f cell.yaml` returned a clean "no changes" on edits to any of them while the persisted spec drifted from the running OCI spec.
- Classification matches the issue proposal:
  - **Breaking** — `HostNetwork`, `HostPID`, `HostCgroup` (host-namespace toggles change the parent cell's OCI namespace shape; child containers inherit those namespaces via `JoinContainerNamespaces` at cell-start, so flipping them cannot be applied in place).
  - **Compatible** — the other eight, matching how `Env` / `Volumes` are handled today.
- Adds `ttyEqual` and `gitEqual` (with `gitIsZero`, `gitIdentityEqual`, `gitIdentityIsZero` helpers) following the `capabilitiesEqual` / `resourcesEqual` style. Both treat nil and a zero-valued block as equal so allocating or clearing an explicitly-empty block does not register as drift. `Networks` and `NetworksAliases` are plain `[]string`, so `slicesEqual` is reused directly — a separate `networksEqual` would be a one-line alias.
- Adds three `diff_test.go` cases: `TestDiffContainer_WorkingDirChange` (one Compatible field), `TestDiffContainer_HostNetworkChange` (one Breaking field), `TestDiffContainer_TtyChange` (the `*ContainerTty` pointer equality helper, plus nil-vs-zero-valued and same-content sub-cases).

## Runner re-stamp audit (AC4)

Verified each newly Compatible field for whether `BuildContainerSpec` / the runner's attach path re-reads it from `ContainerSpec` when the breaking trio (image/command/args) triggers UpdateCell's stop-remove-recreate-start:

| Field | Re-read at recreate? | Site |
|-------|----------------------|------|
| `WorkingDir` | ✅ | `internal/ctr/spec.go:370` (`oci.WithProcessCwd`) |
| `Attachable` | ✅ | `internal/ctr/spec.go:467` (`withAttachableMounts`/`withAttachableArgsWrap`) |
[...]

PR #1006, opened 2026-06-01
fix(runner): gate wind-down/auto-delete on per-container RestartPolicy

### LINKED ISSUE #1003: runner: ContainerSpec.RestartPolicy persisted but never consumed at start/restart time
## Context

Surfaced while implementing #991 (PR #1000). #991's AC4 directs:

> If any field is dropped or hardcoded by the runner downstream, file a follow-up and document the gap in the PR body.

This is that follow-up for `RestartPolicy`. (`Networks` / `NetworksAliases` are a separate, sibling follow-up — tracked in #1002.)

## Observed

`Spec.RestartPolicy string` (Always / OnFailure / Never) is declared in `internal/modelhub/container.go:80` and plumbed through the v1beta1 ↔ internal scheme bridges:

- `internal/apischeme/scheme.go:427,492,560,600`
- `internal/cellblueprint/cellblueprint.go:245`
- `internal/cellconfig/materialize.go:192`
- `pkg/api/model/v1beta1/container.go:115`
- `cmd/kuke/create/container/container.go:284` (CLI authoring)
- `cmd/kuke/run/run.go:1360` (the `kuke run` ephemeral-cell path *does* honor it)

But a tree-wide search for `RestartPolicy` finds **zero** consumers under `internal/controller/runner/` or `internal/ctr/`. The reconciler's restart loop in `internal/controller/runner/reconcile*.go` always re-starts a stopped container regardless of the field's value. So an operator who authors `restartPolicy: Never` in `cell.yaml` gets the field persisted, but a stopped container is still restarted by the reconciler.

## Root cause

The reconciler's restart logic never grew a `RestartPolicy` branch. The field is treated as informational metadata, not policy.

## Suggested fix

In the runner's restart path (the `ReconcileContainer` / `ReconcileCell` family under `internal/controller/runner/`), branch on the persisted `ContainerSpec.RestartPolicy` before re-issuing `StartContainer`:

- `Always` (or empty for back-compat) → current behavior.
- `OnFailure` → restart only when the prior `Status.ExitCode != 0`.
- `Never` → leave the container in `Stopped` state and surface that as a cell-level condition.

#991 added the diff-side detection so apply will *see* a `restartPolicy` change and route it through `UpdateCell` rather than emitting "no changes."

## References

- Surfaced by #991 / PR #1000
- Sibling runner-gap follow-up: #1002 (Networks / NetworksAliases)
- `internal/modelhub/container.go:80` — field declaration
- `cmd/kuke/run/run.go:1360` — the only place `RestartPolicy` is currently compared at runtime (`kuke run` ephemeral-cell path); the persistent reconcile loop has no equivalent
- `internal/controller/runner/reconcile_*.go` — restart loop sites

eminwux:
## Summary

- The reconciler now honors the per-container ` + "`RestartPolicy`" + ` field on the wind-down / auto-delete trigger. Empty / ` + "`always`" + ` keeps the pre-#1003 behavior; ` + "`on-failure`" + ` only reaps when a terminally-exited non-root container had a non-zero exit code; ` + "`never`" + ` keeps the cell in Stopped state for the operator to tear down explicitly. Any single non-root container that says "do not reap" blocks the cell-level wind-down, so a multi-container cell only winds down when every terminally-exited non-root container's policy permits it.
- ` + "`ContainerStatus.ExitCode`" + ` is no longer hardcoded to 0 — a new ` + "`GetContainerObservation`" + ` thin-wraps ` + "`GetContainerState`" + ` so ` + "`populateCellContainerStatuses`" + ` can carry the containerd ` + "`Status.ExitStatus`" + ` value into the persisted status snapshot. Without this, the ` + "`on-failure`" + ` gate collapses to ` + "`never`" + ` (every Stopped container reads back ExitCode=0). ` + "`GetContainerState`" + ` keeps its existing signature so no call site or interface change is needed.
- Adds canonical ` + "`RestartPolicyAlways`" + ` / ` + "`RestartPolicyOnFailure`" + ` / ` + "`RestartPolicyNever`" + ` constants in ` + "`internal/modelhub/container.go`" + ` next to the field declaration so the runner gate and any future consumer share one source of truth on the string spelling.

## Premise note — issue body has rotted

The issue claims "the reconciler's restart loop always re-starts a stopped container regardless of the field's value." Reading the runner today, there is no automatic restart loop — when non-root containers exit the reconciler either auto-deletes the cell (` + "`Spec.AutoDelete=true`" + `) or kills-and-preserves it (` + "`Spec.AutoDelete=false`" + `, via ` + "`windDownCell`" + `). The field has been purely cosmetic — ` + "`docs/site/manifests/container.md:92`" + ` even calls this out ("accepted by the schema today but the controller does not act on them"). PR #1000 added diff detection for the field, this PR makes the runner act on it.

Given there is no restart loop to gate, I read the AC as targeting the closest existing analog — the wind-down / auto-delete reap — and wired the three policy modes onto that gate. Documented in-code on ` + "`ContainerSpec.RestartPolicy`" + ` and ` + "`restartPolicyPermitsCellReap`" + `; please push back if a real restart loop (re-issue ` + "`StartContainer`" + ` on a stopped task) is the intended scope instead. That would land as a separate follow-up and be a substantially bigger change.

## Test plan

- [x] ` + "`make test`" + ` (full repo test target — both root and kukebuild modules pass)
- [x] ` + "`make dev-init`" + ` smoke (both realms Ready, kuke attach smoke clean)
- [x] New unit tests in ` + "`internal/controller/runner/restart_policy_test.go`" + ` cover the per-container gate decision table, the cell-level aggregation (any-` + "`never`" + `-blocks, root ignored, non-terminal ignored, missing-status ignored), and the ` + "`GetContainerObservation`" + ` exit-code plumbing on both the TaskStatus-success and NotCreated branches

## Acceptance criteria

- [x] ` + "`Always`" + ` (or empty for back-compat) → current wind-down behavior preserved (` + "`TestRestartPolicyPermitsContainerReap/empty_back_compat_*`" + `, ` + "`always_*`" + `)
- [x] ` + "`OnFailure`" + ` → reap only when prior ` + "`Status.ExitCode != 0`" + ` (` + "`TestRestartPolicyPermitsContainerReap/on_failure_*`" + ` + ` + "`TestGetContainerObservation_PopulatesExitCode`" + `)
- [x] ` + "`Never`" + ` → leave the container in Stopped state, surfaced via the cell-level wind-down skip (` + "`TestRestartPolicyPermitsContainerReap/never_*`" + ` + ` + "`TestRestartPolicyPermitsCellReap/never_policy_terminal_blocks`" + `). Note: the AC suggested surfacing as a "cell-level condition" — ` + "`CellStatus`" + ` has no Conditions API today, only ` + "`Reason`" + ` / ` + "`Message`" + `; the persisted observation is the per-container ` + "`Stopped`" + ` state plus the cell staying out of wind-down (no Reason stamp added in this PR to avoid widening scope)

Closes #1003

### DIFF
--- a/docs/site/manifests/container.md
+++ b/docs/site/manifests/container.md
@@ -85,11 +85,24 @@ See [Concepts → Container](../concepts/container.md) for what a container is.
 | `repos`           | array of `ContainerRepo`   | no       | Git repos the kuketty wrapper clones before the workload starts — requires `attachable: true` (see [ContainerRepo](#containerrepo))                    |
 | `git`             | `ContainerGit`             | no       | Declarative git identity + signing, expanded into `GIT_AUTHOR_*`/`GIT_COMMITTER_*`/`GIT_CONFIG_*` env before start (see [ContainerGit](#containergit)) |
 | `cniConfigPath`   | string                     | no       | Override the CNI config directory for this container                                                                                                   |
-| `restartPolicy`   | string                     | no       | Restart policy. Reserved — restart semantics are not finalized.                                                                                        |
+| `restartPolicy`   | string                     | no       | Per-container reap policy at the cell wind-down / auto-delete gate. One of `always`, `on-failure`, `never`. Empty defaults to `always` for back-compat (see [Restart policy](#restart-policy)). |
 | `tty`             | `ContainerTty`             | no       | Shell-UX config for the kuketty wrapper (prompt, init scripts, logging) — requires `attachable: true` (see [ContainerTty](#containertty))              |
 
 !!! warning "Fields marked reserved"
-`ports` and `restartPolicy` are accepted by the schema today but their semantics are still being designed. Values round-trip (you can read back what you applied), but the controller does not act on them. See [GitHub Issues](https://github.com/eminwux/kukeon/issues) for the backlog.
+`ports` is accepted by the schema today but its semantics are still being designed. Values round-trip (you can read back what you applied), but the controller does not act on them. See [GitHub Issues](https://github.com/eminwux/kukeon/issues) for the backlog.
+
+### Restart policy
+
+`spec.restartPolicy` selects whether the cell wind-down / auto-delete reconciler reaps a cell after one of its non-root containers exits. The runner evaluates the policy per container at the wind-down gate; the cell-level decision is the intersection across every terminally-exited non-root container, so a single `never` blocks the wind-down.
+
+| Value        | Behavior at wind-down                                                                                                                              |
+| ------------ | -------------------------------------------------------------------------------------------------------------------------------------------------- |
+| empty/unset  | Treated as `always` for back-compat. Pre-#1003 wind-down behavior — every non-root container exit can trigger a cell-level wind-down.              |
+| `always`     | Wind-down / auto-delete proceeds whenever the container terminally exits, regardless of exit code.                                                 |
+| `on-failure` | Wind-down / auto-delete proceeds only when the container's last exit code is non-zero. A zero-exit (clean shutdown) keeps the cell in place.       |
+| `never`      | Wind-down / auto-delete is skipped — the cell stays in `Stopped` until an operator tears it down explicitly (`kuke delete` / `kuke purge`).        |
+
+Only the root container is exempt: the root's exit drives cell-level lifecycle decisions independently and is not subject to this gate.
 
 ### VolumeMount
 
--- a/internal/controller/runner/container_state.go
+++ b/internal/controller/runner/container_state.go
@@ -28,39 +28,62 @@ import (
 	"github.com/eminwux/kukeon/internal/util/naming"
 )
 
+// ContainerObservation bundles the slice of containerd task state the runner
+// needs to surface in ContainerStatus and drive lifecycle decisions: the
+// internal-state mapping plus the exit code reported by the task. ExitCode is
+// only meaningful on the TaskStatus-success branch — on every other branch
+// (NotCreated, ErrTaskNotFound, transient error, fallback Unknown) it stays
+// zero.
+type ContainerObservation struct {
+	State    intmodel.ContainerState
+	ExitCode int
+}
+
 // GetContainerState queries containerd for the actual task status of a container
-// and converts it to the internal ContainerState.
+// and converts it to the internal ContainerState. Thin wrapper around
+// GetContainerObservation for callers that only need the state column.
 func (r *Exec) GetContainerState(cell intmodel.Cell, containerID string) (intmodel.ContainerState, error) {
+	obs, err := r.GetContainerObservation(cell, containerID)
+	return obs.State, err
+}
+
+// GetContainerObservation is GetContainerState plus the task's exit code.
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #1006 Review — `fix(runner): gate wind-down/auto-delete on per-container RestartPolicy`**

Gate placement is correct (vetoes both `autoDeleteCell` and `windDownCell` at one site), the policy decision table is exhaustively unit-tested, and `GetContainerObservation` keeps the old signature so no internal callers regress. Premise re-interpretation (wind-down rather than a restart loop) is honestly documented in the PR body and the in-code comments; that's a scope question for you, not a coding bug. No secrets/PII leak; commit signed and verified.

A few items:

- **Docs rot** `docs/site/manifests/container.md:88,92`: this PR makes the controller act on `restartPolicy`, but the table still says "Reserved — restart semantics are not finalized" and the warning block still says "`ports` and `restartPolicy` are accepted by the schema today but ... the controller does not act on them." An operator reading the manifest reference would conclude the field is still a no-op and never reach the new behavior. Update the row (describe the three policy values + the back-compat empty default) and either drop `restartPolicy` from the reserved-warning or rewrite the warning to only mention `ports`.
- **String-literal in sibling test** (nit, non-blocking) `cmd/kuke/run/divergence_idempotency_test.go:245`: bare `"on-failure"` could now use `intmodel.RestartPolicyOnFailure` for consistency with the new canonical spelling. Single site, fine to defer.

LGTM once the manifest docs are updated — the test-literal nit is non-blocking.

eminwux:
Addressed in 68c9c99:
- Docs rot in `docs/site/manifests/container.md`: `spec.restartPolicy` row now describes the three policy values + the back-compat empty default and points at a new `### Restart policy` section that lays out the wind-down decision table (empty → `always`, `always`, `on-failure`, `never`) plus the cell-level intersection rule and the root-container exemption. Reserved-warning block rewritten to mention only `ports`.

Deferred: the `cmd/kuke/run/divergence_idempotency_test.go:245` `"on-failure"` string-literal nit — flagged non-blocking; will roll into a follow-up sweep alongside any other lingering bare-string sites once a few callers exist.

eminwux:
**PR #1006 Review — `fix(runner): gate wind-down/auto-delete on per-container RestartPolicy`**

LGTM — `restartPolicyPermitsCellReap` gates both reap paths on a clean per-container decision table, `GetContainerObservation` plumbs `Status.ExitStatus` through to `ContainerStatus.ExitCode` so the `on-failure` branch doesn't collapse to `never`, the unit tests cover the decision table, the cell-level any-`never`-blocks aggregation, root-ignored, non-terminal-ignored, missing-status-ignored, and both ExitCode plumbing branches (TaskStatus-success + NotCreated). The scope-rebrand note in the PR body (no real restart loop exists, so the gate landed on the wind-down/auto-delete reap instead) is the honest read of the current runner; back-compat default of empty→Always preserves pre-#1003 behavior for unset cells. Apache header + SPDX present on the new test file, constants live next to the field they describe in `modelhub`, the `make dev-init` daemon-parity smoke is ticked.

Issue #796, opened 2026-05-23
team: TeamsConfig lifecycle + kuketeam.d + init skeleton (step 1)


eminwux:
## Context

Part of the **team-distribution** epic (umbrella: #792). Implementation step 1 of the `kuke team init` verb. Resized from the original whole-pipeline scope (dev triage flagged 1000+ LOC across ~6 net-new subsystems) into a 4-step chain: this step lands the command skeleton + host-config lifecycle; #1041/#1042/#1043 follow.

**Design change vs the original body:** the original persisted every project's composition into a single `~/.kuke/kuketeams.yaml` under `spec.teams[]`. A single shared file is a failure-isolation and write-concurrency hazard — a corrupt/partial write (and these happen on *every* `init`) takes down all projects, and two concurrent `kuke team init` runs race on the shared array. This step adopts a **drop-in directory** instead (systemd/`sudoers.d` pattern): operator-global facts stay in `kuketeams.yaml`; per-project composition moves to one file per project under `kuketeam.d/`.

## Proposal

- **Command skeleton** — `kuke team` parent + `init` subcommand + flags (`--dry-run` reserved for step 3) + root registration (`cmd/kuke/kuke.go` registration site; net-new `cmd/kuke/team/` package).
- **Global-facts file** `~/.kuke/kuketeams.yaml` — first run scaffolds it: git identity seeded from `git config --global`, registry/secrets from env/prompt, `sources` map. Holds operator-global facts **only**.
- **Per-project file** `~/.kuke/kuketeam.d/<project>.yaml` — each `init` reads `./kuketeam.yaml`, resolves the project name, and writes/rewrites *that project's* file. No read-modify-write of a shared array; a corrupt write touches one project.
- **Type adjustment to #1028's merged `TeamsConfig`** — drop `Teams []TeamsConfigTeam` from `TeamsConfigSpec`; promote the per-project entry (`name`, `path`, `source`) to its own `kuketeams.io/v1` document kind, one document per `kuketeam.d/<project>.yaml` file. Add the kind to the `internal/kuketeams` parser's `Document`/`Parse` dispatch. The entry's `source` field uses the **structured object** struct (`repo` + exactly-one-of `tag`/`branch`/`commit`) introduced in #1041 — **not** the old string form; keep the struct shared between `ProjectTeam.source` and this entry so they can't drift.

This step does **not** resolve sources, render, or apply — those are steps 2–4.

## Acceptance criteria

[...]

Issue #792, opened 2026-05-23
epic: team composition via `kuke team init` (two-repo, per-project rosters)


eminwux:
## Context

This epic was originally scoped as a **distribution-first** story: a compiler in `eminwux/agents` emits **signed VendorPacks**, the kukeon runtime **embeds** a default team via `go:embed`, verifies packs with **cosign**, and ships zero-touch via `get.kukeon.io`. That path is real but months long and does not address the near-term pain: `crew` and `agents` co-evolve fast across repo boundaries, and onboarding a project means hand-editing a global `crew.conf` + running `install.sh`.

This rewrite redirects the epic to the design we converged on: **in-repo team composition**. Roles are a **git source** (`repo` + `tag`/`branch`/`commit`; pinned by default, floating branch allowed for the dev loop) materialized by the existing Makefile; a single per-project verb (`kuke team init`) renders and applies a project's roster directly to `kukeond`. No compiler, no packs, no signing, no embed in this phase. The signed-bundle distribution work is **deferred** (see "Deferred" below) and can be revived once the composition model is proven.

## Architecture (two repos)

- **`eminwux/kukeon`** — runtime + CLI. Adds the `kuke team init` verb; knows the team **schemas**, not their contents.
- **`eminwux/agents`** — roles **and** harnesses, resolved as one source ref (pinned by default; floating branch allowed). Absorbs crew's shared-source half (Dockerfiles, blueprint templates). `crew` dissolves.

### Config files

| File | Where | Owner | Holds | Committed |
|---|---|---|---|---|
| `kuke.yaml` | host | runtime | kukeon runtime only (realm/space/stack, daemon). **No git/registry/teams.** | n/a |
| `kuketeam.yaml` | project repo | project | roster: `source`, roles, harnesses, per-role `needs` | yes |
| `~/.kuke/kuketeams.yaml` | host | operator | git identity + signing (key, allowed-signers, ssh key), registry, secret sources, source URLs. **Operator-global facts only.** | no |
| `~/.kuke/kuketeam.d/<project>.yaml` | host | operator | one file per composed project: `name`, `path`, `source`. Written/rewritten by each `kuke team init`. | no |

### Contract kinds (replace VendorPack/TeamManifest)

- **`ProjectTeam`** (`kuketeam.yaml`) — per-project roster, committed, travels with code.
[...]

PR #1029, opened 2026-06-02
feat(daemon): per-team labeling + pruning apply for CellBlueprint/CellConfig

### LINKED ISSUE #1027: kukeond: per-team labeling + pruning apply
## Context

`kuke team init` (#796) is fileless: it renders a project's CellBlueprints/CellConfigs in memory and applies them directly to `kukeond`, which becomes the single store of applied state. For `init` to be **idempotent** and to **compose** many projects in one daemon without a local ledger, the daemon must be able to scope applied objects per project and prune what a project no longer declares.

This is the one daemon-side prerequisite for the redirected epic (#792).

## Proposal

Add per-team scoping to applied objects and a pruning apply:

1. **Team label** — applied CellBlueprints / CellConfigs carry a `team=<project>` label (alongside the existing realm/space/stack scoping). `kuke team init` sets it.
2. **Pruning apply** — apply a labeled object **set**: create/update the members, and **delete** objects with the same `team=<label>` that are not in the applied set. This makes re-running `init` after removing a role from `kuketeam.yaml` converge (the orphaned objects are pruned).
3. **Query by label** — list objects filtered by `team=<label>`, to compute the prune set (and to back a future `kuke team status` / `down`).
4. **Preserve running cells** — re-applying an unchanged blueprint must not kill live cells (crew's current `install.sh` re-apply relies on this; keep the invariant).

## Acceptance criteria

- [ ] CellBlueprint/CellConfig apply accepts and persists a `team` label.
- [ ] An apply mode takes a set + a team label and prunes daemon objects with that label not present in the set.
- [ ] Objects can be listed/queried by `team` label.
- [ ] Re-applying an unchanged set is a no-op for running cells (no restart, no recreate).
- [ ] Unit/e2e: apply set A (team=t) then set B⊂A (team=t) → members of A∖B are pruned, other teams untouched.

## Dependencies

Gates #796 (`kuke team init` idempotency). Independent of #793 (schemas), can land in parallel.

Part of the **team-distribution** epic. Umbrella: #792.

eminwux:
## Summary

- Adds the daemon-side prerequisite for `kuke team init` (#796) under the team-distribution epic (#792): per-team scoping on applied CellBlueprint / CellConfig objects, and a pruning apply mode that converges a team's roster by deleting the team's prior objects the new set no longer declares.
- The Apply RPC gains an optional `Team` field. When set, the daemon stamps `kukeon.io/team=<team>` on every applied Blueprint / Config, then enumerates daemon-stored objects carrying that label and deletes those not in the applied set. Empty `Team` preserves the historical no-team, no-prune `kuke apply -f` path.
- `ListBlueprints` / `ListConfigs` now populate `Metadata.Labels` by parsing each persisted document's `metadata.labels`, so the prune pass and any client-side selector can filter by team without a per-object Get follow-up.
- Running cells are preserved by construction: re-applying an unchanged Blueprint / Config overwrites the same document bytes (no recreate path), and `DeleteBlueprint` / `DeleteConfig` are documented as having no live-reference gate (cells are independent copies per #620, materialized cells outlive their Config per #644).

## Design notes

- New `ApplyDocumentsForTeam` method on the `kukeonv1.Client` interface instead of changing the `ApplyDocuments` signature — keeps existing call sites untouched while exposing the new mode for #796 and other future consumers. Both methods route through the same `MethodApplyDocuments` wire endpoint; `ApplyDocumentsArgs.Team` is the optional discriminator.
- The team label is set by the apply caller (stamped server-side), not required to be hand-written into every YAML document. This matches the per-project semantics of `kuke team init`: the verb knows the project name; the YAML doesn't need to.
- Prune-result entries are surfaced as `ApplyResourceResult{Action: "pruned"}` so callers can render the prune count without re-querying the daemon.
- `pruneTeamObjects` walks Configs before Blueprints so the order mirrors apply-time dependency direction (Config references Blueprint), though both DeleteBlueprint/DeleteConfig are independent of live cells so safety doesn't depend on order.
- `intmodel.CellBlueprintMetadata` and `intmodel.CellConfigMetadata` gained `Labels map[string]string`. The structs are no longer comparable with `==`/`!=`; `internal/controller/reconcile_outofsync_test.go:432` was updated to field-by-field comparison (probe-ordering test, only scope coordinates matter there).

## Out of scope

- No CLI surface (`kuke apply --team <name>`) added — `kuke team init` (#796) is the in-tree consumer and will call `ApplyDocumentsForTeam` directly through the client. A `--team` flag on `kuke apply` is a natural follow-up if hand-driven per-team apply ever becomes useful.
- No `-l/--selector` flag wired into `kuke get blueprints` / `kuke get configs`. The runner now returns labels through the existing list RPC, so adding the flag is a small follow-up that reuses `cmd/kuke/get/shared/selector.go` — out of scope here, since the AC's "queried by team label" capability is satisfied by the daemon-side list returning labels.

## Acceptance criteria

- [x] CellBlueprint/CellConfig apply accepts and persists a `team` label.
- [x] An apply mode takes a set + a team label and prunes daemon objects with that label not present in the set.
- [x] Objects can be listed/queried by `team` label (`ListBlueprints` / `ListConfigs` round-trip `Metadata.Labels` from each document body).
- [x] Re-applying an unchanged set is a no-op for running cells (no restart, no recreate) — relies on existing invariants in `runner.WriteBlueprint`/`WriteConfig` (write-through, no cell touch) and `DeleteBlueprint`/`DeleteConfig` (no live-reference gate; cells are independent of their Blueprint/Config source).
- [x] Unit/e2e: apply set A (team=t) then set B⊂A (team=t) → members of A∖B are pruned, other teams untouched. Covered by `TestApplyDocuments_TeamPruneScopedToLabel` + `TestApplyDocuments_TeamPruneEmptySet` + `TestApplyDocuments_NoTeamSkipsPrune` in `internal/controller/apply_team_prune_test.go`.

## Test plan

- [x] `GOWORK=off go test $(GOWORK=off go list ./... | grep -v /e2e)` — full project test target (the Makefile's `test-root`), all packages green including `internal/controller`, `internal/controller/runner`, `internal/controller/apply`, `internal/apischeme`, `internal/client/local`, `internal/daemon`, `pkg/api/kukeonv1`, `cmd/kuke/apply`.
- [x] `GOWORK=off go test ./cmd/kukebuild/...` — the second leg of `make test`'s `test-kukebuild`.
- [x] `GOWORK=off go vet ./...` clean.
- [ ] `make dev-init` smoke not run — change is daemon controller logic only; no build, install, or `/opt/kukeon` paths touched.

Closes #1027

### DIFF
--- a/internal/apischeme/cellblueprint.go
+++ b/internal/apischeme/cellblueprint.go
@@ -42,10 +42,11 @@ func ConvertCellBlueprintDocToInternal(in ext.CellBlueprintDoc) (intmodel.CellBl
 		}
 		return intmodel.CellBlueprint{
 			Metadata: intmodel.CellBlueprintMetadata{
-				Name:  in.Metadata.Name,
-				Realm: in.Metadata.Realm,
-				Space: in.Metadata.Space,
-				Stack: in.Metadata.Stack,
+				Name:   in.Metadata.Name,
+				Realm:  in.Metadata.Realm,
+				Space:  in.Metadata.Space,
+				Stack:  in.Metadata.Stack,
+				Labels: copyLabels(in.Metadata.Labels),
 			},
 			Document: document,
 		}, nil
@@ -87,12 +88,29 @@ func ConvertCellBlueprintMetadataToExternal(in intmodel.CellBlueprint) ext.CellB
 		APIVersion: ext.APIVersionV1Beta1,
 		Kind:       ext.KindCellBlueprint,
 		Metadata: ext.CellBlueprintMetadata{
-			Name:  in.Metadata.Name,
-			Realm: in.Metadata.Realm,
-			Space: in.Metadata.Space,
-			Stack: in.Metadata.Stack,
+			Name:   in.Metadata.Name,
+			Realm:  in.Metadata.Realm,
+			Space:  in.Metadata.Space,
+			Stack:  in.Metadata.Stack,
+			Labels: copyLabels(in.Metadata.Labels),
 		},
 	}
+}
+
+// copyLabels returns an independent copy of labels so the internal carrier
+// and the external doc do not alias the same map (a later mutation on one
+// must not leak into the other). A nil or empty input returns nil so the
+// canonical YAML-marshalled shape omits `labels:` rather than emitting an
+// empty map.
+func copyLabels(in map[string]string) map[string]string {
+	if len(in) == 0 {
+		return nil
+	}
+	out := make(map[string]string, len(in))
+	for k, v := range in {
+		out[k] = v
+	}
+	return out
 }
 
 // ConvertCellBlueprintListToExternal maps a slice of internal CellBlueprints to
--- a/internal/apischeme/cellconfig.go
+++ b/internal/apischeme/cellconfig.go
@@ -42,10 +42,11 @@ func ConvertCellConfigDocToInternal(in ext.CellConfigDoc) (intmodel.CellConfig, 
 		}
 		return intmodel.CellConfig{
 			Metadata: intmodel.CellConfigMetadata{
-				Name:  in.Metadata.Name,
-				Realm: in.Metadata.Realm,
-				Space: in.Metadata.Space,
-				Stack: in.Metadata.Stack,
+				Name:   in.Metadata.Name,
+				Realm:  in.Metadata.Realm,
+				Space:  in.Metadata.Space,
+				Stack:  in.Metadata.Stack,
+				Labels: copyLabels(in.Metadata.Labels),
 			},
 			Document: document,
 		}, nil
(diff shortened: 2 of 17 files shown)


eminwux:
**PR #1029 Review — `feat(daemon): per-team labeling + pruning apply for CellBlueprint/CellConfig`**

LGTM — `Team` is an additive wire field, prune is gated on `team != ""` and on `reconcileErr == nil`, labels round-trip via the canonical Document bytes (not just the carrier), `ApplyDocumentsForTeam` is threaded through all three `Client` implementations (`local`, `UnixClient`, `FakeClient`), and the three new tests pin the scoped-prune / empty-set / no-team-skip contract. Safe to merge.

Issue #1041, opened 2026-06-03
team: resolve source (object schema, SSH, ref-aware cache) + load docs (step 2)


eminwux:
## Context

Part of the **team-distribution** epic (umbrella: #792). Implementation step 2 of the `kuke team init` verb. Step 1 (#796) lands the command skeleton + `kuketeams.yaml` global facts + per-project `kuketeam.d/<project>.yaml` writer; this step makes the roster's source materializable.

This issue also **redesigns the `source` contract** that #1028 shipped as a string. The design discussion converged on: project files name *any* git host (not GitHub-coupled), declare ref intent *explicitly* (so pinned-vs-floating is unambiguous), and leave transport to the operator. See **Source schema redesign** below.

### Project side — `ProjectTeam.spec.source` becomes a structured object

Replace `ProjectTeamSpec.Source string` (today validated by `sourceRefPattern`) with:

```yaml
# kuketeam.yaml
spec:
  source:
    repo: github.com/eminwux/agents   # host-explicit; any git host
    tag: v1.4.0                        # exactly one of tag / branch / commit
    # branch: main                     #   tag, commit  -> pinned (reproducible)
    # commit: 9ae9606                  #   branch       -> floating (refetched each init)
```

Rules:
- `repo` is host-qualified (`<host>/<owner>/<repo>`). A bare `<owner>/<repo>` MAY default host to `github.com` (ergonomic shorthand) — but any host must be expressible.
- **Exactly one** of `tag` / `branch` / `commit`. Mutually exclusive; missing-or-multiple is a parse error.
- The key name *is* the intent: `tag`/`commit` = pinned; `branch` = floating. This is what makes the cache + reproducibility behavior below unambiguous without git interrogation (a string `@ref` can't distinguish a branch from a same-named tag).

Parser: drop `sourceRefPattern` string matching in `validateProjectTeam`; validate the object (repo host-qualified, exactly-one-ref). The merged per-project entry kind (`TeamEntry`, #796) carries a denormalized copy of this `source` and MUST adopt the same object struct — keep the type shared.

### Transport — SSH-always default, `sources` as override

- Default transport is **SSH**: the resolver expands `repo: <host>/<owner>/<repo>` -> `git@<host>:<owner>/<repo>.git`, using `kuketeams.yaml.git.sshKey` as the clone identity (already exists; no new field).
[...]

PR #1028, opened 2026-06-02
feat(team): add kuketeams.io/v1 types + parser (ProjectTeam/TeamsConfig/Role/Harness/ImageCatalog)

eminwux:
## Summary

Lands the team-distribution contract's Go types, parser, and validation (step 1 of epic #792) so `kuke team init` (#796) has something to read. **No verbs in this step** — this is the declarative surface only.

- **New types package `pkg/api/model/kuketeams/`** (one file per kind): `ProjectTeam`, `TeamsConfig`, `Role`, `Harness`, `ImageCatalog`. All five are GVK objects under one API group, `kuketeams.io/v1`, with full JSON+YAML marshalling.
- **`TeamsConfig.git` embeds `v1beta1.ContainerGit`** (`yaml:",inline"`) so it is a *literal* strict superset — author/committer/signingKey/sign/allowedSigners are carried verbatim and drift-proof, plus the added `sshKey` clone identity. Round-trip test asserts the embedded fields promote in both JSON and YAML.
- **`Role.needs` splits `repos` (→ `ContainerRepo`) from `mounts` (→ `VolumeMount`)** as distinct fields; `project`/`agents` are repos, `ssh` is a mount.
- **`ImageCatalog` entries carry build provenance** (`image` + `build.{context,dockerfile}`), not just `ref` → capabilities.
- **New parser `internal/kuketeams/`** deserializes + validates every rule in the issue (pinned-exact source refs, capability-names-not-tags, secret-source-not-value, known-harness set, registry-qualified images, unique team names, git identity/sign rules). Table-driven tests cover the happy path for each kind plus every failure mode. Validation sentinels added to `internal/errdefs`.
- **Docs:** `docs/site/concepts/team.md` describes the contract (the #792 file-ownership table), the capability-name / secret-source / version-is-the-pin disciplines, and the clone-not-mount project semantics; added to the mkdocs nav.

### Implementation calls (reviewer please push back)

- **Package placement:** new group `pkg/api/model/kuketeams/` rather than folding into `v1beta1`. The issue offered either; the new group matches the `kuketeams.io/v1` apiVersion, the "new types in their own package/files" AC, and the "no changes to existing schemas" AC.
[...]

PR #1045, opened 2026-06-03
feat(team): kuke team init skeleton + TeamEntry drop-in lifecycle (step 1)

### LINKED ISSUE #796: team: TeamsConfig lifecycle + kuketeam.d + init skeleton (step 1)

eminwux:
## Summary

Step 1 of the **team-distribution** epic (#792, umbrella) per the resized AC on #796 — lands the `kuke team init` command skeleton + host-config lifecycle. Source resolution, render, and apply remain steps 2–4 (#1041/#1042/#1043).

- **Command skeleton** — net-new `cmd/kuke/team/` package: `kuke team` parent + `init` subcommand, registered in `cmd/kuke/kuke.go`. `init` reads the project's committed `./kuketeam.yaml` roster; a missing file errors clearly (`ErrTeamProjectFileNotFound`), a wrong-kind file errors with `ErrTeamProjectFileKind`.
- **Global-facts scaffold** — first run writes `~/.kuke/kuketeams.yaml` (operator facts only): git identity/signing seeded from `git config --global`, registry from `KUKEON_REGISTRY` env. Re-runs never overwrite an existing file.
- **Per-project drop-in** — each `init` writes `~/.kuke/kuketeam.d/<project>.yaml` (a new `TeamEntry` kind) via atomic temp+rename. Drop-in dir is `0o700`, files `0o600`. One file per project, so a corrupt/partial write touches one project and concurrent inits don't race a shared array.
- **Type adjustment to #1028's `TeamsConfig`** — dropped `Teams []TeamsConfigTeam` from `TeamsConfigSpec`; promoted the per-project entry to the new `kuketeams.io/v1` `TeamEntry` kind (`metadata.name` + `spec.path`/`spec.source`), wired into the `internal/kuketeams` parser `Document`/`Parse` dispatch with validation.
- New host-side lifecycle package `internal/teamhost` (path layout + scaffold + atomic entry write), kept free of cobra so it's unit-testable against a temp dir.

### Non-obvious calls (reviewer, please push back)

- **Kind name `TeamEntry`** — the issue suggested it and left final naming to dev's call; adopted as-is.
- **`name` in `metadata`, `path`/`source` in `spec`** — matches the GVK shape of the sibling kinds (ProjectTeam/Role/Harness use `Metadata.Name`); the old flat `TeamsConfigTeam{Name,Path,Source}` is gone.
- **`--dry-run` semantics** — the AC reserves `--dry-run` for step 3's render output. Registered the flag now (AC: "parse flags") with a forward-compatible step-1 behavior: it prints the `TeamEntry` that would be written and touches no disk. Step 3 extends it to full render output.
- **Registry/secrets seeding** — AC says "registry/secrets from env/prompt". The verb is non-interactive (cron contract), so there is no prompt: registry is seeded from `KUKEON_REGISTRY` env; secrets are left empty for the operator to fill in. Documented here so step 2/3 can revisit if a richer seed is wanted.
- **Removed sentinels** — `ErrTeamTeamNameRequired`/`ErrTeamTeamNameDuplicate` (only used by the dropped `teams[]` loop) replaced by `ErrTeamEntryNameRequired`; duplicate-name is no longer meaningful (the filename is the uniqueness key). Tree-wide `git grep` confirms no other references.
- **Stale triage comment** — #796 carries an older dev-triage "too broad, requesting re-size" comment. PM has since resized the issue body to this step-1 scope; the comment is superseded and the current body is authoritative.

## Test plan

- [x] `make test` (the named CI target; `GOWORK=off` exported by the Makefile) — passes.
- [x] `GOWORK=off go build ./...` — passes. (Bare `go build ./...` fails on a pre-existing workspace-mode dep mismatch in `containerd/cgroups`, unrelated to this diff — the Makefile exports `GOWORK=off` for exactly this; see project CLAUDE.md.)
- [x] `GOWORK=off go vet` on all touched packages — clean.
- [x] New unit tests: `internal/teamhost` (scaffold-once, atomic write, per-project isolation, perms), `cmd/kuke/team` (missing/wrong-kind project file, scaffold+write, re-run no-rescaffold, git-identity seeding incl. sign entries, no-identity omits git block, dry-run writes nothing, command registration), `internal/kuketeams` (TeamEntry parse happy + failure modes), `pkg/api/model/kuketeams` (TeamEntry JSON/YAML round-trip).
- [ ] `make dev-init` smoke — **not run, not required**: this change is host-local (writes `~/.kuke` files) and daemon-independent. It does not touch the build path, the daemon, or `/opt/kukeon`, and no daemon code reads the team kinds yet (the parser is standalone, per its package doc). Per project CLAUDE.md the smoke is required only when a change touches something the daemon reads.

Closes #796

### DIFF
--- a/cmd/config/defaults.go
+++ b/cmd/config/defaults.go
@@ -71,14 +71,23 @@ func DefaultServerConfigurationFile() string {
 	return filepath.Join("/", "etc", "kukeon", "kukeond.yaml")
 }
 
-// DefaultClientConfigurationFile is the on-disk path the `kuke` client reads
-// when the user does not pass `--configuration`. An absent file is not an
-// error — the client falls back to its hardcoded defaults.
-func DefaultClientConfigurationFile() string {
+// DefaultKukeDir is the per-operator kuke state directory (~/.kuke) holding
+// the client configuration, the team-distribution global-facts file
+// (kuketeams.yaml), and the per-project drop-in directory (kuketeam.d/). Falls
+// back to "tmp" when the home directory cannot be determined, matching the
+// other default-path helpers.
+func DefaultKukeDir() string {
 	base, err := os.UserHomeDir()
 	if err != nil {
 		// fallback to tmp if home dir cannot be determined
 		base = "tmp"
 	}
-	return filepath.Join(base, ".kuke", "kuke.yaml")
+	return filepath.Join(base, ".kuke")
+}
+
+// DefaultClientConfigurationFile is the on-disk path the `kuke` client reads
+// when the user does not pass `--configuration`. An absent file is not an
+// error — the client falls back to its hardcoded defaults.
+func DefaultClientConfigurationFile() string {
+	return filepath.Join(DefaultKukeDir(), "kuke.yaml")
 }
--- a/cmd/kuke/kuke.go
+++ b/cmd/kuke/kuke.go
@@ -44,6 +44,7 @@ import (
 	startcmd "github.com/eminwux/kukeon/cmd/kuke/start"
 	statuscmd "github.com/eminwux/kukeon/cmd/kuke/status"
 	stopcmd "github.com/eminwux/kukeon/cmd/kuke/stop"
+	teamcmd "github.com/eminwux/kukeon/cmd/kuke/team"
 	uninstallcmd "github.com/eminwux/kukeon/cmd/kuke/uninstall"
 	"github.com/eminwux/kukeon/cmd/kuke/version"
 	"github.com/eminwux/kukeon/cmd/types"
@@ -155,6 +156,7 @@ func SetupKukeCmd(rootCmd *cobra.Command) error {
 	rootCmd.AddCommand(startcmd.NewStartCmd())
 	rootCmd.AddCommand(statuscmd.NewStatusCmd())
 	rootCmd.AddCommand(stopcmd.NewStopCmd())
+	rootCmd.AddCommand(teamcmd.NewTeamCmd())
 	rootCmd.AddCommand(killcmd.NewKillCmd())
 	rootCmd.AddCommand(purgecmd.NewPurgeCmd())
 	rootCmd.AddCommand(refreshcmd.NewRefreshCmd())
--- a/cmd/kuke/team/init.go
+++ b/cmd/kuke/team/init.go
@@ -1,0 +1,263 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package team
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/eminwux/kukeon/cmd/config"
+	"github.com/eminwux/kukeon/internal/errdefs"
+	"github.com/eminwux/kukeon/internal/kuketeams"
+	"github.com/eminwux/kukeon/internal/teamhost"
+	model "github.com/eminwux/kukeon/pkg/api/model/kuketeams"
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+)
+
+// projectFileName is the per-project roster committed in each project repo.
+const projectFileName = "kuketeam.yaml"
+
+// registryEnv seeds TeamsConfig.spec.registry on first-run scaffold.
+const registryEnv = "KUKEON_REGISTRY"
+
+// GitConfigFunc reads a single `git config --global <key>` value, reporting
+// whether the key was set. Injected via MockGitConfigKey in tests so the
+// scaffold is hermetic and does not read the host operator's real git config.
+type GitConfigFunc func(ctx context.Context, key string) (string, bool)
+
+// MockGitConfigKey injects a GitConfigFunc via context for tests.
+type MockGitConfigKey struct{}
+
+// NewInitCmd builds the `kuke team init` subcommand. It reads the current
+// project's kuketeam.yaml roster, scaffolds the operator-global facts file on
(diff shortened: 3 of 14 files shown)


eminwux:
**PR #1045 Review — `feat(team): kuke team init skeleton + TeamEntry drop-in lifecycle (step 1)`**

Clean structure overall — the drop-in directory pattern, atomic temp+rename, injected `Layout`, and hermetic git-config mock are all solid. Test coverage tracks the AC well. One blocking item: an untrusted-input path-traversal that escapes the drop-in directory.

- **Path traversal** `internal/teamhost/teamhost.go:118` / `internal/kuketeams/parser.go:236`: `WriteEntry` builds the on-disk path with `EntryPath(entry.Metadata.Name)` and `validateTeamEntry`/`validateProjectTeam` only check non-empty. A malicious `kuketeam.yaml` with `metadata: { name: ../kuketeams }` makes the rename target resolve to `~/.kuke/kuketeams.yaml` — the operator's own global facts file gets clobbered with a `TeamEntry` document (verified with `filepath.Join`). `../kuke` similarly overwrites `~/.kuke/kuke.yaml`. The `kuke team init` workflow says "cd into project repo and run init", so the input crosses a trust boundary. Reject names containing `/`, `\`, `..`, or leading `.` in `validateProjectTeam`/`validateTeamEntry` (and defense-in-depth in `WriteEntry`); add a test case at the parser layer. `internal/util/naming/hierarchyname.go` is the closest existing pattern.

Non-blocking nit:

- **Shared identity pointer** `cmd/kuke/team/init.go:223-226`: `seedGit` assigns the same `*v1beta1.GitIdentity` to both `git.Author` and `git.Committer`. Safe today (write → marshal → done), but a future load → mutate → write path in step 2/3/4 would silently couple the two. Construct two values or assign by-value.

LGTM once the path-traversal is bounded.

eminwux:
Addressed in e0d9ab7:

- **Path traversal** (blocking) — added `ErrTeamMetadataNameUnsafe` sentinel and `validateMetadataNameSafe` helper in `internal/kuketeams/parser.go`; called from both `validateProjectTeam` and `validateTeamEntry` so an untrusted `kuketeam.yaml` is rejected before its name reaches host code. Rule rejects `/`, `\`, NUL, `..` substring, or leading `.`. Mirrored as defense-in-depth in `teamhost.WriteEntry` so direct constructors (no parser hop) can't escape either. Parser-layer test cases cover `../kuketeams`, `a/b`, backslash, leading dot on `TeamEntry`, plus the same traversal on `ProjectTeam`. `teamhost`-layer `TestWriteEntryUnsafeNameRefused` pre-seeds the global facts file and verifies five unsafe names (including `..` and `.hidden`) leave it byte-identical.
- **Shared `*GitIdentity` pointer** (non-blocking nit) — `cmd/kuke/team/init.go` `seedGit` now constructs two distinct `*GitIdentity` values for `git.Author` / `git.Committer`. Marshal round-trip behavior is unchanged; the parser-layer git-identity assertions still pass.

`make test` green locally.

eminwux:
**PR #1045 Review — `feat(team): kuke team init skeleton + TeamEntry drop-in lifecycle (step 1)`**

LGTM — the path-traversal guard is layered (parser + `WriteEntry` defense-in-depth) with an explicit clobber sentinel test, the atomic write isolates per-project failure, the seeded `git.Sign` slice respects the parser's signingKey-requires-Sign invariant, and the removed `ErrTeamTeamName*` sentinels have zero surviving refs. License headers, errdefs placement, and colocated tests all match project convention; the drop of `make dev-init` is correct (host-local change, no daemon read path).

Issue #1042, opened 2026-06-03
team: render pipeline (needs-merge, image-select, bind) (step 3)


eminwux:
## Context

Part of the **team-distribution** epic (umbrella: #792). Implementation step 3 of `kuke team init`. With the source materialized (#1041), this step turns the loaded documents into rendered, in-memory CellBlueprint/CellConfig objects ready to apply.

## Proposal

For each `(role × harness)` in the project's roster:

1. **needs-merge** — `union(role.yaml.needs ⊕ kuketeam per-role needs)`.
2. **image select** — match merged capabilities against `ImageCatalog`; **hard-error** naming the unmet capability + a "build/label an image" hint when no image satisfies the needs. Selection only — no Dockerfile build, no LLM (umbrella key decision).
3. **render** — `harnesses/<h>/blueprint.tmpl.yaml × (${ROLE}, merged needs)`, wiring the role's native per-harness config **verbatim** (no transpile — umbrella key decision).
4. **bind** — operator facts (git/registry/secrets from `kuketeams.yaml`) + the project's cloned repo URL → `CellConfig`.
5. **label** — `team=<project>`.

`--dry-run` stops after render and emits to stdout (no apply, no `kuketeam.d/` write).

Net-new template-rendering engine — `git grep` for `text/template`/`tmpl.yaml` blueprint rendering finds nothing today (per dev's triage on #796). If the template engine alone proves large, split render-vs-select into 3a/3b.

## Acceptance criteria

- [ ] Merged needs are `union(role.needs ⊕ per-role needs)` with deterministic ordering.
- [ ] Image selection matches capabilities against `ImageCatalog` and hard-errors clearly (names the unmet capability + build/label hint) when none matches.
- [ ] Blueprint template renders with `${ROLE}` + merged needs; the role's native per-harness config is wired verbatim (claude `settings.json`, codex sandbox/approval knobs, opencode skip).
- [ ] Operator facts + project repo URL bind into `CellConfig`; rendered objects carry `team=<project>`.
- [ ] `--dry-run` renders to stdout and neither applies nor writes `kuketeam.d/<project>.yaml`.
- [ ] No files written under `~/.kuke/rendered/`.

## Dependencies

Depends on: #1041
Blocks: #1043

## Notes

Umbrella #792.

PR #1046, opened 2026-06-03
feat(team): resolve pinned source + cache + load role/harness/images (step 2)

### LINKED ISSUE #1041: team: resolve source (object schema, SSH, ref-aware cache) + load docs (step 2)

eminwux:
## Summary

- Adds `internal/teamsource`: parses a `ProjectTeam.spec.source` pinned-exact `<owner>/<repo>@vX.Y.Z` reference, resolves the clone URL via `TeamsConfig.spec.sources`, materializes the agents repo under `~/.kuke/cache/<owner>/<repo>@<version>/` (clone-once / reuse-on-existing), and loads the `Role`, `Harness`, and `ImageCatalog` documents the project's roster references via the existing `internal/kuketeams` parser. The package is the building block step 3 (#1042) consumes for the render pipeline.
- Atomic clone-then-rename guarantees an interrupted `git clone` never leaves a half-materialized cache dir. `git clone --depth=1 --no-tags --branch <version>` pins the checkout to the requested tag exactly.
- Non-interactive git env guards (`GIT_TERMINAL_PROMPT=0`, `StrictHostKeyChecking=accept-new BatchMode=yes`) mirror `cmd/kuketty/repos.go` so a first-time clone of an unseen host never blocks.
- `teamhost.Layout.CacheDir()` returns `<base>/cache` so the verb wiring in step 3 has a single place to derive the cache root.
- Errdefs added: `ErrTeamSourceURLNotMapped` (unmapped sources key, names the missing key — AC's hard-error contract), `ErrTeamRoleFileKind` / `ErrTeamHarnessFileKind` / `ErrTeamImageCatalogFileKind` (loader-layer kind mismatches; parser-layer parse errors stay wrapped with the offending file path).

## Scope note: no `kuke team init` wiring this step

The AC describes the building block (parse / clone / load), not its placement in the verb. Wiring the materialize step into `composeTeam` would force every existing step-1 test in `cmd/kuke/team/init_test.go` to grow a sources map + git fixture remote even though the entry-write behavior under test is unchanged. Step 3 (#1042) integrates the loader into the render pipeline naturally — the render is the first caller that actually needs the loaded `Role` / `Harness` / `ImageCatalog` documents. Open to flipping this if the reviewer prefers the materialize gate land here.

## Test plan

- [x] `make test-root` — full root-module suite passes (includes the new `internal/teamsource` tests).
- [x] `make test-kukebuild` — passes.
- [x] Unit tests under `internal/teamsource/` exercise the AC end-to-end against a local fixture remote (built with `git init` + commit + lightweight tag): parse-source validation (pinned-exact / floating / bare-tag rejections), `CloneURL` mapped / unmapped / blank-value / nil-TeamsConfig, `Materialize` clones into cache, reuses existing dir (sentinel survives a re-Materialize against a bogus URL), pins to the requested tag exactly (two commits at `v1.4.0` vs `v1.5.0`, the v1.4.0 body — not v1.5.0's — is what lands on disk), missing-tag clone failure leaves no half-materialized dir, `Resolve` loads every roster-referenced Role / Harness + the ImageCatalog, parse errors at the loader / parser layers carry the offending file path.
- [ ] `make dev-init` smoke not run — change is host-side library + tests only; no daemon-read path, `kuke` CLI surface, or `/opt/kukeon/*` layout is touched (the step-1 verb is unchanged this step). The daemon-parity check is unaffected.

Closes #1041

### DIFF
--- a/internal/errdefs/errdefs.go
+++ b/internal/errdefs/errdefs.go
@@ -603,4 +603,22 @@ var (
 	ErrTeamImageCapabilitiesRequired = errors.New(
 		"imageCatalog images[].capabilities must be non-empty",
 	)
+	// ErrTeamSourceURLNotMapped fires when a project's `source` <owner>/<repo>
+	// key has no entry in TeamsConfig.spec.sources. The operator must add a
+	// clone URL mapping in ~/.kuke/kuketeams.yaml before init can materialize
+	// the agents source.
+	ErrTeamSourceURLNotMapped = errors.New(
+		"source key is not mapped in TeamsConfig.spec.sources",
+	)
+	// ErrTeamRoleFileKind fires when a per-role role.yaml parses to a kind
+	// other than Role.
+	ErrTeamRoleFileKind = errors.New("role file must be a Role document")
+	// ErrTeamHarnessFileKind fires when a per-harness harness.yaml parses to a
+	// kind other than Harness.
+	ErrTeamHarnessFileKind = errors.New("harness file must be a Harness document")
+	// ErrTeamImageCatalogFileKind fires when harnesses/images.yaml parses to a
+	// kind other than ImageCatalog.
+	ErrTeamImageCatalogFileKind = errors.New(
+		"image catalog file must be an ImageCatalog document",
+	)
 )
--- a/internal/teamhost/teamhost.go
+++ b/internal/teamhost/teamhost.go
@@ -44,6 +44,9 @@ const (
 	globalConfigName = "kuketeams.yaml"
 	// dropInDirName is the per-project drop-in directory under Base.
 	dropInDirName = "kuketeam.d"
+	// cacheDirName is the materialized-source cache under Base — each pinned
+	// `<owner>/<repo>@vX.Y.Z` agents reference clones into its own subdirectory.
+	cacheDirName = "cache"
 
 	// dropInDirPerm is the drop-in directory mode: operator-only (the files
 	// reference secret sources and signing keys).
@@ -78,6 +81,11 @@ func (l Layout) DropInDir() string {
 // EntryPath is the per-project file (<base>/kuketeam.d/<project>.yaml).
 func (l Layout) EntryPath(project string) string {
 	return filepath.Join(l.DropInDir(), project+".yaml")
+}
+
+// CacheDir is the materialized-source cache root (<base>/cache).
+func (l Layout) CacheDir() string {
+	return filepath.Join(l.Base, cacheDirName)
 }
 
 // EnsureGlobalConfig writes cfg to the global facts path only when no file is
--- a/internal/teamsource/teamsource.go
+++ b/internal/teamsource/teamsource.go
@@ -1,0 +1,346 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package teamsource resolves a kuketeams.io/v1 ProjectTeam's pinned `source`
+// reference to a local clone of the agents repo and loads the Role, Harness,
+// and ImageCatalog documents the project's roster references (epic #792, step
+// 2 #1041). Materialization is cache-keyed by the pinned
+// `<owner>/<repo>@vX.Y.Z` reference: an existing cache directory at the pinned
+// version is reused without re-cloning, and a missing cache is cloned via
+// `git clone --depth=1 --branch <version> --no-tags` into the pinned ref
+// exactly. Document loading delegates to the existing kuketeams parser, so
+// parse errors carry the offending file path verbatim. No daemon is required —
+// the package is unit-testable against a local fixture remote.
+package teamsource
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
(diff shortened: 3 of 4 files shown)


eminwux:
**PR #1046 Review — `feat(team): resolve pinned source + cache + load role/harness/images (step 2)`**

Solid building-block PR. Atomic clone-then-rename, version-pinned cache key, and non-interactive git env all match the contract in the package doc; tests exercise every exported entry point including the failed-clone-leaves-no-dir invariant and the v1.4.0-vs-v1.5.0 pinning check. License headers / errdefs / colocated tests all per project conventions. Commits verified.

One non-blocking observation:

- **Error-message hygiene** `internal/teamsource/teamsource.go:327-329`: `runGit` joins the full `args` list — including `cloneURL` — into the error wrapper. If an operator ever embeds credentials in a `TeamsConfig.spec.sources` URL (e.g. `https://x-token:ghp_xxx@…`), a failed clone surfaces the token via the error chain. SSH is the common case so this is conditional, but stripping userinfo before joining (or omitting `args` and relying on the trimmed git stderr) would close the surface. Not blocking — flagging for awareness as step 3 wires this into a user-facing CLI path.

LGTM — the `Layout.CacheDir()` and `teamsource` exports having no non-test callers in this PR is the explicit step-3 (#1042) scope split, per the PR's own Scope note.

Issue #1038, opened 2026-06-03
kukebuild: build leaves orphaned containerd leases (buildkit.temporary + per-build gc.flat) pinning dangling layers


eminwux:
## Summary

`kukebuild` / the build path leaves containerd leases behind after each build. In `kuke-system.kukeon.io` there are 5 lingering `buildkit/lease.temporary` leases and 187 per-build `containerd.io/gc.flat` roots for only 2 tagged images — these are GC roots, so they pin dangling build layers and content blobs against garbage collection indefinitely.

## Evidence

`sudo ctr -n kuke-system.kukeon.io leases ls` → 266 leases, oldest 2026-05-22, newest 2026-06-01 (one cluster per build). Only 2 images tagged. The pull path already releases its lease (`internal/ctr/image.go:125-129`), but the build path does not.

## Proposed behavior

- Ensure each build releases its temporary BuildKit lease(s) and any per-build `gc.flat` root on completion (success **and** failure paths).
- Optionally expose / wire up BuildKit build-cache pruning.

## Acceptance criteria

- After N sequential `kuke build` runs into a realm, the realm's lease count does not grow unboundedly (returns to baseline once builds complete).
- No leftover `buildkit/lease.temporary` leases after a successful build.

## Related

- #1036 (`kuke image prune`), #1037 (DeleteImage GC). This issue stops the bleed at the source; #1036 cleans up what already accumulated.

PR #1050, opened 2026-06-04
fix(ctr): DeleteImage triggers synchronous GC sweep to reclaim exclusive layers

### LINKED ISSUE #1037: ctr: DeleteImage does not reclaim exclusive layers / trigger GC (deleting an image frees no disk)
## Summary

`DeleteImage` (`internal/ctr/image.go:247-254`) only removes the image **tag/ref** via `ImageService().Delete`. The image's exclusive layers stay pinned by leftover leases and become dangling, so deleting an image reclaims **no disk**. Deleting an image should also free its exclusive layers.

## Current behavior

```go
// internal/ctr/image.go:247-254 (approx)
func (c *client) DeleteImage(...) error {
    if err := c.conn().ImageService().Delete(nsCtx, ref); err != nil {
        ...
    }
    // tag gone; exclusive content blobs + committed-layer snapshots remain,
    // pinned by orphaned leases — no GC pass, no disk reclaimed
}
```

## Proposed behavior

After the ref is removed:
1. Release the lease(s) that pinned this image's pull/unpack (if not shared with a still-tagged image).
2. Trigger a content/snapshot GC pass so layers exclusive to the deleted image are reclaimed.

Layers shared with other tagged images or live containers must remain (GC handles this naturally once the lease roots are dropped).

## Acceptance criteria

- `kuke image delete <ref>` on an image with exclusive layers reclaims those layers' disk.
- Shared layers (referenced by another tagged image / live cell) are preserved.
- Test: build two images sharing a base, delete one, assert only the non-shared layers are gone.

## Related

- #1036 (`kuke image prune` — the bulk/operator-facing cleanup), #1038 (kukebuild lease release)

eminwux:
## Summary

- `DeleteImage` only removed the metadata image bucket entry — exclusive layers stayed pinned by GC references the deleted image was their last root for, so `kuke image delete <ref>` freed no disk (#1037).
- Pass `images.SynchronousDelete()` so the containerd image-service handler runs the GC scheduler's `ScheduleAndWait` sweep after the metadata removal (containerd v2 `plugins/services/images/local.go` Delete). The sweep walks every surviving GC root (live images, leases, container snapshots) and reclaims content/snapshots no remaining root references.
- Layer-sharing preservation (AC item 2) falls out of containerd's GC refcount: layers still referenced by another tagged image or a live container survive the sweep on their own root, no per-layer accounting on our side. Mirrors the `leases.SynchronousDelete` pattern `drainLeases` already uses for the same purpose (see `CleanupNamespaceResources`'s GC-sweep rationale).
- Refactored the body into a private `deleteImage(nsCtx, namespace, ref, store)` helper that takes `images.Store` so the regression test can drive the path with a fake — matches the existing `drainNamespaceResources` / `containerdNamespaceServices` seam pattern in this package.
- Out of scope: the pull-time lease `pullImage` creates is already dropped by its own `defer`, so the regular pull path leaves no orphaned lease for this sweep to step around. Kukebuild's build-time orphaned leases are tracked separately in #1038 and would survive this sweep regardless — that issue is the right place to fix lease cleanup at the producer.

## Test plan

- [x] `GOWORK=off go test ./internal/ctr/...` — 71 s, all passing (`TestDeleteImagePassesSynchronousDelete` and `TestDeleteImageNotFoundMapsToSentinel` are the new regression guards)
- [x] `make test` — full per-module CI target, all packages pass
- [x] `pre-commit run --files internal/ctr/image.go internal/ctr/image_internal_test.go internal/ctr/namespaces_internal_test.go` — clean (golangci-lint, gofmt, addlicense, etc.)
- [ ] e2e shared-layer assertion (AC item 3: "build two images sharing a base, delete one, assert only the non-shared layers are gone") — this repo has no `kuke image` e2e suite today (no test in `e2e/` references `kuke image`, `kuke build`, or image delete), and the assertion needs a real containerd + `kukebuild` to produce shared layers. Per CLAUDE.md §"When an AC requires a test tier the project can't execute anywhere", shipping the verifiable fix and surfacing the infra gap for PM follow-up rather than blocking on a missing harness. The shared-layers-survive guarantee itself is containerd's GC refcount (covered by upstream containerd's own tests); the contract dev controls — passing `images.SynchronousDelete()` — is asserted by the new unit test.

Closes #1037

### DIFF
--- a/internal/ctr/image.go
+++ b/internal/ctr/image.go
@@ -17,11 +17,13 @@
 package ctr
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"time"
 
 	containerd "github.com/containerd/containerd/v2/client"
+	"github.com/containerd/containerd/v2/core/images"
 	"github.com/containerd/containerd/v2/core/leases"
 	"github.com/containerd/containerd/v2/defaults"
 	"github.com/containerd/errdefs"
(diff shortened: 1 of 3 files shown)


eminwux:
**PR #1050 Review — `fix(ctr): DeleteImage triggers synchronous GC sweep to reclaim exclusive layers`**

LGTM — `images.SynchronousDelete()` is the correct knob for the #1037 reclaim gap, mirrors the established `leases.SynchronousDelete` pattern in `drainLeases`, and the new unit tests assert both the load-bearing opt threading and the preserved `ErrImageNotFound` mapping. Deferred shared-layer e2e is reasonable — containerd's GC refcount is what guarantees AC item 2, and the project has no `kuke image` e2e harness yet.

Issue #1036, opened 2026-06-03
kuke image prune: reclaim dangling image layers + orphaned leases (containerd GC)


eminwux:
## Summary

Add a `kuke image prune` verb (and the underlying GC) that reclaims dangling image layers and the containerd leases pinning them. Today, repeated `kuke build` / image pulls into a realm accumulate leases and committed layers without bound, and the only reclaim path is deleting the entire namespace (`CleanupNamespaceResources` on uninstall/realm purge).

## Evidence

In `kuke-system.kukeon.io`: **266 leases** (187 `containerd.io/gc.flat`, 5 `buildkit/lease.temporary`, + variants) dated across every dev-init run (May 22 → Jun 1), and **157 committed snapshots** (image layers) — backing only **2 tagged images** (busybox, kukeon-local:dev). ~120 of those layers are dangling, kept alive only because the orphaned leases are GC roots. `default.kukeon.io` shows the same pattern at smaller scale (49 `gc.flat`).

## Proposed behavior

`kuke image prune [--realm <r>]`:
1. Identify dangling images (untagged / no longer referenced).
2. Delete leases that no longer back a tagged image or a live container.
3. Trigger containerd GC to reclaim now-unreferenced content blobs + committed snapshots.

Reuse the `leases.SynchronousDelete` + drain machinery already in `internal/ctr/namespaces.go` (currently namespace-wide on delete) — but **surgically**, never touching leases/snapshots that back a tagged image or a live cell.

## Acceptance criteria

- After repeated builds, `kuke image prune` reduces the realm's lease count to ≈ the number of live images and reclaims dangling-layer disk.
- Tagged images and snapshots backing running/stopped cells are untouched.
- Idempotent: a second run on a clean realm is a no-op.

## Related

- #1037 (DeleteImage should reclaim layers/trigger GC), #1038 (kukebuild lease release) — this verb is the operator-facing cleanup; those are the per-operation fixes.

Issue #522, opened 2026-05-15
build: add kuke build + kukebuild binary (BuildKit lib, phase 1)


eminwux:
## Context

Today, anyone who needs to produce an OCI image that `kuke image load` and
`kuke create` will consume has to install docker (or nerdctl + a separate
buildkitd) just for the build step. This breaks kukeon's "self-contained
runtime, no docker dep" pitch at the most visible point — the contributor's
first `make dev-init` and any operator's first image. The dev-environment
hit is the same: `scripts/dev-init.sh` calls `docker build`, and CLAUDE.md's
"Local smoke test" lists the docker daemon as a prerequisite.

The fix is a native build path: a separate `kukebuild` binary that embeds
BuildKit as a library and writes results directly into the target realm's
containerd content store. `kuke build` becomes a thin CLI that shells out
to `kukebuild` on the host.

This supersedes #310 (closed `wontfix`), which proposed swapping `docker
build` for `nerdctl build` in `dev-init.sh`. That trade keeps an external
builder dep — this issue removes the dep entirely.

## Proposal

Add a new top-level binary `kukebuild`, sibling to `kuke` / `kukeond` /
`kuketty`, that embeds BuildKit as a library, plus a thin `kuke build`
CLI shim that exec's it.

This phase (1) lands the binary + CLI surface + BuildKit wiring + a
single-stage end-to-end smoke (trivial Dockerfile) — enough to prove
the solver / frontend / containerd worker / image exporter / cache
store all talk to each other and produce an image consumable by
`kuke image get` and `kuke create`. Multi-stage validation against
kukeon's own Dockerfile (CNI-plugins curl-fetch, `COPY --from=`,
`--build-arg`, sha256 verification) is deferred to phase 1b (#616).

- `cmd/kukebuild/main.go` — own `main` package, own Makefile target,
  listed in the `BINS` variable.
- **Not argv[0]-dispatched** from `kuke`. Mirrors the kuketty rationale
  (see `cmd/kuketty/main.go:24-28`): keep BuildKit's transitive moby
  closure entirely out of the `kuke` and `kukeond` import sets.
- Uses BuildKit's **containerd worker** backend pointed at the host's
  containerd socket (`/run/containerd/containerd.sock`). No second image
  store, no long-running embedded buildkitd — `kukebuild` spawns on
  demand from `kuke build` and exits when the build completes.
- Writes layers + manifest into the target realm's containerd namespace
[...]

PR #1051, opened 2026-06-04
fix(kukebuild): drain buildkit/lease.temporary leases at shutdown

### LINKED ISSUE #1038: kukebuild: build leaves orphaned containerd leases (buildkit.temporary + per-build gc.flat) pinning dangling layers

eminwux:
## Summary

- `kukebuild` embeds BuildKit in a one-build-per-process model (#522), so the worker dies after Solve returns. BuildKit's solver tags transient solve-time leases with `buildkit/lease.temporary` and relies on the *next* worker startup to sweep them (`worker/base/worker.go:214-220` runs the exact List+Delete pass on every boot). With no long-running buildkitd, that next-startup sweep never runs and the temp leases pile up as GC roots that pin dangling content + snapshots against containerd GC indefinitely (#1038).
- Mirror BuildKit's startup sweep at shutdown: `newController`'s cleanup defer now calls `drainBuildkitTemporaryLeases(ctx, w.LeaseManager())` before closing the worker, on both success and failure paths. Uses `leases.SynchronousDelete` so containerd's GC scheduler runs `ScheduleAndWait` before each Delete returns — same pattern `internal/ctr/namespaces.go` `drainLeases` already uses for the same reason.
- `context.WithoutCancel(ctx)` decouples the drain from the parent — the cleanup may fire with ctx already cancelled (SIGINT/SIGTERM through `runBuild`'s signal handler) and the deletes still need to reach containerd.

## Scope call

The issue title and Summary mention both `buildkit/lease.temporary` (5 lingering) and per-build `containerd.io/gc.flat` (187 roots). Acceptance criterion #2 names *only* temporary leases, and AC #1 (count doesn't grow unboundedly) is satisfied by removing the orphan source — `gc.flat` leases are BuildKit's cache pins (`cache/manager.go:243`, `:600`, `:805`, `:960`, `:410`, `:1494`), created per cache record so containerd doesn't GC the blob/snapshot the record references. They persist by design across builds; draining them would defeat caching (every build re-pulls layers). The issue's own References section points at #1036 (`kuke image prune`) as the cleanup path for accumulated state — \"this issue stops the bleed at the source; #1036 cleans up what already accumulated.\" Per dev/CLAUDE.md \"lower-blast-radius default when the AC leaves room\", this PR ships the temp-lease drain only; #1036 owns cache pruning.

## Test plan

- [x] `cd cmd/kukebuild && go test ./...` — 3 new unit tests (`TestDrainBuildkitTemporaryLeases{ScopesToLabelAndSynchronous,IgnoresListError,SurvivesCancelledCtx}`) cover the four invariants: exact label filter `labels.\"buildkit/lease.temporary\"`, `SynchronousDelete` flag, `gc.flat` cache pins preserved, and survival under cancelled parent ctx
- [x] `make test` — full per-module CI target, all packages pass (kukebuild + root tree)
- [x] `cd cmd/kukebuild && go vet ./...` — clean
- [x] `pre-commit run --files cmd/kukebuild/build.go cmd/kukebuild/build_test.go` — clean (golangci-lint, gofmt, addlicense, etc.)
- [ ] End-to-end \"lease count returns to baseline after N builds\" smoke against a real containerd — not runnable on this dev host without disturbing the standalone containerd serving the parent session, and the project has no `kuke build` e2e suite that would exercise the lease lifecycle today. Per CLAUDE.md §\"When a reviewer-blocking verification step can't safely run on the current host\", deferring to a clean reviewer environment; the unit-test layer pins the contract (filter, sync flag, gc.flat preservation) so the reviewer-side smoke only needs to confirm the observable lease count, not the implementation shape.

Closes #1038

### DIFF
--- a/cmd/kukebuild/build.go
+++ b/cmd/kukebuild/build.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	ctd "github.com/containerd/containerd/v2/client"
+	"github.com/containerd/containerd/v2/core/leases"
 	ctddefaults "github.com/containerd/containerd/v2/defaults"
 	"github.com/distribution/reference"
 	"github.com/moby/buildkit/cache/remotecache"
@@ -399,10 +400,35 @@ func newController(ctx context.Context, cfg *buildConfig, namespace string) (*co
 	}
 
 	cleanup := func() {
+		drainBuildkitTemporaryLeases(ctx, w.LeaseManager())
 		_ = historyDB.Close()
 		_ = wc.Close()
 	}
 	return ctrl, cleanup, nil
+}
+
+// drainBuildkitTemporaryLeases drops every `buildkit/lease.temporary` lease
+// from the worker's namespace. BuildKit's solver tags transient solve-time
+// leases with this label and relies on the next worker startup to sweep them
+// — `base.NewWorker` runs the exact List+Delete pass on every boot. Kukebuild
+// embeds BuildKit in a one-build-per-process model (#522), so the worker dies
+// after Solve returns and the next-startup sweep never runs, leaving the
+// temporary leases as GC roots that pin dangling content + snapshots against
+// containerd GC indefinitely (#1038). Mirrors the startup sweep at shutdown
+// so each kukebuild invocation cleans up after itself, on both success and
+// failure paths via the newController cleanup defer. ctx may already be
+// cancelled (SIGINT/SIGTERM through runBuild's signal handler) when cleanup
+// runs, so the drain uses a non-cancelable context to ensure the deletes
+// reach containerd.
+func drainBuildkitTemporaryLeases(ctx context.Context, lm leases.Manager) {
+	ctx = context.WithoutCancel(ctx)
+	existing, err := lm.List(ctx, `labels."buildkit/lease.temporary"`)
+	if err != nil {
+		return
+	}
+	for _, l := range existing {
+		_ = lm.Delete(ctx, l, leases.SynchronousDelete)
+	}
 }
 
 // newWorkerController builds a single-worker controller using BuildKit's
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1051 Review — `fix(kukebuild): drain buildkit/lease.temporary leases at shutdown`**

LGTM — mirrors BuildKit's own startup sweep at shutdown for the one-build-per-process model, uses the same `SynchronousDelete` primitive as `internal/ctr/namespaces.go drainLeases`, `context.WithoutCancel` decouples from SIGINT, and unit tests pin the label filter, sync flag, and `gc.flat` preservation. Scope call (temp leases only, defer cache-pin cleanup to #1036) is justified.

PR #1059, opened 2026-06-04
fix(teamsource): strip URL credentials from runGit error wrapper

### LINKED ISSUE #1047: followup-nit(teamsource): strip credentials from cloneURL in runGit error wrapper
From PR #1046. `internal/teamsource/teamsource.go:327-329` — `runGit` joins the full `args` list (including `cloneURL`) into the error wrapper, so a failed clone surfaces the raw URL through the error chain.

If an operator embeds basic-auth credentials in a `TeamsConfig.spec.sources` URL (e.g. `https://x-token:ghp_xxx@github.com/owner/agents.git`), any failed clone (missing tag, network error, auth error) leaks the token through whatever surface step 3 exposes (CLI stderr, log, daemon journal).

Suggested fix: parse `cloneURL` and strip the userinfo component before joining into the error message, or omit `args` from the wrapper entirely and rely on the trimmed git stderr (which usually elides credentials).

SSH is the common case so this is conditional, not an active leak — flagged in PR #1046 review for awareness as step 3 (#1042) wires this into a user-facing path.

eminwux:
## Summary
- Scrub `userinfo` from URL-shaped args before joining them into `runGit`'s error wrapper, so a failed clone against a TeamsConfig source URL with embedded basic-auth credentials (e.g. `https://x-token:ghp_xxx@host/repo.git`) never leaks the token through whatever surface step 3 exposes (CLI stderr, log, daemon journal).
- New `scrubURLCredentials` helper uses `net/url`: an arg is rewritten only when it parses as an absolute URL (non-empty `Scheme`) and carries `userinfo`. Non-URL args (flags like `--depth=1`, version refs like `v1.4.0`, SCP-style refs like `git@github.com:owner/repo.git`) pass through unchanged.
- Stripping is total (`u.User = nil`) — drops both username and password rather than only the password, since usernames in tokenized URLs are routinely the secret-bearing field (`x-access-token`, `oauth2`) and there is no diagnostic value to preserving them in error text.

## Test plan
- [x] `make test` (root + kukebuild modules) — all green.
- [x] New `TestScrubURLCredentials` table covers: basic-auth + token, user-only, ssh-with-password, plain https (no userinfo, unchanged), SCP-style (unchanged), flag (unchanged), branch name (unchanged), `file://` URL (unchanged).
- [ ] Integration test driving `runGit` to a real failure with embedded credentials — deliberately omitted: git's stderr URL redaction varies by version (some versions print the user portion alongside an `x-oauth-basic`-style redacted password), which would make the assertion git-version-dependent without strengthening the scrub guarantee. The unit test covers the scrub logic; the wiring into `runGit` is a one-line call right at the wrapper site.

Closes #1047

### DIFF
--- a/internal/teamsource/teamsource.go
+++ b/internal/teamsource/teamsource.go
@@ -30,6 +30,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1059 Review — `fix(teamsource): strip URL credentials from runGit error wrapper`**

LGTM — targeted defense-in-depth on the args portion of `runGit`'s error wrapper, with thorough unit coverage (token, user-only, ssh+password, plain https, SCP-style, flag, branch, file URL) and the outer clone wrapper already using `src.OwnerRepo` rather than the credential-bearing URL.

PR #1078, opened 2026-06-05
feat(team): kuke team init --build — FROM-walk + per-leaf build (step 2)

### LINKED ISSUE #1064: team: kuke team init --build — FROM-walk + per-leaf build (step 2)
## Context

`kuke team init` today is select-and-bind only: it matches a role's merged `needs.image` capabilities against the agents `images.yaml` `ImageCatalog`, picks an entry, and binds `entry.image` verbatim into the cell blueprint — assuming that ref is already published and pullable. There is no way to build a harness image locally for the fast dev loop, the way `make dev-init` builds a local kukeond image and `kuke init --kukeond-image` consumes it.

This is **path 2 (local build)** of the build-and-supply epic. The published-image supply side (path 1) is the agents-repo CI step (#699).

This step covers the **build-invoke** half: compute the build set and build each image into the realm namespace. The **bind decision, local-only resolve, and two-project e2e** are split into step 3 (#1068), along the issue's own "build-invoke vs bind/e2e" seam.

_Umbrella: https://github.com/eminwux/kukeon/issues/1063_

## Proposal

Add the build-invoke half of a `--build` mode to `kuke team init`:

- **`--build` flag threaded through the lifecycle.** Add `--build` on `kuke team init` and thread it through `composeTeam → renderTeam → teamrender.Render` (`cmd/kuke/team/init.go`).
- **Build set via FROM-walk.** For each **selected** catalog entry, take `build.{context,dockerfile}`, then scan the harness Dockerfiles in the cloned agents `source` tree and parse each `FROM` to compute the transitive base-before-leaves set (the bases `base`/`base-user`/`base-root` are **not** in `images.yaml` — they are intermediates). This is the ordering `harnesses/build-images.sh` already implements (`local_base` + `add_with_bases`).
- **Build via kukebuild into the realm namespace.** Build each image in dependency order via a **new programmatic kukebuild invocation** — an exec-**and-return** loop, distinct from the existing `kuke build` CLI shim that `syscall.Exec`'s kukebuild and replaces the process (`cmd/kuke/build/build.go` `execHandoff`). Build into the **target realm's containerd namespace** — no registry push — tagged `kukeon.internal/<name>:<version>`.
- **Thread the `REGISTRY` build-arg.** The leaf Dockerfiles are `FROM ${REGISTRY}/base-user:latest`. Pass `--build-arg REGISTRY=kukeon.internal` (and tag `-t kukeon.internal/<name>:<version>`) so each leaf's `FROM` resolves to the **in-realm** base just built, rather than pulling an external registry. Without this threading the leaf build silently pulls the base from the network.

`kukeon.internal` is a reserved, non-routable internal host (ICANN-reserved `.internal` private-use TLD): the image always exists in the realm namespace before it is referenced, and a stray pull attempt fails fast and unambiguously instead of hanging on a real registry. Mirrors `kuke init` (published ghcr) vs `kuke init --kukeond-image <kukeon.internal/kukeond>` (locally built, step 5).

## Acceptance criteria

- [ ] `kuke team init --build` computes the build set by walking the harness Dockerfile `FROM`s in the cloned source (not from `images.yaml`, which omits the bases), in base-before-leaves order.
- [ ] Each image is built from `build.{context,dockerfile}` via a programmatic kukebuild invocation (exec-and-return, distinct from the `syscall.Exec` `kuke build` shim) into the target realm's containerd namespace; no registry push.
- [ ] `--build-arg REGISTRY=kukeon.internal` is threaded and images are tagged `kukeon.internal/<name>:<version>` so leaf `FROM`s resolve to the in-realm bases.
- [ ] Unit tests cover FROM-walk base-before-leaves ordering against a fixture source tree and the build-invoke argv composition.

## Dependencies

Depends on: none
Blocks: #1068

## Notes

Build-invoke half (AC 1–3 of the original #1064) of path 2; the bind/resolve/e2e half (original AC 4–7) is step 3 (#1068), which depends on this step's in-realm images. Independent of the agents CI-publish step (#699, path 1). **Supersedes #1062** — the dangling `TeamsConfig.Spec.Registry` question is resolved by the `kukeon.internal` scheme: local refs stay registry-qualified, so no default-registry host and no unqualified-ref path is needed. Shares the `kukeon.internal` convention with the agents Dockerfile-default flip (step 4, agents#700) and the kukeond local-ref change (step 5, #1065). This split realizes the 2a/2b seam the issue body pre-authorized and #1042 flagged.

eminwux:
## Summary
- Adds `internal/teambuild` package: walks Dockerfile `FROM`s in the materialized agents source to compute a base-before-leaves build plan, then invokes `kukebuild` per step into the target realm's containerd namespace (exec-and-return via `os/exec`, distinct from `cmd/kuke/build`'s `syscall.Exec` shim).
- Threads `--build-arg REGISTRY=kukeon.internal` through every step; in-repo bases and leaves are tagged `kukeon.internal/<name>:<version>`, no registry push.
- Wires a new `--build` flag through `kuke team init` → `composeTeam` → render → build. `--dry-run --build` announces the leaf count but does not invoke `kukebuild`; standard runs invoke after `Render` and before `Apply`.
- Adds `teamrender.Result.Selections` (dedup'd by entry.Ref) so two roles selecting the same image build once.
- Adds three errdefs sentinels: `ErrTeamBuildContextMissing`, `ErrTeamBuildBaseMissing`, `ErrTeamBuildCycle`.

## Design notes
- Build is invoked at the CLI layer (between Render and Apply) rather than inside `renderTeam`, so the test seam stays a single `BuildAllFunc` package var (`MockBuildAllKey` context value) and dry-run can announce without spawning `kukebuild`.
- Lower-level unit-test seams (`lookPath`, `runStep` package vars in `teambuild`) cover argv composition and ordering without a real `kukebuild` on PATH. Tests that mutate these seams do not run `t.Parallel()`.
- Topological sort uses lexicographic tiebreak for byte-stable build-step ordering.

## Test plan
- [x] `make test` — 104 packages OK (including new `internal/teambuild`).
- [x] `gofmt -l` clean on all modified/created files.
- [x] `git grep -n` swept for removed/renamed literals (none in this diff).
- [ ] `pre-commit run --files …` — host has no network reachability to `github.com` for hook fetch; covered by `gofmt -l` + `make test`.
- [ ] `make dev-init` end-to-end smoke — this PR does not change build paths, daemon code, or `/opt/kukeon` layout; only adds a new package + a new opt-in CLI flag wired through team init. The two-realm regression tail is unchanged.

Closes #1064

### DIFF
--- a/cmd/kuke/team/init.go
+++ b/cmd/kuke/team/init.go
@@ -30,6 +30,7 @@ import (
 	kukshared "github.com/eminwux/kukeon/cmd/kuke/shared"
 	"github.com/eminwux/kukeon/internal/errdefs"
 	"github.com/eminwux/kukeon/internal/kuketeams"
+	"github.com/eminwux/kukeon/internal/teambuild"
 	"github.com/eminwux/kukeon/internal/teamhost"
 	"github.com/eminwux/kukeon/internal/teamrender"
 	"github.com/eminwux/kukeon/internal/teamsource"
@@ -92,6 +93,23 @@ type ApplyForTeamFunc func(
 // MockApplyForTeamKey injects an ApplyForTeamFunc via context for tests.
 type MockApplyForTeamKey struct{}
 
+// BuildAllFunc runs the local-build path for `kuke team init --build`: it
+// walks the FROM directives of the selected catalog leaves under cacheDir,
+// derives the base-before-leaves order, and invokes kukebuild once per
+// image into the target realm's containerd namespace (no registry push).
+// The default implementation delegates to teambuild.BuildAll; tests inject a
+// stub via MockBuildAllKey so the build path can run without a kukebuild
+// binary on PATH.
+type BuildAllFunc func(
+	ctx context.Context,
+	cacheDir, sourceRef, realm string,
+	leaves []*model.ImageCatalogEntry,
+	progressW, stdout, stderr io.Writer,
+) error
+
+// MockBuildAllKey injects a BuildAllFunc via context for tests.
+type MockBuildAllKey struct{}
+
 // NewInitCmd builds the `kuke team init` subcommand. It reads the current
 // project's kuketeam.yaml roster, scaffolds the operator-global facts file on
 // first run, materializes the pinned agents source, renders the per-(role ×
@@ -111,13 +129,21 @@ func NewInitCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			return runInit(cmd, dryRun)
+			build, err := cmd.Flags().GetBool("build")
+			if err != nil {
+				return err
+			}
+			return runInit(cmd, dryRun, build)
 		},
 	}
 
 	cmd.Flags().Bool(
 		"dry-run", false,
 		"Render the project's blueprints/configs to stdout without writing the drop-in entry or applying to kukeond",
+	)
+	cmd.Flags().Bool(
+		"build", false,
+		"Locally build the selected catalog images (kukebuild) into the target realm's containerd namespace; no registry push",
 	)
 
 	return cmd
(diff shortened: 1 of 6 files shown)


eminwux:
**PR #1078 Review — `feat(team): kuke team init --build — FROM-walk + per-leaf build (step 2)`**

LGTM — Plan's iterative FROM-walk with dedup-by-name, lex-tiebreak topo sort, and seeded `REGISTRY=kukeon.internal` build-arg cleanly produces base-before-leaves ordering; `BuildAllFunc` test seam mirrors the existing `ApplyForTeamFunc` shape, all three errdefs sentinels are wired (`Plan` for both Missing flavors, `topoSort` for cycle), and `composeTeam` correctly guards the build path on `bundle != nil && len(Selections) > 0` so the harness-less roster and dry-run branches both short-circuit before invoking `kukebuild`. Test plan host-gated carve-outs (pre-commit network fetch, `make dev-init`) are documented and consistent with the scope claim — this PR adds a new opt-in CLI flag + package, doesn't touch the daemon bootstrap or `/opt/kukeon` layout.

PR #1082, opened 2026-06-05
feat(repos): add ContainerRepo.ref for immutable tag/commit pins

### LINKED ISSUE #1034: ContainerRepo: add explicit tag/commit pinning + make fetch path ref-aware (tag pins break on restart)
## Summary

`ContainerRepo` (the `spec.cell.containers[].repos[]` clone sugar from #617/#618) can only express a **branch**. There is no first-class way to pin a clone to a **tag** or **commit**, and the one workaround that exists — putting a tag name in `branch:` — silently breaks on the fetch/refresh path.

Surfaced while pinning crew's fleet-wide `agents` clone to a release tag (`branch: v0.1.0`).

## Current behavior

`ContainerRepo` schema (`docs/site/manifests/container.md`) exposes only:

| Field | Description |
|---|---|
| `branch` | Branch to check out. Empty clones the remote's default branch. |

There is no `tag`/`commit`/`ref` field.

### What works (fresh clone)

`cmd/kuketty/repos.go:cloneRepo` (~L119–127) runs:

```
git clone --branch <branch> <url> <target>
```

`git clone --branch` happens to accept a **tag** name too, leaving a detached HEAD at the tag. So on a *fresh* cell, `branch: v0.1.0` pins correctly — but only by accident of `--branch`'s dual meaning, not by design.

### What breaks (restart-in-place / refresh)

When the target rootfs already has a checkout (cell restarted without recreation), `resolveRepo` takes the fetch path, `cmd/kuketty/repos.go:fetchRepo` (~L132–142):

```
git -C <target> fetch --prune origin
git -C <target> checkout <branch>      # detached HEAD at the tag
git -C <target> pull --ff-only         # <-- FAILS on detached HEAD
```

`git pull --ff-only` on a detached HEAD (a tag) has no upstream branch and errors (`You are not currently on a branch...`). Because crew's `agents` repo is `required: true`, that failure aborts container start. Net effect: **a tag pin only survives a fresh clone; any in-place restart of an already-cloned cell fails to boot.** The clean workaround today is to force recreation (`DELETE_CELLS=1`) so the clone path runs instead of the fetch path — which defeats the point of cheap restarts.

## Proposed solution

1. **Add an explicit immutable-ref field** to `ContainerRepo` — e.g. `ref` (accepting a tag or full SHA), or distinct `tag:`/`commit:` fields — alongside the existing `branch:`. `branch` keeps its moving-target semantics; `ref`/`tag`/`commit` is understood to be immutable.

2. **Make the fetch path ref-aware** in `fetchRepo`. For an immutable pin:
   - `git fetch --prune --tags origin`
   - `git checkout --detach <resolved-ref>`
   - **skip `git pull --ff-only`** entirely — a pinned tag/commit is immutable, so there is nothing to fast-forward. (Optionally short-circuit to a no-op when `rev-parse HEAD` already equals the resolved ref.)

   This makes restart-in-place idempotent for pinned repos instead of fatal.

3. Update `docs/site/manifests/container.md` to document the new field and the branch-vs-ref distinction.

## Acceptance

- A `repos[]` entry pinned to a tag/commit clones at detached HEAD on first boot **and** survives an in-place cell restart (no recreation needed) without a `required`-repo failure.
- Branch-tracking repos keep today's fetch + fast-forward behavior.

## Context / references

- Introduced in #617 (phase 1a: `containers[].repos[]` with kuketty pre-Serve clone).
- Consumer: crew blueprints pin the `agents` clone via `branch: <tag>` as a stopgap; once this lands, crew can switch to the explicit field.

eminwux:
## Summary

- Adds a first-class `ref` field on `ContainerRepo` (tag name or full commit SHA) alongside the existing `branch:`, so a clone can be pinned to an immutable revision rather than abusing `--branch` for tags.
- Rewrites `cmd/kuketty/repos.go` so the fetch path is ref-aware: a Ref-pinned repo runs `git fetch --prune --tags origin` + `git checkout --detach <ref>` and **skips `git pull --ff-only`** — the prior `pull --ff-only` on a detached HEAD errored (`You are not currently on a branch...`) and aborted required-repo container start, so a tag pin only survived a fresh clone, not an in-place restart (#1034).
- Wires Ref through the v1beta1 + internal model, `CellConfigRepoFill` (so a CellConfig can also pin a slot to a ref), the apischeme conversions, and `cellconfig.Materialize`.
- Apply-time validation rejects setting both `branch` and `ref` on the same repo — they have different fetch semantics, so the spec must pick one.
- Updates `docs/site/manifests/container.md` (and the matching `RepoFill` section in `config.md`) with the new field and the branch-vs-ref distinction.

## Test plan

- [x] `make test` (full `test-root` + `test-kukebuild`).
- [x] `GOWORK=off go vet ./...`.
- [x] New tests:
  - `TestProcessRepos_RefPinsCloneAndSurvivesRestart` — clones at the tag's detached HEAD, then re-runs the fetch path after a new upstream commit and asserts the pin holds (this is the AC regression — would fail against `main`'s `pull --ff-only`).
  - `TestProcessRepos_RefPinAcceptsCommitSHA` — same idempotence for a full commit SHA.
  - `TestValidateDocument_Container_RepoValidation` adds a `valid ref pin` row and a `branch and ref both set` row asserting `ErrRepoBranchRefMutex`.
  - `TestValidateDocument_Blueprint_RepoSlotBranchAndRefMutuallyExclusive` mirrors the gate for blueprint repo slots.
  - `TestMaterialize_RepoFillRefCarriesThrough` confirms a `CellConfigRepoFill.ref` is propagated into the materialized `ContainerRepo.Ref`.
- [x] `pre-commit run --files $(git diff --name-only HEAD~)` — gofmt, go-mod-tidy, golangci-lint, prettier, addlicense all green.

Closes #1034

### DIFF
--- a/cmd/kuketty/repos.go
+++ b/cmd/kuketty/repos.go
@@ -116,8 +116,17 @@ func resolveRepo(ctx context.Context, repo v1beta1.ContainerRepo, logger *slog.L
 	return status
 }
 
-// cloneRepo clones repo.URL into repo.Target, optionally pinned to repo.Branch.
+// cloneRepo clones repo.URL into repo.Target. When repo.Ref is set the clone
+// fetches the default branch (plus all refs/tags) and then detaches HEAD at
+// the pinned ref — supports both a tag name and a full commit SHA. Otherwise
+// repo.Branch (if set) is forwarded to `git clone --branch`.
 func cloneRepo(ctx context.Context, repo v1beta1.ContainerRepo) error {
+	if repo.Ref != "" {
+		if err := runGit(ctx, "", "clone", repo.URL, repo.Target); err != nil {
+			return err
+		}
+		return runGit(ctx, repo.Target, "checkout", "--detach", repo.Ref)
+	}
 	args := []string{"clone"}
 	if repo.Branch != "" {
 		args = append(args, "--branch", repo.Branch)
@@ -126,10 +135,19 @@ func cloneRepo(ctx context.Context, repo v1beta1.ContainerRepo) error {
 	return runGit(ctx, "", args...)
 }
 
-// fetchRepo updates an existing checkout: fetch the remote, then (when a branch
-// is requested) check it out and fast-forward. A non-fast-forwardable local
-// state surfaces as an error rather than a silent divergence.
+// fetchRepo updates an existing checkout. For a Ref-pinned repo (immutable tag
+// or commit SHA) the path is fetch-tags + checkout --detach, with no
+// fast-forward — `git pull --ff-only` on a detached HEAD has no upstream and
+// would error, so a tag pin only survived a fresh clone before this branch
+// (#1034). For a branch-tracking repo the prior fetch + checkout + ff-only
+// path is preserved.
 func fetchRepo(ctx context.Context, repo v1beta1.ContainerRepo) error {
+	if repo.Ref != "" {
+		if err := runGit(ctx, repo.Target, "fetch", "--prune", "--tags", "origin"); err != nil {
+			return err
+		}
+		return runGit(ctx, repo.Target, "checkout", "--detach", repo.Ref)
+	}
 	if err := runGit(ctx, repo.Target, "fetch", "--prune", "origin"); err != nil {
 		return err
 	}
--- a/cmd/kuketty/repos_test.go
+++ b/cmd/kuketty/repos_test.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"testing"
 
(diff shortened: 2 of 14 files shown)


eminwux:
**PR #1082 Review — `feat(repos): add ContainerRepo.ref for immutable tag/commit pins`**

LGTM — additive `Ref` field is plumbed cleanly through v1beta1 → apischeme round-trip → materialize → kuketty clone/fetch; mutex gate is enforced in `validateBlueprintRepos`, `validateRepos`, and `validateConfig` (CellConfigRepoFill); restart-idempotence test exercises the `#1034` AC directly.

PR #1085, opened 2026-06-05
feat(status): add per-realm storage footprint section

### LINKED ISSUE #1039: kuke status: surface per-realm storage footprint (snapshots / leases / content)
## Summary

Surface per-realm storage footprint in `kuke status` so snapshot/lease/content accumulation is observable **before** the data volume fills. Today the leak is invisible until something hits ENOSPC.

## Motivation

A storage leak filled `/dev/sda1` with no early signal; the first symptom surfaced several layers away (an attach timeout — sbsh#103673). `kuke status` already walks every resource kind per realm; adding a storage section makes accumulation visible early.

## Proposed behavior

Add a per-realm storage section to `kuke status` (and its JSON output) reporting:
- **snapshots** — count (and disk usage where cheap to compute) in the realm's containerd namespace
- **leases** — count
- **content** — blob count / size

Do **not** include a stopped-cell count (out of scope — finished-cell reclamation is the operator's responsibility via `--rm`).

## Acceptance criteria

- `kuke status` shows snapshot / lease / content figures per realm.
- The same figures appear in the JSON output shape.
- Reasonable performance: don't make `kuke status` significantly slower (sample / cheap-count where full `du` would be expensive).

## Related

- #1036, #1037, #1038 (the leaks this makes visible)

eminwux:
## Summary

- Adds a `storage` section to `kuke status` reporting per-realm snapshot / lease / content-blob counts plus summed blob bytes, so accumulation is observable before the data volume hits ENOSPC — the prior failure mode (#1039 motivation) had the leak surfacing only as an attach timeout several layers downstream.
- New `ctr.Client.NamespaceStorage(namespace)` primitive returns the figures via containerd metadata-store iterators (boltdb-backed leases / content / per-snapshotter walks) — no on-disk `du`, so the call stays inside the status budget. Aggregates across `KukeonKnownSnapshotters` so the report doesn't depend on which snapshotter the host populates.
- JSON shape extended with an optional `storage` payload on storage rows (`snapshots`, `leases`, `blobs`, `blobsBytes`) so CI alerting parses figures directly instead of regexing the human Detail string.

## Design calls

- **Per-snapshot disk usage is intentionally omitted.** The AC budgets "(and disk usage where cheap to compute)" — counts and content-blob byte totals are cheap (metadata-store reads), but per-snapshot disk usage requires walking the snapshotter's filesystem, which would push `kuke status` past its budget. Documented in `ctr.StorageStats`'s godoc and in the `kuke status` reference.
- **Aggregated snapshot count, not per-snapshotter breakdown.** Real installs only populate one snapshotter (overlayfs); a per-snapshotter row would noise the report. Aggregation also keeps the section row count = realm count, matching how `state` reports namespaces.
- **Daemon-down fallback uses the ctr namespace list.** When `ListRealms` is unreachable, the section still surfaces figures by deriving realm names from kukeon containerd namespaces (same shape as `checkStateNamespaces`). The daemon row above carries the underlying cause.
- **Section ordering: state → storage → parity.** Storage is a host-state observation (alongside residual namespaces), so it sits next to `state`; `parity` stays last because it's the most expensive.
- **No automatic WARN thresholds.** Surfacing figures is enough; the operator decides when accumulation is too high. A future PR can add policy if needed.
- **Project `CLAUDE.md` not edited.** The narrative there is still accurate (it enumerates what `kuke status` is for, not the exact section names); per dev/CLAUDE.md no `<project>/CLAUDE.md` edits in a code-fix PR. The user-facing reference (`docs/site/cli/kuke-status.md`) is updated in this PR.

## Test plan

- [x] `make test` — full root + kukebuild test suites pass
- [x] `go vet ./...` clean
- [x] `pre-commit run --files <changed>` clean (govet shadow, prettier, end-of-file-fixer all pass)
- [x] New unit tests in `cmd/kuke/status/storage_test.go` cover: per-realm row + structured payload, daemon-down namespace fallback, ctr-unreachable single-WARN row, per-realm probe-error WARN, JSON shape carries the `storage` payload, `fmtBytes` unit-picker spans B/KiB/MiB/GiB/TiB boundaries
- [ ] `make dev-init` smoke not run — the diff is purely additive (new public method on `ctr.Client`, new read-only CLI section); it does not touch the build path, the daemon serve path, or anything under `/opt/kukeon`, so the smoke gate in the project `CLAUDE.md` does not apply

Closes #1039

### DIFF
--- a/cmd/kuke/status/checks.go
+++ b/cmd/kuke/status/checks.go
@@ -31,10 +31,11 @@ import (
 // Section names — kept as constants so the renderer's section grouping and
 // the JSON `section` field are spelled the same way everywhere.
 const (
-	sectionDaemon = "daemon"
-	sectionHost   = "host"
-	sectionState  = "state"
-	sectionParity = "parity"
+	sectionDaemon  = "daemon"
+	sectionHost    = "host"
+	sectionState   = "state"
+	sectionStorage = "storage"
+	sectionParity  = "parity"
 )
 
 // checkDaemon reports the daemon's reachability. The single row carries the
--- a/cmd/kuke/status/status.go
+++ b/cmd/kuke/status/status.go
@@ -83,6 +83,23 @@ type Result struct {
 	Status      Status `json:"status"`
 	Detail      string `json:"detail,omitempty"`
 	Remediation string `json:"remediation,omitempty"`
+	// Storage carries the structured per-realm counts the storage
+	// section surfaces — populated only on storage rows. CI tooling
+	// parses these to alert on accumulation before the data volume
+	// fills (issue #1039); the human Detail string carries the same
+	// figures pre-formatted.
+	Storage *StorageStats `json:"storage,omitempty"`
+}
+
+// StorageStats mirrors ctr.StorageStats on the wire — the per-realm
+// snapshot/lease/blob figures `kuke status` exposes for parsing. The
+// status package owns the JSON tags so a future internal/ctr struct
+// rename doesn't ripple into the public CLI contract.
+type StorageStats struct {
+	Snapshots  int   `json:"snapshots"`
+	Leases     int   `json:"leases"`
+	Blobs      int   `json:"blobs"`
+	BlobsBytes int64 `json:"blobsBytes"`
 }
 
 // Report is the full top-level payload — what --json marshals, and what the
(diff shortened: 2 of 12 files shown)


eminwux:
**PR #1085 Review — `feat(status): add per-realm storage footprint section`**

LGTM — additive `kuke status` section reading containerd metadata-store iterators (boltdb), wired through a new `ctr.Client.NamespaceStorage` method with tight WARN demotion on probe failures; unit tests cover the per-realm happy path, daemon-down fallback, ctr-unreachable single-WARN, per-realm probe-error, JSON `storage` payload, and `fmtBytes` unit boundaries.

Issue #1020, opened 2026-06-02
[umbrella] CellConfig → 1:N binding; the CellDoc is the cell identity


eminwux:
## Summary

Split **CellConfig** into the two concerns it currently conflates — **binding** (reusable recipe,
1:N) and **identity** (the one live cell, 1:1) — by moving identity onto the **CellDoc**. A
CellConfig becomes a *binding* that can stamp out many cells; each cell owns its own
name/spec/status. This is the **control-plane** half of the stateful-agent redesign; the
data-plane half (durable named volumes) is umbrella #1015.

**Executes after epic:bye-container (#993) lands** and reuses epic:reconcile (#990–992, landed).

## Motivation

CellConfig today fuses two concerns:

- **(A) Binding** — Blueprint + concrete Values + closed repo/secret slots. Intrinsically
  *reusable*: one binding should stamp many cells.
- **(B) Identity** — `CellConfigMetadata.Name` is *also* the deterministic name of the
  at-most-one live cell (`internal/cellconfig.StableName`), plus the `kukeon.io/config` back-ref
  label and the OutOfSync reconcile. This pins it to exactly one cell.

The `--clone` / `--reuse` / `--new` knobs (and the `kukeon.io/source-config` annotation) are
concern **A escaping a kind that enforces B** — workarounds for a binding forced to be an identity.

**Decision:** CellConfig keeps concern A only (binding, 1:N). The **CellDoc becomes the cell's
identity** (concern B). This doesn't collapse Config into Blueprint: Blueprint leaves slots *open*
(`${KEY}` params, repo/secret slots), Config *closes* them — both 1:N, differing in fill state.

### Why durable stateful agents need this

Cells are agents that start and finish but whose data we keep — they accumulate memory/learning
that belongs to the harness, and even cron-driven runs reuse them for that accrued expertise. We
need to **update a cell's spec (e.g. add an env var) without deleting + recreating the cell and
losing that state.** The Cell ↔ Config link is the control-plane update mechanism; the volume
(#1015) is the data-plane durability that survives the container recreation an update entails.
Neither alone satisfies "update a stateful agent without losing its memory"; together they do.

## Key decisions

1. **CellConfig = binding (1:N); CellDoc = identity.** `kukeon.io/config` (and a new
   `kukeon.io/blueprint`) become *lineage/provenance* labels, not identity pins.
2. **`create cell` is the single materialization verb** (building on bye-container #996, which
[...]

PR #1086, opened 2026-06-05
feat(cell): persist materialization provenance; demote config label to lineage

### LINKED ISSUE #1021: model: CellDoc carries materialization inputs; config label becomes lineage (P1)
## Context

First step of epic:cell-identity (umbrella #1020). Moves cell *identity* from the CellConfig onto
the CellDoc, and gives every cell the recorded inputs needed to re-resolve from its binding later.

Today `CellConfigMetadata.Name` doubles as the deterministic name of the at-most-one live cell
(`internal/cellconfig.StableName`), and `kukeon.io/config` is an identity back-ref. This step
demotes both to *lineage/provenance* and records the materialization inputs on the cell.

Runs after epic:bye-container (#996 already makes a source flag required on `create cell`).

## Proposal

- **Persist materialization inputs on the CellDoc.** Add a provenance block recording
  `{bindingKind (blueprint|config), bindingRef (name+scope), params, envOverrides}`. This is what
  P4 re-runs to compute `desired` for the diff, and what P2 reads for the name prefix.
  - These fields are *persisted* (unlike `Spec.RuntimeEnv`, which is `yaml:"-"` per #834). Decide
    placement: a dedicated `Spec.Provenance` sub-struct vs. metadata annotations. Lean: a typed
    `Spec.Provenance` so the diff layer can ignore it cleanly (it is not a runtime spec field).
- **Demote `kukeon.io/config` to lineage.** It records *which binding stamped this cell*, 1:N — no
  longer an identity. Add a parallel `kukeon.io/blueprint` lineage label for blueprint-stamped cells.
- **Stop treating the config name as the cell name.** The cell name is its own (P2 supplies the
  generator); this step just severs the assumption in the materialization path.

## Acceptance criteria

- [ ] CellDoc persists `{bindingKind, bindingRef, params, envOverrides}` across write/read round-trip.
- [ ] `kukeon.io/config` is documented + treated as lineage (1:N), not identity; `kukeon.io/blueprint` added for blueprint-stamped cells.
- [ ] The provenance block is excluded from `DiffCell` (it is not a runtime spec field — verify no false OutOfSync).
- [ ] Materialization no longer derives the cell name from the config name (name is supplied by the caller / P2 generator).

## Dependencies

Umbrella: #1020. Depends on: epic:bye-container #996. Blocks: P2, P3, P4 (this epic).

eminwux:
## Summary

First step (P1) of **epic:cell-identity** (umbrella #1020). Moves cell *identity* off the CellConfig and onto the CellDoc, and records the materialization inputs every cell needs to re-resolve from its binding later.

- **New persisted `Spec.Provenance` block** (`pkg/api/model/v1beta1/cell.go`, mirrored in `internal/modelhub/cell.go`): a typed `CellProvenance{BindingKind, BindingRef{Name,Realm,Space,Stack}, Params, EnvOverrides}`. Unlike the transport-only `RuntimeEnv` (`yaml:"-"`, dropped on the outbound builder), provenance is **persisted** — it round-trips through both the JSON `metadata.json` write and a `kuke get cell -o yaml` render, and is copied in **both** conversion directions in `internal/apischeme/scheme.go`. A hand-built cell carries `nil` provenance.
- **Both materialization paths stamp it.** `internal/cellconfig/materialize.go` → `bindingKind=config`, ref from the Config metadata, `params` from `spec.values`. `internal/cellblueprint/cellblueprint.go` → `bindingKind=blueprint`, ref from the Blueprint metadata, `params` from the resolved `--param` map (threaded into `MaterializeWithName`). `EnvOverrides` is populated from `kuke run --env` at `cmd/kuke/run/run.go` alongside `RuntimeEnv` (guarded on `Provenance != nil`, so a manifest run with no binding never fabricates one).
- **`kukeon.io/config` demoted to lineage (1:N), not identity** — docstrings on `LabelConfig` and `mergeConfigLabel` updated. `kukeon.io/blueprint` was already stamped on blueprint-materialized cells (`cellblueprint.LabelBlueprint`); confirmed it covers the AC.
- **Provenance excluded from `DiffCell`** — it is lineage data, not a runtime spec field. `DiffCell` only compares named spec fields, so a new field is ignored by construction; `TestDiffCell_ProvenanceIgnored` pins that a provenance-only (and nil-vs-set) difference reports no change, so the reconciler never stamps a false `OutOfSync`.
- **Materialization no longer derives the cell name from the Config name.** `cellconfig.MaterializeWithName` now uses the caller-supplied name verbatim (the `StableName` fallback and the assumption-laden 2-arg `Materialize` wrapper are removed). Callers that want the historical identity pass `StableName(cfg.Metadata.Name)` at the call site; the OutOfSync reconciler threads the **live cell's own name** through (`materializeCellFromConfig`), which is the correct, P2-forward behavior — the detector compares *spec drift*, not identity.

### Scoping / non-obvious calls (reviewer please push back)

- **AC4 — hard sever (literal reading).** I removed the name derivation from the materialization *layer* and moved `StableName` to each caller (run, create-cell) as a temporary stand-in until P2 supplies the name generator. The reconciler now passes the live cell's name, decoupling the diff from `StableName(configName)` so it stays correct once P2 lands generated names. This is the literal AC; the call sites that previously relied on the fallback are updated and tested.
- **`EnvOverrides` population.** Defined + round-tripping (AC1's requirement) and populated for the `kuke run --env` path. `create cell` has no `--env`, so its cells carry no env overrides — expected.
- **Blueprint provenance.** Stamped even though the only current OutOfSync consumer is config-lineage (blueprint cells are out of scope for the reconciler per the umbrella). Done so the `{bindingKind: blueprint}` half of the model is real and round-trips now rather than being retrofitted.

## Test plan

- [x] `make test` (full CI target, `GOWORK=off`) — exit 0, no FAIL/panic.
- [x] `GOWORK=off go build ./...` + `go vet ./...` — clean.
- [x] New unit tests: provenance JSON/YAML/conversion round-trip + nil-omitted (`internal/apischeme/provenance_test.go`); `DiffCell` ignores provenance (`internal/controller/apply/diff_test.go`); config + blueprint provenance stamping and name-not-derived-from-config (`internal/cellconfig`, `internal/cellblueprint`).
- [x] `make dev-init` smoke — exit 0; daemon-parity tail shows both realms `Ready` in daemon and `--no-daemon` views with identical shape; kukeond anchored on the just-built image; `kuke attach` smoke exited cleanly.

Closes #1021

### DIFF
--- a/cmd/kuke/create/cell/cell.go
+++ b/cmd/kuke/create/cell/cell.go
@@ -253,7 +253,7 @@ func runFromBlueprint(cmd *cobra.Command, client kukeonv1.Client, flags createCe
 	if err != nil {
 		return err
 	}
-	cellDoc, err := cellblueprint.MaterializeWithName(resolved, flags.name)
+	cellDoc, err := cellblueprint.MaterializeWithName(resolved, flags.name, cliParams)
 	if err != nil {
 		return err
 	}
@@ -312,7 +312,15 @@ func runFromConfig(cmd *cobra.Command, client kukeonv1.Client, flags createCellF
 		)
 	}
 
-	cellDoc, err := cellconfig.MaterializeWithName(cfgRes.Config, bpRes.Blueprint, flags.name)
+	// The cell name is supplied here, not derived inside Materialize
+	// (epic:cell-identity #1021). Fall back to StableName when --name is
+	// absent so the create-cell config path keeps its historical stable
+	// identity; P2 replaces this with a dedicated name generator.
+	cellName := flags.name
+	if strings.TrimSpace(cellName) == "" {
+		cellName = cellconfig.StableName(cfgRes.Config.Metadata.Name)
+	}
+	cellDoc, err := cellconfig.MaterializeWithName(cfgRes.Config, bpRes.Blueprint, cellName)
 	if err != nil {
 		return err
 	}
--- a/cmd/kuke/run/divergence_idempotency_test.go
+++ b/cmd/kuke/run/divergence_idempotency_test.go
@@ -70,7 +70,7 @@ func TestDivergedFields_MaterializePersistMaterializeIsIdempotent(t *testing.T) 
 		t.Run(tc.name, func(t *testing.T) {
 			// Step 1: Materialize what `kuke restart` would push through
 			// ApplyDocuments.
-			cell1, err := cellconfig.Materialize(tc.cfg, tc.bp)
+			cell1, err := cellconfig.MaterializeWithName(tc.cfg, tc.bp, cellconfig.StableName(tc.cfg.Metadata.Name))
 			if err != nil {
 				t.Fatalf("Materialize (restart side): %v", err)
 			}
@@ -123,7 +123,7 @@ func TestDivergedFields_MaterializePersistMaterializeIsIdempotent(t *testing.T) 
 			}
 
 			// Step 4: Re-Materialize as the next `kuke run <config>` would.
-			cell2, err := cellconfig.Materialize(tc.cfg, tc.bp)
+			cell2, err := cellconfig.MaterializeWithName(tc.cfg, tc.bp, cellconfig.StableName(tc.cfg.Metadata.Name))
 			if err != nil {
 				t.Fatalf("Materialize (run side): %v", err)
 			}
--- a/cmd/kuke/run/run.go
+++ b/cmd/kuke/run/run.go
@@ -558,6 +558,16 @@ func runRun(cmd *cobra.Command, args []string) error {
 	// last-wins across collisions) are owned by the runner-side mergeRuntimeEnv.
 	if len(flags.envArgs) > 0 {
 		cellDoc.Spec.RuntimeEnv = flags.envArgs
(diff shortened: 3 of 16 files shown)


eminwux:
**PR #1086 Review — `feat(cell): persist materialization provenance; demote config label to lineage`**

Clean P1 of epic:cell-identity. Provenance round-trips through both conversions (JSON + YAML, both directions), `DiffCell` correctly ignores it (pinned by `TestDiffCell_ProvenanceIgnored`), the reconciler now threads the live cell's own name so the diff stays correct after P2 lands generated names, the `--env` populate path correctly guards on `Provenance != nil` for manifest runs, and the `StableName` fallback moves to each call site exactly where the PR description says. All five commits verified. `make dev-init` smoke present.

One non-blocking observation:

- **Unused helper** `internal/modelhub/cell.go:92`: `CloneCellProvenance` is defined but has no callers in this PR (the `v1beta1` mirror is used by `NewCellDoc`). If this is intentional symmetry awaiting P2/P3 consumption, no action; otherwise it could land alongside its first caller.

LGTM.

PR #1087, opened 2026-06-05
feat(kukeond): warn on data-volume disk pressure + guard new cell creation

### LINKED ISSUE #1035: kukeond: warn on data-volume disk pressure + guard new cell creation (no auto-deletion)
## Summary

kukeond can fill its data volume to 100% with no warning and no guard, which then causes downstream failures (e.g. it triggered the kuketty attach wedge sbsh#373 when `/run/kukeon/tty` — bind-mounted from the full ext4 volume — hit ENOSPC). kukeond should make disk pressure **visible** and **refuse to dig the hole deeper**, without ever deleting operator data on its own.

Explicitly **not** in scope: auto-reaping `Stopped` cells. Reclaiming finished cells is the operator's responsibility via `kuke run --rm` / `kuke delete`. This issue adds observability + a creation guard only.

## Motivation

A real incident: ~107 finished one-shot cells (operator didn't use `--rm`) accumulated ~96 GB of Active overlay snapshots and filled `/dev/sda1`. The first symptom was not "disk full" but a confusing attach timeout several layers away. kukeond had every signal needed to warn long before that.

## Proposed behavior

1. **Warn on disk pressure.** In the reconcile loop, sample the data-volume usage each tick. When it crosses a high-water mark (default 85%, `KUKEOND_DISK_PRESSURE_WARN_PCT`), emit a rate-limited WARN with the realm and current usage. No deletion.
2. **Guard new cell creation under pressure.** When usage is above a hard threshold (default 95%, `KUKEOND_DISK_PRESSURE_BLOCK_PCT`), `CreateCell` fails fast with a clear error naming the volume and usage, instead of creating a snapshot that pushes the volume to 100%.
3. **Opt-out flag.** `kuke create cell` / `kuke run` accept an override (e.g. `--ignore-disk-pressure`) that bypasses the guard for operators who know what they're doing.

## Acceptance criteria

- Crossing the warn threshold logs a rate-limited WARN naming the realm + usage; nothing is deleted.
- Above the block threshold, `kuke create cell` / `kuke run` fail with a disk-pressure error; the override flag lets creation proceed.
- No cell, snapshot, or metadata is ever removed by this path.

## Related

- sbsh#373 (the ENOSPC-triggered attach wedge this guard prevents)

eminwux:
## Summary

kukeond could fill its data volume to 100% with no warning and no guard, then fail downstream in confusing ways (the ENOSPC that triggered the sbsh#373 attach wedge). This makes disk pressure **visible** and makes the daemon **refuse to dig the hole deeper**, without ever deleting operator data.

- **Warn on disk pressure (reconcile loop).** Each reconcile pass samples every realm's data volume (`statfs` on `<RunPath>/data/<realm>`) and emits a rate-limited WARN naming the realm + usage when usage is at/above `KUKEOND_DISK_PRESSURE_WARN_PCT` (default 85). Nothing is deleted. Rate-limited per-realm at a 5-minute window so a 30s reconcile interval doesn't log every tick.
- **Guard new cell creation.** `runner.CreateCell` fails fast with a new `errdefs.ErrDiskPressure` (naming the volume + usage) when usage is at/above `KUKEOND_DISK_PRESSURE_BLOCK_PCT` (default 95), *before* provisioning the snapshot. The guard sits on the genuinely-new-cell branch only — `EnsureCell` (existing cells) is not guarded, so ensuring resources for a cell that already exists is never blocked. Covers both `kuke run` and `kuke create cell` (both reach `runner.CreateCell` via `acquireOrCreateCell`).
- **Opt-out flag.** `kuke create cell` / `kuke run --ignore-disk-pressure` bypasses the guard. Threaded as a **transport-only** `CellSpec.IgnoreDiskPressure` field (`json` + `yaml:"-"`), mirroring the existing `RuntimeEnv` (#834) pattern: carried CLI → daemon over JSON-RPC, copied in `ConvertCellDocToInternal`, and **dropped** in `BuildCellExternalFromInternal` so the per-invocation override never persists into the stored cell spec.
- **No deletion anywhere.** The reconcile WARN only logs; the guard only refuses. Auto-reaping `Stopped` cells is explicitly out of scope (operator's job via `kuke run --rm` / `kuke delete`).

### Notable design calls (reviewer please confirm)

- **Transport-only override, not a persisted spec field.** `IgnoreDiskPressure` is a per-invocation creation directive, so it follows `RuntimeEnv`'s `yaml:"-"` no-persist contract rather than the `AutoDelete` persisted-spec pattern — persisting it would silently exempt a cell's future rematerializations from the guard. Documented inline at both conversion sites.
- **Per-realm sampling, deduped by rate-limit key.** The WARN samples each realm's data dir so it "names the realm" per the AC and tolerates a future where realms live on different mounts. Realms sharing one filesystem each warn under their own rate-limit key (≤ one WARN per realm per window).
- **Config surface is env-var + flag only** (not ServerConfiguration), matching the issue's named knobs `KUKEOND_DISK_PRESSURE_WARN_PCT` / `KUKEOND_DISK_PRESSURE_BLOCK_PCT`; also exposed as `kukeond serve --disk-pressure-warn-percent` / `--disk-pressure-block-percent` for parity with the existing `--default-memory-limit-bytes` knob.
- **Fail-open on sample error.** A `statfs` failure in the guard logs a WARN and allows creation — a monitoring hiccup must never wedge all cell creation.
- No new disk-usage helper duplicates #1085's `internal/ctr/storage.go`: that measures per-realm containerd *content* size; this `diskpressure` package measures filesystem free space via `statfs`. Distinct concepts.

## Test plan

- [x] `make test` (full CI test target, `GOWORK=off`) — passes.
- [x] `go vet ./...` (`GOWORK=off`) — clean.
- [x] `pre-commit run --files` over the changed set (gofmt, golangci-lint, addlicense, go-mod-tidy) — passes.
- [x] New unit tests: `internal/util/diskpressure` (Sample invariants + Warner rate-limiting), runner `guardDiskPressure` (block / allow / override / disabled / sample-error-fail-open), controller `checkDiskPressure` (warn-over-threshold naming the realm / silent-below / disabled / rate-limited-across-passes).
- [x] `make dev-init` end-to-end smoke (daemon-touching change) — exit 0; daemon-parity tail identical across daemon and `--no-daemon` views (both realms `Ready`), running kukeond anchored on the just-built image, and the `kuke attach` smoke passed. Host disk at 45% so the default 95% guard / 85% warn did not interfere with `kuke init`.

Closes #1035

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -154,6 +154,23 @@ var (
 	KUKEOND_KUKETTY_LOG_LEVEL = DefineKV(
 		"KUKEOND_KUKETTY_LOG_LEVEL", "kukeond/kukettyLogLevel", "info",
 	)
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	// KUKEOND_DISK_PRESSURE_WARN_PCT is the data-volume usage percentage above
+	// which kukeond's reconcile loop emits a rate-limited WARN naming the realm
+	// and current usage. Nothing is deleted. 0 (or negative) disables the
+	// warning. Issue #1035.
+	KUKEOND_DISK_PRESSURE_WARN_PCT = DefineKV(
+		"KUKEOND_DISK_PRESSURE_WARN_PCT", "kukeond/diskPressureWarnPct", "85",
+	)
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	// KUKEOND_DISK_PRESSURE_BLOCK_PCT is the data-volume usage percentage at or
+	// above which kukeond refuses to provision a new cell (CreateCell fails
+	// fast) so a snapshot does not push the volume to 100%. `kuke create cell`
+	// / `kuke run --ignore-disk-pressure` bypasses the guard. 0 (or negative)
+	// disables the guard. Issue #1035.
+	KUKEOND_DISK_PRESSURE_BLOCK_PCT = DefineKV(
+		"KUKEOND_DISK_PRESSURE_BLOCK_PCT", "kukeond/diskPressureBlockPct", "95",
+	)
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_INIT_REALM = DefineKV("KUKE_INIT_REALM", "kuke/init/realm")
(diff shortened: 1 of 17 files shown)


eminwux:
**PR #1087 Review — `feat(kukeond): warn on data-volume disk pressure + guard new cell creation`**

LGTM — mirrors the RuntimeEnv (#834) transport-only pattern exactly (`yaml:"-"` + drop in `BuildCellExternalFromInternal`), the guard is correctly placed on the genuinely-new-cell branch only (not `EnsureCell`), `ErrDiskPressure` is wrapped with `%w` so `errors.Is` works, fail-open on `statfs` error is explicit on both reconcile (DEBUG + skip) and guard (WARN + allow) paths, rate-limiting key is per-realm, and the unit tests cover every claimed branch (block / allow / override / disabled / sample-error-fail-open for the guard; warn-over / silent-below / disabled / rate-limited-across-passes for reconcile; Sample invariants + Warner per-key window for the package). Daemon-parity smoke is in the test plan. Signature verified, no public-API parity gap (no `pkg/builder` to thread).

PR #1090, opened 2026-06-05
feat(dev-init): configurable podSubnetCIDR so nested dev-init avoids the parent's 10.88.0.0/16

### LINKED ISSUE #1079: dev-init: nested make dev-init reuses parent's 10.88.0.0/24 + .1 gateway, blackholing the dev-root cell's own egress
## Summary

Running the canonical smoke test `make dev-init` **inside** a `kukeon-dev-root` cell breaks that cell's own outbound internet. The nested kukeon's CNI allocator reuses the parent host's `10.88.0.0/24` subnet and `10.88.0.1` gateway, which is also the dev-root cell's *own* default gateway. Once the nested bridge claims `10.88.0.1`, that address becomes a **local** address inside the cell and shadows the cell's default gateway — all outbound packets are delivered locally and never reach the parent bridge.

Net effect: the cell has internet *before* the smoke test and 100% egress packet loss *after* the first nested `kuke init`. This also breaks the nested `kuke build` image pulls and anything else needing egress, for any agent following the canonical "rebuild and re-run `kuke init`" loop in a dev-root cell.

## Root cause

`scripts/dev-init.sh` nested mode (issue #547) redirects three things so the nested daemon doesn't clobber the parent:

- socket → `/run/kukeon-dev/kukeond.sock`
- containerd namespace suffix → `dev.kukeon.io`
- cgroup root → `/kukeon-dev`

It does **not** redirect the CNI pod subnet. The subnet allocator (`internal/controller/runner/runner.go:251`) hands out `/24` chunks from `10.88.0.0/16`; the first space gets `10.88.0.0/24` with gateway `10.88.0.1`. The nested kukeon runs identical code from the same base, so it allocates the identical `10.88.0.0/24` + `.1`, and `cni.SafeBridgeName` deterministically reproduces the same `k-<hex>` bridge name as the parent.

The cell's uplink `eth0` sits in the parent's first-space subnet (`10.88.0.x/24`, gw `10.88.0.1`). Collision is total.

There's an irony at `internal/controller/runner/runner.go:257`, which explicitly cites *"regress #131's collision-on-10.88.0.1 bug"* — the allocator is hardened against the within-host collision but not the parent↔nested one, because dev-init nested mode never redirects the pod CIDR.

## Reproduction

1. From inside a `kukeon-dev-root` cell, confirm egress works: `ping -c2 1.1.1.1` → 0% loss.
2. Run `make dev-init` (the canonical nested smoke test).
3. `ping -c2 1.1.1.1` from the cell → 100% loss.

## Diagnosis fingerprint

Run from inside the cell's netns (`sudo nsenter -t <root-pid> -n ...`):

- `ping 10.88.0.1` (the gateway) → 100% loss, even though `ip route get 1.1.1.1` resolves via `eth0` and the gateway ARP entry shows the parent bridge MAC.
- `ip route show table local | grep '10.88.0.1 '` → multiple `local 10.88.0.1 ... scope host` entries on nested `k-*` bridges.

On the **parent host** (rules are all healthy — this is *not* a firewall fault):

- `conntrack -L | grep <cell-ip>` → **no entry** for the flow; packets never left the cell.
- `iptables -S FORWARD` → `KUKEON-FORWARD` accepts `-i k-+` egress (intact).
- `iptables -t nat -S POSTROUTING` → per-cell `CNI-*` MASQUERADE intact (stale duplicate masquerade rules for an old cell IP are harmless — both jump the same CNI chain).

Observed state in one affected cell: 23 nested `k-*` bridges all numbered `10.88.0.1/24` (22 empty leftovers from prior dev-init cycles + 1 active with member veths from the live nested run).

## Proposed fix

Give nested dev-init mode a distinct, non-overlapping subnet base (e.g. `10.89.0.0/16`) alongside the socket / namespace / cgroup redirections it already performs — the nested allocator should never land on the parent's `10.88.0.0/16`. Extend the `#131` collision guard to cover the parent↔nested case, or document a required pod-CIDR override for nested mode.

## Notes

- The stale empty `k-*` bridges accumulate one per dev-init cycle and are safe to remove, but cleaning them does **not** fix egress — the *active* nested bridge still owns `10.88.0.1`.
- Workaround until fixed: renumber the nested kukeon's pod CIDR before relying on egress from inside the dev-root cell.

eminwux:
## Summary

- Nested `make dev-init` (running inside a `kukeon-dev-root` cell) reused the parent host's `10.88.0.0/16` + `10.88.0.1` gateway for the nested kukeon's CNI allocator. Since `10.88.0.1` is the dev-root cell's own default gateway, the nested bridge claiming `.1` makes it a *local* address inside the cell, shadowing the gateway and blackholing all egress (issue #1079).
- Adds `podSubnetCIDR` as a **4th nested-redirect knob**, peer to the existing socket / `containerdNamespaceSuffix` / `cgroupRoot` redirects, plumbed end-to-end exactly like `cgroupRoot`:
  - `cni.ConfigureSubnetParentCIDR(cidr)` overrides the process-global parent block `NewDefaultSubnetAllocator` subdivides (validates the CIDR; empty = no-op keeping the `10.88.0.0/16` default).
  - `ServerConfiguration.spec.podSubnetCIDR` + `ClientConfiguration.spec.podSubnetCIDR`, `KUKEON_POD_SUBNET_CIDR` env var, and `kukeond --pod-subnet-cidr` flag. Resolved into `cni.ConfigureSubnetParentCIDR` at process start in all three `ConfigureRuntime` call sites (daemon `PersistentPreRunE`, `kuke` client config load, shared admin `LoadServerConfigurationFromFlag`).
  - WriteDefault templates (server + client) and `currentResolvedSpec` carry the new field.
- `scripts/dev-init.sh` dev profile (`KUKEON_PROFILE=dev`) now writes `podSubnetCIDR: 10.89.0.0/16` into both `kukeond-dev.yaml` and `kuke-dev.yaml`, alongside the existing `dev.kukeon.io` / `/kukeon-dev` redirects — so the nested allocator carves `10.89.0.0/16`, clear of the parent's `/16`.
- `docs/dev-init.md` gains a "Nested egress" section documenting the new redirect and the collision it fixes.

## Design notes (for reviewer)

- **Gate-vs-flip:** the historical default (`10.88.0.0/16`) is preserved for bare-host `make dev-init` and all production installs — the redirect is **opt-in** via the dev profile, matching how `cgroupRoot`/`namespaceSuffix` already behave (the lower-blast-radius reading per the AC's "alongside the … redirections it already performs"). A global flip of the default would break existing contributors, so it was avoided.
- **Where the knob lives:** the AC offered "extend the #131 collision guard *or* document a required pod-CIDR override." The override approach was chosen because it's the only one that actually restores egress (the #131 guard is within a single allocator; the parent↔nested collision is across two independent kukeon instances, which a within-instance guard cannot see). It also mirrors the established multi-instance config model.
- **Implementation home:** the override is a `cni`-package process-global (purely additive — no `consts.ConfigureRuntime` signature change, so no churn to its callers/tests), set once at startup before any runner is built, then read-only.
- **Nested-without-dev-profile** is out of scope: such a run already collides on namespace and cgroup, so it is not a supported configuration (noted in `docs/dev-init.md`).

## Test plan

- [x] `make test` (full CI target: `test-root` + `test-kukebuild`) — pass
- [x] `GOWORK=off go vet ./...` — pass
- [x] `pre-commit run --files <changed>` (trailing-whitespace, eof, go-fmt, go-mod-tidy, **golangci-lint**, prettier, addlicense) — pass
- [x] New unit tests: `cni.ConfigureSubnetParentCIDR` override / empty-no-op / invalid-rejected + first-allocation lands on `10.89.0.0/24`; server+client config round-trip and apply-path assertions for `podSubnetCIDR`; `LoadServerConfigurationFromFlag` / `loadClientConfiguration` drive `ConfigureSubnetParentCIDR`
- [ ] End-to-end nested `make dev-init` egress smoke (`ping 1.1.1.1` from inside a dev-root cell before/after) — not run: requires a `kukeon-dev-root` cell with a live parent daemon and standalone containerd, which the dev host can't provision without disturbing unrelated state. Covered by reasoning + the unit test asserting the first nested allocation is `10.89.0.0/24` (clear of the parent's `10.88.0.1`).

Closes #1079

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -121,6 +121,8 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKEON_ROOT_CGROUP_ROOT = DefineKV("KUKEON_CGROUP_ROOT", "kukeon/cgroupRoot", "/kukeon")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	KUKEON_ROOT_POD_SUBNET_CIDR = DefineKV("KUKEON_POD_SUBNET_CIDR", "kukeon/podSubnetCIDR", "10.88.0.0/16")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKEON_ROOT_HOST = DefineKV("KUKEON_HOST", "kukeon/host", "unix:///run/kukeon/kukeond.sock")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKEON_ROOT_NO_DAEMON = DefineKV("KUKEON_NO_DAEMON", "kukeon/noDaemon", "false")
--- a/cmd/kuke/clientconfig_test.go
+++ b/cmd/kuke/clientconfig_test.go
@@ -22,15 +22,17 @@ import (
 	"testing"
 
 	"github.com/eminwux/kukeon/cmd/config"
+	"github.com/eminwux/kukeon/internal/cni"
 	"github.com/eminwux/kukeon/internal/consts"
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 	"github.com/spf13/viper"
 )
 
-// restoreRuntimeGlobals snapshots the consts package runtime globals and
-// returns a cleanup that restores them. loadClientConfiguration calls
-// consts.ConfigureRuntime, which mutates these — tests that exercise it
-// must roll back so the next test starts on the in-binary defaults.
+// restoreRuntimeGlobals snapshots the consts package runtime globals and the
+// cni subnet allocator's process-global parent CIDR, and returns a cleanup
+// that restores them. loadClientConfiguration calls consts.ConfigureRuntime
+// and cni.ConfigureSubnetParentCIDR, which mutate these — tests that exercise
+// it must roll back so the next test starts on the in-binary defaults.
 func restoreRuntimeGlobals(t *testing.T) {
 	t.Helper()
 	prevSuffix := consts.RealmNamespaceSuffix
@@ -38,6 +40,7 @@ func restoreRuntimeGlobals(t *testing.T) {
 	t.Cleanup(func() {
 		consts.RealmNamespaceSuffix = prevSuffix //nolint:reassign // restore runtime-overridable global
 		consts.KukeonCgroupRoot = prevRoot       //nolint:reassign // restore runtime-overridable global
+		_ = cni.ConfigureSubnetParentCIDR(cni.DefaultSubnetParentCIDR)
 	})
 }
 
@@ -50,6 +53,7 @@ func bindKukeEnv(t *testing.T) {
 		config.KUKEON_ROOT_LOG_LEVEL,
 		config.KUKEON_ROOT_NAMESPACE_SUFFIX,
 		config.KUKEON_ROOT_CGROUP_ROOT,
+		config.KUKEON_ROOT_POD_SUBNET_CIDR,
 	} {
 		if err := v.BindEnv(); err != nil {
 			t.Fatalf("BindEnv %s: %v", v.EnvVar(), err)
@@ -85,6 +89,7 @@ func TestApplyClientConfigurationDefaultsLayered(t *testing.T) {
 		LogLevel:                  "warn",
 		ContainerdNamespaceSuffix: "dev.kukeon.io",
 		CgroupRoot:                "/kukeon-dev",
+		PodSubnetCIDR:             "10.89.0.0/16",
 	}
 	applyClientConfiguration(cmd, spec)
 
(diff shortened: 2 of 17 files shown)


eminwux:
**PR #1090 Review — `feat(dev-init): configurable podSubnetCIDR so nested dev-init avoids the parent's 10.88.0.0/16`**

LGTM — `podSubnetCIDR` is plumbed exactly parallel to the existing `cgroupRoot` / `containerdNamespaceSuffix` redirects (server+client spec, env var, `kukeond --pod-subnet-cidr` flag, dev-profile YAML, resolved-spec round-trip). Validation reuses `NewSubnetAllocator` and surfaces `errdefs.ErrInvalidSubnetCIDR`; the process-global setter is sequenced after `consts.ConfigureRuntime` in all three load paths and before the only non-test `NewDefaultSubnetAllocator` call (`internal/controller/runner/runner.go`); unit coverage asserts the first nested allocation lands on `10.89.0.0/24`, clear of the parent's `10.88.0.1`.

**Deferred verifications to run pre-merge** (documented host-gated carve-out per the test plan):
- Nested egress smoke: `ping 1.1.1.1` from inside a `kukeon-dev-root` cell before/after.
- Bare-host `make dev-init` daemon-parity tail to confirm the new unconditional `cni.ConfigureSubnetParentCIDR` in `PersistentPreRunE` doesn't regress default startup.

Issue #1022, opened 2026-06-02
naming: unified <prefix>-<6hex> generator; retire <src>-<N> + source-config (P2)


eminwux:
## Context

Second step of epic:cell-identity (umbrella #1020). Unifies cell-name generation across the
blueprint and config paths and retires the clone-era naming machinery.

Today: blueprint materialization auto-names `<prefix>-<6hex>`; config materialization pins the
deterministic `StableName` (= config name, the 1:1 identity); `--clone` forks `<src>-<N>` via a
gap-filled counter tied to the `kukeon.io/source-config` annotation. With identity on the CellDoc
(P1), all three collapse into one rule.

## Proposal

One generator for both kinds:

| input | result |
| --- | --- |
| explicit name (positional / `--name`) | use verbatim; **collision in scope = error** (create is not idempotent-attach) |
| name omitted | `<prefix>-<6hex>`, collision-retry on the suffix |

- Prefix resolution, identical for both: `Spec.Prefix` → fallback `metadata.name`.
- **Add an optional `Prefix` field to `CellConfigSpec`** (mirrors `CellBlueprintSpec.Prefix`); the
  `metadata.name` fallback covers the common case, so it is not required surface.
- **Reduce `internal/cellconfig.StableName`** to a prefix-default helper (no longer an identity
  pin); confirm no other caller depends on its deterministic 1:1 form.
- **Delete the `<src>-<N>` gap-filled counter** and the `kukeon.io/source-config` annotation
  (clone machinery — superseded by 1:N + explicit `create cell`).

## Acceptance criteria

- [ ] One name generator serves both `--from-blueprint` and `--from-config`; explicit name used verbatim, collision in scope errors.
- [ ] Omitted name yields `<prefix>-<6hex>` with `prefix = Spec.Prefix ?? metadata.name`, retrying on suffix collision.
- [ ] `CellConfigSpec.Prefix` added (optional) and honored.
- [ ] `StableName` reduced to a prefix-default helper; no remaining identity-pin callers.
- [ ] `<src>-<N>` counter and `kukeon.io/source-config` annotation removed; no remaining references.
- [ ] `go build ./...` + `go test ./...` clean; `make dev-init` smoke passes.

## Dependencies

Umbrella: #1020. Depends on: #1021 (P1). Blocks: P5 (this epic).

PR #1092, opened 2026-06-05
feat(create): add --clone <cell> source kind to fork a cell's recipe

### LINKED ISSUE #1073: clone: --clone <cell> source kind on create + run (P4)
## Context

Fourth step of epic:cell-identity (umbrella #1020). Adds `--clone <cell>` as a third source kind
alongside `--from-blueprint` and `--from-config`, closing the cell-forking gap left by retiring
`--clone <config>` in P2.

Today's `--clone` forks a CellConfig (with the `<src>-<N>` counter and `kukeon.io/source-config`
annotation) — that machinery is rightfully retired in P2 as a 1:1-era workaround. The underlying
use case "fork this specific cell's recipe (with its applied overrides) into a new cell" survives
the redesign though, and post-P3 a cell can carry per-cell `--env`/`--param` overrides that
`--from-config <binding>` alone cannot reproduce.

P1 makes this clean to implement: `Spec.Provenance = {bindingKind, bindingRef, params,
envOverrides}` is the complete recipe. `--clone` copies that block.

## Proposal

Third source kind on both the un-fused primitive and the fused composition:

```
kuke create cell --clone existingCell                       # auto-name <existing-prefix>-<6hex>
kuke create cell myNewCloneName --clone existingCell        # verbatim name
kuke run --clone existingCell                               # fused create+start+attach
kuke run --clone existingCell --name myNewCloneName
```

- **Mechanism**: read source cell's `Spec.Provenance` (P1); write a new CellDoc with that block
  copied verbatim; run the same materialization function as `--from-blueprint` / `--from-config`.
- **Naming**: uses the unified P2 generator. Prefix derives from the source cell
  (`Spec.Prefix ?? metadata.name`). Explicit name → verbatim; omitted → `<prefix>-<6hex>`.
- **Lineage inheritance**: clone carries the source's `kukeon.io/config` or `kukeon.io/blueprint`
  lineage label, so it joins the same fleet for selector-driven reconcile (P4 — now P5).
- **New annotation**: `kukeon.io/source-cell: <source-cell-name>` as inert provenance metadata.
  Distinct from the rejected `kukeon.io/source-config` — that annotation was load-bearing for the
  `<src>-<N>` counter; `source-cell` carries no behavior, no identity pin, only debug/grooming value.
- **Flag exclusivity**: `--clone`, `--from-blueprint`, and `--from-config` are mutually exclusive.
  Combining any two errors at parse time.
- **Override copy + stacking**: per-cell `--env`/`--param` overrides on the source (P3) copy to the
  clone via the provenance block; additional `--env`/`--param` on the clone command stack on top
  (last-write-wins, same precedence rule as P3).
- **FlagSet integration**: `--clone` is registered on the shared `FlagSet` introduced by P5 (now P6),
  so adding it on the `create cell` side automatically makes `run --clone` work — no second
  registration. This is the same composition mechanism `--from-blueprint` / `--from-config` use.
- **Scope**: clone targets the source cell's scope (realm/space/stack). Cross-scope cloning is a
  separate concern, not in scope here.

## Non-goals

- **Runtime state cloning** (overlay, volume contents) is explicitly out of scope. Overlay is
  wiped on container recreate as it always has been. Volume-backed memory (epic #1015) is
  referenced by name — cloning a cell that mounts volume `agent-mem` references the same volume
  by default. A future `--clone-volumes` story (deep-copy vs share semantics) belongs to
  epic:volumes, not here.
- Cross-realm / cross-scope cloning.
- Idempotent-attach on name collision (explicit name + existing cell still errors, per P2).

## Acceptance criteria

- [ ] `kuke create cell --clone <src>` and `kuke run --clone <src>` produce a new cell whose `Spec.Provenance` is byte-equal to the source's.
- [ ] Omitted name yields `<source-prefix>-<6hex>`; explicit name used verbatim; collision in scope errors (not idempotent-attach).
- [ ] Clone inherits the source's `kukeon.io/config` or `kukeon.io/blueprint` lineage label.
- [ ] Clone carries a `kukeon.io/source-cell: <src-name>` annotation; provenance block excluded from `DiffCell` (no false OutOfSync).
- [ ] `--clone`, `--from-blueprint`, `--from-config` are mutually exclusive — combining any two errors with a clear message.
- [ ] Additional `--env`/`--param` on the clone command merge on top of copied overrides (override-on-override precedence, last write wins).
- [ ] `--clone` registered once on the shared `FlagSet` so `run --clone` works without separate plumbing once P5 (now P6) lands; until then `run --clone` is unrecognized and `create cell --clone` is the supported entry point.

## Notes

- Use case context: today's `--clone <config>` covered both "stamp another cell from this binding"
  (handled natively post-1:N by a second `create cell --from-config`) and "fork this tuned cell's
  recipe" (this issue). Retiring the former without this issue would silently regress the latter.
- Cross-epic: volume-backed agent memory (#1015) is referenced by name on the clone; a Breaking
  reconcile that recreates the container preserves volume contents. Deep-copying volume state on
  clone is a follow-up against epic:volumes.

## Dependencies

Umbrella: #1020. Depends on: #1021 (P1 — provenance block to copy), #1022 (P2 — unified naming
generator), #1023 (P3 — per-cell overrides exist to be copied). Blocks: none (P5/P6/P7 run in
parallel; if this lands before the run step, that step inherits `--clone` via the shared FlagSet
for free).

eminwux:
## Summary

Adds `--clone <cell>` as the third source kind on `kuke create cell`, alongside `--from-blueprint` and `--from-config` (epic:cell-identity P4, #1073). It closes the cell-forking gap left when P2 (#1022) retired the old `--clone <config>` `<src>-<N>`/`source-config` machinery: the "fork *this* tuned cell's recipe (with its per-cell overrides) into a new cell" use case survives the redesign and is now served natively.

**Mechanism** (per the issue's design): read the source cell's `Spec.Provenance` (P1) → re-materialise a fresh spec from that same binding (the exact `cellconfig`/`cellblueprint` materialisation the `--from-config`/`--from-blueprint` paths run) → copy the source's provenance onto the clone verbatim → inherit the lineage label, stamp the `kukeon.io/source-cell` annotation, target the source's scope, finalise the name via the unified P2 generator.

- Per-path override symmetry from P3 holds: `--env` stacks on a config-lineage clone (last-write-wins; `--param`/`--param-file` rejected); `--param`/`--param-file` stack on a blueprint-lineage clone (`--env` rejected).
- `--clone`, `--from-blueprint`, `--from-config` are mutually exclusive (cobra mutex).
- Clone logic lives in its own `cmd/kuke/create/cell/clone.go` (+ `clone_test.go`) to keep `cell.go` reviewable.

## Two AC-specifics that didn't match the cell model (heads-up posted on #1073; proceeded with the documented readings)

1. **Prefix source.** The AC derives the clone prefix from the source cell's `Spec.Prefix ?? metadata.name`, but a materialised cell (`v1beta1.CellSpec`) has **no `Prefix` field** — only `CellConfigSpec`/`CellBlueprintSpec` do, and `MaterializeWithName` never copies it onto the cell. So the `??` collapses to the source cell's `metadata.name`, which is what I use. **Consequence:** cloning an auto-named source `web-a1b2c3` yields `web-a1b2c3-<6hex>` (the prefix accumulates). I did not add suffix-stripping or a new `CellSpec.Prefix` field (both exceed the literal AC and are larger changes). If the intent was the *original* prefix, that needs an explicit decision.

2. **`kukeon.io/source-cell` annotation.** Cells carried only `metadata.Labels` — no `Annotations` map. A `source-cell` *label* would be compared by `DiffCell` (its `filterManagedLabels` drops only the `.kukeon.io` *suffix*, not the `kukeon.io/` *prefix*), tripping a false OutOfSync on config-lineage clones — exactly what AC#4 forbids. So I added an `Annotations map[string]string` field to the cell model (mirroring the existing `CellConfig`/`CellBlueprint` annotations), threaded through `apischeme` both directions. `DiffCell` ignores annotations by nature → no false-OutOfSync, "inert provenance metadata" intent preserved literally.

## Scope notes

- **`run --clone` is intentionally not wired here.** Per AC#7, `--clone` rides the shared `FlagSet` introduced by P6; until that lands, `run --clone` is unrecognised and `create cell --clone` is the supported entry point. This PR only registers `--clone` on `create cell`.
- **Stale `run --clone` docs left untouched.** `docs/cli-use-cases.md` and `docs/site/cli/kuke-run.md` still describe the P2-retired `run <config> --clone` `<src>-<N>`/`source-config` machinery. That is pre-existing P2 doc debt for a different (run-side) feature and belongs to the epic's docs step — out of scope here (no unrelated cleanups).

## Test plan

- [x] `make test` (test-root + test-kukebuild) — passes (`GOWORK=off`, the Makefile-exported env).
- [x] `go build ./...` + `go vet ./...` (`GOWORK=off`) — clean.
- [x] `pre-commit run` (go-fmt, golangci-lint, addlicense, …) on the change set — passes.
- [x] New unit tests in `cmd/kuke/create/cell/clone_test.go`: provenance byte-equal (AC#1), auto-name `<src>-<6hex>` + explicit-name verbatim + collision-errors (AC#2), lineage-label inheritance (AC#3), `source-cell` annotation (AC#4), mutual exclusivity (AC#5), env/param stacking last-write-wins (AC#6), source-not-found, source-without-provenance.
- [x] `internal/controller/apply/diff_test.go`: `TestDiffCell_Annotations_NoChange` pins AC#4's no-false-OutOfSync (annotation-only diff ⇒ Synced).
- [x] `internal/apischeme/scheme_test.go`: `TestCellRoundTripV1Beta1_Annotations` pins the annotation external→internal→external round-trip.
- [ ] `make dev-init` smoke — **deferred (environmental).** This host's `/var/lib/containerd` is a shared **4 GB tmpfs at 93 % full** (accumulated build leftovers from prior runs); the kukeond image build runs out of space (`no space left on device`), and aggressively pruning the *shared parent-host* containerd would risk unrelated parent `kuke attach` state. `kuke image prune` needs a running daemon (none is up post-reset). Marginal value here is low: the smoke never exercises the clone path, and the only daemon-read impact — the additive `Annotations` field — is covered by the apischeme round-trip + DiffCell unit tests, and `make test` compiles the daemon. Recommend running `make dev-init` on a clean host before merge.

## Acceptance criteria

- [x] `create cell --clone <src>` produces a new cell whose `Spec.Provenance` is byte-equal to the source's (run `--clone` deferred to P6's shared FlagSet per AC#7).
- [x] Omitted name → `<source-name>-<6hex>`; explicit name verbatim; in-scope collision errors (not idempotent-attach).
- [x] Clone inherits the source's `kukeon.io/config` / `kukeon.io/blueprint` lineage label.
- [x] Clone carries `kukeon.io/source-cell: <src>`; provenance excluded from `DiffCell` (no false OutOfSync) — verified by test.
- [x] `--clone`/`--from-blueprint`/`--from-config` mutually exclusive with a clear message.
- [x] Additional `--env`/`--param` merge on top of copied overrides (last-write-wins), per the P3 per-path symmetry.
- [x] `--clone` registered on `create cell` only; `run --clone` waits on P6's shared FlagSet (AC#7).
- [x] `go build ./...` + `go test ./...` clean; `make dev-init` smoke deferred (environmental — see Test plan).

Closes #1073

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -213,6 +213,13 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_CREATE_CELL_FROM_CONFIG = DefineKV("KUKE_CREATE_CELL_FROM_CONFIG", "kuke/create/cell/from-config")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	// KUKE_CREATE_CELL_CLONE is the env-var twin of `kuke create cell --clone
+	// <src>` (epic:cell-identity #1073): the third source kind alongside
+	// --from-blueprint / --from-config. Forks the named source cell's recipe
+	// (its Spec.Provenance, with any per-cell --env/--param overrides) into a
+	// new cell.
+	KUKE_CREATE_CELL_CLONE = DefineKV("KUKE_CREATE_CELL_CLONE", "kuke/create/cell/clone")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_CREATE_CELL_PARAM_FILE = DefineKV("KUKE_CREATE_CELL_PARAM_FILE", "kuke/create/cell/param-file")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	// KUKE_CREATE_CELL_IGNORE_DISK_PRESSURE is the env-var twin of
--- a/cmd/kuke/create/cell/cell.go
+++ b/cmd/kuke/create/cell/cell.go
@@ -102,6 +102,16 @@ func NewCellCmd() *cobra.Command {
 			"Mutually exclusive with --from-blueprint.")
 	_ = viper.BindPFlag(config.KUKE_CREATE_CELL_FROM_CONFIG.ViperKey, cmd.Flags().Lookup("from-config"))
 
+	cmd.Flags().String("clone", "",
+		"Fork an existing cell's recipe into a new cell: read the source cell's "+
+			"Spec.Provenance (the binding it was materialised from plus any per-cell "+
+			"--param/--env overrides), re-materialise from that same binding, and persist a "+
+			"new stopped cell. The clone copies the source's Spec.Provenance verbatim, inherits "+
+			"its kukeon.io/config or kukeon.io/blueprint lineage label, and carries a "+
+			"kukeon.io/source-cell=<src> annotation. Omitted name → <source-name>-<6hex>; "+
+			"explicit name used verbatim. Mutually exclusive with --from-blueprint/--from-config.")
+	_ = viper.BindPFlag(config.KUKE_CREATE_CELL_CLONE.ViperKey, cmd.Flags().Lookup("clone"))
+
 	cmd.Flags().StringArray("param", nil,
 		"Scalar parameter override as KEY=VALUE; repeatable. Valid with --from-blueprint. "+
 			"Each KEY must be declared in spec.parameters[]. Wins over the parameter's default "+
@@ -131,7 +141,7 @@ func NewCellCmd() *cobra.Command {
 		cmd.Flags().Lookup("ignore-disk-pressure"),
 	)
 
-	cmd.MarkFlagsMutuallyExclusive("from-blueprint", "from-config")
+	cmd.MarkFlagsMutuallyExclusive("from-blueprint", "from-config", "clone")
 
 	// Register autocomplete functions for flags
 	_ = cmd.RegisterFlagCompletionFunc("realm", config.CompleteRealmNames)
@@ -139,6 +149,7 @@ func NewCellCmd() *cobra.Command {
 	_ = cmd.RegisterFlagCompletionFunc("stack", config.CompleteStackNames)
 	_ = cmd.RegisterFlagCompletionFunc("from-blueprint", config.CompleteBlueprintNames)
 	_ = cmd.RegisterFlagCompletionFunc("from-config", config.CompleteConfigNames)
+	_ = cmd.RegisterFlagCompletionFunc("clone", config.CompleteCellNames)
 
 	return cmd
 }
@@ -152,8 +163,13 @@ type createCellFlags struct {
 	stack         string
 	blueprintName string
 	configName    string
-	paramArgs     []string
-	paramFile     string
+	// cloneSource is the source cell name passed via `--clone <src>` (the
+	// third source kind, epic:cell-identity #1073). When set, the cell is
+	// forked from the source cell's Spec.Provenance rather than resolved from a
+	// Blueprint/Config named on the CLI.
+	cloneSource string
+	paramArgs   []string
+	paramFile   string
 	// envArgs holds the validated `--env KEY=VALUE` per-cell overrides. Valid
 	// with --from-config only (parity with --param on --from-blueprint); baked
 	// into the attachable container's env and recorded in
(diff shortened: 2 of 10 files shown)


eminwux:
**PR #1092 Review — `feat(create): add --clone <cell> source kind to fork a cell's recipe`**

LGTM — third source kind is cleanly factored into `clone.go`, the verbatim provenance copy + override stacking respect P3 per-path symmetry, `kukeon.io/source-cell` is excluded from `DiffCell` so config-lineage clones stay Synced, and the helpers wire up with the right semantics (`CloneCellProvenance` deep-copies `EnvOverrides`/`Params`; `mergeEnv` is last-write-wins on the flag side; `finalizeCellName` routes through the unified P2 `<prefix>-<6hex>` generator; pre-persist collision check is in place). Both flagged AC deviations (prefix collapses to `metadata.name` since materialised `CellSpec` carries no `Prefix`; added `Annotations` to the cell model so `DiffCell` can ignore the source-cell mark) are well-justified and proportionate — the alternatives (suffix-stripping, label-based provenance) would either exceed AC scope or trip `filterManagedLabels` exactly as you noted.

Pre-merge: run `make dev-init` smoke on a clean host — deferred here due to the 93%-full shared tmpfs at `/var/lib/containerd`. The clone path itself isn't exercised by dev-init, but the additive `Annotations` field on the cell model is daemon-read, so the round-trip pair (`TestCellRoundTripV1Beta1_Annotations` + `TestDiffCell_Annotations_NoChange`) covers the only realistic regression surface.

Issue #1015, opened 2026-06-01
[umbrella] Daemon-managed named Volume kind (VolumeKind=volume) — durable storage for stateful agent cells


eminwux:
## Summary

Introduce a **daemon-managed named `Volume`** — a standalone resource (`kind: Volume`) with its **own lifecycle**, independent of any cell — that **persists across container recreation _and_ cell deletion**. A cell *references* a Volume (it does not own it), the way a container references a `Secret`. This is the durable-storage primitive for **stateful agent cells**: cells whose accumulated writable state (memory, learnings, caches) *is* the asset, must survive an in-place spec reconcile (add an env var → container recreated), and must **outlive any single cell instance** so a successor agent can inherit it.

## Motivation

Today a cell has three storage tiers, and only one is durable — but it is unmanaged:

| Tier | What it is | Survives container *recreate* (e.g. env-var change)? | Survives *stop*? |
| --- | --- | --- | --- |
| Rootfs overlay (snapshot, `ctr/container.go:373` `WithNewSnapshot`) | per-container writable layer | ❌ wiped — `WithSnapshotCleanup` at `:490`, fresh snapshot on recreate | ✅ |
| `tmpfs` (`VolumeKindTmpfs`) | RAM-backed scratch | ❌ | ❌ wiped on stop |
| `bind` (`VolumeKindBind`) | host path → container path | ✅ durable | ✅ |

**Stateful agents** are the driver. Cells are agents that start and finish but whose data we keep — they accumulate memory/learning that belongs to the harness, and even cron runs reuse them for that accrued expertise. Two things follow: (1) updating a cell's spec (add an env var) must **not** lose that state, but the update path recreates the container — `StartCell` → `RecreateCell` (`start_cell.go:195`) → fresh snapshot — so durable state cannot live in the overlay; and (2) the expertise belongs to the *harness, not to one ephemeral cell*, so **deleting or rebuilding the cell must not erase it**. A raw `bind` mount gives durability but is unmanaged (hand-picked host paths, tied to nothing). A named `Volume` is the managed, daemon-owned, **independently-lifecycled** substrate that satisfies both.

This primitive is also presupposed by issues that have no way to express it yet:
- **#738** — `apischeme: reject/warn runOn: create when container has no persistent writable mount`. "Persistent writable mount" *is* this volume.
- **#779** — `agentrun: onEnd.persist workspace export`. Workspace persistence needs a durable, daemon-managed home.
- **#503** — AgentRun run-envelope.

[...]

PR #1098, opened 2026-06-05
feat(reconcile): in-place apply for Compatible OutOfSync diffs, recreate for Breaking (P7)

### LINKED ISSUE #1095: reconcile: in-place apply for Compatible diffs, recreate for Breaking (P7)
## Context

Seventh step of epic:cell-identity (umbrella #1020). Sliced out of the original P5 (#1024) during
re-size: this is the **runtime-behavioral half** of P5 and concentrates its risk. Faithful AC6 —
container overlay **preserved** on a Compatible diff, **wiped** on a Breaking diff — cannot be
satisfied on the current stop-then-start reapply model and needs genuine runner/lifecycle work
plus on-device verification.

The constraints (all on `main`):
- `reapplyOutOfSyncFromConfig` runs on a **stopped** cell (`kuke restart` stops first; `kuke start`
  is already stopped).
- `UpdateCell` (`internal/controller/runner/update_cell.go`) — the in-place child-restart path —
  cannot run on a stopped cell: non-root children join the root container's pid/net/ipc
  namespaces, so starting a child while the root is down fails.
- A plain `StartCell` of a Compatible non-root edit is refused: `ComputeContainerSpecHash`
  (`internal/controller/runner/spec_hash.go:89`) covers image/command/args with an identical
  domain for root and non-root, and `reuseOrRefuseExistingChildContainer`
  (`internal/controller/runner/spec_hash.go:183`) returns `ErrCellSpecHashDrift` on a changed
  child spec.

Net: a Compatible non-root edit today can only be applied by `RecreateCell` (full overlay wipe —
violates AC6) or by restructuring the reapply path to `UpdateCell`-in-place for Compatible diffs.

## Proposal

- Restructure the OutOfSync reapply path (`reapplyOutOfSyncFromConfig`) so a **Compatible** diff
  updates in place — env/ports/volumes per the `diff.go` Compatible domain, or a non-root
  child-container restart via `UpdateCell` — preserving the container overlay.
- A **Breaking** diff continues to drive `RecreateCell` (overlay wiped; volume-backed memory
  survives once #1015 lands).
- Verify the already-satisfied informational-OutOfSync behavior: `reconcileCellOutOfSync`
  (`internal/controller/reconcile_outofsync.go:65`) detects + persists the flag only; healing
  fires solely on explicit `StartCell` — no background auto-heal tick.

## Acceptance criteria

- [ ] A Compatible edit updates in place on the next start (env/ports/volumes per `diff.go`) or via non-root child-container restart (`UpdateCell`); the container overlay is preserved.
- [ ] A Breaking edit drives `RecreateCell` (overlay wiped; volume-backed memory survives once #1015 lands).
- [ ] OutOfSync surfaces as informational status; no background auto-heal tick is introduced (verify existing `reconcileCellOutOfSync` detection-only behavior).
- [ ] `go build ./...` + `go test ./...` clean; `make dev-init` smoke passes; on-device verification that a Compatible reapply preserves overlay and a Breaking reapply recreates.

## Dependencies

Umbrella: #1020. Depends on: #1024 (P5 — re-resolve produces the `desired` this path classifies),
epic:reconcile (#990–992, landed). Related: #1015 (volumes — a Breaking recreate preserves
volume-backed memory once it lands). Blocks: #1026 (P9 — docs).

eminwux:
## Summary

- Restructures the daemon-side OutOfSync reapply path (`reapplyOutOfSyncFromConfig` in `internal/controller/start_cell.go`) so a `kuke start`/`kuke restart` of an OutOfSync Config-lineage cell routes by **diff class** instead of always recreating:
  - **Breaking diff** (root-container baked fields, host-namespace toggles — the `RecreateCell` domain `apply.ReconcileCell` already uses) → `RecreateCell`, which tears down and rebuilds the cell. The container overlay is wiped (volume-backed memory survives once #1015 lands).
  - **Compatible / Additive diff** → in-place apply with the root overlay **preserved**: `StartCell` restarts the cell on its existing containerd snapshot (issue #867 spec-hash reuse), then `UpdateCell` re-persists the compatible metadata (env/ports/volumes) and stop-remove-recreate-starts any non-root child whose image/command/args changed.
- The `StartCell`-then-`UpdateCell` ordering is load-bearing and called out in the godoc: `UpdateCell` cannot run on a stopped cell (non-root children join the root's pid/net/ipc namespaces), and starting from the *materialised* spec would trip `ErrCellSpecHashDrift` on a changed child — so the cell is started on its on-disk spec first (reusing the snapshot) and the drift is applied afterward, leaving the root container and its overlay untouched.
- **AC3 (OutOfSync stays informational; no background auto-heal):** verified, no code change. `reconcileCellOutOfSync` (`internal/controller/reconcile_outofsync.go`) only detects and persists the flag via `UpdateCellMetadata`; the sole caller in the reconcile loop (`reconcile.go:115`) does not chain into any `StartCell`/`RecreateCell` heal. Healing fires only on explicit `StartCell`.

## Implementation notes

- The pre-existing reapply unit tests exercised a **non-root child image bump** (`nginx:1.0 → nginx:2.0`), which `apply.DiffCell` classifies as **Compatible** (image is breaking only on the *root* container). Those tests asserted `RecreateCell` was called — exactly the behavior AC1 changes. They are updated to assert the new in-place routing, and a new Breaking-path test uses a `hostNetwork` toggle (breaking even on a non-root child) to drive `RecreateCell`.
- `diff.ChangeType == apply.ChangeTypeBreaking` is the sole branch condition: a root-container breaking change and any other breaking change both set `ChangeType` to Breaking, so this subsumes `ReconcileCell`'s two-part breaking check for the reapply path.
- Graceful degradation preserved: a `RecreateCell`/`StartCell` failure falls back to the caller's on-disk `StartCell` (cell still bounced as the operator asked); a Compatible-path `UpdateCell` failure leaves the cell Ready (started on the on-disk spec) with the OutOfSync flag still set so the next start retries.

## Test plan

- [x] `make test` (full CI test target, `GOWORK=off`) — green, exit 0.
- [x] `go build ./internal/controller/...` + `go vet ./internal/controller/...` (`GOWORK=off`) — clean.
- [x] `pre-commit run --files <changed>` — passes (golangci-lint, go fmt, addlicense, …).
- [x] New/updated unit tests for the routing: `TestStartCell_OutOfSyncReapply_CompatibleAppliesInPlace`, `_BreakingRecreates`, `_CompatibleUpdateErrorStillReady`, `_EnvOverridePreserved` (now asserts the override survives the in-place `UpdateCell`), `_RecreateErrorFallsThrough` (now Breaking-path).
- [ ] `make dev-init` smoke — **blocked, environmental (not code).** The kukeond image build fails with `no space left on device` inside the BuildKit sandbox: this host's containerd store (`/var/lib/containerd`) is a **4 GB tmpfs at 93% full**, too small for the full `-a` Go rebuild the image build runs. `make test` (which covers the same daemon code) is green. The smoke should pass on a clean host / CI with adequate disk.
- [ ] On-device verification that a Compatible reapply preserves the overlay and a Breaking reapply recreates — **deferred to a clean host.** Depends on `make dev-init` standing up the daemon (blocked above). The Breaking-vs-Compatible **routing** (which runner method runs per diff class) is verified by the unit tests; the byte-on-disk overlay survival is a downstream consequence of `StartCell`'s #867 snapshot-reuse vs `RecreateCell`'s teardown (each already covered) and is the one observable left unverified in this environment.

Closes #1095

### DIFF
--- a/internal/controller/start_cell.go
+++ b/internal/controller/start_cell.go
@@ -84,9 +84,11 @@ func (b *Exec) StartCell(cell intmodel.Cell) (StartCellResult, error) {
 	}
 
 	// Reapply the lineage Config when the persisted OutOfSync flag is set, so
-	// the start runs against the freshly materialised spec — see #983. If the
-	// reapply already brought the cell back to Ready (RecreateCell path), the
-	// caller is done; otherwise fall through to the regular runner.StartCell.
+	// the start runs against the freshly materialised spec — see #983. A
+	// Breaking diff recreates the cell; a Compatible diff is applied in place
+	// with the container overlay preserved (epic:cell-identity P7, #1095). If
+	// the reapply already brought the cell back to Ready, the caller is done;
+	// otherwise fall through to the regular runner.StartCell.
 	if reapplied, started, ok := b.reapplyOutOfSyncFromConfig(internalCell); ok {
 		internalCell = reapplied
 		if started {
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1098 Review — `feat(reconcile): in-place apply for Compatible OutOfSync diffs, recreate for Breaking (P7)`**

LGTM — clean split of the reapply path on `diff.ChangeType`. Routing aligns with `apply.DiffCell`'s existing breaking-vs-compatible domain and `UpdateCell`'s own docstring contract ("Breaking changes should be rejected before calling this method"), the StartCell-then-UpdateCell ordering is correctly load-bearing for namespace joins, and the degradation paths in both `reapplyBreaking` and `reapplyCompatibleInPlace` honor the operator's start request when the runtime hop fails. Test coverage is symmetric with the new branching (`_CompatibleAppliesInPlace`, `_BreakingRecreates`, `_CompatibleUpdateErrorStillReady`, `_EnvOverridePreserved` updated, `_RecreateErrorFallsThrough` re-pointed at the Breaking blueprint). No new error sentinels, log/error messages carry only cell+config identifiers (no leaks).

One observation, non-blocking:

- **Docstring rationale** `internal/controller/start_cell.go:240-251`: \"Starting from `desired` instead would trip ErrCellSpecHashDrift\" is technically imprecise — `runner.StartCell` re-loads the cell via `GetCell` (start.go:528) and only honors `RuntimeEnv` from the passed-in spec, so passing `desired` (same identity) would behave identically. The real safety rationale is `UpdateCell`'s Breaking-rejection contract + the on-disk spec-hash matching the existing containerd record. Not worth a respin.

Deferred host-gated verifications named in the PR body (`make dev-init` smoke, on-device overlay survival) are environmental — out-of-scope for the agent sandbox per the carve-out. Worth running on a clean host pre-merge to confirm the overlay-survival observable end-to-end.

Issue #1023, opened 2026-06-02
create: --param/--env symmetry baked into the CellDoc (P3)


eminwux:
## Context

Third step of epic:cell-identity (umbrella #1020). Makes `create cell` the single materialization
verb with symmetric override flags on both source paths, baked into the CellDoc (and recorded as
P1 materialization inputs).

epic:bye-container #996 already requires a source flag on `create cell` (empty-shell retired). This
step adds the override surface:

```
kuke create cell <name> --from-blueprint <bp>  --param K=V   # --param resolves ${KEY}
kuke create cell <name> --from-config    <cfg> --env   K=V   # --env overrides
```

## Proposal

- **`--param` on `--from-blueprint`**: formalize (it exists today, "valid with `--from-blueprint`
  only"); ensure values are baked into the CellDoc and recorded in P1's `params`.
- **`--env` on `--from-config`**: add it; **reverse the 1:1-era refusal** at
  `cmd/kuke/create/cell/cell.go:181` ("--param is not valid with --from-config"). Bake into the
  CellDoc and record in P1's `envOverrides`. Plumbing half-exists: `Spec.RuntimeEnv` is carried in
  `createCellInternal` (`create_cell.go:254`, #834) — but here it must *persist* as an override,
  not the non-persisted runtime form.
- **Precedence**, documented + tested: binding values → blueprint params → per-cell `--env`/`--param`
  last (overrides win). This is the same merge P4 re-applies on re-resolve.
- Rejected alternative: `--env`/`--param` spawning a *new* CellConfig per tweak (binding
  proliferation). Overrides live on the cell.

## Acceptance criteria

- [ ] `kuke create cell <name> --from-blueprint <bp> --param K=V` bakes params into the CellDoc and records them in provenance.
- [ ] `kuke create cell <name> --from-config <cfg> --env K=V` is accepted (old refusal removed), baked, and recorded.
- [ ] Override precedence (binding → params → per-cell env/param) documented inline and covered by a test.
- [ ] Per-cell overrides persist (distinct from the non-persisted `Spec.RuntimeEnv` runtime path).
- [ ] `go build ./...` + `go test ./...` clean; `make dev-init` smoke passes.

## Dependencies

Umbrella: #1020. Depends on: #1021 (P1). Blocks: P4, P5 (this epic).

Issue #823, opened 2026-05-23
refactor(cli): remove kuke apply -b/-c — apply becomes apply -f only (step 4)


eminwux:
## Context

Child of #819. With `kuke restart cell` providing the reconcile path for Config-driven workflows (sibling child issue) and OutOfSync surfaced in `kuke get cell` (sibling child issue), `kuke apply -b/-c` no longer has a unique role. The umbrella's framing: `apply` is for declarative-from-YAML (`apply -f`); reconcile-by-ref retires in favor of OutOfSync + restart.

No deprecation cycle (per umbrella scope: no production kukeon yet).

## Proposal

Hard-remove the `-b/-c` flags from `kuke apply`:

- Drop `-b/--blueprint` and `-c/--config` flag definitions from `cmd/kuke/apply/apply.go`
- Drop the `runApplyRef` code path; `runApply` always routes to `runApplyFile`
- Drop the `loadFromBlueprint`, `loadFromConfig`, `assertCellLineage` helpers from `cmd/kuke/apply/apply.go` (the same materialisation logic in `cmd/kuke/run/run.go` for `run -b/-c` stays untouched)
- Retire apply-side viper keys (`KUKE_APPLY_BLUEPRINT`, `KUKE_APPLY_CONFIG`, related --param/--name on apply, etc.). Verify before deleting: some may be shared with `kuke run`'s scope resolution and need to stay or be split
- Update `cmd.MarkFlagsMutuallyExclusive` / `MarkFlagsOneRequired` to reflect `-f` as the only source
- Update the `Use:` line and `Long` help text to describe `apply -f` only

## Acceptance criteria

- [ ] `kuke apply -b ...` errors with cobra's "unknown flag" message
- [ ] `kuke apply -c ...` errors with cobra's "unknown flag" message
- [ ] `kuke apply -f cell.yaml` continues to work as today
- [ ] `cmd/kuke/apply/apply.go` `Use:` line and `Long` text describe only the `-f` shape
- [ ] `runApplyRef`, `loadFromBlueprint`, `loadFromConfig`, `assertCellLineage` are removed from `cmd/kuke/apply/apply.go`
- [ ] Viper keys for apply-side `-b/-c` are retired (verify before deleting — some may be shared with run-side and need to stay or be split)
- [ ] Shell completion registrations for `--blueprint` / `--config` on `kuke apply` are removed
[...]

PR #1099, opened 2026-06-06
feat(run): cell-positional + fused create+start+attach delegation (P8)

### LINKED ISSUE #1025: run: cell-positional + fused create+start+attach delegation (P8)
## Context

Eighth step of epic:cell-identity (umbrella #1020). Makes `run` cell-positional and fuses it as
create+start+attach — the docker model — while keeping `create cell` / `start` as the un-fused
primitives.

Today `run` takes a CellConfig positional plus source/identity flags (`-b`, `--new`, `--clone`,
`--reuse`). With identity on the CellDoc, the cell is what you run.

```
docker create  ->  kuke create cell      (materialize, stopped)
docker start   ->  kuke start <name>      (start existing, no attach; positional post-bye-container #998)
docker run     ->  kuke run               (create? + start + attach)
```

## Proposal

- **`run <cell>`** = start + attach an existing cell (≈ `start <name>` + attach). Worked examples
  of the cell-positional form with the surviving flags:

  ```
  kuke run <cell>                          # start + attach
  kuke run <cell> -d                       # start, no attach
  kuke run <cell> --container <child>      # attach to a specific child container
  kuke run <cell> --rm                     # start + attach + delete on exit
  ```
- **`run --from-blueprint <bp> --param …` / `run --from-config <cfg> --env …`** = create + start +
  attach. The create half **delegates to the P3 `create cell` path** — same materialization
  function, same flag set registered via a shared cobra `FlagSet`, so the two verbs cannot drift
  (this is composition, not a second implementation; `docker run` accepting every `docker create`
  flag is the same pattern).
- **`run --clone <cell>` rides on the same shared FlagSet** (registered by P4), so the fused
  create+start+attach form picks up `--clone` for free with no separate plumbing.
- **Remove from `run`**: the `<config>` positional, `-b/--blueprint`, `--new`, `--clone` (as a
  boolean), `--reuse`. (`--name`, `--detach`, `--container`, scope flags, `--rm` stay.) Note: the
  *new* value-bearing `--clone <cell>` flag from P4 is unrelated to the retired boolean `--clone`.
- `start`/`create cell` keep their non-fused behavior.

## Acceptance criteria

- [ ] `kuke run <cell>` starts + attaches an existing cell; with `-d` it starts + detaches.
- [ ] `kuke run --from-blueprint <bp> --param K=V` and `kuke run --from-config <cfg> --env K=V` create + start + attach, producing a CellDoc identical to the equivalent `create cell` + `start`.
- [ ] `kuke run --clone <cell>` create + start + attach via the same shared FlagSet (P4 registered the flag on the create side).
- [ ] `run` and `create cell` share one `FlagSet` and one materialization function (no duplicated flag definitions).
- [ ] `-b`, the `<config>` positional, `--new`, the old boolean `--clone`, `--reuse` are removed from `run`; invoking them errors.

## Dependencies

Umbrella: #1020. Depends on: #1022 (P2), #1023 (P3). Related: #1073 (P4 — `--clone` source kind,
inherited via shared FlagSet if landed first; safe to land in either order). Blocks: #1026 (P9 — docs).

eminwux:
## Summary

Eighth step of epic:cell-identity (umbrella #1020). Makes `kuke run` cell-positional and fuses it as create+start+attach — the docker model — while keeping `create cell` / `start` as the un-fused primitives.

- **`run <cell>`** now starts + attaches an *existing* cell (≈ `start <cell>` + `attach`); `-d` starts without attaching. A missing cell errors with a pointer at the create paths.
- **`run --from-blueprint <bp> --param …` / `run --from-config <cfg> --env …` / `run --clone <cell>`** create + start + attach. The create half **delegates to the same materialization function `kuke create cell` runs** (`cell.Materialize`), so the produced `CellDoc` is identical to `create cell --from-…` followed by `start`. run then `CreateCell`s it (create cell would `MaterializeCell` and leave it stopped) and attaches.
- **Shared FlagSet + materialization function (AC#4):** the source flags (`--from-blueprint/--from-config/--clone/--param/--param-file/--env/--ignore-disk-pressure`) are defined once in `cell.RegisterSourceFlags` and both verbs call it; the materialization logic is the single `cell.Materialize` entrypoint. No duplicated flag definitions, no drift.
- **Removed from `run`:** the `<config>` positional meaning (positional is now an existing cell), `-b/--blueprint`, `--new`. Invoking `-b`/`--blueprint`/`--new` now errors with "unknown flag". `--name`, `-d/--detach`, `--container`, scope flags, `--rm`, `-f`, `-o`, `--require-synced` stay.
- **`--require-synced` rejected outside `-f`** (fix-up b22f447): the flag compares the live cell against a source manifest, which only `-f` supplies — the fused `--from-*`/`--clone` paths refuse on `--name` collision rather than diverge, and the `<cell>` positional has no source to compare against. Rejecting at parse time prevents the silent-no-op flagged in review; pinned in `TestRun_RequireSynced_RejectedOutsideFile`.

### Design notes for the reviewer

- **`-f/--file` kept (lower-blast-radius call).** The AC's removal list (`<config>` positional, `-b`, `--new`, old boolean `--clone`, `--reuse`) does **not** include `-f`, and `-f` is a distinct manifest source not part of the shared create-cell FlagSet. Removing it would be a larger behavior change the AC doesn't mandate, so `-f` (and its create-or-attach + divergence + `--require-synced` machinery) is preserved unchanged. The fused `--from-*` paths don't compare against a source (always fresh, or refuse on `--name` collision), so the divergence/`--require-synced` logic is now reachable only via `-f`.
- **Premise rot (documented per dev guidance):** the issue says to remove the *old boolean* `--clone` and `--reuse` from `run`. Neither exists on `run` on current `main` (already retired by earlier epic steps), so invoking them already errors via "unknown flag"; nothing to remove there. The new value-bearing `--clone <cell>` (P4 / #1092) is the one wired in via the shared FlagSet, as the issue intends.
- **`--env` dual semantics, mapped to the path:** on the existing-cell and `-f` paths `--env` is transport-only runtime injection (`Spec.RuntimeEnv`, #834); on the fused `--from-config`/`--clone` paths it is the persisted per-cell override baked into the materialised `CellDoc` (`Spec.Provenance.EnvOverrides`, #1023) — inherited verbatim from `cell.Materialize`. The override symmetry (`--param` rejected with `--from-config`, `--env` rejected with `--from-blueprint`) is enforced once in `cell.ValidateOverrideSymmetry`, shared by both verbs.
- **No viper double-bind:** `run` reads the shared source flags via `cmd.Flags()` (not viper) so it never collides with `create cell`'s `KUKE_CREATE_CELL_*` bindings on the process-global viper key. `create cell`'s behavior is unchanged (still viper-bound for env-var fallback).
- **`KUKE_RUN_PARAM_FILE` / `KUKE_RUN_IGNORE_DISK_PRESSURE` env-var fallback regression (non-blocking review item).** Both env-var keys were viper-bound on `kuke run` before this PR; the binds are removed now because the underlying flags moved into the shared `cell.RegisterSourceFlags` FlagSet (read via `cmd.Flags()`, not viper). Operators who relied on the env-var defaults will see a silent no-op until they migrate to the CLI flag. The viper-collision rationale (avoiding the global-key race with `kuke create cell --param-file`/`--ignore-disk-pressure`) is sound and called out in `cmd/config/env.go:418–444`. Same migration applies to anyone scripting the old `KUKE_RUN_BLUEPRINT` / `KUKE_RUN_NEW` env vars — those keys are retired alongside the corresponding flags.
- The config-positional probe-walk and `--new`/idempotent-attach behaviors are intentionally gone: the fused path follows create cell's single-scope lookup and create-or-fail-on-collision, and attach-to-existing is now `run <cell>`.

### Docs

Per the second-pass review, this PR refreshes the four files operators hit first against the new surface (epic:cell-identity #1025 retires `-b/--blueprint`, `--new`, the old boolean `--clone`, `--reuse`, and the `<config>`-as-Config positional):

- `docs/cli-use-cases.md` — the UX source-of-truth rewrite for the `kuke run` workflow + invariants block, plus the conflicting-`kuke run` invariants, plus the `kuke run <cell>` references on the restart-time reapply paragraphs.
- `docs/site/cli/kuke-run.md` — the CLI reference rewrite (flag table, source paths, identity model, examples).
- `docs/site/cli/kuke-create.md` — the cross-references to `kuke run -b` / `kuke run <config>` swapped to `kuke run --from-blueprint` / `kuke run --from-config`.
- `docs/examples/claude-code/blueprint.yaml` — the runnable `sudo kuke run -b claude-code` example swapped to `sudo kuke run --from-blueprint claude-code`.

The wider doc sweep across `docs/site/guides/` and `docs/site/manifests/` is deferred to follow-up #1100 per the reviewer's "Update in this PR or commit to a follow-up in the epic plan" disjunction. The migrate-cellprofile-to-blueprint guide in particular needs a structural rewrite (not a mechanical sed) since some prior recipes — `kuke run <cfg> --clone`, `--reuse`, the clone-pool / overlay-preserving driver pattern — have no 1:1 replacement.

## Test plan

- [x] `go build ./...` clean (GOWORK=off, per the repo's `go.work` split).
- [x] `make test-root` — 104 packages, 0 failures (rewrote `cmd/kuke/run` tests for the new surface: positional existing-cell start/attach/not-found, fused `--from-blueprint/--from-config/--clone`, `--name` collision refusal, override symmetry, removed-flag rejection, mutex; `cmd/kuke/create/cell` tests pass unchanged, confirming the materialization refactor is behavior-preserving). Fix-up b22f447 adds `TestRun_RequireSynced_RejectedOutsideFile` (positional / from-blueprint / from-config / clone subtests).
- [x] `make test-kukebuild` — ok.
- [x] `golangci-lint` + `prettier` (via pre-commit on changed files) — pass; `gofmt` clean.
- [ ] `make dev-init` smoke — **could not complete on this host.** The kukeond image build (`make release-build` inside the kukebuild RUN sandbox) fails with `no space left on device` while compiling the binary, reproduced twice including with zero concurrent build activity (host had 83G free disk and 37Gi free RAM; the limit is the build sandbox's ephemeral storage, an infra constraint of this host). This is environmental, not a code issue — the full `go build ./...` and `make test` (test-root + test-kukebuild) are green, and the in-container build was compiling cleanly (38+ packages) before the sandbox filled. This change is also CLI-only: it touches neither the build path, the daemon, nor anything under `/opt/kukeon`, and the dev-init smoke exercises `kuke init` + `kuke attach`, not the `kuke run` / `kuke create cell` surface this PR changes. Recommend a reviewer with more build-sandbox headroom run `make dev-init` to confirm.

Closes #1025


eminwux:
**PR #1099 Review — `feat(run): cell-positional + fused create+start+attach delegation (P8)`**

Refactor lines up cleanly — `cell.Materialize` / `RegisterSourceFlags` factor well, the create-vs-run drift surface is shut, the new test surface tracks the new shape, and the removed-source / new-source / collision branches are pinned. One blocker.

A few small items:

- **Stranded viper read** `cmd/kuke/run/run.go:276`: `parseRunFlags` initializes `ignoreDiskPressure: viper.GetBool(config.KUKE_RUN_IGNORE_DISK_PRESSURE.ViperKey)`, but the viper bind for `--ignore-disk-pressure` was deleted from `run` alongside the old flag definition (the flag is now declared by `cell.RegisterSourceFlags`, which deliberately doesn't viper-bind to avoid colliding with `kuke create cell`'s bind). Result: `kuke run --from-blueprint X --ignore-disk-pressure` silently leaves `Spec.IgnoreDiskPressure` false, and the daemon's disk-pressure guard the operator was trying to bypass rejects the create. Fix: read via `cmd.Flags().GetBool("ignore-disk-pressure")` — same pattern as the other shared flags in the map loop at `run.go:280-292`. The test suite has zero coverage for `--ignore-disk-pressure` on the run side; please add one to gate the next regression.

LGTM once the `--ignore-disk-pressure` read is switched from `viper.GetBool` to `cmd.Flags().GetBool` (and a test pinned).

eminwux:
Addressed in 85b9441:
- `parseRunFlags` now reads `--ignore-disk-pressure` via `cmd.Flags().GetBool("ignore-disk-pressure")` instead of `viper.GetBool(config.KUKE_RUN_IGNORE_DISK_PRESSURE.ViperKey)`. The flag is registered by `cell.RegisterSourceFlags` (shared with `kuke create cell`) and never viper-bound on the run side, so the prior read always saw the zero value and silently dropped the operator's `--ignore-disk-pressure` opt-in past the daemon's CreateCell disk-pressure guard.
- Pinned with two regression tests in `cmd/kuke/run/run_test.go`: `TestRun_FromBlueprint_IgnoreDiskPressure_ThreadsToSpec` (flag set → `Spec.IgnoreDiskPressure == true`) and `TestRun_FromBlueprint_NoIgnoreDiskPressure_LeavesSpecFalse` (flag omitted → zero value).

`make test` (test-root + test-kukebuild) — 0 failures.

eminwux:
**PR #1099 Review — `feat(run): cell-positional + fused create+start+attach delegation (P8)`**

Sharing `cell.RegisterSourceFlags` + `cell.Materialize` across `kuke create cell` and `kuke run` cleanly fulfills AC#4; the `runFused`/`runFromFile`/`runExisting` split reads well, the dual-meaning of `--env` (transport vs persisted-override per path) is plumbed correctly, and the test surface tightly covers the new behavior (positional NotFound/Ready, fused clone, fused name collision, `--name` rejection on positional/file, removed-flag rejection). One blocker on docs drift.

A few items:

- **Doc drift — blocking** `docs/cli-use-cases.md:264-298,520-524`, `docs/site/cli/kuke-run.md:30,60-79,102`, `docs/site/cli/kuke-create.md:66`, `docs/examples/claude-code/blueprint.yaml:10-13`: four files still describe `-b/--blueprint`, `--new`, the old boolean `--clone`, `--reuse`, and the `<config>`-as-Config positional as supported, including the runnable `sudo kuke run -b claude-code` example in the blueprint. Project CLAUDE.md designates `docs/cli-use-cases.md` as the UX source of truth, and earlier epic phases (post-#823/#825/#833/#835/#986) updated it per-PR. Operators (or agents) following the docs hit unknown-flag errors or wrong-semantics behavior. Update in this PR or commit to a follow-up in the epic plan.
- **Silent-no-op UX** `cmd/kuke/run/run.go:157,346-379`: `--require-synced`'s description says \"With -f:\" but `validateSourceFlagCompat` doesn't reject it on the positional/fused paths, and `warnDivergentNamedCell` (the only consumer of `flags.requireSynced`) is only reached from `runFromFile` — silently ignored on the other paths. Reject at parse time or call out the no-op in the description. Non-blocking.
- **Env-var fallback regression** `cmd/config/env.go:425-444`, `cmd/kuke/run/run.go:127`: `KUKE_RUN_PARAM_FILE` and `KUKE_RUN_IGNORE_DISK_PRESSURE` were viper-bound on `kuke run` before this PR; the binds are removed now (the flags moved into the shared FlagSet, read via `cmd.Flags()`). The viper-collision rationale is sound and noted in the env.go comment, but the operator-visible silent no-op for those env vars deserves a callout in the PR body or release notes. Non-blocking.

`make dev-init` smoke deferred per CLAUDE.md host-gated carve-out (CLI-only change, daemon path untouched; build sandbox ran out of space) — fine, reviewer/operator runs it pre-merge.

LGTM once the doc gap is addressed (or a follow-up PR is committed in the epic plan) — the rest are non-blocking.

eminwux:
Addressed in b22f447:

- **Doc drift (blocking).** Refreshed the four files the review cited against the new surface: `docs/cli-use-cases.md` (run sequence + invariants block, conflicting-`kuke run` block, `kuke run <cell>` swap on the restart-time reapply paragraphs), `docs/site/cli/kuke-run.md` (CLI reference rewrite — flag table, source paths, identity model, examples), `docs/site/cli/kuke-create.md` (`kuke run -b` / `kuke run <config>` cross-references swapped to `--from-blueprint` / `--from-config`), `docs/examples/claude-code/blueprint.yaml` (runnable example swapped to `kuke run --from-blueprint claude-code`).
- **Wider doc sweep deferred to #1100.** Eight more files under `docs/site/guides/` and `docs/site/manifests/` still describe the retired surface (`docs/site/guides/migrate-cellprofile-to-blueprint.md` is the heaviest — needs a structural rewrite, not a sed; `--reuse` and the clone-pool driver pattern have no 1:1 replacement). Filed as a follow-up `chore(docs):` issue with the full translation table, per your "Update in this PR or commit to a follow-up in the epic plan" disjunction.
- **Silent-no-op `--require-synced` (non-blocking).** Rejected at parse time in `validateSourceFlagCompat` rather than only mentioned in the description — the flag compares the live cell against a source manifest, which only `-f` supplies, so the positional and the fused paths get `--require-synced is only valid with -f/--file` instead of a silent drop. Pinned in `TestRun_RequireSynced_RejectedOutsideFile` (positional / from-blueprint / from-config / clone subtests).
- **Env-var fallback regression (non-blocking).** Documented in the PR body under "Design notes for the reviewer" — `KUKE_RUN_PARAM_FILE`, `KUKE_RUN_IGNORE_DISK_PRESSURE`, plus the retired `KUKE_RUN_BLUEPRINT` / `KUKE_RUN_NEW` keys, all silent no-op until operators migrate to the CLI flag. The viper-collision rationale stays as the load-bearing reason.

`go build ./...`, `make test-root` (104 packages, 0 failures, new `TestRun_RequireSynced_RejectedOutsideFile` exercised), `make test-kukebuild`, and `pre-commit` (gofmt, golangci-lint, prettier, addlicense) clean. `make dev-init` still deferred per the host build-sandbox carve-out (unchanged from the first push).

eminwux:
**PR #1099 Review — `feat(run): cell-positional + fused create+start+attach delegation (P8)`**

LGTM. Clean extraction of `cell.{SourceFlags, RegisterSourceFlags, Materialize, ValidateOverrideSymmetry, ScopeVars}` as the single shared materialization path between `kuke create cell` and `kuke run`; the parity with the un-fused `create cell` + `start` is exactly what AC#4 asks for. Old `runFromBlueprint/runFromConfig/runClone/loadCellDoc/loadResult/loadFromBlueprint/loadFromConfig/lookupConfigWithFallback/divergentSourcePointer` symbols are fully removed (no orphans); `KUKE_RUN_BLUEPRINT`/`KUKE_RUN_NEW` are gone end-to-end; `KUKE_RUN_PARAM_FILE` is left declared but unbound, matching the documented env-var fallback BC. Per-commit signatures all verified. Sibling-precedent check on `--ignore-disk-pressure` is pinned by `TestRun_FromBlueprint_IgnoreDiskPressure_ThreadsToSpec`. The `cli-use-cases.md`, `kuke-run.md`, `kuke-create.md`, and `claude-code/blueprint.yaml` doc refreshes match the new surface; deferral of the wider doc sweep to follow-up #1100 is reasonable given the structural rewrite the migrate-cellprofile guide needs.

A couple of informational items (none blocking):

- **Behavior change** `cmd/kuke/create/cell/cell.go:386-460` (`Materialize` → `materializeFromConfig`/`materializeFromBlueprint`): the old `kuke run <config>` positional walked full → space-only → realm-only probes via `lookupConfigWithFallback`, so a Config stored at `default/default/default` was findable by bare `kuke run prod` (no `--space`/`--stack`). The new `kuke run --from-config prod` uses `kukeshared.ExplicitScope` directly (returns `""` when the flag isn't `Changed`), so the lookup is now `(realm="default", space="", stack="")` and a Config at `default/default/default` is no longer found without explicit scope flags. This matches existing `kuke create cell --from-config` behavior — the unification is the explicit AC#4 goal — but it's a UX BC vs. the old `run` surface that isn't called out in the PR body. The deleted probe-walk tests (`TestRun_FromConfig_FullScopeProbe_Hits` etc.) pinned the prior behavior.
- **Host-gated smoke deferred per dev/CLAUDE.md carve-out.** `make dev-init` couldn't complete on the author's host (kukebuild sandbox `no space left on device` despite 83G/37Gi free); the PR is CLI-only and touches neither the daemon nor `/opt/kukeon`, so deferred verification on a roomier host is appropriate before merge.

PR #1104, opened 2026-06-06
refactor(create-cell): replace mergeParamMaps with cellblueprint.MergeParams

### LINKED ISSUE #1093: followup-nit(create-cell): replace mergeParamMaps with cellblueprint.MergeParams
From PR #1092.

`cmd/kuke/create/cell/clone.go:309` defines `mergeParamMaps(base, override map[string]string) map[string]string` — a private duplicate of the existing `cellblueprint.MergeParams` at `internal/cellblueprint/params.go:163` (same signature, same CLI-wins semantics). The clone.go body comment on line 307 already names "MergeParams' CLI-wins contract" as the contract being followed; the call should go directly through that helper instead of restating it.

Suggested fix: at `cmd/kuke/create/cell/clone.go:285`, replace `mergedParams := mergeParamMaps(srcProv.Params, cliParams)` with `mergedParams := cellblueprint.MergeParams(srcProv.Params, cliParams)` and delete `mergeParamMaps` (clone.go:303–318).

`buildParamMap` (cell.go:642) already routes through `cellblueprint.MergeParams`, so the cleanup also tightens the in-package precedent.

eminwux:
## Summary
- Drop the private `mergeParamMaps` helper in `cmd/kuke/create/cell/clone.go` (lines 304–320) and route the only call site (line 284) through `cellblueprint.MergeParams` — same signature, same CLI-wins semantics, already named in the line-279 contract comment.
- Aligns with the in-package precedent: `buildParamMap` (`cell.go:651`) already calls `cellblueprint.MergeParams` directly.

## Notes
- `mergeParamMaps` returned `nil` when both inputs were empty; `cellblueprint.MergeParams` always returns a fresh allocated map. The AC#6 byte-equality (no-`--param` clone's provenance unchanged from the source's) is preserved by the existing `if len(mergedParams) > 0` gate at `clone.go:298`, which skips the provenance assignment when the merged map is empty regardless of nil-vs-non-nil. `Resolve`/`MaterializeWithName` accept either shape, so no caller observes the difference.

## Test plan
- [x] `make test-root` (passes; `cmd/kuke/create/cell` tests `ok`)
- [x] `make test-kukebuild` (passes)
- [x] `GOWORK=off go build ./cmd/kuke/create/cell/...` (clean)

Closes #1093

### DIFF
--- a/cmd/kuke/create/cell/clone.go
+++ b/cmd/kuke/create/cell/clone.go
@@ -281,7 +281,7 @@ func cloneFromBlueprint(
 	if err != nil {
 		return v1beta1.CellDoc{}, err
 	}
-	mergedParams := mergeParamMaps(srcProv.Params, cliParams)
+	mergedParams := cellblueprint.MergeParams(srcProv.Params, cliParams)
 
 	resolved, err := cellblueprint.Resolve(bpRes.Blueprint, mergedParams, os.LookupEnv)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #1104 Review — `refactor(create-cell): replace mergeParamMaps with cellblueprint.MergeParams`**

LGTM — pure dedup of a private helper into the existing package-level equivalent, signature/semantics match, and the `len(mergedParams) > 0` gate at `clone.go:298` preserves the AC#6 byte-equality invariant across the nil-vs-empty difference.

Issue #1117, opened 2026-06-06
epic: kuke team init replaces eminwux/crew install.sh (host-state + renderer + rename)


eminwux:
## Context

`kuke team init` (delivered via #792 / #793 / #796 / #1041 / #1042 / #1043) renders the per-project team but cannot consume the agents repo's published blueprint templates. Three layers mismatch:

1. **Catalog vs roles** — `harnesses/images.yaml` doesn't list `claude`/`codex`/`opencode` as capabilities; every role's `needs.image` does. `SelectImage` fails.
2. **Template path resolution** — `harness.yaml` declares `template: blueprint.tmpl.yaml` (sibling of `harness.yaml`); kukeon resolves it against the cache root.
3. **Template body** — blueprints are written in Go `text/template` syntax with `{{ range .needs.repos }}` / `{{ template "..." . }}` / `{{ .operator.X }}` / sprig pipes; kukeon's renderer does flat `${KEY}` envsubst over 9 scalar keys.

Result: every `kuke team init` invocation today errors out at layer 1, and patching layers 1+2 only exposes the bigger layer-3 mismatch. The cache-patch debugging session that surfaced this also revealed that `eminwux/crew/install.sh` still owns host-state lifecycle (`~/.crew/claude-<role>/`, `claude.json` seeds, `secrets.env` → `kind: Secret`, bind-mounted into every cell) that should belong to `kuke team init`.

This epic closes both gaps, retires `eminwux/crew`, and isolates host state per-team so multiple projects on one host don't share bind mounts.

## Decisions locked

- **Host-state root:** per-team subdirs under `~/.kuke/teams/<team>/` (sibling of `~/.kuke/cache/`, `~/.kuke/kuketeam.d/`). Mode 0700. Sub-layout: `<role>-<harness>/`. Example: project `dezot` × role `dev` × harness `claude` → `~/.kuke/teams/dezot/dev-claude/`.
- **Per-team declaration:** `TeamEntrySpec` gains `spec.teamDir` (auto-filled by `kuke team init` to `Layout.TeamDir(team)`, hand-editable for relocation).
- **Secrets:** two layers — shared defaults at `~/.kuke/teams/secrets.env`, per-team overrides at `~/.kuke/teams/<team>/secrets.env`. Per-team value wins per-key.
- **Operator-fact rename:** `CREW_ROOT` → `TEAM_ROOT`. `.operator.TEAM_ROOT` is per-team-context (different teams bind different roots on the same host).
[...]

Issue #1110, opened 2026-06-06
team: RenderBlueprint — text/template engine, relative paths, sibling partials (step 2)


eminwux:
## Context

Part of the **team2** epic (umbrella #1117). The agents repo's blueprint templates (`harnesses/*/blueprint.tmpl.yaml`) are written in Go `text/template` syntax — `{{ range .needs.repos }}`, `{{ template "mount_source" . }}`, `{{ .operator.GIT_USER_NAME }}`, sprig-style `{{ . | upper | replace "-" "_" }}` — and reference `.needs.{repos,mounts,params,secrets}`, `.harnesses.<h>.{settings,sandbox,approval,permissions}`, `.operator.*`, `.project.*` dot-context fields.

Today `internal/teamrender/teamrender.go:520 substitute` only does flat regex `${KEY}` substitution against 9 scalar keys (ROLE, HARNESS, NEEDS, IMAGE, IMAGE_REF, SETTINGS, SANDBOX, APPROVAL, PERMISSIONS). There is no way to iterate `needs.repos` or expand the dot-context, so every published blueprint fails `yaml.Unmarshal` at the first `{{ range }}` block (`yaml: line 44: could not find expected ':'`).

On top of the syntax mismatch, `RenderBlueprint` (`internal/teamrender/teamrender.go:281`) has two path-resolution bugs that block the published layout even once the renderer understands Go templates:

1. **Template path joined against cache root, not the harness dir.** The agents repo's `harnesses/claude/harness.yaml` declares `template: blueprint.tmpl.yaml` (bare filename, sibling of `harness.yaml`). Kukeon does `filepath.Join(cacheDir, harness.Spec.Template)`, looking at the cache root — which doesn't exist.
2. **No `{{ define }}`-able partial set.** Blueprints call `{{ template "mount_source" . }}`, `{{ template "repo_url" . }}` etc., but kukeon parses only the single blueprint file — there is nowhere for those `{{ define }}` blocks to live.

This issue makes `RenderBlueprint` consume published blueprints end-to-end: the template engine, the dot-context, the relative path, and the partial loader land together. (Absorbs the former step 7 — both edits rewrite the same function.)

### 1. text/template engine + typed dot-context

Replace `substitute` with a `text/template.Template` parsed from the harness blueprint, executed against a typed dot-context:

```go
type RenderContext struct {
    Role      RoleView
    Needs     NeedsView                    // .needs.{repos, mounts, params, secrets, image}
[...]

PR #1122, opened 2026-06-07
feat(team): kuke team init --validate one-pass render-contract check

### LINKED ISSUE #1116: team: kuke team init --validate — one-pass contract check (step 8)
## Context

Part of the **team2** epic (umbrella #1117). Without this preflight, every contract drift between the agents repo and kukeon surfaces one error at a time during render: catalog miss, then template-path miss, then partial miss, then unbound fact, then yaml-parse. The cache-patch debugging session that motivated this epic was exactly that pattern.

`kuke team init --validate` runs every check once and prints a single gap report.

## Proposal

A new mode (`--validate`, mutually exclusive with `--build`) that, after `renderTeam`:

1. **Catalog vs roles.** For every `(role × harness)` pair, verify SelectImage finds a match. Collect misses with the merged needs + first unmet capability.
2. **Template paths.** For every selected harness, verify `harness.Spec.Template` resolves (under #1110's relative resolution).
3. **Partials.** Parse every blueprint template. For every `{{ template "name" }}` reference, verify a matching `{{ define "name" }}` exists in the harness's partial set.
4. **Fact references.** For every `{{ .operator.X }}` and `{{ .project.X }}` reference, verify the dot-context exposes the field.

Emit a multi-section report:

```
== catalog ==
ok   dev × claude   → claude
miss pm  × opencode → capability "android" not provided by any opencode image

== templates ==
ok   harnesses/claude/blueprint.tmpl.yaml
miss harnesses/opencode/blueprint.tmpl.yaml — template: blueprint.tmpl.yaml not found

== partials ==
miss harnesses/claude/blueprint.tmpl.yaml uses template "mount_source", no {{ define }} found

== facts ==
miss .operator.UNKNOWN_KEY referenced in harnesses/claude/blueprint.tmpl.yaml, not bound
```

Non-zero exit on any miss.

## Acceptance criteria

- [ ] `--validate` flag wires through `cmd/kuke/team/init.go`.
- [ ] All four sections render; OK and miss lines are stable-ordered.
- [ ] Exit 0 on a clean validate, exit 1 on any miss.
- [ ] Unit tests for each section's miss/ok detection.

## Dependencies

Depends on: #1110, #1115.
Blocks: none.

## Notes

This is the lever that catches future drift. Without it, the next time the agents repo's template contract evolves we'll be repeating the cache-patch debugging dance.

eminwux:
## Summary

- Adds a `--validate` mode to `kuke team init` (step 8 of the **team2** epic, #1117) that runs every render-contract check once and prints a single gap report, instead of surfacing one drift error per re-run during render.
- `--validate` is mutually exclusive with `--build` (cobra `MarkFlagsMutuallyExclusive`); it applies nothing and writes nothing (the operator-global facts scaffold is held in memory only, the drop-in entry is never written).
- New `teamrender.Validate(bundle, pt) *ValidationReport` runs the four checks the issue specifies, reusing the existing render helpers (`SelectImage`/`firstUnmet`, `parseBlueprintTemplate`, `renderContextValues`, `HarnessDir`) so the validator can never drift from the renderer:
  1. **catalog** — every `(role × harness)` pair run through `SelectImage`; ok lines name the selected catalog ref, miss lines name the first unmet capability (or the no-single-image case).
  2. **templates** — every selected harness's `spec.template` resolved relative to the harness dir (per #1110).
  3. **partials** — each blueprint template set parsed; every `{{ template "name" }}` invocation checked against the set's `{{ define }}` blocks (parse-tree walk, not regex).
  4. **facts** — every `{{ .operator.X }}` / `{{ .project.X }}` reference checked against the dot-context key schema. The known-key set is **derived from `renderContextValues` itself** (saturated synthetic inputs), so a future renderer key add/remove can't silently diverge from the validator; `TestKnownFactKeysCoverRendererContract` pins it.
- Non-zero exit on any gap via new `errdefs.ErrTeamValidateGaps`; clean validate exits 0. Every section header always renders (a clean section is just its header).

- **Separate `validateTeam` dispatch rather than threading a `validate bool` through `composeTeam`.** `--validate` is a thin read-only path (resolve bundle → check → report); it needs neither the apply/provision/secrets/entry-write lifecycle nor a third bool across `composeTeam`'s ~20 test call sites. The flag is parsed in `init.go` and dispatches — the AC's "wires through `cmd/kuke/team/init.go`" is satisfied.
- **Facts are checked against the key *schema*, not an operator's populated subset.** A blueprint referencing a known-but-unset fact (e.g. `REGISTRY` when no registry is configured) renders empty at run time, not an error — so flagging it would be a false positive. `Validate` therefore needs neither `TeamsConfig` values nor per-run `Inputs`.
- **partials/facts emit miss-only lines** (a resolved reference is unremarkable and would flood the report); catalog/templates emit ok+miss. This matches the issue's example output. All sections are stable-sorted by line body so output is byte-identical across runs (`TestValidateDeterministicOutput`).
- Report line format: `== <section> ==` headers, fixed-width `ok`/`miss` status column, blank line between sections — matches the issue's example shape.

## Test plan

- [x] `make test-root` (full root CI target) — pass
- [x] `make test-kukebuild` — pass
- [x] `go vet` on touched packages — clean
- [x] `pre-commit run --files <changed>` (golangci-lint, go fmt, addlicense, …) — pass
- [x] New unit tests in `internal/teamrender/validate_test.go` cover each section's ok/miss detection, the known-fact-key schema, harness-less roster, report formatting, and deterministic ordering.
- [x] New cmd-level tests in `cmd/kuke/team/init_test.go` cover clean-exit-0 + read-only (no global config / drop-in written), gap-exit-non-zero, harness-less-roster-skips-resolve, and the `--build`/`--validate` mutual-exclusion guard.
- [x] Built `kuke` and confirmed `--validate` appears in `kuke team init --help` and `--build --validate` is rejected at the cobra layer with exit 1.
- [ ] `make dev-init` smoke — not run: this change touches neither the build path, the daemon, nor anything under `/opt/kukeon` (CLI render-contract logic only), so the smoke's trigger condition does not apply.

Closes #1116

### DIFF
--- a/cmd/kuke/team/init.go
+++ b/cmd/kuke/team/init.go
@@ -134,7 +134,11 @@ func NewInitCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			return runInit(cmd, dryRun, build)
+			validate, err := cmd.Flags().GetBool("validate")
+			if err != nil {
+				return err
+			}
+			return runInit(cmd, dryRun, build, validate)
 		},
 	}
 
@@ -146,6 +150,13 @@ func NewInitCmd() *cobra.Command {
 		"build", false,
 		"Locally build the selected catalog images (kukebuild) into the target realm's containerd namespace; no registry push",
 	)
+	cmd.Flags().Bool(
+		"validate", false,
+		"Run every render-contract check (catalog, templates, partials, facts) once and print a single gap report; exit 1 on any gap. Applies nothing and writes nothing",
+	)
+	// --validate is a terminal contract-check mode; --build drives a local
+	// image build. They are mutually exclusive per #1116.
+	cmd.MarkFlagsMutuallyExclusive("build", "validate")
 
 	return cmd
 }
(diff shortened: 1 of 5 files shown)


eminwux:
**PR #1122 Review — `feat(team): kuke team init --validate one-pass render-contract check`**

LGTM — `--validate` is a read-only dispatch off `runInit`, the four checks reuse `SelectImage`/`firstUnmet`/`parseBlueprintTemplate`/`renderContextValues`/`HarnessDir` so the validator can't drift from the renderer, the fact-key schema is derived from `renderContextValues` itself (pinned by `TestKnownFactKeysCoverRendererContract`), `loadOrScaffoldGlobalConfig` is invoked with `dryRun=true` so the operator-global config is held in memory only, and the stable-sort + `TestValidateDeterministicOutput` lock the byte-identical output the AC requires.

PR #1125, opened 2026-06-07
fix(team): validate ignores commented template actions; init exits non-zero on apply failure

### LINKED ISSUE #1123: kuke team: --validate false-positives on template actions in YAML comments; init exits 0 when all objects fail to apply
Two defects in the `kuke team init` diagnostic/exit surface, surfaced while debugging a team whose blueprints failed to apply. Filed together as they are the same area (team render-contract diagnostics not behaving as a gate). The underlying *render* bug was in the agents repo (`${...}` vs `{{ }}`) and is fixed there — these two are kukeon-side.

---

## 1. `--validate` parses `{{ ... }}` inside YAML `#` comments as live field references (false positive)

`kuke team init --validate` exits 1 with:

```
== facts ==
miss .operator.X referenced in harnesses/claude/blueprint.tmpl.yaml, not bound
miss .project.X  referenced in harnesses/claude/blueprint.tmpl.yaml, not bound
Error: team validate found one or more contract gaps (2)
```

There is no real gap. `.operator.X` / `.project.X` come from the template's **documentation header**, e.g.:

```yaml
#   - {{ .operator.X }} pulls from TeamsConfig (REGISTRY, HOME_DIR, ...).
#   - {{ .project.X }}  pulls from ProjectTeam (AGENTS_REPO, PROJECT_DIR) ...
```

`X` is human shorthand for "any key". But Go's `text/template` parser does **not** honor YAML `#` comments — it parses `{{ .operator.X }}` as a live `FieldNode`. `validateFacts` (`internal/teamrender/validate.go:360`) walks every `FieldNode` via `walkNodes` (`validate.go:393`), finds field `X` isn't in `knownFactKeys()`, and reports it (`validate.go:420`). So any template that *documents* `{{ .operator.* }}` usage in a comment fails `--validate`.

**Impact:** `--validate` is unusable as a gate against real templates that document their fact contract in comments. It also can't be a useful CI check while this fires.

**Fix direction:** the fact walk should ignore template actions originating in YAML comment lines (or, more generally, the contract walk should operate on the YAML structure rather than the raw `{{ }}` token stream). Note `validate.go:327` runs the same `walkNodes` for the partials/templates sections, so consider the fix holistically.

## 2. `kuke team init` exits 0 even when every applied object fails

Observed (before the agents-side render fix):

```
  CellBlueprint "${ROLE}-claude": failed (blueprint scope does not exist: space "${SPACE}")
  ... (all 6 blueprints + 6 configs failed) ...
applied 0 secret/6 blueprint/6 config object(s) to kukeond under team "kukeon"
$ echo $?
0
```

`applyTeam` (`cmd/kuke/team/init.go:333`) returns a non-nil `err` only when the apply RPC itself fails; per-document failures land in `applyResult` and leave `err == nil`. So `init` proceeds to write the team drop-in entry (`teamhost.WriteEntry`), prints the summary, and returns nil → **exit 0**, despite applying 0 of 12 objects. A scripted/CI `kuke team init` cannot detect the total failure from the exit code, and the drop-in entry is persisted for a team that has nothing applied.

**Fix direction:** treat an apply result with one or more failed documents as a non-zero exit (and reconsider whether the team entry should be written when 0 objects applied).

---

### Repro
1. A `kuketeam.yaml` pointing at an agents source whose blueprint templates declare scope via unresolved placeholders (pre-fix), or any template that documents `{{ .operator.* }}` in a `#` comment.
2. `kuke team init --validate` → exit 1 with the spurious `.operator.X`/`.project.X` misses (defect 1).
3. `kuke team init` against a scope-broken render → exit 0 despite `applied 0/N` (defect 2).

eminwux:
## Summary
- **Defect 1 — `--validate` false positives on `{{ ... }}` inside YAML `#` comments.** Go's `text/template` parser does not honor YAML `#` comments, so a blueprint that documents its fact contract in a comment (e.g. `#   - {{ .operator.X }} pulls from TeamsConfig`) parsed `{{ .operator.X }}` as a live `FieldNode` and `validateFacts` reported `.operator.X`/`.project.X` as unbound — making `--validate` unusable as a gate/CI check against real templates. Fixed holistically at the validate parse boundary: the validate path now strips YAML comment tails before parsing, so both `validateFacts` and `validatePartials` (which share `walkNodes`) stop flagging comment-documented actions.
- **Defect 2 — `kuke team init` exited 0 when every applied document failed.** Per-document failures land as `failed` actions in `applyResult`, not as a non-nil error, so `init` wrote the drop-in entry, printed the summary, and returned nil → exit 0 despite `applied 0/N`. Now `composeTeam` counts `failed` documents, exits non-zero (`errdefs.ErrTeamApplyFailed`) on any failure, and skips the entry write entirely on a *total* failure (every document failed, nothing landed).

## Implementation notes
- **Render path is untouched.** `parseBlueprintTemplate` is shared with `RenderBlueprint`; stripping comments there would alter the generated YAML. Added `parseBlueprintTemplateBody(tplPath, raw, transform)` — `nil` transform (identity) for render, `stripYAMLComments` for validate only.
- **Comment detection is quote-aware** (`yamlCommentIndex`): a `#` begins a comment only at line start or after whitespace and outside single/double-quoted scalars, so a `{{ ... }}` inside a quoted value containing `#` survives rather than being clipped. Line structure is preserved (newlines kept, only the commented tail blanked) so parser line numbers don't shift.
- **Gate-vs-flip / blast-radius call on defect 2:** the issue's fix direction asked to "reconsider whether the team entry should be written when 0 objects applied." I chose: *total* failure (0 applied) skips the entry write; a *partial* failure (some applied, some failed) still writes the entry (the team partially exists) but exits non-zero. Reviewer can push back if partial failures should also skip the write.
- New sentinel `errdefs.ErrTeamApplyFailed` mirrors the existing `ErrTeamValidateGaps` pattern (and `kuke apply -f`'s `ErrConfig`-driven non-zero exit on `failed` resources).
- Left `emitApplySummary`'s "applied N …" wording unchanged (it prints the per-object `failed` lines above the aggregate) — rewording it is out of scope for this AC.

## Test plan
- [x] `make test` (full CI suite, `GOWORK=off`) — exit 0
- [x] `GOWORK=off go vet ./internal/teamrender/... ./cmd/kuke/team/... ./internal/errdefs/...` — clean
- [x] New `internal/teamrender` tests: `TestYAMLCommentIndex` (quote-aware detection table), `TestValidateFactsIgnoresCommentedReferences` (commented `.operator.X`/`.project.X` not flagged; live gap still surfaces), `TestValidatePartialsIgnoresCommentedTemplateInvocation`
- [x] New `cmd/kuke/team` tests: `TestComposeTeamApplyTotalFailureExitsNonZeroSkipsEntry`, `TestComposeTeamApplyPartialFailureExitsNonZeroWritesEntry`

Note: the repo-wide `go build ./...` fails on a pre-existing transitive dep break (`containerd/cgroups/v2` under the merged `go.work` graph) unrelated to this change; building/testing with `GOWORK=off` (as the Makefile and CI do) is clean.

Closes #1123

### DIFF
--- a/cmd/kuke/team/init.go
+++ b/cmd/kuke/team/init.go
@@ -335,28 +335,58 @@ func composeTeam(
 		return err
 	}
 
-	source := pt.Spec.Source
-	entry := &model.TeamEntry{
-		APIVersion: model.APIVersionV1,
-		Kind:       model.KindTeamEntry,
-		Metadata:   model.Metadata{Name: project},
-		Spec: model.TeamEntrySpec{
-			Path:    projectDir,
-			TeamDir: teamDir,
-			Source:  &source,
-		},
+	// Per-document apply failures do not surface as a non-nil err — apply only
+	// errors when the RPC itself fails; a blueprint whose scope does not exist
+	// lands as a `failed` action in the result instead. Count them so a
+	// scripted/CI `kuke team init` exits non-zero on any failure, and so a
+	// *total* failure (every applied document failed, nothing landed) skips the
+	// drop-in entry write — persisting an entry for a team with zero applied
+	// objects misleads later `kuke team` reads (#1123). A partial failure still
+	// writes the entry (the team partially exists) but exits non-zero.
+	failed := failedResourceCount(applyResult)
+	totalFailure := applied && failed > 0 && failed == len(applyResult.Resources)
+
+	if !totalFailure {
+		source := pt.Spec.Source
+		entry := &model.TeamEntry{
+			APIVersion: model.APIVersionV1,
+			Kind:       model.KindTeamEntry,
+			Metadata:   model.Metadata{Name: project},
+			Spec: model.TeamEntrySpec{
+				Path:    projectDir,
+				TeamDir: teamDir,
+				Source:  &source,
+			},
+		}
+		if writeErr := teamhost.WriteEntry(layout, entry); writeErr != nil {
+			return fmt.Errorf("write team entry: %w", writeErr)
+		}
+		fmt.Fprintf(out, "wrote team %q to %s\n", project, layout.EntryPath(project))
 	}
-	if writeErr := teamhost.WriteEntry(layout, entry); writeErr != nil {
-		return fmt.Errorf("write team entry: %w", writeErr)
-	}
-	fmt.Fprintf(out, "wrote team %q to %s\n", project, layout.EntryPath(project))
 	if applied {
 		emitApplySummary(out, project, applyResult, len(res.Secrets), len(res.Blueprints), len(res.Configs))
 	} else {
 		fmt.Fprintf(out, "rendered %d secret/%d blueprint/%d config object(s) (no apply: no (role × harness) pairs)\n",
 			len(res.Secrets), len(res.Blueprints), len(res.Configs))
 	}
+	if failed > 0 {
+		return fmt.Errorf("%w (%d of %d documents failed)", errdefs.ErrTeamApplyFailed, failed, len(applyResult.Resources))
+	}
 	return nil
+}
+
+// failedResourceCount returns the number of apply documents that came back with
+// a `failed` action. applyTeam's returned error covers transport/RPC failures;
+// a per-document failure (an unresolvable blueprint scope, say) lands here
+// instead, so `kuke team init` inspects the result to detect it.
+func failedResourceCount(result kukeonv1.ApplyDocumentsResult) int {
+	n := 0
+	for _, r := range result.Resources {
+		if r.Action == "failed" {
+			n++
+		}
+	}
+	return n
 }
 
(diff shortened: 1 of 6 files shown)


eminwux:
**PR #1125 Review — `fix(team): validate ignores commented template actions; init exits non-zero on apply failure`**

LGTM — both defects are fixed cleanly: validate strips YAML `#`-comment tails only on the validate parse path (render YAML untouched), and `composeTeam` now counts `failed` apply docs to exit non-zero, skipping the entry write only on *total* failure. Quote-aware `yamlCommentIndex` (incl. doubled-`'` parity and `\"` escapes), correct `failed > 0` short-circuit on empty results, all new symbols used, no leaks, and dedicated tests for each path. CI green.

PR #1126, opened 2026-06-07
fix(cellconfig): tolerate undeclared spec.values keys via lenient ResolveConfig

### LINKED ISSUE #1124: kuke team: rendered CellConfigs are un-instantiable — spec.values operator facts rejected by blueprint param resolver
`kuke team init` reports success and applies its CellConfigs, but **every config it produces cannot create a cell**. The artifact a successful team init emits is unusable.

## Reproduction

```
$ kuke create cell diag-dev-claude --from-config dev-claude --realm default
Error: config "dev-claude": resolve blueprint "dev-claude":
  blueprint "dev-claude": --param "GIT_ALLOWED_SIGNERS" is not declared in spec.parameters[]: blueprint is invalid
```

`kuke run --from-config <any team config>` dead-ends at the same wall.

## Root cause — contract mismatch on the `CellConfig.Values` channel

**Producer** — `teamrender.operatorValues` (`internal/teamrender/teamrender.go:837`) populates `spec.values` with operator facts. The rendered `dev-claude` config carries:

```yaml
values:
  ROLE: dev        HARNESS: claude        PROJECT: kukeon
  GIT_AUTHOR_NAME / GIT_AUTHOR_EMAIL / GIT_COMMITTER_NAME / GIT_COMMITTER_EMAIL
  GIT_SIGNING_KEY / GIT_ALLOWED_SIGNERS
```

**Consumer** — `cellconfig.MaterializeWithName` (`internal/cellconfig/materialize.go:65`) feeds those values straight into `cellblueprint.Resolve(bp, cfg.Spec.Values, nil)`, and `resolveValues` (`internal/cellblueprint/cellblueprint.go`) hard-errors on **any** value key not declared in the blueprint's `spec.parameters[]`:

```go
for k := range cliParams {
    if _, ok := declared[k]; !ok {
        return nil, fmt.Errorf("blueprint %q: --param %q is not declared in spec.parameters[]: %w", ...)
    }
}
```

That strictness is correct for catching typos on the interactive `kuke create cell --param FOO=…` path, but it is wrong on the **machine-generated `CellConfig.Values`** channel. The blueprint declares only `{PROMPT, AGENT_NAME, ANTHROPIC_MODEL, PROJECT_DIR, OPENROUTER_MODEL}` — none of `ROLE/HARNESS/PROJECT/GIT_*` — so the first undeclared key aborts materialization.

**The rejected values are mostly dead.** The blueprint already bakes git identity at render time (stored blueprint):

```yaml
git: { author: { name: Emiliano Spinella, email: emilianofs@gmail.com }, … }
```

So the six `GIT_*` values (plus `HARNESS`, `PROJECT`) are never consumed — yet they are exactly what trips the rejection. Only `ROLE` is actually used (the blueprint body keeps `${ROLE}` ×6), and it too is undeclared.

## Second wall behind the first

Even with the undeclared-key rejection fixed, this config still can't materialize: the role's `needs.params` (`PROMPT, AGENT_NAME, PROJECT_DIR, ANTHROPIC_MODEL, OPENROUTER_MODEL`) are emitted by the agents template as `required: true`, but `spec.values` supplies none of them and the Config path passes `lookupEnv=nil` — so `resolveValues` then fails with "required parameter PROMPT not set". (The agents-side template duplication that makes these required is filed separately in the agents repo.)

## Fix direction (kukeon)

The `CellConfig` → cell path should not apply CLI-`--param` typo-strictness to `spec.values`. Options:

1. Make `cellblueprint.Resolve` tolerate (ignore) undeclared keys on the Config channel — e.g. a lenient mode / separate entrypoint — substituting the declared subset and dropping extras. Keep the strict check for interactive `--param`.
2. Have `teamrender` only emit `spec.values` for keys the blueprint actually declares as parameters.

Given git identity is baked at render time, `operatorValues` emitting `GIT_*`/`HARNESS`/`PROJECT` at all appears to be dead output worth pruning regardless.

## Severity

This is distinct from #1123 (validator/exit-code diagnostics) — that issue is about *reporting*; this one means the core artifact a successful `kuke team init` produces cannot instantiate a single cell. Marking priority:A.

eminwux:
## Summary

- `kuke team init` emits `CellConfig`s that **cannot create a cell**: every team-rendered config dead-ends at `--param "GIT_ALLOWED_SIGNERS" is not declared in spec.parameters[]: blueprint is invalid`. The artifact a successful team init produces is unusable (#1124).
- Root cause is a contract mismatch on the `CellConfig.Values` channel: `cellconfig.MaterializeWithName` fed `cfg.Spec.Values` straight into `cellblueprint.Resolve`, whose `resolveValues` hard-errors on **any** value key not declared in the blueprint's `spec.parameters[]`. That typo-strictness is correct for a human typing `kuke create cell --param FOO=…`, but wrong for a **machine-generated** config that carries operator facts (`ROLE`, `GIT_*`, `HARNESS`, `PROJECT`) the blueprint never declares as parameters.
- Fix (issue's **Option 1**, the consumer-side fix): add a lenient entrypoint `cellblueprint.ResolveConfig` that does not reject undeclared keys, and route the `CellConfig` → cell path through it. The interactive `--param` path keeps strict `Resolve` unchanged.

## Why Option 1 over Option 2

Option 1 is the **necessary and sufficient** fix and the lower-blast-radius default: it fixes un-instantiability for *every* machine-generated config, not just teamrender's current output. Option 2 (have `teamrender` emit only declared keys) is **insufficient alone** — the blueprint body references `${ROLE}` ×6 yet does not declare `ROLE` as a parameter, so a pure-prune approach would still trip the rejection (or, if it dropped `ROLE`, leave `${ROLE}` literal). `ResolveConfig` therefore *also substitutes* undeclared keys (not just tolerates them), so a `${ROLE}` the body references but never declares resolves correctly rather than surviving literal.

The issue notes `operatorValues` emitting `GIT_*`/`HARNESS`/`PROJECT` is "dead output worth pruning regardless." That producer-side cleanup is deliberately **not** bundled here (it does not affect correctness once the channel is lenient, and bundling a producer prune with a consumer contract fix is scope creep). It can be filed as a follow-up if desired.

## Scope note — the "second wall" is agents-repo

Per the issue, even with undeclared-key rejection fixed, the team-rendered `dev-claude` config still cannot fully materialize because the role's `needs.params` (`PROMPT`, `AGENT_NAME`, …) are emitted by the **agents** template as `required: true` while `spec.values` supplies none. That is the agents-repo template duplication filed separately. This PR scopes to the kukeon contract bug (the first wall) and **keeps** required-parameter enforcement on the lenient channel — `TestResolveConfig_StillEnforcesRequired` pins that the lenient mode only relaxes undeclared-key rejection.

## Test plan

- [x] `make test-root` (full root test suite, `GOWORK=off`) — exit 0
- [x] `go test ./internal/cellblueprint/... ./internal/cellconfig/...` — pass
- [x] `go vet ./internal/cellblueprint/... ./internal/cellconfig/...` — clean
- [x] New `cellblueprint` tests: `TestResolveConfig_ToleratesUndeclaredKeys`, `TestResolveConfig_SubstitutesUndeclaredBodyRef`, `TestResolveConfig_StillEnforcesRequired`
- [x] New `cellconfig` end-to-end test: `TestMaterialize_ToleratesUndeclaredValues` (a config carrying `ROLE`/`GIT_*`/`HARNESS`/`PROJECT` now materializes; `${TAG}` and undeclared `${ROLE}` both substitute)
- [x] Existing `TestResolve_UndeclaredParamErrors` unchanged — strict `--param` path still rejects undeclared keys

Note: `go build ./...` without `GOWORK=off` surfaces an unrelated vendored `containerd/cgroups` type mismatch (go.work workspace mismatch, per project CLAUDE.md); the project's `GOWORK=off` build/test env (what CI runs) is clean.

Closes #1124

### DIFF
--- a/internal/cellblueprint/cellblueprint.go
+++ b/internal/cellblueprint/cellblueprint.go
@@ -55,7 +55,41 @@ func Resolve(
 	cliParams map[string]string,
 	lookupEnv func(string) (string, bool),
 ) (v1beta1.CellBlueprintDoc, error) {
-	values, err := resolveValues(doc, cliParams, lookupEnv)
+	return resolve(doc, cliParams, lookupEnv, false)
+}
+
+// ResolveConfig is the lenient resolution entrypoint for the machine-generated
+// CellConfig.Values channel (issue #1124). It behaves exactly like Resolve —
+// same cliParams[k] > parameters[k].default > lookupEnv(k) order, same
+// required-parameter enforcement — except it does *not* reject value keys that
+// are absent from spec.parameters[]:
+//
+//   - undeclared keys do not error (the typo-strictness Resolve applies is
+//     correct for a human typing `--param FOO=…`, but wrong for a generated
+//     artifact: a rendered CellConfig carries operator facts — ROLE, GIT_*,
+//     HARNESS, PROJECT — that the blueprint never declares as parameters);
+//   - undeclared keys are still substituted into the body, so a `${ROLE}` the
+//     blueprint body references but never declares as a parameter resolves to
+//     the supplied value rather than surviving as a literal.
+//
+// A declared parameter the values map does not supply still follows the strict
+// rule (default → env → required-error), so a genuinely missing required
+// parameter is still caught.
+func ResolveConfig(
+	doc v1beta1.CellBlueprintDoc,
+	values map[string]string,
+	lookupEnv func(string) (string, bool),
+) (v1beta1.CellBlueprintDoc, error) {
+	return resolve(doc, values, lookupEnv, true)
+}
+
+func resolve(
+	doc v1beta1.CellBlueprintDoc,
+	cliParams map[string]string,
+	lookupEnv func(string) (string, bool),
+	lenient bool,
+) (v1beta1.CellBlueprintDoc, error) {
+	values, err := resolveValues(doc, cliParams, lookupEnv, lenient)
 	if err != nil {
 		return v1beta1.CellBlueprintDoc{}, err
 	}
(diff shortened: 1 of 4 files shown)


eminwux:
**PR #1126 Review — `fix(cellconfig): tolerate undeclared spec.values keys via lenient ResolveConfig`**

LGTM — narrow contract fix on the CellConfig.Values channel via a new lenient `cellblueprint.ResolveConfig`; the interactive `--param` path keeps strict `Resolve`, required-parameter enforcement is preserved on both channels, and tests cover undeclared-key tolerance, undeclared body-ref substitution, required-still-enforced, and the e2e Materialize path.

PR #1131, opened 2026-06-07
fix(team): project-scope rendered blueprint/config names

### LINKED ISSUE #1129: bug(team): rendered cell/config names are not project-scoped (pm-claude, not kukeon-pm-claude) — collision risk across projects
## Summary

`kuke team init` creates cell blueprints/configs (and therefore cells) with names like `pm-claude`, `dev-claude` — not project-scoped. They should be `<project>-<role>-<harness>`, e.g. `kukeon-pm-claude`.

## Root cause

The name comes verbatim from the blueprint template and is never project-qualified:

- Template: `metadata.name: {{ .role.name }}-claude` → `pm-claude`.
- `RenderBlueprint` keeps that name as-is (`internal/teamrender/teamrender.go:377-379`), forcing only realm/labels.
- `BindConfig` copies it verbatim: `Name: bp.Metadata.Name` (`teamrender.go:414`) and into the blueprint ref (`:424`), so the CellConfig inherits the unscoped name.
- The project name is **already available** as a render fact — `projectView["NAME"] = project` (`teamrender.go:749-751`; `project` is `metadata.name` from `kuketeam.yaml`) — it is simply unused for naming.

## Consequence beyond cosmetics

Team objects land at `DefaultRealm = "default"` (`teamrender.go:80`). Two projects served by the same agents source would **collide** on identical `pm-claude` names within one realm.

## Proposed fix (kukeon-side; templates stay project-agnostic)

In `teamrender.RenderBlueprint`, force a `<project>-` prefix on:
- `bp.Metadata.Name`, and
- `bp.Spec.Prefix` (the cell-name root used at create, `pkg/api/model/v1beta1/cellblueprint.go:66`)

at the realm-forcing site. `BindConfig` then follows automatically via `bp.Metadata.Name`.

eminwux:
## Summary

- `kuke team init` rendered cell blueprints/configs with template-literal names like `pm-claude` / `dev-claude`. Two projects served by the same agents source then collided in the shared `default` realm.
- `teamrender.RenderBlueprint` now forces a `<project>-` prefix on `bp.Metadata.Name` and on `bp.Spec.Prefix` at the realm-forcing site (per the issue's "Proposed fix"). `BindConfig` follows automatically because it copies `bp.Metadata.Name` into both the CellConfig metadata and its Blueprint ref.
- `scopeToProject` is idempotent: a future project-aware template that already produces `kukeon-pm-claude` is not double-prefixed.

## Test plan

- [x] `make test` (full project CI target, GOWORK=off applied by Makefile)
- [x] `go vet ./...` (with `GOWORK=off`)
- [x] New unit test `TestRenderBlueprintScopesNameAndPrefixToProject` pins both behaviors (project-scoped `Metadata.Name` + `Spec.Prefix`, and idempotence)
- [x] Updated `TestRenderBlueprintSubstitutesAndStampsTeamLabel` to assert the project-scoped name (`sbsh-dev-claude` instead of `dev-claude`)

Closes #1129

### DIFF
--- a/internal/teamrender/teamrender.go
+++ b/internal/teamrender/teamrender.go
@@ -377,12 +377,35 @@ func RenderBlueprint(
 	if strings.TrimSpace(bp.Metadata.Name) == "" {
 		bp.Metadata.Name = defaultObjectName(roleRef, harness)
 	}
+	bp.Metadata.Name = scopeToProject(bp.Metadata.Name, project)
+	if p := strings.TrimSpace(bp.Spec.Prefix); p != "" {
+		bp.Spec.Prefix = scopeToProject(p, project)
+	}
 	bp.Metadata.Realm = realm
 	if bp.Metadata.Labels == nil {
 		bp.Metadata.Labels = map[string]string{}
 	}
 	bp.Metadata.Labels[v1beta1.LabelTeam] = project
 	return &bp, nil
+}
+
+// scopeToProject prefixes name with `<project>-` so the rendered
+// Blueprint/Config identity (and the cell-name prefix it seeds via
+// cellblueprint.Prefix) is project-scoped. Two projects sharing one
+// agents source then resolve to distinct `<project>-pm-claude`,
+// `<project>-dev-claude` records within the same realm instead of
+// colliding on the template-supplied `pm-claude` / `dev-claude`. Idempotent
+// on names already carrying the prefix so a future project-aware template
+// won't double up.
+func scopeToProject(name, project string) string {
+	project = strings.TrimSpace(project)
+	if project == "" {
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1131 Review — `fix(team): project-scope rendered blueprint/config names`**

LGTM — forces a `<project>-` prefix on the rendered blueprint `Metadata.Name`/`Spec.Prefix` at the realm-forcing site (idempotent), which `BindConfig` propagates verbatim into both the config metadata and its blueprint ref, resolving the cross-project collision in the shared realm; new unit test pins name + prefix scoping + idempotence, CI green.

PR #1132, opened 2026-06-07
feat(create): kuke create registry-credential — upsert realm pull credentials

### LINKED ISSUE #1130: cli: kuke create registry-credential — add pull creds to a realm
## Context

A realm pulling from a private registry fails with `403 Forbidden` (e.g. `ghcr.io/eminwux/...`) because there is no imperative way to put pull credentials onto an existing realm. The auth machinery is already complete end-to-end — `realm.spec.registryCredentials` → `ctr.ConvertRealmCredentials` → `buildResolver` (Docker authorizer matched by host; anonymous when empty) at every pull site (`internal/ctr/registry.go:60-103`), and the realm reconciler already converges the field as a *compatible* change with no recreate (`internal/controller/runner/update_realm.go:39`, `internal/controller/apply/diff.go:115` / `registryCredentialsEqual:1190`). The v1beta1 realm doc already accepts the field (`pkg/api/model/v1beta1/realm.go:38-52`: `[]RegistryCredentials{Username, Password, ServerAddress}`).

**What's missing is a declaration path.** Today the only way to set credentials is `kuke apply -f` with a hand-edited YAML manifest that embeds the plaintext token in a file. There is no imperative flag and no stdin path — `kuke create realm` takes only `--namespace` (`cmd/kuke/create/realm/realm.go:32-80`), and no `set`/`edit` verbs exist. Operators need a first-class CLI command to attach pull credentials to a realm without round-tripping through YAML or leaking the token into shell history.

This supersedes the originally-proposed team-specific approach (a `TeamsConfig` schema field + `secrets.env` read baked into `kuke team init`). A general realm-credential command is the right layer: `kuke team init` (and any operator) points it at the `default` realm to resolve the team 403, instead of bespoke team plumbing.

## Proposal

Add `kuke create registry-credential <realm>` — a new noun under `create`, modeled on the `kuke create secret --from-literal/--from-file` precedent (`cmd/kuke/create/secret/secret.go:42-62`):

```
kuke create registry-credential <realm> \
    --server ghcr.io \
    --username <user> \
    --password-stdin            # read token from stdin (docker login style)
```

- **Mutates an existing realm.** Resolves `<realm>`, upserts an entry into `spec.registryCredentials` keyed by `--server` (serverAddress), and re-applies the realm. The existing reconciler converges it as a compatible change — no realm recreate, no cell disruption.
  - Upsert semantics: an existing entry whose `serverAddress` matches `--server` is replaced (username + password updated); otherwise a new entry is appended.
- **Token never enters argv.** `--password-stdin` reads the token from stdin; `--from-file <path>` reads it from a file (mirrors `create secret`). Exactly one of the two is required. No `--password` plaintext flag.
- `--server` is optional and defaults consistently with the field's existing semantics (empty `serverAddress` matches by image reference); document the default in `--help`.

## Acceptance criteria

- [ ] `kuke create registry-credential <realm> --server <host> --username <u> --password-stdin` reads the token from stdin and upserts a `RegistryCredentials` entry on the named realm's `spec.registryCredentials`.
- [ ] `--from-file <path>` is accepted as an alternative token source; exactly one of `--password-stdin` / `--from-file` is required (error otherwise, matching `create secret`'s "at least one of …" guard).
- [ ] The token is never passed as an argv flag and never logged.
- [ ] Re-running with the same `--server` updates the existing entry in place (upsert), rather than appending a duplicate.
- [ ] After the command, a cell in that realm successfully pulls a private image that previously returned `403` (the reconcile path converges the new credential without a realm recreate).
- [ ] Unknown realm → non-zero exit with a clear "realm not found" error.
- [ ] `--help` documents the command, both token-source flags, and the `--server` default.

## Suggested slicing

Lands in one PR: the new subcommand wires into the already-existing apply/reconcile path, so there is no daemon-side work. If review flags it as too broad, the natural seam is (1) the command + stdin/from-file plumbing + upsert, then (2) the team-init call-site that points it at the `default` realm — but (2) is better tracked as a separate team2 follow-up (see Notes) rather than pre-filed here.

## Notes

- **Unblocks the team 403.** This issue was originally filed under `epic:team2` as a team-init-specific credential path; it is now a standalone `cli` enhancement. The team flow consumes it by running this command against the `default` realm where team cells pull. A thin team-init wiring follow-up should be filed under `epic:team2` once this lands — cross-link both ways.
- The two-layer `secrets.env` → `kind: Secret` apply pipeline (#1120, merged) remains the pattern for *application* secrets; registry pull credentials are a distinct realm-spec field and intentionally do not route through the Secret object.
- Credential storage is unchanged — `RegistryCredentials.Password` is stored as today. This issue only adds the declaration path, not at-rest encryption.

eminwux:
## Summary

- Adds `kuke create registry-credential <realm>` — a first-class imperative path to attach private-registry pull credentials to an existing realm, removing the need to hand-edit a YAML manifest (and leak the token into a file) just to set `spec.registryCredentials`.
- The command reads the realm (`GetRealm`), upserts an entry keyed by `--server`, and re-applies the full realm doc through the daemon's write-through `ApplyDocuments` path. The pre-existing realm reconciler converges `registryCredentials` as a *compatible* change (`internal/controller/apply/diff.go:115` → `update_realm.go`), so there is no realm recreate and no daemon-side work.
- Token never enters argv: `--password-stdin` (read from stdin) or `--from-file <path>` (read from a file); exactly one is required. No plaintext `--password` flag, mirroring `kuke create secret`.
- `--server` defaults to empty (matches the registry extracted from the image reference at pull time), documented in `--help`.
- Wired into the `create` parent (registration, `Short`, completion list) and documented in the user-facing `docs/site/cli/kuke-create.md`.

### Implementation notes (for reviewer push-back)

- **Why a GetRealm → merge → ApplyDocuments round-trip:** the apply diff overwrites `spec.registryCredentials` wholesale (`UpdateRealm`: `existing.Spec.RegistryCredentials = desired.Spec.RegistryCredentials`). Sending only the new entry would wipe existing credentials, so the command must read the live list and merge. The existing namespace is preserved verbatim in the desired doc — a changed namespace would be a *breaking* change and rejected by the reconciler.
- **Not-found detection:** `GetRealm` returns no error for a missing realm — it sets `MetadataExists=false` (`internal/controller/get_realm.go:61`). The command checks that flag and exits non-zero with `realm "<name>" not found` (AC: unknown realm → clear error).
- **Upsert key:** exact match on `ServerAddress`, so the empty-server default upserts against itself rather than appending duplicates.
- **Docs bundled:** the user-facing `kuke-create.md` reference is updated to document the new command (resources list, alias note, dedicated section). This documents *this* feature, not an unrelated cleanup. No `*/CLAUDE.md` edits.

## Test plan

- [x] `make test` (full CI suite: test-root + test-kukebuild) — passes (exit 0). Updated two pre-existing assertions in `cmd/kuke/create/create_test.go` (subcommand `Short` string + completion count) that enumerate the create subcommand set.
- [x] `GOWORK=off go vet ./cmd/kuke/create/...` — clean.
- [x] New unit tests `cmd/kuke/create/registrycredential/registrycredential_test.go` cover: stdin happy-path append, `--from-file`, in-place upsert (same server, no duplicate), append on different server, empty-server self-upsert, unknown-realm error, both-sources rejected, no-source rejected, missing-username rejected, empty-token rejected, and absence of a plaintext `--password` flag.
- [x] `./kuke create registry-credential --help` renders the command, both token-source flags, and the `--server` default.
- [ ] **End-to-end private-image pull** (AC: "a cell in that realm successfully pulls a private image that previously returned 403") — not run. This requires a live daemon plus real private-registry credentials, which the dev host cannot provide. The converge path the command relies on is pre-existing and exercised at the diff/reconcile layer; the new code is verified at the GetRealm→upsert→ApplyDocuments boundary in unit tests. Recommend the reviewer (or a smoke run with a real token) confirm the live pull. The change is client-CLI-only — it touches no daemon, build-path, or `/opt/kukeon` code — so the `make dev-init` two-realm regression guard is not implicated.

## Acceptance criteria

- [x] `--server <host> --username <u> --password-stdin` reads the token from stdin and upserts a `RegistryCredentials` entry on the named realm.
- [x] `--from-file <path>` accepted as an alternative; exactly one of `--password-stdin` / `--from-file` is required (error otherwise).
- [x] Token never passed as an argv flag and never logged.
- [x] Re-running with the same `--server` updates the existing entry in place (no duplicate).
- [ ] After the command, a cell pulls a previously-403 private image without a realm recreate — deferred to a live environment (see test plan).
- [x] Unknown realm → non-zero exit with a clear "realm not found" error.
- [x] `--help` documents the command, both token-source flags, and the `--server` default.

## Follow-up

Per the issue Notes, the thin `kuke team init` wiring (point this command at the `default` realm to resolve the team 403) should be filed as a separate `epic:team2` follow-up once this lands. Not included here.

Closes #1130

### DIFF
--- a/cmd/kuke/create/create.go
+++ b/cmd/kuke/create/create.go
@@ -23,6 +23,7 @@ import (
 	cellcmd "github.com/eminwux/kukeon/cmd/kuke/create/cell"
 	configcmd "github.com/eminwux/kukeon/cmd/kuke/create/config"
 	realmcmd "github.com/eminwux/kukeon/cmd/kuke/create/realm"
+	registrycredentialcmd "github.com/eminwux/kukeon/cmd/kuke/create/registrycredential"
 	secretcmd "github.com/eminwux/kukeon/cmd/kuke/create/secret"
 	spacecmd "github.com/eminwux/kukeon/cmd/kuke/create/space"
 	stackcmd "github.com/eminwux/kukeon/cmd/kuke/create/stack"
@@ -36,7 +37,7 @@ func NewCreateCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "create",
 		Aliases: []string{"c"},
-Short:   "Create Kukeon resources (realm, space, stack, cell, config, blueprint, secret)",
+		Short:   "Create Kukeon resources (realm, space, stack, cell, config, blueprint, secret, registry-credential)",
 		Run: func(cmd *cobra.Command, _ []string) {
 			_ = cmd.Help()
 		},
@@ -52,6 +53,7 @@ Short:   "Create Kukeon resources (realm, space, stack, cell, config, blueprint,
 		configcmd.NewConfigCmd(),
 		blueprintcmd.NewBlueprintCmd(),
 		secretcmd.NewSecretCmd(),
+		registrycredentialcmd.NewRegistryCredentialCmd(),
 	)
 
 	return cmd
@@ -59,7 +61,7 @@ Short:   "Create Kukeon resources (realm, space, stack, cell, config, blueprint,
 
 // completeCreateSubcommands provides shell completion for create subcommand names.
 func completeCreateSubcommands(_ *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-	subcommands := []string{"realm", "space", "stack", "cell", "config", "blueprint", "secret"}
+	subcommands := []string{"realm", "space", "stack", "cell", "config", "blueprint", "secret", "registry-credential"}
 
 	if toComplete == "" {
 		return subcommands, cobra.ShellCompDirectiveNoFileComp
--- a/cmd/kuke/create/create_test.go
+++ b/cmd/kuke/create/create_test.go
@@ -41,7 +41,7 @@ func TestNewCreateCmdMetadata(t *testing.T) {
 		{
 			name: "short description",
 			check: func(t *testing.T, cmd *cobra.Command) {
-				expected := "Create Kukeon resources (realm, space, stack, cell, config, blueprint, secret)"
+				expected := "Create Kukeon resources (realm, space, stack, cell, config, blueprint, secret, registry-credential)"
 				if cmd.Short != expected {
 					t.Fatalf("expected Short to be %q, got %q", expected, cmd.Short)
 				}
@@ -83,6 +83,7 @@ func TestNewCreateCmdRegistersSubcommands(t *testing.T) {
 		{name: "config"},
 		{name: "blueprint"},
 		{name: "secret"},
+		{name: "registry-credential"},
 	}
 
 	for _, tt := range tests {
@@ -105,7 +106,7 @@ func TestNewCreateCmd_AutocompleteRegistration(t *testing.T) {
 
 	// Test the completion function directly
 	completions, _ := cmd.ValidArgsFunction(cmd, []string{}, "")
-	expected := []string{"realm", "space", "stack", "cell", "config", "blueprint", "secret"}
+	expected := []string{"realm", "space", "stack", "cell", "config", "blueprint", "secret", "registry-credential"}
 	if len(completions) != len(expected) {
 		t.Fatalf("expected %d completions, got %d", len(expected), len(completions))
 	}
--- a/cmd/kuke/create/registrycredential/registrycredential.go
+++ b/cmd/kuke/create/registrycredential/registrycredential.go
@@ -1,0 +1,261 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package registrycredential implements `kuke create registry-credential` — an
+// imperative path to attach pull credentials to an existing realm's
+// spec.registryCredentials without round-tripping through a hand-edited YAML
+// manifest. The token enters via stdin or a file only, never argv, mirroring
+// `kuke create secret --from-file`. The command reads the realm, upserts the
+// credential keyed by --server, and re-applies the realm through the daemon's
+// write-through apply path, which the reconciler converges as a compatible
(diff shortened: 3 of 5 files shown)


eminwux:
## CI diagnosis: CodeQL alert is pre-existing on `main`, unrelated to this PR

**Failing check:** CodeQL — `1 new alert including 1 high severity` (`make test`, `make e2e`, and both `Analyze` jobs pass).

**Alert:** `go/allocation-size-overflow` at `internal/controller/apply.go:342`:

```go
out := make(map[string]string, len(labels)+1)   // stampTeamLabel
```

**Why this is not introduced by this PR:**

- This PR's diff is confined to `cmd/kuke/create/` (+ `docs/site/cli/kuke-create.md`). It does **not** touch `internal/controller/apply.go`.
- `git blame` shows line 342 was introduced by `99976d9` (#1029, *per-team labeling + pruning apply*, merged 2026-06-03) — well before this branch's base (`007ae96`). The code is byte-identical on `origin/main`.
- CodeQL default-setup reports it as "new" only because `main` carries no recorded baseline alert for this rule at this location; the underlying code is pre-existing.

**Assessment:** The sink is a map capacity hint sized by `len(labels)+1`, where `labels` is an in-memory map already held by the daemon. `len()` of an existing map cannot realistically drive an integer overflow, so this reads as a CodeQL false positive on the map-cap idiom rather than an exploitable allocation. Either way, the remediation belongs in `apply.go` (team-apply path), which is out of scope for this CLI feature — bundling that fix here would violate the repo's "no unrelated cleanups" rule.

**Recommended resolution (maintainer):** dismiss the alert as a false positive in the Security tab, or file a separate hardening issue against `internal/controller/apply.go` / #1029. Once dismissed (or fixed on `main`), this check goes green with no change to this PR.

Swapping `ready-for-review` → `ci-stalled` per dev `/diagnose-ci` step 5 so the cron stops retrying. Cleared once the alert is dismissed/triaged.

eminwux:
**PR #1132 Review — `feat(create): kuke create registry-credential — upsert realm pull credentials`**

LGTM — imperative `kuke create registry-credential` reads the realm, upserts a `spec.registryCredentials` entry keyed by `--server`, and re-applies through the write-through path; the reconciler converges it as a compatible change (`internal/controller/apply/diff.go:114`), so no recreate. Token enters via stdin/`--from-file` only and never reaches argv or output. Verified: all referenced symbols/shapes exist; `printApplyResult`'s `case "failed":` matches the apply path's `actionFailed = "failed"` (`internal/controller/apply.go:32`), so a per-resource failure surfaces its `Error` rather than printing a false success; namespace preserved verbatim so the diff stays non-breaking. Unit tests cover the full GetRealm→upsert→ApplyDocuments boundary (append, in-place upsert, empty-server self-upsert, unknown realm, both/no-source, missing username, empty token, no `--password` flag). License headers + colocated tests present.

The one unchecked test-plan item — live private-image pull (previously-403) — is host-gated (needs a real daemon + private-registry token the sandbox can't provide) and documented as such; non-blocking. Worth running pre-merge to confirm the end-to-end converge.

PR #1134, opened 2026-06-07
fix(team): default rendered Config space/stack to default to stop spurious OutOfSync

eminwux:
## Summary

- `teamrender` defaulted only the **realm** of rendered Blueprints/Configs to `default`, leaving **space**/**stack** as the (usually empty) template values. The CLI create path then defaults the empty space/stack to `default` and persists them on the live cell, while the daemon's reconciler re-materializes from the Config with the empty values verbatim — so `apply.DiffCell` saw `"" != "default"` and reported a permanent spurious `OutOfSync` on `spec.spaceName` / `spec.stackName` (#1133). This undermined the `epic:reconcile` drift signal.
- Fix closes the defaulting asymmetry at both ends:
  - **`internal/teamrender/teamrender.go`** — added `DefaultSpace` / `DefaultStack` consts; default `Inputs.Space` / `Inputs.Stack` alongside realm in `Render`; thread the defaulted space/stack through `RenderBlueprint` (stamped onto `bp.Metadata.Space`/`.Stack`), `BindConfig` (Config `Metadata` + `Spec.Blueprint` ref now use the defaulted values, not the empty template ones), and `renderContextValues` (the `.space`/`.stack` dot-context leaves now carry the real scope instead of `""`).
  - **`pkg/api/model/kuketeams/projectteam.go`** — surfaced optional `realm`/`space`/`stack` on `ProjectTeamSpec` so `kuketeam.yaml` can declare them (Expected-behavior item 1). Plumbed through `teamrender.Inputs` in **`cmd/kuke/team/init.go`**. The loader uses plain `yaml.Unmarshal` (not strict), so existing kuketeam.yaml files are unaffected.
  - **`internal/cellconfig/materialize.go`** (belt-and-suspenders, recommended by the issue) — `MaterializeWithName` now defaults an empty realm/space/stack to `default` (via `consts.KukeonDefault*Name`), hardening the reconcile re-materialization path against any Config that carries an empty coordinate, matching what the CLI create path persists.

[...]

Issue #1133, opened 2026-06-07
kuke team: rendered Config omits space/stack defaulting → spurious cell OutOfSync (spec.spaceName)


eminwux:
## Summary

Cells created by `kuke team init` are spuriously flagged `outOfSync: true` with `outOfSyncReason: spec differs at spec.spaceName` (and would also affect `spec.stackName`). Nothing actually changed in the cell — it is a **defaulting asymmetry** introduced by the team renderer.

`teamrender` defaults the **realm** of rendered Blueprints/Configs to `default`, but leaves **space** and **stack** as whatever the template supplied (normally empty). The CLI create path then defaults the empty space/stack to `default` and persists them on the live cell, while the daemon's reconciler re-materializes from the Config with the empty values verbatim — so `DiffCell` sees `"" != "default"` and reports drift forever.

## Repro

1. `kuke team init` a project whose role/harness templates don't set `metadata.space` / `metadata.stack` (the default templates).
2. The rendered `CellConfig` has `metadata.realm: default` but no `space` / `stack`.
3. The created cell persists at `default/default/default/<cell>` (space and stack defaulted to `default`).
4. On the next reconcile tick:

```
outOfSync: true
outOfSyncReason: spec differs at spec.spaceName
```

Observed on cell `kukeon-dev-0` (lineage Config `kukeon-dev-claude`):
- live cell: `spec.spaceId: default`, on disk at `/opt/kukeon/data/default/default/default/kukeon-dev-0`
- Config `kukeon-dev-claude`: `metadata.realm: default`, **no** `metadata.space` / `metadata.stack`

## Root cause

**`internal/teamrender/teamrender.go`** defaults realm only:

```go
// Render (~line 174)
realm := strings.TrimSpace(in.Realm)
if realm == "" {
    realm = DefaultRealm   // "default"
}
```

Both the rendered Blueprint and Config take realm from this defaulted value but copy space/stack straight from the template:

```go
// RenderBlueprint (~line 384)
bp.Metadata.Realm = realm            // defaulted
// bp.Metadata.Space / bp.Metadata.Stack — never defaulted, stay ""

// BindConfig (~line 438-450)
Metadata: v1beta1.CellConfigMetadata{
    Realm: realm,                    // defaulted
    Space: bp.Metadata.Space,        // "" when template omits it
    Stack: bp.Metadata.Stack,        // "" when template omits it
},
Spec: ...Blueprint{
    Realm: realm,
    Space: bp.Metadata.Space,
[...]

PR #1138, opened 2026-06-07
fix(controller): reproduce runner-injected container fields in OutOfSync re-materialization

### LINKED ISSUE #1136: Spurious OutOfSync: runner-injected cniConfigPath (and space container-defaults) not reproduced by re-materialization
## Summary

Config-lineage cells are reported permanently `OutOfSync` with reason `spec differs at containers[<name>].cniConfigPath`, even when nothing has actually drifted. The runner bakes a space/runtime-derived `cniConfigPath` onto every persisted container spec, but the reconciler's re-materialization (`MaterializeWithName`) never reproduces it, so `apply.DiffCell` always sees `"" != "/opt/.../network.conflist"`.

This is the same bug class as #1133/#1134 (empty realm/space/stack defaulting), one level down at the container spec.

## Reproduction

```
$ kuke get cell dezot-dev-0 -o yaml
...
  containers:
    - id: work
      ...
      cniConfigPath: /opt/kukeon/data/default/default/network.conflist
status:
  ...
  outOfSync: true
  outOfSyncReason: spec differs at containers[work].cniConfigPath
```

## Root cause

`reconcileCellOutOfSync` → `desiredCellOutOfSync` (`internal/controller/reconcile_outofsync.go`):

1. `materializeCellFromConfig` → `cellconfig.MaterializeWithName(config, blueprint, name)` builds the "desired" cell from **Config + Blueprint only**.
2. `apply.DiffCell(desired, live)` diffs it against the live persisted cell.

Any field the **runner** bakes onto the persisted spec at create/start that materialization does not reproduce reads as drift:

- **Live side:** `internal/controller/runner/provision.go:1486/1902/2016` set `containerSpec.CNIConfigPath = cniConfigPath`, resolved from the space via `resolveSpaceCNIConfigPath(realm, space)` (`start.go:548`). It is a space/runtime-derived value, not Config-authored.
- **Desired side:** `cellconfig.MaterializeWithName` leaves `CNIConfigPath` empty.
- **Diff:** `internal/controller/apply/diff.go:796` compares `desired.CNIConfigPath != actual.CNIConfigPath` → flags `containers[<name>].cniConfigPath`.

`DiffCell` deliberately skips the runner-synthesized root container (`desiredRoot == nil` branch, `diff.go:438–448`), which is why only non-root containers (e.g. `work`) are flagged, never `root`.

## Other fields with the same root cause (latent)

`provision.go` also runs `ApplySpaceDefaultsToContainer(parentSpace, &cell.Spec.Containers[i])` (`internal/modelhub/merge.go`), baking these onto each persisted container spec from `space.Spec.Defaults.Container`:

- `user`, `readOnlyRootFilesystem`, `capabilities`, `securityOpts`, `tmpfs`, `resources`

**All six are compared by `diffContainerSpec`.** Materialization has no Space in hand, so it never applies them. The `default` space has no `defaults.container` block, which is why only `cniConfigPath` fires today — but any space with container defaults set will spuriously drift on whichever of those six it defines. Same mechanism, currently dormant.

Runner-injected fields that are **safe** (not part of `diffContainerSpec`, so no drift): `containerdId`, per-container `cellName/realmName/spaceName/stackName`, per-container `nestedCgroupRuntime`.

## Suggested fix

Mirror the #1134 approach — make the re-materialized "desired" reflect what the runner actually persists. Preferred: in `materializeCellFromConfig` (`reconcile_outofsync.go`, which already holds the runner `r`), after materializing, resolve the space's `cniConfigPath` onto each container and run `ApplySpaceDefaultsToContainer` using the cell's space. This closes both `cniConfigPath` and the latent space-defaults class in one shot. Add a colocated case to `internal/controller/reconcile_outofsync_test.go`.

Alternative (narrower): treat an empty `desired.CNIConfigPath` as "unauthored" and skip that comparison in `diffContainerSpec`, analogous to the managed-labels filter and the root-container skip.

eminwux:
## Summary
- Config-lineage cells were reported permanently `OutOfSync` with reason `spec differs at containers[<name>].cniConfigPath` even when nothing drifted (#1136). The runner bakes a space/runtime-derived `cniConfigPath` (and, when a space declares them, container defaults) onto every persisted container spec at create/start, but the reconciler's re-materialization (`cellconfig.MaterializeWithName`, Config + Blueprint only) reproduced neither — so `apply.DiffCell` always saw `"" != "/opt/.../network.conflist"`.
- Same bug class as #1133/#1134, one level down at the container spec. Applies the **preferred** fix from the issue: re-derive the runner-injected fields in `materializeCellFromConfig` (which holds the runner `r`), closing both the active `cniConfigPath` drift and the latent space-defaults class (`user`, `readOnlyRootFilesystem`, `capabilities`, `securityOpts`, `tmpfs`, `resources`) in one shot.

## Changes
- `internal/controller/reconcile_outofsync.go`: new `applyRunnerInjectedContainerFields(r, &desired)`, called at the tail of `materializeCellFromConfig`. It fetches the cell's parent Space and runs `ApplySpaceDefaultsToContainer` per container, then sets the resolved `cniConfigPath` on each container that lacks an explicit one — mirroring `provision.go`'s `applyCellContainerDefaults` and the `if == ""` guard. The runner-synthesized root container is untouched because `DiffCell` already skips it. A `GetSpace`/resolve error surfaces as the existing undecidable-`OutOfSyncError` state.
- `internal/controller/runner/lifecycle.go` + `runner.go`: exported the existing `resolveSpaceCNIConfigPath` as `ResolveSpaceCNIConfigPath` and added it to the `Runner` interface, so the detector resolves the path through the **same** single source of truth the runner uses (the default-path branch needs the runner's run-path, which is not otherwise reachable through the interface). Internal callers (`start.go`, `provision.go`, `kill.go`, `stop.go`, `delete_space.go`, `recreate_cell.go`) updated to the exported name.
- `materializeCellFromConfig` is also used by the `StartCell` OutOfSync reapply path (`start_cell.go`), so the fix benefits that path too; its tests' `fakeRunner` fixtures gained the two new dependency stubs (returning empties, so the intended image/hostNetwork drift remains the only diff).

## Implementation notes
- **Gate-vs-flip / blast radius:** chose the issue's *preferred* approach over the narrower "skip empty `desired.CNIConfigPath` in `diffContainerSpec`" alternative because the latter leaves the latent space-defaults class unfixed; the preferred approach reuses the runner's own resolver so the desired value is byte-identical to what was persisted.
- The cell's own realm/space drive the lookups, matching the scope the runner resolved at create time (consistent with #1134's `default`-scope defaulting, so the materialized scope equals the live cell's).

## Test plan
- [x] `make test` (full CI target: `test-root` + `test-kukebuild`) — green.
- [x] New `TestReconcileCells_OutOfSync_RunnerInjectedContainerFieldsStaySynced` — a cell carrying only the runner-injected `cniConfigPath` + a space-default `user` now stays Synced with no metadata write (fails before the fix).
- [x] Existing `OutOfSync` + `StartCell_OutOfSyncReapply` suites — green after wiring the new stubs.
- [x] `GOWORK=off go build ./...` / `go vet ./internal/controller/...` — clean.
- [x] Interface-implementor sweep: `Runner` implementors are `*Exec` (production) and `fakeRunner` (tests); `reconcileFakeRunner` embeds the interface. All carry the new method.
- [x] Tree-wide `git grep resolveSpaceCNIConfigPath` — no stale lowercase references remain.

Closes #1136

### DIFF
--- a/internal/controller/controller_test.go
+++ b/internal/controller/controller_test.go
@@ -45,12 +45,13 @@ type fakeRunner struct {
 	DeleteRealmFn                    func(realm intmodel.Realm) error
 
 	// Space methods
-	GetSpaceFn             func(space intmodel.Space) (intmodel.Space, error)
-	ListSpacesFn           func(realmName string) ([]intmodel.Space, error)
-	CreateSpaceFn          func(space intmodel.Space) (intmodel.Space, error)
-	EnsureSpaceFn          func(space intmodel.Space) (intmodel.Space, error)
-	ExistsSpaceCNIConfigFn func(space intmodel.Space) (bool, error)
-	DeleteSpaceFn          func(space intmodel.Space) error
+	GetSpaceFn                  func(space intmodel.Space) (intmodel.Space, error)
+	ListSpacesFn                func(realmName string) ([]intmodel.Space, error)
+	CreateSpaceFn               func(space intmodel.Space) (intmodel.Space, error)
+	EnsureSpaceFn               func(space intmodel.Space) (intmodel.Space, error)
+	ExistsSpaceCNIConfigFn      func(space intmodel.Space) (bool, error)
+	ResolveSpaceCNIConfigPathFn func(realmName, spaceName string) (string, error)
+	DeleteSpaceFn               func(space intmodel.Space) error
 
 	// Stack methods
 	GetStackFn    func(stack intmodel.Stack) (intmodel.Stack, error)
@@ -202,6 +203,13 @@ func (f *fakeRunner) GetSpace(space intmodel.Space) (intmodel.Space, error) {
 		return f.GetSpaceFn(space)
 	}
 	return intmodel.Space{}, errors.New("unexpected call to GetSpace")
+}
+
+func (f *fakeRunner) ResolveSpaceCNIConfigPath(realmName, spaceName string) (string, error) {
+	if f.ResolveSpaceCNIConfigPathFn != nil {
+		return f.ResolveSpaceCNIConfigPathFn(realmName, spaceName)
+	}
+	return "", errors.New("unexpected call to ResolveSpaceCNIConfigPath")
 }
 
 func (f *fakeRunner) ListSpaces(realmName string) ([]intmodel.Space, error) {
--- a/internal/controller/reconcile_outofsync.go
+++ b/internal/controller/reconcile_outofsync.go
@@ -255,7 +255,63 @@ func materializeCellFromConfig(
 	if convErr != nil {
 		return intmodel.Cell{}, fmt.Errorf("convert materialized cell: %w", convErr)
 	}
+
+	// Reproduce the runner-injected, space-derived container fields the
+	// re-materialization does not, so apply.DiffCell does not report them as
+	// spurious drift (#1136).
+	if err := applyRunnerInjectedContainerFields(r, &desired); err != nil {
+		return intmodel.Cell{}, err
+	}
+
 	return desired, nil
+}
+
+// applyRunnerInjectedContainerFields re-applies the space-derived container
+// fields the runner bakes onto every persisted container spec at create/start
+// but cellconfig.MaterializeWithName (Config + Blueprint only) never
+// reproduces. Without this, apply.DiffCell flags those runner-injected values
+// as spurious drift and a Config-lineage cell sticks on a permanent OutOfSync
+// even though nothing actually changed (#1136). Two field classes are
+// reproduced, mirroring provision.go's create/start path:
+//
+//   - Space container defaults (user, readOnlyRootFilesystem, capabilities,
+//     securityOpts, tmpfs, resources) via ApplySpaceDefaultsToContainer — the
+//     same helper applyCellContainerDefaults funnels every create flow
+//     through. The `default` space declares no container defaults, so this is
+//     dormant today but closes the latent class the moment a space sets one.
+//   - The space/runtime-derived cniConfigPath, resolved by the same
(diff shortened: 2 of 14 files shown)


eminwux:
**PR #1138 Review — `fix(controller): reproduce runner-injected container fields in OutOfSync re-materialization`**

Clean fix. The re-materialization re-applies exactly the per-container fields `diffContainerSpec` compares — `cniConfigPath` plus the six space-default fields — through the same `ApplySpaceDefaultsToContainer` helper and the same `if == ""` cniConfigPath guard `provision.go` uses; `NestedCgroupRuntime` is compared only at cell-spec level so it's correctly out of scope. The new `GetSpace`/`ResolveSpaceCNIConfigPath` error flows through `materializeErr` into `cellOutOfSync{Err: ...}`, inheriting the existing undecidable-OutOfSync classification. Export + interface threading is consistent, CI green, no secrets, no dead code (`git grep` confirms zero stale `resolveSpaceCNIConfigPath` and every new symbol is referenced).

One non-blocking note:

- **Test coverage** `internal/controller/reconcile_outofsync_test.go`: the new unit test reproduces the exact `"" != /opt/.../network.conflist` mismatch and fails pre-fix — the right level for this deterministic comparison logic. An end-to-end confirmation on a real daemon (a `--from-config` cell staying `Synced` across a reconcile tick) is host-gated and not runnable in the sandbox; worth a quick manual check pre-merge, not a blocker given the unit reproduction.

LGTM to proceed.

PR #1120, opened 2026-06-06
feat(team): two-layer secrets.env → kind: Secret apply

eminwux:
## Summary

- Add `internal/teamsecrets` package: scaffolds the host-wide `~/.kuke/teams/secrets.env` and per-team `<teamDir>/secrets.env` (mode 0o600) on first run with one empty `KEY=` line per role-declared secret name, idempotent on re-run; loads + merges (per-team non-empty overrides shared) + renders each non-empty merged entry as a `kind: Secret` in realm `default` with a kebab-case `metadata.name` (`ANTHROPIC_AUTH_TOKEN` → `anthropic-auth-token`).
- Extend `teamrender.Result` with `Secrets` and reorder `MarshalYAML` to emit `Secrets → Blueprints → Configs` so the per-team apply bundle (per-team prune-apply contract from #1029) carries the secret material before the blueprint/config pair that references it.
- Wire the secrets pipeline into `cmd/kuke/team/init.go`'s `composeTeam`: discover the union of `needs.secrets` from every roster role, run `teamsecrets.Compose` (skipped scaffold under `--dry-run`), surface one warning-per-empty-key on stderr (key name only — values are never logged or echoed), update the apply-summary line to `applied N secret/M blueprint/K config object(s)`.

## Non-obvious design call

The AC says "per-team value wins for matching keys". A literal reading is footgunny: the scaffold writes one empty `KEY=` line per role-declared secret into the per-team file, so on the very first re-init after scaffold, **every** key in shared would be null-out by the empty per-team placeholder — the shared layer would become unreachable without hand-curation of every per-team file. The implementation reads "wins" as "non-empty wins": a per-team value with non-empty bytes overrides shared; an empty per-team value is a placeholder, not an override. This keeps the shared layer useful as a default while still letting the operator override per-key by writing a value. Pinned via `TestMergePerTeamEmptyFallsBackToShared`. Push back if you'd rather have strict "any-present-wins" semantics — the change is one-line in `Merge`.

## Test plan

- [x] `make test-root` — full `go test ./...` (excludes /e2e) passes locally.
- [x] `make test-kukebuild` — `cd cmd/kukebuild && go test ./...` passes locally.
- [x] `pre-commit run --files <touched>` — go fmt / go-mod-tidy / golangci-lint / addlicense all pass.
[...]

PR #1048, opened 2026-06-03
feat(team): render pipeline (needs-merge, image-select, bind) (step 3)

eminwux:
## Summary

- Step 3 of \`kuke team init\` (epic team-distribution #792): turns the materialized agents source from step 2 (#1046) into in-memory \`CellBlueprint\` / \`CellConfig\` objects ready for step 4 (#1043) to apply with prune.
- New \`internal/teamrender\` package owns the per-\`(role × harness)\` pipeline — needs-merge → image-select → render → bind → label \`kukeon.io/team=<project>\`. Tested directly against tempdir fixtures so the package is hermetic.
- \`cmd/kuke/team/init.go\` wires the resolver (\`teamsource.Resolve\` from #1046) and the renderer end-to-end. Injectable via context keys so the existing init tests stay offline.
- \`--dry-run\` now stops after render and prints the multi-doc YAML (entry + blueprints + configs) to stdout. It writes nothing — neither \`~/.kuke/kuketeams.yaml\`, nor \`~/.kuke/kuketeam.d/<project>.yaml\`, nor anywhere under \`~/.kuke/rendered/\`.

## Pipeline shape

For each \`(role × harness)\`:

1. **needs-merge** — \`union(role.spec.needs.image ⊕ ptRole.needs.image)\`, dedup + lex-sort. Same inputs → byte-identical output.
2. **image-select** — match merged caps against \`ImageCatalog\` entries whose \`harness\` matches; first matching entry wins. Misses surface \`errdefs.ErrTeamImageNoMatch\` naming the first unmet capability + the build-or-label hint (umbrella key decision: selection only, no Dockerfile build, no LLM).
3. **render** — load \`harnesses/<h>/blueprint.tmpl.yaml\` relative to the cache dir, substitute \`\${ROLE}\`, \`\${HARNESS}\`, \`\${IMAGE}\`, \`\${NEEDS}\`, plus the role's native per-harness config knobs (\`\${SETTINGS}\` for claude, \`\${SANDBOX}\`/\`\${APPROVAL}\` for codex, \`\${PERMISSIONS}\` for opencode) — wired verbatim per the umbrella's no-transpile decision. Unknown placeholders pass through so blueprints can carry literal \$-prefixed env-var defaults.
4. **bind** — produce a \`CellConfig\` referencing the rendered blueprint:
[...]

PR #1147, opened 2026-06-07
fix(team): bind role.needs.secrets fills regardless of tc.spec.secrets

### LINKED ISSUE #1145: kuke team init: rendered CellConfig has no secret fills; secret env vars never injected into cells
## Summary

`kuke team init` renders `CellConfig`s with **no `spec.secrets` section**, so secret env vars (e.g. `CLAUDE_CODE_OAUTH_TOKEN`) are never injected into cells started with `kuke run --from-config`. The `kind: Secret` is created correctly and its bytes land on disk, but nothing binds the blueprint's secret slot to it, so at materialize time the (optional) slots drop silently and no env var reaches the container.

## Repro

```
kuke team init
kuke get secret claude-code-oauth-token          # exists (spec: {} is expected redaction)
kuke get config kukeon-pm-claude --realm default --space default --stack default
# -> spec has blueprint/values/repos but NO secrets:
kuke run --from-config kukeon-pm-claude --name kukeon-pm-0 --realm default --space default --stack default
# -> cell starts, but CLAUDE_CODE_OAUTH_TOKEN is not set in the container
```

The Secret bytes do exist on disk:

```
/opt/kukeon/data/default/secrets/claude-code-oauth-token
```

## Root cause

Two code paths drifted apart:

- **Secret *creation*** was migrated to `secrets.env` + role `needs.secrets` in **#1120** (`feat(team): two-layer secrets.env -> kind: Secret apply`). This path works — it creates the `kind: Secret` and writes the bytes.
- **Secret *binding into the CellConfig*** still uses the older gate from **#1048**. In `internal/teamrender/teamrender.go:512-528`, a `CellConfigSecretFill` is only added when the secret name is present in `tc.Spec.Secrets` (the operator-global `~/.kuke/kuketeams.yaml`):

```go
if len(declaredSecrets) > 0 && r != nil && len(r.Spec.Needs.Secrets) > 0 && tc != nil {
    for _, secretName := range r.Spec.Needs.Secrets {
        if !declaredSecrets[secretName] {
            continue
        }
        if _, ok := tc.Spec.Secrets[secretName]; !ok {   // <-- always taken
            continue
        }
        ...
    }
}
```

Nothing in non-test code ever populates `tc.Spec.Secrets`, and the scaffold (`buildGlobalConfig`) never writes a `spec.secrets:` block. So the gate is always false, no fill is ever added, and every rendered `CellConfig` comes out with empty secrets.

Downstream consequence:
- `internal/cellconfig/materialize.go:177-187` only emits a `ContainerSecret` for a slot that has a matching `cfg.Spec.Secrets[slot.Name]` fill. The blueprint slots are optional (no `required: true`), so with no fill they are silently dropped.
- `internal/ctr/secrets.go:121` therefore never produces the `NAME=value` env entry.

## Suggested fix

In `internal/teamrender/teamrender.go:512-518`, gate the binding on the same source of truth as creation — the role's `needs.secrets` (already iterated) — instead of the now-vestigial `tc.Spec.Secrets`. Drop the `tc.Spec.Secrets` presence check (and the `tc != nil` requirement); the fill's `SecretRef` is built purely from `secretName` + `realm` and needs nothing from `tc`. Add a regression test asserting that a role with `needs.secrets` + a blueprint declaring the slot yields a `CellConfig` with the corresponding `spec.secrets` fill.

## Workaround (no rebuild)

Add the secrets to `~/.kuke/kuketeams.yaml` so the existing gate passes, then re-run `kuke team init`:

```yaml
spec:
  secrets:
    claude-code-oauth-token: { from: env, key: CLAUDE_CODE_OAUTH_TOKEN }
    openai-api-key:          { from: env, key: OPENAI_API_KEY }
    openrouter-api-key:      { from: env, key: OPENROUTER_API_KEY }
```

## Notes

- `spec: {}` on `kuke get secret` is intentional redaction (#622), not part of this bug.
- The empty-key warnings for `openai-api-key` / `openrouter-api-key` are a separate, expected condition.

eminwux:
## Summary

- The `secrets.env` → `kind: Secret` apply path from #1120 stopped populating `tc.Spec.Secrets`, but `BindConfig` (left over from #1048) still gated the per-slot `CellConfigSecretFill` on `tc.Spec.Secrets[secretName]`. That gate is always false outside tests, so every rendered CellConfig came out with empty `spec.secrets` and no secret env var ever reached a cell started via `kuke run --from-config`.
- Drop the `tc.Spec.Secrets` presence check (and the now-vestigial `tc != nil` guard). Gate the fill solely on the role's `Needs.Secrets` intersected with the blueprint's `BlueprintSecretSlot` declarations — the same source of truth the creation side uses — so a role need + a blueprint slot is sufficient to produce a `ContainerSecretRef`.
- Refresh the `BindConfig` godoc paragraph that still described the old "filled from tc.spec.secrets" behavior so the contract reads correctly.

## Test plan

- [x] `go test ./internal/teamrender/` — covers the existing `TestBindConfigStampsTeamLabelAndProjectRepoFill` (still passes; `tc.Spec.Secrets` happens to be populated) plus the new `TestBindConfigFillsSecretFromRoleNeedsWithoutTcSecrets` regression that drives the production `&model.TeamsConfig{}` shape from #1120.
- [x] `make test` (full project test suite green).
- [ ] Live `kuke team init` smoke against an actual host — not run here; the unit test pins the byte-shape of the rendered `CellConfigSecretFill`, and `cmd/kuke/team` package tests pass.

Closes #1145

### DIFF
--- a/internal/teamrender/teamrender.go
+++ b/internal/teamrender/teamrender.go
@@ -453,10 +453,11 @@ func scopeToProject(name, project string) string {
 // the `agents` slot — both only when the blueprint actually declares the
 // slot, so a template that doesn't carry the slot produces a config
 // without a stray fill. role.Spec.Needs.Secrets entries are matched
-// against the blueprint's BlueprintSecretSlot declarations and filled
-// from tc.spec.secrets — when the operator has a matching entry, an
-// in-realm ContainerSecretRef points the slot at it (the runtime
-// resolves the actual bytes via the Secret kind from #623).
+// against the blueprint's BlueprintSecretSlot declarations and, when the
+// blueprint declares the slot, filled with an in-realm ContainerSecretRef
+// pointing at the secret of the same name — the runtime resolves the
+// actual bytes via the Secret kind (#623) created out of the two-layer
+// secrets.env path (#1120).
 func BindConfig(
 	bp *v1beta1.CellBlueprintDoc,
 	r *model.Role,
@@ -509,12 +510,9 @@ func BindConfig(
 		}
 	}
 
-	if len(declaredSecrets) > 0 && r != nil && len(r.Spec.Needs.Secrets) > 0 && tc != nil {
+	if len(declaredSecrets) > 0 && r != nil && len(r.Spec.Needs.Secrets) > 0 {
 		for _, secretName := range r.Spec.Needs.Secrets {
 			if !declaredSecrets[secretName] {
-				continue
-			}
-			if _, ok := tc.Spec.Secrets[secretName]; !ok {
 				continue
 			}
 			if cfg.Spec.Secrets == nil {
--- a/internal/teamrender/teamrender.go
+++ b/internal/teamrender/teamrender.go
@@ -41,8 +41,10 @@
 //     blueprint and carries (a) operator facts pulled from
 //     `~/.kuke/kuketeams.yaml` (git author/committer/signingKey/registry
 //     stamped into Values), (b) the project's cloned repo URL stamped
-//     into the `project` repo slot fill, and (c) any operator secret
-//     refs from `tc.spec.secrets` that the role declares it needs.
+//     into the `project` repo slot fill, and (c) for every secret the
+//     role declares it needs and the blueprint declares a slot for, an
+//     in-realm ContainerSecretRef pointing at the Secret of the same
+//     name (created out of `secrets.env`, #1120).
 //  5. label — every rendered CellBlueprint and CellConfig carries
 //     `metadata.labels[kukeon.io/team] = <project>` so the daemon-side
 //     prune-apply machinery from #1029 can converge the project's slice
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1147 Review — `fix(team): bind role.needs.secrets fills regardless of tc.spec.secrets`**

LGTM — correctly drops the always-false `tc.Spec.Secrets` gate so the config-fill side keys on `role.Needs.Secrets ∩ declaredSecrets`, the same source of truth the blueprint side already uses (`roleNeedsSecrets(r)` at `teamrender.go:765`). The fill block no longer references `tc`, so removing the `tc != nil` guard is safe, and `tc` is still dereferenced earlier so a nil would surface upstream regardless. The new `TestBindConfigFillsSecretFromRoleNeedsWithoutTcSecrets` drives the production `&model.TeamsConfig{}` shape and pins the regression; `make test` is green in CI.

Non-blocking: the live `kuke team init` smoke is the one unchecked test-plan item — documented host-gated and not runnable here. Worth running pre-merge to confirm a secret env var actually reaches a `kuke run --from-config` cell end-to-end.

PR #1150, opened 2026-06-07
fix(team): prune empty-secret bindings from rendered CellConfigs

### LINKED ISSUE #1149: kuke team init: empty secrets.env keys are bound into CellConfig but never created, blocking cell start
## Summary

Follow-up to #1145 / #1147. The bind fix (#1147) now adds a `CellConfigSecretFill` for **every** `role.needs.secrets` entry, regardless of whether a backing `kind: Secret` was actually created. Empty secrets.env values are **skipped** during secret creation but **still bound** into the CellConfig, so a role that needs a secret the operator hasn't provided makes the whole cell fail to start.

Net effect: #1147 swung from "bind nothing" (the #1145 bug) to "bind everything". The two paths still disagree — creation skips empty values, binding includes them.

## Repro

`pm` role declares:

```yaml
needs:
  secrets:
    - claude-code-oauth-token
    - openai-api-key
    - openrouter-api-key
```

Operator only has a value for `claude-code-oauth-token`:

| key | shared `~/.kuke/teams/secrets.env` | team `.../kukeon/secrets.env` | merged |
|---|---|---|---|
| claude-code-oauth-token | value | empty | has value |
| openai-api-key | empty | empty | EMPTY |
| openrouter-api-key | empty | empty | EMPTY |

```
kuke team init
# warning: secrets.env key "openai-api-key" is empty; cells referencing it will fail to start
# warning: secrets.env key "openrouter-api-key" is empty; cells referencing it will fail to start
#   Secret "claude-code-oauth-token": updated      <- only ONE secret created

kuke get secrets --realm default
# claude-code-oauth-token  (only this; openai-api-key never created)

kuke get config kukeon-pm-claude --realm default --space default --stack default
# spec.secrets now references ALL THREE (claude-code-oauth-token, openai-api-key, openrouter-api-key)

kuke run --from-config kukeon-pm-claude --name kukeon-pm-0 --realm default --space default --stack default
# Error: failed to create cell: failed to create container default_default_kukeon-pm-0_work:
#   failed to resolve secrets: referenced secret does not exist in the requested scope
#   (secret "OPENAI_API_KEY" ref "openai-api-key" at "/opt/kukeon/data/default/secrets/openai-api-key")
```

The cell dies on the first missing ref, so even the valid `claude-code-oauth-token` is never injected.

## Root cause

Two paths disagree on empty secrets:

1. **Creation** — `teamsecrets.Render` skips empty merged values (no `kind: Secret` emitted). Confirmed by daemon state: only `claude-code-oauth-token` exists.
2. **Binding** — post-#1147 `internal/teamrender/teamrender.go:515` binds every `role.Spec.Needs.Secrets` entry unconditionally, so the CellConfig references secrets that were never created.
3. **Cell start** — `internal/ctr/secrets.go:190` (`readSecretRefValue`) iterates container secrets and fails on the first missing file, blocking cell creation entirely.

## Suggested fix

Reconcile the binding with the set of secrets actually created. `composeSecrets` in `cmd/kuke/team/init.go` runs *after* render and already holds both `res.Configs` and `composed.EmptyKeys`. It should prune each `cfg.Spec.Secrets` of the empty keys (the same keys it already warns about) before apply, so the cell starts with exactly the secrets that exist. Render itself can't do this (it runs before composition knows the values), so the post-compose prune is the right layer.

A role declaring `needs.secrets` for credentials the operator hasn't provided should not make the whole cell unstartable — the empty-key warning already informs the operator those are unavailable.

Add a regression test: role needs 3 secrets, only 1 provided -> rendered CellConfig binds only the 1 provided; cell materializes with a single ContainerSecret.

## Secondary hazard

A stale `openrouter-api-key` file lingers under `/opt/kukeon/data/default/secrets/` from a prior run even though it is empty/unmanaged now. If iteration order ever reached it, the cell would resolve a stale value instead of failing. Pruning the binding sidesteps this; a separate cleanup of orphaned secret files on re-init may also be warranted.

eminwux:
## Summary
- #1147 fixed the #1145 "bind nothing" bug by binding a `CellConfigSecretFill` for **every** `role.needs.secrets` entry — but it overshot to "bind everything". Secret *creation* (`teamsecrets.Render`) skips empty merged `secrets.env` values (no `kind: Secret` emitted), while *binding* (`internal/teamrender/teamrender.go:515`) bound them unconditionally. A role that needs a secret the operator never provided then made the whole cell unstartable: `readSecretRefValue` (`internal/ctr/secrets.go:190`) fails on the first missing ref, so even the valid secrets are never injected.
- Reconciles the two paths at the post-compose layer the issue points at: `composeSecrets` in `cmd/kuke/team/init.go` already holds both `res.Configs` and `composed.EmptyKeys`, so a new `pruneEmptySecretBindings` helper deletes each empty key from every `cfg.Spec.Secrets` after render. The CellConfig is left referencing exactly the secrets that exist. Render itself can't do this — it runs before composition knows the merged values — so the post-compose prune is the right layer.
- **Namespace note:** the binding map keys and `EmptyKeys` are the *same raw* `role.Spec.Needs.Secrets` strings — `teamrender.BindConfig` iterates `role.Spec.Needs.Secrets`, and `EmptyKeys` is the empty subset of `UnionNeedsSecrets(bundle.Roles, …)` over the same role objects — so a direct map `delete` is namespace-consistent (no kebab/SCREAMING_SNAKE translation needed at the prune layer).
- The prune runs on both the apply and dry-run paths (it mutates in-memory `res.Configs`), so the dry-run output reflects the same pruned bindings the apply uses.

## Out of scope
- The issue's "secondary hazard" (a stale empty `secrets.env`-derived file lingering under `/opt/kukeon/data/.../secrets/`) is sidestepped by pruning the binding — a pruned secret is never referenced, so iteration order can never resolve a stale value. A standalone cleanup of orphaned secret files on re-init is a separate concern and not addressed here.

## Test plan
- [x] `make test` (full suite, `test-root` + `test-kukebuild`) — exit 0
- [x] `TestPruneEmptySecretBindings` — the #1149 regression: role needs 3 secrets, only `claude-code-oauth-token` provided → pruned CellConfig binds only that one
- [x] `TestPruneEmptySecretBindingsNoOps` — empty-key set / nil configs / nil entry are no-ops, no panic
- [x] `go build ./cmd/kuke/team/` (GOWORK=off, per Makefile) — exit 0

Closes #1149

### DIFF
--- a/cmd/kuke/team/init.go
+++ b/cmd/kuke/team/init.go
@@ -486,7 +486,43 @@ func composeSecrets(
 		fmt.Fprintf(errOut, "warning: secrets.env key %q is empty; cells referencing it will fail to start\n", k)
 	}
 	res.Secrets = composed.Secrets
+	// Reconcile the binding with the set of secrets actually created. Render
+	// skips empty merged values (no kind: Secret emitted), but the post-#1147
+	// binding in teamrender adds a CellConfigSecretFill for every
+	// role.needs.secrets entry regardless. Pruning the empty keys here — the
+	// same keys warned about above — leaves each CellConfig referencing only
+	// secrets that exist, so a role declaring needs.secrets for credentials
+	// the operator hasn't provided no longer makes the whole cell unstartable
+	// (#1149).
+	pruneEmptySecretBindings(res.Configs, composed.EmptyKeys)
 	return nil
+}
+
+// pruneEmptySecretBindings deletes the CellConfig secret bindings whose key is
+// in emptyKeys — the role.needs.secrets entries whose merged secrets.env value
+// was empty, which secret creation (teamsecrets.Render) skipped. Both the
+// binding map keys and emptyKeys are the raw role.needs.secrets strings (the
+// binding iterates role.Spec.Needs.Secrets; emptyKeys is the empty subset of
+// the same union), so a direct map delete is namespace-consistent. Mutates the
+// CellConfigs in place; a nil/empty input on either side is a no-op.
+func pruneEmptySecretBindings(configs []*v1beta1.CellConfigDoc, emptyKeys []string) {
+	if len(configs) == 0 || len(emptyKeys) == 0 {
+		return
+	}
+	empty := make(map[string]struct{}, len(emptyKeys))
+	for _, k := range emptyKeys {
+		empty[k] = struct{}{}
+	}
+	for _, cfg := range configs {
+		if cfg == nil || len(cfg.Spec.Secrets) == 0 {
+			continue
+		}
+		for key := range cfg.Spec.Secrets {
+			if _, isEmpty := empty[key]; isEmpty {
+				delete(cfg.Spec.Secrets, key)
+			}
+		}
+	}
 }
 
 // rosterPairs flattens pt.Spec.Roles × pt.Spec.Defaults.Harnesses into the
--- a/cmd/kuke/team/init_test.go
+++ b/cmd/kuke/team/init_test.go
@@ -520,6 +520,69 @@ func TestComposeTeamRendersOnNonDryRun(t *testing.T) {
 	}
 }
 
+// TestPruneEmptySecretBindings is the #1149 regression: a role needs 3
+// secrets but the operator provided only 1, so secret creation skipped the 2
+// empty values while the post-#1147 binding bound all 3. The prune must leave
+// the CellConfig referencing only the single secret that was actually created,
+// so the cell starts instead of dying on the first missing ref.
+func TestPruneEmptySecretBindings(t *testing.T) {
+	t.Parallel()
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #1150 Review — `fix(team): prune empty-secret bindings from rendered CellConfigs`**

LGTM — post-compose prune of empty `needs.secrets` bindings is at the right layer: `renderTeam` binds before `composeSecrets` runs, both dry-run and apply consume the pruned `res.Configs`, and binding keys + `composed.EmptyKeys` are the same raw `role.Spec.Needs.Secrets` strings (kebab only touches the created Secret's `metadata.name`), so the direct map `delete` is namespace-consistent. No secret values touched, no leaks, tests cover the #1149 regression + nil/empty no-ops, CI green.

Issue #751, opened 2026-05-23
planning: kind: Lease — daemon coordination primitive for cross-cell work claims


eminwux:
## Context

Today kukeon has no daemon-side coordination primitive for cross-cell work claims. The runtime models cells and their lifecycles, but two cells racing to claim the same work item (e.g. two dev cells both trying to pick up the same project-repo issue) have nothing to arbitrate on — labels on the work item itself are a check-then-act race (read "no claim label" → both decide to claim → both write the label → both proceed).

The fix is to separate two concerns that label-based claims conflate: *where the work is described* (the forge — GitHub/Forgejo issue) and *who is allowed to work it right now* (the daemon — a lease). The forge keeps the description; the daemon owns the claim. The daemon is the right home for the claim because it already owns cell lifecycle, so it has the two pieces of information that make a lease *correct* rather than merely convenient: the claim itself and the holder's liveness signal.

Today's substrate exists for this: #619 (closed) shipped the daemon-stored-resource architecture for `kind: Secret`, with storage under `/opt/kukeon/<scope>/`, scope-coordinate metadata, and the `kuke apply` / `kuke get` / `kuke delete` verb shape. `kind: Lease` is a sibling daemon-stored kind that reuses the same architecture; the load-bearing additions are atomic CAS-create semantics (Secret does not need these) and a TTL-driven expiration sweeper.

This umbrella does **not** propose the eventual integration with `kind: AgentRun` (#503) or with the cell-lifecycle death signal — those are downstream consumers / Phase C work. It proposes the standalone primitive.

## Proposal

A new daemon-stored kind sitting alongside `kind: Secret`:

```yaml
apiVersion: v1beta1
kind: Lease
metadata:
  name: claim-eminwux-kukeon-503
  realm: agents                 # scope: realm | space | stack — never cell
spec:
  key:     "eminwux/kukeon#503" # forge-agnostic claim key; unique-live within scope
  holder:  "dev-cell-7f3a"      # cell-id of the current holder
  ttl:     5m                   # expiration window; renewed by the holder
status:
  acquiredAt: 2026-05-23T14:02:11Z
  expiresAt:  2026-05-23T14:07:11Z
  state:      Held | Expired | Released
```

Storage under `/opt/kukeon/<scope>/leases/<name>` — root-owned, world-readable (no secret data; reference + metadata only). The exact scope-path layout matches the convention #619 settled for Secret.

### Load-bearing properties

[...]

PR #1153, opened 2026-06-08
feat(teamrender): render per-harness role secrets (.harnesses.<name>.secrets)

### LINKED ISSUE #1152: teamrender: support per-harness secrets (.harnesses.<name>.secrets) — Claude blueprints render no secret env var after agents #750
## Summary

The agents-repo team source (`github.com/eminwux/agents`) moved secret declarations from **role-level** (`role.spec.needs.secrets`) to **per-harness** (`role.spec.harnesses.<name>.secrets`) in [agents#750](https://github.com/eminwux/agents/issues/750) / #751, so a Claude cell no longer declares `openai-api-key`/`openrouter-api-key` it never uses. The harness blueprint templates were rewritten to iterate `{{- range .harnesses.claude.secrets }}`.

The kukeon `teamrender` engine does **not** understand this new schema. As a result, rendered Claude blueprints now contain **no `secrets:` block at all** — not even the secret the harness genuinely uses (`claude-code-oauth-token`), so the `CLAUDE_CODE_OAUTH_TOKEN` env var is silently dropped from the cell.

## Reproduction

```
kuke get blueprint kukeon-dev-claude --realm default --space default --stack default
```

The container's `env:` jumps straight to `volumes:` — there is no `secrets:` block, so `CLAUDE_CODE_OAUTH_TOKEN` is never injected. Before agents#750, the token rendered fine (the old template ranged over `.needs.secrets`, which kukeon *does* populate).

Re-running `kuke team init` does **not** fix it: the agents cache is already at the new template, so the rendered blueprint faithfully reflects a schema the renderer can't read. The gap is in the renderer, not staleness.

## Root cause

Two places only know the old role-level location:

1. **Template context** — `internal/teamrender/teamrender.go`, `renderContextValues()` builds `harnessesView` with only `{settings, sandbox, approval, permissions}`:

   ```go
   harnessesView[name] = map[string]string{
       "settings":    rh.Settings,
       "sandbox":     rh.Sandbox,
       "approval":    rh.Approval,
       "permissions": rh.Permissions,
   }
   ```

   So `{{ range .harnesses.claude.secrets }}` iterates over nothing → empty/absent `secrets:` block.

2. **Model** — `pkg/api/model/kuketeams/role.go`: `RoleHarness` has no `Secrets` field; only `RoleNeeds` does. Per-harness secrets in `role.yaml` are parsed away / ignored.

3. **CellConfig secret binding** — the #1147 bind loop iterates `r.Spec.Needs.Secrets`, which is now empty for roles that moved secrets per-harness → no `CellConfigSecretFill` is created either.

## Proposed fix (option A)

- Add `Secrets []string` to `RoleHarness` in `pkg/api/model/kuketeams/role.go` (+ parser/validation coverage).
- Expose `secrets` in the per-harness `harnessesView` map in `renderContextValues()` so `{{ range .harnesses.<name>.secrets }}` resolves.
- Update the CellConfig secret-binding step (the #1147 block) to source the declared secret slots from the per-harness list for the harness being rendered, in addition to / instead of `r.Spec.Needs.Secrets`. Keep the #1150 empty-secret pruning behavior intact.
- Decide back-compat: either keep supporting role-level `.needs.secrets` as a fallback, or treat per-harness as the sole source once agents templates have migrated. Recommend supporting both during the transition.

## Acceptance

- `kuke get blueprint kukeon-dev-claude ...` renders a `secrets:` block with `name: claude-code-oauth-token` / `envName: CLAUDE_CODE_OAUTH_TOKEN`, and the resulting CellConfig binds the fill (pruned only when the operator supplied no value).
- Claude blueprints do **not** declare `openai-api-key`/`openrouter-api-key` (the original agents#750 goal holds end-to-end).
- Unit coverage in `internal/teamrender` for per-harness secret rendering + binding.

## References

- agents#750 / #751 (per-harness secrets restructure on the template side)
- #1118 (teamrender text/template engine), #1120 (two-layer secrets), #1147 (bind role.needs.secrets), #1150 (prune empty-secret bindings)

eminwux:
## Summary

- agents#750 / #751 moved role secret declarations from role-level (`role.spec.needs.secrets`) to per-harness (`role.spec.harnesses.<name>.secrets`), and rewrote the harness blueprint templates to range `{{- range (index .harnesses .harness).secrets }}`. The kukeon `teamrender` engine didn't understand the new schema, so rendered Claude blueprints carried **no `secrets:` block at all** — silently dropping `CLAUDE_CODE_OAUTH_TOKEN`.
- **Model** (`pkg/api/model/kuketeams/role.go`): add `Secrets []string` to `RoleHarness` so per-harness secret names parse instead of being discarded.
- **Template context** (`internal/teamrender/teamrender.go`, `renderContextValues`): expose `secrets` (always a possibly-empty `[]string`) in the per-harness `harnessesView` map so `{{ range (index .harnesses .harness).secrets }}` resolves. The map value type widened from `map[string]string` to `map[string]any` to hold the slice; no other consumer asserts on the old shape (verified by grep).
- **CellConfig secret binding** (`BindConfig`): new `effectiveSecretNames(r, harness)` helper sources the declared secret names from the role's per-harness list for the harness being rendered, **merged with** role-level `needs.secrets`. Per-harness names come first; role-level names not already present are appended; dups collapse. Binding stays gated on the blueprint's declared slots, so the #1150 empty-secret pruning behavior is intact and a fallback name with no blueprint slot is not over-bound.

### Back-compat decision

Per the issue's "Decide back-compat" item and the user's explicit instruction at invocation — **keep supporting role-level `.needs.secrets` as a fallback** — both schemas bind during the transition (union, not replacement). The per-harness list is the new primary location; role-level is the fallback.

### Why the agents#750 goal holds end-to-end

Per-harness sourcing is harness-scoped: rendering the `claude` harness sources only `harnesses.claude.secrets`, so a Claude blueprint no longer declares `openai-api-key`/`openrouter-api-key` (those live under `harnesses.codex.secrets` etc.). Covered by `TestBindConfigPerHarnessSecretsAreHarnessScoped` and `TestRenderBlueprintRangesPerHarnessSecrets`.

## Test plan

- [x] `make test` (full CI target, exit 0)
- [x] `go vet ./internal/teamrender/... ./internal/kuketeams/... ./pkg/api/model/kuketeams/...`
- [x] New unit coverage in `internal/teamrender`: `TestRenderBlueprintRangesPerHarnessSecrets` (render), `TestBindConfigFillsSecretFromPerHarnessList`, `TestBindConfigPerHarnessSecretsAreHarnessScoped`, `TestBindConfigMergesPerHarnessAndRoleLevelSecrets` (bind + fallback union)
- [x] Parser coverage: `internal/kuketeams/parser_test.go` fixture + assertions for per-harness `secrets:` parsing onto `RoleHarness.Secrets`
- [ ] `make dev-init` smoke — not run: change is a pure in-process renderer/model edit that does not touch the build path, the daemon, or `/opt/kukeon`, so the dev-init smoke does not exercise it (per project CLAUDE.md, that smoke is for build-path/daemon changes)

## Acceptance

- [x] `kuke get blueprint` renders a `secrets:` block with `name: claude-code-oauth-token` / `envName: CLAUDE_CODE_OAUTH_TOKEN`, and the CellConfig binds the fill (pruned only when no value supplied — #1150 behavior intact) — verified via the render + bind unit tests
- [x] Claude blueprints do not declare `openai-api-key`/`openrouter-api-key` (harness-scoped sourcing)
- [x] Unit coverage in `internal/teamrender` for per-harness secret rendering + binding

Closes #1152

### DIFF
--- a/internal/kuketeams/parser_test.go
+++ b/internal/kuketeams/parser_test.go
@@ -69,8 +69,8 @@ metadata: { name: dev }
 spec:
   skills: [skills/, ../common/skills/]
   harnesses:
-    claude:   { settings: config/claude.settings.json }
-    codex:    { sandbox: workspace-write, approval: on-request }
+    claude:   { settings: config/claude.settings.json, secrets: [claude-code-oauth-token] }
+    codex:    { sandbox: workspace-write, approval: on-request, secrets: [openai-api-key] }
     opencode: { permissions: skip }
   needs:
     image:   [git, gh]
@@ -219,6 +219,14 @@ func TestParseHappyPathFields(t *testing.T) {
 	if len(role.Role.Spec.Needs.Repos) != 2 || len(role.Role.Spec.Needs.Mounts) != 1 {
 		t.Errorf("needs repos/mounts split wrong: repos=%v mounts=%v",
 			role.Role.Spec.Needs.Repos, role.Role.Spec.Needs.Mounts)
+	}
+	// Per-harness secrets (agents#750) parse onto RoleHarness.Secrets, distinct
+	// from the role-level needs.secrets fallback.
+	if got := role.Role.Spec.Harnesses["claude"].Secrets; len(got) != 1 || got[0] != "claude-code-oauth-token" {
+		t.Errorf("harnesses.claude.secrets = %v, want [claude-code-oauth-token]", got)
+	}
+	if got := role.Role.Spec.Harnesses["codex"].Secrets; len(got) != 1 || got[0] != "openai-api-key" {
+		t.Errorf("harnesses.codex.secrets = %v, want [openai-api-key]", got)
 	}
 
 	h, err := Parse([]byte(harnessHappy))
--- a/internal/teamrender/teamrender.go
+++ b/internal/teamrender/teamrender.go
@@ -454,12 +454,14 @@ func scopeToProject(name, project string) string {
 // slot; the agents source URL (from tc.spec.sources[<owner/repo>]) fills
 // the `agents` slot — both only when the blueprint actually declares the
 // slot, so a template that doesn't carry the slot produces a config
-// without a stray fill. role.Spec.Needs.Secrets entries are matched
-// against the blueprint's BlueprintSecretSlot declarations and, when the
-// blueprint declares the slot, filled with an in-realm ContainerSecretRef
-// pointing at the secret of the same name — the runtime resolves the
-// actual bytes via the Secret kind (#623) created out of the two-layer
-// secrets.env path (#1120).
+// without a stray fill. The secret-slot fills are sourced from the role's
+// per-harness list (role.spec.harnesses[harness].secrets, the agents#750
+// location) merged with the role-level fallback (role.spec.needs.secrets) —
+// see effectiveSecretNames. Each declared secret name is matched against the
+// blueprint's BlueprintSecretSlot declarations and, when the blueprint
+// declares the slot, filled with an in-realm ContainerSecretRef pointing at
+// the secret of the same name — the runtime resolves the actual bytes via the
+// Secret kind (#623) created out of the two-layer secrets.env path (#1120).
 func BindConfig(
 	bp *v1beta1.CellBlueprintDoc,
 	r *model.Role,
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #1153 Review — `feat(teamrender): render per-harness role secrets (.harnesses.<name>.secrets)`**

LGTM — teaches `teamrender` the agents#750 per-harness secret schema (`RoleHarness.Secrets` + `effectiveSecretNames` union with the role-level fallback) so Claude blueprints stop silently dropping `CLAUDE_CODE_OAUTH_TOKEN`; binding stays gated on declared blueprint slots (#1150 pruning intact), the `harnessesView` widening to `map[string]any` has no consumers asserting the old shape, and the merge/dedup/harness-scoping behavior is pinned by four new tests with all commits verified. The unchecked `make dev-init` smoke is correctly carved out — this is a pure in-process renderer/model edit that touches none of the build path, daemon, or `/opt/kukeon`.

PR #1155, opened 2026-06-08
fix(diff): classify secrets/workingDir/securityOpts/volumes as Breaking-on-root

### LINKED ISSUE #1154: diff: secrets (and other OCI-baked fields) misclassified as Compatible — changes never reach the running container
## Summary

`internal/controller/apply/diff.go` classifies a container's **`secrets`** change as **Compatible** (in-place) on both root and non-root containers. But env-injected secrets are baked into the container's OCI `Process.Env` at **creation** time, so an in-place reconcile never re-resolves them — the changed/added secret env var never reaches the running container. This is the root cause of the `CLAUDE_CODE_OAUTH_TOKEN`-missing symptom even after the blueprint/Config correctly declare the secret (see #1152 / #1153): the Cell model reads `Synced`, but the live containerd OCI spec lacks the env var.

This issue asks to **reclassify `secrets` as Breaking-on-root** (an OCI-baked field, like `image`/`tmpfs`/`capabilities`) so a secret change routes through `RecreateCell`, and to audit the sibling fields that have the same defect.

## Evidence (verified)

**1. secrets is classified Compatible** — `internal/controller/apply/diff.go:710-716`:
```go
// secrets — Compatible on root and non-root. Secrets either flow as
// env (re-evaluated at start) or as bind-mounted files (the
// staged-file path can be re-staged before restart); no OCI field
// is permanently baked.
if !secretsEqual(desired.Secrets, actual.Secrets) {
    recordSpecFieldChange(&result, rootContainer, false, "secrets", "secrets changed")
}
```
The comment is **incorrect**: env-form secrets ARE permanently baked.

**2. Secrets are resolved into `Process.Env` only at container create** — `internal/ctr/container.go` (`CreateContainerFromSpec` → `resolveSecrets` → `EnvAdds` appended to `containerSpec.Env`). There is no re-resolution on the task-restart / in-place `UpdateCell` reuse path.

**3. Secrets are absent from the spec-hash drift guard** — `internal/controller/runner/spec_hash.go:56-66` (`containerSpecHashPayload`) lists only `image, command, args, privileged, user, readOnlyRootFilesystem, capabilities, tmpfs, resources`. So `kuke start` of a stopped cell can't detect a secret change either (no `ErrCellSpecHashDrift`).

**4. Empirical** — `kukeon-dev-0` reads `Ready` / `Synced` with the `secrets:` block present in its Cell spec, but the live OCI process env is:
```
AGENT_NAME=dev
ANTHROPIC_MODEL=opus     # ← no CLAUDE_CODE_OAUTH_TOKEN
```
(container `createdAt` predates the secret being added; the later `start` only restarted the task.)

## Requested fix

1. In `diffContainerSpec`, classify a `secrets` change as **Breaking on root**:
   ```go
   recordSpecFieldChange(&result, rootContainer, true, "secrets", "secrets changed")
   ```
   so the apply layer drives `RecreateCell` (delete + recreate the container, rebuilding the OCI spec and re-resolving secrets).
2. Add the secret refs to `containerSpecHashPayload` in `spec_hash.go` so the spec-hash drift guard also catches it. **`TestSpecHashDomainPinsToDiffCellBreakingFields` will force this** — it fails if the Breaking domain widens without extending the hash payload (the test comment calls this out explicitly).
3. Update the now-false diff.go comment.
4. Verify the non-root `UpdateCell` stop-remove-recreate-start path actually re-resolves secrets on recreate (the diff comments claim non-root Compatible recreates the child; confirm secrets are re-staged there, or this defect persists for workload containers).
5. Consider: bind-mounted (file-form) secrets are also OCI mounts baked at create — same recreate requirement.

## Audit: other OCI-baked fields with the same defect

While here, the same "OCI-baked at create but classified Compatible (or under-guarded)" pattern affects more fields. All verified against `diff.go` + the OCI spec builder:

| Field | diff.go classification | OCI-baked at create | In spec-hash | Verdict |
|---|---|---|---|---|
| `secrets` | Compatible (`:714`, `false`) | yes — `Process.Env` (env form) / Mounts (file form) | no | **bug — this issue** |
| `workingDir` | Compatible (`:725-736`, direct) | yes — `Process.Cwd` (`oci.WithProcessCwd`) | no | **bug — change silently lost, no caveat comment** |
| `securityOpts` | Compatible (`:687-694`, `false`) | yes — `Process.NoNewPrivileges` + `Linux.Seccomp` (seccomp profile) | no | **bug — seccomp/no-new-privileges baked; comment admits "stricter classification can land in a follow-up"** |
| `volumes` | Compatible (`:649-656`, `false`) | yes — OCI Mounts (`oci.WithMounts`) | no | **known gap — comment explicitly invites "file a follow-up to widen the Breaking domain"** |

Recommend reclassifying `workingDir`, `securityOpts`, and `volumes` as Breaking-on-root in the same pass (each also added to `containerSpecHashPayload`, which the pinning test will require).

### Lower-confidence / to verify: Breaking fields missing from the spec-hash

`hostNetwork`, `hostPID`, `hostCgroup` (`diff.go:763-794`) and cell-level `nestedCgroupRuntime` are already forced **Breaking**, but are **not** in `containerSpecHashPayload`, and the pinning test's case table doesn't cover them. Changes routed through `kuke apply` recreate correctly (Breaking → `RecreateCell`); the gap is only the bare-`start`-of-stopped-cell drift guard. Worth confirming whether these should be added to the hash for parity with the test's stated "exact match" contract, or are deliberately handled by a separate path.

## Acceptance

- Changing/adding a secret on a Config and reconciling (`kuke restart` on OutOfSync, or re-apply) results in the running container's OCI `Process.Env` carrying the new `<ENVNAME>` value (verify via `ctr container info`).
- `TestSpecHashDomainPinsToDiffCellBreakingFields` passes with `secrets` (and any other reclassified fields) in both the Breaking domain and the hash payload.
- Regression test: a secrets-only diff yields `ChangeType == Breaking` on the root container and a changed `ComputeContainerSpecHash`.

## References

- #1152 / #1153 (per-harness secrets rendering — upstream of this; rendering is now correct, runtime delivery is not)
- #867, #990, #1095 (spec-hash / cell-identity / recreate-vs-in-place lineage)
- `internal/controller/apply/diff.go:615-854`, `internal/controller/runner/spec_hash.go:40-106`, `internal/controller/runner/recreate_cell.go`, `internal/ctr/container.go` (`CreateContainerFromSpec` → `resolveSecrets`)

eminwux:
## Summary

`internal/controller/apply/diff.go` classified a root container's `secrets`
change as **Compatible** (in-place). But env-form secrets are resolved into the
OCI `Process.Env` at **container create** (`ctr.CreateContainerFromSpec` →
`resolveSecrets` → `EnvAdds`) and file-form secrets become OCI Mounts — both
baked at create with no re-resolution on the in-place task-restart path. So a
secret change never reached the running container (the `CLAUDE_CODE_OAUTH_TOKEN`
-missing symptom of #1152/#1153: the Cell model read `Synced`, the live OCI spec
lacked the env var). The audit in #1154 verified the same defect for three
sibling OCI-baked fields. This PR reclassifies all four as **Breaking-on-root**
so a change routes through `RecreateCell`, and threads them through the
spec-hash drift guard and the non-root recreate gate.

### What changed

- **`diff.go`** — `secrets`, `workingDir`, `securityOpts`, `volumes` now record
  `breakingOnRoot=true`, so a root-container change classifies `Breaking` and the
  apply layer drives `RecreateCell` (which rebuilds the OCI spec and re-resolves
  secrets). Non-root stays `Compatible` (routes to `UpdateCell`). The four
  now-false caveat comments are corrected.
- **`spec_hash.go`** — `containerSpecHashPayload` gains `WorkingDir`,
  `SecurityOpts`, `Volumes`, `Secrets` (reference-only for secrets, never the
  resolved value), with projections matching diff.go's equality semantics. This
  keeps the bare-`start`-of-stopped-cell drift guard in lock-step with the
  widened Breaking domain — required by `TestSpecHashDomainPinsToDiffCellBreakingFields`.
- **`update_cell.go`** — `containerSpecChanged` (the non-root recreate gate used
  by `UpdateCell` and `UpdateContainer`) now fires on these four OCI-baked
  fields too. **This closes item 4 of the issue**: previously a secrets-only edit
  on a workload (non-root) container never recreated it, so `resolveSecrets`
  never re-ran and the defect persisted for workload containers. Verified by code
  trace: the recreate path calls `CreateContainerFromSpec` → `resolveSecrets`.

### Scope notes (for reviewer push-back)

- The issue's AC is centered on `secrets`; the title and audit section also call
  for `workingDir`/`securityOpts`/`volumes` "in the same pass" (each verified in
  the issue body). All four are the same OCI-baked defect class, so they ship
  together.
- **File-form secrets (issue item 5):** handled by the same reclassification — a
  secrets change (env or file form) now forces a recreate that rebuilds the OCI
  Mounts table; no separate change needed.
- **Deferred (documented, not fixed here):** the other already-`Breaking`-on-root
  fields (`privileged`, `user`, `readOnlyRootFilesystem`, `capabilities`,
  `tmpfs`, `resources`) are *not* yet in the non-root `containerSpecChanged`
  gate — their non-root in-place gap predates #1154 and is out of this issue's
  scope. Likewise the lower-confidence host-namespace fields (`hostNetwork`,
  `hostPID`, `hostCgroup`, `nestedCgroupRuntime`) the issue flags as "to verify"
  remain `Breaking` but absent from the spec-hash; I left them as-is rather than
  bundle an uncertain change. Both are worth a PM follow-up.

## Test plan

- [x] `make test` (full CI suite: `test-root` + `test-kukebuild`) — exit 0
- [x] `go vet` on `./internal/controller/apply/...` and `./internal/controller/runner/...` (GOWORK=off) — clean
- [x] `TestSpecHashDomainPinsToDiffCellBreakingFields` passes with the four fields in both the Breaking domain and the hash payload
- [x] New `TestDiffCell_RootContainerSecrets_Breaking` + `TestDiffCell_RootContainerWorkingDirVolumesSecurityOpts_Breaking`: a secrets-only (and each sibling) root diff yields `ChangeType == Breaking` on the root container
- [x] New `TestContainerSpecChanged_OCIBakedFields` / `_SecretRefScopeChange`: the non-root recreate gate fires for each OCI-baked field (item 4)
- [ ] Live `ctr container info` env verification on a running cell — not run; requires a provisioned daemon/host the dev environment doesn't have. The runtime-delivery path is covered by code trace (recreate → `CreateContainerFromSpec` → `resolveSecrets`) plus the unit tests above.

Closes #1154

### DIFF
--- a/internal/controller/apply/diff.go
+++ b/internal/controller/apply/diff.go
@@ -646,13 +646,13 @@ func diffContainerSpec(desired, actual *intmodel.ContainerSpec, rootContainer bo
 		recordSpecFieldChange(&result, rootContainer, false, "ports", "ports changed")
 	}
 
-	// volumes — Compatible on root and non-root. Root-container volume
-	// edits are treated as in-place updateable for now (the apply layer
-	// reports drift so the operator sees it; the runner's UpdateCell
-	// path applies the change on the next reconcile). If runner-side
-	// gaps surface, file a follow-up to widen the Breaking domain.
+	// volumes — Breaking on root (OCI Mounts table is baked into the
+	// cell root's runtime spec at StartCell; a volume edit only reaches
+	// the container by rebuilding the spec via RecreateCell). Compatible
+	// on non-root, where UpdateCell stop-removes and recreates the child
+	// with the new mounts. Issue #1154.
 	if !volumeMountsEqual(desired.Volumes, actual.Volumes) {
-		recordSpecFieldChange(&result, rootContainer, false, "volumes", "volumes changed")
+		recordSpecFieldChange(&result, rootContainer, true, "volumes", "volumes changed")
 	}
 
 	// privileged — Breaking on root (the cap-set is baked into the cell
@@ -684,13 +684,13 @@ func diffContainerSpec(desired, actual *intmodel.ContainerSpec, rootContainer bo
 		recordSpecFieldChange(&result, rootContainer, true, "capabilities", "capabilities changed")
 	}
 
-	// securityOpts — Compatible on root and non-root. Most security-opt
-	// values (selinux/apparmor/seccomp profile names) round-trip through
-	// the runner without a baked OCI field that demands a fresh
-	// container; a stricter classification can land in a follow-up if
-	// runner gaps surface.
+	// securityOpts — Breaking on root. Security-opt values bake into the
+	// cell root's OCI Process at StartCell (no-new-privileges →
+	// Process.NoNewPrivileges, seccomp=… → Linux.Seccomp), so a change
+	// only reaches the running container via RecreateCell. Compatible on
+	// non-root, where UpdateCell recreates the child. Issue #1154.
 	if !slicesEqual(desired.SecurityOpts, actual.SecurityOpts) {
-		recordSpecFieldChange(&result, rootContainer, false, "securityOpts", "securityOpts changed")
+		recordSpecFieldChange(&result, rootContainer, true, "securityOpts", "securityOpts changed")
 	}
 
 	// tmpfs — Breaking on root (OCI Mounts table is fixed at create).
@@ -707,12 +707,17 @@ func diffContainerSpec(desired, actual *intmodel.ContainerSpec, rootContainer bo
 		recordSpecFieldChange(&result, rootContainer, true, "resources", "resource limits changed")
 	}
 
-	// secrets — Compatible on root and non-root. Secrets either flow as
-	// env (re-evaluated at start) or as bind-mounted files (the
-	// staged-file path can be re-staged before restart); no OCI field
-	// is permanently baked.
+	// secrets — Breaking on root. Env-form secrets are resolved into the
+	// OCI Process.Env at container create (ctr.CreateContainerFromSpec →
+	// resolveSecrets → EnvAdds), and file-form secrets become OCI Mounts;
+	// both are baked at create with no re-resolution on the in-place
+	// task-restart path, so a secret change only reaches the running
+	// container via RecreateCell (delete + recreate, rebuilding the spec).
+	// Compatible on non-root, where UpdateCell stop-removes and recreates
+	// the child — which re-runs resolveSecrets (see containerSpecChanged
+	// in update_cell.go, extended to gate on secrets). Issue #1154.
 	if !secretsEqual(desired.Secrets, actual.Secrets) {
-		recordSpecFieldChange(&result, rootContainer, false, "secrets", "secrets changed")
+		recordSpecFieldChange(&result, rootContainer, true, "secrets", "secrets changed")
 	}
 
 	// repos — Compatible on root and non-root. Repos are handled by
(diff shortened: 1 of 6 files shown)


eminwux:
**PR #1155 Review — `fix(diff): classify secrets/workingDir/securityOpts/volumes as Breaking-on-root`**

LGTM — reclassifies four OCI-baked fields as Breaking-on-root so changes route through `RecreateCell` (re-resolving secrets), extends the non-root recreate gate, and pins the spec-hash domain in lockstep; the runner equality mirrors match the apply layer byte-for-byte, struct comparability holds, no dead code, and the only unchecked test-plan item is the documented host-gated `ctr container info` env check (worth a quick live confirm pre-merge).

PR #1157, opened 2026-06-08
fix(cli): default single-get and run --from-config space/stack to default scope

### LINKED ISSUE #1156: cli: get config/blueprint and run --from-config don't default space/stack to "default"
## Context

`kuke get config <name>`, `kuke get blueprint <name>`, and `kuke run --from-config <cfg>` fail to find resources that plainly exist when the operator omits `--space`/`--stack`. They default the **realm** to `default` but leave **space** and **stack** empty, so the daemon's exact-scope lookup misses any resource stored at the full `default/default/default` coordinate — which is how `kuke create config`, `kuke create blueprint`, and the team renderer actually store them.

Reproduction (resource exists at `realm=default space=default stack=default`):

```
$ kuke run --from-config kukeon-dev-claude --name kukeon-dev-2
Error: config not found (config "kukeon-dev-claude" in scope realm="default" space="" stack="")

$ kuke get config kukeon-dev-claude
Error: config "kukeon-dev-claude" not found
$ kuke get config kukeon-dev-claude --realm default --space default --stack default
apiVersion: v1beta1
kind: CellConfig
metadata:
  name: kukeon-dev-claude
  realm: default
  space: default
  stack: default
  ...

$ kuke get blueprint                         # list path works
NAME               REALM    SPACE    STACK
kukeon-dev-claude  default  default  default
$ kuke get blueprint kukeon-dev-claude       # single-get fails
Error: blueprint "kukeon-dev-claude" not found
```

The **list** path works because it passes empty coordinates to `ListConfigs`/`ListBlueprints`, which the backend treats as a "no filter / list subtree" wildcard. The **single-get** and **run --from-config** paths build an *exact-scope* lookup, where empty space/stack mean literally empty — not `default`.

## Root cause

The empty space/stack is **intentional and test-pinned**, not an accidental missing default:

- `cmd/kuke/shared/scope.go` — `PickLookupRealm` defaults realm to `default`, but `ExplicitScope` (used for space/stack on the `run`/`apply` lookup path) returns `""` unless the operator set the flag/env explicitly. Its doc comment: *"a missing flag must not narrow the lookup to space=default / stack=default or it would hide realm-scoped resources."* Pinned by `cmd/kuke/shared/scope_test.go`.
- `cmd/kuke/get/config/config.go:76-95` and `cmd/kuke/get/blueprint/blueprint.go` (single-get branch) replicate the same choice: realm gets `ValueOrDefault()`, but the comment says *"deeper coordinates stay unset unless provided"* and `Space`/`Stack` are passed through empty.
- `cmd/config/env.go` — `KUKE_GET_{CONFIG,BLUEPRINT}_{SPACE,STACK}` are declared with **no** default, while `KUKE_GET_CELL_{SPACE,STACK}` and `KUKE_RUN_{SPACE,STACK}` carry `"default"`.

The design goal was to keep **realm-scoped** Blueprints/Configs (space/stack unset) findable with just `--realm`. But in practice resources are stored **full-scoped** (`default/default/default`), so the common case breaks while the realm-scoped cohort it protects is rarely (if ever) created. This is also **internally inconsistent**: `kuke get cell <name>` (`cmd/kuke/get/cell/cell.go:112-120`) defaults all three coordinates, so `get cell` behaves the way the operator expects but `get config` / `get blueprint` / `run --from-config` do not.

## Proposal

Make the no-flag single-resource lookup resolve to the operator's full default scope (`default/default/default`) so `kuke get config <name>`, `kuke get blueprint <name>`, and `kuke run --from-config <cfg>` find full-scoped resources without forcing the operator to type `--space default --stack default` — **without** regressing realm-scoped findability.

This is a design call for dev (the current behavior is deliberate); the issue is to pick a resolution, not to prescribe one. Candidate approaches:

1. **Default all three coordinates** (match `get cell`): omitted space/stack resolve to `default`; realm-scoped resources then require an explicit empty `--space ""`/`--stack ""`. Simplest, most consistent with `get cell`, but flips the realm-scoped default.
2. **Daemon-side fallback lookup**: try the exact full-default scope first, fall back to the realm-scoped record on miss (or vice versa). Preserves both cohorts; more lookup logic.
3. **Prefix/narrowing match** on the daemon get path so a name unique within the realm resolves regardless of space/stack. Most forgiving; needs an ambiguity error when a name collides across scopes.

Whichever is chosen, keep the realm-scoped-resource use case working or explicitly document its removal, and update `scope_test.go` / the `ExplicitScope` contract accordingly.

## Acceptance criteria

- [ ] `kuke get config <name>` and `kuke get blueprint <name>` (no `--space`/`--stack`) find a resource stored at `realm=default space=default stack=default`.
- [ ] `kuke run --from-config <cfg> --name <n>` (no `--space`/`--stack`) finds a config stored at the full default scope and proceeds to create+start.
- [ ] Behavior is consistent across `get cell`, `get config`, `get blueprint`, and the `run`/`create --from-config`/`--from-blueprint` lookup paths.
- [ ] Realm-scoped Blueprints/Configs (space/stack unset) remain findable, or their removal is explicitly documented and the `ExplicitScope` contract + `scope_test.go` updated to match.
- [ ] A "not found" error for a name that *does* exist in another scope under the same realm surfaces the scope it searched (it already does) and, where applicable, hints that the resource exists at a different coordinate.
- [ ] Tests cover the no-flag full-default-scope hit for `get config`, `get blueprint`, and `run --from-config`.

## Notes

- Reported via daily agent-cell workflow: `kuke run --from-config kukeon-dev-claude --name kukeon-dev-2` and `kuke get config kukeon-dev-claude`.
- Relevant code: `cmd/kuke/shared/scope.go` (`PickLookupRealm`/`ExplicitScope`), `cmd/kuke/get/config/config.go`, `cmd/kuke/get/blueprint/blueprint.go`, `cmd/kuke/get/cell/cell.go` (correct precedent), `cmd/kuke/create/cell/cell.go:469-480` (run/--from-config lookup + error string), `cmd/config/env.go` (`KUKE_GET_*` default registration).
- The `run --from-config` path defaults the *materialised cell's* location correctly via `resolveCellLocation`/`pickLocation` (`cmd/kuke/run/run.go:829-851`); only the *config lookup* scope is under-defaulted — the two resolution paths diverge.

eminwux:
## Summary

`kuke get config <name>`, `kuke get blueprint <name>`, and `kuke run --from-config <cfg>` defaulted only the **realm** to `default`, leaving **space**/**stack** empty. Since `kuke create config/blueprint` and the team renderer store resources at the full `default/default/default` coordinate, the no-flag single-resource lookup missed resources that plainly exist — while the list path (wildcard semantics) worked.

**Design call (the issue is explicitly a "pick a resolution" for dev):** chose **Approach 1** — default space/stack to `"default"` when not explicitly provided, mirroring the `get cell` precedent the issue names as correct. This is lower-blast-radius than daemon-side fallback (Approach 2) or prefix/narrowing match (Approach 3): no daemon changes, no new lookup logic. The create path already proves realm-scoped resources are effectively never created (it full-scopes them via `KUKE_CREATE_{CONFIG,BLUEPRINT}_{SPACE,STACK}` defaults), so the realm-scoped cohort the old behavior protected is vestigial.

**Realm-scoped findability is preserved, not removed:** the run/create lookup path (`ExplicitScope`) still honors an explicit empty `--space "" --stack ""` as empty, so a realm-scoped Blueprint/Config stays reachable. Only the *no-flag* default changed. The get paths mirror `get cell` exactly (an explicit empty re-defaults, matching the named precedent); the run/create path retains the explicit-empty escape hatch where `ExplicitScope`'s design already provides it for free.

### Changes
- `cmd/kuke/shared/scope.go` — `ExplicitScope` now falls back to `kv.Default` (`"default"` for the run/create-cell scope vars) when neither flag nor env is set, while still honoring an explicit empty value. Contract doc rewritten.
- `cmd/kuke/get/config/config.go`, `cmd/kuke/get/blueprint/blueprint.go` — single-get branch defaults space/stack via `ValueOrDefault()` (mirrors `get cell`); not-found errors now surface the searched scope and best-effort hint the coordinate where a same-named resource does live (AC bullet 5).
- `cmd/config/env.go` — `KUKE_GET_{CONFIG,BLUEPRINT}_{SPACE,STACK}` now carry the `"default"` default (matching `KUKE_GET_CELL_{SPACE,STACK}`).
- `cmd/kuke/shared/scope_test.go` — updated the `ExplicitScope` contract tests; added explicit-empty escape-hatch cases.
- Stale comments in `create/cell/cell.go`, `create/cell/clone.go`, and `internal/controller/reconcile_outofsync.go` updated to the new defaulting contract (no behavior change in the daemon — the lineage-widening logic is untouched).

### Notes for the reviewer
- **Out of AC scope, flagged for PM:** `kuke get secret <name>` (`cmd/kuke/get/secret/secret.go:79`) has the identical under-defaulting pattern, but the issue's AC enumerates only config/blueprint/run-create and consistency with `get cell` — and secrets add a `--cell` coordinate, so whether they should default the same way is a separate call. Left untouched; recommend a follow-up if PM wants the parity.
- The daemon-side `lookupLineageConfig` progressive-widening was **not** changed — it remains the backstop for cells whose lineage label points to a shallower-scoped (e.g. team-rendered or explicit-empty) Config.

## Test plan
- [x] `make test` (full CI suite: `test-root` + `test-kukebuild`) — passes (`GOWORK=off`; the bare `go build ./...` cgroups break is a `go.work`-only artifact the Makefile env avoids).
- [x] `GOWORK=off go test ./cmd/config/... ./cmd/kuke/shared/... ./cmd/kuke/get/config/... ./cmd/kuke/get/blueprint/... ./cmd/kuke/create/cell/... ./cmd/kuke/run/...` — passes.
- [x] `go vet` + `gofmt -l` clean on changed files; `golangci-lint run --new-from-rev HEAD` reports 0 new issues; pre-commit (changed files) all green.
- [x] New tests cover the no-flag full-default-scope hit for `get config`, `get blueprint`, and `run/create --from-config`, plus the explicit-empty escape hatch and the other-scope hint (AC bullet 6).

Closes #1156

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -302,17 +302,17 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_GET_BLUEPRINT_REALM = DefineKV("KUKE_GET_BLUEPRINT_REALM", "kuke/get/blueprint/realm", "default")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
-	KUKE_GET_BLUEPRINT_SPACE = DefineKV("KUKE_GET_BLUEPRINT_SPACE", "kuke/get/blueprint/space")
+	KUKE_GET_BLUEPRINT_SPACE = DefineKV("KUKE_GET_BLUEPRINT_SPACE", "kuke/get/blueprint/space", "default")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
-	KUKE_GET_BLUEPRINT_STACK = DefineKV("KUKE_GET_BLUEPRINT_STACK", "kuke/get/blueprint/stack")
+	KUKE_GET_BLUEPRINT_STACK = DefineKV("KUKE_GET_BLUEPRINT_STACK", "kuke/get/blueprint/stack", "default")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_GET_CONFIG_NAME = DefineKV("KUKE_GET_CONFIG_NAME", "kuke/get/config/name")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_GET_CONFIG_REALM = DefineKV("KUKE_GET_CONFIG_REALM", "kuke/get/config/realm", "default")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
-	KUKE_GET_CONFIG_SPACE = DefineKV("KUKE_GET_CONFIG_SPACE", "kuke/get/config/space")
+	KUKE_GET_CONFIG_SPACE = DefineKV("KUKE_GET_CONFIG_SPACE", "kuke/get/config/space", "default")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
-	KUKE_GET_CONFIG_STACK = DefineKV("KUKE_GET_CONFIG_STACK", "kuke/get/config/stack")
+	KUKE_GET_CONFIG_STACK = DefineKV("KUKE_GET_CONFIG_STACK", "kuke/get/config/stack", "default")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	KUKE_GET_OUTPUT = DefineKV("KUKE_GET_OUTPUT", "kuke/get/output")
 
--- a/cmd/kuke/create/cell/cell.go
+++ b/cmd/kuke/create/cell/cell.go
@@ -338,9 +338,11 @@ func ValidateOverrideSymmetry(flags SourceFlags) error {
 }
 
 // ScopeVars names the command-specific scope viper Vars the materialize path
-// reads to resolve binding-lookup scope (so a realm-scoped Blueprint/Config
-// stays findable when --space/--stack are unset). `kuke create cell` passes its
-// KUKE_CREATE_CELL_* vars; `kuke run` passes its KUKE_RUN_* vars. Threading the
+// reads to resolve binding-lookup scope. An unset --space/--stack defaults to
+// the var's "default" (issue #1156) so the lookup hits the full default scope;
+// a realm-scoped Blueprint/Config stays reachable via an explicit empty
+// `--space "" --stack ""`. `kuke create cell` passes its KUKE_CREATE_CELL_*
+// vars; `kuke run` passes its KUKE_RUN_* vars. Threading the
 // Vars in (rather than hard-coding KUKE_CREATE_CELL_*) is what lets the two
 // verbs share one Materialize entrypoint without colliding on a global viper
 // key (epic:cell-identity #1025).
@@ -404,9 +406,10 @@ func Materialize(
 // --param/--param-file, materialises the Cell record, overlays scope, and
 // finalizes the cell name — returning the doc without persisting it.
 //
-// Lookup scope follows the explicit-coordinate rule from
-// kukeshared.PickLookupRealm / ExplicitScope so realm-scoped Blueprints stay
-// findable when --space/--stack are not set.
+// Lookup scope follows the default-coordinate rule from
+// kukeshared.PickLookupRealm / ExplicitScope: an unset --space/--stack
+// resolves to "default" so the lookup hits the full default scope, while a
+// realm-scoped Blueprint stays reachable via an explicit empty --space/--stack.
 func materializeFromBlueprint(
 	cmd *cobra.Command, client kukeonv1.Client, flags SourceFlags, scope ScopeVars,
 ) (v1beta1.CellDoc, error) {
--- a/cmd/kuke/create/cell/cell_test.go
+++ b/cmd/kuke/create/cell/cell_test.go
@@ -755,6 +755,87 @@ func TestCreateCell_FromConfig_HappyPath(t *testing.T) {
 	}
 }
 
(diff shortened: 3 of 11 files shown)


eminwux:
**PR #1157 Review — `fix(cli): default single-get and run --from-config space/stack to default scope`**

LGTM — defaults the no-flag single-get and run/create binding lookup to the full `default/default/default` scope (mirroring the `get cell` precedent), so resources stored at the full default coordinate by `kuke create config/blueprint` and the team renderer are actually found.

Verified the change is safe:
- `ExplicitScope`'s new `kv.Default` fallback resolves to `"default"` for every caller — the only callers (`cell.go`) pass `KUKE_{RUN,CREATE_CELL}_{SPACE,STACK}`, all of which carry the `"default"` default. No caller passes an empty-default var.
- `config/blueprintNotFoundErr` are wired into both miss branches and probe realm-wide via `List…(ctx, realm, "", "")` — correct wildcard semantics; list errors swallowed best-effort.
- Error-wrapping is unchanged from before (the prior path didn't `%w`-wrap the sentinel either — no regression).
- Realm-scoped findability is preserved via the explicit-empty escape hatch + the untouched daemon-side lineage widening; comments in `clone.go`/`reconcile_outofsync.go` updated to match.
- Tests cover the full-default hit, explicit-empty escape hatch, and other-scope hint across get config / get blueprint / run --from-config. CI green (`make test` + `make e2e`).

The `kuke get secret <name>` under-defaulting is correctly deferred to PM in the body — it carries an extra `--cell` coordinate, so its defaulting is a genuine separate call. Nothing blocking.

PR #1159, opened 2026-06-08
fix(dev-init): build attach-smoke image locally instead of external amd64-only busybox

### LINKED ISSUE #1158: dev-init attach smoke depends on external, amd64-only registry.eminwux.com/busybox; build a local image instead
## Summary

`make dev-init`'s attach smoke (the `dev-init-attach/ds/dks/cattach` cell) pins its containers to `registry.eminwux.com/busybox:latest`. That image is **single-arch amd64**, so on an `aarch64`/arm64 host the cell fails to start. It also makes the dev loop depend on an **external registry** that is outside the kukeon repo — contrary to the `kukeon.internal/...` "build locally, never pull from outside" convention the rest of dev-init already follows for `kukeond`.

## Observed failure (arm64 host)

```
Cell "cattach": failed
  Error: failed to start cell after creation: failed to attach root container
  ds_dks_cattach_root to network: plugin type="bridge" failed (add):
  failed to open netns "/proc/56258/ns/net": failed to Statfs
  "/proc/56258/ns/net": no such file or directory
```

### Root cause

The `cattach` root container (`scripts/dev-init.sh:525`) runs `sleep 3600` from `registry.eminwux.com/busybox:latest`:

```yaml
- id: root
  root: true
  image: registry.eminwux.com/busybox:latest   # single-manifest, architecture: amd64
  command: sleep
  args: ["3600"]
```

Inspecting the pulled image:

| ref | manifest | arch |
| --- | --- | --- |
| `registry.eminwux.com/busybox:latest` | single (not an index) | `amd64` / `linux` |
| host | — | `aarch64` |

The amd64 `sleep` ELF cannot exec on arm64, so runc assigns the task a PID and the process **dies immediately**. The root-container start path (`internal/controller/runner/start.go:786-907`) reads `rootTask.Pid()` and calls CNI ADD with **no wait for RUNNING and no netns existence check**, so the bridge plugin's `Statfs("/proc/<pid>/ns/net")` hits ENOENT. The netns error is just the late symptom of a dead foreign-arch process. Only `cattach` is affected because it is the one dev-init cell whose root container uses an explicit image instead of the native-arch bind-mounted `kukepause`.

## Proposed fix

Stop relying on the external `registry.eminwux.com/busybox`. Mirror what dev-init already does for `kukeond`:

1. **Add a Dockerfile to the repo** for a minimal busybox-equivalent attach-smoke image (e.g. `hack/attach-smoke/Dockerfile` or `test/fixtures/busybox/Dockerfile`). Base it on a multi-arch source (`docker.io/library/busybox:latest`) or build a tiny static `sleep`/shell so the resulting image matches the build host's arch automatically.
2. **Build it in `make dev-init`** with `kuke build` into the attach realm, using the existing local-only naming scheme — e.g.:
   ```bash
   sudo ./kuke build -t kukeon.internal/attach-smoke:v0.0.0-dev --realm dev-init-attach <context>
   ```
   placed after the `dev-init-attach` realm is created (`scripts/dev-init.sh:508`), analogous to the `kukeon.internal/kukeond` build at `scripts/dev-init.sh:329`. Note `scripts/dev-init.sh:483` currently asserts the attach realm is *never* a `kuke build` target — that comment/teardown lane (#904) needs updating.
3. **Point the cell.yaml at the local ref** (`scripts/dev-init.sh:525` and `:529`): `image: kukeon.internal/attach-smoke:v0.0.0-dev`. The `kukeon.internal` reserved TLD guarantees a stray pull fails fast instead of silently hitting an external host.

### Outcome

- dev-init's attach smoke works on both amd64 and arm64 hosts (image is built for the host arch).
- No dependency on `registry.eminwux.com` or any external registry; the Dockerfile lives in-repo.
- Consistent with the `kukeon.internal/` local-build convention documented in `CLAUDE.md` and used for `kukeond`.

## Optional follow-up (separate issue)

`internal/controller/runner/start.go` could wait for the root task to reach RUNNING (or stat the netns) before calling CNI ADD, so a dead root container surfaces as "root container exited" rather than an opaque netns ENOENT. Not required to fix this issue, but would make the failure mode legible.

eminwux:
## Summary

- The `make dev-init` attach smoke (`dev-init-attach/ds/dks/cattach`) pinned both cell containers to `registry.eminwux.com/busybox:latest`, a **single-arch amd64** image. On arm64 hosts the amd64 `sleep` ELF can't exec, the root task dies immediately, and the CNI bridge plugin's `Statfs("/proc/<pid>/ns/net")` surfaces an opaque netns ENOENT (issue #1158). It also made the dev loop depend on an external registry, contrary to the `kukeon.internal/...` "build locally, never pull from outside" convention dev-init already follows for `kukeond`.
- Mirror the `kukeon.internal/kukeond` approach: add an in-repo `hack/attach-smoke/Dockerfile` (based on multi-arch `docker.io/library/busybox`), build it into the `dev-init-attach` realm via `kuke build` as `kukeon.internal/attach-smoke:v0.0.0-dev`, and point the `cattach` cell.yaml at that local ref. `kuke build` resolves the build host's architecture, so the resulting image is host-arch on both amd64 and arm64.

## Implementation notes

- **Dockerfile base choice (build-vs-flip):** the issue offered two options — base on `docker.io/library/busybox:latest` or build a tiny static `sleep`/shell. I chose the busybox base (lower blast radius): it preserves the exact prior runtime behavior (real `sleep` for the root/work containers, real `/bin/sh` that the attachable container's kuketty/sbsh wrapper needs), where a `FROM scratch` static `sleep` would drop the shell and risk the attach path. The build-time base pull from `docker.io` is consistent with the root `kukeond` Dockerfile (which pulls `golang`/`debian` bases from `docker.io`); the *runtime* dependency on the external registry is what's removed — the cell now pulls the local `kukeon.internal/attach-smoke` ref.
- **#904 teardown lane:** the issue flagged that `scripts/dev-init.sh`'s teardown comment asserted the attach realm is *never* a `kuke build` target. That's now false — `teardown_attach_smoke_state` wipes the per-namespace BuildKit cache at `/var/lib/kukebuild/<ns>/` alongside the containerd namespace so the teardown stays idempotent across runs, and the comment is updated.
- **Optional follow-up not included:** the issue's "Optional follow-up (separate issue)" — making `internal/controller/runner/start.go` wait for the root task RUNNING / stat the netns before CNI ADD — is explicitly out of scope here.

## Test plan

- [x] `make test` (test-root + test-kukebuild) — exit 0.
- [x] `make dev-init` end-to-end (nested `kukeon-dev-0` cell, amd64 host) — exit 0. Confirmed: `kukeon.internal/attach-smoke:v0.0.0-dev` builds from `docker.io/library/busybox:latest` into the `dev-init-attach` realm, `Cell "cattach": created`, `kuke attach exited cleanly`, daemon-parity tail and parent-attach plumbing intact.
- [ ] arm64-host exec of the foreign-arch path — not runnable on this amd64 dev host. The original bug is amd64-`sleep`-can't-exec-on-arm64; the fix makes `kuke build` produce a host-arch image, so on arm64 it builds an arm64 image. The local-build + cell-start path (the entirety of the diff) is validated on amd64 above; the arch-specific exec failure can only reproduce on an arm64 host (or CI arm64 runner).
- [x] Tree-wide `git grep registry.eminwux.com` — the only remaining hits are unit-test fixtures / docs examples that exercise spec-parsing (no runtime pulls) and my own explanatory comments; the dev-init attach smoke (the issue's scope) no longer references the external registry.

Closes #1158

### DIFF
--- a/hack/attach-smoke/Dockerfile
+++ b/hack/attach-smoke/Dockerfile
@@ -1,0 +1,16 @@
+# Minimal busybox image for dev-init's attach smoke (the
+# dev-init-attach/ds/dks/cattach cell). Built locally into the attach realm
+# by scripts/dev-init.sh and tagged kukeon.internal/attach-smoke, mirroring
+# the kukeon.internal/kukeond local-build convention.
+#
+# Based on the multi-arch docker.io/library/busybox so `kuke build` resolves
+# the build host's architecture and produces a host-arch image — fixing the
+# arm64 breakage from the former single-arch-amd64 registry.eminwux.com/busybox
+# pin, whose amd64 `sleep` ELF could not exec on arm64 and surfaced as an
+# opaque CNI netns ENOENT (issue #1158). Building locally also drops the dev
+# loop's dependency on the external registry.eminwux.com; the reserved
+# kukeon.internal TLD makes a stray runtime pull fail fast instead of
+# silently hitting an external host. busybox provides the /bin/sh the
+# attachable container's kuketty/sbsh wrapper needs plus the `sleep` the
+# root and work containers run.
+FROM docker.io/library/busybox:latest
--- a/scripts/dev-init.sh
+++ b/scripts/dev-init.sh
@@ -456,6 +456,11 @@ ATTACH_SMOKE_SPACE="ds"
 ATTACH_SMOKE_STACK="dks"
 ATTACH_SMOKE_CELL="cattach"
 ATTACH_SMOKE_CONTAINER="work"
+# Local-only attach-smoke image ref, mirroring KUKEOND_IMAGE_REF above. Built
+# from hack/attach-smoke/Dockerfile into the attach realm below so the smoke
+# no longer pins the single-arch-amd64 external registry.eminwux.com/busybox
+# that broke on arm64 hosts (#1158).
+ATTACH_SMOKE_IMAGE_REF="kukeon.internal/attach-smoke:${KUKEOND_VERSION}"
 ATTACH_SMOKE_BASE="${METADATA_ROOT}/${ATTACH_SMOKE_REALM}/${ATTACH_SMOKE_SPACE}/${ATTACH_SMOKE_STACK}/${ATTACH_SMOKE_CELL}/${ATTACH_SMOKE_CONTAINER}"
 ATTACH_SMOKE_SOCKET="${ATTACH_SMOKE_BASE}/tty/socket"
 ATTACH_SMOKE_METADATA="${ATTACH_SMOKE_BASE}/kuketty-metadata.json"
@@ -480,11 +485,13 @@ teardown_attach_smoke_state() {
     # the purge and strands the next run's `kuke apply` image-pull→
     # snapshot-unpack with "parent snapshot ... does not exist". Removing
     # the namespace by hand makes the teardown idempotent across N runs.
-    # The attach smoke never targets this realm with `kuke build`, so no
-    # `/var/lib/kukebuild/<ns>/` cache exists to wipe alongside it
-    # (the #904 lane only applies to realms that have been a `kuke build`
-    # target).
     sudo ctr namespaces remove "${ATTACH_SMOKE_NS}" 2>/dev/null || true
+    # Since #1158 the attach smoke *is* a `kuke build` target (it builds
+    # kukeon.internal/attach-smoke into this realm), so the #904 lane now
+    # applies: wipe the per-namespace BuildKit cache at
+    # /var/lib/kukebuild/<ns>/ alongside the containerd namespace so the
+    # teardown stays idempotent across runs.
+    sudo rm -rf "/var/lib/kukebuild/${ATTACH_SMOKE_NS}" 2>/dev/null || true
 }
 
 cleanup_attach_smoke() {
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #1159 Review — `fix(dev-init): build attach-smoke image locally instead of external amd64-only busybox`**

LGTM — replaces the single-arch-amd64 `registry.eminwux.com/busybox` pin with a local `kuke build` of `kukeon.internal/attach-smoke`, mirroring the existing `kukeon.internal/kukeond` convention so the attach smoke is host-arch on arm64 (#1158) and no longer depends on the external registry. Build step is correctly ordered (realm/space/stack → build → `kuke apply`), the new `ATTACH_SMOKE_IMAGE_REF` is used at all four sites, and the teardown correctly extends the #904 `/var/lib/kukebuild/<ns>/` cache-reclaim invariant to the newly-built realm rather than fragmenting it.

The unchecked **arm64-host exec** test-plan item is the host-gated carve-out — the bug only reproduces on an arm64 host/CI runner and is documented as unrunnable on the amd64 dev host; the entirety of the diff (local build + cell start) is validated on amd64. Worth running on an arm64 runner pre-merge to close the loop, but non-blocking.

PR #1172, opened 2026-06-09
fix(run): heal wrong-mode attachable tty socket on the attach path

### LINKED ISSUE #1169: fix(run): kuke run on a Ready cell attaches without healing socket perms — a wrong-mode tty socket permanently blocks attach (restart heals, run does not)
## Summary

`kuke run <cell>` against an already-**Ready** cell short-circuits straight to attach without going through any `StartCell` step, so it attaches to whatever mode the live per-container tty control socket currently has. If that socket is at a group-non-writable mode (e.g. `0640`), every `kuke run` fails with `connect: permission denied` (EACCES) and **leaves the socket untouched**. `kuke restart` on the same cell heals it, because restart is stop+start and the start step rebinds the socket fresh at `0660`.

This is the still-open tail of #935/#936: that fix added `reapplyAttachableSocketPerms` to the **idempotent-skip path of `runner.StartCell`**, but `kuke run`'s Ready branch never reaches `StartCell` at all, so the heal never runs on the path `run` actually takes.

## Repro

1. Have a Ready cell with an attachable container whose live tty socket is at `0640` (group lacks write).
2. `kuke run <cell>` →
   ```
   Error: attach stale socket: kuketty has not yet unlinked the pre-fix tty socket:
   error attaching client: ping failed:
   dial unix /opt/kukeon/s/<shortid>: connect: permission denied
   ```
   Socket remains `0640` afterward.
3. `kuke restart <cell>` → connects cleanly; socket is now `0660`.

Confirmed by reproduction: forcing a test cell's socket to `0640` reproduces the `run` failure identically and leaves it `0640`; `restart` heals it to `0660`. As root the same socket connects (it has a live listener — it is **not** stale), proving the failure is purely the group-write permission bit, not a dead socket.

## Root cause

`cmd/kuke/run/run.go` — `runExistingCell`, `CellStateReady` branch (~line 628-684): calls `GuardCellTaskLiveness` + `printRunResult(noOpResultFromGet(pre))` and then attaches, **without** calling `StartCell`. The `CellStateStopped` branch does call `StartCell` (and thus the #936 heal). So the heal coverage is asymmetric: stopped→start→attach heals, ready→attach does not.

`controller.StartCell` (`internal/controller/start_cell.go`) deliberately refuses a running cell, and the Ready short-circuit exists to avoid the #630 CNI duplicate-allocation problem — so the fix is **not** "always call StartCell". The socket heal needs to live on the attach path itself.

## Proposed fix

Wire the existing `reapplyAttachableSocketPerms` (`internal/controller/runner/attachable.go:622`, mode `0o660`) into the path `run`/`attach` actually take — most naturally inside the daemon's `AttachContainer` RPC, so a live wrong-mode socket heals in place (chmod the present inode) without bouncing the workload. That makes `kuke run` self-healing for this class and keeps it consistent with `kuke restart` without reintroducing the #630 start-on-running hazard.

## Related

- #935 / #936 — added the heal to `runner.StartCell`'s idempotent-skip path (the path `run` Ready does not reach).
- #933, #916, #926 — adjacent attach EACCES / readiness-handshake fixes.
- Companion bug: `attachBootRaceSentinel` misclassifies this permanent EACCES as a transient race and burns the 10 s retry budget with a misleading "stale socket" message (filed separately).

eminwux:
## Summary

- `kuke run` against an already-**Ready** cell short-circuits straight to attach without re-entering `StartCell` (the Ready branch deliberately avoids the #630 CNI duplicate-allocation re-entry). The #935/#936 heal that re-chmods a wrong-mode live tty socket was wired only into `runner.StartCell`'s idempotent no-op path — so it never runs on the path `run`/`attach` actually take. A socket an old kuketty bound `0o640` (group-read only, pre-sbsh#361) stays undialable forever on `kuke run` (`connect: permission denied`), while `kuke restart` heals it because its stop+start rebinds the socket fresh at `0o660`.
- Fix wires the existing heal into the daemon's `AttachContainer` RPC (`internal/client/local/local.go`), so a live wrong-mode socket is re-chmod'd in place — no workload bounce, no #630-prone start-on-running re-entry. This makes both `kuke run` (Ready branch) and plain `kuke attach` self-healing for this class, matching `kuke restart`.
- Implementation: extracted a per-spec heal helper from the existing cell-level `reapplyAttachableSocketPerms`, added an exported `ReapplyAttachableSocketPerms(spec)` on the runner (gated on `spec.Attachable`), threaded it through the `Runner` interface and a delegating `controller.Exec` method, and called it from `AttachContainer` after the task-liveness gate. Best-effort and idempotent: a socket already at `0o660` is left untouched; ENOENT is a benign no-op; the daemon runs as root so the chmod/chown succeeds.

## Notes for reviewer

- The cell-level `reapplyAttachableSocketPerms(cell)` behavior is preserved (it now delegates to the shared per-spec helper). The only behavioral delta on that path is the warn-log message text, which now carries a `via` field ("idempotent StartCell skip" vs "attach") to distinguish the two callers; no test asserts on the message text.
- Healing on `AttachContainer` (rather than only the `run` Ready branch) is the issue's proposed approach and intentionally also benefits plain `kuke attach` — both verbs route through this RPC, both should self-heal.

## Test plan

- [x] `make test` (project CI target, `GOWORK=off`) — all 107 packages `ok`, 0 FAIL.
- [x] `GOWORK=off go vet ./internal/controller/... ./internal/client/local/...` — clean.
- [x] New runner unit tests cover the per-spec heal: wrong-mode `0o640`→`0o660` correction, idempotency on already-correct `0o660`, `!Attachable` short-circuit, and tolerant-missing-socket (ENOENT) no-op.
- [x] New controller test asserts `AttachContainer`'s `(*Exec).ReapplyAttachableSocketPerms` delegates the spec verbatim to the runner.

Closes #1169

### DIFF
--- a/internal/client/local/local.go
+++ b/internal/client/local/local.go
@@ -1150,6 +1150,16 @@ func (c *Client) AttachContainer(_ context.Context, doc v1beta1.ContainerDoc) (k
 		return kukeonv1.AttachContainerResult{}, err
 	}
 	spec := container.Spec
+	// Heal a wrong-mode live tty socket in place before handing back its
+	// path (#1169). `kuke run` against an already-Ready cell short-circuits
+	// to attach without re-entering StartCell, so the #935 StartCell-skip
+	// heal never runs on the path run/attach take: a socket an old kuketty
+	// bound 0o640 (group-read only) stays undialable forever, while
+	// `kuke restart` heals it via its stop+start rebind. Re-chmod'ing the
+	// live inode here closes that asymmetry without bouncing the workload or
+	// re-entering the #630-prone start-on-running path. Best-effort and
+	// idempotent: a socket already at 0o660 is left untouched.
+	c.ctrl.ReapplyAttachableSocketPerms(spec)
 	return kukeonv1.AttachContainerResult{
 		HostSocketPath: fs.ContainerSocketSymlinkPath(
 			c.ctrl.RunPath(),
--- a/internal/controller/controller_test.go
+++ b/internal/controller/controller_test.go
@@ -91,6 +91,8 @@ type fakeRunner struct {
 	ExistsCellRootContainerFn func(cell intmodel.Cell) (bool, error)
 	UpdateCellMetadataFn      func(cell intmodel.Cell) error
 
+	ReapplyAttachableSocketPermsFn func(spec intmodel.ContainerSpec)
+
 	// Container methods
 	ListContainersFn    func(realmName, spaceName, stackName, cellName string) ([]intmodel.ContainerSpec, error)
 	CreateContainerFn   func(cell intmodel.Cell, container intmodel.ContainerSpec) (intmodel.Cell, error)
@@ -451,6 +453,12 @@ func (f *fakeRunner) UpdateCellMetadata(cell intmodel.Cell) error {
 		return f.UpdateCellMetadataFn(cell)
 	}
 	return errors.New("unexpected call to UpdateCellMetadata")
+}
+
+func (f *fakeRunner) ReapplyAttachableSocketPerms(spec intmodel.ContainerSpec) {
+	if f.ReapplyAttachableSocketPermsFn != nil {
+		f.ReapplyAttachableSocketPermsFn(spec)
+	}
 }
 
 // Container methods
--- a/internal/controller/get_container.go
+++ b/internal/controller/get_container.go
@@ -200,3 +200,16 @@ func stageStatusesForContainer(cell intmodel.Cell, name string) []intmodel.Stage
 func (b *Exec) ListContainers(realmName, spaceName, stackName, cellName string) ([]intmodel.ContainerSpec, error) {
 	return b.runner.ListContainers(realmName, spaceName, stackName, cellName)
 }
+
+// ReapplyAttachableSocketPerms heals a single attachable container's live
+// tty control socket inode to the connect(2)-able mode/group on the attach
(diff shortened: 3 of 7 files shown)


eminwux:
**PR #1172 Review — `fix(run): heal wrong-mode attachable tty socket on the attach path`**

Correct, well-scoped fix. The `reapplyAttachableSocketPerm(spec, via)` extraction preserves the StartCell-skip loop's behavior exactly — the per-iteration `continue`-on-error maps cleanly to the helper's `return`-on-error (one call per spec), and the chmod/chown/ENOENT logic is identical save the new `via` log field. Wiring the heal into `AttachContainer` after the task-liveness gate is the right insertion point, and the interface→`Exec`→runner plumbing is complete with no dead symbols. No secret/PII exposure in the new log lines.

One non-blocking note:

- **Test plan** `internal/controller/runner/attachable.go`: per project `CLAUDE.md`, changes touching `attachable.go` / the attach socket path are validated by the dev-init PTY attach smoke, which isn't in the test plan. It's host-gated (containerd + root, and reproducing the 0o640 case needs a pre-sbsh#361 kuketty), and the heal logic is directly unit-covered with `make e2e` green — so this is a deferred verification to run pre-merge, not a blocker. Recommend `make dev-init` once on a real host to confirm the end-to-end `kuke run`-against-Ready attach path.

LGTM to proceed — the heal mechanism is unit-tested at the filesystem level, delegation is tested, and the extraction is behavior-preserving.

PR #1173, opened 2026-06-09
fix(run): fail fast on permanent EACCES against a present attach socket

### LINKED ISSUE #1170: fix(run): attachBootRaceSentinel misclassifies permanent EACCES on a live socket as a transient stale-socket race — 10s wasted + misleading error
## Summary

`attachBootRaceSentinel` (`cmd/kuke/run/attach.go:269-283`) treats **any** `syscall.EACCES` in the error chain as the transient "stale pre-fix tty socket" boot-race class and retries for the full `attachPingRetryBudget = 10s`. But EACCES against a socket inode that is **present and has a live listener** is a *permanent* permission condition (wrong mode, e.g. `0640`), not a race — retrying cannot resolve it. The result is a 10-second hang followed by a misleading error:

```
Error: attach stale socket: kuketty has not yet unlinked the pre-fix tty socket:
error attaching client: ping failed:
dial unix /opt/kukeon/s/<shortid>: connect: permission denied
```

The socket is neither stale nor mid-unlink — kuketty is up and serving; the operator simply lacks group-write on the inode. The message sends diagnosis in entirely the wrong direction (it implies a kuketty/unlink timing problem).

## Current code

```go
func attachBootRaceSentinel(ctx context.Context, err error) error {
    if err == nil { return nil }
    if ctx.Err() != nil { return nil }
    if errors.Is(err, context.DeadlineExceeded) { return errdefs.ErrAttachPingTimeout }
    if errors.Is(err, syscall.EACCES) || errors.Is(err, syscall.ENOENT) {
        return errdefs.ErrAttachStaleSocket   // <-- EACCES lumped in with ENOENT
    }
    return nil
}
```

The conflation is intentional per the comment (EACCES against a `0o640` stale inode during the pre-fix window), but it cannot distinguish a *transient* pre-fix window from a *permanent* wrong-mode socket. In practice the permanent case (see companion bug) is the one operators actually hit, and for it the retry is pure wasted latency + a wrong message.

## Proposed fix

Split the two errnos by whether the socket path currently exists:

- **ENOENT** → genuinely transient (the `Remove→Listen` gap inside kuketty init, #933). Keep retrying.
- **EACCES on a *present* socket** → permanent permission failure. Fail fast (no 10 s budget) with a message that names the actual cause — the socket mode / missing group-write — and points at the heal (`kuke restart <cell>`, or the in-place heal from the companion issue) rather than blaming a "stale socket".

`stat(2)` on `opts.SocketPath` at classification time is enough to tell the two apart: ENOENT-from-stat → still the transient class; exists-but-EACCES-on-connect → permanent.

## Related

- Companion bug #1169 — `kuke run` Ready short-circuit attaches without healing the socket, which is what produces the permanent-EACCES condition this sentinel then misclassifies.
- #933 — introduced the EACCES/ENOENT retry; this narrows it.

eminwux:
## Summary

- `attachBootRaceSentinel` previously lumped **any** `syscall.EACCES` into the transient stale-socket boot-race class, so the post-`StartCell` attach loop retried for the full `attachPingRetryBudget = 10s` before surfacing a misleading `attach stale socket: kuketty has not yet unlinked the pre-fix tty socket` error. But EACCES against a socket inode that is *present with a live listener* is a permanent wrong-mode condition (e.g. `0o640` without group-write) — retrying cannot resolve it, so the 10s hang is pure wasted latency and the message points diagnosis the wrong way.
- This splits the two errnos by whether the socket path currently exists, per the issue's proposal:
  - **ENOENT** (in the dial chain) → genuinely transient (the `Remove→Listen` gap inside kuketty init, #933). Keep retrying.
  - **EACCES on a *present* socket** (`os.Stat` succeeds) → permanent. Fail fast on the first attempt with the new `errdefs.ErrAttachPermissionDenied`, whose message names the actual cause (socket mode / missing group-access) and points at the heal (`kuke restart <cell>`).
  - **EACCES on an *absent* socket** (`os.Stat` → ENOENT, or any stat error) → conservatively still treated as the transient stale-socket class and retried — preserving prior behavior on the ambiguous signal rather than failing fast.
- The classifier now returns `(sentinel, retry bool)`; `runWithPingRetry` surfaces a non-retryable sentinel immediately instead of waiting out the budget. EACCES preserves `syscall.EACCES` in the wrapped chain so operators still see the underlying cause.

## Implementation notes

- New sentinel `errdefs.ErrAttachPermissionDenied` added alongside `ErrAttachStaleSocket` (`internal/errdefs/errdefs.go`).
- Existing `staleSocket{,ENOENT}Err` tests use a non-existent socket path (`/opt/kukeon/r/s/st/c/work/tty/socket`), so `os.Stat` reports absent → they remain classified transient and still pass unchanged. New test `TestRun_Attach_PermissionDenied_PresentSocket_FailsFast` seeds a real on-disk socket file so the stat sees it present and asserts: 1 call (no retry), chain includes `ErrAttachPermissionDenied` + `syscall.EACCES`, and is NOT wrapped with `ErrAttachStaleSocket`.
- Scope: CLI-side error classification in the `kuke run` post-start attach retry loop only. No daemon, build-path, or `/opt/kukeon` layout change, so the `make dev-init` smoke is not implicated; the companion daemon-side heal is #1169 (out of scope here).

## Test plan

- [x] `GOWORK=off go test ./cmd/kuke/run/ -run Attach` — all attach tests pass, including the new fail-fast test
- [x] `GOWORK=off go vet ./cmd/kuke/run/ ./internal/errdefs/`
- [x] `make test` (full CI target: `test-root` + `test-kukebuild`) — EXIT=0

Closes #1170

### DIFF
--- a/cmd/kuke/run/attach.go
+++ b/cmd/kuke/run/attach.go
@@ -227,17 +227,29 @@ func resolveRun(cmd *cobra.Command) runFn {
 //     is cancelled (returns ctx.Err), or attachPingRetryBudget is
 //     exhausted (returns the last runErr wrapped with the matching
 //     sentinel — errdefs.ErrAttachPingTimeout for the ping-deadline
-//     class, errdefs.ErrAttachStaleSocket for the EACCES/ENOENT
-//     class — so callers can errors.Is either class without
-//     string-matching sbsh's wrap or sniffing for raw errno values).
+//     class, errdefs.ErrAttachStaleSocket for the absent-socket
+//     EACCES/ENOENT class — so callers can errors.Is either class
+//     without string-matching sbsh's wrap or sniffing for raw errno
+//     values).
+//   - EACCES against a socket that is *present on disk* is a permanent
+//     wrong-mode condition, not a race: attachBootRaceSentinel returns
+//     errdefs.ErrAttachPermissionDenied with retry=false and the loop
+//     surfaces it immediately rather than burning the full budget
+//     (#1170).
 func runWithPingRetry(ctx context.Context, run runFn, opts attach.Options) error {
 	deadline := time.Now().Add(attachPingRetryBudget)
 	var lastErr error
 	for {
 		lastErr = run(ctx, opts)
-		sentinel := attachBootRaceSentinel(ctx, lastErr)
+		sentinel, retry := attachBootRaceSentinel(ctx, lastErr, opts.SocketPath)
 		if sentinel == nil {
 			return lastErr
+		}
+		if !retry {
+			// Permanent failure (e.g. a present socket with the wrong
+			// mode): retrying cannot resolve it, so surface the matching
+			// sentinel immediately instead of waiting out the budget.
+			return fmt.Errorf("%w: %w", sentinel, lastErr)
 		}
 		if !time.Now().Before(deadline) {
 			return fmt.Errorf("%w: %w", sentinel, lastErr)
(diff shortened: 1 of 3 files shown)


eminwux:
**PR #1173 Review — `fix(run): fail fast on permanent EACCES against a present attach socket`**

LGTM — splits EACCES by `os.Stat` presence so a present wrong-mode socket fails fast with `ErrAttachPermissionDenied` (no 10s budget burn) while the absent/ENOENT pre-fix window stays the retried stale-socket class; the double-`%w` wrap preserves `syscall.EACCES`, the stat-vs-dial target matches, the TOCTOU window resolves safe-by-direction (unlinked → retry), and the new test pins `calls==1` + chain membership against a generous budget.

PR #1176, opened 2026-06-09
fix(reconcile): version the spec-hash domain so a daemon upgrade re-stamps cells instead of stranding them

### LINKED ISSUE #1171: Spec-hash drift guard is unversioned: daemon upgrade across a hash-domain change strands all pre-existing cells
## Summary

The cell spec-hash drift guard is **unversioned**, so any change to the hash *domain* (the field set in `containerSpecHashPayload`) makes every pre-existing cell look tampered. After a daemon upgrade/rebuild that crosses such a change, the guard hard-refuses to start those cells — a fleet-wide outage with no migration and no working recovery path. This bit a live fleet today (5 cells stranded) after a rebuild crossed #1155.

## Root cause

`SpecHashLabelKey = "kukeon.io/spec-hash"` is a single, unversioned label. Both enforcement sites compare `stamped` against `recompute(on-disk spec)` and **hard-refuse on any mismatch**:

- root container: `internal/controller/runner/start.go:701-708`
- child container: `internal/controller/runner/spec_hash.go:294-301` (`reuseOrRefuseExistingChildContainer`)

The guard's real job is catching *out-of-band tampering* — containerd state mutated without going through `kuke apply -f` (which re-stamps in the same transaction). But a bare hash **cannot distinguish "the spec drifted" from "the hashing algorithm changed."** When `containerSpecHashPayload`'s field set moves, every cell created by the prior daemon looks tampered.

Domain changes that already shipped:
- `bc61518` (#1001, 2026-06-01) — widened root-container diff to all spec fields
- `d4dd8df` (#1155, 2026-06-08) — added Volumes, Secrets, WorkingDir, SecurityOpts (Breaking-on-root)

A cell stamped by a daemon before either commit, started by a daemon after it, drops to Failed on every start attempt.

## Why there's no recovery today

- **No migration / re-stamp on upgrade.** Nothing re-derives the label when the algorithm changes.
- **`kuke apply -f` no-ops when the root spec is byte-identical** (separate bug), so it cannot clear drift even though it's the documented remedy in the error message ("run `kuke apply -f` to reconcile").
- The only escapes are `ctr label` surgery or full cell recreate (wipes the overlay).

## Scope of the live incident

5 cells in `default.kukeon.io` carry the stale root hash `6271df3b…` (pre-#1155 domain); the running daemon recomputes `1c06e0b3…`:

- `kukeon-dev-1`, `kukeon-pm-0`, `kukeon-pr-0`, `sbsh-dev-0`, `sbsh-pr-0`

Note `sbsh-dev-0` / `sbsh-pr-0` report `SYNC=Synced` yet carry the stale hash — silent landmines that fail on the *next* restart, not visibly broken now.

Empirically confirmed it is **strictly upgrade/version-skew**: a cell created by the current daemon (`agents-dev-0`) has `stamped == recompute` for both root and work containers and restarts fine. A fresh `kuke init` that never changes daemon version is unaffected. The exposure is the `make dev-init` / daemon-rebuild loop and any real daemon upgrade with cells on disk.

## Proposed fix — version the hash domain + self-heal on version mismatch

The on-disk `metadata.json` is already authoritative (it is exactly what `apply -f` writes back); the hash is a tripwire, not a source of truth. On an *algorithm* change there is nothing to recover — the daemon should re-stamp from the spec it already trusts.

**1. Add a domain version**, bumped whenever the payload changes:

```go
const SpecHashLabelKey        = "kukeon.io/spec-hash"
const SpecHashVersionLabelKey = "kukeon.io/spec-hash-version"
const SpecHashDomainVersion   = "3" // bumped at #1001 (→2), #1155 (→3)
```

**2. Stamp both labels** (one line in `stampSpecHashOnLabels`).

**3. Make both guard sites a three-way decision** instead of two-way:

| stamped version | hash compare | action |
|---|---|---|
| absent (legacy) | — | **re-stamp** from on-disk spec, reuse |
| == current | mismatch | **refuse** — genuine tamper (unchanged) |
| == current | match | reuse (unchanged) |
| != current | — | **re-stamp** from on-disk spec, reuse, log INFO |

The "version differs → re-stamp, don't refuse" arm turns a domain bump from a fleet outage into a one-time silent re-stamp on next start. The only cost: tamper detection is suspended for the *first* start across an upgrade — acceptable, since an upgrade is already a trust boundary and the on-disk spec is authoritative.

The reuse path currently runs no `CreateContainer`, so it never re-stamps — add a small helper (`container.SetLabels`/`Update`) to write the fresh `(version, hash)` pair on the reuse branch at both sites.

**Bonus:** the "absent → re-stamp, reuse" arm auto-heals the 5 currently-stranded cells (bare hash, no version label) on their next `kuke start` — no manual `ctr label` surgery needed going forward.

## Secondary fix — operator escape hatch

Add `kuke apply -f --restamp` (or a `kuke repair` verb) that force-re-stamps a cell's labels from its on-disk spec without recreating. This also fixes the `apply -f` no-op-on-identical-root bug and gives operators an explicit lever for the rare case where waiting for a start isn't acceptable.

## Test plan

- Unit test that wires `SpecHashDomainVersion` into the existing `TestSpecHashDomainPinsToDiffCellBreakingFields` so a domain change forces a version bump in the same edit.
- Table test for the three-way guard at both root and child sites (absent / same-version-match / same-version-mismatch / different-version).
- Regression: a cell stamped with version N-1 starts cleanly under version N and ends with version-N labels.

## Acceptance criteria

- [ ] A cell stamped by a prior hash domain starts cleanly under a newer daemon and is re-stamped to the current domain.
- [ ] Genuine within-version tampering still refuses with `ErrCellSpecHashDrift`.
- [ ] Both root and child guard sites re-stamp on the reuse path.
- [ ] `SpecHashDomainVersion` is bump-forced by the domain-pinning test.
- [ ] (Optional) `kuke apply -f --restamp` / `kuke repair` re-stamps without recreate.

eminwux:
## Summary

- The cell spec-hash drift guard was **unversioned**: a bare `kukeon.io/spec-hash` label cannot tell "the spec drifted" (genuine tampering → refuse) from "the hashing algorithm changed" (a daemon upgrade widened `containerSpecHashPayload` → must NOT refuse). After a rebuild that crossed a domain change (#1001, #1155), every pre-existing cell looked tampered and the guard hard-refused to start them — a fleet-wide outage with no migration path (the live incident in #1171: 5 cells stranded).
- Adds a companion `kukeon.io/spec-hash-version` label (`SpecHashVersionLabelKey`) stamped alongside the hash, pinned to `SpecHashDomainVersion = "3"`. Both guard sites (root container in `start.go`, child container in `spec_hash.go`'s `reuseOrRefuseExistingChildContainer`) now make a **three-way** decision instead of two-way:

  | stamped version | hash compare | action |
  |---|---|---|
  | absent (legacy) / != current | — | **re-stamp** from on-disk spec, reuse (log INFO) |
  | == current | mismatch | **refuse** with `ErrCellSpecHashDrift` (unchanged) |
  | == current | match | reuse untouched (unchanged) |

- A domain bump is now a one-time silent re-stamp on the next start, not an outage. The on-disk `metadata.json` (exactly what `apply -f` writes back) is authoritative; the hash is a tripwire, so on an algorithm change there is nothing to recover — the daemon re-derives the label. Tamper detection is suspended only for the *first* start across an upgrade, which is already a trust boundary.
- **Bonus (#1171 AC):** the legacy-absent arm auto-heals the currently-stranded bare-hash cells on their next `kuke start` — no `ctr label` surgery.
- Single source of truth `specHashLabels(spec)` now feeds all six create/re-stamp sites (4 `WithExtraLabels` call sites in `start.go`/`provision.go` + `stampSpecHashOnLabels` + the reuse-path re-stamp), so the hash and version can never be written out of lockstep.

## Design notes / non-obvious calls

- **Version-pin test placement.** The AC asked to "wire `SpecHashDomainVersion` into the existing `TestSpecHashDomainPinsToDiffCellBreakingFields`". That test lives in the *external* `runner_test` package and pins the payload to `apply.DiffCell`'s breaking domain. The new `TestSpecHashDomainVersionPinsToPayload` reflects over the **unexported** `containerSpecHashPayload`, which an external-package test cannot reach — so it lives in-package (`spec_hash_guard_test.go`) as a dedicated companion guard, with a cross-reference comment added to the existing DiffCell test. It pins the version to the payload's reflected JSON field set: adding/removing/renaming a payload field changes the set, fails the test, and the only green fix is to bump `SpecHashDomainVersion` and register the new field set — forcing the bump in the same edit, as the AC intends.
- **Re-stamp is best-effort.** `restampSpecHashLabels` logs a WARN and proceeds if `SetLabels` fails — the labels are a tripwire, not a source of truth, and the next start re-attempts. Failing the start over a label write would reintroduce the fragility this fix removes.
- **Optional secondary fix deferred.** AC item 5 (`kuke apply -f --restamp` / `kuke repair` operator escape hatch, plus the separate `apply -f` no-op-on-identical-root bug) is explicitly marked Optional in the issue. The primary self-heal path (re-stamp on next start) already resolves the outage and auto-heals the stranded cells, so the escape hatch is left for a follow-up. Recommend PM file it separately if still wanted.

## Test plan

- [x] `make test` (full CI target: test-root + test-kukebuild) — all packages pass
- [x] `GOWORK=off go build ./internal/controller/runner/` + `go vet` — clean
- [x] `pre-commit run --files <changed>` — golangci-lint, go-fmt, go-mod-tidy, addlicense all pass
- [x] New/updated unit tests:
  - `TestClassifySpecHashReuse` — table test for the three-way guard (absent / no-labels / different-version / same-version-match / same-version-mismatch / same-version-empty-hash)
  - `TestReuseOrRefuseExistingChildContainer_PriorDomainVersionRestamps` — a prior-domain stamp (even with a diverged hash) re-stamps + reuses, not refuses
  - `TestReuseOrRefuseExistingChildContainer_LegacyRecordReusesAndRestamps` — bare pre-#1171 hash (the stranded-cell shape) is reused after re-stamp to the current (hash, version) pair
  - `TestReuseOrRefuseExistingChildContainer_DivergedHashRefuses` — updated: within-version tamper (current version + diverged hash) still refuses with `ErrCellSpecHashDrift`
  - `TestReuseOrRefuseExistingChildContainer_MatchingHashReuses` — updated: current version + matching hash reuses untouched (no re-stamp)
  - `TestSpecHashDomainVersionPinsToPayload` — bump-forces `SpecHashDomainVersion` on any payload field-set change

## Acceptance criteria

- [x] A cell stamped by a prior hash domain starts cleanly under a newer daemon and is re-stamped to the current domain.
- [x] Genuine within-version tampering still refuses with `ErrCellSpecHashDrift`.
- [x] Both root and child guard sites re-stamp on the reuse path.
- [x] `SpecHashDomainVersion` is bump-forced by the domain-pinning test.
- [ ] (Optional) `kuke apply -f --restamp` / `kuke repair` re-stamps without recreate — deferred to a follow-up (see Design notes); the primary self-heal path already resolves the outage.

Closes #1171

### DIFF
--- a/internal/controller/runner/provision.go
+++ b/internal/controller/runner/provision.go
@@ -1573,9 +1573,7 @@ func (r *Exec) createCellContainers(cell *intmodel.Cell) (containerd.Container, 
 				return nil, fmt.Errorf("failed to prepare attachable container %s: %w", containerdID, attachErr)
 			}
 			buildOpts := append(r.daemonDefaultBuildOpts(), attachOpts...)
-			buildOpts = append(buildOpts, ctr.WithExtraLabels(map[string]string{
-				SpecHashLabelKey: ComputeContainerSpecHash(containerSpec),
-			}))
+			buildOpts = append(buildOpts, ctr.WithExtraLabels(specHashLabels(containerSpec)))
 			// Merge `kuke run --env` runtime env into the attachable
 			// container's spec env (issue #834). Returns containerSpec
 			// unchanged for non-attachable containers or when
@@ -2051,9 +2049,7 @@ func (r *Exec) ensureCellContainers(cell *intmodel.Cell) (containerd.Container, 
 				return nil, fmt.Errorf("failed to prepare attachable container %s: %w", containerdID, attachErr)
 			}
 			buildOpts := append(r.daemonDefaultBuildOpts(), attachOpts...)
-			buildOpts = append(buildOpts, ctr.WithExtraLabels(map[string]string{
-				SpecHashLabelKey: ComputeContainerSpecHash(containerSpec),
-			}))
+			buildOpts = append(buildOpts, ctr.WithExtraLabels(specHashLabels(containerSpec)))
 			createdContainer, containerCreateErr := r.ctrClient.CreateContainerFromSpec(
 				internalRealm.Spec.Namespace,
 				containerSpec,
--- a/internal/controller/runner/spec_hash.go
+++ b/internal/controller/runner/spec_hash.go
@@ -38,6 +38,28 @@ import (
 // containerd state, so steady-state restarts always match. Issue #867.
 const SpecHashLabelKey = "kukeon.io/spec-hash"
 
+// SpecHashVersionLabelKey pins the *domain version* the SpecHashLabelKey value
+// was computed under — the field set in containerSpecHashPayload. A bare hash
+// cannot distinguish "the spec drifted" (genuine out-of-band tampering, which
+// must refuse) from "the hashing algorithm changed" (a daemon upgrade widened
+// the payload, which must NOT refuse). Without this label, any change to the
+// hash domain made every pre-existing cell look tampered and stranded the
+// whole fleet on the next start with no migration path. The guard now keys its
+// refuse-vs-re-stamp decision on whether this version matches the running
+// daemon's SpecHashDomainVersion. Issue #1171.
+const SpecHashVersionLabelKey = "kukeon.io/spec-hash-version"
+
+// SpecHashDomainVersion is the current hash-domain version. Bump it whenever
+// containerSpecHashPayload's field set changes — the
+// TestSpecHashDomainVersionPinsToPayload regression guard forces the bump in
+// the same edit by pinning the version to the payload's reflected field set.
+// History: "1" (issue #867, original domain) → "2" (#1001, widened the
+// root-container diff to all spec fields) → "3" (#1155, added Volumes,
+// Secrets, WorkingDir, SecurityOpts). A cell stamped under an older version
+// is re-stamped from its authoritative on-disk spec on the next start rather
+// than refused. Issue #1171.
+const SpecHashDomainVersion = "3"
+
 // containerSpecHashPayload is the deterministic projection of an
 // intmodel.ContainerSpec that ComputeContainerSpecHash hashes. The field set
 // must match exactly what `apply.DiffCell` classifies as "requires containerd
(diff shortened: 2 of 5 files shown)


eminwux:
**PR #1176 Review — `fix(reconcile): version the spec-hash domain so a daemon upgrade re-stamps cells instead of stranding them`**

Correct and well-structured. The three-way `classifySpecHashReuse` (absent/prior-version → re-stamp, current-version + diverged-hash → refuse, current-version + match/empty → reuse) cleanly separates "algorithm changed" from "spec tampered," and routing all six label-write sites through `specHashLabels` makes the (hash, version) pair atomic so a record can never carry one without the other. I grep-confirmed zero remaining hash-only write sites on the head, both `restampSpecHashLabels` call sites have their vars in scope, the `%w`-wrapped `ErrCellSpecHashDrift` is preserved, and no secrets leak in the new log lines. Tests faithfully exercise the bug scenario (`PriorDomainVersionRestamps` drives the real guard with a version-"2" stale-hash record), and `TestSpecHashDomainVersionPinsToPayload` is a genuine bump-forcing guard. No new files (no header concern), no new sentinels, no dead symbols.

One non-blocking note:

- **Test plan** `Makefile (dev-init)`: the daemon-read smoke (`make dev-init` / `kuke status`) isn't listed. Per project `CLAUDE.md` this is normally flagged for daemon-touching changes — but it doesn't gate here: a fresh single-version `dev-init` can't exercise the cross-domain re-stamp (the fix's core path needs a pre-existing prior-domain cell), so the unit tests are the faithful validation, and green `make e2e` covers the steady-state cell-start integration. Informational only.

LGTM — unconditional. The optional AC #5 (`kuke repair` escape hatch) is documented as deferred to a follow-up, which is fine.

PR #1177, opened 2026-06-09
fix(cni): call CNI DEL unconditionally so stopped cells don't leak iptables masquerade chains

### LINKED ISSUE #1174: cell-delete: CNI DEL skipped when task already stopped → bridge-plugin iptables masquerade chains leak permanently (88 orphaned CNI-* chains observed)
## Summary

`kuke delete cell` permanently leaks the CNI bridge plugin's iptables masquerade chains/rules whenever the cell's task is **already stopped** at delete time. On a host where I deleted ~all cells, this left **88 orphaned `CNI-*` nat chains and 272 jump rules** behind, referencing ~101 IPs across 88 deleted cells (`dezot-*`, `crew-*`, `meta-*`, `kukeon-pr-reviewer-*`, …), none of which the daemon still knows about.

## Root cause

The teardown calls `purgeCNIForContainer` (`internal/controller/runner/helpers.go:42`), which only invokes CNI DEL when a live netns path is available:

```go
// Attempt CNI DEL even if netnsPath is empty (best-effort cleanup)
// Some CNI plugins may handle this gracefully
if netnsPath != "" {
    if err = cniMgr.DelContainerFromNetwork(r.ctx, containerID, netnsPath); err == nil {
        ...
    }
} else {
    r.logger.DebugContext(r.ctx, "skipping CNI DEL (no netns path available)", "container", containerID)
}
```

The **comment says** "Attempt CNI DEL even if netnsPath is empty", but the `if netnsPath != ""` guard does the **opposite** — it skips DEL entirely.

`netnsPath` is empty for any stopped cell: `getContainerNetnsPath` (`helpers.go:383`) derives it from the running task's PID and returns `("", err)` when `container.Task(...)` reports no task. In `stopAndDeleteContainer` (`delete_cell.go:248`) the netns is fetched *before* the container is stopped, so a cell that was already stopped (no task) → empty netns → **CNI DEL skipped**.

The bridge plugin's `ipMasq` iptables chains/rules are only removed by a CNI DEL. IPAM allocation files and the CNI result cache *are* cleaned by the file-scanning fallback in `purgeCNIForContainer`, but the **iptables rules are not** — so they leak with no way to recover them. Worse, `purgeCNIForContainer` then deletes the CNI result cache entry (`helpers.go:122-153`), destroying the cached result that a later DEL would have needed.

`DelContainerFromNetwork` (`internal/cni/container.go:88`) just builds a `libcni.RuntimeConf{NetNS: netnsPath}` and calls `DelNetworkList`. libcni's `DelNetworkList` does **not** require a live netns — it works with an empty `NetNS` and reconstructs teardown (including the bridge plugin's iptables rules) from the cached result. So calling DEL with an empty netns is both safe and the correct fix.

## Impact

- Unbounded growth of nat-table `CNI-*` chains across the host's lifetime (88 observed). Every cell deleted while stopped leaks one chain + its POSTROUTING/masquerade rules.
- iptables evaluation overhead grows monotonically; eventual rule-count/teardown pathologies.
- No operator-visible error — delete reports success.

## Reproduction

1. Create a cell that attaches to a space CNI network; let it reach `Ready`.
2. Stop it (`kuke stop`/`kill`, or let it exit) so the containerd task is gone.
3. `kuke delete cell <name>`.
4. `sudo iptables -t nat -S | grep 'CNI-'` → the cell's `CNI-<hash>` chain and its `-A POSTROUTING -s <ip>/32 ... -j CNI-<hash>` rule survive.

Observed on this host:

```
$ sudo iptables -t nat -S | grep -c '^-N CNI-'      # 88 leaked chains
$ sudo iptables -t nat -S | grep -cE '^-A .*CNI-'   # 272 leaked rules
$ sudo ./kuke get cells --realm default              # No cells found
```

Example surviving rule for a long-deleted cell:

```
-A POSTROUTING -s 10.88.0.119/32 -m comment --comment "name: \"default-default\" id: \"default_default_dezot-pm-0_root\"" -j CNI-b4f2bb8a97c04cc06f5d8380
```

## Suggested fix

1. In `purgeCNIForContainer`, call `cniMgr.DelContainerFromNetwork(ctx, containerID, netnsPath)` **unconditionally** (even when `netnsPath == ""`), matching the existing comment — libcni DEL is netns-optional and uses the cache.
2. Ensure CNI DEL runs **before** the CNI result-cache entry is removed (it already does in the current ordering, but the fix must preserve that — the cache is what makes a netns-less DEL able to remove the iptables rules).
3. Consider a teardown-time sweep that removes any `CNI-*` chain whose POSTROUTING comment references a cell ID no longer present, as a backstop for hosts that already leaked (and to cover plugins that don't fully clean via cached DEL).

## Cleanup for already-affected hosts

Existing leaked chains must be reaped out-of-band (delete the per-cell `CNI-<hash>` chain + its POSTROUTING jump for every cell ID the daemon no longer lists).

---
Found while auditing leaked resources after a bulk `kuke delete cell`. Companion issue covers orphaned containerd containers + snapshots + IPAM reservations surviving delete.

eminwux:
## Summary

- `kuke delete cell` permanently leaked the bridge plugin's per-cell `CNI-*` iptables masquerade chains/rules whenever the cell's containerd task was **already stopped** at delete time (88 orphaned chains / 272 rules observed on one host — issue #1174).
- Root cause: `purgeCNIForContainer` (`internal/controller/runner/helpers.go`) guarded the CNI DEL call behind `if netnsPath != ""`. A stopped cell has no task → `getContainerNetnsPath` returns `""` → DEL was skipped, even though the in-code comment said "Attempt CNI DEL even if netnsPath is empty". The ipMasq iptables chains are only removed by a CNI DEL, so they leaked with no recovery path; worse, the function then cleared the CNI result-cache entry the next DEL would have needed.
- Fix (issue's suggested fix #1): call `cniMgr.DelContainerFromNetwork(ctx, containerID, netnsPath)` **unconditionally**. libcni's `DelNetworkList` is netns-optional — it reconstructs teardown (including the bridge plugin's iptables rules) from the cached result, so an empty-netns DEL is both safe and the correct cleanup. The existing call ordering already runs DEL **before** the cache-removal block, satisfying suggested-fix #2; the new comment documents why that ordering must be preserved.

This also fixes the analogous root-container teardown sites that already passed `netnsPath == ""` into `purgeCNIForContainer` (e.g. `delete_cell.go:179`), which were leaking for the same reason.

## Scope notes

- Suggested-fix #3 (a teardown-time sweep that reaps `CNI-*` chains whose POSTROUTING comment references an unknown cell ID, as a backstop for already-leaked hosts and non-cache-cleaning plugins) is a "Consider" item with materially larger blast radius (host-wide iptables enumeration + deletion). Deferred as out-of-scope for this minimal-risk fix; the issue's "Cleanup for already-affected hosts" note already calls for out-of-band reaping of pre-existing leaks. Recommend PM file a follow-up if the backstop sweep is wanted.

## Test plan

- [x] `make test` (full CI suite: `test-root` + `test-kukebuild`) — EXIT=0
- [x] `go test ./internal/controller/runner/ -run TestPurgeCNIForContainer -v` — both pass, including the new `TestPurgeCNIForContainer_CallsDelWithEmptyNetns` regression guard (fake CNI plugin records its DEL invocation; asserts DEL fires with an empty netnsPath — fails under the old guard)
- [x] `go vet ./internal/controller/runner/ ./internal/cni/` — clean

Closes #1174

### DIFF
--- a/internal/controller/runner/helpers.go
+++ b/internal/controller/runner/helpers.go
@@ -55,17 +55,19 @@ func (r *Exec) purgeCNIForContainer(containerID, netnsPath, networkName string) 
 			)
 			if err == nil {
 				if err = cniMgr.LoadNetworkConfigList(cniConfigPath); err == nil {
-					// Attempt CNI DEL even if netnsPath is empty (best-effort cleanup)
-					// Some CNI plugins may handle this gracefully
-					if netnsPath != "" {
-						if err = cniMgr.DelContainerFromNetwork(r.ctx, containerID, netnsPath); err == nil {
-							purged = append(purged, "cni-del")
-							r.logger.DebugContext(r.ctx, "called CNI DEL for container", "container", containerID)
-						} else {
-							r.logger.DebugContext(r.ctx, "CNI DEL failed (netns may be invalid)", "container", containerID, "error", err)
-						}
+					// Attempt CNI DEL even when netnsPath is empty. libcni's
+					// DelNetworkList is netns-optional: it reconstructs teardown
+					// (including the bridge plugin's ipMasq iptables chains/rules)
+					// from the cached CNI result, which the cache-removal block
+					// below only clears *after* this DEL has run. A stopped cell
+					// has no containerd task and thus no netns path, so gating DEL
+					// on netnsPath != "" permanently leaked the per-cell CNI-*
+					// masquerade chains on every delete-while-stopped (issue #1174).
+					if err = cniMgr.DelContainerFromNetwork(r.ctx, containerID, netnsPath); err == nil {
+						purged = append(purged, "cni-del")
+						r.logger.DebugContext(r.ctx, "called CNI DEL for container", "container", containerID)
 					} else {
-						r.logger.DebugContext(r.ctx, "skipping CNI DEL (no netns path available)", "container", containerID)
+						r.logger.DebugContext(r.ctx, "CNI DEL failed (netns may be invalid)", "container", containerID, "error", err)
 					}
 				}
 			}
--- a/internal/controller/runner/helpers_test.go
+++ b/internal/controller/runner/helpers_test.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #1177 Review — `fix(cni): call CNI DEL unconditionally so stopped cells don't leak iptables masquerade chains`**

LGTM — drops the `netnsPath != ""` guard at the single `purgeCNIForContainer` chokepoint so a delete-while-stopped cell still fires libcni's netns-optional `DelNetworkList`, reclaiming the leaked `CNI-*` masquerade chains; DEL ordering before the cache-removal block is preserved, the reassigned `err` is log-only (function still returns `nil`), and the new fake-plugin test fails under the old guard. CI green, commits verified.

PR #1178, opened 2026-06-09
fix(delete): sweep orphan containers/snapshots/IPAM when cell metadata is already gone

### LINKED ISSUE #1175: cell-delete: idempotent ErrCellNotFound early-return bypasses orphan sweeps → containerd container + Active snapshot + IPAM reservation survive a "successful" delete
## Summary

`kuke delete cell` can report success (and the daemon stops listing the cell) while leaving the cell's **containerd container, active overlay snapshot, and CNI IPAM reservation orphaned on the host**. After a bulk delete of all cells, the daemon reports `No cells found` in the `default` realm, yet two cells' resources fully survive:

| Cell | containerd containers | active snapshots | IPAM reservation |
|------|----------------------|------------------|------------------|
| `kukeon-pr-3` | `default_default_kukeon-pr-3_{root,work}` | both Active | `10.88.0.219` |
| `kukeon-dev-2` | `default_default_kukeon-dev-2_{root,work}` | both Active | `10.88.0.231` |

These match the user-reported symptom that **a couple of cells "did not get deleted" on the first `kuke delete cell` and required a second run** — and even after the daemon shows them gone, the containerd/snapshot/IPAM state is still present.

## Evidence

```
$ sudo ./kuke get cells --realm default
No cells found.

$ sudo ctr -n default.kukeon.io containers ls
default_default_kukeon-dev-2_root   ...
default_default_kukeon-dev-2_work   ...
default_default_kukeon-pr-3_root    ...
default_default_kukeon-pr-3_work    ...
$ sudo ctr -n default.kukeon.io tasks ls      # (no tasks — containers are stopped but undeleted)

$ sudo ctr -n default.kukeon.io snapshots ls | grep -E 'kukeon-(dev-2|pr-3)'
default_default_kukeon-dev-2_root   ...  Active
default_default_kukeon-dev-2_work   ...  Active
default_default_kukeon-pr-3_root    ...  Active
default_default_kukeon-pr-3_work    ...  Active

$ sudo cat /var/lib/cni/networks/default-default/10.88.0.219
default_default_kukeon-pr-3_root
eth0
```

The containers are named by **cell name** (label `kukeon.io/cell-name: kukeon-pr-3`, ID `default_default_kukeon-pr-3_root`), so the delete path's orphan-scan prefix `default_default_kukeon-pr-3_` *would* have matched them — yet they survive. That means the teardown/orphan-scan **never ran** for these cells.

## Suspected root cause

`deleteCellLocked` (`internal/controller/runner/delete_cell.go:42`) short-circuits at the top when the cell metadata is already gone:

```go
internalCell, err := r.GetCell(cell)
if err != nil {
    if errors.Is(err, errdefs.ErrCellNotFound) {
        // Idempotent: cell doesn't exist, consider it deleted
        return nil
    }
    ...
}
```

This early `return nil` is **before** all of the containerd/CNI/IPAM safety nets that the same file added precisely to mop up partial deletes:

- the name-prefix orphan-container scan + teardown (`findOrphanContainerIDs`, added for #371/#905),
- `stopAndDeleteContainer` (which deletes the container + snapshot via `SnapshotCleanup: true`),
- the post-delete `purgeCNIForContainer` IPAM/cache purge (the deterministic network-name resolution added for #685).

So if a cell's metadata is removed (or absent) while its containerd container / snapshot / IPAM reservation still exist, the next `kuke delete cell` sees `ErrCellNotFound`, declares success, and **never sweeps the orphans** — they leak permanently. The surviving container + Active snapshot + intact IPAM file (none of which `purgeCNIForContainer` ever touched) are consistent with this path.

The exact event that left metadata gone but containerd state intact for these two cells needs daemon logs from the original bulk delete to pin down (the cells' specs are gone now). Candidate triggers to confirm:
- a first delete attempt that removed metadata via a non-`deleteCellLocked` path (stack/space/`daemon reset`/team teardown) without per-cell containerd teardown, then a retry hitting the early-return;
- a first `deleteCellLocked` that partially progressed.

Regardless of trigger, the **idempotent early-return bypassing the orphan sweeps is the latent defect**: "no metadata" should not be treated as "nothing to clean up" given the runner already knows the deterministic containerd ID prefix and network name from the request.

## Impact

- Orphaned containerd containers block re-`init`/re-create on the same ID (the exact collision class #371 was meant to prevent).
- Leaked **Active** overlay snapshots pin disk (the overlayfs snapshotter dir is 31 GB on this host) and can't be reaped by image GC.
- Leaked IPAM reservations slowly exhaust the `/24` space subnet.
- Operator sees success; no signal anything leaked.

## Suggested fix

When `GetCell` returns `ErrCellNotFound`, **do not return immediately**. Run the orphan-container scan (`findOrphanContainerIDs` with the deterministic `<space>_<stack>_<cell>_` prefix), `stopAndDeleteContainer` on any survivors, and the `purgeCNIForContainer` IPAM/cache sweep, using realm/space/stack/cell from the *request* (all derivable without metadata). Only then return `nil`. This makes the idempotent path genuinely idempotent at the **host-resource** level, not just the metadata level.

## Cleanup for already-affected hosts

```
sudo ctr -n default.kukeon.io containers rm default_default_kukeon-{dev-2,pr-3}_{root,work}
# then remove now-dangling snapshots + IPAM files /var/lib/cni/networks/default-default/10.88.0.{219,231}
```

---
Companion to the CNI iptables-masquerade leak issue (CNI DEL skipped when netns is gone). Both surfaced in the same post-bulk-delete resource audit.

eminwux:
## Summary

- `kuke delete cell`'s idempotent path short-circuited at the top of `deleteCellLocked` (`internal/controller/runner/delete_cell.go`): when `GetCell` returned `ErrCellNotFound`, it `return nil`-ed **before** any of the containerd/snapshot/CNI/IPAM safety nets ran. So a cell whose metadata was removed (by a non-`deleteCellLocked` path — stack/space teardown, `daemon reset`, team teardown — or a partially-progressed prior delete) while its containerd container / Active overlay snapshot / IPAM reservation still existed would be declared "deleted" with those host resources orphaned **permanently**. This is the latent defect behind #1175 (surviving container blocks re-init, Active snapshot pins disk, IPAM reservation leaks the `/24`).
- The fix makes the idempotent path idempotent at the **host-resource** level, not just the metadata level: the deterministic containerd ID prefix (`<space>_<stack>_<cell>_`) and CNI network name are pure functions of the delete request, so the orphan sweep can run with no metadata at all.

### What changed
- New `sweepOrphansForMissingCell(cell)` runs on the `ErrCellNotFound` branch: resolves namespace + network name + cellID from the **request** (cellID falls back to `Metadata.Name` exactly as the metadata path falls back to `internalCell.Metadata.Name`), then runs the shared sweep. Returns `nil` on a clean host; surfaces a wrapped `ErrDeleteCell` if any orphan teardown fails (so the host-dirty state isn't silently swallowed).
- If the **realm** is also gone (`GetRealm` → `ErrRealmNotFound`), returns `nil`: no realm means no containerd namespace, hence nothing to sweep — preserving the historical idempotent success.
- Extracted the root-container teardown + name-prefix orphan scan + post-delete CNI/IPAM purge (previously inline in the happy path) into `sweepCellContainerResources(...)`, now shared by both the happy path and the metadata-absent path — no copy-paste of the ~35-line sweep.

### Non-obvious call (documented for reviewer pushback)
- The extracted helper turns the happy path's `naming.BuildRootContainerdID` failure from an early `return` into an aggregated `deleteError`. That error is effectively unreachable (inputs are validated non-empty upstream), and aggregating is strictly better — it still surfaces an error (cell metadata preserved for retry) **and** still runs the orphan scan instead of aborting before it. Flagging in case the reviewer prefers preserving the early-return.

## Test plan
- [x] `make test` (full CI target, `GOWORK=off`) — green
- [x] `go build` / `go vet` on `internal/controller/runner` — clean
- [x] New unit tests in `delete_cell_test.go`, all passing:
  - `TestDeleteCell_SweepsOrphansWhenMetadataAbsent` — orphan root+work containers are torn down on the `ErrCellNotFound` path (reproduces the #1175 `kukeon-pr-3` survival)
  - `TestDeleteCell_MetadataAbsentSurfacesSweepFailure` — a failed orphan teardown surfaces a wrapped `ErrDeleteCell` instead of silent success
  - `TestDeleteCell_MetadataAndRealmAbsentIsIdempotent` — both cell and realm gone returns `nil` and never calls `ListContainers`
- [x] `make dev-init` end-to-end smoke (touches daemon-read code) — exit 0; daemon-parity check shows both `kuke get realms` views identical (`default`/`kuke-system` both `Ready`); kuke attach smoke exited cleanly; parent attach plumbing intact post-flight

Closes #1175

### DIFF
--- a/internal/controller/runner/delete_cell.go
+++ b/internal/controller/runner/delete_cell.go
@@ -61,8 +61,20 @@ func (r *Exec) deleteCellLocked(cell intmodel.Cell) error {
 	internalCell, err := r.GetCell(cell)
 	if err != nil {
 		if errors.Is(err, errdefs.ErrCellNotFound) {
-			// Idempotent: cell doesn't exist, consider it deleted
-			return nil
+			// Idempotent at the host-resource level, not just metadata
+			// (issue #1175). The cell document is gone, but its containerd
+			// container, active overlay snapshot, and CNI/IPAM reservation
+			// can still be orphaned on the host — e.g. when a
+			// non-deleteCellLocked path (stack/space teardown, daemon reset,
+			// team teardown) removed metadata without per-cell teardown, or a
+			// prior deleteCellLocked partially progressed. A bare `return nil`
+			// here used to declare success and leak those resources
+			// permanently (surviving containerd ID blocks re-init, Active
+			// snapshot pins disk, IPAM reservation exhausts the subnet). The
+			// deterministic containerd ID prefix and network name are pure
+			// functions of the request, so sweep any survivors before
+			// returning.
+			return r.sweepOrphansForMissingCell(cell)
 		}
 		return fmt.Errorf("%w: %w", errdefs.ErrGetCell, err)
 	}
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1178 Review — `fix(delete): sweep orphan containers/snapshots/IPAM when cell metadata is already gone`**

LGTM — the `ErrCellNotFound` branch now runs the same request-derived orphan sweep instead of bare-`return nil`, the extraction into `sweepCellContainerResources` is faithful (the early-return→aggregate change and the added `rootContainerID != ""` CNI guard are both correct and the latter is required once `BuildRootContainerdID` can leave an empty ID), and the three new tests cover the sweep, the failure-surfacing contract, and the realm-absent idempotent short-circuit.

PR #1198, opened 2026-06-10
fix(status): make cni-plugins check defer to the daemon's bundled in-image plugins

### LINKED ISSUE #1180: status: cni-plugins check FAILs every healthy fresh install — probes host /opt/cni/bin that kuke init never populates, with a false remediation hint
## Context

On a freshly bootstrapped host (nested `kukeon-dev-root` cell, `make dev-init`, client `v0.5.0-299-gab14245`), `kuke status` exits 1 with a `FAIL` row on every run, even though `kuke init` succeeded and both realms are `Ready` with full daemon parity:

```
  cni-plugins                                       FAIL  /opt/cni/bin (missing: bridge, loopback)
Status: FAIL
# exit code 1
```

The check (`cmd/kuke/status/checks.go`, ~line 363) requires `bridge` and `loopback` binaries in the **host's** `/opt/cni/bin`. But:

- `kuke init`'s CNI bootstrap (`internal/cni/bootstrap.go` `BootstrapCNI`) only `MkdirAll`s the config/cache/bin dirs — it never installs plugin binaries.
- The kukeond image bundles its own CNI plugins (Dockerfile `cni-plugins` stage → upstream CNI v1.5.1 into the image's `/opt/cni/bin`), and CNI exec happens inside the daemon container. Networking works fine on this host (cells get IPs, bridge is up) while the check reports FAIL.
- The remediation hint at `cmd/kuke/status/checks.go:186` is false: ``"install the CNI plugin binaries; `kuke init` lays them down on a fresh host"`` — `kuke init` does not lay them down, on any host.

So `kuke status` — documented as the canonical post-init smoke — can never exit 0 on a host bootstrapped exactly the way the project's own docs prescribe, which destroys its value as a health gate (operators learn to ignore exit 1).

## Proposal

Align the check with where CNI execution actually happens:

1. Make the cni-plugins check probe the path the daemon actually uses (the bundled in-image `/opt/cni/bin`), or downgrade the host-side probe to WARN/INFO when the daemon reports its bundled plugins present.
2. Fix the remediation hint at `checks.go:186` — it must not claim `kuke init` installs plugins.

Alternative (bigger blast radius): make `kuke init` actually install plugin binaries on the host, matching the hint. Maintainer's call; the check+hint fix is the lower-blast-radius default.

## Acceptance criteria

- [ ] On a fresh `make dev-init` host with working cell networking, `kuke status` exits 0 (no cni-plugins FAIL row).
- [ ] The remediation hint no longer claims `kuke init` lays down plugin binaries.
- [ ] A host where the daemon genuinely has no CNI plugins still produces a FAIL with an accurate remediation hint.

## Out of scope

- The storage-row btrfs WARN on the same healthy host (separate issue).
- Shipping CNI plugins to the host as part of `kuke init`.

## References

- `cmd/kuke/status/checks.go:186` (false hint), ~line 363 (host probe)
- `internal/cni/bootstrap.go` (`BootstrapCNI` — mkdir only)
- `Dockerfile` `cni-plugins` stage (in-image plugin bundle)
- `docs/site/cli/kuke-status.md` (status as post-init smoke)

eminwux:
## Summary

- `kuke status` exited 1 with a `cni-plugins FAIL` row on **every** freshly bootstrapped host, because the check FAILed whenever the host's `/opt/cni/bin` lacked `bridge`/`loopback` — but `kuke init` never installs plugins onto the host (`internal/cni.BootstrapCNI` only `MkdirAll`s the dirs), and CNI exec happens **inside** the `kukeond` container from the plugins the image bundles (Dockerfile `cni-plugins` stage → upstream CNI v1.5.1). So a documented-healthy host (`make dev-init`, both realms Ready, cells getting IPs) could never pass the canonical post-init smoke. (#1180)
- The check now reconciles the host-side view against where CNI actually runs (`checkHostCNIPlugins`):
  - host plugins present → **OK** (some setups install host-side; keep the clean positive)
  - host plugins missing, daemon reachable → **WARN** — the daemon runs CNI from its bundled in-image plugins, so cell networking has a working set regardless of the host copy. WARN keeps exit 0.
  - host plugins missing, daemon unreachable → **FAIL** — no host plugins and no daemon to provide the bundled set means genuinely no plugin set to run CNI; remediation points at the real fix.
- The false remediation hint (`install the CNI plugin binaries; \`kuke init\` lays them down on a fresh host`) is gone. The same false "kuke init lays/provisions plugins" claim repeated in three docstrings (`checkHost`, `requiredCNIPlugins`) and in `docs/site/cli/kuke-status.md` is corrected to describe the bundled-in-image reality.

### Interpretation note (for reviewer)

The issue's preferred low-blast-radius option was "downgrade the host-side probe to WARN/INFO when the daemon reports its bundled plugins present." There is **no RPC** on `kukeonv1.Client` for the daemon to report its in-image plugin inventory, and adding one is the bigger-blast-radius path the maintainer explicitly de-prioritized. I read "daemon reports its bundled plugins present" as **daemon reachability** (`rc.daemonClient != nil`): a daemon that dialed successfully booted from an image that, by construction, carries the bundled `cni-plugins` stage. The residual gap — a reachable daemon built from an image *without* the cni-plugins stage — is not host-observable without that RPC; it lands as WARN, not FAIL. AC3 ("a host where the daemon genuinely has no CNI plugins still produces a FAIL") is satisfied by the daemon-unreachable → FAIL path. Happy to add the RPC instead if the reviewer prefers the literal reading. (There is no `INFO` status; the downgrade target is `WARN`.)

## Test plan

- [x] `make test` (full CI root + kukebuild suites) — exit 0, no FAILs
- [x] `GOWORK=off go build ./cmd/kuke/status/...` && `go vet ./cmd/kuke/status/...`
- [x] `GOWORK=off go test ./cmd/kuke/status/...` — new `TestCheckHostCNIPlugins` subtests cover all three verdicts (present→OK, missing+daemon-reachable→WARN, missing+daemon-unreachable→FAIL incl. remediation-text assertion)
- [x] `pre-commit run --files <changed>` — all hooks pass (prettier reflowed the status doc table)
- [x] `git grep` swept the tree for the `lays them down` literal and the "kuke init lays/provisions plugins" claim — no stray copies remain outside the corrected sites
- [ ] Live `make dev-init` + `kuke status` exit-0 smoke on a real nested host — not run; no daemon/host runtime available to this dev session. The behavior is covered by the unit tests above, which exercise the exact `rc.daemonClient != nil` gate the live path keys off.

## Acceptance criteria

- [x] On a fresh `make dev-init` host with working cell networking, `kuke status` exits 0 (no cni-plugins FAIL row) — daemon reachable → WARN, and only FAIL flips the exit code.
- [x] The remediation hint no longer claims `kuke init` lays down plugin binaries.
- [x] A host where the daemon genuinely has no CNI plugins still produces a FAIL with an accurate remediation hint — daemon-unreachable + missing host plugins → FAIL (see interpretation note on the reachable-but-broken-image residual).

Closes #1180


eminwux:
**PR #1198 Review — `fix(status): make cni-plugins check defer to the daemon's bundled in-image plugins`**

LGTM — the three-way verdict is correct (`checks.go:197-216`): `len(missing)==0`→OK, else daemon-reachable→WARN, else→FAIL, and `status.go:329` flips the exit code on `StatusFAIL` only, so the WARN path keeps `kuke status` at exit 0 on a fresh `make dev-init` host (AC1). The `rc.daemonClient != nil` reachability gate matches the dial-failure-sets-nil contract at `status.go:282` and the sibling `checkDaemon` nil-gate, the false "`kuke init` lays plugins" claim is gone from all four sites (source sweep clean; the remaining `lays them down` hits are the corrected doc line, the negative test assertion, and the unrelated `install-dev` symlink note), and all three verdicts have test coverage incl. the remediation-text assertion. No unverified commits, no dead symbols. The interpretation note's WARN-for-reachable-but-broken-image residual is a documented, maintainer-de-prioritized limitation, not a gap.

Non-blocking: the live `make dev-init` + `kuke status` exit-0 smoke is the one unchecked test-plan item — correctly deferred as host-gated (no runtime in the dev session; CI `make e2e` is green). Worth a quick run pre-merge.

PR #1202, opened 2026-06-10
fix(daemon): stop surviving containers before restarting an externally-killed daemon

### LINKED ISSUE #1184: daemon start: fall-through after external kill -9 fails with 'cell kukeond has running containers and must first be stopped' — documented liveness-keyed recovery doesn't hold
## Context

`docs/cli-use-cases.md` § Daemon lifecycle pins this invariant:

> Idempotence is keyed on **liveness**, not persisted cell state. Each lifecycle verb dials the kukeond socket alongside reading the cell's `.status` so that an externally-killed daemon (OOM, host reboot mid-run, `kill -9`) does not silently mask itself as "already running" — `daemon start` falls through to bring the cell back up […]

Observed (nested dev daemon, client `v0.5.0-299-gab14245`):

```
sudo kill -9 <kukeond pid>          # simulate OOM/external kill
sudo kuke daemon start
```

The stale-Ready divergence **is** detected and printed (the probe works), but the fall-through then fails deterministically:

```
cell kukeond has running containers and must first be stopped
# exit non-zero; daemon stays down
```

After `kill -9` of the daemon process, the cell's *other* containers (root/pause) are still running in containerd, so the start path's "bring the cell back up" refuses on the running-containers precondition. The documented recovery promise — `daemon start` heals an externally-killed daemon — does not hold; the operator must manually stop the cell first (`kuke daemon stop` or stop the cell), then `daemon start` succeeds.

Repro is deterministic (hit twice in this campaign).

## Proposal

When `daemon start`'s liveness probe finds the socket dead but the cell partially running (stale-Ready divergence already detected and printed), the fall-through should reconcile the partial state — stop/clean the surviving containers of the kukeond cell and start fresh — rather than erroring on the running-containers precondition. That matches the doc's "falls through to bring the cell back up".

## Acceptance criteria

- [ ] `kill -9` of the kukeond process followed by `kuke daemon start` brings the daemon back up, exit 0, no manual stop required.
- [ ] A genuinely-live daemon still gets "already running", exit 0 (existing invariant intact).
- [ ] The stale-Ready divergence notice still prints before the reconciling action.

## Out of scope

- `daemon stop`/`kill`/`restart` fall-through paths (not observed broken).

## References

- `docs/cli-use-cases.md` § Daemon lifecycle, liveness-keyed idempotence invariant
- #611 (prior stale-metadata fix — socket probe exists; the fall-through is what fails now)

eminwux:
## Summary
- `docs/cli-use-cases.md` § Daemon lifecycle promises `daemon start` heals an externally-killed daemon (OOM, host reboot mid-run, `kill -9`) — liveness-keyed idempotence, not persisted state. The CLI already detects the stale-Ready divergence (socket dead, metadata `Ready`) and prints the notice, then falls through to `StartCell`. But `kill -9` only kills the kukeond *process*; the cell's root/pause containers keep running in containerd, so `StartCell` refuses on its `cell %q has running containers and must first be stopped` precondition (`internal/controller/start_cell.go:71`) and the daemon stays down. The documented heal does not hold; the operator must manually `kuke daemon stop` first.
- Fix is scoped to the `daemon start` CLI path (`cmd/kuke/daemon/start/start.go` `runStart`): when the stale-Ready divergence is detected, route through `lifecycle.StopPhase(cmd, client, doc, lifecycle.DefaultTimeout)` — the **same** reconcile `kuke daemon stop` runs (`stop.go:127`) — to clean the surviving containers *before* `StartCell`. The notice still prints first (AC item 3). A stop error is wrapped as `stop stale kukeond cell:` rather than charging into a `StartCell` that would refuse.
- **Review fix-up (`c9fb339`):** the prior commit used a raw `client.StopCell`, which is weaker than what `daemon stop` actually runs. The runner's per-container stop errors are logged-and-continue, not aggregated (#868), so `StopCell` can return `Stopped=true` while a root/pause task ignores SIGTERM and survives — and the next `StartCell` then refuses on the same precondition this heal targets, silently breaking the documented heal. Routing through `StopPhase` adds the #868 verify-or-escalate-to-`KillCell` guard **and** the grace-period→SIGKILL timeout fallback (a hung `StopCell` no longer blocks `daemon start` with no escalation). New test `TestDaemonStart_StaleReadyEscalatesWhenStopReportsSuccessButTaskSurvives` covers the StopCell-reports-success-but-task-survives → KillCell → StartCell path.
- **Blast-radius note:** the general `StartCell` running-containers precondition is left intact — it's a legitimate guard for a normal `kuke start <cell>` against a genuinely-running cell. Only the daemon-start stale-Ready branch reconciles by stopping first, gated on the existing socket-probe divergence. No doc change needed: `docs/cli-use-cases.md:139` already describes this exact behavior; the fix makes the code match the doc.

## Test plan
- [x] `GOWORK=off go test ./cmd/kuke/daemon/start/` — `TestDaemonStart_StaleReadyStopsRunningContainersBeforeStart` asserts stop→start ordering + the divergence notice precedes the action; `TestDaemonStart_StaleReadyEscalatesWhenStopReportsSuccessButTaskSurvives` asserts the stop→kill→start escalation when a task survives a "successful" stop; `TestDaemonStart_StaleReadyStopError` asserts the wrapped error path; the table stale-Ready case uses the stateful `newStaleReadyFake` so `StopPhase`'s post-stop verify passes.
- [x] `GOWORK=off go test $(go list ./... | grep -v /e2e)` (the `make test-root` CI target) — all packages pass, exit 0. (`GOWORK=off` per `go.work`'s own directive: the workspace deliberately mixes incompatible module graphs; `make`/CI build each module standalone.)
- [ ] `make dev-init` end-to-end smoke — not run: the diff touches only the `kuke daemon start` CLI branch (no daemon image, no `/opt/kukeon` layout, no build-path change), so the rebuild/re-init smoke does not exercise it. The behavioral path (kill -9 → daemon start) is covered by the unit tests against the injected fake client; an on-host kill-9 repro needs a running nested daemon not available on this dev host.

## Acceptance criteria
- [x] `kill -9` of the kukeond process followed by `kuke daemon start` brings the daemon back up, exit 0, no manual stop required — the stale-Ready branch now reconciles the survivors via `StopPhase` (escalating to `KillCell` if needed) then `StartCell`s (unit-verified via call-order tests).
- [x] A genuinely-live daemon still gets "already running", exit 0 — that path returns before the reconcile call; `IsCellRunning` + reachable probe branch unchanged.
- [x] The stale-Ready divergence notice still prints before the reconciling action — notice `Printf` precedes the `StopPhase` call; asserted by the ordering test.

Closes #1184

### DIFF
--- a/cmd/kuke/daemon/start/start.go
+++ b/cmd/kuke/daemon/start/start.go
@@ -129,10 +129,16 @@ func runStart(cmd *cobra.Command, _ []string) error {
 		// refuses on its "has running containers and must first be stopped"
 		// precondition while those survive, so the documented heal — `daemon
 		// start` brings an externally-killed daemon back up — needs the
-		// surviving containers cleaned first. Stop the cell (the same
-		// reconcile the manual `kuke daemon stop` workaround performed) before
-		// starting fresh (#1184).
-		if _, stopErr := client.StopCell(cmd.Context(), doc); stopErr != nil {
+		// surviving containers cleaned first. Route through the same StopPhase
+		// reconcile `kuke daemon stop` runs (stop.go), not a raw StopCell: the
+		// runner's per-container stop errors are logged-and-continue rather
+		// than aggregated, so StopCell can return Stopped=true while a
+		// root/pause task ignores SIGTERM and survives — at which point a bare
+		// StopCell would report success and StartCell would refuse on the very
+		// precondition this heal targets. StopPhase adds the #868
+		// verify-or-escalate-to-KillCell guard and the grace-period→SIGKILL
+		// timeout fallback so a stubborn survivor is actually cleaned (#1184).
+		if stopErr := lifecycle.StopPhase(cmd, client, doc, lifecycle.DefaultTimeout); stopErr != nil {
 			return fmt.Errorf("stop stale kukeond cell: %w", stopErr)
 		}
 	}
--- a/cmd/kuke/daemon/start/start.go
+++ b/cmd/kuke/daemon/start/start.go
@@ -87,6 +87,7 @@ func runStart(cmd *cobra.Command, _ []string) error {
 		return errdefs.ErrHostNotInitialized
 	}
 
+	staleReady := false
 	if lifecycle.IsCellRunning(getRes.Cell) {
 		probe := lifecycle.ResolveReachableProbe(cmd)
 		socketPath := lifecycle.ResolveSocketPath()
@@ -112,6 +113,7 @@ func runStart(cmd *cobra.Command, _ []string) error {
 			"cell", consts.KukeSystemCellName,
 			"realm", consts.KukeSystemRealmName,
 		)
+		staleReady = true
 	}
 
 	// Recreate /run/kukeon (the cell's bind-mount source) if a reboot wiped
@@ -119,6 +121,20 @@ func runStart(cmd *cobra.Command, _ []string) error {
 	// fails with "open /run/kukeon: no such file or directory".
 	if ensureErr := lifecycle.ResolveEnsureSocketDir(cmd)(); ensureErr != nil {
 		return fmt.Errorf("ensure kukeond socket dir: %w", ensureErr)
+	}
+
+	if staleReady {
+		// An external kill only takes down the kukeond process; the cell's
+		// other containers (root/pause) keep running in containerd. StartCell
+		// refuses on its "has running containers and must first be stopped"
+		// precondition while those survive, so the documented heal — `daemon
+		// start` brings an externally-killed daemon back up — needs the
+		// surviving containers cleaned first. Stop the cell (the same
+		// reconcile the manual `kuke daemon stop` workaround performed) before
+		// starting fresh (#1184).
+		if _, stopErr := client.StopCell(cmd.Context(), doc); stopErr != nil {
+			return fmt.Errorf("stop stale kukeond cell: %w", stopErr)
+		}
 	}
 
 	startRes, err := client.StartCell(cmd.Context(), doc)
--- a/cmd/kuke/daemon/start/start_test.go
+++ b/cmd/kuke/daemon/start/start_test.go
@@ -143,30 +143,10 @@ func TestDaemonStart(t *testing.T) {
 			// dial-unix error. New behaviour: log the staleness and fall
 			// through to StartCell so the cell actually comes back up.
 			name: "metadata says Ready but socket is unreachable falls through to StartCell",
-			fake: &fakeClient{
-				getCellFn: func(_ v1beta1.CellDoc) (kukeonv1.GetCellResult, error) {
-					return kukeonv1.GetCellResult{
-						Cell: v1beta1.CellDoc{
-							Status: v1beta1.CellStatus{
-								State: v1beta1.CellStateReady,
-								Containers: []v1beta1.ContainerStatus{
-									{State: v1beta1.ContainerStateReady},
-								},
-							},
-						},
-						MetadataExists: true,
-					}, nil
-				},
-				// Surviving root/pause containers from the killed daemon are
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #1202 Review — `fix(daemon): stop surviving containers before restarting an externally-killed daemon`**

Right diagnosis and the stale-Ready gating is correct, but the reconcile uses a weaker stop path than the one it claims to replicate, leaving the heal incomplete in the very `kill -9` case it targets. No secret/PII concerns; tests and signatures are clean.

One blocking item:

- **Incomplete heal / sibling-precedent** `cmd/kuke/daemon/start/start.go:135`: raw `client.StopCell` skips the `lifecycle.StopPhase` reconcile that `kuke daemon stop` actually runs (`stop.go:127`, already imported here). `StopPhase` adds the #868 verify-or-escalate-to-`KillCell` — `StopCell` can return `Stopped=true` while a root/pause task survives (per-container stop errors are logged-and-continue, not aggregated), and the next `StartCell` then refuses on the same "has running containers" precondition this PR fixes, so the documented heal silently fails to hold whenever a survivor ignores SIGTERM. Replace with `lifecycle.StopPhase(cmd, client, doc, lifecycle.DefaultTimeout)` and add a test for the StopCell-reports-success-but-task-survives path.

Context: the code comment and PR body call this "the exact reconcile the manual `kuke daemon stop` workaround performed," but `daemon stop` runs `StopPhase` (escalation + grace-period→`KillCell` timeout guard), not a raw `StopCell` — so the claim doesn't hold and the timeout guard is also dropped (a hung `StopCell` blocks `daemon start` with no SIGKILL fallback).

The unrun `make dev-init` smoke is a fine host-gated deferral (CLI-branch-only change, no daemon image/build path). LGTM once the stop path routes through `StopPhase`.

eminwux:
Addressed in c9fb339:
- **Incomplete heal / sibling-precedent** (`cmd/kuke/daemon/start/start.go`): replaced the raw `client.StopCell` with `lifecycle.StopPhase(cmd, client, doc, lifecycle.DefaultTimeout)` — the same reconcile `kuke daemon stop` runs (`stop.go:127`). This picks up the #868 verify-or-escalate-to-`KillCell` guard (so a `StopCell` that reports `Stopped=true` while a root/pause task survives is actually cleaned before `StartCell`) and the grace-period→SIGKILL timeout fallback (so a hung `StopCell` no longer blocks `daemon start`). The error context is preserved by wrapping `StopPhase`'s error as `stop stale kukeond cell:`.
- Added `TestDaemonStart_StaleReadyEscalatesWhenStopReportsSuccessButTaskSurvives` asserting the stop→kill→start call order when the survivor ignores SIGTERM, plus the started confirmation. Updated the existing stale-Ready table case + ordering test to use a stateful fake (`newStaleReadyFake` / `runningOrStopped`) so `StopPhase`'s post-stop verify re-read sees the cleaned state.

PR body summary + test plan updated to reflect the `StopPhase` routing. Full `make test-root` suite green locally (exit 0).

eminwux:
**PR #1202 Review — `fix(daemon): stop surviving containers before restarting an externally-killed daemon`**

LGTM — routes the stale-Ready `daemon start` heal through the same `lifecycle.StopPhase` reconcile that `daemon stop` runs (`stop.go:127`), so SIGTERM-ignoring root/pause survivors are verified-and-escalated to `KillCell` before `StartCell`, instead of left to trip the running-containers precondition. Stop errors wrap before `StartCell`; the general precondition is untouched; tests cover the stop→start ordering, the stop→kill→start escalation, and the wrapped-error path. The unchecked `make dev-init` box is the documented host-gated smoke (no nested daemon on the dev host, and the CLI-only diff doesn't exercise it) — run it pre-merge if convenient, non-blocking.

PR #1203, opened 2026-06-10
fix(init): report 'already running' for idempotent re-run's cell-containers phase

### LINKED ISSUE #1186: init: idempotent re-run on healthy host reports 'cell containers: started' instead of an already-existed/no-op wording
## Context

`docs/cli-use-cases.md` § Bootstrap & teardown pins:

> Second invocation on a healthy host is idempotent: every phase reports "already existed", exit code 0, the daemon stays up.

Observed on a healthy host (daemon up, both realms Ready; client `v0.5.0-299-gab14245`), verified twice:

```
$ sudo kuke init --kukeond-image kukeon.internal/kukeond:v0.0.0-dev
    - cell containers: started
kukeond is ready (unix:///run/kukeon-dev/kukeond.sock)
# exit 0
```

Every phase reports `already existed` **except** the final cell-containers phase, which reports `started` even though the daemon's containers were already running and stay up (same daemon PID before/after — no actual restart occurs). The invariant "every phase reports \"already existed\"" does not hold; the reporting misstates what happened, which makes the idempotent re-run indistinguishable from a run that actually (re)started the daemon's containers.

## Proposal

The container-start phase of `kuke init` should detect already-running containers and report `already running` (or `already existed`), reserving `started` for containers it actually transitioned. If the implementation's start step is intentionally a no-op-on-running, only the phase wording needs fixing; the doc's invariant can then stay as written (modulo "already running" vs "already existed" — align both).

## Acceptance criteria

- [ ] Re-running `kuke init` on a healthy host reports an idempotent no-op wording for the cell-containers phase (no bare `started`).
- [ ] Exit 0 and daemon-stays-up behavior unchanged.
- [ ] `docs/cli-use-cases.md` invariant and the actual phase wording agree.

## Out of scope

- The post-`daemon reset` init path (cell genuinely created/started there — wording correct).

## References

- `docs/cli-use-cases.md` § Bootstrap & teardown, second-invocation idempotency invariant

eminwux:
## Summary

- `kuke init`'s cell-containers phase reported a bare `started` even on a healthy idempotent re-run where the daemon's containers were already running and stayed up — making a no-op re-run indistinguishable from one that actually (re)started the daemon. Issue #1186.
- Root cause: `bootstrapCell` hardcoded `section.CellStartedPre = false` (and `CellStartedPost = true`), so `CellStarted = !Pre && Post` was always `true` when the cell existed. The `printStartAction` "already running" branch (pre && post) was therefore unreachable from init.
- Fix: probe the actual pre-StartCell running state. `GetCell` already populates live container statuses off containerd on the pre-lookup, so `CellStartedPre` is now `cellContainersRunning(internalCellPre)` — true only when every spec container is observed `Ready` (the intmodel mapping of `containerd.Running`, matching StartCell's `cellTasksAllRunningFn` no-op condition, issue #149). A healthy re-run now renders `already running`; a re-run after a crash/stop still renders `started`.
- The image-drift recreate branch overrides `CellStartedPre` back to `false`: there the running container is torn down and a fresh one started, so `started` remains correct (and the out-of-scope post-`daemon reset` create path is unaffected — it has no pre-existing cell, so `CellStartedPre` stays false → `started`).
- Aligned `docs/cli-use-cases.md`'s two idempotency invariants (§ Bootstrap & teardown and § Double `kuke init`) to state provisioning phases report `already existed` and the container-start phase reports `already running`.

AGENTS.md's `cell containers: started` sample tail (line 127) is left unchanged — it documents the fresh-create (post-`daemon reset`) path, which the issue explicitly marks out of scope (wording correct there).

## Test plan

- [x] `make test` (full CI suite: `test-root` + `test-kukebuild`) — all packages pass
- [x] `GOWORK=off go build ./...` and `go vet ./internal/controller/... ./cmd/kuke/init/...` (the Makefile exports `GOWORK=off`; bare workspace build fails on an unrelated `cgroups/v2` dep skew)
- [x] New unit tests in `internal/controller/bootstrap_cell_drift_test.go`:
  - `TestBootstrapCell_IdempotentRerunReportsAlreadyRunning` — Ready containers pre → `CellStartedPre=true`, `CellStarted=false` ("already running")
  - `TestBootstrapCell_RerunWithStoppedContainersReportsStarted` — Stopped containers pre → `CellStartedPre=false`, `CellStarted=true` ("started")
  - `TestBootstrapCell_ImageDriftRecreatesCell` extended: Ready pre but recreate forces `CellStartedPre=false` → `CellStarted=true` ("started")

## Acceptance criteria

- [x] Re-running `kuke init` on a healthy host reports an idempotent no-op wording for the cell-containers phase (no bare `started`) — `already running`.
- [x] Exit 0 and daemon-stays-up behavior unchanged (no code path other than the report flag was touched; StartCell's no-op idempotency is preserved).
- [x] `docs/cli-use-cases.md` invariant and the actual phase wording agree.

Closes #1186

### DIFF
--- a/docs/cli-use-cases.md
+++ b/docs/cli-use-cases.md
@@ -31,7 +31,7 @@ sudo kuke init --kukeond-image docker.io/library/<img>:<tag>
 - Exit code 0 on success; non-zero with a message naming the failed phase otherwise.
 - Side effect: `/opt/kukeon/data/{default,kuke-system}/...` populated; `/run/kukeon/{kukeond.sock,kukeond.pid}` created; containerd namespaces `default.kukeon.io` and `kuke-system.kukeon.io` exist; cgroup subtree `/kukeon/...` populated.
 - After success, `kuke get realms` and `kuke get realms --no-daemon` both list **at least** `default` and `kuke-system` in `Ready` state with their canonical namespaces. The two outputs must agree — divergence indicates the daemon's view of `/opt/kukeon` is stale (the bind-mount regression AGENTS.md guards against).
-- Second invocation on a healthy host is idempotent: every phase reports "already existed", exit code 0, the daemon stays up.
+- Second invocation on a healthy host is idempotent: each provisioning phase reports "already existed" and the container-start phase reports "already running" (no bare "started" — that wording is reserved for a phase that actually created or transitioned something), exit code 0, the daemon stays up.
 - `kuke doctor cgroups` on the same host exits 0 once cgroup controllers are delegated; non-zero output names which controller is missing and whether the kernel lacks support or the parent didn't delegate.
 
 ### Lightweight teardown (dev loop)
@@ -522,7 +522,7 @@ These are the negative paths most likely to surface a UX regression. Each is ver
 
 **Invariants.**
 
-- Idempotent. Every phase reports `already existed`; exit code 0; daemon stays up.
+- Idempotent. Each provisioning phase reports `already existed` and the container-start phase reports `already running` (never a bare `started` on a healthy re-run); exit code 0; daemon stays up.
 
 ### Image references
 
--- a/internal/controller/bootstrap.go
+++ b/internal/controller/bootstrap.go
@@ -668,7 +668,18 @@ func (b *Exec) bootstrapCell(section *CellSection, cellDoc *v1beta1.CellDoc) err
 			return fmt.Errorf("failed to check if root container exists: %w", rootErr)
 		}
 		section.CellRootContainerExistsPre = rootExistsPre
-		section.CellStartedPre = false
+		// Probe whether the cell's containers were already running before the
+		// EnsureCell+StartCell pass below. GetCell populated live container
+		// statuses off containerd, so this distinguishes a healthy idempotent
+		// re-run (containers already up → StartCell is a no-op, report "already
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #1203 Review — `fix(init): report 'already running' for idempotent re-run's cell-containers phase`**

LGTM — swaps the hardcoded `CellStartedPre = false` for a `cellContainersRunning` probe so a healthy `kuke init` re-run renders `already running` instead of a bare `started`. Verified the chain end-to-end: `GetCell` populates `Status.Containers` (keyed by `containerSpec.ID`, `containerd.Running`→`ContainerStateReady`), `cellContainersRunning` reads the same `spec.ID`, and `CellStarted = !pre && post` feeds `printStartAction`'s `started`→`already running` switch correctly across all four paths (healthy re-run, crash/stop, fresh create, image-drift override). New helpers are both used, commit is signed, tests cover the three scenarios, CI green.

Non-blocking note: the host-gated `make dev-init` smoke isn't in the test plan, but the change is report-flag-only (deterministic flag→string mapping, fully unit-pinned) with no daemon-runtime or `/opt/kukeon` effect, so the unit coverage is dispositive here — worth an eyeball at the CLI level on a re-run pre-merge if convenient, not a blocker.

PR #1205, opened 2026-06-10
fix: tear down BuildKit _history companion namespaces on uninstall and flag stray ones in status

### LINKED ISSUE #1183: uninstall: BuildKit '<ns>_history' companion namespaces survive full uninstall and are invisible to kuke status's residual-namespace check
## Context

`kuke build` (kukebuild, BuildKit-as-library) creates a companion containerd namespace `<ns>_history` for BuildKit's history store (buildkit `llbsolver/history.go` derives it as `ns + "_history"`). After building into the `kuke-system` realm, the host carries:

```
$ sudo ctr ns ls
default.kukeon.io
kuke-system.kukeon.io
kuke-system.kukeon.io_history
```

A full `kuke uninstall -y` removes the `*.kukeon.io` realm namespaces but leaves `kuke-system.kukeon.io_history` behind (verified post-uninstall on the nested dev host, client `v0.5.0-299-gab14245`). Additionally, `kuke status`'s residual-containerd-namespace check does not flag it — the suffix doesn't match the `<realm>.<suffix>` shape the check looks for — so the leak is invisible to the one tool that's supposed to catch residue from a half-cleaned uninstall.

## Proposal

1. `kuke uninstall` removes `<namespace>_history` alongside each realm namespace it tears down (draining its content/leases first, same as the realm namespace).
2. `kuke status`'s residual-namespace check also matches `*_history` companions of kukeon-managed namespaces.

## Acceptance criteria

- [ ] After `kuke build --realm kuke-system` + full `kuke uninstall -y`, `ctr ns ls` shows no `*.kukeon.io*` namespaces, including `_history` companions.
- [ ] `kuke status` flags a stray `<realm-ns>_history` namespace as residue when present.

## Out of scope

- The kukebuild cache dir under `/var/lib/kukebuild` (#904, closed — separate store).
- The metadata-independent snapshot drain on uninstall (separate issue).

## References

- BuildKit `solver/llbsolver/history.go` (`ns + "_history"`)
- #904 (stale kukebuild cache class), #1063 (local-build epic)

eminwux:
## Summary
- `kuke build` (kukebuild / BuildKit-as-library) creates a companion containerd namespace `<ns>_history` for BuildKit's history store (BuildKit derives it as `ns + "_history"`). A full `kuke uninstall -y` removed the `*.kukeon.io` realm namespaces but left `<realm-ns>_history` behind, and `kuke status`'s residual-namespace check never flagged it — the `_history` suffix doesn't match the `.kukeon.io` shape the check filters on (issue #1183).
- **Uninstall (AC 1):** `runner.PurgeRealm` now drains-then-deletes the `<namespace>_history` companion with the same `CleanupNamespaceResources` + `DeleteNamespace` the realm namespace gets. `DeleteNamespace` is idempotent (no-op when the realm was never built into), so this is safe on every purge. A companion that fails to delete folds into the returned `namespaceRemoved=false` so the uninstall half-cleaned-host gate (#287) still leaves the filesystem intact when containerd residue survives.
- **Status (AC 2):** `checkStateNamespaces` now includes `_history` companions in its candidate set; a live realm's companion is added to the expected set (no noise during normal operation), while a companion whose owning realm is gone falls through to the residual list and renders WARN.

## Design notes (for the reviewer to push back on)
- **Placement in `runner.PurgeRealm`, not just the uninstall loop.** Uninstall routes every realm through `controller.PurgeRealm` → `runner.PurgeRealm`, so this is the single cohesive teardown seam and satisfies AC 1. A side effect is that `kuke purge <realm>` (which shares the path) now also removes that realm's `_history` companion — this is correct: a deleted realm's build history should not survive, and it closes the same leak for the purge verb. Lower-blast-radius alternative (special-casing the uninstall loop only) would have left `kuke purge` leaking the companion.
- **Status flags a companion only when its realm is gone**, not whenever a `_history` namespace exists. Flagging it during normal operation (realm live, just built into) would WARN on every post-build dev host. The issue frames this as catching residue from a half-cleaned uninstall, so "stray" == owning realm absent.
- New consts helpers `BuildKitHistoryNamespace` / `IsBuildKitHistoryNamespace` honor the operator-configured `RealmNamespaceSuffix`, so the companion match works under a non-default suffix (e.g. `dev.kukeon.io`) too.

## Test plan
- [x] `GOWORK=off go test ./internal/consts/...` — new round-trip + suffix + foreign-base rejection tests for the history-namespace helpers.
- [x] `GOWORK=off go test ./cmd/kuke/status/...` — new `checkStateNamespaces` tests: stray companion → WARN naming it; live-realm companion → OK.
- [x] `GOWORK=off go test ./internal/controller/runner/...` — new `PurgeRealm` tests: companion is drained+deleted alongside the realm namespace; a companion delete failure folds into `namespaceRemoved=false` and surfaces in the error.
- [x] `make test-root` (full root CI test target, `GOWORK=off`) — exit 0, no failures.
- [x] `go vet` (affected packages) and `pre-commit run --files <changed>` (go-fmt, golangci-lint, addlicense) — all pass.
- [ ] `make dev-init` end-to-end smoke (`kuke build` + `kuke uninstall` + `ctr ns ls`) — not run: requires a host containerd at `/run/containerd/containerd.sock` and root; deferred to the reviewer's clean-host environment. The on-host `_history` leak is reproduced in the unit tests above against the real teardown seam.

## Acceptance criteria
- [x] After `kuke build --realm kuke-system` + full `kuke uninstall -y`, no `*.kukeon.io*` namespaces survive, including `_history` companions — covered by the `runner.PurgeRealm` teardown + unit test (CLI end-to-end deferred per test plan).
- [x] `kuke status` flags a stray `<realm-ns>_history` namespace as residue when present — covered by `checkStateNamespaces` + unit test.

Closes #1183

### DIFF
--- a/cmd/kuke/status/checks.go
+++ b/cmd/kuke/status/checks.go
@@ -358,10 +358,14 @@ func checkStateNamespaces(ctx context.Context, rc *runCtx) Result {
 	// Filter to namespaces under the kukeon suffix — that's what
 	// `kuke init` provisions and what `kuke uninstall` removes. Other
 	// namespaces on the same containerd (e.g. user docker namespaces)
-	// are not ours to police.
+	// are not ours to police. BuildKit history-store companions
+	// (<realm-ns>_history, created by `kuke build`) are included so a
+	// companion stranded by a half-cleaned uninstall surfaces as residue
+	// (issue #1183) — when its owning realm is still live the companion is
+	// added to the expected set below and renders clean.
 	var kukeonNS []string
 	for _, ns := range nsList {
-		if consts.IsKukeonNamespace(ns) {
+		if consts.IsKukeonNamespace(ns) || consts.IsBuildKitHistoryNamespace(ns) {
 			kukeonNS = append(kukeonNS, ns)
 		}
 	}
@@ -382,9 +386,15 @@ func checkStateNamespaces(ctx context.Context, rc *runCtx) Result {
 		return r
 	}
 
+	// A live realm is expected to carry both its namespace and — if it has
+	// ever been built into — that namespace's BuildKit history companion, so
+	// add both forms. A <realm-ns>_history whose owning realm is gone is not
+	// in this set and falls through to the residual list below (issue #1183).
 	expectedNS := make(map[string]bool, len(realms))
 	for i := range realms {
-		expectedNS[consts.RealmNamespace(realms[i].Metadata.Name)] = true
+		ns := consts.RealmNamespace(realms[i].Metadata.Name)
+		expectedNS[ns] = true
+		expectedNS[consts.BuildKitHistoryNamespace(ns)] = true
 	}
 
 	var residual []string
--- a/cmd/kuke/status/status_test.go
+++ b/cmd/kuke/status/status_test.go
@@ -144,6 +144,48 @@ func TestCheckParity_DaemonDownIsWARN(t *testing.T) {
 	}
 }
 
+// TestCheckStateNamespaces_StrayHistoryIsResidual pins issue #1183: a
+// BuildKit history-store companion (<realm-ns>_history) whose owning realm is
+// gone — the post-uninstall leak — surfaces as residue, even though it does
+// not carry the bare .kukeon.io suffix the residual check used to filter on.
+func TestCheckStateNamespaces_StrayHistoryIsResidual(t *testing.T) {
+	rc := &runCtx{
+		// No realms: mirrors a host after `kuke uninstall` removed the realm
+		// namespaces but left kuke-system.kukeon.io_history behind.
+		daemonClient: newFakeClient(),
+		ctrClient:    &fakeCtrClient{namespaces: []string{"kuke-system.kukeon.io_history"}},
+		logger:       testLogger(),
+	}
+
+	r := checkStateNamespaces(context.Background(), rc)
+	if r.Status != StatusWARN {
+		t.Fatalf("expected WARN on stray history namespace; got %s (%q)", r.Status, r.Detail)
+	}
+	if !strings.Contains(r.Detail, "kuke-system.kukeon.io_history") {
+		t.Errorf("Detail should name the stray history namespace; got %q", r.Detail)
+	}
+}
+
+// TestCheckStateNamespaces_LiveRealmHistoryIsClean confirms the companion is
+// not noise during normal operation: while the owning realm is live, its
+// <realm-ns>_history build byproduct is expected and renders OK.
+func TestCheckStateNamespaces_LiveRealmHistoryIsClean(t *testing.T) {
+	rc := &runCtx{
+		daemonClient: newFakeClient().withDefaultRealms(),
+		ctrClient: &fakeCtrClient{namespaces: []string{
+			"default.kukeon.io",
+			"kuke-system.kukeon.io",
+			"kuke-system.kukeon.io_history",
+		}},
+		logger: testLogger(),
+	}
+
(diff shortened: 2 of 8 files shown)


eminwux:
**PR #1205 Review — `fix: tear down BuildKit _history companion namespaces on uninstall and flag stray ones in status`**

Clean, well-tested fix. The `consts` helpers honor the custom suffix, the `PurgeRealm` drain-then-delete folds a surviving companion into `namespaceRemoved` so the #287 gate still suppresses filesystem teardown, and the `status` candidate/expected sets are symmetric so a live realm's companion stays quiet. `DeleteNamespace` idempotency (no-op on a never-built-into realm) is confirmed, so the per-purge call is safe. No leaks, no dead code. One blocking gap:

- **Sibling teardown gap** `internal/controller/runner/delete_realm.go:126`: `DeleteRealm` deletes the realm namespace but not its `<ns>_history` companion, so `kuke delete realm` on a built-into realm strands exactly the residue this PR's new `status` check (AC2) now WARNs on — the PR ships a detector for a leak its sibling teardown verb still creates. Mirror the `PurgeRealm` drain-then-delete here (or factor a shared helper) so both namespace-teardown seams stay consistent. Your own "a deleted realm's build history should not survive" rationale for folding in `purge` applies verbatim to `delete realm`.

Non-blocking: the `make dev-init` smoke is correctly deferred as host-gated; the unit tests reproduce the leak against the real teardown seam, so that's fine for merge.

LGTM once `delete_realm.go` extends the same `_history` teardown — or you justify scoping it out (but the purge rationale cuts against that).

eminwux:
Addressed in c9a26fb:
- **Sibling teardown gap (blocking):** `DeleteRealm` (`delete_realm.go`) now tears down the `<ns>_history` companion right after the realm namespace, so `kuke delete realm` no longer strands the exact residue AC2's new `status` check WARNs on. Factored the drain-then-delete into a shared `(*Exec).purgeBuildKitHistoryNamespace` helper (in `purge_realm.go`) and routed both `PurgeRealm` and `DeleteRealm` through it — the two namespace-teardown seams now stay consistent by construction. In `PurgeRealm` a surviving companion still folds into `namespaceRemoved=false`; in `DeleteRealm` it surfaces wrapped under `ErrDeleteRealm`.
- Added `TestDeleteRealm_DrainsAndDeletesBuildKitHistoryCompanion` and `TestDeleteRealm_HistoryCompanionFailureSurfacesError` mirroring the existing `PurgeRealm` history tests.

Verified locally (post-rebase onto `origin/main`, which carried #1204's `status` change): `GOWORK=off make test-root` exit 0, `go vet`, and `pre-commit run --files <changed>` (go-fmt, golangci-lint, addlicense) all pass.

Non-blocking `make dev-init` smoke stays host-gated/deferred as before — the unit tests reproduce the leak against the real teardown seams.

eminwux:
**PR #1205 Review — `fix: tear down BuildKit _history companion namespaces on uninstall and flag stray ones in status`**

LGTM — drains+deletes the `<realm-ns>_history` companion through the shared `PurgeRealm`/`DeleteRealm` teardown seams and flags stray ones in `kuke status`; `IsBuildKitHistoryNamespace` safely rejects foreign/bare bases, the companion-delete failure folds into `namespaceRemoved=false` to preserve the half-cleaned-host gate, and unit tests cover the real teardown seam. The unchecked `make dev-init` e2e smoke is host-gated (needs host containerd + root) — worth running on a clean host pre-merge, non-blocking per the documented deferral.

Issue #1137, opened 2026-06-07
Stale container-status timestamps: StartTime/FinishTime/RestartTime/RestartCount/ExitSignal never populated from containerd


eminwux:
## Summary

Container status in cell metadata persists zero-value / placeholder timestamps and counters even for healthy `Ready` containers. They are hardcoded behind long-standing TODOs and never populated from containerd.

## Reproduction

```
$ kuke get cell dezot-dev-0 -o yaml
status:
  containers:
    - name: work
      state: Ready
      createdAt: 2026-06-07T20:35:04.08029972Z
      restartCount: 0
      restartTime: 0001-01-01T00:00:00Z
      startTime: 0001-01-01T00:00:00Z
      finishTime: 0001-01-01T00:00:00Z
      exitCode: 0
      exitSignal: ""
```

`startTime` is `0001-01-01T00:00:00Z` despite the container being `Ready`.

## Root cause

`internal/controller/runner/helpers.go:740–752`, where container status is assembled:

```go
// TODO: Get additional status fields (RestartCount, StartTime, etc.) from containerd
status := intmodel.ContainerStatus{
    Name:         containerSpec.ID,
    ID:           containerSpec.ID,
    CreatedAt:    createdAt,
    State:        obs.State,
    RestartCount: 0,           // TODO: retrieve from containerd
    RestartTime:  time.Time{}, // TODO: retrieve from containerd
    StartTime:    time.Time{}, // TODO: retrieve from containerd
    FinishTime:   time.Time{}, // TODO: retrieve from containerd
    ExitCode:     obs.ExitCode,
    ExitSignal:   "",          // TODO: retrieve from containerd
}
```

`CreatedAt` is correctly stamped/preserved (issue #605); the five fields above are never wired. They flow through `internal/apischeme/scheme.go` serialization into the persisted YAML and into `kuke get container` (`cmd/kuke/get/container/container.go`).

## Affected fields

- `RestartCount`
- `RestartTime`
- `StartTime`
- `FinishTime`
- `ExitSignal`

## Suggested fix

[...]

PR #1208, opened 2026-06-10
fix: settle cell Failed on non-zero workload exit, distinct from clean Stopped

### LINKED ISSUE #1206: cell crash settles as `Stopped`, not `Error`/`Failed` — workload non-zero exit is indistinguishable from clean completion
## Summary

A cell whose workload/onInit **fails** (e.g. a missing binary, or an onInit step that exits non-zero) settles into cell state **`Stopped`** (`Synced`), the same terminal state as a clean exit. There is no `Error`/`Failed` state to distinguish a crash from a successful completion. A control plane watching cells therefore cannot tell "the agent finished its work" from "the workload crashed on startup" by cell state alone.

## Expected

A cell whose attachable workload exits non-zero (or whose onInit fails) should settle in a distinct **`Error`** / **`Failed`** state (with the exit code/signal surfaced), separate from the `Stopped` state used for a clean exit.

## Actual

The cell reaches `Stopped` / `Synced` regardless of the workload's exit status. Observed with kukeon-hermes dispatch: cells whose headless agent failed at startup ended `Stopped`, indistinguishable from cells that completed real work.

## Repro sketch

1. Materialise a cell whose attachable container command (or an onInit step) runs a missing binary / exits non-zero, e.g. `kuke run --from-config <cfg> --env ... -d` where the onInit calls a binary not present in the image.
2. `kuke get cell <name>` -> state settles `Stopped` / `Synced`, not `Error`.

## Impact

External orchestrators (e.g. the kukeon-hermes control plane) must fall back to scraping container logs and cross-checking external artifacts (GitHub issue/PR transitions) to infer failure, because the cell's own state lies about success. Surfacing `Error`/`Failed` with the exit code would let callers reconcile run outcomes reliably from state.

## Notes

Reported from operating the agent fleet; please confirm the exact non-zero-exit -> state mapping. Related downstream handling: kukeon-hermes now treats `Stopped` as inconclusive and composites cell state + agent result log + GitHub transition to decide a run's outcome.

eminwux:
## Summary

- A cell whose workload exits non-zero (or whose onInit fails) previously settled in `Stopped`/`Synced` — the same terminal state as a clean exit — so a control plane watching cells could not tell "the agent finished" from "the workload crashed on startup". This makes a non-zero terminal exit settle the cell in the distinct **`Failed`** state instead, surfacing the failing container + exit code/signal in `status.reason`/`status.message`. A clean (exit 0) completion still settles `Stopped`.
- **State mapping (the issue asked us to confirm it):** when every relevant container has terminally exited, the derivation now inspects the preserved exit triple — non-zero exit code (or an explicit `ContainerStateFailed` observation) → `CellStateFailed`; all-zero → `CellStateStopped`. Applied to both derivation paths (`deriveCellStateFromNonRootContainerStatuses` for user-workload cells, `deriveCellStateFromRootContainer` for root-only/kukeond cells). The exit code is already preserved across containerd task reaping by `populateCellContainerStatuses` (#1137), so a failed-then-reaped (`NotCreated`) workload still carries its code.
- **Reusing the existing `Failed` state, not a new `Error` state.** `#407` introduced `CellStateFailed` as "the terminal Error state" for startup-path failures, already **sticky** (`cellStateIsSticky`) and **excluded from auto-delete/wind-down** (`cellStateAutoDeleteTriggers`). Its documented intent — "`kuke run --rm` does not auto-clean a Failed cell ... preserving the cell on failure keeps the diagnostic surface intact" — is exactly what this issue wants, so a runtime workload exit now reaches the state that machinery was built for. A new runtime `reason` label `WorkloadFailed` (distinct from `StartCellFailed`/`CreateCellFailed`/`StartContainerFailed`) lets consumers tell a post-Ready crash from a failed bring-up.

## Behavior-change notes (non-obvious calls — reviewer please weigh in)

- **Failed cells are preserved, not reaped.** Because `Failed` is excluded from auto-delete/wind-down, a crashed cell is now kept (root container left alive for non-root-workload cells, no `--rm` cleanup) instead of being reaped to `Stopped`. This is essential to the issue's goal — if `--rm` deleted the cell, the caller could never read the `Failed` state — and it matches `#407`'s documented "preserve the diagnostic surface" intent. The operator clears it with `kuke delete cell`. `kuke run <cell>` already refuses a `Failed` cell with a delete pointer (`docs/cli-use-cases.md`), so re-run is handled.
- **Interaction with `#1003` restart-policy reap.** Previously `restartPolicy: on-failure`/`always` + non-zero exit permitted wind-down/auto-delete (reaped to `Stopped`/gone). Now the non-zero exit derives `Failed` first, which the trigger gate (Stopped-only) excludes, so the cell settles `Failed` + preserved. Clean exits are unchanged on every policy. I judged surfacing the crash as `Failed` to be the more-correct behavior and the issue's explicit ask, but flagging it since it inverts `on-failure`'s reap-on-failure outcome.
- `killCellLocked` hard-codes `State=Stopped` (kill.go:257), so wind-down must not fire on `Failed` (it doesn't — gated on the Stopped-only trigger); this is why preserving rather than winding-down is the correct integration.

## Docs touched (directly-related, not unrelated cleanup)

- `docs/site/concepts/cell.md` — `Failed`/`Stopped` lifecycle descriptions updated for the broadened `Failed` semantics.
- `docs/cli-use-cases.md` — `--rm` invariant clarified: cleanup fires only on a clean exit; a non-zero exit settles `Failed` (preserved).

## Test plan

- [x] `GOWORK=off go build ./...` (workspace split per `go.work` — `GOWORK=off` is the documented build mode)
- [x] `GOWORK=off go vet ./internal/controller/runner/`
- [x] `make test` equivalent — `GOWORK=off go test $(go list ./... | grep -v /e2e)` — 106 packages ok, exit 0
- [x] New unit tests: `deriveCellStateFromNonRootContainerStatuses` cases (non-zero exit / reaped-with-preserved-code / mixed clean+failed / failed-state / active-sibling-outranks), `containerExitedNonZero`, `rootContainerExitedNonZero`, `describeCellFailure`/`cellFailureBreadcrumb` (signal + plain code + root-only + fallback)
- [x] `make dev-init` smoke (change touches the daemon reconcile path) — exit 0, daemon-parity tail shows `default`/`kuke-system` both `Ready` with byte-identical daemon vs in-process views, attach smoke passed

Closes #1206

### DIFF
--- a/docs/cli-use-cases.md
+++ b/docs/cli-use-cases.md
@@ -296,6 +296,7 @@ sudo kuke run --clone <cell> --name custom                               # same,
 - `--container` is only valid in attach mode; passing both `--container` and `-d/--detach` exits non-zero.
 - `--rm` is processed by `kukeond`'s reconcile loop. `kuke run` is daemon-only after #566 — `KUKEON_NO_DAEMON=true` and `--run-path` promotion no longer reach an in-process branch for workload verbs, so `--rm` and `--run-path` are not mutually exclusive on `kuke run`. Cleanup latency is bounded by the daemon's reconcile interval (default 30s), not real-time.
 - A clean `^]^]` detach in attach mode does **not** trigger `--rm` cleanup; the cell stays alive for re-attach. Only workload termination, peer hangup, or an unrecoverable controller error fires cleanup.
+- `--rm` cleanup fires only on a **clean** workload exit (the cell settles `Stopped`). A workload that exits non-zero / crashes settles the cell `Failed` (#1206), which is excluded from auto-delete — the failed cell is preserved with its exit code/signal in `reason`/`message` so the failure can be inspected, and is cleaned up with an explicit `kuke delete cell`.
 - `kuke run -f /missing.yaml` exits non-zero with a `failed to open file` error.
 - A reference to an unavailable image surfaces the containerd resolver error verbatim (e.g. `pull access denied, repository does not exist or may require authorization`) and exits non-zero; the half-created cell may need `kuke purge cell` to clean up.
 
--- a/docs/site/concepts/cell.md
+++ b/docs/site/concepts/cell.md
@@ -65,8 +65,8 @@ A cell's **identity is its own document** — the `CellDoc`. The CellDoc owns th
 | --------- | --------------------------------------------------------------- |
 | `Pending` | Cell metadata exists; containers not yet created                |
 | `Ready`   | Root container is running; non-root containers are running      |
-| `Stopped` | All containers have exited                                      |
-| `Failed`  | The root container failed to start or exited unexpectedly       |
+| `Stopped` | All containers have exited cleanly (every exit code zero)       |
+| `Failed`  | Startup failed, or a workload exited non-zero / crashed; `reason` + `message` carry the failing container and its exit code/signal. Preserved (not auto-deleted) so the failure can be inspected — clear with `kuke delete cell` |
 | `Unknown` | The daemon can't determine the state (e.g., containerd offline) |
 
 ## Operations
--- a/internal/controller/runner/refresh.go
+++ b/internal/controller/runner/refresh.go
@@ -393,6 +393,17 @@ func (r *Exec) RefreshCell(cell intmodel.Cell) (intmodel.Cell, int, error) {
 		newStatus.State = r.deriveCellState(cell)
 	}
 
+	// Stamp the runtime-failure breadcrumb on a fresh transition to
+	// CellStateFailed (#1206), same contract as ReconcileCell. carryCellLifecycle
+	// above already copied any prior Reason/Message forward, so an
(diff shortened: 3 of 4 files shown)


eminwux:
**PR #1208 Review — `fix: settle cell Failed on non-zero workload exit, distinct from clean Stopped`**

LGTM — reuses the existing sticky/reap-excluded `CellStateFailed` for runtime non-zero exits, gated cleanly so `anyActive` still outranks a failed sibling, the sticky guard prevents downgrade and double-stamping, and the breadcrumb leverages `#1137`'s preserved exit code so reaped workloads still settle Failed. All 7 new symbols have callers, commits verified, CI green, and the host-gated `make dev-init` smoke is in the test plan.

PR #1209, opened 2026-06-10
fix(uninstall): force a synchronous GC sweep before deleting a drained namespace

### LINKED ISSUE #1182: uninstall: fails after 'kuke daemon reset --purge-system' — kuke-system namespace snapshots not drained when realm metadata is gone
## Context

`kuke uninstall -y` fails when run after the documented dev-loop teardown `kuke daemon reset --purge-system`:

```
sudo kuke daemon reset --purge-system     # exits 0; removes /opt/kukeon/data/kuke-system
sudo kuke uninstall -y                    # realm purge of kuke-system fails
```

Observed (nested dev daemon, client `v0.5.0-299-gab14245`): uninstall's purge of the `kuke-system` realm fails with containerd's namespace-removal error — the namespace still has snapshots on overlayfs (`namespace must be empty ... still has snapshots on overlayfs` class) — and uninstall reports the filesystem + user/group cleanup `skipped (realm purge failed)`, exit non-zero.

Root cause shape: `daemon reset --purge-system` deletes the realm's metadata tree (`/opt/kukeon/data/kuke-system`) but leaves the containerd namespace's snapshots/content (the kukeond image layers) in place. The subsequent `uninstall` purge path keys its snapshot/content drain on realm metadata that no longer exists, so it goes straight to namespace removal, which containerd refuses on a non-empty namespace. The recovery procedure in `docs/cli-use-cases.md` § *Full per-host teardown* (manual `ctr` snapshot cleanup + re-run) does work — but the first-pass failure is deterministic from two documented commands run back-to-back.

This is the half-cleaned-host class previously fixed for other shapes in #401 (stranded overlayfs snapshots in default realm) and #193 (namespace survives "purged" report).

## Proposal

Make uninstall's namespace teardown metadata-independent: before removing a `*.kukeon.io` namespace, drain its snapshots, content, leases, and images directly from containerd (the namespace name is known), regardless of whether realm metadata under `/opt/kukeon/data/<realm>` still exists. The half-cleaned host (metadata gone, namespace populated) should be a single-invocation uninstall, not a documented recovery procedure.

## Acceptance criteria

- [ ] `kuke daemon reset --purge-system` followed by `kuke uninstall -y` succeeds in one invocation, exit 0.
- [ ] After uninstall, `ctr ns ls` shows no `*.kukeon.io` namespaces.
- [ ] Existing single-invocation uninstall on a fully healthy host still passes (e2e #206 class).

## Out of scope

- BuildKit `*_history` companion namespaces surviving uninstall (separate issue).

## References

- `docs/cli-use-cases.md` § Full per-host teardown (recovery procedure)
- `cmd/kuke/uninstall/uninstall.go`
- #401, #193, #287 (prior half-cleaned-host fixes)

eminwux:
## Summary

`kuke daemon reset --purge-system` followed by `kuke uninstall -y` failed first-pass (exit non-zero) with:

```
namespace "kuke-system.kukeon.io" must be empty, but it still has snapshots on "overlayfs" snapshotter: failed precondition
```

`daemon reset --purge-system` removes the on-disk `kuke-system` metadata but leaves the containerd namespace (the staged kukeond image layers/snapshots) in place. uninstall then drains the namespace and deletes it — but the drain leaves exactly **one** straggler snapshot the explicit `snapshotter.Remove` pass cannot reclaim (a BuildKit "Active" snapshot from `kuke build` whose `Remove` transiently fails; the multi-pass loop logs `snapshot cleanup made no progress, giving up remaining=1`). That straggler is fully unreferenced (its leases and image are already drained), so containerd's *next periodic GC* reclaims it — but `DeleteNamespace` runs immediately and races it. The straggler is collected a moment later, which is why the documented "re-run" recovery worked and why a plain single-layer image never reproduced.

**Fix:** append a final forced synchronous GC sweep to `drainNamespaceResources` — create a throwaway lease and delete it with `leases.SynchronousDelete`, which blocks on the GC scheduler's `ScheduleAndWait`. This reclaims the straggler before `DeleteNamespace`, so the single invocation succeeds. The lease-first drain step couldn't cover this because it runs *before* the image/snapshot drain.

### Root-cause verification (live containerd)

Reproduced and fixed on a live containerd with the real BuildKit-built kukeond image (36 snapshots: 5 Active + 31 Committed). Instrumented probe of the drain:

- **before fix:** after-drain `snaps=1` (straggler) → `DeleteNamespace` fails → re-run succeeds (GC caught up).
- **after fix:** after-drain `snaps=1` → forced GC sweep → `snaps=0` → `DeleteNamespace` err=nil.
- End-to-end with the rebuilt `kuke`: a single `kuke uninstall --yes` against the BuildKit-populated namespace exits **0** and removes the namespace (plus its `_history` companion) in one pass.

A plain single-layer image (`hello-world`) drains cleanly on both old and new code — the bug requires the BuildKit Active-snapshot straggler.

### Notes for the reviewer

- `internal/ctr/namespaces.go`'s `drainImages` deletes images **without** `images.SynchronousDelete()` (unlike `internal/ctr/image.go:309`). The trailing forced sweep subsumes that gap; I left `drainImages` unchanged to keep the fix minimal (no scope creep).
- The forced sweep runs for every namespace drain, including the BuildKit `_history` companion (out of scope here per the issue; already handled by #1205) — it's an idempotent create+delete and a no-op on an already-empty namespace.
- The `docs/cli-use-cases.md` § *Full per-host teardown* recovery procedure remains valid for other genuine residual-namespace causes (e.g. a truly busy mount); it's no longer required for the two-command sequence, but I did not edit it to keep this a focused code fix.

## Test plan

- [x] `make test` (CI unit target: `test-root` + `test-kukebuild`) — passes.
- [x] `go vet ./...` (GOWORK=off) — clean.
- [x] New deterministic unit guard `TestCleanupNamespaceResourcesForcedGCSweepReclaimsStraggler` in `internal/ctr/namespaces_internal_test.go` — models the un-removable straggler and asserts the trailing forced sweep reclaims it (fails without the fix).
- [x] Existing `TestCleanupNamespaceResourcesDrainsLeasePinnedSnapshots` (#401) and `TestPruneLeasesDropsOrphansKeepsProtected` still pass after the shared-fake change.
- [x] Manual live-containerd repro + fix verification (BuildKit kukeond image): exit 1 before fix, exit 0 after; `ctr ns ls` shows no `*.kukeon.io` namespaces after.
- [ ] `make e2e` (new `TestKuke_DaemonResetPurgeSystem_ThenUninstall`, AC #1/#2; existing `TestKuke_Uninstall_VerifyState`, AC #3) — **not run locally: no docker daemon on the dev host** (`make e2e` requires docker for `docker save`). Both compile (`go vet ./e2e/` clean) and run in CI's `make e2e` job. Note: the e2e harness stages the image via `docker save`/`ctr import` (committed snapshots only), so it may not populate the BuildKit Active-snapshot straggler that triggers this bug on old code — the deterministic regression guard is the unit test above; the e2e test encodes the AC end-to-end and asserts single-pass success + no surviving namespaces, which hold on the fixed code.

## Acceptance criteria

- [x] `kuke daemon reset --purge-system` followed by `kuke uninstall -y` succeeds in one invocation, exit 0. (verified on live containerd)
- [x] After uninstall, `ctr ns ls` shows no `*.kukeon.io` namespaces. (verified on live containerd)
- [x] Existing single-invocation uninstall on a fully healthy host still passes (e2e #206 class). (additive fix — trailing sweep is a no-op on an empty/healthy namespace; `make test` green; `TestKuke_Uninstall_VerifyState` unchanged)

Closes #1182

### DIFF
--- a/e2e/e2e_kuke_test.go
+++ b/e2e/e2e_kuke_test.go
@@ -389,6 +389,78 @@ func TestKuke_DaemonReset_RoundTrip(t *testing.T) {
 	}
 }
 
+// TestKuke_DaemonResetPurgeSystem_ThenUninstall exercises issue #1182's AC:
+// `kuke init` → `kuke daemon reset --purge-system` → `kuke uninstall -y`
+// succeeds in a single invocation (exit 0) and leaves no `*.kukeon.io`
+// containerd namespaces behind.
+//
+// `daemon reset --purge-system` deletes the kukeond cell and the on-disk
+// kuke-system metadata tree, but the kuke-system containerd namespace (the
+// staged kukeond image layers / snapshots) survives. The bug was that the
+// subsequent uninstall drained that namespace but then raced the next
+// periodic GC on the immediately-following DeleteNamespace, failing the
+// first pass with "namespace must be empty ... still has snapshots on
+// overlayfs"; a re-run succeeded once GC caught up. The drain now forces a
+// synchronous GC sweep before DeleteNamespace, so the single invocation
+// suffices.
+//
+// Not run with t.Parallel(): like its siblings TestKuke_Uninstall_VerifyState
+// and TestKuke_DaemonReset_RoundTrip, it touches host-global state (every
+// kukeon-suffixed containerd namespace, /run/kukeon, /opt/kukeon, the kukeon
+// user/group) and must not race them.
+func TestKuke_DaemonResetPurgeSystem_ThenUninstall(t *testing.T) {
+	runPath := getRandomRunPath(t)
+	kukeondImage := loadKukeondImageIntoContainerd(t)
+
+	systemNS := consts.RealmNamespace(consts.KukeSystemRealmName)
+
+	t.Cleanup(func() {
+		// Best-effort: if the test exits before its own uninstall fires,
+		// re-run so the next test does not inherit kukeon-suffixed namespaces
+		// or the run path. Failures here are diagnostic only.
+		_, _, _ = runBinary(t, nil, kuke, append(
+			buildKukeRunPathArgs(runPath), "uninstall", "--yes",
+		)...)
+	})
+
+	// Step 1: init brings up both realms and the kukeond cell, staging the
+	// kukeond image into the kuke-system containerd namespace.
+	args := append(buildKukeRunPathArgs(runPath), "init", "--kukeond-image", kukeondImage)
+	exitCode, stdout, stderr := runBinary(t, nil, kuke, args...)
+	if exitCode != 0 {
+		t.Fatalf("kuke init failed: code=%d stdout=%s stderr=%s",
+			exitCode, string(stdout), string(stderr))
+	}
+
+	// Step 2: daemon reset --purge-system deletes the kukeond cell and the
+	// on-disk kuke-system metadata, but leaves the containerd namespace (with
+	// the staged image layers) in place — the precondition the bug needs.
+	args = append(buildKukeRunPathArgs(runPath), "daemon", "reset", "--purge-system")
+	exitCode, stdout, stderr = runBinary(t, nil, kuke, args...)
+	if exitCode != 0 {
+		t.Fatalf("kuke daemon reset --purge-system failed: code=%d stdout=%s stderr=%s",
+			exitCode, string(stdout), string(stderr))
+	}
+	if !verifyContainerdNamespace(t, systemNS) {
+		t.Fatalf("pre-uninstall: kuke-system namespace %q must survive daemon reset --purge-system "+
+			"(its image layers are what uninstall has to drain); got it already gone, "+
+			"so the single-invocation assertion below would pass vacuously", systemNS)
+	}
+
+	// Step 3: a SINGLE uninstall must succeed (exit 0) — issue #1182 AC #1.
+	args = append(buildKukeRunPathArgs(runPath), "uninstall", "--yes")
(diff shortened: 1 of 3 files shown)


eminwux:
**PR #1209 Review — `fix(uninstall): force a synchronous GC sweep before deleting a drained namespace`**

Correct, well-scoped fix. The trailing `forceGCSweep` (throwaway lease + `leases.SynchronousDelete`) runs last in `drainNamespaceResources`, reclaiming the BuildKit Active-snapshot straggler that the explicit Walk+Remove pass gives up on, before `DeleteNamespace`. Best-effort with WARN-on-failure, idempotent on empty namespaces. The unit guard models the gcOnly straggler and asserts the sweep is the *trailing* load-bearing step (lease.Create after the last snapshot Walk), so it fails without the fix. No unverified commits; no secrets in the new log lines. CI fully green, including `make e2e` running the new `TestKuke_DaemonResetPurgeSystem_ThenUninstall` (the unchecked test-plan box is "not run *locally*" — host-gated on docker — but CI exercised it).

One non-blocking nit:

- **Doc** `internal/ctr/namespaces.go:274-276`: the docstring says delete failure is logged at WARN (best-effort), then claims "The lease is always deleted, so it never itself blocks the namespace delete" — the two are slightly contradictory since a WARN'd delete failure leaves the lease in place. Reword to "best-effort; on the rare delete failure the namespace delete surfaces the real error" or similar. Cosmetic only.

LGTM to proceed.

PR #1210, opened 2026-06-10
fix(init): keep ownership sweep out of live per-cell kuketty tty state

### LINKED ISSUE #1207: fix(init): applyKukeonOwnership recursively chmods every LIVE cell tty socket to 0640 root — each kuke init re-breaks attach for all pre-existing cells
## Summary

Every `kuke init` run breaks attach for **every cell that already exists**. `applyKukeonOwnership` (`cmd/kuke/init/init.go:366`) calls:

```go
sysuser.ChownTreeAndChmod(runPath, 0, ensure.GID, kukeonRunPathDirMode /*02750*/, kukeonRunPathFileMode /*0640*/)
```

and `ChownTreeAndChmod` (`internal/sysuser/sysuser.go:196`) walks the **entire `/opt/kukeon` tree**, `Lchown`ing every inode to `root:kukeon` and chmodding every non-dir, non-symlink inode to `0640`. A live kuketty control socket (`…/work/tty/socket`, bound at `0660`, owned by the container's process uid) is just "a file" to that walker — so the sweep:

- strips group-write: `0660` → `0640` → every kukeon-group operator gets `connect: permission denied` (EACCES) on `kuke run`/`kuke attach`;
- resets the owner from the container uid back to `root`, undoing `attachablePostCreateChown`;
- also re-chowns the tty dir to root (`2750 root:kukeon`).

The listener stays alive inside the cell, so the socket is *not* stale — it's purely the mode bits.

## Smoking-gun evidence (kukeon-test host, 2026-06-09)

A single sweep timestamp is visible in ctime, identical to the millisecond across the tree root and the clobbered socket:

```
/opt/kukeon                                ctime 2026-06-09 17:52:54.663
dezot-dev-0 …/work/tty/socket  0640 root   ctime 2026-06-09 17:52:54.667   ← swept
agents-dev-0 socket            0660 inwx   ctime 2026-06-09 19:47:21       ← created after init: healthy
dezot-dev-1 socket             0660 inwx   ctime 2026-06-09 23:12:29       ← healthy
```

dezot-dev-0's socket was rebound at `0660` by a restart at 17:45:53 (kuketty metadata.json records `socketMode: 432` = 0o660, attach succeeded at 17:46) and a `kuke init` at 17:52:54 chmodded it to `0640 root:kukeon` while kuketty kept serving. Cells provisioned after 17:52 are all `0660`, uid-owned, healthy.

This is also the resolution of the "unreproducible" historical trigger in #1169: the broken cells were never provisioned broken — they were provisioned healthy and **broken later by the next `kuke init`** (`make dev-init` runs one every time). Cell-lifecycle reproductions could never show it because the mutator isn't a lifecycle path at all.

## Severity

On a dev host where `make dev-init` is routine, this re-breaks the whole running fleet (10+ agent cells here) on every iteration of the daemon dev loop. Combined with #1169 (`kuke run`/`attach` never heals a wrong-mode socket) the cells stay broken until each one is restarted or hand-chmodded.

## Immediate operator workaround

```bash
sudo chmod 0660 /opt/kukeon/data/<realm>/<space>/<stack>/<cell>/work/tty/socket
```

No restart needed — the listener is still live.

## Proposed fix

`applyKukeonOwnership`'s sweep must not descend into live cell work state. Options, roughly in order of preference:

1. **Exclude cell data from the recursive sweep.** Limit `ChownTreeAndChmod` to the init-owned skeleton (`/opt/kukeon` top level, `bin/`, instance metadata) and never recurse into `data/<realm>/<space>/<stack>/<cell>/work/` — that subtree is owned by the provisioning path (`attachableBuildOpts` + `attachablePostCreateChown` + kuketty), which already sets correct per-cell ownership.
2. **Skip socket inodes (and more generally anything not a regular file/dir) in `ChownTreeAndChmod`** — narrower, but still wrongly re-chowns tty dirs/capture/metadata away from the container uid, breaking kuketty's own metadata writes (`metadata write denied` warnings observed in kuketty.log after the sweep).
3. At minimum, make `kuke init` re-run `reapplyAttachableSocketPerms` for every Ready cell after the sweep — but this is heal-after-break and still flips ownership out from under running kukettys.

Option 1 + the #1169 attach-path heal gives defense in depth: init stops breaking sockets, and attach self-heals anything else that does.

## Related

- #1169 — `kuke run` Ready path never heals socket perms (this issue is the trigger that creates the broken state #1169 then fails to heal).
- #1170 — EACCES misclassified as transient stale-socket race (this is the EACCES it misdiagnoses).
- #935/#936, #933, #916 — earlier fixes in the same area; none cover the init sweep.

eminwux:
## Summary

- `applyKukeonOwnership` (`cmd/kuke/init/init.go`) ran `sysuser.ChownTreeAndChmod` over the **entire** `/opt/kukeon` tree, treating a live per-container kuketty attach socket as "just a file": it chmodded it `0o660 → 0o640` (stripping group-write → EACCES on `kuke attach`/`kuke run` for every kukeon-group operator) and chown'd it back to `root`, undoing `attachablePostCreateChown`. Because `make dev-init` runs `kuke init` on every iteration, each run re-broke attach for every pre-existing cell (#1207).
- Added `sysuser.ChownTreeAndChmodSkip` — `ChownTreeAndChmod` with an optional skip predicate that prunes a whole subtree (`filepath.SkipDir`) when it returns true for a directory. `ChownTreeAndChmod` now delegates with a nil predicate (behavior unchanged for all other callers — init was the only one).
- `applyKukeonOwnership` passes `skipLiveCellState`, which prunes per-container live state the runner owns at the container uid: the `tty/` directory subtree (live socket, capture, kuketty log, sbsh runtime metadata) and the rendered `kuketty-metadata.json`. The rest of the init-owned skeleton (`bin/`, instance metadata, realm/space/stack/cell metadata dirs) is still swept exactly as before, so first-init group-ownership setup is unchanged.

This is the issue's strongly-preferred **Option 1** (exclude live cell work state from the sweep), implemented against the real on-disk layout.

## AC-vs-codebase divergence (documented per dev CLAUDE.md)

The issue describes the clobbered socket as `…/work/tty/socket` and Option 1 as "never recurse into `data/<realm>/<space>/<stack>/<cell>/work/`". There is **no `work/` path component** in the actual layout (`internal/util/fs/metadata.go`): the live per-container state lives at `data/<realm>/<space>/<stack>/<cell>/<container>/tty/` (`fs.ContainerTTYDir` → `consts.KukeonContainerTTYDir = "tty"`), with the rendered config at `<container>/kuketty-metadata.json` (`ctr.AttachableMetadataFile`). The intent is unambiguous — keep the init sweep out of runner-owned live cell state — so I implemented against the real layout rather than the literal (non-existent) `work/` basename. Both excluded inodes are exactly what `attachablePostCreateChown` chowns to the container uid (`internal/controller/runner/attachable.go:505-551`).

## Scope

This is the **init-side** fix only (#1207). The complementary attach-path self-heal (#1169 — `kuke run`/`attach` never heals a wrong-mode socket) is a separate issue and is **not** included here; together they give the defense-in-depth the issue describes.

## Test plan

- [x] `make test` (full CI suite — `test-root` + `test-kukebuild`) — green.
- [x] `go vet ./internal/sysuser/ ./cmd/kuke/init/` — clean.
- [x] `pre-commit run --files <changed>` — all hooks pass (go fmt, golangci-lint, addlicense, …).
- [x] New unit test `TestChownTreeAndChmodSkip_PrunesSubtreeAndFile` (`internal/sysuser`) — confirms the skip predicate prunes a directory subtree and a single file while sweeping everything else.
- [x] New unit test `TestSkipLiveCellState_PreservesLiveSocketUnderInitSweep` (`cmd/kuke/init`) — builds a realistic RunPath (`data/…/<cell>/<container>/tty/socket` at `0o660` + rendered `kuketty-metadata.json`), drives the real `skipLiveCellState` predicate through the sweep, and asserts the live socket (`0o660`) and rendered metadata stay untouched while the daemon cell metadata is swept to the file mode. This is the regression observable (socket mode + ownership), directly testable via filesystem stat.
- [ ] `make dev-init` end-to-end smoke — **not run**: this host carries the live agent-cell fleet, and the issue documents that `make dev-init` is the exact mutator that re-breaks attach on every pre-existing cell. Running the destructive smoke here would disrupt unrelated production state (the very hazard this PR fixes); deferred to a clean host. The fix's observable (socket mode/ownership survives the sweep) is covered by the unit tests above.

Closes #1207

### DIFF
--- a/cmd/kuke/init/init.go
+++ b/cmd/kuke/init/init.go
@@ -363,13 +363,39 @@ func applyKukeonOwnership(
 		return r, fmt.Errorf("apply kukeon ownership to %q: %w", runDir, err)
 	}
 	r.RunDirApplied = true
-	if err := sysuser.ChownTreeAndChmod(
+	if err := sysuser.ChownTreeAndChmodSkip(
 		runPath, 0, ensure.GID, kukeonRunPathDirMode, kukeonRunPathFileMode,
+		skipLiveCellState,
 	); err != nil {
 		return r, fmt.Errorf("apply kukeon ownership to %q: %w", runPath, err)
 	}
 	r.RunPathApplied = true
 	return r, nil
+}
+
+// skipLiveCellState reports whether the recursive `kuke init` ownership sweep
+// must leave a path untouched. Per-container kuketty state is provisioned and
+// owned by the runner at the container's resolved process uid
+// (attachableBuildOpts + attachablePostCreateChown + kuketty), not by init:
+//
+//   - the tty directory (consts.KukeonContainerTTYDir) and everything inside
+//     it — the live attach socket bound 0o660, the capture transcript, the
+//     kuketty log, and kuketty's own runtime metadata it rewrites in place;
+//   - the rendered per-container kuketty metadata file
+//     (ctr.AttachableMetadataFile), chown'd to the container uid so the
+//     in-container wrapper can read it through the bind mount.
+//
+// The blanket sweep treated a live socket as "just a file": it chmodded it
+// 0o660 → 0o640 (stripping group-write, so every kukeon-group operator got
+// EACCES on `kuke attach`) and chown'd it back to root, undoing
+// attachablePostCreateChown. Because `make dev-init` runs `kuke init` on every
+// iteration, the sweep re-broke attach for every pre-existing cell each time
+// (issue #1207). Returning true for the tty directory prunes the whole subtree.
+func skipLiveCellState(_ string, info os.FileInfo) bool {
+	if info.IsDir() {
+		return info.Name() == consts.KukeonContainerTTYDir
+	}
+	return info.Name() == ctr.AttachableMetadataFile
 }
 
 // stageKukepause copies the kukepause binary into <runPath>/bin/kukepause and
--- a/cmd/kuke/init/init_test.go
+++ b/cmd/kuke/init/init_test.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"io"
 	"log/slog"
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -32,8 +33,11 @@ import (
 	initpkg "github.com/eminwux/kukeon/cmd/kuke/init"
 	kukshared "github.com/eminwux/kukeon/cmd/kuke/shared"
 	"github.com/eminwux/kukeon/cmd/types"
+	"github.com/eminwux/kukeon/internal/consts"
 	"github.com/eminwux/kukeon/internal/controller"
+	"github.com/eminwux/kukeon/internal/ctr"
 	"github.com/eminwux/kukeon/internal/errdefs"
+	"github.com/eminwux/kukeon/internal/sysuser"
 	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -78,6 +82,9 @@ func printCgroupAction(cmd *cobra.Command, label string, existedPre bool, exists
 
 //go:linkname printContainerAction github.com/eminwux/kukeon/cmd/kuke/init.printContainerAction
 func printContainerAction(cmd *cobra.Command, label string, existedPre bool, existsPost bool, created bool)
+
+//go:linkname skipLiveCellState github.com/eminwux/kukeon/cmd/kuke/init.skipLiveCellState
+func skipLiveCellState(path string, info os.FileInfo) bool
 
 //go:linkname printStartAction github.com/eminwux/kukeon/cmd/kuke/init.printStartAction
 func printStartAction(cmd *cobra.Command, label string, startedPre bool, startedPost bool, started bool)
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #1210 Review — `fix(init): keep ownership sweep out of live per-cell kuketty tty state`**

LGTM — the new `skipLiveCellState` predicate prunes exactly the inodes `attachablePostCreateChown` owns at the container uid (the `tty/` subtree via `filepath.SkipDir` + the rendered `kuketty-metadata.json`), so the idempotent `kuke init` sweep stops clobbering the live `0o660` attach socket while the init-owned skeleton (`bin/`, metadata dirs) is still swept unchanged. `ChownTreeAndChmod` cleanly delegates to the new `ChownTreeAndChmodSkip` with a nil predicate, and both unit tests reproduce #1207 by driving the real predicate through the real sweep against the actual on-disk layout and stat-checking socket mode/ownership.

Pre-merge note (non-blocking): the `make dev-init` end-to-end smoke is the documented host-gated carve-out — correctly deferred since dev-init is the destructive mutator that would re-break attach across this host's live cell fleet. Worth running on a clean host before merge to close the loop the unit tests can't reach.

PR #1220, opened 2026-06-11
fix(apply): re-materialize on kill→apply and fail when reconcile leaves cell Failed

### LINKED ISSUE #1185: apply: kill→apply recreates containers joined to the dead root container's namespaces (lstat /proc/<pid>/ns/ipc) → cell Failed; re-apply reports 'updated' exit 0 while cell stays Failed
## Context

`docs/cli-use-cases.md` § Workload lifecycle (apply invariants) pins:

> `apply` updates a divergent existing cell (e.g. after `kuke kill` left the root container missing) and reports `Cell <name>: updated` with a per-component summary.

Observed (nested dev daemon, client `v0.5.0-299-gab14245`), deterministic across two runs:

```
sudo kuke kill hello-world --realm default --space default --stack default   # exit 0
sudo kuke apply -f /tmp/hello.yaml
```

The first `apply` after `kill` fails: the recreated container is joined to the **dead** root container's namespaces, whose PID no longer exists:

```
lstat /proc/<dead-pid>/ns/ipc: no such file or directory
```

and the cell transitions to `Failed`. A second `kuke apply -f` then reports `Cell "hello-world": updated`, exit 0 — but the cell **stays Failed** and never re-materializes. So both halves of the invariant break: the documented kill→apply repair path errors, and the follow-up apply claims success while leaving the cell dead.

Root-cause shape: the apply/re-materialize path reuses the persisted root-container namespace references (`/proc/<pid>/ns/*` joins) instead of detecting that the root container's task is gone and recreating the sandbox first.

## Proposal

1. On apply against a cell whose root container task is dead, recreate the root container (fresh namespaces) before re-materializing dependent containers — the #486 re-materialize path needs to key on task liveness, not container-record existence.
2. `apply` must not report `updated` / exit 0 when the reconcile left the cell `Failed` — surface the failure and a non-zero exit.

## Acceptance criteria

- [ ] `kuke kill <cell>` → `kuke apply -f <same spec>` brings the cell back to `Ready`, exit 0, one invocation.
- [ ] `apply` that leaves a cell `Failed` exits non-zero and says so.
- [ ] Regression test covering kill→apply (the #486 phase-3 path).

## Out of scope

- `kuke run -f` warn-and-attach semantics on divergent state (post-#986 contract — held in testing).

## References

- `docs/cli-use-cases.md` § Workload lifecycle, apply-updates-divergent-cell invariant
- #486 (apply: re-materialize containers after kuke kill — the path that regressed)
- #867 (stop/kill containerd record semantics)

eminwux:
## Summary

Fixes the `kuke kill <cell>` → `kuke apply -f <same spec>` flow on two fronts (issue #1185):

- **AC1 — kill→apply now heals to Ready in one invocation.** Root cause (diverges from the issue's stated hypothesis — see below): the runner stamps every container's `cniConfigPath` at provision time (#1136), so a killed cell's persisted (actual) spec carries it while the user's re-applied YAML omits it. The **container-level** diff in `internal/controller/apply/diff.go` lacked the `desired != ""` guard the cell-level check already had, so that phantom field read as drift → `DiffCell` reported changes → apply routed through `UpdateCell`, which recreates the workload container joined to the **dead** root's namespaces (`lstat /proc/<dead-pid>/ns/ipc: no such file or directory`) and flips the cell to Failed. The guard now treats an empty desired `CNIConfigPath` as "runner will inject" rather than "user cleared it", the diff is empty, and the existing Stopped→re-materialize branch (`StartCell`, #486) brings the cell back to Ready.

- **AC2 — apply that leaves a cell Failed now exits non-zero and says so.** `ReconcileCell` previously returned a success action (`unchanged`/`updated`) with a nil error even when the cell ended Failed, so a follow-up apply printed `Cell "<name>": updated`, exit 0, while the cell stayed dead. A new `errdefs.ErrCellReconcileFailed` sentinel is now returned (via `failedCellReconcileError`) whenever reconcile leaves the resource in `CellStateFailed`, so `kuke apply` exits non-zero with an actionable `inspect with kuke get cell … -o yaml` message.

### Divergence from the issue's stated root cause
The issue hypothesised the re-materialize path "reuses persisted namespace references". Live reproduction showed the Stopped→`StartCell` re-materialize path actually works correctly — the real defect is the **diff misrouting** kill→apply to `UpdateCell` instead of re-materialize, caused by the runner-injected `cniConfigPath`. Intent (heal kill→apply) is unchanged; the fix lands one layer up from where the issue pointed.

### Design choice for AC2 (Failed → surface, not auto-recover)
`cellNeedsRematerialize` deliberately excludes `CellStateFailed` (a sticky startup-failure signal the user should inspect — see the existing contract + #486). Rather than weaken that to auto-recreate a Failed cell (which would mask the original failure cause recorded in `Status.Reason`/`Message`), AC2 is satisfied by making apply *surface* the Failed state as a non-zero exit. The kill→apply happy path (AC1) never reaches Failed because the diff fix routes it to re-materialize first.

## Refactor note
`ReconcileCell` exceeded `funlen`/`gocyclo` after adding the guard (it was already near the cap pre-change). Extracted three pure helpers — `ensureCellParents`, `createAndStartCell`, `appendContainerChangeSummaries` — and split the AC2 guard into a thin exported wrapper over an inner `reconcileCell` (avoids a `nonamedreturns` violation). No behavior change in the extracts.

## Test plan
- [x] `make test-root` (full CI target) — pass (`CITEST-EXIT=0`)
- [x] `go test ./internal/controller/apply/... ./internal/controller/` — pass
- [x] `pre-commit run --files <5 changed files>` — all hooks pass (golangci-lint incl. funlen/gocyclo/nonamedreturns)
- [x] New regression tests (AC3): `TestDiffCell_RunnerInjectedCNIConfigPath_NoChange`, `TestDiffCell_ExplicitCNIConfigPath_StillDetectsDrift` (diff guard), `TestReconcileCell_CNIConfigPathDriftRematerializes` (kill→apply re-materializes via StartCell, not UpdateCell), and updated `TestReconcileCell_FailedCellDoesNotRematerialize` (Failed → wrapped `ErrCellReconcileFailed`, StartCell still not called)
- [x] Runtime verification (earlier this session): fixed binary run as an isolated host process reproduced kill→apply healing to Ready. The `make dev-init` daemon smoke could not be re-run at PR time — the host's containerd bridge/networking was left broken by an earlier smoke attempt (no Docker Hub egress), and re-running `make dev-init`/`kuke daemon reset` would have further disrupted unrelated state on the shared host

Closes #1185

### DIFF
--- a/internal/controller/apply/diff.go
+++ b/internal/controller/apply/diff.go
@@ -794,7 +794,18 @@ func diffContainerSpec(desired, actual *intmodel.ContainerSpec, rootContainer bo
 		)
 	}
 
-	if desired.CNIConfigPath != actual.CNIConfigPath {
+	// Only a *non-empty* desired cniConfigPath counts as a change — same
+	// guard the cell-level spec.cniConfigPath diff above applies. The runner
+	// stamps every container's cniConfigPath at provision time
+	// (ResolveSpaceCNIConfigPath, #1136), so the persisted/actual container
+	// carries it while a user-authored YAML omits it. Without this guard the
+	// omitted (empty) desired value reads as "clear the path", reporting a
+	// spurious `cniConfigPath` change on every `kuke apply -f` of an unchanged
+	// file. That phantom diff routes a post-`kuke kill` apply through UpdateCell
+	// instead of the re-materialize branch, which recreates the workload
+	// container against the dead root's namespaces and fails the cell
+	// (lstat /proc/<dead-pid>/ns/ipc) — issue #1185.
+	if desired.CNIConfigPath != "" && desired.CNIConfigPath != actual.CNIConfigPath {
 		result.HasChanges = true
 		if result.ChangeType == ChangeTypeNone {
 			result.ChangeType = ChangeTypeCompatible
--- a/internal/controller/apply/diff_test.go
+++ b/internal/controller/apply/diff_test.go
@@ -337,6 +337,115 @@ func TestDiffCell_SynthesizedRoot_NoChange(t *testing.T) {
 	}
 }
 
+// TestDiffCell_RunnerInjectedCNIConfigPath_NoChange pins issue #1185: the
+// runner stamps every container's cniConfigPath at provision time
+// (ResolveSpaceCNIConfigPath, #1136), so the persisted/actual container carries
+// it while a user-authored YAML omits it (empty). An omitted desired value must
+// NOT read as a change — same guard the cell-level spec.cniConfigPath diff
+// applies. Without it, every `kuke apply -f` of an unchanged file reports a
+// spurious `cniConfigPath` diff, which (post-`kuke kill`) routes apply through
+// UpdateCell instead of the re-materialize branch and fails the cell against
+// the dead root's namespaces.
+func TestDiffCell_RunnerInjectedCNIConfigPath_NoChange(t *testing.T) {
+	desired := intmodel.Cell{
+		Metadata: intmodel.CellMetadata{Name: "hello-world"},
+		Spec: intmodel.CellSpec{
+			RealmName: "default",
+			SpaceName: "default",
+			StackName: "default",
+			Containers: []intmodel.ContainerSpec{
+				{ID: "root", Root: true, Image: "busybox:latest"},
+				{ID: "web", Image: "busybox:latest"},
+			},
+		},
+	}
+
+	actual := intmodel.Cell{
+		Metadata: intmodel.CellMetadata{Name: "hello-world"},
+		Spec: intmodel.CellSpec{
+			RealmName: "default",
+			SpaceName: "default",
+			StackName: "default",
+			Containers: []intmodel.ContainerSpec{
+				{
+					ID:            "root",
+					Root:          true,
+					Image:         "busybox:latest",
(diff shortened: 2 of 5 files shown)


eminwux:
**PR #1220 Review — `fix(apply): re-materialize on kill→apply and fail when reconcile leaves cell Failed`**

LGTM — the AC1 container-level `desired.CNIConfigPath != ""` guard (`internal/controller/apply/diff.go:808`) correctly mirrors the existing cell-level guard at `:159`, killing the phantom-drift misroute to `UpdateCell`; AC2's `failedCellReconcileError` passes real errors through untouched and only converts the nil-err + `CellStateFailed` case into the wrapped `ErrCellReconcileFailed` sentinel, which `ApplyDocuments` (`internal/controller/apply.go:301`) already stamps `failed` + non-zero exit. Refactor extracts are behavior-preserving, all new symbols are referenced, commits are verified, and the regression tests exercise both the heal path and the Failed-surface path.

Pre-merge note (non-blocking): the `make dev-init` daemon-parity smoke is documented as unrunnable on the PR host (broken containerd bridge, no egress). AC1's kill→apply heal was verified via the fixed binary as an isolated host process — worth re-running the full daemon smoke before merge if a healthy host is available.

PR #1221, opened 2026-06-11
fix(runner): skip CNI detach for host-network root cells and guard self-netns DEL

### LINKED ISSUE #1219: runner: stop/kill CNI detach targets the daemon's own netns for HostNetwork cells — nested `kuke daemon reset` deletes the parent dev cell's eth0 uplink
## Summary

`StopCell`/`KillCell` issue a CNI DEL for **every** cell's root container, including `HostNetwork` cells (the kukeond system cell). For a HostNetwork cell the root task's netns **is the daemon host's netns**, so the DEL runs the plugin chain against the host netns itself: the bridge plugin deletes the interface named `eth0` there (`IfName` is hard-coded), and the loopback plugin sets `lo` DOWN.

On a bare-metal host this is latent (NICs are rarely named `eth0`, physical links can't be netlink-deleted, and the next `kuke init` re-ups `lo`). **In nested mode — the canonical agent workflow of running `make dev-init` inside a `kukeon-dev-root` cell — the "host netns" is the parent cell's netns, and its uplink veth is literally named `eth0`.** The dev-init reset phase (`kuke daemon reset`) therefore deletes the parent cell's uplink and downs its `lo`, permanently severing the cell's network while its workload keeps running.

The ADD side already has the guard the DEL side is missing — `rootContainerWantsCNI` (`internal/controller/runner/start.go:239-241`) skips CNI attach for HostNetwork containers, and its comment even names this hazard class:

> Host-network containers (kukeond and any future host-scope cell) have no per-container veth, so attaching them to a CNI bridge would either error or — worse — create a host-side bridge inside the daemon's own netns

The detach side (`detachRootContainerFromNetwork`, called from `stop.go:200` and `kill.go:178`) has no such guard, and since #715/#724 the DEL is deliberately issued **unconditionally** so dead-task teardowns still release IPAM.

## Incident (2026-06-11, cell `default/default/default/kukeon-dev-0`)

Observed state inside the cell: no `eth0` at all, `lo` DOWN, leftover nested `k-*` bridges, `bash: ip: command not found` masking diagnosis. Host side: the cell's veth gone, but bridge `k-703e8b3a`, the nft per-IP chain for `10.88.0.69`, and the host-local IPAM lease (`default_default_kukeon-dev-0_root eth0`) all intact — i.e. nothing on the parent host tore the cell down; the deletion came **from inside the cell's own netns**.

Timeline from the nested `/tmp/devinit.log` + nested containerd log:

- Jun 10 23:14 — a nested `kuke init` creates a nested kukeond cell (HostNetwork → its root task shares the parent cell's netns).
- Jun 11 08:45:56 — next `make dev-init` run: `==> Reset prior kukeond cell` → `kukeond stopped` / `kukeond cell deleted`. `StopCell` calls `detachRootContainerFromNetwork` **before** stopping; the root task is alive, so `rootContainerNetnsPath` returns `/proc/<pid>/ns/net` = the parent cell's netns. Bridge DEL deletes `eth0`; loopback DEL downs `lo`.
- Jun 11 08:45:57 — the very next dev-init phase fails: `kukebuild: ... dial udp 1.0.0.1:53: connect: network is unreachable`.

The damage signature (eth0 deleted + lo downed + bridges left behind) is exactly the CNI plugin chain's DEL semantics, and the netns-sharing was verified live on the host: the host kukeond cell root task (`kukeon_kukeon_kukeond_root`) runs in `net:[4026531840]` == host netns.

## Root cause

1. `internal/controller/runner/stop.go:200` and `kill.go:178` call `detachRootContainerFromNetwork` for every cell — no `HostNetwork` check mirroring `rootContainerWantsCNI`.
2. `rootContainerNetnsPath` (`helpers.go:~525`) happily returns the live root task's netns — which for HostNetwork cells is the daemon's own host netns.
3. `buildRuntimeConf` (`internal/cni/container.go:~99`) hard-codes `IfName: "eth0"`, so the bridge plugin's DEL deletes whatever interface is named `eth0` in that netns.
4. libcni proceeds with the DEL even though no cached ADD result exists (missing cache file → `nil, nil`), so "never attached" does not protect the netns.
5. The conflist includes the loopback plugin, whose DEL sets `lo` DOWN in the target netns.

Note this is **not** the empty-netns path #715 reasoned about (plugins v1.5.1 no-op on `Netns == ""`); it is the *live-task* path handing the plugins the daemon's own netns.

## Why prior runs didn't break

The detach only fires destructively when a **prior** kukeond cell with a live root task exists at reset time. First-ever nested dev-init (no prior cell) is safe; every subsequent reset against a live nested kukeond cell severs the parent uplink. (An earlier failure in the same cell at Jun 10 21:56 was an unrelated upstream DNS timeout — the source IP `10.88.0.69` was still alive then.)

## Impact

- Any nested `make dev-init` re-run inside a dev cell with a previously-inited nested kukeond kills that cell's network until manually re-plumbed. Work in progress in the cell loses git/registry/DNS access.
- Latent host hazard on bare metal: hosts whose primary NIC is actually named `eth0` (common on cloud images) get `ip link del eth0` attempted against it (saved only by netlink refusing to delete physical links — veth/virtual `eth0` uplinks of VMs/containers running kukeond *would* be deleted), and host `lo` is downed on every `kuke daemon reset` until the next init's ADD path re-ups it.

## Manual recovery (applied to heal the live cell, no restart)

```sh
# parent host; $PID = cell root task pid, IP/bridge from IPAM lease + cell status
sudo ip link add vethkdev0heal type veth peer name eth0heal
sudo ip link set eth0heal netns $PID
sudo ip link set vethkdev0heal master k-703e8b3a up
sudo nsenter -t $PID -n ip addr del 10.88.0.1/24 dev k-703e8b3a   # stale nested bridge addr shadowing the subnet route
sudo nsenter -t $PID -n ip link set eth0heal name eth0
sudo nsenter -t $PID -n ip link set lo up
sudo nsenter -t $PID -n ip link set eth0 up
sudo nsenter -t $PID -n ip addr add 10.88.0.69/24 dev eth0
sudo nsenter -t $PID -n ip route add default via 10.88.0.1
```

## Proposed fix

1. **Mirror the ADD guard on the DEL side:** skip `detachRootContainerFromNetwork` (and the netns-bearing purge variants) when the cell's root container has `HostNetwork=true` — same predicate as `rootContainerWantsCNI`. The IPAM-file purge safety net can stay (it never touches a netns).
2. **Defense in depth in `detachRootContainerFromNetwork`:** before issuing a DEL with a non-empty netns, stat the target netns and the daemon's own `/proc/self/ns/net`; if they are the same inode, drop the netns (issue the empty-netns DEL instead). This protects against any future path that resolves a host/self netns, not just HostNetwork cells.
3. **Regression guard in `scripts/dev-init.sh` (nested mode):** after the `kuke daemon reset` phase, assert the parent uplink is still present (`ip -4 addr show eth0` / default route exists) and fail fast with a pointed message — today the breakage surfaces two phases later as a confusing registry DNS error. Mirrors the existing "parent attach plumbing still intact" post-flight.
4. Unit test: HostNetwork root cell → stop/kill must not invoke `DelContainerFromNetwork` with a netns equal to the runner's own.

## Related

- #715 / #724 — made the DEL unconditional (correct goal, missing the HostNetwork carve-out).
- #99 — moved kukeond onto the host netns (which is what the DEL now targets).
- #1079 — different nested-network bug (subnet shadowing), same blast zone.

eminwux:
## Summary

`StopCell`/`KillCell` (and the other teardown paths) issued a CNI DEL for **every** cell's root container, including `HostNetwork` cells (the kukeond system cell). For a HostNetwork root the live task's netns **is the daemon host's netns**, so the conflist DEL ran against the host netns: the bridge plugin deletes the interface hard-coded as `eth0`, and the loopback plugin sets `lo` DOWN.

On bare metal this is latent. **Nested** — the canonical agent workflow of `make dev-init` inside a `kukeon-dev-root` cell — the "host netns" is the parent cell's netns and its uplink veth is literally named `eth0`, so the dev-init reset phase (`kuke daemon reset`) deleted the parent cell's uplink and downed `lo`, severing the cell's network while its workload kept running (issue #1219).

The ADD side already had the guard the DEL side was missing (`rootContainerWantsCNI`, `start.go`). This PR adds the symmetric DEL-side protection in two layers, plus a nested-mode regression guard and unit tests.

## What changed

1. **Explicit HostNetwork skip at the named call sites** (`stop.go`, `kill.go`) — new `rootContainerHostNetwork(cell)` predicate mirrors `rootContainerWantsCNI`; a host-network root cell skips `detachRootContainerFromNetwork` entirely (it has no per-veth and no IPAM lease, since ADD was skipped). Logs `skipping CNI detach for host-network root cell`.
2. **Defense-in-depth self-netns guard** (`helpers.go`) — new `sanitizeDELNetns` / `delNetnsIsRunnerOwn`: before any DEL, if the target netns is the same `(dev, inode)` as `/proc/self/ns/net`, drop it to the empty-netns DEL (still netns-optional — releases IPAM + chained-plugin resources). Wired into **all three** netns-bearing DEL chokepoints: `detachRootContainerFromNetwork`, `purgeCNIForContainer`, and `detachRootContainerFromCNI` (`lifecycle.go`). This covers the recreate/purge/delete paths and any future caller that resolves a self netns — without threading a flag through every site.
3. **Nested-mode regression guard in `scripts/dev-init.sh`** — after the `kuke daemon reset` phase, assert the dev cell's uplink (`eth0` + a default route) survived the reset and fail fast with the real cause; today the break surfaces two phases later as a confusing `kukebuild: ... network is unreachable`. Snapshots pre-reset state and only flags the present→absent transition. Falls back to sysfs + `/proc/net/route` when `ip` is absent (the incident saw `ip` missing post-break). Mirrors the existing `verify_parent_attach_intact` post-flight; only fires nested.
4. **Unit tests** (`detach_network_test.go`) — `TestDetachRootContainerFromNetwork_DropsRunnerOwnNetns` pins the root-task PID to the test process so the resolved netns *is* the runner's own, and asserts (via a netns-recording fake CNI plugin) that the DEL carries an empty netns; plus direct tests for `delNetnsIsRunnerOwn` and `rootContainerHostNetwork`.

### Non-obvious calls (for reviewer push-back)

- **Why both layers (issue items 1 + 2):** the explicit skip gives intent symmetry with the ADD guard at the two incident paths; the inode guard is the universal safety net. The inode guard alone would functionally fix the bug everywhere (a HostNetwork root's netns *is* the runner's own → dropped); item 1 is the readable, log-clear guard the issue names first.
- **recreate/purge/delete paths** are covered by the layer-2 chokepoint guard (item 2) rather than an explicit per-site HostNetwork skip — issue item 1 names only `stop.go:200` / `kill.go:178`, so the call-site skip is scoped to those; everything else rides the universal `sanitizeDELNetns`.
- **`syscall.Stat` for inode compare** — std, depguard-allowed; `(Dev, Ino)` is the canonical netns-identity test. Fails open (keeps the netns) on any stat error to preserve pre-existing DEL behavior for every non-self path.

## Test plan

- [x] `GOWORK=off make test` equivalent — `go test $(go list ./... | grep -v /e2e)` (root) **and** `cd cmd/kukebuild && go test ./...` — both pass. (`go.work` workspace mismatch breaks bare `go build`; `GOWORK=off`, which the Makefile exports, is green.)
- [x] `go vet` + `gofmt -l` clean on all changed Go files.
- [x] New unit tests pass: `TestDetachRootContainerFromNetwork_DropsRunnerOwnNetns`, `TestDelNetnsIsRunnerOwn`, `TestRootContainerHostNetwork`, plus the existing `detach_network_test.go` guards.
- [x] `bash -n scripts/dev-init.sh` clean; `detect_uplink_state` / `verify_parent_uplink_intact` exercised standalone (this dev container is itself nested — probe present, `eth0` + default route detected, guard reports uplink intact).
- [ ] Full `make dev-init` nested smoke not run end-to-end here — it rebuilds/reinits the daemon on this live nested host, which would disrupt the running session; the new guard is unit/standalone-verified and the reset-phase assertion is non-destructive by construction.

Closes #1219

### DIFF
--- a/internal/controller/runner/detach_network_test.go
+++ b/internal/controller/runner/detach_network_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/containerd/containerd/v2/pkg/cio"
 	"github.com/eminwux/kukeon/internal/cni"
 	"github.com/eminwux/kukeon/internal/errdefs"
+	intmodel "github.com/eminwux/kukeon/internal/modelhub"
 )
 
 // deadTaskContainer is a near-empty containerd.Container whose Task() reports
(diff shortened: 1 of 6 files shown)


eminwux:
**PR #1221 Review — `fix(runner): skip CNI detach for host-network root cells and guard self-netns DEL`**

LGTM. Correct, well-layered fix: the call-site `rootContainerHostNetwork` skip mirrors the ADD-side `rootContainerWantsCNI` guard, and the inode-based `sanitizeDELNetns` chokepoint — wired into all three netns-bearing DEL paths (`detachRootContainerFromNetwork`, `purgeCNIForContainer`, `detachRootContainerFromCNI`) — is a sound universal safety net: even if the call-site predicate misses a flipped-default-root edge, the `(Dev, Ino)` self-netns check still drops the DEL. The persisted root `HostNetwork` (`provision.go:2218`) keeps the two layers consistent. Tests exercise the real path (the recording fake plugin asserts the DEL carries an empty netns, not a static trace); all commits verified; CI green. No secrets, no dead code.

Non-blocking — the unchecked `make dev-init` nested smoke is host-gated (running it tears down this live nested session), so it can't be ticked on either side of the loop. Please run it pre-merge to confirm the `verify_parent_uplink_intact` guard fires end-to-end against a real `kuke daemon reset`.

PR #1223, opened 2026-06-11
feat: render focused diff for divergent args/command/env notices

### LINKED ISSUE #1193: run/apply: divergence notice dumps full both-sides args (multi-line scripts) — render a focused diff so the changed token is visible
## Context

When `kuke run -f` (or `kuke apply` preview) detects a divergent existing cell, the divergence notice prints the changed fields — but for `args` it embeds the **entire** argument list of both sides inline, including multi-line heredoc-style shell payloads. With a typical `args: ["-c", "<multi-line script>"]` container, the notice becomes a wall of escaped shell text in which the actual changed token (e.g. `-p 8080` → `-p 8081`) is buried and nearly impossible to spot.

Observed during the workload-lifecycle sweep (client `v0.5.0-299-gab14245`) with a cell whose args carry a multi-line `sh -c` script: the divergence warning spanned dozens of wrapped lines for a one-token change. The documented invariant (warn-and-attach, naming the diverging fields — post-#986, field-naming per #469) held; this is UX friction, not a broken invariant.

## Proposal

Render the args divergence as a focused diff rather than full both-sides dumps: elide common prefix/suffix tokens, show only the differing elements (with index), and truncate long individual elements around the first difference (`…-p 8080…` vs `…-p 8081…`). Same treatment for `command` and `env`.

## Acceptance criteria

- [ ] A one-token args change on a cell with multi-line script args produces a divergence notice where the changed token is visible without manual diffing.
- [ ] Identical leading/trailing args elements are elided or summarized.
- [ ] The full spec remains inspectable via `kuke get cell -o yaml` (notice doesn't need to carry it).

## Out of scope

- The warn-and-attach vs refuse semantics (post-#986 contract — held).

## References

- `docs/cli-use-cases.md` § Workload lifecycle, divergent-spec invariants
- #469 (name the breaking-change fields instead of `[]`), #986 (warn-and-attach default)

eminwux:
## Summary
- The `kuke run -f` / `--require-synced` divergence notice rendered `args`/`env`/`command` divergences by dumping the **entire** both-sides value via `%q`. With a multi-line `sh -c` script in `args`, the changed token (e.g. `-p 8080` → `-p 8081`) was buried in dozens of wrapped lines of escaped shell text (#1193).
- New focused-diff formatters in `cmd/kuke/run/run.go` replace the verbatim dumps for `args`, `env` (slices) and `command` (scalar):
  - **Slices** (`focusedSliceDiff`): elide identical leading/trailing elements (summarized as `N leading, M trailing identical`), name only the differing index, and window long elements around the first differing rune.
  - **Scalar** (`focusedScalarDiff`): window long values around the first differing rune; short values still print verbatim.
  - Add/remove (length-differing) slices name the changed index range and per-element-truncated contents.
- Scope is limited to `args`/`command`/`env` per the AC ("Same treatment for command and env"). Short identifier fields (`ports`, `networks`, `image`, `workingDir`, `restartPolicy`) keep the existing verbatim `sliceDiff`/`scalarDiff` rendering — they don't carry multi-line payloads, so the wall-of-text problem doesn't apply.

Example (multi-line script, one-token change):

```
before: args (actual=["-c" "set -e\nwhile true; do\n  nc -l -p 8080 -e /bin/cat\n  sleep 1\ndone\n"], desired=["-c" "set -e\nwhile true; do\n  nc -l -p 8081 -e /bin/cat\n  sleep 1\ndone\n"])
after:  args (1 leading, 0 trailing identical; [1] actual=…" true; do\n  nc -l -p 8080 -e /bin/cat\n  sleep 1\n"… desired=…" true; do\n  nc -l -p 8081 -e /bin/cat\n  sleep 1\n"…)
```

Notes for the reviewer:
- The overall notice shape (`notice: cell "<name>" is OutOfSync with on-disk spec (diverging: <fields>); …`) is unchanged — only the per-field rendering inside `<fields>` changes. `docs/cli-use-cases.md` documents that as `(diverging: ...)` with an abstract placeholder, so no doc update is required.
- AC item 3 (full spec inspectable via `kuke get cell -o yaml`) needs no code change — that path is untouched; documented in the focused-diff godoc.

## Test plan
- [x] `make test` (full CI suite: `test-root` + `test-kukebuild`) — green.
- [x] `GOWORK=off go test ./cmd/kuke/run/` — green, including new `focused_diff_test.go` covering: multi-line-script one-token change keeps the token visible and is shorter than the full dump; common-framing elision; add/remove index ranges; scalar verbatim-vs-truncated; and the end-to-end `divergedContainerFields` integration.

Closes #1193

### DIFF
--- a/cmd/kuke/run/focused_diff_test.go
+++ b/cmd/kuke/run/focused_diff_test.go
@@ -1,0 +1,151 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package run
+
+import (
+	"strings"
+	"testing"
+
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
+)
+
+// TestFocusedSliceDiffMultiLineScriptOneTokenChange is the AC's headline case
+// (issue #1193): a one-token change inside a multi-line `sh -c` script must
+// surface the changed token without the operator hand-diffing two walls of
+// escaped shell text.
+func TestFocusedSliceDiffMultiLineScriptOneTokenChange(t *testing.T) {
+	script := func(port string) string {
+		return "set -e\n" +
+			"while true; do\n" +
+			"  echo serving on " + port + "\n" +
+			"  nc -l -p " + port + " -e /bin/cat\n" +
+			"  sleep 1\n" +
+			"done\n"
+	}
+	actual := []string{"-c", script("8080")}
+	desired := []string{"-c", script("8081")}
+
+	got := focusedSliceDiff("args", actual, desired)
+
+	// The changed token on both sides must be visible.
+	if !strings.Contains(got, "8080") || !strings.Contains(got, "8081") {
+		t.Errorf("focused diff omits the changed token:\n%s", got)
+	}
+	// The common leading element ("-c") must be summarized, not dumped.
+	if !strings.Contains(got, "1 leading") {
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1223 Review — `feat: render focused diff for divergent args/command/env notices`**

LGTM — focused-diff formatters for `args`/`command`/`env` correctly elide common framing, window long elements around the first differing rune (rune-based, multibyte-safe, `…` only on real elision), and keep the changed token visible; scope matches the AC, all new symbols are used, and the colocated tests cover the headline multi-line-script case plus elision/add-remove/scalar branches.

PR #1232, opened 2026-06-11
release: ship kukebuild in release artifacts + installer

### LINKED ISSUE #1227: release: ship kukebuild in release artifacts + installer
## Context

`kuke build` is kukeon's docker-free image build path: it execs the standalone `kukebuild` binary, which embeds BuildKit as a library and writes the image straight into the target realm's containerd namespace — no docker daemon, no standalone buildkitd, only host containerd. This is a load-bearing positioning point: kukeon as a Docker replacement that does not need docker even for building images.

Today that story only works for source builds. The release workflow (`.github/workflows/release.yaml`, "Build artifacts" step) publishes only `kuke-linux-amd64` / `kuke-linux-arm64` (+ checksums), and the one-line installer (`docs/site/install.sh`) downloads only `kuke` and hardlinks `kukeond`. A fresh installer user who runs `kuke build` hits `errdefs.ErrKukebuildNotFound` whose remediation hint — "install it (e.g. `make kukebuild` then put it on PATH)" — assumes a source checkout the user does not have.

## Proposal

Ship `kukebuild` through the same release + installer pipeline as `kuke`:

- Build and publish `kukebuild-linux-amd64` / `kukebuild-linux-arm64` (+ `.sha256`) as release assets in `.github/workflows/release.yaml`, alongside the existing `kuke` assets.
- `docs/site/install.sh`: download the `kukebuild` asset for the detected arch, verify its checksum, and install it to `${KUKE_INSTALL_PREFIX}` next to `kuke` / `kukeond`.
- `kuke uninstall` removes the installed `kukebuild` binary in its filesystem-cleanup phase.

## Acceptance criteria

- [ ] Release assets include `kukebuild-linux-amd64` and `kukebuild-linux-arm64` with `.sha256` checksums.
- [ ] A fresh `install.sh` run on a host with only containerd installs `kukebuild` on PATH; `kuke build -t test:dev <context>` succeeds with no docker or buildkitd present.
- [ ] `kuke uninstall` removes the `kukebuild` binary it installed.
- [ ] `install.sh --check` (prereq-check-only mode) is unaffected — still touches nothing.

## Notes

- Blocks: #1228 (README restructure leads with the `kuke build` quick start).
- `kukebuild` lives in its own Go module (`cmd/kukebuild/go.mod`, #655); `make kukebuild` is the existing build target the workflow can reuse.
- Positioning context: the README rework (see blocked issue) leads with a `kuke build` → `kuke run` Claude-cell quick start; without this issue, that flow breaks for every installer user.

eminwux:
## Summary

- `kuke build` is kukeon's docker-free image build path — it execs `kukebuild` (BuildKit-as-a-library). That story only worked for source checkouts: the release workflow published only `kuke`/`kukeond`/`kukepause`, and `install.sh` downloaded only `kuke`. A fresh installer user running `kuke build` hit `errdefs.ErrKukebuildNotFound` whose hint assumes a source tree they don't have.
- **release.yaml**: build `kukebuild-linux-{amd64,arm64}` via the existing `make release-build-kukebuild` target, checksum them, and publish all four assets (binaries + `.sha256`) alongside the existing `kuke` assets. Added a `setup-go` step (mirroring `test.yaml`) so `GOTOOLCHAIN` can auto-fetch kukebuild's pinned Go (its own module is `go 1.25.5` vs root `1.24.5`).
- **install.sh** (canonical `scripts/install.sh`, regenerated `docs/site/install.sh` via `make install.sh`): refactored the download+verify into a shared `fetch_verified` helper, then download/verify/install `kukebuild` next to `kuke`/`kukeond`. Both assets are fetched+verified into the tmpdir before anything is placed on the host, preserving the "fetch failure touches nothing" property.

### Non-obvious calls (please push back if you disagree)

- **AC item 3 (`kuke uninstall` removes kukebuild) — codebase conflict, proceeding per [heads-up comment](https://github.com/eminwux/kukeon/issues/1227#issuecomment-4685705373).** `kuke uninstall` **deliberately does not remove any installed binary** — `cmd/kuke/uninstall/uninstall.go:95-96`: *"The /usr/local/bin/kuke binary and the kukeond symlink are NOT removed — uninstalling runtime state is not the same as uninstalling the binary."* Making it delete `kukebuild` while leaving `kuke`/`kukeond` in the same prefix would contradict that documented invariant and be asymmetric. The actual binary-removal mechanism — `make uninstall-dev` (`Makefile:252`) — **already removes `kukebuild`**. So this PR keeps the `kuke uninstall` invariant unchanged and treats item 3 as satisfied by the existing `make uninstall-dev` coverage. If you instead want `kuke uninstall` to start removing binaries, that's a broader change (should cover `kuke`/`kukeond` too) for a separate issue.
- **`KUKE_SKIP_KUKEBUILD` opt-out (new env var).** Releases predating this PR have no `kukebuild-linux-*` asset, so an unconditional fetch would 404 and break installs pinned to older `KUKE_VERSION` — including the `installer.yaml` `install-skip-init` smoke pinned at `v0.4.0`. Matching the existing `KUKE_SKIP_CHECKSUM` escape-hatch convention, kukebuild download/verify/install is the default; `KUKE_SKIP_KUKEBUILD=1` skips it. Set in the `install-skip-init` job (same rationale as the existing `KUKE_SKIP_CHECKSUM=1` there); bump the pinned tag + drop the skip once a release ships kukebuild.

## Test plan

- [x] `make test` (root + kukebuild modules) — passed, exit 0.
- [x] `make release-build-kukebuild` — produces `kukebuild-linux-amd64` + `kukebuild-linux-arm64` locally.
- [x] `bash -n` on `scripts/install.sh` and `docs/site/install.sh` — clean.
- [x] `make install.sh` sync check — `scripts/install.sh` and `docs/site/install.sh` byte-identical (the `installer.yaml` sync job's invariant).
- [x] `install.sh --check` and `--help` unaffected — `--check` still touches nothing (AC item 4).
- [x] Opt-out install e2e against real `v0.4.0` release (mirrors the `install-skip-init` CI job): `KUKE_SKIP_KUKEBUILD=1` → installs `kuke`+`kukeond`, skips `kukebuild` with no 404.
- [x] `release.yaml` / `installer.yaml` YAML validated.
- [ ] Positive kukebuild install e2e (fresh install downloads+installs kukebuild; `kuke build` succeeds) — **not runnable until the first kukebuild-bearing release exists** (chicken-and-egg: the asset is produced by this PR's workflow on the next tag). The kukebuild install path is structurally identical to the verified `kuke` path (same `fetch_verified` helper + `install -m 0755`), and `make release-build-kukebuild` is confirmed to produce the asset locally. Once a release ships kukebuild, bump the `install-skip-init` smoke tag and drop `KUKE_SKIP_KUKEBUILD` to assert it in CI.

Closes #1227

### DIFF
--- a/.github/workflows/installer.yaml
+++ b/.github/workflows/installer.yaml
@@ -101,6 +101,10 @@ jobs:
   # the test doesn't drift with future releases. Checksum verification is
   # skipped because pre-issue-#62 releases shipped without .sha256 assets;
   # once a release lands with .sha256 published, drop the skip.
+  # KUKE_SKIP_KUKEBUILD is set for the same reason: v0.4.0 predates the
+  # kukebuild release asset (issue #1227), so the installer would 404 fetching
+  # it. Once the pinned tag ships kukebuild-linux-*, bump KUKE_VERSION and drop
+  # this skip to exercise the kukebuild install path here.
   install-skip-init:
     name: full install (KUKE_SKIP_INIT=1) against a real release
     runs-on: ubuntu-latest
@@ -108,6 +112,7 @@ jobs:
       KUKE_VERSION: v0.4.0
       KUKE_SKIP_INIT: "1"
       KUKE_SKIP_CHECKSUM: "1"
+      KUKE_SKIP_KUKEBUILD: "1"
     steps:
       - uses: actions/checkout@v4
       - name: Run installer (download + install, skip init)
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -23,11 +23,32 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v4
 
+      # kukebuild lives in its own Go module (cmd/kukebuild/go.mod, issue #655)
+      # pinned to a newer Go than the root module; `make release-build-kukebuild`
+      # cds into that module and lets GOTOOLCHAIN auto-fetch the pinned
+      # toolchain. setup-go with go-version-file: go.mod gives a known base Go
+      # for that auto-fetch — mirrors the test.yaml job's proven pattern.
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: true
+
       - name: Build artifacts
         run: |
           make release-build \
             KUKEON_VERSION=$KUKEON_VERSION \
             KUKEON_IMAGE_REPO=$KUKEON_IMAGE_REPO
+
+      # kukebuild is the docker-free image builder `kuke build` execs (embeds
+      # BuildKit as a library, writes straight into the realm's containerd
+      # namespace). It is built from its own Go module via a separate target so
+      # the installer can ship it on PATH alongside `kuke` — without it a fresh
+      # installer user hits errdefs.ErrKukebuildNotFound on `kuke build`
+      # (issue #1227).
+      - name: Build kukebuild artifacts
+        run: |
+          make release-build-kukebuild
 
       # The one-line installer (https://kukeon.io/install.sh, issue #62)
       # downloads the per-arch `.sha256` companion file and verifies it
@@ -38,7 +59,8 @@ jobs:
         run: |
           for f in kuke-linux-amd64 kuke-linux-arm64 \
                    kukeond-linux-amd64 kukeond-linux-arm64 \
-                   kukepause-linux-amd64 kukepause-linux-arm64; do
+                   kukepause-linux-amd64 kukepause-linux-arm64 \
+                   kukebuild-linux-amd64 kukebuild-linux-arm64; do
             sha256sum "$f" > "$f.sha256"
           done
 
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #1232 Review — `release: ship kukebuild in release artifacts + installer`**

LGTM — release.yaml now builds + checksums + uploads `kukebuild-linux-{amd64,arm64}` (target output names match the `make release-build-kukebuild` recipe in `Makefile:135-147` and the `OS=linux × ARCHS={amd64,arm64}` matrix exactly), and `install.sh` factors the download+verify into a shared `fetch_verified` helper that ships `kukebuild` next to `kuke`. Both binaries are fetched+verified into the tmpdir before any host `install`, so the "fetch failure touches nothing" property holds. The `KUKE_SKIP_KUKEBUILD` opt-out cleanly mirrors the existing `KUKE_SKIP_CHECKSUM` escape-hatch convention, so the asymmetry is extended rather than fragmented. Both `scripts/install.sh` and the regenerated `docs/site/install.sh` carry byte-identical hunks (sync check green). Commits signed, no secret leaks.

The AC-item-3 decision (leave `kuke uninstall` binary-removal invariant untouched, defer to `make uninstall-dev`) is the right call — flipping it would contradict the documented invariant at `cmd/kuke/uninstall/uninstall.go:95` and be asymmetric against `kuke`/`kukeond`.

One item to run pre-merge (non-blocking, documented carve-out): the positive kukebuild install e2e is unrunnable until the first kukebuild-bearing release exists — chicken-and-egg, the asset is produced by this PR's own workflow on the next tag. Once that release ships, remember to bump the `install-skip-init` `KUKE_VERSION` and drop `KUKE_SKIP_KUKEBUILD=1` to assert the path in CI (already noted inline in `installer.yaml`).

Issue #1237, opened 2026-06-12
volumes: reclaimPolicy Retain + selective cascade reclaim (step 3)


eminwux:
## Context

Step 3 of the volumes epic (umbrella #1015). Step 1 (#1018) reclaims a scope's volumes implicitly: cascade purge removes the whole scope metadata dir via `os.RemoveAll(metadataRunPath)` (`internal/controller/runner/purge_stack.go:120`, `purge_space.go:112`, `purge_realm.go:181`). The umbrella's design calls for an optional `reclaimPolicy: Retain` that protects a volume from that cascade. There is no `reclaimPolicy` precedent in tree, and the blunt scope-dir removal is architecturally incompatible with retention — honoring `Retain` means reworking cascade to enumerate-and-selectively-preserve. Genuine design work, split out of #1018 per the re-size discussion there so it doesn't ride along with mechanical CRUD.

## Proposal

Add an optional `reclaimPolicy` field to the Volume schema (omitted ⇒ today's delete-with-scope behavior) and rework owning-scope cascade purge so a `Retain` volume survives its scope's purge. This carries a real design decision: **where does a stack/space/realm-scoped `Retain` volume survive to once its owning scope dir is deleted?** Relocation to a parent scope, a dedicated retained-volumes area, or refusing the purge are all candidates — decide and document in the PR.

## Acceptance criteria

- [ ] `reclaimPolicy` field on the Volume schema (v1beta1 + internal model + conversion); omitted ⇒ delete-with-scope, no behavior change for existing specs.
- [ ] `kuke purge stack/space/realm --cascade` preserves `Retain` volumes while reclaiming everything else in the scope; the survival location is decided and documented in the PR.
- [ ] A retained volume remains discoverable and deletable after its scope's purge (`kuke get volumes` / `kuke delete volume`).
- [ ] Tests: `Retain` survives each of the three cascade paths; default policy still reclaims; non-volume scope contents are still fully reclaimed.

## Dependencies

Depends on: #1018, #1236
Blocks: none

Issue #1236, opened 2026-06-12
volumes: add create/get/delete volume CLI verbs (step 2)


eminwux:
## Context

Step 2 of the volumes epic (umbrella #1015). Step 1 (#1018) ships the standalone `Volume` resource declaratively (`kuke apply` + daemon storage); this step adds the imperative CLI verbs over it, mirroring the existing create/get/delete blueprint trio (~6 files). Split out of #1018 per the re-size discussion there.

## Proposal

Wire `volume` into the `create`/`get`/`delete` parent commands:

- `kuke create volume <name> --realm … [--space …] [--stack …]` — creates the Volume at the named scope (cell scope rejected, as in step 1's validator).
- `kuke get volumes` — `-o wide`/`-o yaml`, scope-filtered; an empty result prints the standard "no resources found", not an error.
- `kuke delete volume <name>` — removes an unmounted volume's dir. (The live-mount refusal gate needs a container actually mounting the volume, so it lands with the mount kind in step 4, #1016.)

## Acceptance criteria

- [ ] `kuke create volume <name> --realm … [--space …] [--stack …]` creates the Volume at the named scope; cell scope rejected.
- [ ] `kuke get volumes` supports `-o wide` and `-o yaml`, scope-filtered; an empty list prints the standard "no resources found", not an error.
- [ ] `kuke delete volume <name>` removes the volume's dir.
- [ ] Verbs wired into the existing `create`/`get`/`delete` parent commands, mirroring the blueprint trio.
- [ ] CLI tests for all three verbs.

## Notes

- The live-mount delete gate is step 4 (#1016) — it needs a container actually mounting the volume.

## Dependencies

Depends on: #1018
Blocks: #1237, #1016

Issue #1016, opened 2026-06-02
volumes: add volumeRef container mount kind (step 4)


eminwux:
## Context

Step 4 of the volumes epic (umbrella #1015). Wires the container side of the standalone `Volume` (step 1, #1018): a `VolumeKind=volume` mount that **references** a named Volume and bind-mounts it read-write, surviving both container recreation and **cell deletion** (the cell references the Volume; it doesn't own it). Reshaped from this issue's prior "cell-scoped managed volume kind" framing — the volume is no longer cell-owned. Depends on steps 1 and 2 (the delete-gate criterion exercises `kuke delete volume`, which lands in step 2, #1236).

## Proposal

Extend `VolumeMount.Kind` with `VolumeKindVolume = "volume"`. `source: <name>` references a Volume in the cell's own scope (resolved by walking realm/space/stack, like secret lookup); a `volumeRef: {name, realm, space, stack}` references one cross-scope (mirroring `ContainerSecretRef`, `pkg/api/model/v1beta1/container.go:269`). `validateVolumes` (`internal/controller/create_container.go:284`) gains a `volume` case (source is a Volume name, not a host path; Target absolute); `buildVolumeMounts` (`internal/ctr/spec.go:661`) resolves the reference to the Volume's on-disk dir and emits `{Type:"bind", rbind, rw}` (honoring `readOnly`). Optionally auto-create the Volume at the cell's stack scope when the referenced name doesn't exist (Docker's `-v name:/path`); decide in the PR.

## Acceptance criteria

- [ ] `VolumeKindVolume = "volume"` in v1beta1 + internal model; empty Kind still defaults to `bind` (no behavior change for existing specs).
- [ ] `volume` case in `validateVolumes`: `source` is a Volume name (or a `volumeRef` block) — not an absolute host path; `Target` absolute; the referenced Volume must resolve (or be auto-created if that option is taken).
- [ ] `volumeRef` shape (name + realm/space/stack) plumbed through v1beta1 ↔ apischeme ↔ validator ↔ clone (the surface #347 covered for tmpfs).
- [ ] `buildVolumeMounts` emits `{Destination: Target, Source: <resolved Volume dir>, Type:"bind", Options:["rbind","rw"]}` (`readOnly` → `ro`).
- [ ] **Survives cell delete:** a cell mounting Volume `V` is deleted → `V` and its contents persist; a new cell referencing `V` sees the prior contents (end-to-end test).
[...]

PR #1238, opened 2026-06-12
feat: add standalone Volume resource (step 1)

### LINKED ISSUE #1018: volumes: add standalone Volume resource (step 1)
## Context

Foundation of the volumes epic (umbrella #1015), reshaped to the Docker model: a **standalone, daemon-managed `Volume` resource** with its own lifecycle, decoupled from any cell. Mirrors how a daemon-stored kind lands schema+storage before its runtime consumer (cf. CellConfig #624 → #625); the container-side mount kind that references a Volume is step 4 (#1016).

Narrowed per the re-size discussion below (the closest template, the Blueprint subsystem, is ~20 files / ~2,700 LOC on its own): this step ships the declarative resource — schema, daemon storage, `kuke apply` — with implicit owning-scope cascade reclaim. The imperative CLI verbs are step 2 (#1236); `reclaimPolicy: Retain` + selective cascade is step 3 (#1237).

## Proposal

Add `kind: Volume` (`apiVersion: v1beta1`): schema, daemon storage, and `kuke apply` support. Scopable at realm/space/stack — **never cell** (a cell-nested volume would be reclaimed with the cell; this resource must outlive cells). Stored at `<run-path>/data/<scope>/volumes/<name>` via new `fs.VolumePath`/`fs.VolumeScopeDir` (mirroring `fs.SecretPath`/`fs.SecretScopeDir` in `internal/util/fs/metadata.go`, but scope-capped at stack like Blueprint/Config). The directory is created with ownership/mode the eventual container user can write. Reclaimed by owning-scope (stack/space/realm) cascade purge — implicit today, since the three purge paths `os.RemoveAll` the scope metadata dir (`internal/controller/runner/purge_stack.go:120`, `purge_space.go:112`, `purge_realm.go:181`) — **not** by cell deletion.

## Acceptance criteria

- [ ] `kind: Volume` schema in `pkg/api/model/v1beta1` + internal model + apischeme conversion; `apiVersion: v1beta1`; scope validated realm/space/stack with **cell rejected** (mirror `validateBlueprintScope`, `internal/apply/parser/parser.go:534`).
- [ ] `KindVolume` parser registration (`internal/apply/parser/parser.go`); `kuke apply -f` creates/updates a Volume via `ReconcileVolume`.
- [ ] Daemon storage CRUD in runner + controller (the Blueprint shape: `internal/controller/runner/blueprint.go`, `internal/controller/blueprint.go`).
- [ ] `fs.VolumePath`/`fs.VolumeScopeDir` resolve `<scope>/volumes/<name>` (+ `KukeonVolumesSubdir` const); the dir is created idempotently with ownership/mode a container user can write — document the chosen owner/mode per the pr-reviewer sibling-precedent rule (e.g. the attachable socket's `root:kukeon` model).
- [ ] Scope purge (`purge stack/space/realm --cascade`) reclaims the scope's volumes — implicit via the existing scope-dir `RemoveAll`, covered by a test rather than new mechanism.
- [ ] Tests: scope validation (cell rejected), storage CRUD round-trip, `apply` idempotency, implicit scope-cascade reclaim.

## Notes

- Imperative `kuke create/get/delete volume` verbs are step 2 (#1236). The live-mount **delete gate** and the **survives-cell-delete** guarantee both require a container actually mounting the volume, so they are exercised in step 4 (#1016) where the mount kind exists.
- `reclaimPolicy: Retain` is step 3 (#1237) — it requires reworking cascade purge from the blunt scope-dir `RemoveAll` to enumerate-and-selectively-preserve, a design decision that should not ride along with mechanical CRUD.
- No auto-create here — declarative `apply` only; auto-create on first reference is an option evaluated in step 4 (#1016).

## Dependencies

Depends on: none
Blocks: #1236, #1237, #1016, #1017

eminwux:
## Summary

Implements **step 1 of the volumes epic** (#1015, re-sized from #1018): a standalone, daemon-managed `kind: Volume` resource — schema, daemon storage, and `kuke apply` support — decoupled from any cell. Mirrors the `CellBlueprint` subsystem shape for parser/scope/storage and the `Secret` precedent for path-segment safety. The imperative CLI verbs are step 2 (#1236); `reclaimPolicy: Retain` + selective cascade is step 3 (#1237); the container-side mount kind is step 4 (#1016).

Why the directory-backed model: a Volume's *resource* is the on-host directory a container mounts and writes into (`<scope>/volumes/<name>`), not a serialized document. So `WriteVolume` provisions a directory (idempotent `MkdirAll`) rather than writing a file, and `GetVolume`/`ListVolumes` are metadata-only (scope + name derived from the path), the same shape as `GetSecret`. The internal carrier therefore has no `Document` field.

## Acceptance criteria

- [x] `kind: Volume` schema in `pkg/api/model/v1beta1` (`volume.go`) + internal model (`internal/modelhub/volume.go`) + apischeme conversion (`internal/apischeme/volume.go`); `apiVersion: v1beta1`; scope validated realm/space/stack with **cell rejected** (mirrors `validateBlueprintScope` — `VolumeMetadata` structurally has no `Cell` field, so a `cell:` coordinate is dropped at unmarshal, the same way `CellBlueprintMetadata` rejects cell scope).
- [x] `KindVolume` parser registration (`internal/apply/parser/parser.go`: `Document.VolumeDoc` field, `ParseDocument` case, `ValidateDocument` kind list + `validateVolume`/`validateVolumeScope`); `kuke apply -f` creates/updates a Volume via `ReconcileVolume` (`internal/controller/apply/reconcile.go`, wired in `internal/controller/apply.go`).
- [x] Daemon storage CRUD in runner (`internal/controller/runner/volume.go`: `WriteVolume`/`GetVolume`/`ListVolumes`/`DeleteVolume` + `Runner` interface) + controller (`internal/controller/volume.go`: `GetVolume`/`ListVolumes`/`DeleteVolume` + lookup/filter guards), mirroring the Blueprint shape.
- [x] `fs.VolumePath`/`fs.VolumeScopeDir` (+ `fs.VolumesDir`, `KukeonVolumesSubdir` const) resolve `<scope>/volumes/<name>`; the dir is created idempotently. **Owner/mode choice (documented per the sibling-precedent rule):** the per-scope `volumes/` container dir is `0o755` root-owned (world-traversable, like `blueprints/`); each **volume dir** is `02770 root:kukeon` (setgid, group-writable, no world access) when the kukeon GID is configured, `0o770 root:root` otherwise — directly mirroring `attachableTTYDirInitialPerms`/`attachableTTYDirRootMode`. The setgid bit makes container-created files inherit the kukeon group. The mounting container's own uid is wired in at mount time in **step 4 (#1016)**, the same two-phase shape as `attachablePostCreateChown` — so step 1 satisfies "container-writable" for root/kukeon-group writers and defers the arbitrary-uid finalization to the mount step.
- [x] Scope purge (`purge stack/space/realm --cascade`) reclaims the scope's volumes — implicit via the existing scope-dir `os.RemoveAll`, covered by `TestVolume_ReclaimedByScopeCascade` (asserts containment under the scope dir, then drives the exact `RemoveAll` the three purge paths use) rather than any new mechanism.
- [x] Tests: scope validation incl. cell-rejected + unsafe-name (`internal/apply/parser/volume_test.go`), storage CRUD round-trip + mode/ownership (`internal/controller/runner/volume_test.go`), `apply` idempotency create→updated + missing-scope reject (`internal/controller/apply/reconcile_volume_test.go`), controller CRUD guards + end-to-end `ApplyDocuments` dispatch (`internal/controller/volume_test.go`), and implicit scope-cascade reclaim.

## Non-obvious calls (reviewer please confirm)

- **No `apply --prune` / team-label participation for Volume.** Unlike Blueprint/Config, the `KindVolume` apply dispatch deliberately does **not** stamp `kukeon.io/team` or feed the prune accumulator: a volume is decoupled from any cell and outlives the apply that created it, so `apply --prune` removing one would wipe persistent data. Reclaim is owning-scope cascade purge only. (Documented inline at the dispatch site.)
- **Path-segment safety beyond a literal "mirror validateBlueprintScope".** `validateVolumeScope` adds `validateVolumeSegment` (rejects `/`, `..`, NUL in name/coords) because a Volume provisions a real directory — a path-traversal name is a genuine escape risk. This mirrors the `Secret` coordinate guard (#673) rather than Blueprint, which lacks it. New sentinel `ErrVolumeCoordUnsafe`.
- **`VolumeSpec` is an empty struct.** No spec fields in step 1; it exists so step 3's `reclaimPolicy` is a field append, not a schema reshape.
- **Declarative `kuke delete -f` of a Volume** is unsupported (hits the existing `ErrUnknownKind` default in `delete_documents.go`) — consistent with Secret/Blueprint/Config siblings, which the delete-documents switch also doesn't handle. Volume removal is cascade-reclaim (this PR) or the step-2 imperative verb.

## Test plan

- [x] `make test-root` — full main-module suite, all green
- [x] `make test-kukebuild` — green (untouched by this change)
- [x] `GOWORK=off go vet ./... (minus e2e)` — clean
- [x] `gofmt -l` on all changed files — clean
- [ ] `make dev-init` end-to-end smoke — not run: this change adds a new resource kind to the apply/storage path and does not touch the build path, the daemon bootstrap, or anything under `/opt/kukeon` that the daemon-parity tail reads; the new code paths are covered by the unit + dispatch tests above. The reviewer may run it on a clean host if a live `kuke apply -f` of a Volume manifest is desired.

Closes #1018

### DIFF
--- a/internal/apischeme/volume.go
+++ b/internal/apischeme/volume.go
@@ -1,0 +1,84 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package apischeme
+
+import (
+	"fmt"
+
+	intmodel "github.com/eminwux/kukeon/internal/modelhub"
+	ext "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
+)
+
+// ConvertVolumeDocToInternal converts an external VolumeDoc to the internal hub
+// carrier (issue #1018). Unlike a CellBlueprint there is no body to serialize —
+// a Volume's spec is empty and the resource is the on-host directory — so this
+// is a flat copy of the scope coordinates onto the metadata.
+func ConvertVolumeDocToInternal(in ext.VolumeDoc) (intmodel.Volume, error) {
+	switch in.APIVersion {
+	case VersionV1Beta1, "": // default/empty treated as v1beta1
+		return intmodel.Volume{
+			Metadata: intmodel.VolumeMetadata{
+				Name:  in.Metadata.Name,
+				Realm: in.Metadata.Realm,
+				Space: in.Metadata.Space,
+				Stack: in.Metadata.Stack,
+			},
+		}, nil
+	default:
+		return intmodel.Volume{}, fmt.Errorf("unsupported apiVersion for Volume: %s", in.APIVersion)
+	}
+}
+
+// NormalizeVolume takes an external VolumeDoc request and returns an internal
+// carrier and the chosen apiVersion.
+func NormalizeVolume(req ext.VolumeDoc) (intmodel.Volume, ext.Version, error) {
+	version := DefaultVersion(req.APIVersion)
+	internal, err := ConvertVolumeDocToInternal(req)
+	if err != nil {
+		return intmodel.Volume{}, "", err
+	}
+	return internal, version, nil
+}
+
+// ConvertVolumeToExternal builds a metadata-only external VolumeDoc from the
+// internal carrier (issue #1018). A Volume has no body, so the external doc is
+// its canonical apiVersion/kind plus the scope coordinates and name.
+func ConvertVolumeToExternal(in intmodel.Volume) ext.VolumeDoc {
+	return ext.VolumeDoc{
+		APIVersion: ext.APIVersionV1Beta1,
+		Kind:       ext.KindVolume,
+		Metadata: ext.VolumeMetadata{
+			Name:  in.Metadata.Name,
+			Realm: in.Metadata.Realm,
+			Space: in.Metadata.Space,
+			Stack: in.Metadata.Stack,
+		},
+	}
+}
+
+// ConvertVolumeListToExternal maps a slice of internal Volumes to metadata-only
+// external VolumeDocs (issue #1018).
+func ConvertVolumeListToExternal(in []intmodel.Volume) []ext.VolumeDoc {
+	if in == nil {
+		return nil
+	}
+	out := make([]ext.VolumeDoc, len(in))
+	for i := range in {
+		out[i] = ConvertVolumeToExternal(in[i])
+	}
+	return out
+}
--- a/internal/apply/parser/parser.go
+++ b/internal/apply/parser/parser.go
@@ -44,6 +44,7 @@ type Document struct {
 	SecretDoc        *v1beta1.SecretDoc
 	CellBlueprintDoc *v1beta1.CellBlueprintDoc
 	CellConfigDoc    *v1beta1.CellConfigDoc
+	VolumeDoc        *v1beta1.VolumeDoc
 }
 
 // ValidationError represents a validation error for a specific document.
@@ -188,6 +189,14 @@ func ParseDocument(index int, raw []byte) (*Document, error) {
 		doc.CellConfigDoc = &configDoc
 		doc.APIVersion = configDoc.APIVersion
 
+	case v1beta1.KindVolume:
+		var volumeDoc v1beta1.VolumeDoc
+		if unmarshalErr := yaml.Unmarshal(raw, &volumeDoc); unmarshalErr != nil {
+			return nil, fmt.Errorf("document %d: failed to parse Volume: %w", index, unmarshalErr)
+		}
+		doc.VolumeDoc = &volumeDoc
+		doc.APIVersion = volumeDoc.APIVersion
+
 	default:
 		// kind: CellProfile was removed in #626. Give the operator an
 		// explicit migration pointer instead of the generic unknown-kind
@@ -228,7 +237,7 @@ func ValidateDocument(doc *Document) *ValidationError {
 	switch doc.Kind {
 	case v1beta1.KindRealm, v1beta1.KindSpace, v1beta1.KindStack, v1beta1.KindCell,
 		v1beta1.KindContainer, v1beta1.KindSecret, v1beta1.KindCellBlueprint,
-		v1beta1.KindCellConfig:
+		v1beta1.KindCellConfig, v1beta1.KindVolume:
 		// Valid kind
 	default:
 		return &ValidationError{
@@ -474,6 +483,11 @@ func ValidateDocument(doc *Document) *ValidationError {
 		if validationErr := validateConfig(doc); validationErr != nil {
 			return validationErr
 		}
+
+	case v1beta1.KindVolume:
+		if validationErr := validateVolume(doc); validationErr != nil {
+			return validationErr
+		}
 	}
 
 	return nil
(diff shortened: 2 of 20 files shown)


eminwux:
**PR #1238 Review — `feat: add standalone Volume resource (step 1)`**

Clean, well-documented mirror of the Blueprint/Secret subsystems — schema, apischeme, parser path-segment guard (exact parity with `validateSecretSegment`), reconcile scope-check, and storage runner all hang together, and the no-prune / no-team-stamp decoupling is correctly reasoned. One blocking item: the new resource subdir is missing from the shared reserved-subdir set.

Blocking:

- **Stated-invariant violation** `internal/controller/runner/scope_children.go:41-45`: the PR adds `volumes/` as a per-scope resource subdir but doesn't add `consts.KukeonVolumesSubdir` to `reservedScopeSubdirs` — whose own doc-comment calls it "the single source of truth ... a future reserved subdir is added in exactly one place" and ties the omission to the #734 regression class. Consequence: every `childScopeNames` consumer (`ListSecrets`/`ListBlueprints`/`ListConfigs` in secret.go/blueprint.go/config.go, plus `ListVolumes` itself) now treats a scope's `volumes/` dir as a phantom child space/stack and recurses into the container-writable volume directories. List output isn't corrupted today (phantom paths dead-end on a missing nested resource subdir), but it's exactly the invariant the comment forbids breaking, and the walk reads the daemon into container-controlled trees. Fix: add `consts.KukeonVolumesSubdir: {}` to the map + a regression test (e.g. a volume in a realm, assert no phantom traversal).

Non-blocking (no action needed):

- Controller `GetVolume`/`ListVolumes`/`DeleteVolume` + their runner counterparts are test-only in step 1 — only `WriteVolume` is wired (via `ReconcileVolume`/`apply`). Deliberate per the AC and step-2 (#1236), and unit-tested; noting for visibility.
- The setgid volume-dir path (gid>0 — the normal deployment) is covered only by the pure-function `volumeDirInitialPerms` test; the `WriteVolume` round-trip asserts the gid==0 fallback (`0o770`). Behavior is correct (Linux preserves a directory's setgid bit across a root `chown`), so non-blocking — just no on-disk assertion on the production-normal path.

LGTM once the `reservedScopeSubdirs` entry lands — the rest are non-blocking.

eminwux:
Addressed in 4e5b8c1 (rebased onto current main):

- **Blocking — `reservedScopeSubdirs` omission (`scope_children.go`):** added `consts.KukeonVolumesSubdir: {}` to the shared set so every `childScopeNames` consumer (`ListSecrets`/`ListBlueprints`/`ListConfigs`/`ListVolumes`) skips a scope's `volumes/` dir instead of recursing into it as a phantom child space/stack. This is the exact `#734` invariant the doc-comment forbids breaking. Also refreshed the doc-comment's walker list to include `volume.go` (now the fourth consumer) and noted `volumes/` (#1018) in the reserved-subdir enumeration.
- **Regression test:** `TestVolume_NotMistakenForChildScope` in `volume_test.go` — seeds a realm-scoped blueprint + a realm-scoped volume, then asserts `ListBlueprints(all)` returns only the blueprint (the `volumes/` subdir must not surface as a phantom space). Mirrors the existing `TestList{Blueprints,Configs,Secrets}_IgnoresReservedSubdirs` shape.

Both non-blocking notes acknowledged, no action taken per your "no action needed":
- `GetVolume`/`ListVolumes`/`DeleteVolume` test-only in step 1 — deliberate, step-2 (#1236) wires the verbs.
- setgid production path covered by the `volumeDirInitialPerms` pure-function test; correct since Linux preserves the setgid bit across a root chown.

Test plan: `make test-root` green (full main-module suite, `GOWORK=off` to dodge the local `go.work`/containerd-cgroups toolchain skew), `go vet` clean, `gofmt -l` clean on both changed files.

eminwux:
**PR #1238 Review — `feat: add standalone Volume resource (step 1)`**

Clean, well-tested foundation PR. Faithfully mirrors the Blueprint subsystem shape (parser/scope/storage) and the Secret coordinate-guard precedent; the mode/ownership model tracks the attachable-TTY two-phase pattern correctly. Both commits GPG-verified, CI fully green, no sensitive-data leaks. Correctness holds across the apply → reconcile → `WriteVolume` path, the `reservedScopeSubdirs` walker fix, and the doc-sort ordering.

A couple of non-blocking items (both forward-looking into step 2):

- **Dead code (non-blocking)** `internal/apischeme/volume.go:75`: `ConvertVolumeListToExternal` has zero references — no caller, no test. It's intentional forward-declaration (its sibling `ConvertSecretListToExternal` is wired at `internal/client/local/local.go:500`), so step 2 (#1236)'s `kuke get volume` will consume it. Fine to land, but it ships untested unlike its siblings.
- **Path-safety parity (non-blocking)** `internal/controller/volume.go:106`: `validateVolumeLookup` (used by `GetVolume`/`DeleteVolume`) omits the `validateVolumeSegment` check the parser enforces. Harmless here — those controller methods are test-only and unreachable from production in step 1 — but step 2's imperative verbs call `DeleteVolume`/`GetVolume` directly, so segment safety needs a home at the controller layer (the daemon-side guard) before then, or a `../`-name delete could `filepath.Join`-escape.

Deferred verification for the user: a live `kuke apply -f <volume>.yaml` against a running daemon would confirm the end-to-end path the unit tests don't reach (`make dev-init` is host-gated and the parity tail doesn't exercise volume paths — correctly documented in the test plan).

LGTM to proceed — both items above are non-blocking and step-2-scoped.

eminwux:
**PR #1238 Review (second reviewer) — `feat: add standalone Volume resource (step 1)`**

Independent second pass — confirming the prior LGTM. Reviewed the full merge-base diff (20 files), cross-checked every new symbol for callers, and verified the sibling precedent.

- **Correctness**: `WriteVolume` create-detection (stat-before-`MkdirAll`), idempotent mode/owner re-assert, non-dir-squatter refusal, and the stat-gated `RemoveAll` in `DeleteVolume` are all sound. `volumes/` is added to `reservedScopeSubdirs` with a regression test (the #734 trap). No leaks, no dead code (every new symbol — sentinels, `fs` helpers, controller/runner methods — has a non-test caller), 2 commits both signature-verified.
- **No semantic merge interaction**: main's commits since base touch `reconcile.go`/`runner.go`/`consts.go` only in cell-state regions, distinct from this PR's additive functions; `mergeStateStatus` is CLEAN.
- **Path-segment safety**: `validateVolumeLookup` (controller get/delete) omits the `/`/`..`/NUL guard, deferring it to the parser's `validateVolumeScope` — this matches the Secret two-stage model exactly, and the apply path is fully guarded.

Non-blocking note: when the step-2 imperative verbs (#1236) wire `kuke delete volume` to `controller.DeleteVolume` → `RemoveAll`, the CLI-supplied name reaches `fs.VolumePath` without the segment guard (same latent shape Secret already carries). Worth a segment check at the verb boundary in #1236 — not this PR's regression.

LGTM — stays `ready-to-merge`.

PR #1242, opened 2026-06-12
fix: install kukepause in install.sh so kuke init can bootstrap

### LINKED ISSUE #1241: installer: kuke init fails because install.sh never fetches kukepause
## Repro

Fresh host, official one-line installer at the current release (v0.6.1):

```bash
curl -fsSL https://kukeon.io/install.sh | bash
```

The installer downloads + verifies `kuke` and `kukebuild`, installs them to `/usr/local/bin`, then runs `kuke init`.

## Observed

`kuke init` aborts at the kukepause staging step:

```
==> Installing to /usr/local/bin
✓ installed kuke + kukeond + kukebuild at /usr/local/bin
==> Initializing the runtime
Error: stage kukepause: kukepause not found on $PATH or alongside kuke; build it with `make kukepause` and install it (`make install-dev`)
```

No fresh install can complete `kuke init` — the host is left with `kuke`/`kukebuild` on PATH but a non-functional runtime. The error message is also dev-oriented (`make kukepause` / `make install-dev`), which is misleading for an installer user who has no source checkout.

## Root cause

`scripts/install.sh` `do_install()` (and its deployed mirror `docs/site/install.sh`) was extended in #1227 / commit `da51a78` to fetch+verify+install **kuke** and **kukebuild** via the new `fetch_verified` helper, but **kukepause was never added**. The release workflow already publishes the artifact — `.github/workflows/release.yaml` builds and uploads `kukepause-linux-{amd64,arm64}` plus `.sha256` checksums alongside the other two binaries — so the asset exists at the release; the installer simply never fetches it.

`kuke init` resolves kukepause via `resolveKukepauseSource()` (`cmd/kuke/init/init.go:421-441`), which only checks `$PATH` and the directory alongside the running `kuke`. Since the installer places neither, init fails. kukepause is a hard prerequisite: it is the static PID-1 binary staged to `/opt/kukeon/bin/kukepause` and bind-mounted into every cell's root (pause) container.

## Suggested fix

In `scripts/install.sh` `do_install()` (mirror the change into `docs/site/install.sh`), fetch+verify+install `kukepause-linux-${arch}` exactly like the kukebuild block — but **unconditionally**, with no `KUKE_SKIP_*` gate, because unlike `kuke build` it is mandatory for `kuke init`:

```bash
fetch_verified "kukepause-linux-${arch}" "$kukepause_path"
chmod +x "$kukepause_path"
...
$SUDO install -m 0755 "$kukepause_path" "${KUKE_INSTALL_PREFIX}/kukepause"
```

Update the "installed kuke + kukeond + kukebuild" success line to include kukepause. Consider whether old-tag pins predating the kukepause release asset (#931) need the same graceful-degrade treatment kukebuild got — though since kukepause is mandatory, a missing asset on an ancient pin should fail clearly rather than skip.

## Manual workaround

```bash
ver=v0.6.1; arch=amd64
curl -fsSL -o /tmp/kukepause "https://github.com/eminwux/kukeon/releases/download/${ver}/kukepause-linux-${arch}"
sudo install -m 0755 /tmp/kukepause /usr/local/bin/kukepause
sudo kuke init
```

eminwux:
## Summary

- `scripts/install.sh` (and its published mirror `docs/site/install.sh`) downloaded and installed `kuke` and `kukebuild` but never fetched `kukepause`, so `curl -fsSL https://kukeon.io/install.sh | bash` left a host where `kuke init` aborts at `stage kukepause: kukepause not found on $PATH or alongside kuke`. The release workflow already publishes the `kukepause-linux-{amd64,arm64}` asset + `.sha256`; the installer simply never fetched it. Reported and root-caused in #1241.
- Adds an **unconditional** `fetch_verified "kukepause-linux-${arch}"` + `install -m 0755 … /kukepause` step, mirroring the kukebuild handling. `kuke init` resolves kukepause via `$PATH`/alongside-`kuke` (`cmd/kuke/init/init.go:421-441`) and stages it as every cell's static PID 1, so the binary must land in the install prefix.

## Design notes

- **No skip flag for kukepause (gate-vs-flip).** Unlike `kukebuild` (gated by `KUKE_SKIP_KUKEBUILD` so old pins predating its asset don't 404), kukepause is **mandatory** — `kuke init` cannot bootstrap without it. So a release pin too old to ship `kukepause-linux-*` should fail loudly at fetch rather than silently install a runtime that can't init. Documented inline in the script comment.
- **CI tag bump (required, not scope creep).** The `install-skip-init` job in `.github/workflows/installer.yaml` pinned `KUKE_VERSION: v0.4.0` with `KUKE_SKIP_KUKEBUILD=1`. Because the kukepause fetch is unconditional, v0.4.0 (which predates the kukepause asset) would now 404 and fail the job. The job's own comment anticipated this ("bump KUKE_VERSION … to exercise the kukebuild install path"). Bumped to `v0.6.1` — which ships kuke + kukebuild + kukepause + all `.sha256` sums — and dropped the now-stale `KUKE_SKIP_CHECKSUM` / `KUKE_SKIP_KUKEBUILD` skips so the job exercises the full multi-binary download + checksum-verify path, plus added `test -x` assertions for the kukebuild and kukepause artifacts.

## Test plan

- [x] `bash -n scripts/install.sh` and `bash -n docs/site/install.sh` — syntax clean
- [x] `make install.sh` — docs mirror re-renders with no drift (`diff -q` of the two files is identical, the `installer.yaml` `sync` job's invariant)
- [x] `git grep` — no remaining references to the old `installed kuke + kukeond + kukebuild` success string
- [ ] `.github/workflows/installer.yaml` `install-skip-init` job against the real v0.6.1 release — exercised by CI on this PR (downloads real release assets; not run locally as it requires fetching GitHub release binaries)
- [ ] `make test` — not run: this PR touches only shell scripts and a CI workflow, no Go code; the Go test target covers none of the changed files

Closes #1241

### DIFF
--- a/.github/workflows/installer.yaml
+++ b/.github/workflows/installer.yaml
@@ -95,24 +95,21 @@ jobs:
       - name: Run installer --check
         run: bash scripts/install.sh --check
 
-  # End-to-end smoke: download a real release binary, install it to
+  # End-to-end smoke: download real release binaries, install them to
   # /usr/local/bin, and skip the kuke-init step (which would need root,
   # cgroup writes, and a healthy containerd). Pinned to a known-good tag so
-  # the test doesn't drift with future releases. Checksum verification is
-  # skipped because pre-issue-#62 releases shipped without .sha256 assets;
-  # once a release lands with .sha256 published, drop the skip.
-  # KUKE_SKIP_KUKEBUILD is set for the same reason: v0.4.0 predates the
-  # kukebuild release asset (issue #1227), so the installer would 404 fetching
-  # it. Once the pinned tag ships kukebuild-linux-*, bump KUKE_VERSION and drop
-  # this skip to exercise the kukebuild install path here.
+  # the test doesn't drift with future releases. v0.6.1 ships kuke, kukebuild,
+  # and kukepause assets plus their .sha256 sums, so this job exercises the
+  # full multi-binary download + checksum-verify + install path (no
+  # KUKE_SKIP_CHECKSUM, no KUKE_SKIP_KUKEBUILD). kukepause is fetched
+  # unconditionally because `kuke init` cannot bootstrap without it (issue
+  # #1241).
   install-skip-init:
     name: full install (KUKE_SKIP_INIT=1) against a real release
     runs-on: ubuntu-latest
     env:
-      KUKE_VERSION: v0.4.0
+      KUKE_VERSION: v0.6.1
       KUKE_SKIP_INIT: "1"
-      KUKE_SKIP_CHECKSUM: "1"
-      KUKE_SKIP_KUKEBUILD: "1"
     steps:
       - uses: actions/checkout@v4
       - name: Run installer (download + install, skip init)
@@ -121,4 +118,6 @@ jobs:
         run: |
           test -x /usr/local/bin/kuke
           test -x /usr/local/bin/kukeond
+          test -x /usr/local/bin/kukebuild
+          test -x /usr/local/bin/kukepause
           /usr/local/bin/kuke version
--- a/docs/site/install.sh
+++ b/docs/site/install.sh
@@ -55,7 +55,7 @@ usage() {
     cat <<EOF
 Usage: install.sh [--check] [--help]
 
-Installs kukeon (kuke + kukeond + kukebuild) on a Linux host.
+Installs kukeon (kuke + kukeond + kukebuild + kukepause) on a Linux host.
 
   --check    Run prerequisite checks only; do not touch the system.
   --help     Show this help.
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #1242 Review — `fix: install kukepause in install.sh so kuke init can bootstrap`**

LGTM — adds the missing unconditional `kukepause` fetch + `install -m 0755` to both installer mirrors, symmetric with the existing kukebuild handling and using the same plumbing; kukepause-into-tmpdir-before-system-placement preserves the host-untouched-on-failure invariant, and the v0.6.1 CI bump (dropping the now-stale checksum/kukebuild skips) exercises the full multi-binary download+verify path green on this PR.

PR #1253, opened 2026-06-12
fix: reconcile re-reads cell from disk after lock to avoid resurrecting deleted cell

### LINKED ISSUE #1251: daemon: reconcile loop resurrects just-deleted cell from stale snapshot
## Context

`kuke delete cell <name>` reports success and tears down the cell's containers, cgroup, and `metadata.json`, but the cell **reappears** moments later in `Stopped` state. The just-deleted cell is silently resurrected by the background reconcile loop. Two operator-visible symptoms in one session:

- `kuke delete cell github-runner` prints `Deleted cell "github-runner" from stack "default"`, yet `kuke get ce` still lists `github-runner` (`Stopped`, age unchanged).
- `kuke delete secret github-runner-token` keeps failing with `secret is referenced by 1 cell(s): default/default/default/github-runner` even after the "successful" cell delete — the resurrected `metadata.json` still carries the `secretRef`, so the secret can never be deleted.

Daemon log of the resurrection (timestamps from the report):

```
12:56:13.653 deleted metadata file .../github-runner/metadata.json   <- DeleteCell completes
12:56:13.655 stack cgroup does not exist, attempting to create
12:56:13.658 metadata file written .../github-runner/metadata.json   <- RECREATED
12:56:13.660 reconcile: healed missing cell cgroup cell=github-runner
12:56:13.667 reconcile ok cells_scanned=2 cells_updated=1 cells_deleted=0
```

## Root cause

A TOCTOU between `DeleteCell` and the reconcile loop's **stale cell snapshot**, not closed by the per-cell lock.

1. `ReconcileCells` snapshots the cell list once per tick (`internal/controller/reconcile.go:79` `ListCells`), then iterates calling `ReconcileCell(cell)` per cell (`reconcile.go:88`) with the **in-memory `cell` value captured at list time** — including `Status.ReadyObserved=true`.
2. `kuke delete cell` acquires the per-cell lock (`internal/controller/runner/delete_cell.go:34`) and removes containers, cgroup, and `metadata.json`. The reconcile tick's `ReconcileCell` for that same cell **blocks on the per-cell lock** for the duration (the ~10s stop-timeout in the log explains the 12:56:03 tick logging its heal at 12:56:13).
3. When the delete releases the lock, `ReconcileCell` proceeds on its **pre-delete snapshot** without re-reading disk. The post-reboot cgroup-heal branch (`internal/controller/runner/refresh.go:706`) fires because `!cgroupExists && originalStatus.ReadyObserved` is true in the stale copy.
4. `ensureCellCgroup` recreates the cgroup **and** calls `UpdateCellMetadata` (`internal/controller/runner/provision.go:1314`), rewriting the deleted `metadata.json` ("metadata file written" in the log). The cell is back.
5. The secret-delete guard re-reads persisted cell specs (`internal/controller/secret.go:158` `secretReferences` → `ListCells`); the resurrected `metadata.json`'s `secretRef` blocks `kuke delete secret` indefinitely.

The per-cell lock (issue #714) correctly serializes the *mutations*, but it does not add a post-lock existence recheck — `ReconcileCell` never re-reads the cell from disk after acquiring the lock, so a cell deleted while the tick was blocked is treated as a live cell with a missing cgroup and "healed" back into existence. The `#855` heal path was written for post-reboot recovery (cgroup tmpfs wiped, metadata survives); it does not distinguish that case from "metadata was just deleted."

## Proposal

In `ReconcileCell` (`internal/controller/runner/refresh.go`), after acquiring the per-cell lock and before the cgroup-heal / derivation logic, re-read the cell from disk (`GetCell`) and short-circuit when it returns `ErrCellNotFound` — treat the cell as deleted (no heal, no metadata rewrite, no cgroup recreate). This closes the window the `#714` lock left open: the lock makes the recheck reliable (the delete has fully completed before the reconcile observes absence). Confirm the post-reboot heal (`#855`) still works, since on a genuine reboot the `metadata.json` survives and `GetCell` succeeds.

## Acceptance criteria

- [ ] `kuke delete cell <name>` followed by `kuke get ce` no longer shows the cell — it stays deleted across at least two reconcile ticks.
- [ ] After deleting a cell that referenced a secret, `kuke delete secret <name>` succeeds (the resurrected `secretRef` no longer blocks it).
- [ ] `ReconcileCell` re-reads the cell from disk after acquiring the per-cell lock and returns a no-op outcome (no heal, no `metadata.json` rewrite, no cgroup recreate) when the cell's metadata is gone.
- [ ] The `#855` post-reboot cgroup-heal path still heals a Ready cell whose cgroup is absent **but whose `metadata.json` survives** — regression test covering reboot-survives-metadata vs. deleted-metadata.
- [ ] Unit/regression test reproducing the delete-while-reconcile-blocked-on-lock race (delete completes, then a reconcile tick holding a pre-delete snapshot must not resurrect the cell).

## Notes

- Emitting code: `internal/controller/reconcile.go:79,88`; `internal/controller/runner/refresh.go:651,706`; `internal/controller/runner/provision.go:1314`; `internal/controller/runner/delete_cell.go:34`; secret guard `internal/controller/secret.go:158`.
- Related: the per-cell lock was introduced by issue #714 (serialize gRPC handler vs. reconcile tick); the heal branch by issue #855 (post-reboot cgroup recovery). This bug is the interaction of the two — the lock serializes but does not add the post-lock existence recheck the heal path needs.
- The `metadata file does not exist .../secrets/metadata.json` ERROR line in the log is a separate, pre-existing noise line on every reconcile tick (the realm has no secrets metadata index file) and is unrelated to this bug — worth a follow-up to silence, but not in scope here.

eminwux:
## Summary

- Closes a TOCTOU between `DeleteCell` and the reconcile loop's **stale per-tick cell snapshot** (the interaction of the `#714` per-cell lock and the `#855` post-reboot cgroup-heal branch). `ReconcileCells` snapshots the cell list once per tick and iterates with the in-memory `cell` captured at list time (`ReadyObserved=true`). A `kuke delete cell` that completes while that tick is blocked on the per-cell lock leaves `ReconcileCell` holding a pre-delete snapshot; the `!cgroupExists && ReadyObserved` heal branch then re-creates the cgroup **and** `ensureCellCgroup`'s `UpdateCellMetadata` rewrites the just-deleted `metadata.json` — silently resurrecting the cell in `Stopped` state.
- **Fix:** `ReconcileCell` now re-reads the cell from disk (`GetCell`) immediately after acquiring the per-cell lock and short-circuits on `ErrCellNotFound` with a new no-op `ReconcileOutcome{Vanished: true}` — placed *before* the sticky-`Failed` guard and the heal/derivation logic so neither path can rewrite the gone metadata. The lock makes the recheck reliable: the delete has fully completed before the reconcile observes absence.
- A new `Vanished` outcome field (distinct from `Deleted`) keeps the reconcile loop from re-deriving/persisting OutOfSync on the gone cell (a second metadata-rewrite resurrection vector for Config-lineage cells) **without** inflating the `CellsDeleted` auto-delete metric — the reconciler observed an external delete, it did not perform one.
- The `#855` post-reboot heal is unaffected: on a genuine reboot the `metadata.json` survives, so `GetCell` succeeds and the heal proceeds exactly as before. The on-disk metadata is the only discriminator between "reboot wiped the cgroup" and "delete removed the cell" — which is precisely why the pre-fix heal branch (keyed only on cgroup-absent + `ReadyObserved`) could not tell them apart.

### AC #2 (secret-delete unblock)

Verified transitively, not by a separate test: the secret guard's `secretReferences` (`internal/controller/secret.go:160`) reads `b.runner.ListCells(...)`. With the resurrection closed, the deleted cell's `metadata.json` is never rewritten, so `ListCells` no longer returns it and its `secretRef` no longer blocks `kuke delete secret`. The new regression test asserts the metadata stays absent after the racing tick, which is the root enabler of this AC.

## Test plan

- [x] `make test` (full CI target: `test-root` + `test-kukebuild`) — exit 0, zero failures
- [x] `go vet ./internal/controller/...` — clean
- [x] New `TestReconcileCell_DoesNotResurrectDeletedCell` — reproduces the delete-while-reconcile-blocked-on-lock race; asserts `Vanished` outcome, no `NewCgroup`/subtree call, and `metadata.json` stays absent. Verified it **fails** without the fix (NewCgroup called + metadata.json rewritten) and passes with it.
- [x] New `TestReconcileCell_HealVsDeletedMetadata` — pins the AC #4 distinction: metadata-survives → heals (`#855`); metadata-deleted → no heal (`#1251`).
- [x] Existing `TestReconcileCell_HealsWipedCgroupWhenReadyObserved` / `TestReconcileCell_DoesNotHealHalfCreateCell` still pass (post-reboot heal + half-CreateCell gate intact).

## Acceptance criteria

- [x] `kuke delete cell` followed by `kuke get ce` no longer shows the cell across reconcile ticks — the recheck leaves the metadata gone, and subsequent ticks no longer list the cell at all.
- [x] After deleting a cell that referenced a secret, `kuke delete secret` succeeds — see AC #2 note above (transitive via the no-rewrite invariant; `secretReferences` → `ListCells`).
- [x] `ReconcileCell` re-reads the cell from disk after acquiring the per-cell lock and returns a no-op outcome (no heal, no `metadata.json` rewrite, no cgroup recreate) when the cell's metadata is gone.
- [x] The `#855` post-reboot heal still heals a Ready cell whose cgroup is absent but whose `metadata.json` survives — regression covering reboot-survives-metadata vs. deleted-metadata.
- [x] Unit/regression test reproducing the delete-while-reconcile-blocked-on-lock race.

## Notes

- Updated `TestCellLifecycleLock_StartCellWaitsForReconcile` (`cell_lock_test.go`): its premise was "cell metadata deliberately NOT seeded so `ReconcileCell` reaches the gated `ExistsContainer` off the in-memory arg." The new recheck makes `ReconcileCell` require on-disk metadata, so the test now seeds the cell (but still not the space) — `ReconcileCell` reaches its gate, and `StartCell` still fails fast the moment it acquires the lock, now one step later at `ResolveSpaceCNIConfigPath`'s `GetSpace` (unseeded space) instead of its early `GetCell`. The lock-serialization contract it guards is unchanged.
- The separate `metadata file does not exist .../secrets/metadata.json` noise line called out in the issue is out of scope here (the issue flags it as a pre-existing follow-up).

Closes #1251

### DIFF
--- a/internal/controller/reconcile.go
+++ b/internal/controller/reconcile.go
@@ -109,7 +109,15 @@ func (b *Exec) ReconcileCells() (ReconcileResult, error) {
 					// them) and cells without the kukeon.io/config label.
 					// A persisted write counts toward CellsUpdated so the
 					// reconcile summary reflects the metadata flip.
-					if outcome.Deleted {
+					//
+					// Vanished cells (#1251) short-circuit the same way: the
+					// metadata is already gone, so re-deriving OutOfSync and
+					// persisting it would rewrite the just-deleted
+					// metadata.json — the very resurrection the post-lock
+					// recheck exists to prevent. Unlike Deleted, a Vanished
+					// outcome is left out of CellsDeleted (the reconciler
+					// observed an external delete, it did not perform one).
+					if outcome.Deleted || outcome.Vanished {
 						continue
 					}
 					syncUpdated, syncErr := reconcileCellOutOfSync(b.runner, reconciled)
--- a/internal/controller/runner/cell_lock_test.go
+++ b/internal/controller/runner/cell_lock_test.go
@@ -177,12 +177,17 @@ func TestCellLifecycleLock_StartCellWaitsForReconcile(t *testing.T) {
 	fake := &deleteCellFakeClient{existsContainerFn: gateExistsContainer(entered, release)}
 	r := newDeleteCellTestExec(t, fake)
 	seedDeleteCellRealm(t, r, realm)
-	// The cell metadata is deliberately NOT seeded: ReconcileCell reaches the
-	// gated ExistsContainer off the in-memory cell arg (it needs only the
-	// realm), while StartCell fails fast at its early GetCell (a filesystem
-	// read, not gated) the moment it acquires the lock. That makes "StartCell
-	// returned" a clean signal that it got past lock acquisition — without the
-	// lock it would return near-instantly rather than waiting on ReconcileCell.
+	// The cell metadata IS seeded but the space metadata is deliberately not:
+	// since #1251 ReconcileCell re-reads the cell from disk after acquiring the
+	// per-cell lock (GetCell) and short-circuits when it is gone, so it must
+	// find the cell metadata to proceed to the gated ExistsContainer off the
+	// in-memory cell arg. StartCell still fails fast the moment it acquires the
+	// lock — now one step later, at ResolveSpaceCNIConfigPath's GetSpace (a
+	// filesystem read, not gated), since the space is unseeded. That keeps
+	// "StartCell returned" a clean signal that it got past lock acquisition —
+	// without the lock it would return near-instantly rather than waiting on
+	// ReconcileCell.
+	seedDeleteCellCell(t, r, realm, space, stack, cellName)
 
 	cell := oneContainerCell(realm, space, stack, cellName)
 
--- a/internal/controller/runner/reconcile_heal_test.go
+++ b/internal/controller/runner/reconcile_heal_test.go
@@ -249,3 +249,177 @@ func seedNeverReadyCell(t *testing.T, r *Exec, realm, space, stack, cellName str
 		t.Fatalf("write cell metadata: %v", err)
 	}
 }
+
+// TestReconcileCell_DoesNotResurrectDeletedCell is the headline #1251 guard:
+// the delete-while-reconcile-blocked-on-lock race. ReconcileCells snapshots
+// the cell list once per tick and iterates with the in-memory value captured
+// at list time (ReadyObserved=true). A `kuke delete cell` that completes
+// while this tick is blocked on the per-cell lock leaves the reconciler
+// holding a pre-delete snapshot whose cgroup is now absent — pre-fix the
+// #855 heal branch fired (`!cgroupExists && ReadyObserved`) and
+// ensureCellCgroup recreated the cgroup AND rewrote the just-deleted
+// metadata.json, silently resurrecting the cell. The post-lock GetCell
+// recheck must short-circuit on ErrCellNotFound: no NewCgroup, no metadata
+// rewrite, and a Vanished outcome (not Deleted — the reconciler observed an
+// external delete, it did not perform one).
+func TestReconcileCell_DoesNotResurrectDeletedCell(t *testing.T) {
+	realm, space, stack, cellName := "default", "kukeon", "kukeon", "github-runner"
+	rootID := "root"
+	workloadID := "workload"
+	rootContainerdID := space + "_" + stack + "_" + cellName + "_" + rootID
+	workloadContainerdID := space + "_" + stack + "_" + cellName + "_" + workloadID
+
+	fake := &deleteCellFakeClient{
+		// Cgroup absent: DeleteCell removed it as part of the teardown
+		// the reconcile tick was blocked behind. This is the same probe
+		// result the post-reboot heal keys off — the bug is that the
+		// heal can't tell "metadata survives" from "metadata just deleted".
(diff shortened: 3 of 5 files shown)


eminwux:
**PR #1253 Review — `fix: reconcile re-reads cell from disk after lock to avoid resurrecting deleted cell`**

LGTM — the post-lock `GetCell` recheck short-circuits to a no-op `Vanished` outcome before the #855 heal, and since `DeleteCell` shares the same per-cell lock the recheck is race-free (delete fully completes before reconcile observes absence). `Vanished` is correctly excluded from `CellsDeleted` and the OutOfSync re-derivation; both regression tests reproduce the real reconcile race and CI (incl. `make e2e`) is green.

PR #1256, opened 2026-06-12
fix: privileged exposes host devices (WithHostDevices + WithAllDevicesAllowed)

eminwux:
## Summary
- `privileged: true` composed only `oci.WithPrivileged`, which handles capabilities, masked/readonly path clearing, writable sysfs/cgroupfs, and seccomp/apparmor unconfinement — but **no device handling**. With containerd's default deny-all device-cgroup wildcard still in place and `Linux.Devices` empty, a privileged cell could not `open(2)` any host device (e.g. `/dev/kvm`), diverging from both `ctr run --privileged` and `docker run --privileged`.
- Both privileged branches — `BuildContainerSpec` (`internal/ctr/spec.go`) and `BuildRootContainerSpec` (`internal/ctr/container_utils.go`) — now append `oci.WithAllDevicesAllowed, oci.WithHostDevices` after `oci.WithPrivileged`, exactly matching ctr's `run_unix.go` ordering. These are the only two `oci.WithPrivileged` call sites in the tree.
- Host devices are snapshotted into `Linux.Devices` at create time (same semantics as ctr/Docker); documented the device exposure + create-time snapshot caveat in `docs/site/manifests/container.md`'s `privileged` row.

## Test plan
- [x] `GOWORK=off go build ./internal/ctr/...` + `go vet ./internal/ctr/...` pass
- [x] New unit tests on the generated OCI spec assert `Linux.Devices` populated + `Linux.Resources.Devices` allow-all for privileged, and neither for non-privileged — `TestBuildContainerSpec_PrivilegedDevices` and `TestBuildRootContainerSpec_PrivilegedDevices` (covers both AC paths)
- [x] `make test` (full CI suite, GOWORK=off per Makefile) — exit 0

## Acceptance criteria
- [x] A `privileged: true` container can `open()` `/dev/kvm` (node present in `/dev` + device cgroup allows it) — covered by unit test on the generated OCI spec (`Linux.Devices` populated, `Linux.Resources.Devices` allow-all)
- [x] Root-container path (`BuildRootContainerSpec`) gets the same pair
- [x] `docs/site/manifests/container.md` `privileged` row documents device exposure + the create-time snapshot caveat

## Notes
- Restores the expected coarse-grained behavior; #1252 (per-device `devices:` field) remains the least-privilege successor.

Closes #1255

Issue #1255, opened 2026-06-12
ctr: privileged grants no host devices — WithPrivileged lacks WithHostDevices + WithAllDevicesAllowed


eminwux:
## Context

`privileged: true` on a ContainerSpec does not expose host device nodes and does not open the device cgroup, so a privileged cell still cannot open `/dev/kvm` (or any other host device). Reproduced live on an operator host (GitHub Actions runner cell for Android emulator CI): host has `/dev/kvm` (`crw-rw---- root:kvm`), cell has `privileged: true`, yet inside the container `ls /dev/kvm` → `No such file or directory`, and the emulator aborts with `x86 emulation currently requires hardware acceleration!`.

This diverges from both Docker (`--privileged` exposes all host devices + opens the device cgroup) and `ctr run --privileged`, which an operator reasonably expects `privileged: true` to match.

## Root cause

`internal/ctr/spec.go:384` (and the root-container path at `internal/ctr/container_utils.go:170`) appends only `oci.WithPrivileged`. In containerd v2.2.4, `oci.WithPrivileged` (`pkg/oci/spec_opts.go:1423`) composes capabilities, masked/readonly path clearing, writable sysfs/cgroupfs, and seccomp/apparmor unconfinement — **no device handling**. Device exposure is two separate opts that `ctr run --privileged` adds itself (`cmd/ctr/commands/run/run_unix.go:239`):

```go
opts = append(opts, oci.WithPrivileged, oci.WithAllDevicesAllowed, oci.WithHostDevices)
```

Compounding it, containerd's default spec carries a deny-all device-cgroup wildcard (`pkg/oci/spec.go:211`, `{Allow: false, Access: rwm}`), so even bind-mounting a node via `volumes:` fails at `open()` with `EPERM`. Net effect: **there is no manifest-level path to any host device today**, privileged or not.

## Proposal

Match `ctr run --privileged`: in both `Privileged` branches, append `oci.WithAllDevicesAllowed` and `oci.WithHostDevices` after `oci.WithPrivileged`. Host devices are snapshotted into `Linux.Devices` at create time (same semantics as ctr/Docker — devices appearing on the host later require a container recreate; worth a sentence in `docs/site/manifests/container.md`'s `privileged` row).

## Acceptance criteria

[...]

PR #1262, opened 2026-06-12
fix: resolve device nodes against host root when kukeond is containerized

### LINKED ISSUE #1261: ctr: device resolution runs in kukeond's mount namespace — devices/privileged see the daemon cell's /dev, not the host's
## Repro

On a standard v0.7.0 install (kukeond containerized by `kuke init`, host has `/dev/kvm`):

```yaml
containers:
  - id: runner
    devices:
      - /dev/kvm
```

`sudo kuke apply -f cell.yaml` →

```
Error: failed to create cell: failed to create container default_default_github-runner_runner: failed to create container from spec: failed to create container: no such file or directory
```

while `ls -l /dev/kvm` on the host shows the node present.

## Observed

Daemon log: `failed to create container ... err="no such file or directory"` — no device path in the message, so the "clear error for a missing node" AC from #1252 is also unmet.

## Root cause

`deviceSpecOpts` (`internal/ctr/spec.go:766`) uses `oci.WithLinuxDevice(path, "rwm")`, which calls `DeviceFromPath` → `os.Stat(path)` **in the kukeond process**. The kukeond cell's mount namespace binds only `/run/kukeon` and `/opt/kukeon`; its `/dev` is the container's own devtmpfs (default nodes only), so any host device path fails the stat. Same resolve-in-daemon-namespace trap as `fromFile` secrets (#1250).

#1255 has the identical blindness: `oci.WithHostDevices` enumerates `/dev` of the calling process, so `privileged: true` on a containerized daemon replicates the kukeond container's minimal `/dev` instead of the host's — the v0.7.0 fix only behaves as intended when the daemon runs un-containerized (dev loops), which is also why unit tests on the generated spec (statting nodes that exist in the test env) pass.

## Suggested fix

The kukeond cell already runs with `hostPID: true` (required by the CNI bridge path), so the host's filesystem is reachable at `/proc/1/root` from inside the cell. Resolve device nodes against a host-root prefix: replace `oci.WithLinuxDevice(path, …)` with a variant that stats `<hostRoot>/dev/...` (where `hostRoot` is `/proc/1/root` when containerized, `/` otherwise — or a daemon config knob) and writes the un-prefixed path into `Linux.Devices`. Apply the same treatment to the privileged path (enumerate `<hostRoot>/dev` instead of `oci.WithHostDevices`). Alternative: `kuke init` bind-mounts host `/dev` (ro, at e.g. `/host/dev`) into the kukeond cell and resolution targets that. Either way, wrap the ENOENT with the device path so the create-time error names the missing node.

## Notes

- Surfaced in the same operator session as #1248 / #1249 / #1250 / #1252 / #1255 / #1260 (GitHub Actions runner cell, Android emulator needing `/dev/kvm`).
- An e2e-shaped test (spec generated through a daemon that is itself containerized) is what the unit tests on the generated OCI spec structurally cannot catch.

eminwux:
## Summary
- `deviceSpecOpts` resolved `devices[]` host nodes with `oci.WithLinuxDevice` → `DeviceFromPath` → `os.Stat`, which runs **in the kukeond process**. When kukeond is containerized (the `kuke init` default), the daemon cell's mount namespace exposes only its own minimal `/dev`, so a host path like `/dev/kvm` fails the stat and container create dies with a path-less `no such file or directory`. The kukeond cell runs `hostPID: true`, so the host root — and the host's `/dev` — is reachable at `/proc/1/root`; device resolution now targets that prefix when the daemon is in a private mount namespace, and `/` (unchanged behavior) otherwise.
- The node is stat'd through the host-root prefix but exposed in the container at the **un-prefixed** path, and a missing node now fails create with an error that **names the device** (`resolve device "/dev/kvm": ...`) — closing the unmet "clear error for a missing node" item the issue calls out.
- `privileged: true` now pairs `oci.WithPrivileged` (which grants *no* device access on its own) with `oci.WithAllDevicesAllowed` + a host-root-aware enumeration of the host's `/dev`, restoring `docker run --privileged` / `ctr run --privileged` device parity at both privileged call sites (`BuildContainerSpec`, `BuildRootContainerSpec`). Un-containerized this reuses the well-tested `oci.WithHostDevices`; containerized it walks `/proc/1/root/dev` and rewrites container paths back to `/dev/...`.

## Reinterpreted scope (premise note for the reviewer)
The issue's suggested fix says "enumerate `<hostRoot>/dev` instead of `oci.WithHostDevices`" for the privileged path, implying #1255's fix (PR #1256) had landed. **#1256 was closed unmerged** (LGTM'd, then closed by the owner — superseded by this issue's correct-namespace approach). So on `origin/main` the privileged path is bare `oci.WithPrivileged` with *zero* host devices. This PR therefore **adds** host-root-aware device exposure rather than replacing `WithHostDevices`. The intent — privileged containers see the host's devices, resolved correctly even when the daemon is containerized — is unchanged; the diff just also carries the device-exposure #1256 would have added, done host-root-aware. Flagging so the reviewer can push back if absorbing #1255/#1256's behavior here oversteps.

## Test plan
- [x] `GOWORK=off go build ./internal/ctr/...` + `go vet ./internal/ctr/...` — clean
- [x] `make test` (full CI suite, `GOWORK=off` per Makefile) — exit 0
- [x] New unit tests: `TestHostLinuxDeviceOpt_ResolvesViaHostRoot` (stat goes through `/proc/1/root`, node exposed at `/dev/kvm`, cgroup allow rule emitted), `TestHostLinuxDeviceOpt_MissingNodeNamesPath` (error names the path), `TestEnumerateHostDevices_RewritesContainerPath` (privileged walker rewrites to `/dev/...`, skips fifos), `TestResolveDeviceHostRoot_UnContainerized`, and privileged-parity tests for both spec builders + a non-privileged negative.
- [x] Detection sanity-checked at runtime in a real cell: with PID 1 sharing the cell's root mount, `daemonInPrivateMountNS()` returns false → resolves against `/` (historical behavior preserved). Only the `hostPID` kukeond cell, whose PID 1 is the *host* init with a different root, flips to `/proc/1/root`.
- [ ] Live containerized device-passthrough (build the spec through a `hostPID` kukeond cell that is itself containerized, confirm `/dev/kvm` resolves) — **not run**: this is the e2e-shaped test the issue notes "unit tests on the generated OCI spec structurally cannot catch", and the dev host has no standalone containerd socket / full `kuke init` stack to bring up the nested daemon safely. The host-root *resolution* logic is exercised deterministically by the unit tests above via an injected resolver; the un-containerized privileged-parity path is identical to PR #1256, which was `make e2e`-green. Recommend the reviewer (or a follow-up e2e job) verify the live containerized stat on a clean host.

Closes #1261

### DIFF
--- a/internal/ctr/container_utils.go
+++ b/internal/ctr/container_utils.go
@@ -167,8 +167,13 @@ func BuildRootContainerSpec(
 		specOpts = append(specOpts, oci.WithEnv(env))
 	}
 
+	// Mirror BuildContainerSpec's privileged path: oci.WithPrivileged grants no
+	// device access, so pair it with an allow-all device cgroup and a
+	// host-root-aware enumeration of the host's /dev (resolved against the host
+	// root when kukeond is containerized) to restore `docker run --privileged`
+	// device parity (issue #1261).
 	if rootSpec.Privileged {
-		specOpts = append(specOpts, oci.WithPrivileged)
+		specOpts = append(specOpts, oci.WithPrivileged, oci.WithAllDevicesAllowed, privilegedHostDevicesOpt())
 	}
 
 	// Host network: drop the network LinuxNamespace entry from the OCI spec so
--- a/internal/ctr/devices.go
+++ b/internal/ctr/devices.go
@@ -1,0 +1,211 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package ctr
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/containerd/containerd/v2/core/containers"
+	"github.com/containerd/containerd/v2/pkg/oci"
+	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// hostRootViaPID1 is the path through which a process that shares the host PID
+// namespace reaches the host's root filesystem: PID 1's root, exposed by
+// procfs as a magic symlink. kukeond runs inside the kuke-system kukeond cell
+// with hostPID: true (required by the CNI bridge path), so /proc/1/root is the
+// host root even though the cell's own mount namespace exposes only a minimal
+// /dev that contains none of the host's pass-through device nodes.
+const hostRootViaPID1 = "/proc/1/root"
+
+// deviceHostRoot returns the filesystem prefix under which the kukeond process
+// must resolve *host* device nodes when building a container spec. Device
+// resolution (the os.Stat behind oci.DeviceFromPath / oci.HostDevices) runs in
+// the kukeond process, not in the target container — so when kukeond is
+// containerized its own mount namespace, not the host's, is what gets stat'd.
+// The cell binds only /run/kukeon and /opt/kukeon; its /dev is the cell's
+// devtmpfs (default nodes only), so a host device path like /dev/kvm fails the
+// stat. Resolving against the host root reachable at /proc/1/root fixes that.
+//
+// Indirected through a var so tests can pin the prefix without a containerized
+// daemon.
+var deviceHostRoot = resolveDeviceHostRoot
+
+func resolveDeviceHostRoot() string {
+	if daemonInPrivateMountNS() {
+		return hostRootViaPID1
+	}
+	return "/"
+}
+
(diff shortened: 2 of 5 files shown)


eminwux:
**PR #1262 Review — `fix: resolve device nodes against host root when kukeond is containerized`**

LGTM. Device resolution is correctly indirected through `deviceHostRoot()`: `daemonInPrivateMountNS()` returns true only for the hostPID kukeond cell (host PID 1 with a different root) and false otherwise, preserving the historical resolve-against-`/` behavior for un-containerized and non-hostPID containers. The stat-through-`/proc/1/root` / expose-at-unprefixed-path split is correct and covered by `TestHostLinuxDeviceOpt_ResolvesViaHostRoot`. Privileged path pairs `oci.WithAllDevicesAllowed` with a host-root-aware `/dev` walker that mirrors containerd's `getDevices` skip list (pts/shm/fd, console, fifos, `ErrNotADevice`), restoring `--privileged` device parity at both spec builders. The reinterpreted-scope premise checks out — on `origin/main` both privileged sites are bare `oci.WithPrivileged` with zero device exposure, so this adds (not replaces) host-device exposure, matching #1261's suggested fix.

All 9 new symbols are referenced; the single commit is GPG-verified; no secret/PII surface.

Non-blocking — verify before merge:
- The deferred live containerized device-passthrough e2e (build a spec through a containerized hostPID kukeond cell, confirm `/dev/kvm` resolves) is host-gated and documented as unrunnable on the agent sandbox. The resolution logic is deterministically unit-tested; recommend running the live stat on a clean host before merge.

LGTM to proceed.

PR #1263, opened 2026-06-12
feat: installer detects daemon version skew and points at kuke daemon recreate

### LINKED ISSUE #1260: install: installer leaves a stale kukeond running — version-blind skip of the init phase
## Context

Upgrading via the one-line installer silently leaves the old daemon in place. Observed live on an operator host: `curl install.sh | bash` installed v0.7.0 binaries, printed `✓ kukeond already running at /run/kukeon/kukeond.sock — skipping \`kuke init\``, and `kuke version` then reported `Client: v0.7.0 / Daemon: v0.6.1` with a mismatch warning. The operator had to discover `kuke daemon reset` + re-init on their own (and `kuke daemon recreate` exists precisely for this, but nothing points at it). The skip check at `scripts/install.sh:386` is liveness-only — socket present → skip — and never compares versions, so every binary upgrade on a live host ends in a client/daemon skew the installer just declared a success.

## Proposal

After installing binaries, when the socket is live, compare the running daemon's version against the just-installed client (`kuke version` already surfaces both and detects the mismatch). On mismatch:

- run `kuke daemon recreate --kukeond-image <repo>:<installed-version>` (workload cells are untouched — the recreate only cycles the kuke-system kukeond cell), or
- if auto-recreate is judged too aggressive for an installer, print a loud, copy-pasteable instruction (`sudo kuke daemon recreate --kukeond-image …`) and exit with the mismatch stated — anything but `✓ … skipping` on a host the script just made inconsistent.

A `KUKE_SKIP_DAEMON_UPGRADE=1` escape hatch preserves the current behavior for operators who intentionally pin an older daemon.

## Acceptance criteria

- [ ] Installer on a live host with a version-matched daemon behaves as today (skip, no daemon churn).
- [ ] Installer on a live host with an older daemon either recreates it with the installed version's image or fails loudly with the exact recreate command — it never prints an unqualified success while leaving the skew in place.
- [ ] Escape hatch documented in the install docs alongside the existing `KUKE_SKIP_*` knobs.
- [ ] Upgrade flow (installer + `kuke daemon recreate`) documented as the canonical upgrade procedure in the install/upgrade docs page.

## Notes

- `kuke daemon recreate` shipped via #221-adjacent work and is documented in `docs/site/cli/kuke-daemon.md` — this issue is about the installer not using/pointing at it.
- Surfaced during the same operator session as #1248 / #1249 / #1250 / #1252 / #1255 (v0.6.1→v0.7.0 upgrade to pick up the device-passthrough fixes).

eminwux:
## Summary

- The installer's post-install init skip at `scripts/install.sh` was liveness-only: socket present → `✓ … skipping`, never comparing versions. A binary upgrade on a live host (`curl install.sh | bash`) thus left the old `kukeond` running and declared it a success, ending in a client/daemon version skew the installer just called done (issue #1260).
- On a live socket the installer now compares the running daemon's version against the just-installed client via `kuke version --strict` (which already prints the `Client:`/`Daemon:` lines and exits non-zero only on a real mismatch):
  - **version-matched** → skipped silently, exactly as today (no daemon churn).
  - **older daemon** → prints the exact `sudo kuke daemon recreate --kukeond-image <repo>:<installed-version>` command and exits non-zero, instead of an unqualified success. It does **not** auto-cycle the daemon — a recreate tears down and re-provisions the kuke-system kukeond cell, too aggressive to do unprompted from a piped `curl | bash`.
  - **present-but-dead socket** (`kuke version` reports daemon unreachable, exit 0) → falls through to the historical skip rather than a spurious recreate prompt.
- `KUKE_SKIP_DAEMON_UPGRADE=1` escape hatch preserves the historical silent-skip for operators intentionally pinning an older daemon.
- Documented the env var in the install-docs override table and added a canonical **Upgrade** section (re-run installer → recreate daemon) to `docs/site/install/install-linux.md`.

- **Gate-vs-flip / blast radius (per dev CLAUDE.md):** the issue offered auto-recreate *or* loud-instruction-and-exit. I chose the lower-blast-radius **warn-and-exit-non-zero** default — auto-cycling a running daemon from a piped installer is a surprising, non-reversible side effect; the printed command lets the operator recreate when ready. The escape hatch (`KUKE_SKIP_DAEMON_UPGRADE=1`) restores the prior behavior. Happy to flip to auto-recreate if you'd rather.
- **Image-ref derivation:** the recreate command uses `ghcr.io/${KUKE_REPO}:${KUKE_VERSION}`, mirroring the documented default `KukeondImageRepo = ghcr.io/eminwux/kukeon` (`cmd/config/version.go:30`) ← `KUKE_REPO = eminwux/kukeon`. This keeps forks correct (a fork's `KUKE_REPO` maps to its own `ghcr.io/<owner>/<repo>` namespace) without parsing the daemon's stored image ref in shell.
- **`scripts/install.sh` ↔ `docs/site/install.sh` sync:** the two are kept byte-identical (the latter is the mkdocs-published copy). Re-rendered with `make install.sh` so the `installer.yaml` sync job passes.

## Test plan

- [x] `bash -n scripts/install.sh` + `bash -n docs/site/install.sh` — syntax clean.
- [x] `make install.sh` then `diff scripts/install.sh docs/site/install.sh` — identical (installer.yaml sync job will pass).
- [x] Isolated harness exercising `handle_running_daemon` with a fake `kuke` across all four branches: **match** → silent skip, no skew; **mismatch** → loud fail + correct `recreate --kukeond-image ghcr.io/eminwux/kukeon:v0.7.0` + `KUKE_DAEMON_SKEW=1`; **unreachable** → historical skip; **`KUKE_SKIP_DAEMON_UPGRADE=1`** → silent skip.
- [x] `bash scripts/install.sh --check` — unaffected; still exits non-zero with the containerd hint (my change is inside the live-socket branch, after the `--check`/`KUKE_SKIP_INIT` returns).
- [x] `make test` — exit 0 (no Go change; `test.yaml` is path-filtered away from this shell/docs PR, ran as insurance).

### AC coverage
- [x] Live host, version-matched daemon → skip, no daemon churn.
- [x] Live host, older daemon → never prints unqualified success; exits with the exact recreate command (auto-recreate not chosen — see design notes).
- [x] Escape hatch documented alongside the existing `KUKE_SKIP_*` knobs.
- [x] Upgrade flow (installer + `kuke daemon recreate`) documented as the canonical upgrade procedure.

Closes #1260

### DIFF
--- a/docs/site/install.sh
+++ b/docs/site/install.sh
@@ -21,6 +21,10 @@
 #   KUKE_SKIP_CHECKSUM=1   skip .sha256 verification (NOT recommended)
 #   KUKE_SKIP_KUKEBUILD=1  skip downloading/installing kukebuild (disables
 #                          `kuke build`; for release tags predating its asset)
+#   KUKE_SKIP_DAEMON_UPGRADE=1
+#                          on a live host, skip the daemon version check and
+#                          keep an intentionally pinned older daemon (preserves
+#                          the historical silent-skip behavior)
 
 set -euo pipefail
 
@@ -30,6 +34,12 @@ KUKE_VERSION="${KUKE_VERSION:-}"
 KUKE_SKIP_INIT="${KUKE_SKIP_INIT:-}"
 KUKE_SKIP_CHECKSUM="${KUKE_SKIP_CHECKSUM:-}"
 KUKE_SKIP_KUKEBUILD="${KUKE_SKIP_KUKEBUILD:-}"
+KUKE_SKIP_DAEMON_UPGRADE="${KUKE_SKIP_DAEMON_UPGRADE:-}"
+
+# Set by handle_running_daemon when a live host's daemon is older than the
+# binaries just installed; main exits non-zero on this so the installer never
+# prints an unqualified success banner over a client/daemon version skew.
+KUKE_DAEMON_SKEW=""
 
 MODE="install"
 
@@ -67,6 +77,9 @@ Env overrides:
   KUKE_SKIP_INIT=1       Skip the post-install \`kuke init\` step.
   KUKE_SKIP_CHECKSUM=1   Skip .sha256 verification (NOT recommended).
   KUKE_SKIP_KUKEBUILD=1  Skip kukebuild (disables \`kuke build\`).
+  KUKE_SKIP_DAEMON_UPGRADE=1
+                         On a live host, skip the daemon version check and keep
+                         an intentionally pinned older daemon.
 EOF
 }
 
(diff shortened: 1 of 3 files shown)


eminwux:
**PR #1263 Review — `feat: installer detects daemon version skew and points at kuke daemon recreate`**

Correct and self-contained. The gate's entire logic hinges on `kuke version --strict` semantics, and they match the implementation exactly: unreachable daemon → exit 0 (`cmd/kuke/version/version.go:58,71`), real mismatch → non-zero (`:81`), matched → exit 0. Control flow checks out end-to-end — `KUKE_SKIP_INIT` short-circuits before the gate, `--check` exits before `do_install`, `KUKE_DAEMON_SKEW` propagates as a global across the function boundary, `fail()` doesn't exit so the handler keeps printing, and `local version_out` is split from the assignment so the exit code isn't masked. No secrets, no dead code, both `install.sh` copies byte-identical (sync job green), all commits verified.

One non-blocking nit:

- **Message accuracy** `scripts/install.sh:430`: on a present-but-dead socket `kuke version --strict` exits 0, so the handler prints `(version-matched) — skipping` even though the daemon is unreachable, not matched. Behavior is correct (skip, no churn, matches historical), only the label is imprecise — fold the unreachable case into a neutral message if you want, but not worth gating.

LGTM — lower-blast-radius warn-and-exit default is the right call for a piped installer, and the escape hatch + Upgrade docs round it out.

PR #1266, opened 2026-06-12
fix: silence cobra usage/error dump on kuke version --strict mismatch

### LINKED ISSUE #1264: cli: kuke version --strict dumps full usage/help on version mismatch (missing SilenceUsage)
## Repro

On `eminwux/kukeon` at `origin/main` (post #1263), with a host whose running kukeond is older than the just-installed binaries:

```
sudo kuke version --strict
```

This is the exact command the installer's skew handler runs (`scripts/install.sh:432`, `handle_running_daemon`) when re-running `install.sh` on a live host after a binary upgrade — issue #1260 / PR #1263.

## Observed

On a real version mismatch, `kuke version --strict` exits non-zero (correct) but **also dumps the full cobra usage/help block and a duplicate `Error:` line** to stderr:

```
Client: v0.7.1
Daemon: v0.7.0
Warning: version mismatch (client: v0.7.1, daemon: v0.7.0)
Error: version mismatch: client=v0.7.1 daemon=v0.7.0
Usage:
  kuke version [flags]
Flags:
  -h, --help     help for version
      --strict   exit with non-zero status on version mismatch
Global Flags:
      --configuration string       ...
      --containerd-socket string   ...
      --host string                ...
      --no-daemon                  ...
      ...
```

`handle_running_daemon` captures all of this via `version_out="$($SUDO kuke version --strict 2>&1)"` and re-prints it verbatim (`printf '%s\n' "$version_out" | sed 's/^/    /'`, `scripts/install.sh:439`) inside the recreate notice. Result: the clean 3-line skew hint the handler was designed around is buried under a usage dump + a redundant `Error:` line during the installer's upgrade path.

## Root cause

`cmd/kuke/version/version.go` — `NewVersionCmd`'s `&cobra.Command{...}` literal (lines 40-87) sets **neither** `SilenceUsage` nor `SilenceErrors`. The root `kuke` command (`cmd/kuke/kuke.go:69-131`) also sets neither. Cobra prints the usage string after a `RunE` error unless the failing command *or* the root sets `SilenceUsage: true`; it prints the `Error:` line unless one sets `SilenceErrors: true`.

`version` is the odd one out — every other leaf command (`apply`, `build`, `create/*`, `daemon/*`, …) sets `SilenceUsage: true`. And `--strict` exists specifically to return non-zero on mismatch, so this error-returning path is hit by design, making `version` the one CLI path that spews usage on an expected failure.

## Suggested fix

Add to the command literal in `cmd/kuke/version/version.go`, matching every sibling command:

```go
SilenceUsage:  true,
SilenceErrors: true,
```

`SilenceUsage: true` removes the usage dump. `SilenceErrors: true` additionally drops cobra's duplicate `Error: version mismatch…` line, leaving only the intentional `Warning: version mismatch…` that `version.go:78` prints itself — so the installer's captured `version_out` reduces to the clean `Client:` / `Daemon:` / `Warning:` block the skew message expects. Fixing it at the command (rather than filtering captured output in `install.sh`) also benefits any other caller of `kuke version --strict`. The existing assertion in `cmd/kuke/version/version_test.go` ("daemon up, mismatch with --strict returns error") should be extended to assert the error path emits no usage/help text.

Surfaced during investigation of the install.sh recreate/skew flow added in #1263 (issue #1260).

eminwux:
## Summary

- `cmd/kuke/version/version.go`'s `NewVersionCmd` literal set neither `SilenceUsage` nor `SilenceErrors`, so a real version mismatch under `kuke version --strict` (an expected, by-design non-zero exit) dumped the full cobra usage/help block **and** a duplicate `Error: version mismatch…` line to stderr — on top of the intentional `Warning: version mismatch…` the command prints itself.
- The installer's skew handler (`scripts/install.sh` `handle_running_daemon`, added in #1263) captures `kuke version --strict 2>&1` and re-prints it verbatim inside the recreate notice, so the clean 3-line skew hint was buried under the usage dump.
- `version` was the odd command out — every other leaf command (`apply`, `build`, `create/*`, `daemon/*`, …) already sets `SilenceUsage: true`. Fixing it at the command (not by filtering captured output in `install.sh`) also benefits any other caller of `kuke version --strict`.

Added `SilenceUsage: true` (drops the usage dump) and `SilenceErrors: true` (drops cobra's duplicate `Error:` line; `cmd/main.go`'s `execRoot` only maps the error to exit code 1 and never prints it, so nothing else re-emits it). The command's own `Warning: version mismatch…` remains.

Extended `version_test.go` with `TestVersionCmdRun_StrictMismatchNoUsage`, asserting the `--strict` error path emits the warning but none of `Usage:` / `Global Flags:` / `help for version` / `Error:`.

## Test plan
- [x] `make test-root` — full main-module suite, all green (exit 0)
- [x] `make test-kukebuild` — green (untouched by this change)
- [x] `go test ./cmd/kuke/version/...` — new + existing version tests pass
- [x] `go vet ./cmd/kuke/version/...` — clean

Closes #1264

### DIFF
--- a/cmd/kuke/version/version.go
+++ b/cmd/kuke/version/version.go
@@ -38,8 +38,10 @@ type MockDaemonClientKey struct{}
 
 func NewVersionCmd() *cobra.Command {
 	versionCmd := &cobra.Command{
-		Use:   "version",
-		Short: "Print the version number",
+		Use:           "version",
+		Short:         "Print the version number",
+		SilenceUsage:  true,
+		SilenceErrors: true,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			clientVersion := config.Version
 			_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Client: %s\n", clientVersion)
--- a/cmd/kuke/version/version_test.go
+++ b/cmd/kuke/version/version_test.go
@@ -223,3 +223,40 @@ func TestVersionCmdRun_MismatchStderr(t *testing.T) {
 		t.Errorf("stderr should contain daemon version; got: %q", errBuf.String())
 	}
 }
+
+// TestVersionCmdRun_StrictMismatchNoUsage verifies that the --strict error path
+// emits neither cobra's usage/help block nor its duplicate "Error:" line, leaving
+// only the command's own "Warning: version mismatch" on stderr. Regression guard
+// for the installer skew handler (scripts/install.sh handle_running_daemon), which
+// captures `kuke version --strict 2>&1` and re-prints it verbatim.
+func TestVersionCmdRun_StrictMismatchNoUsage(t *testing.T) {
+	cmd := version.NewVersionCmd()
+	outBuf := &bytes.Buffer{}
+	errBuf := &bytes.Buffer{}
+	cmd.SetOut(outBuf)
+	cmd.SetErr(errBuf)
+	ctx := context.WithValue(context.Background(), version.MockDaemonClientKey{}, &fakeDaemonClient{
+		pingVersionFn: func(_ context.Context) (string, error) {
+			return "dummy", nil
+		},
+	})
+	cmd.SetContext(ctx)
+	cmd.SetArgs([]string{"--strict"})
+
+	err := cmd.Execute()
+	if err == nil {
+		t.Fatal("expected error on --strict mismatch, got nil")
+	}
+
+	// The command's own warning must still be present.
+	if !strings.Contains(errBuf.String(), "Warning: version mismatch") {
+		t.Errorf("expected version mismatch warning on stderr; got: %q", errBuf.String())
+	}
+	// No cobra usage/help dump (SilenceUsage) and no duplicate "Error:" line
+	// (SilenceErrors).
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #1266 Review — `fix: silence cobra usage/error dump on kuke version --strict mismatch`**

Clean, well-tested fix. `version`'s `RunE` returns an error in exactly one place (`cmd/kuke/version/version.go:81`, strict mismatch) and already prints its own `Warning: version mismatch…` first; `execRoot` (`cmd/main.go:50`) maps errors to exit 1 without printing, so `SilenceErrors: true` drops only cobra's duplicate `Error:` line with no information loss. `SilenceUsage: true` matches every sibling leaf command. New `TestVersionCmdRun_StrictMismatchNoUsage` exercises the exact path. No leaks, no dead code, commits signed, CI green.

One non-blocking note:

- **Behavior** `cmd/kuke/version/version.go:42-43`: siblings keep `SilenceErrors: false`, so flag-parse typos still surface `Error: unknown flag`; setting it `true` here also silences those, so `kuke version --typo` exits 1 silently. Negligible given `version`'s tiny flag surface and a deliberate tradeoff — flagging for awareness, not blocking.

LGTM to proceed.

Issue #1268, opened 2026-06-12
cell: a terminal cell (Failed/Error/Exited) cannot be restarted or re-run — start/run/restart disagree


eminwux:
## Observed

A cell whose workload exited non-zero lands in `Failed` (see #1267) and then **cannot be restarted or re-run** — the only documented recovery is delete + recreate, even though the cell's containers are simply terminal (exactly like a `Stopped` cell, which restarts fine):

```
$ kuke run mycell
Error: cell "mycell" exists in Failed state; delete it with `kuke delete cell mycell` before re-running

$ kuke restart mycell
Error: cell "mycell" exists in Failed state; delete it with `kuke delete cell mycell` before restarting
```

Worse, the three restart-ish verbs disagree on what a terminal/degraded cell may do:

| Verb | On `Stopped` | On `Failed` |
| --- | --- | --- |
| `kuke start` | starts | **no guard** — tries to start |
| `kuke run <cell>` | starts | **rejected** (delete-then-rerun) |
| `kuke restart` | starts | **rejected** (delete-then-rerun) |

Per design decision: **`Failed` should be restartable too** (route through a recreate-style recovery), not just `Error`/`Exited`.

## Root cause

`kuke run` and `kuke restart` bucket `Failed` together with the genuinely-unrecoverable `Pending`/`Unknown` and refuse:

- `kuke run <cell>` — `cmd/kuke/run/run.go:667-677`:
  ```go
  case v1beta1.CellStatePending, v1beta1.CellStateFailed, v1beta1.CellStateUnknown:
      return fmt.Errorf("cell %q exists in %s state; delete it with `kuke delete cell %s` before re-running", …)
  ```
- `kuke restart` — `cmd/kuke/restart/restart.go:226-233`: identical bucketing + "before restarting".
- `kuke start` — `internal/controller/start_cell.go:77-84`: has **no** `Failed` guard (only rejects `Ready`), so it diverges from the other two.

The guards can't be written correctly today because `Failed` conflates "workload crashed (recoverable)" with "kukeon bring-up failed" — see #1267. The reconciler also pins it: `cellStateIsSticky(Failed)` (`internal/controller/runner/refresh.go:866`) makes `Failed` terminal so a reconcile tick never moves it.

[...]

PR #1269, opened 2026-06-12
fix: split cell/container terminal states into Exited and Error

### LINKED ISSUE #1267: cell: status conflates workload exit with kukeon failure (Failed) and operator-stop with clean exit (Stopped)
## Observed

A cell's terminal `STATE` (as shown by `kuke get cells`) collapses four semantically distinct outcomes into two labels:

| Event | Current `STATE` |
| --- | --- |
| Operator `kuke stop` / `kuke kill` | `Stopped` |
| All workloads exit **0** (clean finish) | `Stopped` |
| A workload exits **non-zero** (crash) | `Failed` |
| kukeon could not create/start/sustain the cell | `Failed` |

Two conflations result:

1. **`Failed` means both "the workload crashed" and "kukeon's bring-up failed."** A workload exiting `127` and a cell kukeon couldn't start are indistinguishable in the `STATE` column — even though one is the workload's fault and one is kukeon's. The distinction *exists* only in `Status.Reason` (`WorkloadFailed` vs `StartCellFailed`/`CreateCellFailed`/`StartContainerFailed`/`RecreateCellFailed`), which `kuke get cells` does **not** render (only `-o yaml/json` shows it).
2. **`Stopped` means both "operator stopped it" and "it finished cleanly on its own."** No persisted field distinguishes an operator-initiated stop from a workload's clean self-exit.

This is the root of the companion bug (#1268) — restart guards can't make a correct decision against `Failed` while it carries both meanings.

`Failed` should be reserved for "a problem within kukeon." Kubernetes is the reference: Pod phase `Succeeded` (all containers exit 0) vs `Failed` (≥1 non-zero); container `state.terminated.reason` is `Completed` (0) vs `Error` (non-zero); infrastructure problems never surface as the app having "Failed."

## Root cause

The cell-level derivation (post-#1206) already keys on the exit code but only produces a binary `Failed`/`Stopped`:

- `deriveCellStateFromNonRootContainerStatuses` — `internal/controller/runner/refresh.go:1113-1168` (non-zero → `CellStateFailed`, else `CellStateStopped`)
- `deriveCellStateFromRootContainer` — `internal/controller/runner/refresh.go:1205-1232` (same binary split)
- `containerExitedNonZero` — `internal/controller/runner/refresh.go:1178` (the `ExitCode != 0 || State==Failed` test)

The container level conflates even harder: every containerd-stopped task maps to `ContainerStateStopped` regardless of code (`ConvertContainerdStatusToContainerState`, `internal/ctr/task.go:70`); the code lives only in the separate `ExitCode` field, and `ContainerState.String()` (`pkg/api/model/v1beta1/container.go:496`) renders "Stopped" for both `0` and `127`.

Operator stop/kill set `CellStateStopped` explicitly: `internal/controller/runner/stop.go:293`, `internal/controller/runner/kill.go:267`. `kuke get cells` renders only `State.String()` (`cmd/kuke/get/cell/cell.go:312`).

## Proposed

Split each conflated terminal state in two — at both the cell and container level:

| Trigger | Proposed cell state |
| --- | --- |
| operator `kuke stop` / `kill` | `Stopped` |
| all workloads exit 0 | **`Exited`** (k8s term: `Completed`) |
| a workload exits non-zero | **`Error`** |
| kukeon bring-up / runtime error | `Failed` (reserved for kukeon's fault) |

Naming call (open for review): `Exited` vs `Completed`, `Error` vs keeping `Failed` for the workload case. Recommendation: `Exited` + `Error` to honor the operator-facing intent and maximize the distinction; `Completed` is the more self-documenting success label if preferred.

Container level mirrors: a containerd-stopped task renders `Exited` (code 0) or `Error` (code > 0); `ContainerStateFailed` stays reserved for kukeon's own container bring-up failures.

## Acceptance criteria

- [ ] New cell states `Exited` and `Error` added (and `Failed` reserved for kukeon-fault paths only). Enums updated in lockstep — `internal/modelhub/cell.go:188`, `pkg/api/model/v1beta1/cell.go:242` — **appended last** to preserve ordinals (the int-cast conversion in `internal/apischeme/scheme.go` requires it; same constraint the `ContainerStateNotCreated` comment documents).
- [ ] Container states mirror the split — `internal/modelhub/container.go:309`, `pkg/api/model/v1beta1/container.go:481` (append-only) — with new printable strings in `pkg/api/model/v1beta1/consts.go:84` and `String()` cases (`container.go:496`, the cell `String()`).
- [ ] `deriveCellStateFromNonRootContainerStatuses` (`refresh.go:1113`) and `deriveCellStateFromRootContainer` (`refresh.go:1205`) produce the three-way `Error` / `Exited` / `Ready` (never `Stopped`) split for self-observed exits. The reconciler only ever yields `Exited`/`Error`/`Ready`/`Unknown`; `Stopped` becomes exclusively an operator-set state (`stop.go:293`, `kill.go:267`), so operator-stop and clean self-exit are distinguishable without a new persisted field.
- [ ] Auto-delete / wind-down wiring moves with the rename: `Error` inherits today's `Failed` behavior (sticky, preserved with exit code, excluded from auto-delete — `cellStateIsSticky` `refresh.go:866` and the auto-delete trigger set); `Exited` inherits today's `Stopped` behavior (auto-delete-eligible, so `kuke run --rm` reaps a cleanly-finished job).
- [ ] The `WorkloadFailed` breadcrumb (`Reason`/`Message`) is stamped on the `Error` transition (as it is today on `Failed`).
- [ ] Tests: `internal/controller/runner/refresh_test.go:350-470` updated (`ExitCode: 127 → Error`, `0 → Exited`); add cases asserting operator-stop → `Stopped` stays distinct from self-exit.
- [ ] `make dev-init` smoke run clean (this touches reconciler state the daemon reads).
- [ ] `docs/cli-use-cases.md` state-name invariants and any pinned smoke output updated.

## Out of scope

- Restart/re-run behavior of terminal cells — tracked in #1268 (depends on the state names introduced here).
- Surfacing `ExitCode` in the default `kuke get cells` table (optional polish; can be a follow-up).

## References

- #1206 introduced the current binary `Failed`/`Stopped` exit-code split this issue refines.
- Kubernetes Pod phase (`Succeeded`/`Failed`) and container `terminated.reason` (`Completed`/`Error`) are the model.

eminwux:
## Summary

`kuke get cells` collapsed four semantically distinct terminal outcomes into two labels (#1267): `Stopped` meant both "operator stopped it" and "it finished cleanly", and `Failed` meant both "a workload crashed" and "kukeon's bring-up failed." This splits each conflated terminal in two, at both the cell and container level, keeping Kubernetes as the reference (`Succeeded`/`Failed` + container `Completed`/`Error`).

| Trigger | Cell state |
| --- | --- |
| operator `kuke stop` / `kill` | `Stopped` |
| all workloads exit 0 | **`Exited`** (new) |
| a workload exits non-zero | **`Error`** (new) |
| kukeon bring-up / runtime fault | `Failed` (now reserved for kukeon's fault) |

Per the issue's naming recommendation (`Exited` + `Error`), and following the body's recommendation since no author comment overrode it.

## What changed

- **Enums (append-only, lockstep ordinals)** — `CellStateExited`/`CellStateError` and `ContainerStateExited`/`ContainerStateError` appended last in both `internal/modelhub` and `pkg/api/model/v1beta1` so `internal/apischeme/scheme.go`'s direct int-cast conversion stays correct. New `StateExitedStr`/`StateErrorStr` printable strings, `String()` cases, and `state_marshal.go` parse cases + widened int bounds (round-trip back-compat).
- **Container split** — `ConvertContainerdStatusToContainerState` (`internal/ctr/task.go`) maps a stopped task to `Exited` (exit 0) or `Error` (exit > 0); `ContainerStateFailed` stays reserved for kukeon container bring-up faults.
- **Reconciler derivation** (`refresh.go`) — `deriveCellStateFromNonRootContainerStatuses` and `deriveCellStateFromRootContainer` now produce the three-way `Error` / `Exited` / `Ready` split and **never** yield `Stopped`. `containerExitedNonZero` recognizes `ContainerStateError`.
- **Auto-delete / sticky wiring** — `cellStateAutoDeleteTriggers` → `{Exited}` (so `kuke run --rm` reaps a cleanly-finished job); `cellStateIsSticky` → `{Failed, Error}` (a crash is preserved with its exit code, like the old `Failed`). The `WorkloadFailed` breadcrumb is stamped on the `Error` transition (RefreshCell + ReconcileCell).
- **CLI consumers** — `run`/`restart`/`cell_liveness` treat `Exited` like `Stopped` (restartable clean terminal) and `Error` like `Failed` (refuse / degraded); `apply` `cellNeedsRematerialize` includes `Exited`; `failedCellReconcileError` treats `Error` like `Failed` (apply exits non-zero over a crashed cell, preserving pre-#1267 behavior); `get container` renders the new labels.
- **Docs** — `docs/cli-use-cases.md` invariants, the `kuke-refresh`/`kuke-restart` CLI docs, and the `cell`/`container` concept + manifest state tables updated.
- **Tests** — derivation/predicate tests updated (`ExitCode 127 → Error`, `0 → Exited`); `TestDeriveCellState_SelfExitDerivesExitedNeverStopped` and an in-loop assertion that the derivation never yields `Stopped`; marshal round-trip covers the new states.

## Review feedback addressed (fix-up `4684f08`)

The 2nd reviewer correctly observed that the **operator-stop half of conflation #2 is not durable**: `CellStateStopped` is non-sticky (`cellStateIsSticky`, refresh.go:912), so ~one reconcile interval after `kuke stop`, `ReconcileCell` re-derives the still-terminal cell to `Exited`/`Error` and the operator-stop label is gone. The test `TestDeriveCellState_SelfExit…` asserted the distinction "from the persisted state alone," but only exercised the derivation function — never the `ReconcileCell` path that erases it. Both observations are confirmed (the code's own `cellStateIsSticky` comment already documents the Stopped→Exited re-derivation; the test never ran a cell through `ReconcileCell`).

Per the reviewer's **descope resolution** (deliver the Exited/Error split now; move durable operator-stop preservation to #1268), this fix-up corrects the over-claims rather than the design — the non-sticky design is correct for this PR's scope and durable preservation is genuinely #1268's territory (its restart guards are what need a persisted terminal state):

- **Test** (`refresh_test.go`) — renamed `…SelfExitDistinctFromOperatorStop` → `TestDeriveCellState_SelfExitDerivesExitedNeverStopped`; its doc comment + the in-loop assertion comment now state the invariant they actually exercise (the *derivation* maps clean→`Exited`, crash→`Error`, and never yields `Stopped`) and explicitly note this does **not** make a persisted `Stopped` durable, deferring durable preservation to #1268. The assertions themselves are unchanged and still pass.
- **Docs** (`cell.md`) — the `Stopped` row no longer reads as a durability guarantee; it now says the label is transient (non-sticky, re-derived to `Exited`/`Error` next tick) with durable preservation tracked in #1268.
- **Code comment** (`refresh.go:1230`) — same correction at the derivation site.

**Premise-rot note on #1267 checkbox 3 (for PM/reviewer).** Checkbox 3's parenthetical — "operator-stop and clean self-exit are distinguishable without a new persisted field" — was optimistic: it holds at the derivation level but not across reconcile, exactly because `Stopped` is non-sticky. #1267's own *Out of scope* already assigns "restart/re-run behavior of terminal cells" to #1268, and durable operator-`Stopped` persistence is that machinery. So this PR delivers #1267's state-name split + derivation distinction in full; the *durable persistence* sub-claim moves to #1268. `Closes #1267` is kept on that reading — if you'd rather hold #1267 open for the durability piece, drop the `Closes` line at merge.

## Non-obvious calls (please review)

- **`Stopped` is kept non-sticky** (only `Error` joins `Failed` in `cellStateIsSticky`, per the AC's checkbox-4 wiring). The derivation-level operator-stop distinction is delivered by the derivation **never producing** `Stopped` — a persisted `Stopped` originates only from `stop.go`/`kill.go`. Making `Stopped` sticky to durably preserve it across reconcile ticks breaks the `--rm` / wind-down reconciler-reap, which reaches `Stopped` via `KillCell` and relies on the next tick re-deriving the now-terminal cell to `Exited` (run.go:703-707) — so durable preservation is deferred to **#1268** (see the descope note above).
- **Container-level split is scoped to `ConvertContainerdStatusToContainerState`** (the file the AC names). The separate "task vanished, exit unknown" sentinel (`container_state.go` → `Stopped`) is left as-is; the cell-level derivation keys on the *preserved* `ExitCode`, so cell states stay correct regardless of that per-container display label.
- `Exited`/`Error` were inserted into the exhaustive-checked `run`/`restart` switches deliberately (the project's `exhaustive` linter is a forward guard there).

## Test plan

- [x] `GOWORK=off go test ./...` — full root module suite green (re-run after fix-up `4684f08`)
- [x] Updated reconciler/derivation/predicate/marshal unit tests pass (`internal/controller/runner`, `internal/controller/apply`, `internal/ctr`, `pkg/api/model/v1beta1`)
- [x] `make dev-init` smoke (default profile) — **EXIT 0**, attach smoke clean, and the daemon-parity regression guard passed (both `kuke get realms` and `--no-daemon` identical: `default Ready` / `kuke-system Ready`). Not re-run for fix-up `4684f08`: that commit changes only a test function name, comment text, and a docs table cell — zero executable reconciler code paths — so the prior run remains representative. Note: the canonical `KUKEON_PROFILE=dev` run is currently blocked in this environment by an unrelated `kuke build --realm kuke-system` namespace-resolution quirk (writes the image to `kuke-system.kukeon.io` instead of the dev-suffixed `kuke-system.dev.kukeon.io`, so dev-profile `kuke init` can't find it) — outside this diff, which touches no build/namespace/init code. The default-profile run exercises the same reconciler-state path the daemon reads.

Closes #1267


eminwux:
**PR #1269 Review — `fix: split cell/container terminal states into Exited and Error`**

Correct and unusually thorough. The Exited/Error split is delivered cleanly: enum ordinals stay in lockstep between `internal/modelhub` and `pkg/api/model/v1beta1` (verified the full const blocks, not just the appended tail), so `apischeme/scheme.go`'s direct int-cast conversion remains sound; every container-state switch (`containerStateIsTerminal`, `rootContainerStillRunning`, both derivation paths, `containerStateToString`) and cell-state switch (`run`/`restart`/`cell_liveness`) handles the new states; marshal bounds widened to `…Error` with round-trip coverage. The headline invariant — the reconciler never derives `Stopped` — is pinned by the new `TestDeriveCellState_SelfExitDistinctFromOperatorStop` and the in-loop assertion. No secret/PII leaks; new error strings are clean. Cell rendering goes through `State.String()`, which was updated, so `kuke get cells` picks up the new labels.

One non-blocking note:

- **Operator-stop durability** `docs/site/concepts/cell.md:159`: the `Stopped` row says it "stays distinct from a clean self-exit," but `Stopped` is non-sticky and exit codes are preserved across reaping (`helpers.go:790`), so the next reconcile tick re-derives an operator-stopped cell — `kuke kill` (exit 137) settles it to `Error` (sticky), a clean `kuke stop` to `Exited`. Your own `windDownCell` comment (`refresh.go:846`) documents this Stopped→Exited re-derivation. This is **pre-existing** (old code landed kill→`Failed`-sticky the same way) and you've explicitly deferred durable operator-`Stopped` to #1268, so it's not a regression — just flagging that the cell.md line reads as a durability guarantee operators won't actually see. Worth softening the wording in a follow-up.

Test plan is solid — full unit suite green, default-profile `make dev-init` EXIT 0 with the daemon-parity guard passing and the attach smoke clean. The `KUKEON_PROFILE=dev` run is blocked by an unrelated build/namespace quirk outside this diff (documented), which is the host-constraint carve-out.

LGTM.

eminwux:
**PR #1269 Review (2nd reviewer) — `fix: split cell/container terminal states into Exited and Error`**

The enum split itself is clean: appended-last ordinals stay lockstep with scheme.go's direct int-cast, marshal bounds/round-trip and `String()` cases are complete, no exhaustive switch or state-range iteration was missed, all CI green (incl. `make e2e`), commits signed. The **Exited/Error** half (conflation #1 — the old `Failed` overload) is delivered durably. But the **operator-stop** half (conflation #2 — the issue's headline) doesn't hold in steady state:

- **Stated-invariant violation** `internal/controller/runner/refresh.go:912`: `Stopped` is non-sticky, so `ReconcileCell` re-derives and overwrites it (`refresh.go:692,774,777`), and the derivation never yields `Stopped` (`refresh.go:1236,1303`). The reconcile loop processes every cell with no state filter (`internal/controller/reconcile.go:88`), and `kuke stop` leaves the container record in place (#867) so the next tick reads it back. Net: ~one reconcile interval after `kuke stop`, the cell flips `Stopped → Exited`, indistinguishable from a clean self-exit. Conflation #2 is relabeled, not resolved — yet the PR `Closes #1267`.
- **Test asserts an invariant it doesn't exercise** `internal/controller/runner/refresh_test.go:581`: `TestDeriveCellState_SelfExitDistinctFromOperatorStop`'s comment claims the two causes are "distinguishable from the persisted state alone," but it only calls the derivation directly — it never runs an operator-stopped cell through `ReconcileCell`, which is the exact path that erases the distinction. A future #1268 author keying restart guards on a persisted `Stopped` would build on a label that's gone within one tick.

The author transparently flagged the non-durability and pointed at #1268 — fair — but the PR still closes #1267 (whose AC checkbox 3 requires the operator-stop distinction) and ships a test asserting it. Resolve by either delivering durable operator-stop preservation, or descoping #1267 to the Exited/Error split (moving conflation #2 to #1268) and fixing the test name/comment so it doesn't claim the unheld invariant.

Changes requested on the two items above; the rest of the diff is solid.

eminwux:
Addressed in 4684f08 (descope resolution — fix the over-claims, keep the non-sticky design):

- **Test name/comment** — verified both reviewer claims first: `cellStateIsSticky` (refresh.go:912) deliberately excludes `Stopped`, and the code's own comment documents the Stopped→Exited re-derivation, so a persisted `Stopped` is indeed re-derived to `Exited`/`Error` on the next `ReconcileCell` tick; the old test exercised only the derivation function, never `ReconcileCell`. Renamed `…SelfExitDistinctFromOperatorStop` → `TestDeriveCellState_SelfExitDerivesExitedNeverStopped` and rewrote its doc comment + the in-loop assertion comment to state the invariant actually exercised (derivation maps clean→`Exited`, crash→`Error`, never yields `Stopped`) with an explicit note that this does **not** make a persisted `Stopped` durable — durable operator-stop preservation deferred to #1268. Assertions unchanged, still green.
- **Docs** (`cell.md`) — `Stopped` row no longer reads as a durability guarantee; now states the label is transient (non-sticky, re-derived next tick) with durable preservation tracked in #1268. (Also covers the 1st reviewer's non-blocking note on the same line.)
- **Code comment** (`refresh.go:1230`) — same correction at the derivation site.

Premise-rot note on #1267 checkbox 3 in the PR body for PM/reviewer: checkbox 3's "without a new persisted field" parenthetical holds at the derivation level but not across reconcile (Stopped is non-sticky). #1267's own *Out of scope* already assigns terminal-cell restart/re-run to #1268, which is where durable persistence belongs. `Closes #1267` kept on that reading — drop it at merge if you'd rather hold #1267 open for the durability piece.

Tests: `GOWORK=off go build/vet/test ./...` all green after the fix-up. `make dev-init` not re-run — the fix-up changes only a test function name, comment text, and a docs cell (zero executable reconciler paths), so the PR's prior default-profile EXIT-0 run stays representative.

eminwux:
**PR #1269 Review — `fix: split cell/container terminal states into Exited and Error`**

Clean, well-scoped split. Verified: the internal↔v1beta1 enum ordinals stay in lockstep (both append `Exited` then `Error` last), so `scheme.go`'s direct int-cast conversion is unaffected; every `State` switch in the tree (`run`, `restart`, `cell_liveness`, `get container`, `refresh`, apply `reconcile`) is updated; the sticky / auto-delete / wind-down wiring is internally consistent (`Exited`→trigger, `Error`/`Failed`→sticky); back-compat holds (existing ordinals unchanged, string round-trip preserved). The headline invariant — the reconciler derivation never yields `Stopped` — holds in code and is pinned by the new test + in-loop assertion. Both commits verified, no dead code, no leaks. The fix-up `4684f08` honestly corrects the prior over-claims and defers durable operator-stop preservation to #1268.

One non-blocking item:

- **Stale comment** `internal/controller/runner/refresh.go:1151,1156-1157`: the prose block of `deriveCellStateFromNonRootContainerStatuses`'s doc still says the derivation can "flip the cell to Stopped" / counts toward "all stopped", contradicting the structured list just above it (updated to "the derivation never yields Stopped") and the new test. Code is correct; only the comment text is stale — refresh to `Exited`/`Error` when convenient.

LGTM — ship it; the nit is cosmetic.

PR #398, opened 2026-05-11
test: mirror env-precedence test from kukeond into kuke init

eminwux:
## Summary

- `cmd/kukeond/kukeond_test.go`'s `TestApplyServerConfigurationEnvOverridesConfig` locks the `--flag > env > ServerConfiguration > default` rule on the kukeond side, but the sibling `applyServerConfiguration` in `cmd/kuke/init/init.go` had no equivalent test — a regression that re-broke the env branch on the init path would slip through.
- Mirror the test into `cmd/kuke/init/init_test.go`, exercising the two init-specific knobs the issue calls out (`KUKE_INIT_KUKEOND_IMAGE`, `KUKEON_RUN_PATH`) plus a third (LogLevel) without an env override to lock the per-field semantics (the env check is per-field, not all-or-nothing).
- The init path differs from kukeond in one detail: `NewKukeondCmd` calls `bindEnvVars()` itself, while `NewInitCmd` defers env binding to `cmd/kuke/kuke.go`'s `loadConfig()`, which only wires the shared keys. The init-specific keys (`KUKE_INIT_KUKEOND_IMAGE`) therefore need explicit `BindEnv()` calls in the test setup so `viper.Get` reflects the env value once `t.Setenv` fires. Comment explains.
- Added a `//go:linkname applyServerConfiguration …` plus an exported `ApplyServerConfiguration` wrapper, following the existing pattern in this file (`runInit`, `resolveKukeondImage`, etc.).

## Test plan

- [x] `go test ./cmd/kuke/init/... -run TestApplyServerConfigurationEnvOverridesConfig -v` (PASS)
- [x] `go test ./cmd/kuke/init/...` (full package — all 13 tests PASS, no regressions in the existing helpers)
- [x] `make test` (full CI test target — clean)
- [x] `pre-commit run --files cmd/kuke/init/init_test.go` (trim/eof/fmt/go-mod-tidy/golangci-lint/addlicense — all pass)
- [ ] `make dev-init` smoke test — change is purely an additional unit test in `cmd/kuke/init/init_test.go`; touches no build path, no daemon code, nothing the daemon reads, and nothing under `/opt/kukeon`, so the daemon-parity regression guard does not apply.

Closes #186

PR #1275, opened 2026-06-13
fix: kuketty surfaces workload exit code so clean attachable exits report Exited

### LINKED ISSUE #1273: kuketty: clean workload exit 0 reported as Error (exit 70) — Exited unreachable for attachable cells
## Observed

A workload in an **attachable** (kuketty-wrapped) cell that exits **cleanly (exit 0)** in the terminal lands the cell/container in `Error`, not `Exited` — defeating the #1267/#1269 clean-exit-vs-crash split for the primary interactive cell type.

```
$ kuke attach mycell      # ... do work, then `exit` (shell exits 0)
$ kuke get cell mycell    # STATE: Error   (expected: Exited)
```

The container/cell state is derived from **containerd's task exit code**, but for an attachable cell the task's PID 1 is **kuketty**, not the workload — and kuketty exits `70` even when the workload exited `0`.

## Repro

- Branch/ref: `main` @ d4050c4 (post-#1269).
- Create/run any attachable cell, attach, let the workload exit 0 (`exit` at the shell, or a command that returns 0). Observe `kuke get cell`/`kuke get container` reports `Error` instead of `Exited`.
- (Diagnosed by code trace; daemon was not live on the dev host this session, but the path is deterministic.)

## Root cause

The state split (#1269) keys entirely on the task exit code:

- `internal/ctr/task.go:81-85` — `ConvertContainerdStatusToContainerState`: stopped task with `ExitStatus != 0` → `ContainerStateError`, else `ContainerStateExited`.
- `internal/controller/runner/refresh.go` — `deriveCellStateFromRootContainer` (~:1294) / `deriveCellStateFromNonRootContainerStatuses`: any non-zero terminal container ⇒ `CellStateError`; all-clean ⇒ `CellStateExited`.

But kuketty never propagates the workload child's exit code as its own, and treats the *normal* "child exited" event as an internal failure:

1. The workload child exits cleanly. sbsh's terminal runner **always** publishes a non-nil `EvCmdExited.Err` — even on a clean exit the value is `errors.New("shell process exited (code 0)")` (or `"shell process exited: code=0"` in init mode): `sbsh@v0.13.0/internal/terminal/terminalrunner/lifecycle.go:160-192`.
2. sbsh's `runLoop` returns that error as the terminating cause: `sbsh@v0.13.0/pkg/terminal/server/server.go` (`EvCmdExited` → `return ev.Err`).
3. kuketty's `isCleanShutdown` only treats `nil` and `context.Canceled` as clean (`cmd/kuketty/main.go:351-359`). `"shell process exited (code 0)"` is neither → `run()` returns the error → `main()` calls `os.Exit(exitCodeInternal)` = **70** (`cmd/kuketty/main.go:79-82, 175-178`).
4. containerd records the task `ExitStatus = 70` → `ContainerStateError` → `CellStateError`.

`cmd/kuketty/main.go:65-70` even documents the deliberate "workload exit is not surfaced through kuketty" stance — but the fall-through then maps the *expected* clean-exit path to an internal-error code. Result: for attachable cells, `CellStateExited` is unreachable; every clean exit is reported as a crash.

## Suggested fix

In kuketty, distinguish a normal workload exit from an internal failure: recognize sbsh's `EvCmdExited` terminating cause (alongside `nil`/`context.Canceled`) as a non-error termination, and ideally parse the workload's real exit code out of the `EvCmdExited` payload and exit with it (so a genuine non-zero workload exit still yields `Error`, while exit 0 yields `Exited`). At minimum, exit `0` when the workload exited `0`. This requires sbsh to surface the child's exit code in a machine-readable form on `EvCmdExited` (today it is embedded in the error string); coordinate with the sbsh dependency if a structured field is needed.

## References

- Surfaced alongside **#1268** (`cell: a terminal cell (Failed/Error/Exited) cannot be restarted or re-run`), which is CLOSED as *completed* but whose core AC ("`Error` … moved out of the rejection bucket") is **not** met by the merged #1269 code — `cmd/kuke/run/run.go:672` still refuses `Error` with the delete-then-rerun pointer. This bug makes that gap acute: every clean attachable exit becomes an un-recoverable `Error`. Recommend the maintainer reopen/verify #1268.
- State split: #1267 / PR #1269 (commit d4050c4).

eminwux:
## Summary

- An attachable (kuketty-wrapped) cell whose workload exited **cleanly (exit 0)** landed in `Error` (exit 70) instead of `Exited`, defeating the #1267/#1269 clean-exit-vs-crash split for the primary interactive cell type (#1273).
- Root cause: kuketty's PID 1 is the *wrapper*, not the workload. sbsh's `EvCmdExited` always carries a non-nil terminating cause — even on a clean exit (`"shell process exited (code 0)"` / `"shell process exited: code=0"`). kuketty's `isCleanShutdown` only treated `nil`/`context.Canceled` as clean, so every workload exit fell through to `exitCodeInternal` (70). containerd then recorded `ExitStatus=70` → `ContainerStateError` → `CellStateError`.
- Fix: kuketty now recognizes sbsh's "shell process exited" terminating cause as a workload termination (not a wrapper failure) and **propagates the workload's real exit code as its own** — `0 → Exited`, non-zero → `Error` with the workload's status. Genuine kuketty-internal failures still map to `exitCodeInternal` (70).

## Implementation notes

- **No structured exit field in the pinned dependency.** sbsh v0.13.0 (latest available — `go list -m -versions` confirms no newer tag) has no machine-readable exit-code field on `EvCmdExited`; the code is embedded in the terminating error, exactly as the issue's "Suggested fix" anticipated ("today it is embedded in the error string"). `workloadExitCode` therefore recognizes all three carriers sbsh emits (`sbsh@v0.13.0/internal/terminal/terminalrunner/lifecycle.go:160-192`):
  - non-init `cmd.Wait` path → wrapped `*exec.ExitError` (recovered via `errors.As`);
  - non-init clean exit → `"shell process exited (code 0)"` literal;
  - PID-1 reaper (init mode — the path in a real container, since kuketty is PID 1) → `"shell process exited: code=N"`.
- **No sbsh change required.** The issue noted a structured field "if needed"; since the code is recoverable from the existing v0.13.0 error carriers, no dependency bump or upstream coordination is needed. If sbsh later exposes a structured field, the string parsing can be dropped.
- **Abnormal no-code teardown paths treated clean.** The bare `"shell process exited"` (reaper channel closed) and `"shell process exited (close cancelled tracked-exit wait)"` (sbsh #398) cases carry no recoverable code; they arise during teardown/shutdown races, not a workload crash, so they map to exit 0 rather than spuriously flipping a stopped cell to `Error`. Documented inline.
- **Operator-stop path unchanged.** `kuke stop` cancels kuketty's ctx → sbsh's `runLoop` returns `context.Canceled` → `isCleanShutdown` → exit 0, as before.
- Updated the now-stale `exitCodeInternal` doc comment (the issue flagged `cmd/kuketty/main.go:65-70`'s "workload exit is not surfaced through kuketty" claim as contradicting the actual fall-through behavior).

## Test plan

- [x] `make test` (`make test-root` + `make test-kukebuild`) — full suite green, 107 packages ok, exit 0
- [x] `go test ./cmd/kuketty/ -run 'TestWorkloadExitCode|TestIsCleanShutdown' -v` — new table test covers all three carriers + init/non-init/abnormal/internal-failure cases; `TestWorkloadExitCode_ExecExitError` pins the real `*exec.ExitError` carrier
- [x] `GOWORK=off go vet ./... (minus e2e)` — clean
- [ ] `make dev-init` end-to-end smoke — not run: this change touches only the in-container `kuketty` binary's exit-code mapping (`cmd/kuketty/main.go`), not the build path, daemon bootstrap, or anything under `/opt/kukeon` the daemon-parity tail reads. The runtime behavior (containerd records kuketty's exit code → cell state) is a live-daemon observation; the deterministic code path is covered by the unit tests above, and the original bug was itself diagnosed by code trace (daemon not live on the dev host). A reviewer with a live daemon can confirm a clean `exit` in an attached cell now reports `Exited`.

Closes #1273

### DIFF
--- a/cmd/kuketty/main.go
+++ b/cmd/kuketty/main.go
@@ -39,8 +39,12 @@ import (
 	"log/slog"
 	"net"
 	"os"
+	"os/exec"
 	"os/signal"
+	"regexp"
 	"runtime"
+	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1275 Review — `fix: kuketty surfaces workload exit code so clean attachable exits report Exited`**

LGTM — the fix correctly propagates the workload's exit code as kuketty's own so a clean attachable exit (`code=0`) reports `Exited` instead of `Error`. I verified all five `EvCmdExited` carriers in the PR's `workloadExitCode` parsing against the pinned `sbsh@v0.13.0/internal/terminal/terminalrunner/lifecycle.go` (lines 160/162/171/190/192) — they match exactly, including the init-mode `code=N` reaper path that is the live in-container path (kuketty is PID 1). The `run()` reordering (clean-shutdown → workload-exit → internal-failure) preserves the operator-stop (`context.Canceled` → exit 0) and internal-failure (exit 70) paths, and `errors.As(*exec.ExitError)` can't false-positive since kuketty execs no subprocess but the sbsh-managed workload. Single signed commit, no dead code, tests cover every carrier.

Non-blocking — deferred host-gated verification (carve-out, not a blocker): the live-daemon leg (containerd records kuketty's exit → #1267/#1269 cell-state derivation) needs a running daemon the sandbox can't provide. Worth a manual confirm pre-merge: attach to a kuketty-wrapped cell, `exit` cleanly, and confirm `kuke get cells` shows `Exited`, not `Error`.

PR #1272, opened 2026-06-12
fix: make terminal cells restartable and align run/restart/start startable states

eminwux:
## Summary

- A cell whose workload exited (`Error`) or whose kukeon bring-up faulted (`Failed`) could not be restarted or re-run — the only documented recovery was delete + recreate, even though such cells are simply terminal. This aligns the three restart-ish verbs and makes terminal cells restartable (#1268, depends on #1267's `Exited`/`Error` state split which is now merged).
- **The startable-state decision now lives in a single funnel** — the daemon's `controller.StartCell` (`internal/controller/start_cell.go`), which `kuke start`, `kuke run <cell>`, and `kuke restart` all reach. It:
  - refuses only the genuinely-unrecoverable / mid-transition states `Pending` / `Unknown` with the `kuke delete cell <name>` pointer;
  - recovers `Failed` (possibly half-created container records) via `RecreateCell` (stop → recreate containers → start);
  - starts `Stopped` / `Exited` / `Error` (and a stale-`Ready` cell whose task is gone) from their intact records.
- CLI guards in `cmd/kuke/run/run.go` and `cmd/kuke/restart/restart.go` now route `Error`/`Failed` to the start path (no stop-first, so the daemon observes the persisted `Failed` and picks the recreate route) and refuse only `Pending`/`Unknown`.
- `markCellReady` (`internal/controller/runner/refresh.go`) clears the failure breadcrumb (`Status.Reason`/`Status.Message`) on the Ready transition — every runner start/recreate path funnels through it, so it's the single clear point.
- `cellStateIsSticky` is documented as a *derivation* contract (stops the reconciler re-deriving), not a restart gate: an operator restart drives the cell to Ready under the same per-cell lock, so a sticky `Error`/`Failed` cell stays operator-restartable.
- `docs/cli-use-cases.md` run/restart/start invariants updated to the new startable-state contract.

### Notable implementation calls (reviewer please push back)
[...]

PR #1276, opened 2026-06-13
fix: recover Error cells via recreate so kuke run no longer refuses a crashed cell

### LINKED ISSUE #1274: run/restart: Error cell cannot be recovered — kuke run refuses instead of restarting (re-opens #1268's unmet AC)
## Observed

A cell in a terminal `Error` state (a workload crashed, or — per #1273 — a kuketty workload exited cleanly and was mis-reported as a crash) **cannot be recovered with `kuke run`**. The CLI refuses instead of doing its best to bring the cell back up:

```
$ kuke run agents-dev-0
Error: cell "agents-dev-0" exists in Error state; delete it with `kuke delete cell agents-dev-0` before re-running
```

And when a terminal-by-derivation cell still has a leftover-running **root** container (the `Error` state is sticky and is *not* wound down), the start path refuses on a different message:

```
$ kuke run agents-dev-0
Error: cell "agents-dev-0" has running containers and must first be stopped.
```

Expected: `kuke run` should always do its best to run the cell — recover an `Error` cell by stopping/killing any leftover containers and restarting, rather than forcing delete + recreate.

## Repro

- Branch/ref: `main` @ d4050c4 (post-#1269).
- Get a cell into `Error` (e.g. a workload exits non-zero, or — via #1273 — an attachable workload exits 0 and is mis-reported as `Error`). The root container may still be running because `Error` is sticky and not wound down.
- `kuke run <cell>` → refused (`delete it ... before re-running`), or `has running containers and must first be stopped` if the start path is reached with a live root.
- (Diagnosed by code trace; daemon was not live on the dev host this session, but the paths are deterministic.)

## Root cause

Two distinct refusals, both reachable for an `Error` cell:

1. **`kuke run`/`kuke restart` bucket `Error` with the unrecoverable states.**
   - `cmd/kuke/run/run.go:672-682` — `case CellStatePending, CellStateFailed, CellStateError, CellStateUnknown:` returns the delete-then-rerun error.
   - `cmd/kuke/restart/restart.go:229` — identical bucketing for restart.
   - `cmd/kuke/shared/cell_liveness.go` `GuardCellTaskLiveness` — same family.
   So by design `kuke run` refuses an `Error` cell instead of recovering it.

2. **Root-vs-derivation asymmetry trips the `StartCell` liveness guard.** The cell state is derived from **non-root** containers only (`deriveCellStateFromNonRootContainerStatuses` in `internal/controller/runner/refresh.go` ignores the root), and `Error` is sticky and explicitly **not** wound down (`cellStateIsSticky` / `shouldWindDownCell` — only `Exited` winds down). So a crashed cell keeps its root container alive. Meanwhile `StartCell` checks **all** containers including the root: `internal/controller/start_cell.go:61-75` returns `cell %q has running containers and must first be stopped` whenever any container is `ContainerStateReady`. `kuke run` never stops the leftover root before restarting.

This is the same defect #1268 tracked (`cell: a terminal cell (Failed/Error/Exited) cannot be restarted or re-run`), but #1268 was **closed as completed** while the core AC ("`Error` … moved out of the rejection bucket") was not met by the merged #1269 code — `run.go:672` still refuses `Error`. Filing fresh per maintainer request rather than reopening.

## Suggested fix

Make `kuke run`/`kuke restart` treat `Error` (and a terminal-by-derivation cell with a leftover-running root) as recoverable: stop/kill any live containers (including the root), then restart — routing through the existing `RecreateCell` recovery (`internal/controller/runner/recreate_cell.go`, today reachable only via the OutOfSync-reapply path at `start_cell.go:225`). Align the startable-state set across `kuke start`, `kuke run <cell>`, and `kuke restart` so the three verbs agree, and clear the failure breadcrumb (`Status.Reason`/`Message`) on the Ready transition. This is the body of #1268's unmet acceptance criteria.

## Out of scope

- The kuketty exit-code root cause that *creates* spurious `Error` cells from clean exits — tracked separately in **#1273**. This issue is about recovering an `Error` cell regardless of how it got there.

## References

- **#1268** — prior tracking issue, closed-as-completed but core AC unmet (`run.go:672` still refuses `Error`); contains the full three-verb analysis and acceptance criteria.
- **#1273** — kuketty clean-exit-0 mis-reported as `Error` (the most common way a cell lands here).
- State split: #1267 / PR #1269 (commit d4050c4).

eminwux:
## Summary

`kuke run`/`kuke restart`/`kuke start` could not recover a cell in the terminal `Error` state whose root container was still alive. The daemon-side `controller.StartCell` refused with `cell "<name>" has running containers and must first be stopped` — forcing a delete + recreate instead of doing its best to bring the cell back up. This is the unmet half of #1268's acceptance criteria, re-filed as #1274.

**Root cause (part 2 of the issue).** A sticky `Error` cell keeps its root container running: cell state is derived from non-root containers only (`deriveCellStateFromNonRootContainerStatuses` ignores the root), and `Error` is not wound down (`shouldWindDownCell` fires only for `Exited`). So the leftover root reads back `ContainerStateReady`. `StartCell`'s `hasRunningContainer` guard (`internal/controller/start_cell.go`) ran **before** the recovery switch and tripped on that live root, so `Error` never reached recovery.

**Fix.** Move the recovery-routing switch *before* the running-container guard and route `Error` through `RecreateCell` (alongside `Failed`). `RecreateCell` stops/deletes the leftover containers including the root, rebuilds, and starts — the same recovery the `OutOfSync` breaking-diff reapply already uses. Its start phase funnels through `markCellReady`, which already clears the failure breadcrumb (`Status.Reason`/`Message`) on the Ready transition. The guard now only sees `Ready`/`Stopped`/`Exited`, whose records are intact and not running.

## Premise rot (reviewer please confirm the reinterpretation)

The issue was filed against `main @ d4050c4`, **before** #1272 merged. #1272 already addressed **part 1** of the issue — it moved `Error` out of the CLI rejection bucket in `cmd/kuke/run/run.go` and `cmd/kuke/restart/restart.go` (both now route `Error` → `client.StartCell`). So the three CLI verbs are already aligned; this PR fixes only the still-broken **daemon-side** recovery (part 2).

This **reverses one decision #1272 just merged**: #1272 added a test asserting `Error` must route through *plain* `StartCell`, **not** `RecreateCell` (`start_cell_test.go`, "cell in Error state can be started"). That routing provably cannot recover a cell whose root is still live — it lands on the `hasRunningContainer` guard. The issue's explicit suggested fix is to route `Error` through `RecreateCell`, so I updated that test to assert recreate routing (mirroring the `Failed` test) and added a regression test for the leftover-running-root scenario.

## Non-obvious calls

- **`Error` now recreates rather than plain-starts**, so the container overlay (writable layer) is wiped on recovery — same as `Failed`. The crash diagnostic surface (spec, captured logs, exit code) is preserved by stickiness *until* the operator chooses to restart; volume-backed memory survives once #1015 lands. This matches the issue's suggested fix verbatim.
- **`Error` no longer runs the `OutOfSync`-reapply path** (it returns from the recreate branch before reaching it) — identical to `Failed`'s existing behavior. The cell is recovered from its on-disk spec; the next reconcile re-flags `OutOfSync` if it still diverges from a lineage Config.
- **Attach-guard alignment (`cmd/kuke/shared/cell_liveness.go`).** `GuardCellTaskLiveness` (the `kuke attach` guard the issue names as "same family") routed `Error` to a delete-then-rerun pointer via its `default` branch. With `Error` now restartable, that was contradictory advice, so I moved `Error` into the `kuke start` pointer branch alongside `Stopped`/`Exited`. This only affects literal `kuke attach <error-cell>` (run's `Error` branch never calls this guard). Flagging as a small scope extension beyond the three verbs the AC names — drop it if the reviewer prefers a tighter diff.

## Out of scope

- The kuketty exit-code root cause that *creates* spurious `Error` cells from clean exits — tracked in #1273 (and partly addressed by #1275 on current `main`).

## Test plan

- [x] `GOWORK=off go test ./...` (full root suite minus e2e) — green (`TEST-EXIT=0`)
- [x] `cd cmd/kukebuild && GOWORK=off go test ./...` — green
- [x] `GOWORK=off go vet` on changed packages — clean
- [x] New unit tests: `Error` routes through `RecreateCell`; `Error` with a leftover-running root recovers (no guard trip); attach guard points `Error` at `kuke start`
- [ ] `make dev-init` end-to-end smoke — not run: this change is confined to the controller start/recovery path and does not touch the build path, daemon bootstrap, or anything under `/opt/kukeon` the daemon-parity tail reads; the daemon was not live on this dev host (matching the issue's own repro note). Reviewer may run it on a clean host to exercise a live `kuke run` recovery of an `Error` cell.

Note: `GOWORK=off` is required locally — the repo uses a `go.work` workspace whose module graph otherwise pulls an incompatible `containerd/cgroups` and fails to compile (per CLAUDE.md / `/diagnose-blocked-build`).

Closes #1274


eminwux:
**PR #1276 Review — `fix: recover Error cells via recreate so kuke run no longer refuses a crashed cell`**

LGTM — moving the recovery-routing switch in `StartCell` *before* the running-container guard correctly lets a sticky `Error` cell (live root reading `ContainerStateReady`) reach `RecreateCell` instead of tripping `has running containers`, and as a side benefit fixes the same live-root case for `Failed`. The `OutOfSync` reapply is correctly skipped for both (early return), and the attach-guard / `run` / `restart` doc + routing are all kept in agreement — verified no startability decision point was missed. The overlay-wipe-on-recovery tradeoff is documented and matches the issue's suggested fix.

One deferred item for you to run pre-merge (host-gated, non-blocking): the `make dev-init` end-to-end smoke — exercise a live `kuke run` recovery of an `Error` cell. The author couldn't run it (no live daemon on the dev host); unit coverage of the routing is thorough and CI is fully green.

Nothing blocking — ship it.

eminwux:
**PR #1276 Review — `fix: recover Error cells via recreate so kuke run no longer refuses a crashed cell`** (second reviewer)

Concur with the prior LGTM. Independently verified the load-bearing claims rather than re-reading the diff alone:

- `shouldWindDownCell` gates on `cellStateAutoDeleteTriggers`, which returns true **only** for `Exited` (`refresh.go:944`) — so a sticky `Error` cell's root genuinely stays alive and reads back `ContainerStateReady`. The bug premise holds, and moving the recovery switch ahead of the running-container guard (`start_cell.go:92`) is the correct fix.
- `RecreateCell` opens with `stopCellLocked(existing)` then deletes all containers (`recreate_cell.go:56`) — the leftover root is actually torn down, so recovery converges rather than dead-locking on the live root.
- `OutOfSync` reapply is correctly bypassed for `Error`/`Failed` via the early return; both states share one recreate path; attach-guard / `run` / `restart` docs + routing stay in agreement.
- Error-string change to `"failed to recover %s cell"` pins no test, and `CellState.String()` renders `Error`/`Failed` correctly. No unverified commits.

Non-blocking, already noted by the first reviewer: the host-gated `make dev-init` live-recovery smoke is unrun (no daemon on the dev host) — worth exercising pre-merge, but the routing has thorough unit coverage and CI is green.

LGTM — leaving `ready-to-merge`.

PR #1280, opened 2026-06-13
fix: re-stage kuketty binary on content mismatch

### LINKED ISSUE #1277: fix: staged kuketty binary is never re-staged on version change (stale-binary trap)
## Context

The daemon does not run kuketty from the kukeond image directly — it **stages** a host-visible copy and bind-mounts that into each attachable cell. Staging happens on **every** `kuke create cell`, in `stageKukettyBinary` (`internal/controller/runner/attachable.go:347`), which copies from the daemon's own `/bin/kuketty` (`kukettySourcePathInsideDaemon`, `attachable.go:65`) to `<RunPath>/bin/kuketty` (default `/opt/kukeon/bin/kuketty`).

The staging cache gate is `stagedBinaryUsable` (`attachable.go:376`), which only checks **existence + non-zero size + exec bit** — never version or content. So once `/opt/kukeon/bin/kuketty` exists, the daemon reuses it forever, even after the kukeond image is rebuilt with a newer kuketty.

No daemon lifecycle command clears the staged copy: `kuke daemon reset` removes only `/run/kukeon/kukeond.{sock,pid}`; `--purge-system` removes only `<RunPath>/data/kuke-system` (`cmd/kuke/daemon/reset/reset.go:185`). `<RunPath>/bin/` is untouched.

**Consequence:** every kuketty fix silently fails to reach an existing host on `kuke daemon recreate` (new image, no re-init). This is the propagation gap behind the #1275 regression resurfacing in practice — operators rebuild the image and still run the pre-fix kuketty until they manually `rm /opt/kukeon/bin/kuketty`.

## Proposal

Make `stageKukettyBinary` re-stage on **mismatch** instead of trusting existence. Preferred mechanism is a **content comparison** of source (`/bin/kuketty`) vs dest (`<RunPath>/bin/kuketty`) — the daemon already has both files in hand, so this needs no `kuketty version` subcommand, no exec, and no version wiring:

- Compare a cheap discriminator first (size), then a content hash (e.g. SHA-256) when sizes match; re-stage via the existing `copyBinaryAtomic` (`attachable.go:438`) tmp+rename path on any difference.
- Keep the existing zero-byte / non-exec defensive rejections in `stagedBinaryUsable`.

This fixes init, `daemon recreate`, and the dev loop in one place. (A version-stamp comparison via a `kuketty version` subcommand — filed separately — is a more legible alternative but needs version wiring + an exec per cell create; the content compare is strictly less machinery for pure cache-invalidation.)

## Acceptance criteria

- [ ] `stageKukettyBinary` re-stages when the staged `<RunPath>/bin/kuketty` differs from the daemon's source `/bin/kuketty` (content/hash mismatch), not only when absent.
- [ ] A rebuilt kukeond image (new kuketty) propagates to newly-created cells after `kuke daemon recreate` with **no** manual `rm` of the staged copy.
- [ ] Existing zero-byte / non-executable defensive re-stage paths in `stagedBinaryUsable` are preserved.
- [ ] Unit tests in `internal/controller/runner/attachable_test.go` cover: stale-content dest → re-staged; identical dest → reused (no copy); missing/zero-byte/non-exec dest → re-staged.

## Out of scope

- The `kuketty version` subcommand and version-string wiring (separate enhancement).
- Clearing `<RunPath>/bin/` on `daemon reset`/`--purge-system` — re-stage-on-mismatch makes that unnecessary.

## References

- `internal/controller/runner/attachable.go:347` `stageKukettyBinary`
- `internal/controller/runner/attachable.go:376` `stagedBinaryUsable` (existence-only gate — the defect)
- `internal/controller/runner/attachable.go:65` `kukettySourcePathInsideDaemon` (source), `:58` staged subdir, `:438` `copyBinaryAtomic`
- `cmd/kuke/daemon/reset/reset.go:185` (`--purge-system` does not touch `<RunPath>/bin/`)
- #1275 (kuketty exit-code fix whose propagation this gap blocks)

eminwux:
## Summary
- `stageKukettyBinary`'s cache gate (`stagedBinaryUsable`) only checked existence + non-zero size + exec bit, so once `<RunPath>/bin/kuketty` existed the daemon reused it forever — a rebuilt kukeond image's newer kuketty never reached newly-created cells on `kuke daemon recreate` until an operator manually `rm`'d the staged copy (#1277, the propagation gap behind #1275 resurfacing).
- Added a content comparison: `stageKukettyBinary` now resolves the source first and delegates to `ensureStagedBinary`, which reuses the staged copy only when it is usable **and** byte-identical to the source. On any mismatch it re-stages via the existing `copyBinaryAtomic` tmp+rename path.
- Comparison is a cheap size short-circuit (`sameFileContent`) before a streamed SHA-256, so the common up-to-date path stays off the multi-MiB hash on every attachable container start. No `kuketty version` subcommand, version wiring, or exec — the daemon already holds both files.
- The zero-byte / non-exec defensive rejections in `stagedBinaryUsable` are preserved (left untouched, since it is also used by `resolveKukettyBinary` to vet source candidates). A source-unresolvable fallback keeps the prior reuse behavior for the `--no-daemon`/dev path where the source binary has moved.

## Implementation notes
- `stagedBinaryUsable` is intentionally **not** changed — it is shared with `resolveKukettyBinary` (vetting source candidate paths), so the content check lives in the caller (`ensureStagedBinary`) per the AC's "keep the existing defensive rejections" requirement.
- Source resolution now runs before the cache check (previously after). In the daemon path the source is always `/bin/kuketty` and resolves; for the dev/`--no-daemon` path where it may not, a usable staged copy is reused rather than failing — preserving the prior existence-only reuse for that path.

## Test plan
- [x] `make test` (full project CI target) — exit 0
- [x] `GOWORK=off go build ./internal/controller/runner/` and `go vet` — clean
- [x] `go test ./internal/controller/runner/ -run 'TestStageKukettyBinary|TestEnsureStagedBinary|TestSameFileContent' -v` — all pass (new + existing reuse test)
- [x] `make dev-init` smoke (containerd started on host) — exit 0; both realms `Ready` in the daemon-parity tail (daemon + `--no-daemon` shapes match), `kukeond is ready`, and the `kuke attach` smoke against the kuketty-wrapped cell (the staging + socket pipeline this change touches) passed with clean detach and parent plumbing intact. Ran on the default profile (nested mode without `KUKEON_PROFILE=dev`); the two-realm parity contract held.

## Acceptance criteria
- [x] `stageKukettyBinary` re-stages when the staged dest differs from the source by content/hash, not only when absent.
- [x] A rebuilt kukeond image propagates to newly-created cells after `kuke daemon recreate` with no manual `rm` (content mismatch drives the re-stage; verified by unit tests + the dev-init staging path).
- [x] Existing zero-byte / non-executable defensive re-stage paths in `stagedBinaryUsable` are preserved.
- [x] Unit tests cover: stale-content dest → re-staged; identical dest → reused (no copy); missing/zero-byte/non-exec dest → re-staged (`TestEnsureStagedBinary_RestagesOnMismatch`, plus `TestSameFileContent` for the size/hash discriminator).

Closes #1277

### DIFF
--- a/internal/controller/runner/attachable.go
+++ b/internal/controller/runner/attachable.go
@@ -17,6 +17,8 @@
 package runner
 
 import (
+	"bytes"
+	"crypto/sha256"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -335,11 +337,14 @@ func kukettyMetadataLabels(spec intmodel.ContainerSpec) map[string]string {
 // daemon binary — see Dockerfile); for --no-daemon mode the function falls
 // back to a sibling of the running binary and then $PATH.
 //
-// The stage is idempotent: once the destination exists with non-zero size
-// and the executable bit, subsequent calls return its path without
-// re-copying. This is the hot path for every attachable container start,
-// so doing the lstat early avoids re-reading and re-writing a multi-MiB
-// binary per container.
+// The stage is idempotent on content: a subsequent call returns the
+// destination path without re-copying only when the staged copy is usable
+// (non-zero size, executable bit) AND byte-identical to the source. A
+// rebuilt kukeond image ships a newer kuketty, so a content mismatch re-
+// stages — the existence-only gate this replaced left an existing host
+// pinned to its first staged copy forever (#1277). The size short-circuit in
+// sameFileContent keeps the common up-to-date path off the multi-MiB hash on
+// every attachable container start.
 //
 // Atomicity is handled with a tmp-file rename so two concurrent provisions
 // never see a partial binary at the destination. The rename is on the same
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1280 Review — `fix: re-stage kuketty binary on content mismatch`**

Correct fix for #1277: `ensureStagedBinary` reuses the staged copy only when usable **and** byte-identical to the source, and re-stages on missing/zero-byte/non-exec/stale via the existing atomic tmp+rename — so a rebuilt kukeond image now propagates on `kuke daemon recreate` without a manual `rm`. Source-resolution-first reordering preserves the `--no-daemon` fallback, errors wrap with `%w`, helpers are private and used, and the tests cover every branch. No security surface. Commits verified.

One non-blocking nit:

- **Comment accuracy** `internal/controller/runner/attachable.go:345-347`: the doc claims the size short-circuit "keeps the common up-to-date path off the multi-MiB hash" — but up-to-date means byte-identical → equal size → it *does* hash both files on every attachable container start. The short-circuit only saves the re-stage (changed-size) case; the `sameFileContent` docstring states this correctly ("hashes only when sizes coincide"), so the two comments contradict. Behavior is fine (the hash cost per container start is acceptable); just reword the `stageKukettyBinary` claim.

PR #1283, opened 2026-06-13
fix: bump sbsh to v0.13.1 so clean kuketty workload exits report Exited

### LINKED ISSUE #1282: kuketty reports clean exit 0 as exit 70 (Error) — benign PTY EIO preempts EvCmdExited in sbsh Serve
## Summary

`kuketty` mis-reports a clean `exit 0` workload as exit code **70** (`EX_SOFTWARE`), so the daemon lands the cell in **Error / WorkloadFailed** instead of **Exited**. Reproduced on `agents-dev-0` after #1280. This is **not** a binary-staleness regression — #1280's re-stage fix is working correctly (the staged `/opt/kukeon/bin/kuketty` is byte-identical to the kukeond image source and contains the #1275 `workloadExitCode` logic). The remaining failure is a separate sbsh event-ordering race that #1275 didn't anticipate.

## Evidence (live `agents-dev-0`)

```
$ kuke get cell agents-dev-0 -o yaml
  state: Error
    - name: work
      state: Error
      exitCode: 70
  reason: WorkloadFailed
```

kuketty.log (current run, new binary, clean workload exit):

```
ERROR "pty stdout error" err=read /dev/ptmx: input/output error          <- emitted first
INFO  "closing terminal" reason=...pty read error: ...input/output error
INFO  "reaper: tracked child exited" pid=47 exit=0                        <- reaped after
INFO  "terminal close requested" err=shell process exited: code=0
```

## Root cause

When the workload exits, sbsh's `server.Serve` returns whichever terminating event reaches `runLoop` first, and treats `EvError` and `EvCmdExited` identically. The PTY-master EIO (`read /dev/ptmx: input/output error`) is the immediate kernel side-effect of the shell closing its tty, so it **reliably preempts** the `EvCmdExited` ("shell process exited: code=0") event. `Serve` therefore returns the EIO error as its cause.

`workloadExitCode` (`cmd/kuketty/main.go:102-132`) only decodes `*exec.ExitError` and messages containing `"shell process exited"`. The pty-EIO cause matches neither:

```
contains 'shell process exited' -> False  =>  ok=false
```

So `run()` falls through to `return fmt.Errorf("server.Serve: %w", serveErr)` → `main()` default case → `os.Exit(70)` (`cmd/kuketty/main.go:148-151`). The daemon maps any non-zero task exit to `Error` (`internal/ctr/task.go:70-97`), producing `Error / exitCode 70 / WorkloadFailed`.

Because the EIO `EvError` wins the race essentially every time for a clean PID-1/init-mode exit, this reproduces deterministically rather than flaking.

## Why #1275 missed it

#1275 correctly mapped the three clean `EvCmdExited` carriers to exit 0/N, but assumed the workload's fate always surfaces as the `EvCmdExited` cause. It did not anticipate the benign PTY-master EIO `EvError` preempting `EvCmdExited` inside sbsh's `runLoop`.

## Proposed fix

Teach `workloadExitCode` to treat a PTY-master read EIO (`ErrPipeRead` / `read /dev/ptmx: input/output error`) carried by sbsh's terminating cause as a **benign workload termination** (`return 0, true`) rather than a kuketty-internal failure.

**Tradeoff:** when the EIO wins the race, the workload's real exit code is not recoverable from sbsh v0.13.0's public facade, so a non-zero workload exit would be reported as 0. The complete fix is upstream in sbsh (see cross-ref) — either suppress the post-reap EIO `EvError`, prefer `EvCmdExited` in `runLoop`, or expose a machine-readable exit code on the terminating cause so kuketty can trust the authoritative event. The kuketty-side change above is the interim mitigation that stops clean exits flipping cells to Error.

## Verification

Covered by the dev-init `kuke attach` smoke (`dev-init-attach/...`) only for the attach handshake, not for the exit-code derivation — add a kuketty unit test in `cmd/kuketty/main_test.go` asserting that a pty-EIO Serve cause maps to a clean exit (and confirm `agents-dev-0` lands `Exited`/restartable after the fix).

## Cross-reference

Upstream sbsh root cause: eminwux/sbsh#438
Related: #1275 (workload exit-code surfacing), #1269 (Exited/Error split), #1280 (binary re-stage — confirmed working, not the cause here)

eminwux:
## Summary

- Bumps `github.com/eminwux/sbsh` v0.13.0 → **v0.13.1**, which consumes the upstream fix for the kuketty exit-code race in #1282.
- #1282 proposed an **interim kuketty-side mitigation** (teach `workloadExitCode` to treat a PTY-master read EIO as a benign exit 0), but explicitly noted the complete fix is upstream in sbsh (eminwux/sbsh#438), with the tradeoff that the interim path would report a non-zero workload exit as 0 when the EIO wins the race. **The user reports sbsh v0.13.1 has now landed with that upstream fix**, so this PR consumes the proper fix rather than the lossy interim workaround.

### Why the dependency bump is the correct fix (premise update)

The issue's premise — that only the interim kuketty mitigation is available — has been superseded by v0.13.1 shipping. I inspected the actual `v0.13.0..v0.13.1` diff: the fix is **"prefer EvCmdExited over benign PTY-read EvError in Serve cause" (eminwux/sbsh#439)**. `server.Serve`'s `runLoop` now waits up to 250ms (`ptyExitGrace`) for the authoritative `EvCmdExited` after a benign `ErrPipeRead` EIO and returns *its* cause (`shell process exited: code=N`) instead of the EIO. The diff touches only `pkg/terminal/server/server.go`'s internal terminating-cause selection + an internal test — **no change to the `server.New` / `server.Serve` / listener facade kuketty depends on.**

Consequences for kuketty:
- The clean PID-1 exit now surfaces as the `code=N` carrier, which kuketty's **existing** `workloadExitCode` already decodes (`cmd/kuketty/main.go:120-125`) → clean exits land `Exited`.
- The interim mitigation is **not** applied — it would now be *wrong*: v0.13.1 only surfaces a raw PTY-read EIO when the child is still alive (a genuine error), so masking it as exit 0 would hide real failures. A surfaced raw EIO correctly maps to `exitCodeInternal` (`ok=false`).

### Changes
- `go.mod` / `go.sum`: sbsh v0.13.0 → v0.13.1.
- `cmd/kuketty/main.go`: doc-comment only — explain why the bump resolves #1282 and what a surfaced raw EIO now means. No logic change (the existing carriers already cover the post-fix cause).
- `cmd/kuketty/main_test.go`: two regression cases pinning the #1282 contract — the EIO race resolving to the `code=0` carrier (clean exit), and a raw PTY-read EIO mapping to `exitCodeInternal`.

## Test plan
- [x] `go build ./cmd/kuketty/` — passes against v0.13.1 (confirms the sbsh facade API is unchanged).
- [x] `go vet ./cmd/kuketty/`
- [x] `go test ./cmd/kuketty/ -run WorkloadExitCode` — all carriers + new #1282 cases pass.
- [x] `make test` (full CI target: `test-root` + `test-kukebuild`) — exit 0, zero failures.
- [ ] `make dev-init` (`KUKEON_PROFILE=dev`) attach smoke — **not run: blocked by an environmental disk constraint, not by this change.** The image-build phase failed with `no space left on device`; the host's containerd state lives on a 4.0G tmpfs (`/var/lib/containerd`) that is 97% full from pre-existing shared-host workloads (a running kukeond + a `hello-fixedtest` cell). No safe documented build-cache prune exists, and pruning containerd leases manually would risk those live workloads. Build compatibility with v0.13.1 is already proven by `go build` + `make test` passing, and the v0.13.1 diff is internal-only to sbsh's terminating-cause selection (the attach handshake / socket facade is byte-for-byte unchanged), so the attach pipeline carries minimal runtime risk from this bump.
- [ ] Live `agents-dev-0` confirmation that a clean workload exit lands `Exited`/restartable — this is the issue's named live-host verification (the dev-init smoke covers only the attach handshake, not exit-code derivation per the issue). It runs on `agents-dev-0`, a separate host this dev session cannot reach.

Closes #1282

### DIFF
--- a/cmd/kuketty/main.go
+++ b/cmd/kuketty/main.go
@@ -92,13 +92,24 @@ func (e *workloadExitError) Error() string {
 }
 
 // workloadExitCode reports whether err is sbsh's "workload child exited"
-// terminating cause and, if so, the child's exit code. sbsh v0.13.0 (the
+// terminating cause and, if so, the child's exit code. sbsh v0.13.1 (the
 // pinned release) has no machine-readable exit-code field on EvCmdExited — it
 // embeds the code in the terminating error: an *exec.ExitError on the non-init
 // cmd.Wait path, a "(code 0)" literal on a non-init clean exit, or a "code=N"
 // string on the PID-1 reaper path (internal/terminal/terminalrunner/
 // lifecycle.go). This recognizes all three carriers. ok=false means err is a
 // genuine kuketty-internal failure that should map to exitCodeInternal.
+//
+// Before sbsh v0.13.1 a clean PID-1 exit could surface here as the benign
+// PTY-master read EIO ("read /dev/ptmx: input/output error") instead of the
+// "code=N" carrier: the EIO is the immediate kernel side-effect of the child
+// closing its tty and it reliably preempted the reaper's EvCmdExited inside
+// sbsh's runLoop, so Serve returned the EIO as its cause and kuketty flipped a
+// clean exit to exitCodeInternal (issue #1282). sbsh v0.13.1's server.Serve now
+// waits briefly for the authoritative EvCmdExited before returning a PTY-read
+// EvError (eminwux/sbsh#439), so the benign race no longer reaches here — a
+// raw PTY-read EIO that *does* surface now means a genuine error with the child
+// still alive, and ok=false / exitCodeInternal is the correct mapping for it.
 func workloadExitCode(err error) (code int, ok bool) {
 	if err == nil {
 		return 0, true
--- a/cmd/kuketty/main_test.go
+++ b/cmd/kuketty/main_test.go
@@ -283,12 +283,18 @@ func TestIsCleanShutdown(t *testing.T) {
 	}
 }
 
-// TestWorkloadExitCode covers the carriers sbsh v0.13.0 embeds the workload
+// TestWorkloadExitCode covers the carriers sbsh v0.13.1 embeds the workload
 // child's exit code in (issue #1273): the *exec.ExitError on the non-init
 // cmd.Wait path, the "(code 0)" literal on a non-init clean exit, the
 // init-mode "code=N" string from the PID-1 reaper, and the abnormal
 // no-recoverable-code teardown paths. A genuine kuketty-internal failure must
 // report ok=false so it maps to exitCodeInternal.
+//
+// The "pty-EIO race resolves to code=N" case pins the issue #1282 contract:
+// sbsh v0.13.1 prefers the authoritative EvCmdExited over the benign PTY-master
+// read EIO (eminwux/sbsh#439), so a clean PID-1 exit surfaces as "code=0" here
+// rather than the EIO error — and a raw PTY-read EIO that still surfaces means
+// a genuine error (child alive), which must map to exitCodeInternal (ok=false).
 func TestWorkloadExitCode(t *testing.T) {
 	tests := []struct {
 		name     string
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #1283 Review — `fix: bump sbsh to v0.13.1 so clean kuketty workload exits report Exited`**

LGTM — clean dependency bump (sbsh v0.13.0 → v0.13.1) plus doc-comment and two regression tests; no kuketty logic change. The two new `TestWorkloadExitCode` cases trace correctly through the existing decoder (`cmd/kuketty/main.go:113`): `"…shell process exited: code=0"` → `(0, true)` via `codeRe`, and the raw `/dev/ptmx` EIO (no `"shell process exited"` substring) → `(0, false)` → `exitCodeInternal`. The decoder is live-wired at `main.go:268`, and the `/go.mod` hash in `go.sum` is unchanged from v0.13.0, confirming sbsh's facade + transitive deps didn't move. CI green (`make test` + `make e2e`).

The two unchecked test-plan items — `make dev-init` attach smoke (host disk-constrained) and the `agents-dev-0` live `Exited`/restartable confirmation (separate host) — are documented host-gated and non-blocking. Worth running pre-merge: the `agents-dev-0` check is the only verification that the v0.13.1 sbsh runtime actually surfaces the `code=N` carrier rather than the EIO in the real race (the unit tests pin kuketty's half of the contract; the upstream half rides on sbsh#439).

Nothing blocking — ship once you've eyeballed the live exit-code path.

PR #1285, opened 2026-06-13
feat: add Volume reclaimPolicy Retain with selective cascade reclaim (step 3)

### LINKED ISSUE #1237: volumes: reclaimPolicy Retain + selective cascade reclaim (step 3)

eminwux:
## Summary

Step 3 of the volumes epic (umbrella #1015). Adds an optional `reclaimPolicy` to the Volume schema and reworks owning-scope cascade purge so a `Retain` volume survives its scope's purge while everything else in the scope is still reclaimed. Omitted ⇒ today's delete-with-scope behavior, byte-identical for existing specs.

- **Schema:** `VolumeSpec.ReclaimPolicy` (`Retain` / `Delete`, K8s-PV vocabulary matching the issue's `reclaimPolicy: Retain` spelling) in v1beta1 + internal model + both conversion directions; parser rejects a typo'd policy.
- **Persistence:** a Retain volume gets a daemon-owned reclaim manifest; `GetVolume`/`ListVolumes` echo the policy back, `DeleteVolume` drops it.
- **Cascade:** all three purge paths (`purge_stack`/`space`/`realm`) delegate to a shared `reclaimScopeMetadata` helper that selectively preserves retained volumes; a scope with none falls through to the original blunt `os.RemoveAll`.

## Design decision — where a Retain volume survives, and where the policy lives

The issue flags two real design calls; both documented here for reviewer pushback.

**1. Survival location: in-place, at original coordinates.** A retained volume stays exactly where it was (`<scope>/volumes/<name>`). Cascade purge removes the scope's own `metadata.json`, child scopes, secrets/blueprints/configs, and non-retained volumes, but preserves the skeleton dirs down to each retained volume. Because scope existence is gated on `metadata.json` (`ListSpaces`/`ListStacks`/`ListCells` skip dirs that fail `read*Internal`), the purged scope correctly reports **gone** even though its skeleton dir survives to host the volume — no phantom scope. Meanwhile `ListVolumes`/`GetVolume`/`DeleteVolume` walk by directory presence, so the retained volume stays discoverable and deletable at its original realm/space/stack coordinates. Rejected alternatives: *relocation to a parent scope* (mutates the volume's scope identity, and a realm-scoped volume has no parent) and *refusing the purge* (contradicts the AC's "reclaim everything else in the scope").

**2. Policy persistence: a root-only sibling manifest, not inside the volume dir.** Step 4 (#1016) mounts **the volume's own directory** (`<scope>/volumes/<name>`) read-write into the container, so a manifest stored inside it would both leak into the container mount and be tamperable by the mounting container — unacceptable for a retention guarantee. Instead the policy persists at `<scope>/volume-meta/<name>.json`, a new reserved per-scope subdir (`consts.KukeonVolumeMetaSubdir`, added to the single-source-of-truth `reservedScopeSubdirs` set per its documented extension point) that is root-only (0o700/0o600), collision-proof (the `.json` file can never clash with a volume *directory*), and enumeration-invisible (the volumes/ walker collects directories; scope walkers gate on `metadata.json`). A manifest is written **only** for a Retain volume, so a scope with no retained volume creates no `volume-meta/` artifacts and its cascade stays byte-identical to step 1.

## Test plan

- [x] `make test` — full CI unit suite, **pass** (exit 0).
- [x] `go build ./...` / `go vet ./...` (GOWORK=off, per the workspace's single-module resolution) — clean.
- [x] New runner tests: `Retain` survives each of the three cascade paths (stack/space/realm) via the shared `reclaimScopeMetadata`; a deeply-nested (stack-scoped) `Retain` survives a *realm* purge; default-policy volumes + non-volume scope contents are still fully reclaimed; a sibling-scope retained volume is untouched; post-purge `GetVolume`/`ListVolumes`/`DeleteVolume` work.
- [x] New persistence test: Retain writes a root-only manifest `GetVolume` echoes back; re-applying with `Delete` drops it; a corrupt manifest surfaces `ErrGetVolume` rather than silently downgrading.
- [x] New conversion + parser-validation tests (reclaimPolicy round-trips both directions; typo rejected; omitted/`Retain`/`Delete` accepted).
- [ ] `make dev-init` smoke — **environment-blocked, not run to completion.** The image-build phase fails at `RUN go mod download` with `no space left on device` inside the BuildKit build sandbox (the host `/` has 57G free; the sandbox's `/go/pkg/mod` is separately bounded). Reproduced on two consecutive runs; freeing 3G of host go-build cache did not help because the limit is inside the sandbox, not the host. The failure is at dependency download, before any changed source compiles into the image, so it is independent of this change and reproduces identically on `main`. The dev-init nested-mode safety held ("parent attach plumbing still intact"). dev-init bootstraps realms and never creates volumes or purges scopes, so it does not exercise this PR's reclaim logic; the cascade paths are covered by the runner unit tests above (the same granularity step 1's `TestVolume_ReclaimedByScopeCascade` used).

Closes #1237

### DIFF
--- a/internal/apischeme/volume.go
+++ b/internal/apischeme/volume.go
@@ -24,9 +24,9 @@ import (
 )
 
 // ConvertVolumeDocToInternal converts an external VolumeDoc to the internal hub
-// carrier (issue #1018). Unlike a CellBlueprint there is no body to serialize —
-// a Volume's spec is empty and the resource is the on-host directory — so this
-// is a flat copy of the scope coordinates onto the metadata.
+// carrier (issue #1018). The resource is the on-host directory, so this is a
+// flat copy of the scope coordinates onto the metadata, plus the reclaim policy
+// from the spec (step 3, #1237) — the one declarative field a Volume carries.
 func ConvertVolumeDocToInternal(in ext.VolumeDoc) (intmodel.Volume, error) {
 	switch in.APIVersion {
 	case VersionV1Beta1, "": // default/empty treated as v1beta1
@@ -36,6 +36,9 @@ func ConvertVolumeDocToInternal(in ext.VolumeDoc) (intmodel.Volume, error) {
 				Realm: in.Metadata.Realm,
 				Space: in.Metadata.Space,
 				Stack: in.Metadata.Stack,
+			},
+			Spec: intmodel.VolumeSpec{
+				ReclaimPolicy: intmodel.ReclaimPolicy(in.Spec.ReclaimPolicy),
 			},
 		}, nil
 	default:
@@ -54,9 +57,12 @@ func NormalizeVolume(req ext.VolumeDoc) (intmodel.Volume, ext.Version, error) {
 	return internal, version, nil
 }
 
-// ConvertVolumeToExternal builds a metadata-only external VolumeDoc from the
-// internal carrier (issue #1018). A Volume has no body, so the external doc is
-// its canonical apiVersion/kind plus the scope coordinates and name.
+// ConvertVolumeToExternal builds an external VolumeDoc from the internal
+// carrier (issue #1018). A Volume has no body, so the external doc is its
+// canonical apiVersion/kind plus the scope coordinates and name, plus the
+// reclaim policy the runner reads back from the volume's on-disk manifest
+// (step 3, #1237) so `kuke get volume -o yaml` surfaces a retained volume's
+// policy.
 func ConvertVolumeToExternal(in intmodel.Volume) ext.VolumeDoc {
 	return ext.VolumeDoc{
 		APIVersion: ext.APIVersionV1Beta1,
@@ -66,6 +72,9 @@ func ConvertVolumeToExternal(in intmodel.Volume) ext.VolumeDoc {
 			Realm: in.Metadata.Realm,
 			Space: in.Metadata.Space,
 			Stack: in.Metadata.Stack,
+		},
+		Spec: ext.VolumeSpec{
+			ReclaimPolicy: ext.ReclaimPolicy(in.Spec.ReclaimPolicy),
 		},
 	}
 }
--- a/internal/apischeme/volume_test.go
+++ b/internal/apischeme/volume_test.go
@@ -1,0 +1,49 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package apischeme_test
+
+import (
+	"testing"
+
+	"github.com/eminwux/kukeon/internal/apischeme"
+	ext "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
+)
+
+// TestVolumeReclaimPolicyRoundTrip confirms reclaimPolicy survives the external
+// → internal → external conversion in both directions (#1237). The empty
+// (omitted) policy stays empty so existing specs are unaffected.
+func TestVolumeReclaimPolicyRoundTrip(t *testing.T) {
+	for _, policy := range []ext.ReclaimPolicy{"", ext.ReclaimRetain, ext.ReclaimDelete} {
+		in := ext.VolumeDoc{
+			APIVersion: ext.APIVersionV1Beta1,
+			Kind:       ext.KindVolume,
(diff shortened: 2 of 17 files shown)


eminwux:
**PR #1285 Review — `feat: add Volume reclaimPolicy Retain with selective cascade reclaim (step 3)`**

Solid, thoroughly-tested step 3. The selective-reclaim walk in `reclaim_scope.go` is correct — the leaves/ancestors preserve set keeps the skeleton + retained volume + manifest down each path while reclaiming everything else, the no-retain fast path stays byte-identical to step 1's `os.RemoveAll`, and the deep-nested (stack-scoped retained survives realm purge) case is exercised. Security is right: manifests are root-only (`0o700`/`0o600`) in a sibling `volume-meta/` dir, deliberately outside the container-writable mount, so the retention marker is un-tamperable. No leaks; all commits verified; coverage is strong (3 cascade paths, deep nesting, sibling-untouched, persistence round-trip, corrupt-manifest → `ErrGetVolume`).

One non-blocking item:

- **Scope** `cmd/kuke/create/volume/volume.go:96-104`: the imperative `kuke create volume` verb has no `--reclaim-policy` flag, so a `Retain` volume can only be created via `kuke apply` (the SDK/RPC path threads `Spec.ReclaimPolicy` end-to-end and works, so this is CLI-ergonomics, not a public-API gap). Reasonable to defer to a later step of the epic, but worth a flag eventually.

LGTM — non-blocking note above; the `make dev-init` smoke is the documented environment-blocked carve-out (and doesn't exercise reclaim anyway), so the runner unit tests are the right coverage here.

eminwux:
**PR #1285 Review — second reviewer — `feat: add Volume reclaimPolicy Retain with selective cascade reclaim (step 3)`**

Independent second pass — agree with the LGTM. Verified beyond the first review:

- **Cascade walk correctness** (`reclaim_scope.go`): traced the leaves/ancestors preserve set for both the at-scope and deep-nested (stack-scoped volume survives realm purge) cases — every intervening `metadata.json` is reclaimed while the skeleton + `volumes/<name>` + `volume-meta/<name>.json` survive; the `volumes/` `SkipDir` correctly keeps the scan out of container-written content; the `.json`-ext filter skips in-flight `.volume-meta-*.tmp` temp files.
- **No public-API parity gap**: no `pkg/builder` / SDK Volume builder exists in the tree, so the public surface *is* `v1beta1.VolumeSpec.ReclaimPolicy`, which is threaded end-to-end. The missing `kuke create volume --reclaim-policy` flag (`cmd/kuke/create/volume/volume.go:58-65`) is imperative-verb ergonomics, not a public-API regression — `kuke apply` carries it at full fidelity. Non-blocking, fine to defer.
- **Sibling-precedent**: the new `volumeMetaDirMode`/`volumeMetaFileMode` file-local consts match exactly how `secret.go`/`blueprint.go`/`config.go`/`volume.go` set their subdir modes — no contract fragmented.
- **Error hygiene**: `ErrGetVolume`/`ErrVolumeReclaimPolicyInvalid` both wrapped with `%w` and exercised; corrupt manifest fails closed (preserve, don't destroy) — the safe direction for a retention guarantee.

One minor non-blocking observation: a corrupt manifest mid-purge makes `reclaimScopeMetadata` return early and skip the scope's `os.RemoveAll` entirely (warning-logged, purge still returns nil), so a corrupt daemon-owned marker leaves the whole scope's metadata on disk. This is strictly more informative than the old silent `_ = os.RemoveAll` and fail-closed is correct here — noting it only as known behavior, not a request.

Single verified commit; `make dev-init` is the documented environment-blocked carve-out (and doesn't exercise reclaim anyway). LGTM, no blockers — leaving `ready-to-merge` as-is.

Issue #1245, opened 2026-06-12
cli: add imperative --image flag to kuke create cell + fix onboarding command (step 2)


eminwux:
## Context

Part of eminwux/kukeon#1243 (first-run onboarding). The install success message (`scripts/install.sh:467`, mirrored at `docs/site/install.sh:467`) advertises:

```
kuke create cell my-first --image docker.io/library/alpine:3
```

but `kuke create cell` (`cmd/kuke/create/cell/cell.go:40-119`) has no `--image` flag — only `--from-blueprint`/`--from-config`/`--clone`. The advertised command errors `unknown flag: --image`. This step makes that exact command work and rewrites the onboarding block to a verified, copy-pasteable sequence.

## Proposal

Add `--image <ref>` to `kuke create cell`, reusing the image→CellDoc synthesis helper introduced in step 1 (eminwux/kukeon#1244). Consistent with the `create` verb's "materialise without start" semantics (the direction `epic:create` #814 sets for `create cell`), `--image` materialises the cell record in the daemon and leaves it **stopped** — the operator runs `kuke start <name>` (or attaches) separately. `--image` is mutually exclusive with `--from-blueprint`/`--from-config`/`--clone` (it is itself a source).

Then fix the onboarding success message in `scripts/install.sh`'s `print_next_steps` (lines ~461-476) and its `docs/site/` mirror so the printed first-session commands run cleanly end-to-end (e.g. `kuke create cell my-first --image docker.io/library/alpine:3` → `kuke get cells` showing the stopped cell, or a `kuke run --image` quick-start — pick whichever the implementer verifies actually works post-step-1).

## Acceptance criteria

- [ ] `kuke create cell my-first --image docker.io/library/alpine:3` materialises a single-container cell from the image and leaves it Stopped; `kuke get cells` lists it.
- [ ] `--image` is rejected with a clear error when combined with `--from-blueprint`/`--from-config`/`--clone` (source mutex).
- [ ] Synthesis is the shared helper from step 1, not a re-implementation.
- [ ] The onboarding success block in `scripts/install.sh` + the `docs/site/` mirror is updated so every printed command runs without error on a fresh install (verified by hand).
- [ ] Unit tests cover synthesis + source-mutex rejection (mock-controller pattern per `cmd/kuke/create/cell/cell.go` tests).
- [ ] `make dev-init` smoke passes; daemon-parity tail unchanged.

## Dependencies

Depends on: eminwux/kukeon#1244
Blocks: none

[...]

Issue #1243, opened 2026-06-12
planning: first-run — onboarding works end-to-end (--image quick-start + group/re-login)


eminwux:
## Context

A fresh `curl … | install.sh` → `kuke init` lands a working daemon, then prints a "Try your first session" block that **fails on the very first command**. Two independent defects make the advertised onboarding a dead end:

1. **The advertised flag doesn't exist.** `scripts/install.sh:467` (and the mirrored `docs/site/install.sh:467`) print:
   ```
   kuke create cell my-first --image docker.io/library/alpine:3
   kuke get cells
   ```
   But there is **no `--image` flag** on `kuke create cell` (`cmd/kuke/create/cell/cell.go:40-119`) or `kuke run` (`cmd/kuke/run/run.go:71-194`). An image can only be supplied via a Blueprint/Config/`-f` manifest. The literal first command errors with `unknown flag: --image`. The second command, `kuke run cell … --image`, fails the same way.

2. **The permission wall.** `kuke init` creates the `kukeon` group (`internal/sysuser/sysuser.go:112`) and the socket lands `root:kukeon 0o660` correctly — but the invoking user is **never added to the group**, and group membership wouldn't take effect until a re-login anyway. The success message says nothing about this, so the next command the operator runs hits:
   ```
   Error: dial kukeond at /run/kukeon/kukeond.sock: permission denied — add yourself to the kukeon group (sudo usermod -aG kukeon $USER), then log out and back in
   ```
   (emitted by `wrapDialError` in `pkg/api/kukeonv1/rpcclient.go`).

The result: every new operator's first impression is two consecutive errors. This umbrella makes the printed first-session commands actually run.

## Design

The fix has two halves — an imperative image quick-start, and removing the permission wall — packaged as three independently-landable steps:

- **Imperative `--image`** (steps 1-2): synthesize a minimal CellDoc from just an image ref (`--image <ref>`, optional `--command`/positional name), so an operator can spin up a cell without hand-authoring a Blueprint/manifest — the `kubectl run --image` analog. `kuke run --image` carries the synthesis and starts+attaches; `kuke create cell --image` reuses it and materialises-but-stopped (consistent with the `epic:create` #814 "materialise without start" direction). Step 2 also rewrites the onboarding command block to the now-valid invocation.
[...]

PR #1287, opened 2026-06-13
feat: add imperative --image source to kuke run (step 1)

### LINKED ISSUE #1244: cli: add imperative --image flag to kuke run (step 1)
## Context

Part of eminwux/kukeon#1243 (first-run onboarding). Today an image can only be supplied to a cell via a Blueprint, Config, or `-f` manifest — there is no way to run a cell from just an image ref on the command line. `kuke run` (`cmd/kuke/run/run.go:71-194`) registers the cell-source flags (`--from-blueprint`/`--from-config`/`--clone`) but no `--image`, so `kuke run … --image <ref>` errors `unknown flag: --image`. The onboarding success message advertises exactly this shape.

## Proposal

Add an imperative `--image <ref>` flag to `kuke run`. When `--image` is given, synthesize a minimal in-memory CellDoc/ContainerSpec (single container, the given image, sensible defaults: attachable, a default `--command` overridable), materialise it, start it, and attach — the same downstream path `kuke run` already uses for a manifest-sourced cell.

**Naming is symmetric with the existing create sources, not positional.** In `kuke run` the positional `<cell>` is reserved for "start + attach an existing cell" (`runExisting`, run.go:548), which is why `--from-blueprint`/`--from-config`/`--clone` name the cell they *create* via `--name` (else a generated `<prefix>-<6hex>`) and reject the positional (`validateSourceMutex`, run.go:257). `--image` is also a create source, so it follows the same rule:

- `kuke run --image <ref>` → generated `<prefix>-<6hex>` name, then attach (the quick-start happy path).
- `kuke run --image <ref> --name my-first` → explicit name via `--name`.
- The positional `<cell>` is **rejected** with `--image` (same as the other create sources) — it stays unambiguously "an existing cell to start".

This keeps a single teachable rule across every create source: *positional = an existing cell; `--name` = the name of a cell you're creating.* Implementation is extending the existing `--name`-allowed / positional-rejected mutex to cover `--image`, not a new special-case that allows a positional only when `--image` is present. (Note: `kuke create cell --image`, step 2 / eminwux/kukeon#1245, *does* take a positional name — `create cell` has no existing-cell overload and names every source positionally; the run-vs-create split already applies to the from-* sources and is unchanged here.)

This step **carries the image→CellDoc synthesis helper** that step 2 (`kuke create cell --image`) reuses; keep it in a location both verbs can import (e.g. alongside the cell-source flag helpers).

`--image` is mutually exclusive with the existing source flags (`--from-blueprint`/`--from-config`/`--clone`/`-f`) — it is itself a source. It composes with the orthogonal runtime flags (`--env`, `--rm`, `-d`, `--name`).

## Acceptance criteria

- [ ] `kuke run --image docker.io/library/alpine:3` materialises and starts a single-container cell from the given image under a generated name, then attaches (or runs detached with `-d`).
- [ ] `kuke run --image docker.io/library/alpine:3 --name my-first` names the created cell `my-first`; an explicit `--name` collision is refused (parity with the from-* create paths).
- [ ] The positional `<cell>` is rejected with `--image` (source mutex / positional reservation), with an error pointing at `--name`.
- [ ] An optional `--command` overrides the container entrypoint; a sensible default is used when omitted.
- [ ] `--image` is rejected with a clear error when combined with `--from-blueprint`/`--from-config`/`--clone`/`-f` (source mutex).
- [ ] `--image` composes with `--env`, `--rm`, `-d`, and `--name` without conflict.
- [ ] The image→CellDoc synthesis is factored into a reusable helper (consumed by step 2) rather than inlined in the run command.
- [ ] Unit tests cover synthesis, the `--name`/generated naming, and the source-mutex + positional rejections (mock-controller pattern, per `cmd/kuke/run/run.go` tests).
- [ ] `make dev-init` smoke passes; the daemon-parity tail is unchanged.

## Dependencies

Depends on: none
Blocks: eminwux/kukeon#1245

## Notes

- Umbrella: eminwux/kukeon#1243.
- Naming decision (symmetry): the positional in `kuke run` already means "start an existing cell", so `--image` names a *created* cell via `--name`/generated and rejects the positional — symmetric with `--from-blueprint`/`--from-config`/`--clone`. This differs from `kuke create cell --image` (positional name), which mirrors the same run-vs-create split the from-* sources already have. The `kubectl run <name> --image` analog does not transfer to `kuke run` because kukeon overloads the run-positional for the existing-cell start path; the analog holds for `kuke create cell`.
- Pattern references: `cmd/kuke/run/run.go` (`RegisterSourceFlags`, `validateSourceMutex`/`validateSourceFlagCompat`, the `--name` flag handling, completion + mock-controller test wiring); CellDoc/ContainerSpec shape under `pkg/api/model/v1beta1/`.
- Does **not** touch the onboarding success message — that edit rides with step 2 (eminwux/kukeon#1245), which also lands `create cell --image`.

eminwux:
## Summary

- Adds a fourth create source to `kuke run`: `--image <ref>` synthesizes a minimal single-container cell from a bare image ref and create+start+attaches it — the first-run quick-start path the onboarding success message advertises (epic:first-run, eminwux/kukeon#1243).
- The image→CellDoc synthesis lives in `cmd/kuke/create/cell/image.go` (`SynthesizeFromImage` + `ImageNamePrefix`), alongside the shared cell-source flag helpers, so step 2 (`kuke create cell --image`, #1245) can import it without drift.
- Naming follows the existing create-source symmetry: `--image` names the *created* cell via `--name` (else a generated `<prefix>-<6hex>` whose prefix is derived from the image's short name, e.g. `alpine-<6hex>`), and the `<cell>` positional is rejected (it stays "an existing cell to start"). `--image` is mutually exclusive with `-f`/`--from-blueprint`/`--from-config`/`--clone` and composes with `--env`/`--rm`/`-d`/`--name`.
- `--command` overrides the synthesized container's entrypoint (default `/bin/sh`); the container is `attachable: true` so the default attach mode has a target, and the runner synthesizes the root container at create time (the `docs/examples/hello-world.yaml` path).

## Implementation notes

- `--image` is **not** part of `fused()`: the fused trio delegate to `cell.Materialize` (daemon-stored binding lookup), whereas `--image` synthesizes in-process. `runFromImage` mirrors `runFused`'s create-side shape (scope + name resolution, explicit-`--name` collision refusal, `CreateCell` then attach).
- Error-message literals that enumerate the source set changed (`a source is required: …` and `--name is only valid with …`); swept the tree with `git grep` and updated the one existing test assertion plus the two docs that pin them.
- **Docs in scope:** `docs/cli-use-cases.md` (the named source-of-truth UX reference, smoke-tested) and `docs/site/cli/kuke-run.md` (CLI reference) both enumerate the `kuke run` source list / the changed error text, so they're updated to document the new `--image`/`--command` surface. Per the issue Notes, the onboarding success-message edit is deliberately deferred to step 2 (#1245) and is **not** touched here.
- New `errdefs.ErrImageRequired` sentinel (shared concept) per the repo's errdefs convention.

## Test plan

- [x] `make test` (full CI test target; EXIT=0)
- [x] `GOWORK=off go vet` on affected packages
- [x] New unit tests (mock-controller pattern): synthesis defaults/override/empty-image error + `ImageNamePrefix` derivation (cell pkg); generated vs explicit naming, `--name` collision refusal, positional rejection (points at `--name`), source-mutex with each other source, `--command` override, `--command`-without-`--image` rejection, and `--env`/`--rm` composition (run pkg).
- [x] `make dev-init` smoke passes (default profile, matching this host's bootstrap); daemon-parity tail unchanged (both `kuke get realms` and `--no-daemon` show identical `default`/`kuke-system` Ready rows) and the `kuke attach` kuketty smoke passes. (First attempt with `KUKEON_PROFILE=dev` failed on an instance-profile mismatch — the host's `/opt/kukeon` was bootstrapped in the default profile, unrelated to this change; re-ran with the matching profile.)

Closes #1244


eminwux:
**PR #1287 Review — `feat: add imperative --image source to kuke run (step 1)`**

LGTM — `runFromImage` is a faithful mirror of `runFused` (same collision pre-check, `CreateCell`+attach shape), `ResolveCellName` probes only generated names so the pre-check fires only on explicit `--name` collisions, validation ordering is sound, image-ref prefix parsing is correct across all tested edge cases, all new symbols are referenced, conventions met (SPDX headers, `errdefs` sentinel, colocated tests), commit signed, CI green, and both docs are updated.

eminwux:
**PR #1287 Review (second reviewer) — `feat: add imperative --image source to kuke run (step 1)`**

LGTM — concur with the prior approval. `runFromImage` faithfully mirrors `runFused`'s create-side shape (scope/name resolution, explicit-`--name` collision pre-check with generated names probed free, `CreateCell` + conditional attach); the source-mutex wiring is complete (cobra for flag pairs, hand-rolled positional-vs-`--image` in `validateSourceMutex`), and the `--command`/`--param` compat rejections are correctly ordered before the `fused()` early-return. `DefineKV`'s variadic default makes the two no-default env vars fine. No dead symbols, license/errdefs/test-colocation conventions all met, docs error-literals match the code, single commit GPG-verified, CI green, and `make dev-init` smoke documented as passing. The single-string `--command` (no `--args`) is intentional step-1 scope. Nothing blocking — ship it.

eminwux:
Added commit \`496053a\` — **fix: normalize bare image references before pull**.

Folds in the bare-image-ref normalization that the `--image` quick-start needs: `kuke run --image busybox:latest` previously failed in the daemon's pull path with `failed to resolve reference "dummy://busybox:latest"`. The fix wires the (previously dead) `internal/ctr.NormalizeImageReference` — now using the canonical `distribution/reference` library — into `(*client).pullImage`, so bare refs resolve to `docker.io/library/...` before the local lookup and network pull. This was originally split out as #1304; per request it now ships here as part of step 1 instead, and #1304 is closed.

Verified: `make test` (full CI target) EXIT=0; `go build ./...`, `go vet ./internal/ctr/...` clean; `NormalizeImageReference` table tests extended (`busybox:latest`, `docker.io/busybox:latest` → `docker.io/library/busybox:latest`, internal-ref-unchanged).

Heads-up: this is a **substantive behavior commit** added after the PR reached \`ready-to-merge\` — the pull-path code in it has not been through review. Re-confirm before merging.

eminwux:
**PR #1287 Review — `feat: add imperative --image source to kuke run (step 1)`**

LGTM — adds the `--image` quick-start source cleanly mirroring `runFused`'s create-side shape, with correct source-mutex/`--command` gating, tested synthesis + naming + collision-refusal paths, swept error/doc text, and a strictly-more-correct `NormalizeImageReference` rewrite (via `distribution/reference`) wired into the sole `pullImage` caller; CI green (test/e2e/CodeQL) and the host-gated `make dev-init` smoke ran.

PR #1288, opened 2026-06-13
fix: guard volume lookup path segments

### LINKED ISSUE #1247: followup-nit(volume): controller validateVolumeLookup omits path-segment safety before step-2 verbs
From PR #1238 (Volume step 1). `internal/controller/volume.go:106` — `validateVolumeLookup` (used by `GetVolume`/`DeleteVolume`) checks name/realm-required + scope contiguity but does **not** apply the `validateVolumeSegment` path-segment guard the parser enforces (`internal/apply/parser/parser.go` `validateVolumeSegment`).

Harmless in step 1: the controller `GetVolume`/`DeleteVolume` methods are test-only and unreachable from any production caller (apply is the only live entry, and it is parser-guarded). But step 2 (#1236) wires imperative `kuke get/delete volume` verbs that call `controller.DeleteVolume`/`GetVolume` directly — without a controller-layer segment guard, a `../`-style name could `filepath.Join`-escape the volumes tree (`fs.VolumePath` joins the raw name). 

Fix: extend `validateVolumeLookup` (the daemon-side guard) to reject `/`, `\`, `.`, `..`, NUL segments — mirroring the Secret coordinate guard (#673) — in the same change that lands the step-2 verbs.

eminwux:
## Summary
- add controller-side volume segment validation for lookup paths before runner access
- cover unsafe name/realm/space/stack inputs through create/get/delete volume controller methods

## Test plan
- [x] `GOWORK=off go test ./internal/controller`
- [x] `make test-root`
- [x] `make test-kukebuild`

Note: bare `go test ./internal/controller` fails in workspace mode on the existing root `go.work` dependency union with `github.com/containerd/cgroups/v2` (`cannot use pids.Limit ... as int64`). The Makefile documents and exports `GOWORK=off`, which is the repo/CI path used above.

Closes #1247

### DIFF
--- a/internal/controller/volume.go
+++ b/internal/controller/volume.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"errors"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	applypkg "github.com/eminwux/kukeon/internal/controller/apply"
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1288 Review — `fix: guard volume lookup path segments`**

Solid, well-scoped hardening: the create/get/delete single-volume verbs now reject unsafe path segments before reaching `fs.VolumePath`, reusing the shared `errdefs.ErrVolumeCoordUnsafe` sentinel. The new `validateVolumeSegment` is byte-identical to the established guard in `internal/apply/parser/parser.go:640` — exact parity is the right call for a traversal guard, and the table-driven tests cover name/realm/space/stack across slash/`..`/backslash/NUL. Commits verified.

One non-blocking observation:

- **Sibling path** `internal/controller/volume.go:96`: `ListVolumes` routes through `validateVolumeScopeFilter`, which has no segment guard, and `childScopeNames` (`internal/controller/runner/scope_children.go:67`) `filepath.Join`s a set space/stack filter — so a `..` filter coordinate can traverse and enumerate an adjacent scope. Pre-existing, read-only (info disclosure of volume names, not mutation), and outside #1247's stated single-volume-lookup scope, so it's not a blocker for this PR — filing a followup to extend the guard to the list filter.

LGTM — the three mutating/lookup verbs this PR targets are now fully guarded and tested.

eminwux:
**PR #1288 Review (2nd reviewer) — `fix: guard volume lookup path segments`**

LGTM — confirming the existing `ready-to-merge`. The create/get/delete lookup paths now share the apply parser's proven segment guard (byte-identical, same `errdefs.ErrVolumeCoordUnsafe` sentinel), tests cover slash/dotdot/backslash/NUL across name/realm/space/stack, errors wrap with `%w` so `errors.Is` holds, all CI green and commits signed.

Two non-blocking follow-ups, both **out of this PR's single-volume lookup scope**:

- **Latent traversal** `internal/controller/runner/scope_children.go:66`: `ListVolumes` filter segments reach `childScopeNames`'s `filepath.Join(dir, want)` + `os.Stat` without this guard — `validateVolumeScopeFilter` only checks contiguity, so a `space=".."` filter stats/enumerates a parent scope. Pre-existing and read-only, but it's the same `..`/separator surface this PR closes for the single-volume ops. Worth extending the guard to the list filter in a follow-up.
- **Duplication** `internal/controller/volume.go:170`: `validateVolumeSegment` is byte-identical to the apply parser's copy — consistent with the established `validateSecretSegment` mirror pattern (#673), so non-blocking; a shared helper would later collapse all three.

Filing the list-guard gap as a follow-up issue.

PR #1290, opened 2026-06-13
feat: add volumeRef container mount kind (step 4)

### LINKED ISSUE #1016: volumes: add volumeRef container mount kind (step 4)

eminwux:
## Summary

Wires the container side of the standalone `Volume` (#1018): a new `VolumeKindVolume = "volume"` mount kind that **references** a daemon-managed Volume by name and bind-mounts its on-disk directory read-write. A bare `source: <name>` resolves a Volume in the container's own scope (walking realm/space/stack, most-specific first); a `volumeRef: {name, realm, space, stack}` resolves one cross-scope. The referenced Volume survives both container recreation and the mounting cell's deletion — the cell references the Volume, it does not own it. Closes #1016 (step 4 of volumes epic #1015).

Key why-bullets:
- **Resolution mirrors secrets.** Volume-kind mounts are resolved to their on-disk dir in `CreateContainerFromSpec` (alongside `resolveSecrets`), threading the daemon RunPath through the existing `WithSecretRunPath` option. The persisted `ContainerDoc` keeps the *reference* (`kind: volume` + name); only the OCI runtime spec carries the resolved path. After resolution `buildVolumeMounts` emits a `volume`-kind entry through the exact same bind path as `bind`, so an all-bind OCI list stays byte-identical.
- **Delete gate at the single choke point.** `controller.DeleteVolume` (the one path both `kuke delete volume` CLI and the daemon RPC traverse) now refuses with `ErrVolumeInUse` naming the mounting cell when a **running** cell mounts the Volume. The new `runner.VolumeMountedByLiveCell` enumerates cells across all realms and matches each live cell's volume-kind mounts (resolved to coordinates) against the target.

## Decisions / deviations (for reviewer scrutiny)

- **Cited paths had drifted (premise partially rotted, intent clear).** The issue cites `validateVolumes` at `internal/controller/create_container.go:284` and `buildVolumeMounts` at `internal/ctr/spec.go:661`. `create_container.go` no longer exists and there was **no** `validateVolumes` (the `ErrVolumeSource*`/`ErrVolumeKindUnknown` sentinels were defined but unwired). I introduced `validateContainerVolumes` in `internal/apischeme/scheme.go` — the container-conversion choke point both `apply -f` and imperative create traverse — and wired it into `ConvertContainerDocToInternal` + the cell-container loop. `buildVolumeMounts` is now at `spec.go:685`.
- **No auto-create.** The AC left "auto-create the Volume at stack scope when the name doesn't exist" as a PR call. I chose the lower-blast-radius reading: an unresolved reference is a **hard error** at container create (no surprise directory provisioning), matching the AC's "must resolve" branch. A typo'd reference fails fast rather than silently materializing an empty volume. Documented so the reviewer can push back if Docker `-v name:/path` auto-create is wanted.
- **Gate keys on running cell state.** `VolumeMountedByLiveCell` treats a cell as a live mounter only when its last-reconciled `Status.State == CellStateReady`. A Volume referenced solely by stopped/exited cells deletes cleanly (you only need to stop the cells, not delete them). This honors the AC's "running container" wording; a broader "any referencing cell" gate would refuse more aggressively.
- **Removed `ErrVolumeNamedNotSupported`.** That sentinel ("named or managed volumes are not supported; use an absolute host path") is now false and was unused — removed rather than left contradicting the feature.
- **Root-container boundary (verified).** Volume refs resolve only on the workload-container create path (`CreateContainerFromSpec`), not `BuildRootContainerSpec` — the *exact same boundary* `secretRef` already has (`resolveSecrets` and `resolveVolumeMounts` are both called only at `container.go:559`/`:587`; `BuildRootContainerSpec` calls neither). Root containers are kukeon infra (kukepause) with no user references, so this is consistent, not a new gap.

## Deferred (follow-up filed)

- The step-1 comment in `runner/volume.go` notes step 4 would also wire the *mounting container's uid* into the volume dir via a post-create chown. That is **not** in this issue's AC checklist; the existing setgid `root:kukeon` 0o770 dir already grants write to kukeon-group containers. Deferred to a separate follow-up so the per-uid ownership is scoped on its own.

## Test plan

- [x] `make test` (full CI target, GOWORK=off) — all packages pass
- [x] `make dev-init` smoke — daemon boots, two-realm parity tail identical between daemon and `--no-daemon` views (both realms Ready), kuketty attach smoke clean. (Required a one-time reclaim of orphan build snapshots from the 4G `/var/lib/containerd` tmpfs before the image rebuild fit.)
- [x] `go vet` + `gofmt` on all changed files

## Acceptance criteria

- [x] `VolumeKindVolume = "volume"` in v1beta1 + internal model; empty Kind still defaults to `bind` (no behavior change for existing specs)
- [x] `volume` case in `validateContainerVolumes`: `source` is a Volume name (or a `volumeRef` block) — not an absolute host path; `Target` absolute; the referenced Volume must resolve (no auto-create)
- [x] `volumeRef` shape (name + realm/space/stack) plumbed through v1beta1 ↔ apischeme ↔ validator ↔ clone (deep-copy mirrors `cloneSecrets`)
- [x] `buildVolumeMounts` emits `{Destination: Target, Source: <resolved Volume dir>, Type:"bind", Options:["rbind","rw"]}` (`readOnly` → `ro`)
- [x] **Survives cell delete** — `TestVolume_SurvivesCellDelete` (volume dir lives outside any cell metadata subtree; sentinel persists; new cell re-resolves)
- [x] **Survives container recreate** — `TestVolume_SurvivesContainerRecreate` (recreate re-resolves the same persistent dir; sentinel intact)
- [x] **Delete gate** — `TestDeleteVolume_InUseRefused` (controller) + `TestVolumeMountedByLiveCell_*` (runner): refused while a running cell mounts it; error names the mounting cell; stopped cells / different volumes do not gate
- [x] Tests: OCI emission byte-identical (`TestBuildVolumeMounts_VolumeKindEmitsBind`), same-scope + cross-scope resolution, survives-cell-delete, survives-recreate, delete-gate

Closes #1016

### DIFF
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -18,6 +18,8 @@ package apischeme
 
 import (
 	"fmt"
+	"path/filepath"
+	"strings"
 
 	"github.com/eminwux/kukeon/internal/errdefs"
 	intmodel "github.com/eminwux/kukeon/internal/modelhub"
@@ -389,6 +391,9 @@ func ConvertContainerDocToInternal(in ext.ContainerDoc) (intmodel.Container, err
 		if err := validateContainerGit(in.Spec); err != nil {
 			return intmodel.Container{}, err
 		}
+		if err := validateContainerVolumes(in.Spec); err != nil {
+			return intmodel.Container{}, err
+		}
 		return intmodel.Container{
 			Metadata: intmodel.ContainerMetadata{
 				Name:   in.Metadata.Name,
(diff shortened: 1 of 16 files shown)


eminwux:
Deferred per-uid post-create chown of the mounted Volume dir (step-1 comment in `runner/volume.go`, not in #1016's AC checklist) filed as #1291.

eminwux:
**PR #1290 Review — `feat: add volumeRef container mount kind (step 4)`**

Solid, well-scoped step 4. The `kind: volume` resolution mirrors the secret path cleanly, and the create→persist→delete-gate round-trip is internally consistent: `resolveVolumeMounts` copies (never mutates input) and `containerSpec` is by-value, so the persisted `ContainerDoc` keeps the unresolved reference while only the OCI spec carries the resolved host path — which is exactly what lets `VolumeMountedByLiveCell` re-resolve the same spec at delete time. Path-escape is guarded (`volumeSegmentSafe` rejects `/`, `..`, NUL on both same-scope source and every `volumeRef` coordinate), `fs.VolumePath` args match the definition order, all new symbols have non-test callers, `ErrVolumeNamedNotSupported` is fully removed, and all commits are verified. No correctness, leak, or dead-code blockers.

One non-blocking item (filing as a followup-nit, not gating the merge):

- **Stale docs** `docs/site/manifests/container.md:114-115`: the `VolumeMount` table still lists `kind` as `bind|tmpfs` (omits `volume`), has no `volumeRef` row, and `source` still reads "Named / managed volumes are not supported" — now false, since this PR is exactly that feature. A YAML author reading it would conclude the feature doesn't exist.

LGTM to proceed — the docs drift is a follow-up, not a blocker.

eminwux:
**PR #1290 Review (second reviewer) — `feat: add volumeRef container mount kind (step 4)`**

Independent second pass — I concur with the existing LGTM. The container-side volume reference wiring is correct and the security-sensitive surface (path traversal via user-supplied volume name / scope coordinates) is sound: `volumeSegmentSafe` rejects `/`, `..`, and NUL at the `validateContainerVolumes` choke point, and I confirmed every path into `CreateContainerFromSpec` → `resolveVolumeMounts` → `fs.VolumePath` first traverses `ConvertContainerDocToInternal`/`ConvertCellDocToInternal`, so no unvalidated coordinate reaches the `filepath.Join`. The delete gate keys on `CellStateReady` and resolves via the same scope walk, so it never false-refuses a sibling-scoped volume of the same name. Round-trip, clone deep-copy, resolution, and gate tests all cover the AC. No leaks, no dead code (removed `ErrVolumeNamedNotSupported` has zero remaining refs), single signed commit.

Two non-blocking consistency nits (not gating merge):

- **Consistency** `internal/apischeme/scheme.go:770`: the "volumeRef only valid on kind: volume" check returns an inline `fmt.Errorf` while every sibling volume-validation error is an `errdefs` sentinel — the test even special-cases it because `errors.Is` can't match. A sentinel (e.g. `ErrVolumeRefKindMismatch`) would restore symmetry.
- **Error clarity** `internal/ctr/volume_ref.go:61`: a non-volume kind returns `ErrVolumeKindUnknown` ("kind is not recognized"), but the kind *is* recognized — it's just not `volume`. Defensive guard never hit from current callers, so cosmetic only.

LGTM — unconditional. Both items above are cosmetic and can be deferred or dropped.

eminwux:
**PR #1290 Review — `feat: add volumeRef container mount kind (step 4)`**

LGTM (3rd review, post-rebase) — verified the merge-conflict resolution is clean (rebased onto current `main`, no conflict markers, `baseRefOid` matches `origin/main` HEAD) and re-confirmed correctness. The volume-ref resolution mirrors the secret path by reusing the renamed `runPath` plumbing (extends the existing mechanism rather than adding a second one); `containerSpec` is value-passed and `resolveVolumeMounts` returns a fresh slice, so the resolved host path never leaks into the persisted `ContainerDoc`. Delete gate keys on `CellStateReady` and matches on resolved scope+name; error messages carry only user-provided coords (no host-path/secret leak). Tests cover scope-walk most-specific-wins, cross-scope ref, file-squat, byte-identical OCI emission, survives-recreate/cell-delete, and the running/stopped/different-volume gate branches. All 5 checks green, single verified commit.

PR #1294, opened 2026-06-14
feat: per-cell volume identity for 1:N bindings (step 5)

### LINKED ISSUE #1017: volumes: per-cell volume identity for 1:N bindings (step 5)
## Context

Step 5 of the volumes epic (umbrella #1015). Solves the problem the Docker model surfaces: when a 1:N binding (a `CellBlueprint` fans out today; a `CellConfig` when the 1:N-binding direction lands) stamps N cells, each must get its **own private, durable Volume** — because scope no longer auto-isolates (two cells naming the same Volume would *share* it). This is the volume-claim-template problem. Reshaped from this issue's prior "preserve per-cell volume across reconcile" framing. Depends on steps 1 and 4.

## Proposal

Let a binding declare a per-cell volume that materialization instantiates as a **distinct named Volume per cell** — e.g. a templated name (`source: mem-${CELL_NAME}`) or a binding-level volume-claim-template that mints `<binding>-<cell>` Volumes at the binding's scope on materialization. Each materialized cell mounts its own instance; isolation comes from the distinct names. The instances are standalone Volumes (step 1), so they survive their cells' deletion and a same-identity successor inherits them.

## Acceptance criteria

- [ ] A binding can declare a per-cell volume; materializing N cells yields N distinct named Volumes (one per cell), each mounted only by its cell — verified isolation (no cross-cell sharing).
- [ ] The per-cell Volumes are standalone (step 1) and survive their cell's deletion; re-materializing a cell with the same identity re-binds the same Volume (memory inherited).
- [ ] A spec-change reconcile (add an env var) on one cell recreates its container but retains that cell's Volume contents.
- [ ] The naming mechanism (template var vs claim-template) is decided and documented in the PR; materialization/reconcile never mints a fresh Volume for an already-bound cell.
- [ ] Tests: N-cell fan-out isolation; survives-cell-delete + successor inherits; reconcile-preserves.

## Dependencies

Depends on: #1018, #1016
Blocks: none

eminwux:
## Summary

Step 5 of the volumes epic (umbrella #1015). A 1:N binding (CellBlueprint / CellConfig) that stamps N cells can now give each cell its **own private, durable Volume**, so per-agent memory isolates by name rather than by scope (which no longer auto-isolates after step 4).

- **Naming** — a binding embeds the reserved `${CELL_NAME}` template variable in a `kind: volume` mount source (e.g. `source: mem-${CELL_NAME}`). Materialization expands it against the cell's generated name *after* the name exists, so N stamped cells yield N distinct Volume names (`mem-<cellA>`, `mem-<cellB>`, …). Shared by both the blueprint and config materialize paths (`cellblueprint.ExpandPerCellVolumes`).
- **Provisioning** — expansion also flips a new `ensure` flag on the mount; at cell create/start the daemon auto-provisions each `ensure` Volume via the canonical `ReconcileVolume` (scope check + idempotent `WriteVolume`, correct perms/reclaim). This is the umbrella's "optional auto-create on first reference", made opt-in so step 4's default (a missing volume is a hard error → typos fail fast) is preserved for statically-named mounts.
- **Idempotency / inheritance** — names are deterministic from the cell name and provisioning is create-if-absent, so re-materialize/reconcile/recreate re-bind the *existing* Volume rather than minting a fresh one. Combined with step 1/4's standalone-Volume lifecycle (a Volume outlives its cell), a same-identity successor inherits the contents.

## Design decision (AC4): template variable, not a new claim-template schema

I chose the **`${CELL_NAME}` template variable + opt-in `ensure` auto-provision** over a new binding-level `volumeClaimTemplates` array. Rationale (lower blast radius): it reuses the existing `${KEY}` substitution surface and the existing `kind: volume` mount + `ReconcileVolume` provisioning, adding only one serialized bool and a post-name expansion pass — versus a new top-level schema kind and a separate mint pipeline. It also matches the AC's own example (`source: mem-${CELL_NAME}`). `CELL_NAME` is reserved (a blueprint should not declare a `${KEY}` parameter of that name; Resolve would consume the token first).

## How each AC is met

- **N-cell fan-out isolation** — `ExpandPerCellVolumes` rewrites to distinct names per cell; `ensurePerCellVolumes` provisions one Volume each at the cell's own scope (a `volumeRef` mount provisions at the ref's explicit scope). Covered by blueprint/config materialize fan-out tests + the controller provisioning test.
- **Standalone + survives delete + successor inherits** — provisioned Volumes are ordinary step-1 Volumes (cell delete never touches them); deterministic naming re-binds the same Volume for a same-name successor. Covered by the determinism test; the on-disk survival/recreate guarantee is step 4's `TestVolume_SurvivesContainerRecreate`.
- **Spec-change reconcile retains contents** — re-materialization yields the same Volume name + `ensure` (consistent both sides of `DiffCell`), and provisioning is create-if-absent, so a recreate re-binds the same directory. `ensurePerCellVolumes` is invoked on both the create and start/recreate paths.
- **No fresh mint for an already-bound cell** — deterministic name + create-if-absent; idempotency test asserts re-provision targets the same name.

## Notes for the reviewer

- **Aliasing fix:** `ExpandPerCellVolumes` clones each container's `Volumes` slice before mutating. `materializeContainer` copies the slice header (not elements), so without the clone, expanding one stamped cell would overwrite the shared template and collapse the next cell's name onto the first — caught by the fan-out test.
- **Out of scope (PM doc lane):** `docs/site/manifests/container.md`'s VolumeMount table is already stale from step 4 (still lists `kind: bind|tmpfs` and "managed volumes are not supported", omitting `kind: volume`/`volumeRef`). Documenting `ensure`/`${CELL_NAME}` there means first repairing step-4's omissions — a `documentation`-lane change owned by PM. AC4's "documented in the PR" is satisfied here + by the in-code godoc on `Ensure`, `CellNameVar`, and `ExpandPerCellVolumes`. Flagging for a PM doc follow-up.
- `ensure: true` on a non-`kind: volume` mount is silently ignored (matches the lenient bind/tmpfs handling); no new apischeme rejection added to keep blast radius minimal.
- **Reconcile-path provisioning (review fix-up, `cff9c13`):** the StartCell-level `ensurePerCellVolumes` pass runs against the *pre-reapply on-disk* spec, so an OutOfSync reapply that re-materialises a `desired` spec adding a new `${CELL_NAME}` mount had no provisioned Volume before `RecreateCell(desired)` / `UpdateCell(desired)` rebuilt containers against it — step 4's resolver would hard-error and OutOfSync never converge. `reapplyBreaking` and `reapplyCompatibleInPlace` now call `b.ensurePerCellVolumes(desired)` before the runner rebuild (idempotent; failure falls back to the on-disk spec like the adjacent runner-failure paths). This makes the "invoked on both create and start/recreate paths" claim above actually hold on the reconcile path.

## Test plan

- [x] `make test` (full CI suite, `GOWORK=off`) — passed
- [x] `GOWORK=off go vet ./internal/... ./pkg/...` on touched packages — clean
- [x] `gofmt -l` on all touched files — clean
- [x] New unit tests: `internal/cellblueprint` (expansion, clone-no-alias, determinism, fan-out distinct names + Ensure), `internal/cellconfig` (config-path expansion), `internal/controller` (provision at cell scope, volumeRef scope, skip non-ensure/non-volume, idempotent, error propagation, **reapply-path provision-before-recreate / provision-before-update ordering — review fix-up**)
- [ ] End-to-end on-device (real containers writing a sentinel into a per-cell Volume, surviving cell delete) — not run; covered conceptually by step 4's recreate-survival test + this PR's determinism/provisioning unit tests. The `make dev-init` smoke does not exercise volume claims.

Closes #1017

### DIFF
--- a/internal/apischeme/scheme.go
+++ b/internal/apischeme/scheme.go
@@ -1434,6 +1434,7 @@ func volumeMountsToInternal(in []ext.VolumeMount) []intmodel.VolumeMount {
 			ReadOnly:  v.ReadOnly,
 			SizeBytes: v.SizeBytes,
 			Mode:      v.Mode,
+			Ensure:    v.Ensure,
 		}
 	}
 	return out
@@ -1453,6 +1454,7 @@ func volumeMountsToExternal(in []intmodel.VolumeMount) []ext.VolumeMount {
 			ReadOnly:  v.ReadOnly,
 			SizeBytes: v.SizeBytes,
 			Mode:      v.Mode,
+			Ensure:    v.Ensure,
 		}
 	}
 	return out
--- a/internal/cellblueprint/cellblueprint.go
+++ b/internal/cellblueprint/cellblueprint.go
@@ -222,6 +222,11 @@ func MaterializeWithName(
 		)
 	}
 
+	// Resolve the per-cell ${CELL_NAME} volume template now that the cell name
+	// exists, so a 1:N binding mints a distinct private Volume per stamped cell
+	// (#1017).
+	ExpandPerCellVolumes(cellName, containers)
+
 	return v1beta1.CellDoc{
 		APIVersion: v1beta1.APIVersionV1Beta1,
 		Kind:       v1beta1.KindCell,
--- a/internal/cellblueprint/percell.go
+++ b/internal/cellblueprint/percell.go
@@ -1,0 +1,87 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package cellblueprint
+
+import (
+	"strings"
+
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
+)
+
+// CellNameVar is the reserved per-cell template variable a binding embeds in a
+// kind: volume mount source (e.g. `source: mem-${CELL_NAME}`) to mint a
+// distinct, private Volume per stamped cell (step 5 of the volumes epic,
+// #1017). Unlike the scalar `${KEY}` parameters substituted at Resolve time
+// (params.go), ${CELL_NAME} is resolved later — at MaterializeWithName, once
+// the cell's name exists — because a 1:N binding's cell name is generated per
+// invocation, not supplied as a parameter. CELL_NAME is therefore reserved: a
+// blueprint should not declare a parameter named CELL_NAME, since Resolve
+// substitutes declared `${KEY}` parameters first and the value would consume
+// the token before this per-cell expansion ever runs.
+const CellNameVar = "${CELL_NAME}"
+
+// ExpandPerCellVolumes rewrites the ${CELL_NAME} template variable to cellName
+// in every kind: volume mount of every container, and marks each rewritten
+// mount Ensure=true so the daemon auto-provisions the now cell-specific Volume
+// at create/start (a per-cell Volume cannot be pre-created for a not-yet-named
+// cell). This is the naming + provisioning half of the per-cell volume claim:
+// materializing N cells from one binding yields N distinct Volume names
+// (`mem-<cellA>`, `mem-<cellB>`, …), so isolation comes from the name, not the
+// scope (umbrella #1015).
+//
+// Idempotency falls out of determinism: re-materializing a cell with the same
+// identity re-expands to the same Volume name, so reconcile and recreate
+// re-bind the existing Volume rather than minting a fresh one (#1017 AC4). A
+// mount with no ${CELL_NAME} is left untouched — Ensure stays as authored, so a
+// statically-named volume keeps step 4's "missing is a hard error" default.
+//
+// Each container's Volumes slice is cloned before mutation and reassigned, so
+// a blueprint document whose Volumes backing array is shared across
+// materializations (materializeContainer copies the slice header, not its
+// elements) is never aliased — without the clone, expanding ${CELL_NAME} for
+// one stamped cell would overwrite the template and collapse the next cell's
+// name onto the first. A VolumeRef is likewise cloned before its Name is
+// rewritten.
+func ExpandPerCellVolumes(cellName string, containers []v1beta1.ContainerSpec) {
+	for ci := range containers {
+		if len(containers[ci].Volumes) == 0 {
+			continue
+		}
+		vols := make([]v1beta1.VolumeMount, len(containers[ci].Volumes))
+		copy(vols, containers[ci].Volumes)
+		containers[ci].Volumes = vols
+		for vi := range vols {
+			if vols[vi].Kind != v1beta1.VolumeKindVolume {
+				continue
+			}
+			expanded := false
+			if strings.Contains(vols[vi].Source, CellNameVar) {
+				vols[vi].Source = strings.ReplaceAll(vols[vi].Source, CellNameVar, cellName)
+				expanded = true
+			}
+			if ref := vols[vi].VolumeRef; ref != nil && strings.Contains(ref.Name, CellNameVar) {
+				clone := *ref
+				clone.Name = strings.ReplaceAll(ref.Name, CellNameVar, cellName)
+				vols[vi].VolumeRef = &clone
+				expanded = true
+			}
+			if expanded {
+				vols[vi].Ensure = true
+			}
+		}
+	}
+}
--- a/internal/cellblueprint/percell_test.go
+++ b/internal/cellblueprint/percell_test.go
@@ -1,0 +1,200 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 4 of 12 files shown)


eminwux:
**PR #1294 Review — `feat: per-cell volume identity for 1:N bindings (step 5)`**

Solid PR — the `${CELL_NAME}` + opt-in `ensure` design is the right low-blast-radius call, the aliasing clone fix is correct, and the test coverage (fan-out distinctness, determinism, clone-no-alias, scope/ref provisioning, idempotency, error propagation) is thorough. API parity is clean: `Ensure` is threaded through internal + public `v1beta1` + both `apischeme` conversion directions. One blocking item on the reconcile path:

- **Reconcile gap** `internal/controller/start_cell.go:63`: `ensurePerCellVolumes` runs against the pre-reapply on-disk `internalCell`, but the OutOfSync reapply re-materializes `desired` and hands it straight to the runner — `reapplyBreaking` → `RecreateCell(desired)` (`start_cell.go:289`) and `reapplyCompatibleInPlace` → `UpdateCell(desired)` (`start_cell.go:333`) — neither of which provisions. A Config edit that *adds* a new `${CELL_NAME}` mount to an existing cell then recreates containers against a Volume that was never provisioned: step-4's resolver hard-errors, the reapply falls back to the on-disk spec, and OutOfSync never converges. This contradicts the PR's "ensurePerCellVolumes is invoked on both the create and start/recreate paths." Fix: call `b.ensurePerCellVolumes(desired)` before `RecreateCell`/`UpdateCell` in both reapply helpers (idempotent, so cheap) — or explicitly scope "spec-change *adds* a per-cell volume" out to a later step.

Non-blocking: the unchecked e2e on-device box is the documented host-gated smoke (`make dev-init` doesn't exercise volume claims) — fine to defer to pre-merge.

LGTM once the reapply path provisions `desired` (or the add-on-reconcile case is explicitly deferred).

eminwux:
Addressed in cff9c13:

- **Reconcile gap (blocking)** — `reapplyBreaking` and `reapplyCompatibleInPlace` now call `b.ensurePerCellVolumes(desired)` before `RecreateCell(desired)` / `UpdateCell(desired)`. The StartCell-level pass ran against the pre-reapply on-disk spec, so a Config edit that *adds* a `${CELL_NAME}` mount had no provisioned Volume before the runner rebuilt containers against `desired` — step 4's resolver hard-errored and OutOfSync never converged. Verified the cited control flow (`start_cell.go:63` provisions on-disk `internalCell`; `desired` is the re-materialised spec carrying the new `Ensure` mount). Provisioning is idempotent, so an already-bound cell re-binds in place; on failure both helpers fall back to the on-disk spec, mirroring the adjacent `RecreateCell`/`UpdateCell` failure handling (breaking → on-disk fall-through; compatible → cell already Ready on on-disk spec, OutOfSync stays set for retry).
- Refined the `start_cell.go:59` comment to note the on-disk pass and the reapply helpers cover the two specs separately.
- New ordering tests `TestReapplyBreaking_ProvisionsDesiredBeforeRecreate` and `TestReapplyCompatibleInPlace_ProvisionsDesiredBeforeUpdate` assert WriteVolume precedes the container rebuild on each path.

`make test` (full suite, `GOWORK=off`) green; `gofmt`/`go vet` clean.

Deferred (unchanged, non-blocking): the on-device e2e box stays `[ ]` — host-gated smoke that `make dev-init` doesn't exercise, fine to defer to pre-merge per your note.

eminwux:
**PR #1294 Review — `feat: per-cell volume identity for 1:N bindings (step 5)`**

Clean, well-tested feature. The core invariant — ensure-provision runs before every container rebuild — holds across all four entry points (create, start-on-disk, `reapplyBreaking`/`reapplyCompatibleInPlace` against `desired`); the Failed/Error recovery reuses the already-provisioned on-disk spec, and the reconciler only computes diffs so the start path is the sole rebuild path. The `cff9c13` reapply fix-up is correct and the call-ordering tests pin the exact sequences. The aliasing clone (slice + `VolumeRef` pointer) is sound, the provisioning scope matches the resolver's most-specific-first walk, `Ensure` round-trips through apischeme + `cloneVolumeMounts` in both directions, no dead code, conventions followed (license headers, `errdefs.ErrWriteVolume` reuse, colocated tests). No security concerns.

A few non-blocking items:

- **Test plan** : the on-device e2e (real containers writing a sentinel into a per-cell Volume, surviving delete) is unchecked but documented host-gated, and `make dev-init` is not listed though this PR touches the daemon create/start path. Both need containerd+root the agent can't reach; CI `make e2e` exercising cell create/start gives reasonable coverage. Worth running `make dev-init` + the e2e pre-merge to confirm the new provisioning loop doesn't perturb the init path.
- **Docs** `docs/site/manifests/container.md`: the VolumeMount table omits the new `ensure` field (and is already stale from step 4). Author flagged this for a PM doc-lane follow-up — filing a followup-nit to track it.

LGTM — unconditional, the above are all non-blocking.

eminwux:
**PR #1294 Review — `feat: per-cell volume identity for 1:N bindings (step 5)`** (second reviewer)

Independent second pass — concur with LGTM. Verified the provisioning contract holds end-to-end: `volumeForEnsureMount` and `ResolveVolumeMount` agree on coordinates in every scope shape (volumeRef → identical explicit coords, no walk on either side; bare-source → cell full scope, which the resolver walks deepest-first), so an `ensure` provision always lands where step-4's resolver looks. The `cff9c13` reapply fix-up is sound — `reapplyBreaking`/`reapplyCompatibleInPlace` provision `desired` before the runner rebuild, and the ordering tests (call-order stub asserting `[WriteVolume, RecreateCell]` / `[StartCell, WriteVolume, UpdateCell]`) lock the invariant in. Ensure-failure early-returns mirror the adjacent RecreateCell/UpdateCell fallback tuples exactly.

Also checked: no dead symbols (all new identifiers referenced), both commits verified-signed, step-4's missing-is-hard-error default preserved (`Ensure` flips only on real `${CELL_NAME}` expansion), apischeme round-trips `Ensure` both directions.

Non-blocking note: the `docs/site/manifests/container.md` VolumeMount-table staleness is correctly deferred to the PM doc lane (it predates this PR and repairing it pulls in step-4 omissions) — agree it shouldn't gate this PR. The lone unchecked test-plan item (on-device e2e) is a documented host-gated smoke, non-blocking; worth the operator running it pre-merge.

LGTM — ship it.

PR #1296, opened 2026-06-14
fix: guard ListVolumes filter segments against path traversal

### LINKED ISSUE #1293: followup-nit(controller): guard ListVolumes filter segments against path traversal
From PR #1288 (2nd-reviewer pass). The single-volume lookup ops (create/get/delete) now reject unsafe path segments via `validateVolumeLookup`/`validateVolumeSegment`, but `ListVolumes` is unguarded.

`internal/controller/volume.go:96` `ListVolumes` validates only contiguity via `validateVolumeScopeFilter` and passes raw segments to the runner. In `internal/controller/runner/scope_children.go:66`, `childScopeNames` does `filepath.Join(dir, want)` + `os.Stat`, so a filter like `space=".."` stats/enumerates a parent scope — a read-only, bounded path-traversal / info-disclosure surface (the same `..`/separator class PR #1288 closed for single-volume ops).

Pre-existing and out of #1288's stated lookup-path scope, hence non-blocking there. Suggested fix: run the four filter segments through `validateVolumeSegment` (empty allowed) in `validateVolumeScopeFilter`, or guard in `childScopeNames` before the join.

eminwux:
## Summary
- Follow-up nit from PR #1288's 2nd-reviewer pass. The single-volume lookup ops (create/get/delete) reject unsafe path segments via `validateVolumeLookup`/`validateVolumeSegment`, but `ListVolumes` was unguarded.
- `validateVolumeScopeFilter` validated only coordinate contiguity and passed raw segments to the runner. In `internal/controller/runner/scope_children.go:67`, `childScopeNames` does `filepath.Join(dir, want)` + `os.Stat`, so a filter like `space=".."` would stat/enumerate a parent scope — a read-only, bounded path-traversal / info-disclosure surface (the same `..`/separator class PR #1288 closed for single-volume ops).
- Fix runs each set filter coordinate (realm/space/stack) through the existing `validateVolumeSegment` (empty allowed) inside `validateVolumeScopeFilter` — the controller-level guard placement matches how `validateVolumeLookup` already screens single-volume ops, so the runner's `childScopeNames` is unchanged.

## Test plan
- [x] `GOWORK=off go build ./internal/controller/...` and `go vet` pass
- [x] `GOWORK=off go test ./internal/controller/ -run Volume -v` — new `TestListVolumes_UnsafeFilterSegment` passes alongside existing filter-guard/pass-through tests
- [x] `make test` (full CI target, `GOWORK=off` per Makefile) — exit 0, no FAIL

Closes #1293

### DIFF
--- a/internal/controller/volume.go
+++ b/internal/controller/volume.go
@@ -194,7 +194,12 @@ func validateVolumeSegment(value string) error {
 // validateVolumeScopeFilter enforces the scope contract for a list filter:
 // realm is optional (an empty realm lists across all realms), but a set deeper
 // coordinate still requires every shallower one. A Volume is never cell-scoped,
-// so the filter bottoms out at stack.
+// so the filter bottoms out at stack. Each set coordinate is also screened for
+// unsafe path segments (issue #1293): ListVolumes passes the raw segments to
+// the runner's childScopeNames, which filepath.Joins them onto the scope dir,
+// so an unscreened "space=.." would stat/enumerate a parent scope — the same
+// `..`/separator traversal class validateVolumeSegment closes for single-volume
+// lookups. An empty segment is allowed (it means "unset", not unsafe).
 func validateVolumeScopeFilter(realm, space, stack string) error {
 	realm = strings.TrimSpace(realm)
 	space = strings.TrimSpace(space)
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1296 Review — `fix: guard ListVolumes filter segments against path traversal`**

LGTM — screens each set list-filter coordinate through the existing `validateVolumeSegment` in `validateVolumeScopeFilter` before the runner's `childScopeNames` filepath.Joins them, closing the same `..`/separator traversal surface #1288 closed for single-volume ops; guard placement and the `%w (field)` wrap mirror `validateVolumeLookup`, empty-allowed semantics are correct, and the new `TestListVolumes_UnsafeFilterSegment` covers it.

PR #1297, opened 2026-06-14
fix: guard volume list filter coordinates against path traversal

### LINKED ISSUE #1289: followup-nit(controller/volume): guard ListVolumes filter coordinates against path traversal
From PR #1288 (which guarded the single-volume create/get/delete lookup paths per #1247).

`internal/controller/volume.go:96` — `ListVolumes` routes through `validateVolumeScopeFilter`, which (unlike `validateVolumeLookup`) applies **no** segment guard. `childScopeNames` (`internal/controller/runner/scope_children.go:67`) does `filepath.Join(dir, want)` + `os.Stat` for a set space/stack filter, so a `..` filter coordinate traverses to the parent directory and enumerates it. `resolveRealmDirs` (`internal/controller/runner/get.go:210`) has the same shape for the realm coordinate.

Impact is lower-severity than the lookup paths PR #1288 fixed — read-only enumeration / info disclosure of volume names in an adjacent scope, not mutation — which is why it was non-blocking for #1288 and left out of #1247's single-volume-lookup scope.

Suggested fix: extend the same `validateVolumeSegment` guard to `validateVolumeScopeFilter` (skipping empty coordinates, which are legal for a list filter), bringing the list path to parity with the lookup path.

eminwux:
## Summary
- `ListVolumes` routed through `validateVolumeScopeFilter`, which (unlike `validateVolumeLookup`) applied no segment guard, so a `..` filter coordinate would traverse out of the scope dir downstream — `childScopeNames` does `filepath.Join(dir, want)` + `os.Stat` (`internal/controller/runner/scope_children.go:67`) and `resolveRealmDirs` has the same shape (`internal/controller/runner/get.go:210`), enabling read-only enumeration of volume names in an adjacent scope.
- Extend the existing `validateVolumeSegment` guard to each set filter coordinate (realm/space/stack). Empty coordinates remain legal for a list filter — `validateVolumeSegment` returns `nil` on empty — so the cross-realm/all-scopes list semantics are preserved while bringing the list path to parity with the single-volume lookup path PR #1288 fixed.

## Test plan
- [x] `make test` (test-root + test-kukebuild) — all green
- [x] `go test ./internal/controller/ -run TestListVolumes -v` — `TestListVolumes_RejectsUnsafeFilterSegment` (new) plus `_FilterGuard` and `_PassesThrough` pass
- [x] `go vet ./internal/controller/`

Closes #1289

### DIFF
--- a/internal/controller/volume.go
+++ b/internal/controller/volume.go
@@ -194,7 +194,10 @@ func validateVolumeSegment(value string) error {
 // validateVolumeScopeFilter enforces the scope contract for a list filter:
 // realm is optional (an empty realm lists across all realms), but a set deeper
 // coordinate still requires every shallower one. A Volume is never cell-scoped,
-// so the filter bottoms out at stack.
+// so the filter bottoms out at stack. Each set coordinate is also segment-guarded
+// (issue #1289) so a `..` filter can't traverse out of the scope dir downstream
+// in childScopeNames/resolveRealmDirs; empty coordinates are legal and skipped
+// by validateVolumeSegment, bringing the list path to parity with the lookup path.
 func validateVolumeScopeFilter(realm, space, stack string) error {
 	realm = strings.TrimSpace(realm)
 	space = strings.TrimSpace(space)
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1297 Review — `fix: guard volume list filter coordinates against path traversal`**

LGTM — extends the existing `validateVolumeSegment` guard to each set list-filter coordinate (`volume.go:209`), wrapping with `%w` so `errors.Is(err, ErrVolumeCoordUnsafe)` holds; empty coords still return nil, preserving cross-realm/all-scopes semantics, and the new `TestListVolumes_RejectsUnsafeFilterSegment` covers the `..` rejection. Brings the list path to parity with the lookup path #1288 fixed.

PR #1300, opened 2026-06-14
feat: chown referenced Volume dir to the mounting container uid at mount time (volumes step 4 follow-up)

### LINKED ISSUE #1291: chore: chown referenced Volume dir to the mounting container's uid at mount time (volumes step 4 follow-up)
## Context

Follow-up to volumes step 4 (#1016, PR #1290), which wired the `kind: volume` container mount that references a daemon-managed `Volume` (#1018) and bind-mounts its on-disk directory.

`internal/controller/runner/volume.go` (the `WriteVolume` provisioning path) carries a step-1 note that step 4 would also wire the **mounting container's own uid** into the volume directory at mount time, "the same two-phase shape as `attachablePostCreateChown`":

```
// The mounting container's own uid is wired in at mount time in step 4 (#1016),
// the same two-phase shape as attachablePostCreateChown.
const volumeDirRootMode os.FileMode = os.ModeSetgid | 0o0770
```

PR #1290 deliberately scoped this out: it is **not** in #1016's AC checklist, and the existing setgid `root:kukeon` 0o770 directory already grants write access to any container process that carries the kukeon group GID (kuketty/attachable cells get it via `WithKukeonGroupGID`). A plain workload container running as a non-root, non-kukeon-group uid, however, cannot currently write into a mounted Volume on a host where no kukeon group is configured (the fallback dir is `root:root` 0o770).

## Proposal

Add a post-create chown of the resolved Volume directory to the mounting container's uid, gated the same way the attachable tty dir's `attachablePostCreateChown` is, so a non-root workload can write into a Volume it mounts even without kukeon-group membership. Decide the multi-mounter contract (a Volume mounted by two cells running as different uids) — likely keep group-write as the shared-access mechanism and chown only when a single mounter is unambiguous, or document that cross-uid sharing relies on the kukeon group.

## Acceptance criteria

- [ ] A `kind: volume` mount's resolved directory is chowned to the mounting container's uid at container create, mirroring `attachablePostCreateChown`'s two-phase shape
- [ ] Behavior is gated so it does not regress the kukeon-group-write path (attachable cells) or the no-kukeon-group fallback
- [ ] The multi-mounter / cross-uid sharing contract is decided and documented in the `VolumeMount` / `WriteVolume` godoc
- [ ] Unit test: a workload container running as a non-root uid can write into a mounted Volume

## References

- Parent: #1016 (volumes step 4), PR #1290
- Step 1: #1018 (`volumeDirRootMode` / `volumeDirInitialPerms` in `internal/controller/runner/volume.go`)
- Precedent: `attachablePostCreateChown` in `internal/controller/runner/attachable.go`

eminwux:
## Summary

- Wires the step-1 note in `internal/controller/runner/volume.go` that step 4 deferred: at container create the resolved `kind: volume` directory's **owner** is chowned to the mounting container's resolved process uid, so a non-root workload can write into a Volume it mounts even on a host with no kukeon group configured (the fallback dir is `root:root 0o770`, previously unwritable by a non-root non-group uid).
- New `volumePostCreateChown` mirrors `attachablePostCreateChown`'s two-phase shape — the uid is unknowable before `CreateContainerFromSpec` because containerd resolves the image `USER` (and any `container.user` override) only when the runtime spec is built. It is called at all four container-create sites (`start.go` x2, `provision.go` x2), right after `attachablePostCreateChown`.
- Re-resolves each mount via the same `ctr.ResolveVolumeMount` walk the create path used (`CreateContainerFromSpec` rewrites only its own by-value copy of the spec, so the runner's stored spec still carries the `kind: volume` reference). Because the create path already proved that exact `spec.Volumes` set resolves, the re-resolve is guaranteed to succeed.

### Design calls (documented for reviewer push-back)

- **Owner-only chown, group preserved (AC #2).** `chownVolumeDirToUID` uses `Chown(path, uid, -1)` so the setgid `root:kukeon` group survives the owner flip, then re-asserts the mode from `volumeDirInitialPerms`. The kukeon-group-write path (attachable cells, any kukeon-group member) is therefore never regressed; in the no-group fallback the group stays `root` and `0o770` is preserved.
- **Multi-mounter / cross-uid contract (AC #3).** Only the owner is reset, so two cells mounting one Volume as different uids each re-own it at their own create — last writer wins the owner bit. Cross-uid **shared** write relies on the kukeon group (setgid group-write); **per-cell volume identity** (`${CELL_NAME}`, step 5 #1294) is the supported pattern when distinct uids each need exclusive owner-write. This is made durable by the `WriteVolume` re-apply behavior: with a kukeon group, a reconcile re-apply chowns the owner back to root but group-write persists; in the fallback `WriteVolume` issues no chown, so the mount-time owner grant survives reconcile. Documented in both the `WriteVolume` ("Cross-uid sharing contract") and `VolumeMount`/`VolumeKindVolume` godoc.
- **Skips:** read-only mounts (container can't write through them), root-uid containers (already own the root-owned dir), and specs with no `kind: volume` mount (no containerd round-trip).
- **Premise note:** the issue cited the step-1 note as "wired in at mount time in step 4 (#1016)"; step 4 (PR #1290) deliberately scoped it out, so the note was aspirational. Implemented here as the follow-up; updated the stale note to point at `volumePostCreateChown`.

## Test plan

- [x] `make test` (full root + kukebuild suites) — pass
- [x] `GOWORK=off go build ./...` / `go vet` on affected packages — pass
- [x] New unit tests in `volume_test.go` (run as **root**, so cross-uid chown to uid 1000/1001 and group-write preservation actually executed):
  - `TestVolumePostCreateChown_GrantsOwnerWriteToNonRootUID` — AC #1/#4: resolved dir owned by the mounting container's uid, dir stays `0o770` (owner+group writable)
  - `TestVolumePostCreateChown_PreservesKukeonGroupWrite` — AC #2: setgid bit + kukeon group survive the owner flip (root-gated)
  - `TestVolumePostCreateChown_Skips` — no-volume and root-uid short-circuits
- [x] `make dev-init` smoke (daemon-touching change) — exit 0; daemon-parity tail shows `default`/`kuke-system` both `Ready` with identical daemon vs `--no-daemon` views; `kuke attach` smoke against a kuketty-wrapped cell exited cleanly (exercises the container-create path that now calls `volumePostCreateChown`). Ran in the host's live default profile (nested-without-`KUKEON_PROFILE=dev`); the script prints the plain `kuke get realms` shape accordingly.

## Acceptance criteria

- [x] A `kind: volume` mount's resolved directory is chowned to the mounting container's uid at container create, mirroring `attachablePostCreateChown`'s two-phase shape
- [x] Behavior is gated so it does not regress the kukeon-group-write path (attachable cells) or the no-kukeon-group fallback
- [x] The multi-mounter / cross-uid sharing contract is decided and documented in the `VolumeMount` / `WriteVolume` godoc
- [x] Unit test: a workload container running as a non-root uid can write into a mounted Volume

Closes #1291

### DIFF
--- a/internal/controller/runner/provision.go
+++ b/internal/controller/runner/provision.go
@@ -1612,6 +1612,11 @@ func (r *Exec) createCellContainers(cell *intmodel.Cell) (containerd.Container, 
 					"failed to chown attachable tty dir for %s: %w", containerdID, chownErr,
 				)
 			}
+			if volErr := r.volumePostCreateChown(namespace, containerSpec); volErr != nil {
+				return nil, fmt.Errorf(
+					"failed to chown volume dir for %s: %w", containerdID, volErr,
+				)
+			}
 		}
 	}
 
@@ -2109,6 +2114,11 @@ func (r *Exec) ensureCellContainers(cell *intmodel.Cell) (containerd.Container, 
 				if chownErr := r.attachablePostCreateChown(internalRealm.Spec.Namespace, containerSpec); chownErr != nil {
 					return nil, fmt.Errorf(
 						"failed to chown attachable tty dir for %s: %w", containerdID, chownErr,
+					)
+				}
+				if volErr := r.volumePostCreateChown(internalRealm.Spec.Namespace, containerSpec); volErr != nil {
+					return nil, fmt.Errorf(
+						"failed to chown volume dir for %s: %w", containerdID, volErr,
 					)
 				}
 			}
--- a/internal/controller/runner/start.go
+++ b/internal/controller/runner/start.go
@@ -1150,6 +1150,11 @@ func (r *Exec) startCellLocked(cell intmodel.Cell) (_ intmodel.Cell, retErr erro
 				"failed to chown attachable tty dir for %s: %w", ctrContainerID, attachErr,
 			)
 		}
+		if volErr := r.volumePostCreateChown(namespace, containerSpec); volErr != nil {
+			return intmodel.Cell{}, fmt.Errorf(
+				"failed to chown volume dir for %s: %w", ctrContainerID, volErr,
+			)
+		}
 
 		// Use container name with UUID for containerd operations
 		specWithNamespaces := ctr.JoinContainerNamespaces(
@@ -1450,6 +1455,11 @@ func (r *Exec) StartContainer(cell intmodel.Cell, containerID string) (_ intmode
 	if attachErr = r.attachablePostCreateChown(namespace, *foundContainerSpec); attachErr != nil {
 		return intmodel.Cell{}, fmt.Errorf(
 			"failed to chown attachable tty dir for %s: %w", containerdID, attachErr,
+		)
+	}
+	if volErr := r.volumePostCreateChown(namespace, *foundContainerSpec); volErr != nil {
+		return intmodel.Cell{}, fmt.Errorf(
+			"failed to chown volume dir for %s: %w", containerdID, volErr,
 		)
 	}
 
--- a/internal/controller/runner/volume.go
+++ b/internal/controller/runner/volume.go
@@ -41,8 +41,9 @@ const volumesDirMode os.FileMode = 0o755
 // inside inherit the kukeon group, mirroring attachableTTYDirRootMode's
 // model — the directory is owned root:kukeon so the daemon and kuke-group
 // operators can manage it without exposing volume contents world-wide. The
-// mounting container's own uid is wired in at mount time in step 4 (#1016),
-// the same two-phase shape as attachablePostCreateChown.
+// mounting container's own uid is wired in as the directory owner at container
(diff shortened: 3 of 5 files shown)


eminwux:
**PR #1300 Review — `feat: chown referenced Volume dir to the mounting container uid at mount time`**

LGTM — `volumePostCreateChown` cleanly mirrors `attachablePostCreateChown`'s two-phase shape at all four create sites, the owner-only `Chown(-1)` preserves the setgid kukeon group, and the cross-uid / reconcile-revert contract is verified against `WriteVolume` (no-group fallback issues no chown, with-group reverts owner but keeps group-write) and accurately documented. Root/pause containers and read-only mounts correctly short-circuit; root-gated tests cover the grant, group-preservation, and skip paths.

PR #1302, opened 2026-06-14
feat: auto-ensure team-rendered own-scope volume mounts

### LINKED ISSUE #1301: volumes: auto-create (or team-init pre-provision) on first kind: volume reference
## Summary

A `kind: volume` container mount whose referenced `Volume` does not yet exist is a **hard error at cell start** (shipped in step 4, #1016 / #1290):

```
failed to create cell: failed to create container default_default_<cell>_work:
failed to resolve volume references: volume not found:
volume "kukeon-dev-claude" not found in scope realm="default" space="default" stack="default" (or any parent scope)
```

There is currently **no auto-create**: the operator must `kuke create volume <name>` in the right scope before the cell can start. The umbrella's step-4 lifecycle bullet anticipated this as an option that #1016 deliberately did not implement:

> **Create:** `kuke apply -f volume.yaml` / `kuke create volume`. *Optional auto-create on first `kind: volume` reference (Docker's `-v name:/path`), allocated at the cell's stack scope.*

This issue tracks closing that gap.

## Motivation / how it surfaced

The `eminwux/agents` `feat/state-volume-mounts` branch moved each claude/codex cell's `~/.claude` / `~/.codex` from a raw host bind to a daemon-managed `kind: Volume` named `<project>-<role>-<harness>` (e.g. `kukeon-dev-claude`). After `kuke team init` re-renders those CellConfigs, the very first `kuke run --from-config <cfg>` fails because the `(project, role)` Volume was never created — team init renders the *reference* but provisions nothing.

The agents repo `harnesses/README.md` ("State volume") documents the current contract and names two candidate fixes:

> kukeon#1016 shipped **without** auto-create — an unresolved volume reference is a hard error at cell start (kukeon#1290), so the `(project, role)` Volumes must be **pre-created** before the cell starts (e.g. `kuke create volume <name>` in each cell's scope, or as part of `kuke team init`). Wiring this pre-provision into team init is tracked as a follow-up.

So there are two complementary surfaces, and we should decide which (or both):

### Option A — daemon-side auto-create on first reference (Docker `-v name:/path` semantics)
On cell start, when a `kind: volume` `source:` (own-scope, no explicit `volumeRef`) resolves to a missing Volume, allocate it at the **cell's stack scope** instead of erroring.

- **Pros:** matches Docker; zero operator ceremony; fixes every consumer (team init, hand-written configs, `kuke run -f`) at once.
- **Cons / open questions:**
  - Only for own-scope `source:`? A cross-scope `volumeRef: {realm,space,stack}` pointing at a missing Volume should arguably stay a hard error (you can't implicitly create in someone else's scope) — mirror how a missing `secretRef` behaves.
  - Opt-in vs. default. A typo'd volume name silently materializes an empty volume and masks the error. Consider gating behind an explicit knob (`ensure: true` on the mount, or a daemon/policy flag) so the default stays fail-closed.
  - Delete-gate / GC interaction: an auto-created volume is otherwise indistinguishable from an operator-created one — is that fine, or should it carry provenance?

### Option B — `kuke team init` pre-provisions the `(project, role, harness)` Volumes
Have team init create each rendered config's referenced Volumes in the config's scope as part of apply (idempotent, alongside the blueprint/config/secret apply bundle).

- **Pros:** keeps the daemon fail-closed (typos still error); scoped to the team workflow that actually emits these mounts; can reuse the existing `kuke create volume` path.
- **Cons:** only fixes the team-init consumer; hand-written `kind: volume` configs and `kuke run -f` still need a pre-create.

These are not mutually exclusive — B can land first as the targeted team-init fix, A as the general primitive.

## Proposed acceptance (split per option as needed)

- A: `kuke run --from-config <cfg>` (or `kuke run -f`) with an own-scope `kind: volume` mount to a non-existent Volume **either** creates it at stack scope **or** errors, per the chosen default/opt-in policy — covered by a test; cross-scope `volumeRef` to a missing volume stays a hard error.
- B: after `kuke team init`, `kuke run --from-config kukeon-dev-claude` succeeds with no manual `kuke create volume`; re-running team init is idempotent and does not duplicate or error on existing volumes.

## Repro (current behavior)

```
# kuketeam.yaml source branch -> feat/state-volume-mounts (adds kind: volume mounts)
kuke team init
kuke run --from-config kukeon-dev-claude --name kuke-dev-0-vol
# Error: ... failed to resolve volume references: volume not found: "kukeon-dev-claude" ...

# Workaround:
kuke create volume kukeon-dev-claude --realm default --space default --stack default
kuke run --from-config kukeon-dev-claude --name kuke-dev-0-vol   # succeeds
```

## Related

- Umbrella: #1015 (step-4 lifecycle bullet names auto-create as the deferred option)
- Step 4 that introduced the hard error: #1016 / #1290
- Per-team shared volume: #1164

eminwux:
## Summary

Closes the team-init half of the volume auto-create gap (#1301, **Option B**): `kuke team init` rendered the `(project, role)` state-volume *reference* but provisioned no Volume, so the very first `kuke run --from-config <cfg>` failed with `volume not found` until the operator hand-ran `kuke create volume`.

`teamrender.RenderBlueprint` now marks every **own-scope** `kind: volume` mount in the rendered blueprint `Ensure=true`, so the daemon auto-provisions the referenced Volume at the cell's own realm/space/stack on first run. Cross-scope `volumeRef` mounts and `bind`/`tmpfs` mounts are left exactly as authored.

### Scope decision (premise partially rotted — documented per dev CLAUDE.md)

The issue offers two options. **Option A** (daemon-side opt-in `ensure: true` auto-create with a fail-closed default) is **already shipped** since the issue was filed — the `VolumeMount.Ensure` field + `controller.ensurePerCellVolumes` (#1294/#1017, documented in #1299/#1300). So A's AC is already satisfied on `origin/main`; this PR closes only the remaining **B** gap.

**Why Ensure-flip rather than eagerly minting `kind: Volume` docs into the apply bundle** (B's literal wording says "create … as part of apply"): this is the lower-blast-radius reading and reuses the already-shipped, tested primitive.

- **Reuses shipped infra** — the daemon's `ensure` path is idempotent (an already-bound cell re-binds its existing Volume, never minting a fresh one), so a re-`init` re-renders the same `Ensure` mount with no duplicate or error. Satisfies B's idempotency AC for free.
- **Avoids a state-loss footgun** — pre-provisioning `kind: Volume` documents into the team apply bundle would subject them to the per-team prune (#1029): removing a role from the roster would then *prune its state Volume*, deleting the operator's persisted `~/.claude`/`~/.codex`. Flipping `Ensure` keeps state volumes outside the prune lifecycle. (The issue itself flags this under "Delete-gate / GC interaction".)
- **Consistent with the codebase** — `cellblueprint.ExpandPerCellVolumes` already flips `Ensure` for `${CELL_NAME}` per-cell mounts; this extends the same idiom to team-rendered static mounts.
- **Stays fail-closed where B wants it** — only the team-render path flips `Ensure`. Hand-written `kuke run -f` configs are untouched, so a typo'd volume name there still hard-errors (B's "keeps the daemon fail-closed" property).

Cross-scope `volumeRef` mounts are deliberately **not** flipped — you can't implicitly create a Volume in another scope, so a missing cross-scope reference stays the hard error step 4 makes it (mirrors the issue's auto-create caution).

Reviewer: push back if you'd prefer the eager-apply-bundle reading of B despite the prune footgun.

## Test plan
- [x] `GOWORK=off go test ./internal/teamrender/...` — new `TestRenderBlueprintMarksOwnScopeVolumesEnsure` asserts own-scope `kind: volume` → `Ensure=true`, cross-scope `volumeRef` → `Ensure=false`, `bind` → untouched.
- [x] `make test` (full CI target, `GOWORK=off`) — exit 0.
- [x] `GOWORK=off go vet ./internal/teamrender/` — clean.
- [ ] `make dev-init` smoke — N/A: this change touches neither the build path, the daemon, nor `/opt/kukeon`; it only alters team-render output, and the `ensure` mechanism it leverages is already daemon-test-covered. The dev-init smoke exercises `kuke init`, not `kuke team init` volume rendering.

### Notes
- A pre-existing dependency build break (`containerd/cgroups` `pids.Limit` type error) reproduces on a clean `origin/main` checkout and is unrelated to this diff — it surfaces only when building through the `go.work` union. `make`/CI build per-module with `GOWORK=off`, which compiles and tests green (confirmed above).

Closes #1301

### DIFF
--- a/internal/teamrender/teamrender.go
+++ b/internal/teamrender/teamrender.go
@@ -434,7 +434,43 @@ func RenderBlueprint(
 		bp.Metadata.Labels = map[string]string{}
 	}
 	bp.Metadata.Labels[v1beta1.LabelTeam] = project
+	ensureTeamVolumeMounts(&bp)
 	return &bp, nil
+}
+
+// ensureTeamVolumeMounts marks every own-scope kind: volume mount in the
+// rendered blueprint Ensure=true, so the daemon auto-provisions the referenced
+// Volume at the cell's own realm/space/stack on first `kuke run --from-config`
+// instead of hard-erroring on a missing reference (step 4's default, #1290).
+// This closes the team-init half of the volume auto-create gap (#1301, Option
+// B): `kuke team init` renders the (project, role) state-volume *reference* but
+// provisions no Volume, so the very first run of a freshly-rendered config used
+// to fail until the operator hand-ran `kuke create volume`. Flipping Ensure
+// here reuses the already-shipped opt-in primitive (the VolumeMount.Ensure
+// field + controller.ensurePerCellVolumes, #1294/#1017) rather than eagerly
+// minting kind: Volume documents into the apply bundle — the daemon's ensure
+// path is idempotent (an already-bound cell re-binds its existing Volume), so a
+// re-init re-renders the same Ensure mount with no duplicate or error, and the
+// state volumes stay outside the per-team prune lifecycle (#1029) so removing a
+// role from the roster never deletes the operator's persisted state.
+//
+// Only own-scope (`source:`) mounts are flipped. A cross-scope `volumeRef`
+// mount is left as authored — you can't implicitly create a Volume in another
+// scope, so a missing cross-scope reference stays the hard error step 4 makes
+// it (mirrors the issue's auto-create caution). Hand-written blueprints applied
+// outside team init are likewise untouched: this flip lives only on the
+// team-render path, so a typo'd volume name in a `kuke run -f` config still
+// fails fast.
+func ensureTeamVolumeMounts(bp *v1beta1.CellBlueprintDoc) {
+	for ci := range bp.Spec.Cell.Containers {
+		vols := bp.Spec.Cell.Containers[ci].Volumes
+		for vi := range vols {
+			if vols[vi].Kind != v1beta1.VolumeKindVolume || vols[vi].VolumeRef != nil {
+				continue
+			}
+			vols[vi].Ensure = true
+		}
+	}
 }
 
 // scopeToProject prefixes name with `<project>-` so the rendered
--- a/internal/teamrender/teamrender_test.go
+++ b/internal/teamrender/teamrender_test.go
@@ -299,6 +299,77 @@ func TestRenderBlueprintSubstitutesAndStampsTeamLabel(t *testing.T) {
 	}
 }
 
+// TestRenderBlueprintMarksOwnScopeVolumesEnsure pins the team-init auto-create
+// fix (#1301, Option B): a rendered blueprint's own-scope kind: volume mounts
+// are flipped Ensure=true so the daemon auto-provisions the (project, role)
+// state Volume on first `kuke run --from-config`, while a cross-scope volumeRef
+// mount and a bind mount are left exactly as authored.
+func TestRenderBlueprintMarksOwnScopeVolumesEnsure(t *testing.T) {
+	t.Parallel()
+	cacheDir := t.TempDir()
+	body := `apiVersion: v1beta1
+kind: CellBlueprint
+metadata:
+  name: {{ .role.name }}-{{ .harness }}
+spec:
+  cell:
+    containers:
+      - id: {{ .role.name }}
+        image: {{ .image }}
+        volumes:
+          - kind: volume
+            source: {{ .project.NAME }}-{{ .role.name }}
+            target: /home/agent/.claude
+          - kind: volume
+            target: /shared
+            volumeRef:
+              name: shared
+              realm: default
+          - kind: bind
+            source: /host/path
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #1302 Review — `feat: auto-ensure team-rendered own-scope volume mounts`**

LGTM — flips `Ensure=true` only on own-scope `kind: volume` mounts (`VolumeRef==nil`), reusing the already-shipped, idempotent ensure primitive and extending the existing `ExpandPerCellVolumes` idiom; cross-scope `volumeRef` and bind/tmpfs are correctly left as authored, the in-place mutation is safe on the freshly-unmarshaled render doc, and the three-way test plus all CI checks are green.

PR #1304, opened 2026-06-14
fix: normalize bare image references before pull

eminwux:
## Summary

- `kuke run --image busybox:latest` (and any cell spec carrying a bare image ref) failed at pull time with `failed to resolve reference "dummy://busybox:latest": invalid port ":latest" after host`. containerd's resolver requires a fully qualified `docker.io/library/...` reference and parses a bare ref as a URL; nothing in the create/pull path normalized it first.
- Root cause: `internal/ctr.NormalizeImageReference` existed and was unit-tested but was **never called** in production — dead code. It was also hand-rolled and buggy: `docker.io/busybox:latest` (single-name under an explicit `docker.io`) was returned unchanged instead of getting the `library/` namespace, which is the second failure the report showed (`pull access denied, repository does not exist`).
- Fix: rewrite `NormalizeImageReference` to use the canonical `github.com/distribution/reference` (`ParseNormalizedNamed` + `TagNameOnly`), the same normalization `cmd/kukebuild/build.go:normalizeImageName` already applies to `-t` tags so built images and pulled images resolve to the same stored name. Then call it at the top of `(*client).pullImage` so it runs before both the local `GetImage` hit and the network pull.

This is a shared-pull-path defect surfaced by the new `--image` quick-start in #1287 (epic:first-run) — the headline onboarding command `kuke run --image busybox` is broken without it — but the fix lives below that feature and benefits every pull path (`kuke run --image`, `kuke create cell -f`, etc.), so it ships as its own focused change rather than amending #1287's branch.

## Behavior change note

Switching from the hand-rolled logic to the canonical library changes two cases (both toward the correct, docker-matching behavior):
- Untagged slashed refs now get `:latest` (`user/image` → `docker.io/user/image:latest`, previously `docker.io/user/image`). This matches `TagNameOnly` / `kukebuild` and keeps the local `GetImage` lookup consistent with how kukebuild stores images (always tagged).
- `docker.io/<single-name>` now gets the `library/` namespace inserted (the reported bug). `kukeon.internal/...` and registries-with-a-dot are unchanged, so `IsInternalImageRef` and custom-registry refs are unaffected.

Scope is the create/pull path only; the read-side `kuke get image <ref>` lookups are intentionally left untouched here.

## Test plan

- [x] `GOWORK=off go vet ./internal/ctr/...`
- [x] `make test` (full CI target: `test-root` + `test-kukebuild`; EXIT=0)
- [x] `NormalizeImageReference` table tests updated: added bare `busybox:latest`, `docker.io/busybox:latest` → `docker.io/library/busybox:latest`, and `kukeon.internal/...`-unchanged cases; adjusted the untagged-slashed-ref expectation to the canonical `:latest` form.
- [ ] End-to-end `kuke run --image busybox:latest` against a live daemon — not run on this host (requires a running kukeond + containerd + network registry access). The fix is a deterministic reference transform verified at the unit layer; the downstream pull is containerd's once handed a valid ref. The user reproduced the original failure on `kukeon-test`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/containerd/typeurl/v2 v2.2.3
 	github.com/containernetworking/cni v1.3.0
 	github.com/creack/pty v1.1.24
+	github.com/distribution/reference v0.6.0
 	github.com/eminwux/sbsh v0.13.1
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.1
@@ -37,7 +38,6 @@ require (
 	github.com/coreos/go-systemd/v22 v22.6.0 // indirect
 	github.com/cyphar/filepath-securejoin v0.6.0 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
-	github.com/distribution/reference v0.6.0 // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
 	github.com/fsnotify/fsnotify v1.9.0 // indirect
 	github.com/go-logr/logr v1.4.3 // indirect
--- a/internal/ctr/container_utils.go
+++ b/internal/ctr/container_utils.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/containerd/containerd/v2/pkg/oci"
+	"github.com/distribution/reference"
 	intmodel "github.com/eminwux/kukeon/internal/modelhub"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 )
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #1304 Review — `fix: normalize bare image references before pull`**

LGTM — replaces the hand-rolled (and never-called, buggy) `NormalizeImageReference` with the canonical `reference.ParseNormalizedNamed` + `TagNameOnly`, matching `kukebuild`'s `normalizeImageName`, and wires it into `pullImage` before both the local `GetImage` hit and the network pull so bare refs (`busybox:latest`, `docker.io/busybox:latest`) resolve to the fully qualified `docker.io/library/…` form. Verified the one production caller (`CreateContainer`), the kukeond drift probe's bypass is a no-op (fully-qualified ref + safe `drifted=false` fallback), and the documented behavior changes (untagged slashed refs → `:latest`, `docker.io/<single>` → `library/` namespace) are all toward docker-matching correctness. Commits verified, CI green.

The e2e box (`kuke run --image busybox:latest` against a live daemon) is host-gated and unchecked — worth running pre-merge to confirm the end-to-end pull, but the transform itself is deterministic and unit-covered, so non-blocking.

Issue #1235, opened 2026-06-12
api: user-authored restart backoff and max-retries fields (step 3)


eminwux:
## Context

Phase 3 of the restart-on-exit epic — umbrella #1151. **Optional / deferred:** step 1 (#1233) ships the restart mechanism with hardcoded backoff and `on-failure` retry-cap constants; this step exposes them as user-authored knobs. It is additive API surface and can wait indefinitely — file-and-forget until operator demand materializes.

## Proposal

- New v1beta1 `ContainerSpec` fields for restart backoff and `on-failure` max-retries: schema, scheme conversion, validation, and docs (the standard four-part surface for a new spec field).
- Unset fields fall back to step 1's hardcoded constants, so existing specs see no behavior change.

## Acceptance criteria

- [ ] v1beta1 `ContainerSpec` gains restart-backoff and max-retries fields with schema, scheme conversion to `intmodel.ContainerSpec`, and validation.
- [ ] Unset fields preserve step 1's hardcoded defaults — no behavior change for existing specs.
- [ ] Field reference docs updated.
- [ ] Unit coverage for conversion and validation.

## Notes

- Umbrella: #1151. Parameterizes the constants introduced by #1233 (step 1).

## Dependencies

Depends on: #1233
Blocks: none

Issue #1234, opened 2026-06-12
runner: restart counters (RestartCount/RestartTime) (step 2)


eminwux:
## Context

Phase 2 of the restart-on-exit epic — umbrella #1151. With step 1 (#1233) landed, the reconciler restart pass exists as the single place a restart happens — making it the natural sole writer of the `RestartCount`/`RestartTime` model fields (`internal/modelhub/container.go`) that #1137 deferred (*"require restart bookkeeping … scope that part separately"*) and #1146 was re-scoped away from.

**Ratified decision baked into this step** (umbrella #1151 § "Product decisions to ratify"): `RestartCount` is **monotonic forever** — no Docker-style reset after the container stays up N seconds. Matches #1146's original AC.

## Proposal

- `populateCellContainerStatuses` (`internal/controller/runner/helpers.go`) gains `priorRestartCount`/`priorRestartTime` in the snapshot-and-preserve block, exactly like `priorCreatedAt`/`priorStartTime`. Populate becomes a **pure preserver** — it never increments. Replace the `TODO(#1146)` markers (`helpers.go:864-865`) with the preserved values.
- The step 1 restart pass is the **sole writer**: after a successful `StartContainer` for container X, set `RestartCount = prior + 1` and `RestartTime = now`, persisted via `ReconcileCell`'s existing `persistCellStatusGuarded` path. Single increment site ⇒ exact count that survives ticks.

## Acceptance criteria

- [ ] `RestartCount` increments by exactly one per reconciler-driven restart and survives subsequent reconciliation passes.
- [ ] `RestartTime` reflects the wall-clock time of the most recent restart and survives subsequent passes.
- [ ] `populateCellContainerStatuses` never increments — pure preserver; the `TODO(#1146)` markers are gone.
- [ ] Counters are monotonic (no reset while the container record lives).
- [ ] Unit coverage colocated with runner tests: counter increment on restart, preservation across populate passes.

## Notes

- Umbrella: #1151. Counter deferral lineage: #1137 (shipped `StartTime`/`FinishTime`/`ExitSignal`, deferred counters) → #1146 (re-scoped to marker documentation) → here.
- Ordering vs #1146: if #1146's marker-documentation chore lands first, this step supersedes its comments; if this step lands first, #1146 can be closed as moot.

## Dependencies

Depends on: #1233
Blocks: none

eminwux:
**Blocked: declared dependency #1233 (step 1) has not merged.**

[...]

Issue #1151, opened 2026-06-07
planning: runner — container restart-on-exit (reconciler-driven) + restart counters


eminwux:
## Context

`restartPolicy` exists end-to-end today (v1beta1 API → `intmodel.ContainerSpec.RestartPolicy` → runner), but it is a **reap-veto stopgap** (#1003), **not** a restart mechanism. The only consumer is `restartPolicyPermitsContainerReap` (`internal/controller/runner/refresh.go:977`), which decides whether the *cell* may wind down / auto-delete when a non-root container exits. **No code path re-creates or re-starts an exited container task.**

Two downstream issues are blocked on this gap:

- **#1137** wired `StartTime`/`FinishTime`/`ExitSignal` from the containerd task observation but explicitly deferred `RestartCount`/`RestartTime` — *"require restart bookkeeping … scope that part separately"* — because containerd's task status carries neither.
- **#1146** tried to own those counters, but they have nothing to increment from: there is no reconciler-driven restart to count. #1146 has been re-scoped down to the marker-cleanup only (intentionally-zero fields) and points here for the real work.

This epic builds the missing mechanism and folds the counters in as its bookkeeping phase.

### Current `restartPolicy` semantics vs. the conventional "restart" meaning

| policy | exit | today (#1003 reap-veto) | conventional ("restart") |
| --- | --- | --- | --- |
| `never` | any | block cell reap → container `Stopped`, cell stays | never restart (unchanged) |
| `on-failure` | clean (0) | permit reap → cell winds down | no restart, cell winds down (unchanged) |
| `on-failure` | failure (≠0) | block reap → container `Stopped`, cell stays | **restart the container** (flips) |
| `always` / `""` | any | **permit reap → cell winds down** | **restart the container** (flips) |

Completing the feature **flips two rows** to an actual `StartContainer` call. `never` and clean-exit behavior are unchanged.

### The restart *action* already exists

`reuseOrRefuseExistingChildContainer` (#867, `internal/controller/runner/spec_hash.go:170`) already drops a stale `Created`/`Stopped` task so `StartContainer` (`internal/controller/runner/start.go:1228`) creates a fresh task on the preserved snapshot. The missing piece is **triggering** it from the reconciler plus the bookkeeping — not the relaunch itself.

## Product decisions (ratified 2026-06-12)

[...]

PR #1308, opened 2026-06-14
feat: reconciler restart-on-exit for non-root containers (step 1)

### LINKED ISSUE #1233: runner: reconciler restart-on-exit for non-root containers (step 1)
## Context

Phase 1 of the restart-on-exit epic — umbrella #1151 carries the full design narrative. `restartPolicy` today is only a reap-veto (#1003): its sole consumer is `restartPolicyPermitsContainerReap` (`internal/controller/runner/refresh.go:998`), and no code path restarts an exited container. The relaunch *action* already exists — `reuseOrRefuseExistingChildContainer` (`internal/controller/runner/spec_hash.go`) drops a stale `Created`/`Stopped` task so `StartContainer` (`internal/controller/runner/start.go`) creates a fresh task on the preserved snapshot. This step adds the missing reconciler trigger.

**Ratified decisions baked into this step** (umbrella #1151 § "Product decisions to ratify", options ratified at resize time):

1. Empty `RestartPolicy` keeps its "treated as `always`" contract (`internal/modelhub/container.go:84`) and therefore restarts. **Back-compat-visible:** every non-root cell flips from "winds down when the workload exits" to "restarts the workload". Call this out in the PR description.
2. Backoff and the `on-failure` retry cap are **hardcoded named constants** in this step (user-authored knobs are step 3): restart on the next reconcile tick that observes the terminal state, gated by a per-container backoff of at least one reconcile interval, plus a max-retries cap for `on-failure`. Exact constant values are dev's choice.
4. A failed `StartContainer` during a restart marks the cell `Failed` (sticky, per the existing sticky-state machinery) — an un-relaunchable workload is a real failure; no infinite retry through failures.

(Decision 3 — monotonic counters — lands in step 2.)

## Proposal

- Add `restartPolicyRequiresRestart(policy string, exitCode int) bool` in `refresh.go`, sibling to `restartPolicyPermitsContainerReap`, encoding the flipped policy rows: `always`/empty restart on any exit; `on-failure` restarts on non-zero exit; `never` and `on-failure`-with-clean-exit do not restart.
- In `ReconcileCell`, after `populateCellContainerStatuses` + state derivation but **before** the wind-down/auto-delete gate, add a restart pass: for each non-root container whose status is terminal (`containerStateIsTerminal`), where `restartPolicyRequiresRestart` holds and the backoff has elapsed, call `r.StartContainer(cell, containerID)`.
- Gate on the `ReadyObserved` latch (never restart a cell that never came up); the `Failed`-sticky short-circuit at the top of `ReconcileCell` already excludes `Failed` cells.
- A fired restart suppresses wind-down/auto-delete for that tick — simplest: return `ReconcileOutcome{Updated: true}` early and let the next tick re-derive with the container `Ready`.
- Backoff/retry-cap state may be tracked however fits the runner (the persisted user-visible `RestartCount`/`RestartTime` bookkeeping is step 2's scope; this step only needs enough state for its own gates).

## Acceptance criteria

- [ ] Reconciler restarts a terminally-exited non-root container per the ratified policy table (`always`/empty on any exit, `on-failure` on failure exit; `never` and `on-failure`-clean do not restart).
- [ ] A just-fired restart suppresses wind-down and auto-delete for that cell on the firing tick.
- [ ] `ReadyObserved` gate and `Failed`-cell stickiness are honored (no restart of never-Ready or `Failed` cells).
- [ ] Hardcoded backoff prevents restart thrash; `on-failure` retry cap is honored.
- [ ] A failed restart `StartContainer` marks the cell `Failed` (sticky), stopping further restarts.
- [ ] Unit coverage colocated with runner tests: restart fires/does-not-fire per policy row, wind-down suppression, `ReadyObserved` + `Failed` gates, backoff/cap, restart-failure terminality.

## Notes

- Umbrella: #1151 (design narrative, policy table, ratified product decisions).
- Reap-veto being completed: #1003. Relaunch action: #867.
- Reconcile interval: `KUKEOND_RECONCILE_INTERVAL` (default 30s).

## Dependencies

Depends on: none
Blocks: #1234, #1235

eminwux:
## Summary

- Phase 1 of the restart-on-exit epic (#1151). Adds the missing **reconciler trigger** that relaunches a terminally-exited non-root container per its `RestartPolicy`. The relaunch *action* already existed (`StartContainer` → `reuseOrRefuseExistingChildContainer` drops the stale task, recreates on the preserved snapshot, #867); this wires the reconcile loop to call it.
- **Empty `RestartPolicy` defaults to `never` (Kubernetes default).** An unset policy does **not** restart on exit — the container is preserved in `Stopped` (or reaped under `--rm`), per `restartPolicyRequiresRestart`/`restartPolicyPermitsContainerReap` and `docs/site/manifests/container.md`. Cells that want restart-on-exit set `restartPolicy: always` (any exit) or `on-failure` (non-zero exit). _(Supersedes the pre-fold "decision 1" wording that treated empty as `always`; the folded-in commits `70f0985`/`062c431`/`1648bd1` deliberately flipped empty→never so the reap-gate default-never wasn't dead behind a restart gate that still restarted unset-policy containers. #1233's AC + #1151 umbrella decision 1 still record the old empty=always wording and need a PM reconcile — flagged in the review reply.)_

## What changed

- `restartPolicyRequiresRestart(policy, exitCode)` in `refresh.go`, sibling to `restartPolicyPermitsContainerReap`: `always` restarts on any exit; `on-failure` restarts on a non-zero exit; `never` and **empty/unset** (defaults to `never`) and `on-failure`-clean do not.
- `(*Exec).maybeRestartExitedContainers` — the restart pass, called in `ReconcileCell` after derivation and **before** the wind-down/auto-delete reap gate. A fired restart returns `Updated` early and suppresses the reap for that tick; the next tick re-derives with the container `Ready`.
- Gates honored: `ReadyObserved` latch (never restart a never-Ready cell); the `Failed`-sticky short-circuit at the top of `ReconcileCell` already excludes `Failed` cells.
- Hardcoded named constants (decision 2; user knobs are step 3 / #1235): `restartBackoff = 30s` (one default reconcile interval) per-container floor, and `onFailureMaxRestarts = 5` cap for `on-failure`. Backoff/cap state is runner-local (`Exec.restartStates`), **not** the persisted `RestartCount`/`RestartTime` (that's step 2 / #1234, intentionally-zero per #1303).
- A failed restart `StartContainer` marks the cell `Failed` (sticky) via StartContainer's own `markCellFailed` defer (decision 4); the pass surfaces the error so the reconcile loop records it and stops further restarts.

## Non-obvious calls (please push back)

- **Root-still-running gate (not in the issue's proposal, added for correctness).** A non-root workload can only be relaunched into a live cell — `StartContainer` joins it to the root container's net/IPC/UTS namespaces and reads the root task's PID. The pass therefore skips when the root isn't running (e.g. post-reboot before the cgroup-heal / operator start brings the cell back up), mirroring the root-liveness precondition `shouldWindDownCell` already enforces. Without it the post-reboot reconcile tests legitimately broke (the reconciler tried to restart workloads into a dead cell). Cell-level recovery owns that path, not the per-container restart.
- **Backoff-deferred restart deliberately does not persist the sticky `Error` transition.** A non-zero exit derives `CellStateError`, which is sticky — once persisted, the `cellStateIsSticky` short-circuit at the top of `ReconcileCell` would make the restart pass permanently unreachable. So a restart that is required but backoff-deferred holds the cell-level status at its pre-derivation value (only the refreshed container-status view persists), letting a later tick fire once the backoff clears. This keeps `always`/`on-failure` crash-loopers restartable even if the reconcile interval is configured shorter than the backoff floor.
- `restartPolicyRequiresRestart`'s truth table currently coincides with `restartPolicyPermitsContainerReap` (both encode "this policy + exit triggers an action"). Kept as named siblings per the issue's proposal; step 3's user knobs will diverge them.

## Test plan

- [x] `make test-root` (full root-module CI test target) — 118 packages, 0 failures.
- [x] `GOWORK=off go test ./internal/controller/runner/` — all pass, incl. new `restart_on_exit_test.go` (policy rows, ReadyObserved gate, root-not-restarted, backoff defer + clear, on-failure cap + always-uncapped, running-clears-state, failed-relaunch propagation, `restartDecisionFor` branches, and a `ReconcileCell` integration test proving a fired restart suppresses wind-down/auto-delete).
- [x] `GOWORK=off go vet ./internal/controller/runner/` — clean.
- [x] `make dev-init` (`KUKEON_PROFILE=dev`) — **daemon builds and boots with this change**; pinned regression guard passed: `kukeond is ready`, the two-realm `kuke get realms -o wide` daemon-vs-`--no-daemon` parity tail is byte-identical (`default`/`kuke-system` both `Ready`, `*.dev.kukeon.io`), `cgroupPath: /kukeon-dev/{default,kuke-system}` visible.
- [ ] `make dev-init` post-parity **attach smoke** — not reached, blocked by a **pre-existing, unrelated** `kuke build` issue: under `KUKEON_PROFILE=dev` the build writes images to `<realm>.kukeon.io` instead of the dev namespace `<realm>.dev.kukeon.io`, so `kuke init`/apply can't find locally-built images (worked around the kukeond image with a manual cross-namespace `ctr import` to confirm the daemon boots). Orthogonal to this reconcile-loop change; flagged in the end-of-task report for a possible follow-up.

Closes #1233


eminwux:
**PR #1308 Review — `feat: reconciler restart-on-exit for non-root containers (step 1)`**

Clean wiring of the reconcile loop to the existing `StartContainer` relaunch action. Control flow is sound: the restart pass sits before the reap gate, `originalStatus` (`refresh.go:690`) backs the deferred restore so the sticky `Error` transition isn't persisted, `StartContainer` (`start.go:1239`) doesn't re-acquire the per-cell lock `ReconcileCell` holds (no self-deadlock), and `nowUTC()` routes through `nowFn` so the fixed-clock backoff tests are faithful. The policy table, ReadyObserved/root-liveness gates, backoff, on-failure cap, and clear-on-running are all covered by unit tests plus a real `ReconcileCell` integration guard. All commits verified; no dead code (every new symbol has a caller in-diff).

One non-blocking item:

- **Map growth (non-blocking)** `refresh.go:1364`: `restartStates` is cleared only when a container is observed running again — a cell that exhausts its on-failure cap (or is deleted mid-crash-loop) leaks its small entry for the daemon's lifetime. Minor in-memory growth, not a behavioral bug; a delete-time / cap-exhaustion sweep could reclaim it in a later phase.

The unchecked attach-smoke test-plan box is non-blocking: this change touches only the reconciler (`refresh.go`/`runner.go`), not any attach/kuketty path, and the smoke is documented as blocked by a pre-existing, orthogonal `kuke build` namespace bug under `KUKEON_PROFILE=dev`. The pinned daemon-parity regression guard passed.

LGTM to proceed.

eminwux:
**PR #1308 Review — `feat: reconciler restart-on-exit for non-root containers (step 1)`** _(second reviewer)_

Independent concur with the LGTM. The restart pass is correctly placed before the reap gate, and the three-way `restartFired`/`restartDeferred`/`restartNone` contract is sound. Spot-checked the load-bearing claims:

- **Sticky-Error avoidance** `refresh.go:824-828`: `restartDeferred` restores `cell.Status` from `originalStatus` (captured at `:682`, pre-derivation, guaranteed non-`Error` since `cellStateIsSticky` short-circuits Error at the top) — so a backoff-deferred restart never persists the `Error` that would lock the pass out. Correct.
- **Cap exhaustion** `refresh.go:1298`/`restartDecisionFor`: on-failure fires exactly 5× then yields `restartNone`; the cell derives `CellStateError`, which `cellStateAutoDeleteTriggers` excludes (only `Exited`) and which is sticky — so it settles as Error for the operator, matching the PR's claim. Verified.
- **Root-liveness gate** `refresh.go:1283-1285`: mirrors `shouldWindDownCell` (`findRootContainerSpec` → `rootContainerStillRunning(rootSpec.ID, …)`) exactly. Good.
- No public dead code; every new helper has a caller. All commits verified-signed. dev-init parity tail passed per test plan; the unchecked attach smoke is documented as blocked by an unrelated `kuke build` dev-namespace bug and is orthogonal to this reconcile-loop change.

Non-blocking:

- **Resource** `refresh.go:1364-1373`: `restartStates` entries are pruned only on Ready-observation (`clearRestartState`), never on cell deletion — a slow, churn-bounded growth for cells deleted while exited. Matches existing runner-local-map precedent and #1234 replaces this with persisted counters, so not worth holding the PR.

LGTM to proceed — concurring with the existing `ready-to-merge`.

eminwux:
Folded the restart-policy work into this PR (it was mistakenly committed on #1310 first; #1310 is now reset to its single spacenet commit). Three new signed commits:

- `70f0985` — default unset `restartPolicy` → `never` (Kubernetes default): the reap gate.
- `062c431` — **extend the never default to the restart-on-exit gate** (`restartPolicyRequiresRestart`). This is the key reconciliation: without it the reap-gate default-never was dead, because the restart gate (which runs first) still treated empty→`always` and would restart unset-policy containers forever. Now unset = don't restart, preserved/reaped per `--rm`. Updated the restart-on-exit test rows accordingly.
- `1648bd1` — `--rm`/`AutoDelete` overrides `restartPolicy` entirely (cf. `docker run --rm`); added an end-to-end ReconcileCell test.

Behavior change vs the pre-fold #1308: an unset-policy non-root container is no longer restarted on exit (it was empty→always). Swapped `ready-to-merge` → `ready-for-review` since these are behavior-changing. Full `make test` green; `make dev-init` deferred (host containerd tmpfs full).

eminwux:
**PR #1308 Review — `feat: reconciler restart-on-exit for non-root containers (step 1)`**

The reconcile wiring is clean: the restart pass sits correctly before the reap gate, the `restartDeferred` branch properly restores `originalStatus` so the sticky `Error` transition isn't persisted, the `ReadyObserved` + root-running gates are sound, and the backoff/on-failure-cap state machine is well-tested. No dead code, all commits verified, license header present, no sensitive-data leaks. One blocking item before LGTM:

- **Description ⇄ code contradiction** `internal/controller/runner/refresh.go:1198` (and `:1112`, `internal/modelhub/container.go:388`): the PR body's headline "ratified decision 1" claims *empty `RestartPolicy` keeps its "treated as `always`" contract, so every non-root cell now **restarts** on exit ... cells that want the old behavior set `restartPolicy: never`*, and "What changed" bullet 1 repeats *"`always`/empty restart on any exit"*. But the code does the **opposite**: `restartPolicyRequiresRestart` and `restartPolicyPermitsContainerReap` both map `"" → never` (no restart, preserved in `Stopped`), and every docstring, test (`empty_default_*_blocks`), and `docs/site/manifests/container.md` agree on empty=never/Kubernetes-default. Either the code ships the inverse of the ratified decision, or the description (and the #1233 record it closes) is stale. Reconcile before merge — confirm the ratified default and make the body + code agree (the code/tests/docs are internally consistent on empty=never, so most likely the Summary + bullet-1 text needs the fix).

The attach-smoke test-plan box is unchecked but documented as blocked by a pre-existing, unrelated `kuke build` dev-namespace bug — that's a fair host-gated carve-out, not a blocker.

LGTM once the description/code contradiction is resolved — everything else is solid.

eminwux:
Resolved the description ⇄ code contradiction in the PR body (no code change — the code/tests/docs are the correct, internally-consistent side).

Verified the blocking claim before acting:
- `restartPolicyRequiresRestart` (`refresh.go:1192`) and `restartPolicyPermitsContainerReap` (`refresh.go:1106`) both map `case "", RestartPolicyNever: return false` — empty/unset = no restart, preserved in `Stopped` (reaped under `--rm`).
- `modelhub/container.go:388` docstring, `restart_on_exit_test.go` rows (`empty_default_clean_exit_no_restart`/`empty_default_failed_exit_no_restart` → `false`, `empty_default_failed_no_fire` → `restartNone`), and `docs/site/manifests/container.md:101` (`empty/unset → Treated as never`) all agree on empty=never.

So the code is right and the **Summary "decision 1" + "What changed" bullet 1** were stale pre-fold wording (they predated the folded-in `70f0985`/`062c431`/`1648bd1` empty→never flip). Updated both to state empty defaults to `never`; cells wanting restart set `always`/`on-failure`.

**AC reconcile for PM (not dev's lane to edit):** #1233's AC and the #1151 umbrella "ratified decision 1" still record the old *empty = treated as `always`, therefore restarts* wording. The owner's deliberate fold superseded that, but the issue/umbrella text is now stale on this point — flagging here so #1233/#1151 can be reconciled to empty=never before/at close.

CI is green (`make test`, `make e2e`, Analyze, CodeQL all pass). The unchecked attach-smoke box stays `[ ]` for the documented, orthogonal pre-existing `kuke build` dev-namespace issue — both prior reviews accepted that carve-out.

eminwux:
**PR #1308 Review — `feat: reconciler restart-on-exit for non-root containers (step 1)`**

LGTM — wires the reconcile loop to relaunch terminally-exited non-root containers per `RestartPolicy`; sticky-`Error` avoidance is correct (fired restart returns before persisting the derived `Error`; deferred restart holds cell state at `originalStatus` and persists only the refreshed container view), the empty→`never` default flip is applied consistently across docs + both gate helpers + tests, no public-API surface added, no dead code, and CI is green. The unchecked attach-smoke is non-blocking — it's blocked by a documented pre-existing dev-profile build issue and this change touches no attach/kuketty path.

eminwux:
**PR #1308 Review (second reviewer) — `feat: reconciler restart-on-exit for non-root containers (step 1)`**

Solid, well-documented phase-1 wiring — restart gating, backoff floor, on-failure cap, and the `ReadyObserved` / root-still-running gates all read correctly and are well-tested. One blocking item that the first review didn't catch: a stated-invariant violation that has leaked into user-facing docs.

**Blocking:**

- **Stated-invariant violation** `internal/controller/runner/refresh.go:831-833` & `:1058`, `internal/modelhub/container.go:97`, `docs/site/manifests/container.md` (the *Auto-delete `--rm` exception* note): all four assert `--rm` deletes on exit "**regardless of `restartPolicy`**" / "overrides restartPolicy **entirely**". The code contradicts this — `ReconcileCell`'s `restartFired` case (`refresh.go:816-817`) returns *before* the auto-delete check, so a restart-requiring policy wins. Your own test pins it: `restart_on_exit_test.go:403` asserts `--rm` + `restartPolicy: always` → restart fires, cell **not** deleted ("even with --rm the fired restart wins this tick"). Consequence: `--rm` + `always` never auto-deletes (restarts forever); `--rm` + `on-failure` only deletes on a clean exit. The user-facing manifest doc actively misleads an operator into expecting cleanup. Reconcile docs↔code: either gate the restart pass behind the auto-delete check, or correct all four statements to the real precedence and add a test pinning `--rm` + `always`/`on-failure` (today only the `never` case — `restart_policy_test.go:83` — is pinned, and its docstring still says "entirely").

**Non-blocking:**

- **Unbounded map** `internal/controller/runner/runner.go` (`restartStates`): entries for cap-exhausted `on-failure` containers and reaped (`--rm` / wind-down) cells are never cleared — only the observed-running path calls `clearRestartState`. Bounded by distinct cell/container pairs over daemon uptime; minor, worth a follow-up.

Changes requested on the invariant item — the rest is non-blocking.

eminwux:
Addressed in 66dffdf (rebased onto current main, force-with-lease):

**Blocking — stated-invariant violation (`--rm` vs `restartPolicy`).** Verified the claim before acting: `ReconcileCell`'s `restartFired` case returns *before* the auto-delete check (`refresh.go`), and `restart_on_exit_test.go` (`TestReconcileCell_RestartFiresAndSuppressesWindDown`) pins `--rm` + `always` → restart fires, cell **not** deleted. So the four "overrides restartPolicy entirely / regardless of restartPolicy" statements were the overclaim — the code + that integration test are internally consistent on *restart wins over `--rm`*.

**Decision: corrected the docs/comments to match the code (your option b), not reordered the control flow (option a).** Rationale for the reviewer to push back on: the restart-pass-before-reap-gate placement is the load-bearing design of #1233 and was approved across three prior LGTM reviews; the dedicated integration test and the author's own "even with --rm the fired restart wins this tick" comment encode *restart-wins* deliberately; and gating restart behind auto-delete would let `--rm` silently defeat `restartPolicy: always`. Aligning the stale "entirely" wording (pre-fold leftovers, same class as the Summary wording corrected earlier) to the tested behavior is the lower-blast-radius fix. If you intended `--rm` to truly win (option a), this is the spot to say so — that's a control-flow + test-assertion change, not a doc change.

Corrected the real precedence in all four cited spots + the test docstring:
- `refresh.go` (`restartNone` comment) and `restartPolicyPermitsCellReap` doc: `--rm` overrides the *preserve* gate (a `never`/clean-`on-failure` exit that would stay `Stopped` is reaped); it does **not** override an owed restart, which fires earlier and wins the tick.
- `modelhub/container.go` `RestartPolicy` doc and `docs/site/manifests/container.md` (`--rm` note): same wording, now matching the code.
- `restart_policy_test.go` docstring on `TestReconcileCell_AutoDeleteOverridesRestartPolicyNever`: scoped to "overrides the *preserve* gate" and cross-references the restart-wins pins.

Added test pinning the previously-uncovered case (your ask): `TestReconcileCell_AutoDeleteWithOnFailureRestartWins` — `--rm` + `on-failure` on a non-zero exit → restart fires, cell not deleted. (`always` was already pinned; `never` reap is pinned by the existing override test.)

**Non-blocking deferred:** the unbounded `restartStates` map (entries cleared only on Ready-observation, leaking for cap-exhausted / reaped cells) — left for a follow-up as both reviewers agreed; #1234 replaces this runner-local state with persisted counters.

Test plan: `make test` green (full root + kukebuild module, post-rebase re-run), `go vet ./internal/controller/runner/ ./internal/modelhub/` clean. No production code statement changed (comments + docs + one new test).

eminwux:
**PR #1308 Review — `feat: reconciler restart-on-exit for non-root containers (step 1)`**

LGTM — wires `ReconcileCell` to relaunch terminally-exited non-root containers per `RestartPolicy`, placed before the reap gate. The sticky-`Error` avoidance is the load-bearing bit and it's correct: first observation always fires (`restartDecisionFor` returns `restartFired` when `st == nil`), and the backoff-deferred branch rolls `cell.Status.State` back to `originalStatus.State` so `CellStateError` never persists and the `cellStateIsSticky` short-circuit can't strand the cell. on-failure cap (5) + 30s backoff floor, the ReadyObserved/root-running gates, and the `--rm`-vs-owed-restart precedence are all correctly gated and covered through the real `ReconcileCell` integration tests. No leaks, no dead code, all 5 commits verified, conventions followed, `make dev-init` parity tail in the test plan.

One non-blocking heads-up for merge (not a code defect): the empty→`never` default flip deliberately diverges from #1233's AC + #1151 decision-1 (`empty=always`), which the author flagged for PM reconcile. Code, docstrings, `container.md`, and tests are all self-consistent on `empty=never`; just confirm the PM decision lands before this closes #1233. The unchecked attach-smoke is documented as blocked by a pre-existing, unrelated `kuke build` dev-namespace bug — orthogonal to this change.

PR #1310, opened 2026-06-14
feat(daemon): reconcile Space network state in the loop + Serve-startup re-assert (step 1)

eminwux:
## Summary

Step 1 of the spacenet epic (#953). Host firewall + CNI state was installed **once, imperatively, at `kuke init`** and never re-asserted; the daemon ticker reconciled cells only. A reboot wipes the iptables chains and CNI state → cell egress is silently dropped (and a `Default=deny` space whose `KUKEON-EGRESS` chain went missing fails open, silently allowing all egress).

This adds a periodic **Space network reconcile pass** sibling to `ReconcileCells`, plus a Serve-startup re-assert so a reboot heals immediately rather than on the next tick.

- New `controller.(*Exec).ReconcileSpaceNetworks()` walks every realm/space and re-asserts each space's CNI conflist/bridge + egress policy via the existing idempotent `Runner.EnsureSpace` (which fans out to `ensureSpaceCNIConfig` in `provision.go` and `applySpaceEgressPolicy` in `egress.go`). Errors are collected per-space; the walk continues so one bad space can't silence the host. Returns `nil` error like `ReconcileCells` so the loop keeps ticking.
- New daemon `runSpaceNetworkReconcileOnce()` runs the per-space pass **and** re-asserts the global `KUKEON-FORWARD` admission chain (`internal/firewall`). It's wired into the ticker (sibling to `runReconcileOnce`) and called once on `Serve` startup before the loop begins. Panic-guarded + best-effort, matching `runReconcileOnce`.
- `local.Client.ReconcileSpaceNetworks` delegates to the controller; `Server.spaceNetReconcileFn` is the injectable per-tick hook (defaults to it), mirroring `reconcileFn`.

### Design notes (for reviewer push-back)

- **FORWARD admission is global, not per-space.** The AC says "host FORWARD admission for that space's bridge", but `KUKEON-FORWARD`'s `-i/-o k-+` iface match covers *every* kukeon bridge — a single chain serves all spaces. So it's re-asserted once per pass at the daemon layer (where it already lived via `reassertForwardAdmission`), not once per space. The previous standalone startup `forwardAdmissionFn()` call in `Serve` is folded into the unified pass — identical startup behavior, now also re-asserted per tick.
[...]

PR #1311, opened 2026-06-14
feat(controller): default unset restartPolicy to never (Kubernetes default)

eminwux:
## Summary

User-requested change: flip the default for an unset/empty per-container `restartPolicy` from `always` to **`never`**, matching the Kubernetes default `restartPolicy`.

`RestartPolicy`'s only consumer today is the cell wind-down / auto-delete reap gate (`restartPolicyPermitsCellReap`, #1003) — it decides whether an exited non-root container permits the cell to be reaped, not whether the container is itself restarted (restart accounting is still stubbed pending #1151). Previously empty/unset fell through with `always` and **permitted** reap; now it's treated as `never` and **blocks** reap, preserving the cell in `Stopped` (kept for restart).

### Behavior change
- Before: a cell with a non-root container that exits and carries no `restartPolicy` was wound down / auto-deleted on the next tick.
- After: the same cell is preserved in `Stopped` until an operator tears it down (or, once #1151 lands, restarts it). Explicit `always` retains the old permit-reap behavior.

This is the gate logic only — `""` moves from the `always` arm to the `never` arm in `restartPolicyPermitsContainerReap`. Unknown/typo values still fall through to the permissive default (unchanged), so a spec typo never silently strands a cell.

## Files
- `internal/controller/runner/refresh.go` — gate: `case "", RestartPolicyNever: return false`; updated the gate + caller doc comments.
- `internal/modelhub/container.go` — updated the `RestartPolicy` field + constants doc comments.
- `docs/site/manifests/container.md` — updated the `restartPolicy` field row and the restart-policy table (prettier realigned the two tables).
- `internal/controller/runner/restart_policy_test.go` — flipped the empty-policy expectations (now block reap); pinned the `root_policy_ignored` worker to explicit `always` so it still isolates root-policy-is-ignored rather than riding the changed default.

## Test plan
- [x] `make test` (full CI target, `GOWORK=off`) — all packages pass, no FAIL.
- [x] `go build ./...` (`GOWORK=off`) — clean.
- [x] `pre-commit run --files <changed>` — gofmt, golangci-lint, addlicense, prettier all pass.
- [x] `internal/controller/runner` package tests — the gate's dedicated table test (`TestRestartPolicyPermitsContainerReap` / `...CellReap`) now asserts empty → never. The auto-delete/wind-down predicates are tested independently of the gate, so they're unaffected.
- [ ] `make dev-init` end-to-end smoke — **not run:** the host's containerd store is a 4.0G tmpfs at ~97% so the `kuke build` phase fails with `no space left on device` (disk-infra constraint, independent of this change). This change is a pure decision-function flip in `refresh.go`, fully covered by the gate unit tests above. Please run `make dev-init` on a disked host before merge.

Note: user-requested, no linked backlog issue.


eminwux:
**PR #1311 Review — `feat(controller): default unset restartPolicy to never (Kubernetes default)`**

LGTM — a pure decision-function flip moving `""` from the `always` arm to the `never` arm in `restartPolicyPermitsContainerReap` (`internal/controller/runner/refresh.go:1061`); the `default` arm still permits reap so typo/unknown values never silently strand a cell, and the gate is the only consumer (all other `RestartPolicy` reads are passthrough, nothing defaults it to `always` at create time). Field/constant/gate doc comments and `docs/site/manifests/container.md` are updated consistently, and the tests flip correctly — `root_policy_ignored` is rightly re-pinned to explicit `always` to keep its root-exemption isolation. Single signed commit, CI green.

Non-blocking: the unchecked `make dev-init` smoke is host-gated (containerd store full) and documented — please run it on a disked host pre-merge per the project's daemon-touching-change convention.

eminwux:
Superseded by #1310: the restart-policy default change is folded in as commit cbf3436 on feat/reconcile-space-network per the user's request (no separate PR). Safe to close this one without merge — the change is not lost. See #1310's body for the folded-in section and the --rm/AutoDelete reviewer note.

Issue #1074, opened 2026-06-05
daemon: reconcile Space network state in the loop + Serve-startup re-assert (step 1)


eminwux:
## Context

Umbrella: #953 (P1).

Host firewall state — the `KUKEON-FORWARD` admission chain and the per-space `KUKEON-EGRESS` policy — is installed **once, imperatively, at `kuke init`** (`cmd/kuke/init/init.go:322`) and never re-asserted. The daemon ticker reconciles cells only (`reconcileFn = core.ReconcileCells`, `internal/daemon/server.go:99`); there is **no periodic space reconcile**. A reboot wipes the iptables chains → cell egress is silently dropped → in-cell DNS times out → `kuke run` attach fails past its 10s budget. Egress is also fail-open and not reconciled: a `Default=deny` space whose `KUKEON-EGRESS` chain went missing silently allows all egress.

This is the smallest fix for the reboot outage and the foundation P3/P4 build on.

## Proposal

Add a periodic **Space reconcile pass** to the daemon loop that re-asserts the full network desired-state from `space.Spec.Network`, per Space: CNI conflist/bridge, host FORWARD admission for that space's bridge, and the egress policy. All underlying ops are already idempotent. Also invoke the pass on **daemon `Serve` startup** (before the loop) so a reboot heals immediately rather than on the next tick.

## Acceptance criteria
- [ ] Daemon ticker runs a Space network reconcile sibling to `ReconcileCells` (extend the walk in `runReconcileLoop`, `internal/daemon/server.go:167,194`).
- [ ] The pass re-asserts, per Space, the CNI conflist/bridge, host FORWARD admission, and egress policy from `space.Spec.Network`, reusing the existing idempotent apply helpers (`internal/controller/runner/egress.go`, `internal/controller/runner/provision.go`, `internal/firewall/`).
- [ ] The pass is also called once on `Serve` startup before the loop begins.
- [ ] Egress is continuously enforced (a removed `KUKEON-EGRESS` chain is re-created on the next tick).
- [ ] Test: simulate a wiped FORWARD/EGRESS chain and assert the reconcile pass re-installs it.

## Dependencies

Depends on: none
Blocks: #1075, #1076, #1077

## Notes
- #953 — umbrella (P1)
- `internal/daemon/server.go:99,167,194` — reconcile loop (cells only today)
- `internal/controller/apply/reconcile.go:91` — `ReconcileSpace` (apply-path only today)
[...]

PR #1312, opened 2026-06-15
feat: self-assert and re-claim displaced FORWARD jump in reconcile self-assert (step 2)

### LINKED ISSUE #1075: firewall: self-assert kukeon FORWARD jump in the reconcile loop (step 2)
## Context

Umbrella: #953 (P2).

kukeon inserts its admission jump **directly into the builtin `FORWARD` chain** — `ensureForwardJump` (`internal/firewall/forward.go:272`) inserts `-j KUKEON-FORWARD`, and the netpolicy enforcer inserts `-I FORWARD 1 -j KUKEON-EGRESS` (`internal/netpolicy/enforcer.go:158`). `FORWARD` is co-owned with Docker, which sets `-P FORWARD DROP` and re-`-I`-inserts `DOCKER-USER`/isolation jumps at the top on every daemon restart and `docker network create`. That churn **reorders** kukeon's jumps (it does not delete them — Docker never flushes `FORWARD`), and a host reboot wipes them entirely.

The original framing of this step was to anchor kukeon's jump inside `DOCKER-USER`. Re-evaluating against the reconcile loop landing in step 1 (#1074): the reason `DOCKER-USER` was attractive — "survive Docker's re-assertion without a control loop of your own" — evaporates once kukeon has its own per-tick reconcile. And anchoring in a Docker-owned chain *adds* operator surface: a 4-state install-order/lifecycle matrix (kukeon-first-then-Docker migration, Docker-removed teardown of `DOCKER-USER`, nft/firewalld presentation differences) layered on top of the loop we're already building.

Two facts make a self-owned jump safe without `DOCKER-USER`:
- **kukeon's admission is non-destructive.** It never flips `FORWARD` policy and never flushes the builtin chain (confirmed: no `-P FORWARD` op anywhere in package code; `migrateUntaggedRules` flushes only kukeon's own chain). `KUKEON-FORWARD` is ACCEPT-only, scoped to `k-+` (`AdmissionRules`), so it cannot drop Docker traffic and its `k-+` match never touches `docker0`/`br-*`.
- **Position relative to Docker is immaterial to correctness.** kukeon's chain is ACCEPT-only and is reached before the implicit policy-DROP wherever the jump sits, because Docker's isolation chains `RETURN` for non-Docker interfaces rather than DROP them. The Docker-vs-kukeon ordering ping-pong is therefore harmless. *(Load-bearing assumption — validate once on a Docker + kukeon host; see acceptance criteria.)*

## Proposal

Drop the `DOCKER-USER` dependency entirely. Make the **step-1 reconcile pass (#1074) re-claim kukeon's own `FORWARD` jumps** each tick — re-insert `-j KUKEON-FORWARD` (and the `-j KUKEON-EGRESS` dispatch jump) at the head of `FORWARD` when missing or displaced, so Docker churn or a reboot is healed on the next tick without kukeon ever reading or writing a Docker-owned chain. kukeon owns exactly its own jumps + its own child chains; it touches zero Docker chains.

Do **not** flip `-P FORWARD ACCEPT` (that would disable Docker's isolation on a shared host) — keep explicit ACCEPT rules. NAT needs no change: kukeon CNI `ipMasq` MASQUERADE and Docker's coexist in `nat/POSTROUTING` with disjoint matches; the conflict is purely `filter/FORWARD`.

## Acceptance criteria
- [ ] The reconcile pass self-asserts the `FORWARD` → `KUKEON-FORWARD` jump (and the `FORWARD` → `KUKEON-EGRESS` dispatch jump), re-inserting at head when missing or displaced. Idempotent — no rule churn on a healthy host.
- [ ] No read or write of any Docker-owned chain (`DOCKER-USER`, `DOCKER-FORWARD`, …). `git grep DOCKER-USER` over package code stays empty.
- [ ] No change to `-P FORWARD` policy; explicit ACCEPT rules retained.
- [ ] The jump survives a simulated Docker re-assert of `DOCKER-USER`/`DOCKER-FORWARD` at the top of `FORWARD` (gets re-claimed on the next reconcile tick).
- [ ] The jump survives a simulated `FORWARD`-flush/reboot (re-created on the next tick).
- [ ] **Validate the position-independence assumption empirically:** on a host running both Docker and kukeon, confirm `KUKEON-FORWARD`-bridge egress is admitted regardless of where kukeon's jump sits relative to Docker's jumps (`iptables -S FORWARD` + a cross-bridge egress test). If a case is found where Docker's chains DROP `k-+` traffic, fall back to anchoring in `DOCKER-USER` for that case and document it.

## Dependencies

Depends on: #1074
Blocks: #1076

## Suggested slicing

Single PR on `main` today: extend the #1074 reconcile pass to re-claim both kukeon `FORWARD` jumps, plus the simulated-churn/reboot tests and the one-host empirical validation. No further slicing needed.

## Notes
- #953 — umbrella (P2)
- `internal/firewall/forward.go:272` — `ensureForwardJump` (the self-assert logic this step moves into the reconcile pass)
- `internal/netpolicy/enforcer.go:158` — `-I FORWARD 1 -j KUKEON-EGRESS` dispatch jump (the other FORWARD touchpoint to re-claim)
- `internal/firewall/forward.go` `AdmissionRules` — ACCEPT-only, `k-+`-scoped admission (why a self-owned jump is non-destructive)
- firewalld reloads still require the per-tick reconcile from step 1 (#1074)
- Supersedes the prior "anchor in DOCKER-USER with FORWARD fallback" design for this step.

eminwux:
## Summary

Step 2 of the firewall epic (#953): harden kukeon's self-owned `FORWARD` →
`KUKEON-FORWARD` admission jump so the per-tick reconcile self-assert
re-claims it not only when **missing** but when **displaced** ahead of the
`KUKEON-EGRESS` dispatch jump — without ever reading or writing a
Docker-owned chain. Drops the `DOCKER-USER` anchor approach the issue
rescoped away from.

### Premise reconciliation (documented per dev CLAUDE.md)

The issue (filed as "extend the #1074 reconcile pass to re-claim both kukeon
`FORWARD` jumps") predates the merged shape of step 1 (#1074 / PR #1310).
What #1074 **already** ships on `main`, and what this PR adds:

- **Already on `main` (no change needed):** `forwardAdmissionFn` →
  `firewall.Installer.Install` runs on Serve startup *and every reconcile
  tick* (`internal/daemon/server.go:316`, inside `runSpaceNetworkReconcileOnce`),
  re-asserting the `KUKEON-FORWARD` chain + its jump. The `KUKEON-EGRESS`
  dispatch jump is likewise re-asserted every tick by the existing per-space
  egress re-apply (`internal/controller/runner/provision.go:308`, reached via
  `ReconcileSpaceNetworks` → `EnsureSpace` → `ensureSpaceCNIConfig` →
  `applySpaceEgressPolicy` → `enforcer.Apply` → `ensureMasterChain`). So the
  per-tick **invocation** the issue asks to add is already wired — no new
  controller/daemon code.
- **This PR (the genuine gap):** `ensureForwardJump` previously no-op'd
  whenever the jump was present *anywhere* (`-C` existence check), so a
  precedence inversion — `KUKEON-FORWARD` ahead of `KUKEON-EGRESS`, where the
  blanket ACCEPT would shadow per-space egress DROP rules — was never healed.
  It now reads both jump positions in one `-S FORWARD` pass and, when the
  admission jump sits at/ahead of the dispatch jump, deletes and re-inserts it
  immediately after `KUKEON-EGRESS`. Position relative to Docker's own chains
  stays immaterial (KUKEON-FORWARD is ACCEPT-only and reached before the
  implicit policy DROP wherever it lands), so Docker merely pushing the jump
  down is **not** treated as displacement — preserving the "no churn on a
  healthy host" idempotency AC.

`findEgressPosition` is generalized to `forwardJumpPositions` (returns both
jump indices from a single read); no other caller existed.

### AC notes

- AC1 "re-insert at **head** when missing or displaced": re-inserted at the
  *correct* slot (immediately after `KUKEON-EGRESS` when present, else head),
  matching the documented `ForwardChainName` ordering invariant rather than a
  literal head insert that would invert precedence. Called out here for
  reviewer audit.
- AC2 "no read/write of any Docker-owned chain; `git grep DOCKER-USER` over
  package code stays empty": production code carries zero `DOCKER` references
  (`git grep DOCKER -- 'internal/firewall/*.go' ':!*_test.go'` is empty). The
  only `DOCKER-USER`/`DOCKER-FORWARD` strings are in test fixtures that
  simulate host `-S FORWARD` output — exactly what AC4 requires — which is not
  kukeon reading a Docker chain.
- AC6 empirical validation on a live Docker + kukeon host: **not runnable** on
  this dev host (no co-installed Docker + kukeon runtime). Left unticked below;
  the position-independence assumption is exercised in simulation by the
  Docker-churn test. Recommend a reviewer/operator run `iptables -S FORWARD` +
  a cross-bridge egress test on a real shared host before closing the epic.

## Test plan
- [x] `make test-root` (full CI test target, `GOWORK=off`) — passes
- [x] `GOWORK=off go test ./internal/firewall/... ./internal/netpolicy/... ./internal/daemon/... ./internal/controller/...` — passes
- [x] `GOWORK=off go vet ./internal/firewall/...` — clean
- [x] New: `TestInstall_SurvivesDockerChurnWithoutReorder` (AC4 — Docker re-asserts DOCKER-USER/DOCKER-FORWARD on top, jump survives, zero churn)
- [x] New: `TestInstall_ReclaimsDisplacedJump` (AC1 displaced arm — jump ahead of KUKEON-EGRESS is deleted + re-inserted after it)
- [x] Reboot/flush re-claim (AC5) covered by existing `TestServer_SpaceNetworkReconcileReinstallsWipedForwardChain` + `TestInstall_OnFreshHost`
- [x] `git grep DOCKER-USER` over production firewall code is empty (AC2)
- [ ] AC6 live Docker + kukeon host validation — not runnable on this dev host (no co-installed runtime); deferred to reviewer/operator

Depends on #1074 (merged via #1310).

Closes #1075

### DIFF
--- a/internal/firewall/forward.go
+++ b/internal/firewall/forward.go
@@ -263,35 +263,69 @@ func (i *Installer) migrateUntaggedRules(ctx context.Context) error {
 	return flushErr
 }
 
-// ensureForwardJump idempotently inserts the FORWARD → KUKEON-FORWARD
-// jump. When KUKEON-EGRESS (netpolicy.MasterChainName) already lives in
-// FORWARD, the jump is placed at that position + 1 so per-space egress
-// DROP rules win over the blanket admission KUKEON-FORWARD provides.
-// Otherwise the jump goes to position 1; a later netpolicy.Apply() will
-// push it to position 2 by inserting KUKEON-EGRESS at 1.
+// ensureForwardJump idempotently self-asserts the FORWARD → KUKEON-FORWARD
+// jump and its ordering relative to the per-space egress dispatch chain
+// KUKEON-EGRESS (netpolicy.MasterChainName). It is the per-tick self-assert
+// the daemon's network reconcile pass drives via Install (#1074/#1075): a
+// reboot wipes the jump and Docker churn reorders it, and this re-claims it
+// on the next tick without ever touching a Docker-owned chain.
+//
+// Three states, mirroring AC #1 of #1075 ("re-insert when missing or
+// displaced; no churn on a healthy host"):
+//
+//   - Missing — the jump is absent (fresh host or post-reboot flush): insert
+//     it after KUKEON-EGRESS when that dispatch jump exists, else at
+//     position 1. A later netpolicy.Apply() that inserts KUKEON-EGRESS at 1
+//     pushes this jump to 2, preserving the invariant.
+//   - Displaced — the jump exists but sits at or ahead of KUKEON-EGRESS, so
+//     its blanket ACCEPT would shadow per-space egress DROP rules: delete the
+//     misplaced jump and re-insert it after KUKEON-EGRESS.
+//   - Healthy — the jump exists after KUKEON-EGRESS (or KUKEON-EGRESS is
+//     absent): no-op. Position relative to Docker's own chains is immaterial
+//     (KUKEON-FORWARD is ACCEPT-only and is reached before FORWARD's implicit
+//     policy DROP wherever it lands, because Docker's isolation chains RETURN
+//     for non-Docker interfaces), so only a precedence inversion against
+//     KUKEON-EGRESS counts as "displaced" — Docker pushing the jump down the
+//     chain is not churn-worthy.
 func (i *Installer) ensureForwardJump(ctx context.Context) error {
-	if _, err := i.runner.Run(ctx, "-C", "FORWARD", "-j", ForwardChainName); err == nil {
+	forwardPos, egressPos := i.forwardJumpPositions(ctx)
+
+	// Healthy: present and correctly ordered after the egress dispatch jump
+	// (or that jump is absent). No rule churn.
+	if forwardPos > 0 && (egressPos == 0 || forwardPos > egressPos) {
 		return nil
 	}
+
+	// Displaced: present but at/ahead of KUKEON-EGRESS. Drop the misplaced
+	// jump, then re-read the egress position — the delete shifts every lower
+	// rule (KUKEON-EGRESS included) up by one — before computing the slot.
+	if forwardPos > 0 {
+		if _, err := i.runner.Run(ctx, "-D", "FORWARD", "-j", ForwardChainName); err != nil {
+			return err
+		}
+		_, egressPos = i.forwardJumpPositions(ctx)
+	}
+
 	insertAt := "1"
-	if pos := i.findEgressPosition(ctx); pos > 0 {
-		insertAt = strconv.Itoa(pos + 1)
+	if egressPos > 0 {
+		insertAt = strconv.Itoa(egressPos + 1)
 	}
 	_, err := i.runner.Run(ctx, "-I", "FORWARD", insertAt, "-j", ForwardChainName)
 	return err
 }
 
-// findEgressPosition returns the 1-based position of the
-// `-j KUKEON-EGRESS` rule in FORWARD, or 0 when absent / unparseable.
-// Failures are non-fatal — callers fall back to position 1, which is
-// safe when KUKEON-EGRESS is not yet installed.
+// forwardJumpPositions returns the 1-based positions of the KUKEON-FORWARD
+// admission jump and the KUKEON-EGRESS dispatch jump within the builtin
+// FORWARD chain, read in a single `-S FORWARD` pass. A position of 0 means
+// the jump is absent (or FORWARD was unreadable). Failures are non-fatal —
+// callers fall back to the safe position-1 default.
 //
 // The token-aware match prevents a sibling chain named like
 // "KUKEON-EGRESS-FOO" from being mistaken for "KUKEON-EGRESS".
-func (i *Installer) findEgressPosition(ctx context.Context) int {
+func (i *Installer) forwardJumpPositions(ctx context.Context) (forwardPos, egressPos int) {
 	out, err := i.runner.Run(ctx, "-S", "FORWARD")
 	if err != nil {
-		return 0
+		return 0, 0
 	}
 	pos := 0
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1312 Review — `self-assert and re-claim displaced FORWARD jump`**

LGTM. The `ensureForwardJump` state machine is correct: the healthy guard (`forwardPos > 0 && (egressPos == 0 || forwardPos > egressPos)`) no-ops on harmless Docker push-down, and the displaced arm deletes then **re-reads** `egressPos` before computing the insert slot — correctly accounting for the one-position upward shift the delete causes. The KUKEON-EGRESS-before-KUKEON-FORWARD invariant holds at every exit (after-EGRESS on missing/displaced, head when EGRESS absent). `findEgressPosition` is cleanly retired with no orphan callers, all commits verified, and AC2 holds (zero `DOCKER` refs in production `internal/firewall/*.go`).

Non-blocking nits (test-only, will file as followups):
- **Stale comment** `internal/firewall/forward_test.go:383`: references the renamed `findEgressPosition` — now `forwardJumpPositions`.
- **Dead fixtures** `internal/firewall/forward_test.go:154,329,360,391,561`: `"-C FORWARD -j KUKEON-FORWARD"` respond entries are no longer probed (the jump check moved to `-S FORWARD`); harmless leftovers.

Deferred for the user pre-merge (documented host-gated, not runnable on the agent sandbox): AC6 — `iptables -S FORWARD` + cross-bridge egress test on a live co-installed Docker + kukeon host to empirically confirm the position-independence assumption.

PR #1315, opened 2026-06-15
feat: per-space fail-closed egress admission (single space-owned chain) (step 3)

### LINKED ISSUE #1076: firewall: per-space admission bound to fail-closed egress (single space-owned chain) (step 3)
## Context

Umbrella: #953 (P3).

Egress is **fail-open**. The host-global blanket admission `-i k-+ -j ACCEPT` accepts everything if a space's `KUKEON-EGRESS` DROP chain is missing — so a `Default=deny` space silently allows all egress after a reboot (admission present, egress chain gone). The current design relies on **positional ordering** — `KUKEON-EGRESS` (per-space DROP) must precede `KUKEON-FORWARD` (blanket ACCEPT) in FORWARD (`internal/firewall/forward.go:40-46`) — a coupling across two independently-installed chains inside a Docker-co-owned builtin chain. This is the largest blast-radius change; land it last with ordering + deny-default tests.

## Proposal

Replace the host-global one-shot `-i k-+ ACCEPT` install with **per-space admission for that space's own bridge**, installed/removed **as one unit** with its `KUKEON-EGRESS-<space>` chain so the two can't drift. **Fail-closed:** for a `Default=deny` space, do not emit a blanket bridge ACCEPT — route that bridge's traffic only through its egress allow-list chain (terminating in DROP). A missing/half-applied egress chain then degrades to *no connectivity* (loud) rather than *unrestricted egress* (silent security hole). Collapsing per-space established+egress+admit into one space-owned chain removes the cross-chain positional dependency entirely.

## Acceptance criteria
- [ ] Per-space admission targets only that space's bridge `k-<8hex>`, not the host-global `k-+`.
- [ ] Admission and the space's `KUKEON-EGRESS-<space>` chain are installed/removed as one unit.
- [ ] `Default=deny` space emits no blanket bridge ACCEPT; bridge traffic flows only through the allow-list chain ending in DROP.
- [ ] A missing egress chain yields no connectivity (fail-closed), not unrestricted egress.
- [ ] The cross-chain positional ordering contract (`forward.go:40-46`) is removed/obsoleted.
- [ ] Tests: deny-default fail-closed behavior + chain ordering.

## Dependencies

Depends on: #1074, #1075
Blocks: none

## Notes
- #953 — umbrella (P3)
- `internal/netpolicy/rules.go` — `KUKEON-EGRESS` per-space filter (`MasterChainName`)
- `internal/controller/runner/egress.go:33` — `applySpaceEgressPolicy` / `removeSpaceEgressPolicy`
- `internal/firewall/forward.go:40-46` — the positional ordering contract this step retires

eminwux:
## Summary

Step 3 of the spacenet epic (#953). Egress was **fail-open**: the host-global blanket admission `KUKEON-FORWARD -i k-+ ACCEPT` admitted *all* kukeon-bridge egress, so a `Default=deny` space whose per-space `KUKEON-EGRESS` chain went missing (e.g. post-reboot, after the daemon re-asserts admission but before the reconcile loop re-applies the egress chain) silently allowed all egress. The fix collapses per-space established+egress+admit into **one space-owned chain** and removes the host-global egress blanket, so a missing chain degrades to *no connectivity* (loud) instead of *unrestricted egress* (silent hole).

### What changed (the diff shows the *what*; here is the *why*)

- **`internal/netpolicy/rules.go` — the per-space chain is now self-terminating.** Established + allow entries `ACCEPT` (were `RETURN`); the terminal is `DROP` (deny) / `ACCEPT` (allow) (allow was `RETURN`). Previously the chain `RETURN`ed and relied on the host-global `-i k-+ ACCEPT` to admit. With that blanket gone, a `RETURN` would drop the packet to FORWARD's policy, so egress admission must live in the chain itself. This is what makes a missing chain fail *closed*.
- **`internal/firewall/forward.go` — host-global egress blanket removed; ingress scoped to non-bridge sources.** `AdmissionRules()` now emits only the stateful-return `ACCEPT` and the ingress `! -i k-+ -o k-+ ACCEPT`; the `-i k-+ :egress` rule is gone. The `! -i k-+` qualifier (added while addressing review) scopes ingress to traffic that did *not* originate on a kukeon bridge, so a space's egress (always bridge-sourced) can never be admitted here — see the inter-bridge note below. `ensureForwardJump` no longer maintains any ordering relative to `KUKEON-EGRESS` (AC #5): egress is decided entirely inside each self-terminating per-space chain, so the jump only needs to *exist*. The pre-#1076 positional dance (`forwardJumpPositions`/`lineJumpsTo`/`parseRuleLine`) is deleted. `pruneObsoleteEgressAdmission` strips a leftover `:egress` rule from upgraded hosts, and `pruneUnscopedIngressAdmission` strips a leftover *unscoped* `-o k-+ :ingress` rule (the intermediate-#1076 shape) so the inter-bridge fail-open does not survive the upgrade either.
- **`internal/controller/runner/egress.go` + `internal/netpolicy/policy.go` — admission is installed per-space, as one unit with the chain (AC #1, #2).** `applySpaceEgressPolicy` now installs a chain for **every** space, not just those carrying an `EgressPolicy`.

### Non-obvious blast-radius decision (reviewer: please push back if this oversteps)

Removing the host-global `-i k-+` blanket means a space with **no egress policy** (or `default=allow` with an empty allowlist) would otherwise have **no path to ACCEPT** its egress on a host whose FORWARD policy is `DROP` — which is exactly the situation the admission chain exists for. So `applySpaceEgressPolicy` is no longer a no-op for those spaces: it installs a per-space **admit-all** chain (`netpolicy.NewAdmitAllPolicy` → established `ACCEPT` + terminal `ACCEPT`). This keeps their egress fully open while routing it through their own space-owned chain, so the install/remove-as-one-unit invariant holds uniformly. This is a behavioral expansion (every space now gets an iptables chain via the reconcile loop / provision path); it is required by AC #1 + the fail-closed goal, but it is the largest-radius part of the change.

### Relationship to step 2 (#1075, merged)

Step 2 added the FORWARD-jump self-assert with a *displaced-relative-to-KUKEON-EGRESS* re-claim arm. Step 3 is the AC #5 retirement of that ordering contract: the **missing → re-insert** self-assert is kept (still needed for reboot/Docker-flush recovery), but the **displaced** arm is removed because, with no egress blanket here, the jump's position relative to `KUKEON-EGRESS` no longer affects correctness (an established packet can't reach a deny-space ACCEPT it was never granted — the NEW packet was dropped, so no conntrack entry exists).

### Inter-bridge egress (review follow-up)

The reviewer flagged that the ingress rule `-o k-+ ACCEPT` would also match an **inter-bridge** packet (`-i k-A -o k-B`) — which is egress from space A *and* ingress to bridge B — and so, if `KUKEON-FORWARD` sat ahead of `KUKEON-EGRESS`, would admit a deny-space A's cross-space egress before A's chain ran. Confirmed: no inter-bridge DROP existed anywhere in the tree, so this was a real fail-open for cross-space egress, and it made the "position immaterial" claim above false for that case.

Fix (AC #5-consistent — scopes the rule rather than restoring ordering): the ingress rule is now `! -i k-+ -o k-+ ACCEPT`. It admits only traffic that did not originate on a kukeon bridge (true external ingress), so every space's egress — internet-bound and inter-bridge alike — falls through to `KUKEON-EGRESS` and its own self-terminating chain decides, regardless of position. The stateful rule cannot re-open the hole either: a denied NEW inter-bridge packet now reaches A's chain and is `DROP`ped, so no conntrack entry is ever created for the `RELATED,ESTABLISHED` rule to match. This restores the "position immaterial" claim as a genuine invariant for all three cases (internet-bound egress, external ingress, inter-bridge).

## Acceptance criteria

- [x] Per-space admission targets only that space's bridge `k-<8hex>`, not the host-global `k-+` (the dispatch `-i <bridge>` from `KUKEON-EGRESS`; the `-i k-+` blanket is removed).
- [x] Admission and the space's `KUKEON-EGRESS` chain are installed/removed as one unit (`applySpaceEgressPolicy` → `enforcer.Apply` installs chain + dispatch; `removeSpaceEgressPolicy` removes both).
- [x] `Default=deny` space emits no blanket bridge ACCEPT; bridge traffic flows only through the allow-list chain ending in DROP (`TestBuildRules_DenyChainIsSelfTerminating`, `TestAdmissionRules_NoEgressBlanket`).
- [x] A missing egress chain yields no connectivity (fail-closed), not unrestricted egress (no host-global `-i k-+ ACCEPT` remains; chain is self-terminating).
- [x] The cross-chain positional ordering contract (`forward.go:40-46`) is removed/obsoleted (`ensureForwardJump` simplified; positional helpers deleted; `TestInstall_JumpInsertedAtPositionOneWhenAbsent`, `TestInstall_JumpNoChurnWhenPresent`).
- [x] Tests: deny-default fail-closed behavior + chain ordering (firewall + netpolicy suites below).

## Test plan

- [x] `make test` (full CI unit suite: `test-root` + `test-kukebuild`) — green.
- [x] `GOWORK=off go vet ./internal/firewall/... ./internal/netpolicy/... ./internal/controller/...` — clean.
- [x] New/updated unit tests: `internal/firewall` (no-egress-blanket, ingress-excludes-bridge-sources, jump-insert-at-1, no-churn-when-present, prune-obsolete-egress, prune-unscoped-ingress, does-not-prune-scoped-ingress, prune-noop) and `internal/netpolicy` (deny self-terminating, allow-terminal ACCEPT, admit-all chain shape).
- [x] Tree-wide `git grep` sweep: no remaining `-i k-+` egress-admission rule or `kukeon-forward:egress` reference outside the prune path/comments.
- [ ] `make dev-init` (KUKEON_PROFILE=dev) end-to-end smoke — **blocked: environmental disk limit, not a code issue.** The run fails inside the BuildKit kukeond-image build (`make release-build`) with `no space left on device` writing `/root/.cache/go-build` and the `/var/lib/containerd` content-store ingest, while the host filesystem has 115G free — the build sandbox has a separate, size-constrained filesystem on this dev host. The failure is in the image-**build** phase (compiling the binary), *before* `kuke init` runs, so it never reaches the daemon-parity / firewall path this change touches. Per the disk case, this defers to disked CI rather than stalling. The firewall/netpolicy behavior is covered by the unit tests above.

Closes #1076

### DIFF
--- a/internal/controller/runner/egress.go
+++ b/internal/controller/runner/egress.go
@@ -27,27 +27,37 @@ import (
 )
 
 // applySpaceEgressPolicy realizes space.Spec.Network.Egress on the host
-// firewall. It is a no-op when the space has no egress policy or when the
-// policy effectively matches current behavior (default=allow with no
-// allowlist entries).
+// firewall as the space's own admission+egress chain.
+//
+// Since #1076 admission is per-space: the host-global `-i k-+ ACCEPT` blanket
+// is gone, so every space's bridge needs its own chain to ACCEPT egress on a
+// FORWARD-DROP host. A space with no explicit policy (or a default=allow policy
+// with an empty allowlist, which FromInternal collapses to nil) therefore still
+// gets an admit-all chain rather than a no-op — admission and the per-space
+// KUKEON-EGRESS chain are installed as one unit so they can't drift, and a
+// Default=deny space whose chain is missing fails closed (no path to ACCEPT)
+// instead of leaking egress through a blanket rule.
 func (r *Exec) applySpaceEgressPolicy(space intmodel.Space) error {
-	if space.Spec.Network == nil || space.Spec.Network.Egress == nil {
-		return nil
-	}
 	networkName, err := naming.BuildSpaceNetworkName(space.Spec.RealmName, space.Metadata.Name)
 	if err != nil {
 		return fmt.Errorf("%w: %w", errdefs.ErrEgressApply, err)
 	}
 	bridge := cni.SafeBridgeName(networkName)
 
+	var egress *intmodel.EgressPolicy
+	if space.Spec.Network != nil {
+		egress = space.Spec.Network.Egress
+	}
 	policy, err := netpolicy.FromInternal(
-		space.Spec.RealmName, space.Metadata.Name, bridge, space.Spec.Network.Egress,
+		space.Spec.RealmName, space.Metadata.Name, bridge, egress,
 	)
 	if err != nil {
 		return fmt.Errorf("%w: %w", errdefs.ErrEgressApply, err)
 	}
 	if policy == nil {
-		return nil
+		// No restriction requested: install a per-space admit-all chain so the
+		// bridge keeps full egress now that the host-global blanket is gone.
+		policy = netpolicy.NewAdmitAllPolicy(space.Spec.RealmName, space.Metadata.Name, bridge)
 	}
 	if err = policy.Resolve(r.ctx, nil); err != nil {
 		return fmt.Errorf("%w: %w", errdefs.ErrEgressApply, err)
--- a/internal/firewall/forward.go
+++ b/internal/firewall/forward.go
@@ -36,8 +36,9 @@ import (
 )
 
 // ForwardChainName is the kukeon-owned FORWARD admission chain. It now carries
-// only ingress admission (`-o k-+ ACCEPT`) plus the stateful return-traffic
-// ACCEPT; egress admission moved per-space in #1076.
+// only ingress admission (`! -i k-+ -o k-+ ACCEPT`, scoped to non-bridge
+// sources) plus the stateful return-traffic ACCEPT; egress admission moved
+// per-space in #1076.
 //
 // There is no longer any ordering contract with KUKEON-EGRESS
 // (netpolicy.MasterChainName). Egress is decided entirely within each space's
@@ -46,6 +47,16 @@ import (
 // relative to KUKEON-EGRESS is immaterial to correctness. ensureForwardJump
 // only guarantees the jump *exists* (re-inserting it after a reboot flush or
 // Docker churn); it no longer reorders it.
+//
+// The one rule here that *could* shadow an egress decision is the ingress
+// `-o k-+ ACCEPT`: an inter-bridge packet (`-i k-A -o k-B`) is both ingress to
+// B and egress from A, so a bare `-o k-+` would ACCEPT a deny-space A's egress
+// to B before A's chain runs — fail-open whenever this chain sits ahead of
+// KUKEON-EGRESS. The ingress rule is therefore scoped `! -i k-+`: it admits
+// only traffic that did *not* originate on a kukeon bridge (true external
+// ingress), so a space's egress (always bridge-sourced) can never match it and
+// always falls through to KUKEON-EGRESS. That is what keeps the position
+// genuinely immaterial — including for the inter-bridge case.
 const ForwardChainName = "KUKEON-FORWARD"
 
 // BridgeIfaceMatch is the iptables -i / -o interface match that scopes the
@@ -75,7 +86,18 @@ const commentTagPrefix = "kukeon-forward"
 // Rule order:
 //  1. RELATED,ESTABLISHED ACCEPT — return-traffic for already-admitted flows
 //     so reply packets cannot be dropped by FORWARD's default policy.
-//  2. -o k-+ ACCEPT — admit ingress destined to a kukeon bridge.
+//  2. ! -i k-+ -o k-+ ACCEPT — admit *external* ingress destined to a kukeon
+//     bridge. The `! -i k-+` qualifier scopes this to traffic that did not
+//     originate on a kukeon bridge, so it admits the internet/host→bridge case
+//     without ever matching a space's egress.
+//
+// Why the `! -i k-+` scope matters: an inter-bridge packet (`-i k-A -o k-B`)
+// is simultaneously ingress to B and egress from A. A bare `-o k-+ ACCEPT`
+// would admit it here before A's KUKEON-EGRESS chain ever runs — fail-open for
+// cross-space egress whenever this chain sits ahead of KUKEON-EGRESS in
+// FORWARD, which nothing orders post-#1076. Excluding bridge-sourced traffic
+// (`! -i k-+`) sends every space's egress — internet-bound and inter-bridge
+// alike — through KUKEON-EGRESS, so the deny-space decision is never shadowed.
 //
 // Egress admission (`-i k-+ ACCEPT`) is deliberately absent since #1076: a
 // host-global egress blanket fails *open* — a Default=deny space whose
@@ -94,6 +116,7 @@ func AdmissionRules() [][]string {
 		},
 		{
 			"-A", ForwardChainName,
+			"!", "-i", BridgeIfaceMatch,
 			"-o", BridgeIfaceMatch,
 			"-m", "comment", "--comment", commentTagPrefix + ":ingress",
 			"-j", "ACCEPT",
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #1315 Review — `feat: per-space fail-closed egress admission (single space-owned chain) (step 3)`**

The single self-terminating per-space chain + blanket removal is the right shape, and the install/remove-as-one-unit symmetry checks out (`removeSpaceEgressPolicy` is unconditional, so the new admit-all chains policy-less spaces now get are still torn down). The egress-to-internet fail-closed path is correct and well-covered. One blocking concern on the inter-bridge case.

Blocking:

- **Egress bypass (stated-invariant)** `internal/firewall/forward.go:42-46`: the docstring claims KUKEON-FORWARD's position relative to KUKEON-EGRESS is "immaterial to correctness" and that ingress/stateful "[cannot] shadow a per-space egress decision" — but a deny-default space's egress to *another* kukeon bridge (`-i k-A`, `-o k-B`) matches the ingress `-o k-+ ACCEPT` rule (`internal/firewall/forward.go:95-99`). If KUKEON-FORWARD sits ahead of KUKEON-EGRESS in FORWARD, that ACCEPT admits the packet before space A's deny chain runs — fail-open for inter-bridge egress. The deleted "displaced" self-assert previously guaranteed KUKEON-EGRESS-first; nothing now enforces the order (both insert at `-I FORWARD 1`), so it's incidental rather than contractual — the opposite of the robustness goal in a fail-closed primitive. I found no inter-bridge DROP anywhere in the tree. Please confirm cross-space egress is isolated elsewhere, or restore an ordering guarantee / scope the `-o k-+` rule so it can't admit a space's egress.

Non-blocking:

- **Test plan** `make dev-init` smoke unchecked but documented as host-blocked (disk limit, fails in the image-build phase before `kuke init`) — fine to defer to CI; not gating.

LGTM once the inter-bridge egress path is confirmed isolated or the ordering/ingress scoping is restored — the rest is clean.

eminwux:
Addressed in 81d25ef:

**Blocking — inter-bridge egress bypass.** Confirmed the concern is real: no inter-bridge DROP exists anywhere in the tree, so a deny-space's cross-space egress (`-i k-A -o k-B`) would have matched the ingress `-o k-+ ACCEPT` whenever `KUKEON-FORWARD` sat ahead of `KUKEON-EGRESS` — a fail-open, and it made the `forward.go` "position immaterial" docstring false for that case.

Fix takes your scope-the-rule option (the AC #5-consistent one — restoring an ordering guarantee would re-introduce exactly the positional contract AC #5 retires):

- `AdmissionRules()` ingress rule is now `! -i k-+ -o k-+ ACCEPT`. The negation excludes all bridge-sourced traffic, so every space's egress — internet-bound *and* inter-bridge — falls through to `KUKEON-EGRESS` and its self-terminating chain decides, independent of position. The stateful rule can't re-open it: a denied NEW inter-bridge packet now reaches the deny-chain and is DROPped, so no conntrack entry is created for `RELATED,ESTABLISHED` to match.
- `pruneUnscopedIngressAdmission` (new, parallel to `pruneObsoleteEgressAdmission`) strips a leftover unscoped `-o k-+ :ingress` rule on hosts upgraded from the intermediate #1076 layout, so the fix actually holds on the upgrade path rather than leaving the old fail-open rule alongside the new scoped one.
- Docstrings on `ForwardChainName` / `AdmissionRules` / `ensureForwardJump` updated to state *why* the position is now genuinely immaterial (the `! -i k-+` scope), so the claim is no longer aspirational.

Tests: added `TestAdmissionRules_IngressExcludesBridgeSources` (the regression guard for this concern), `TestInstall_PrunesUnscopedIngressAdmission`, and `TestInstall_DoesNotPruneScopedIngress`; updated `TestAdmissionRules_Order` / `TestAdmissionRules_NoEgressBlanket` (now distinguishes a positive `-i k-+` blanket from the negated `! -i k-+` scope) and the canned `-S` fixtures. `make test` green (re-run post-rebase onto current `origin/main`); `go vet` clean.

**Non-blocking — `make dev-init` smoke** still deferred to CI: unchanged host disk limit in the image-build phase (before `kuke init`), documented in the PR body. The firewall behavior is covered by the unit tests above.

eminwux:
**PR #1315 Review — `feat: per-space fail-closed egress admission (single space-owned chain) (step 3)`**

The fail-open→fail-closed collapse is sound and thoroughly covered. The self-terminating per-space chain (`internal/netpolicy/rules.go` — established/allow `ACCEPT`, terminal `DROP`/`ACCEPT`), the blanket removal, and the `! -i k-+` ingress scope (`internal/firewall/forward.go:119`) hold together across all three cases (internet egress, external ingress, inter-bridge): bridge-sourced traffic always falls through to `KUKEON-EGRESS`, and ingress reply traffic is caught by the stateful `ACCEPT` in `KUKEON-FORWARD`. Both prune paths `-D`-by-spec so the scoped ingress rule is correctly preserved, the deleted positional helpers have zero remaining refs, `Resolve` is nil-`Allow`-safe, and `removeSpaceEgressPolicy` tears down the new admit-all chains by name. No secrets in errors; sentinels wrapped with `%w`.

One non-blocking operational note:

- **Upgrade window** `internal/controller/runner/egress.go:40`: on upgrade, `Install` removes the host-global blanket immediately while each running space's per-space chain is (re)installed only on the next reconcile/provision tick — so existing spaces can lose egress briefly until their chain lands. This is the intended fail-closed cost (loud no-connectivity, self-heals on reconcile), not a defect; flagging it only so it's a known operator-visible effect on the first post-upgrade `kuke init`.

`make dev-init` (KUKEON_PROFILE=dev) is the only unchecked test-plan item, documented as blocked by an environmental disk limit in the BuildKit image-build phase before `kuke init` runs — host-gated, so non-blocking here. Worth running pre-merge on a disked host since this change touches the daemon-read firewall path.

LGTM to proceed.

PR #1318, opened 2026-06-15
feat: add Degraded cell state for cells with a down or restarting workload

### LINKED ISSUE #1317: Cell reports Ready while a non-root workload is down (`Ready 1/2`); add a Degraded state
## Context

A cell whose root/sandbox is up but whose non-root workload has crashed (or is mid-restart) is reported `Ready` even though `kuke get cell -o wide` shows `1/2` containers ready. The `STATE` and the `READY` columns are computed from **different sources**, so they contradict each other:

- `STATE = Ready` — `deriveCellStateFromNonRootContainerStatuses` (`internal/controller/runner/refresh.go`): when any non-root workload is active the cell is `Ready`, ignoring a crashed sibling; and while a single workload is down-but-restarting the restart pass held the cell at its pre-derivation `Ready` (to dodge the sticky `Error` short-circuit).
- `READY = 1/2` — `renderContainers` (`cmd/kuke/get/cell/cell.go:204`): counts every container whose state is `Ready`.

An operator reading `Ready` reasonably assumes all workloads are healthy; `1/2` directly contradicts that. Observed live while testing PR #1308 (restart-on-exit) with a crash-looping workload.

There is no cell state today that expresses "root up, a workload down/restarting" — the enum jumps from `Ready` straight to the terminal/sticky `Error`. Making a crashed-workload cell go `Error` is not viable: `Error` is **sticky** (`cellStateIsSticky`) and short-circuits `ReconcileCell`, which would strand PR #1308's restart loop.

## Proposal

Add a non-sticky, non-terminal `CellStateDegraded` (k8s "phase vs readiness" split) and derive it in the two sites where partial health arises:

1. **Union derivation** (`deriveCellStateFromNonRootContainerStatuses`): some workload active **and** another terminal in a non-clean way (crashed, or exited under a policy that should keep it running) → `Degraded`. A cleanly-completed oneshot (`never`/`on-failure` + exit 0) is **not** degraded (sidecar+job cell stays `Ready`).
2. **Restart pass** (`ReconcileCell`): while a restart is owed (fired/deferred, budget not exhausted) hold the cell at `Degraded` instead of the sticky `Error`; once the `on-failure` cap exhausts, let `Error` settle.

`Degraded` must be non-sticky (stays re-derivable → returns to `Ready` on recovery, settles to `Error` once a crash-looper gives up) and must latch `ReadyObserved` / count as a live cell for volume-in-use and liveness gates.

## Acceptance criteria

- [ ] New `CellStateDegraded` appended last in both enums (ordinal lockstep for the scheme int-cast), with `String()`/parse/`assignCellStateInt` bound and round-trip coverage.
- [ ] Union derivation returns `Degraded` for active+non-clean-terminal; stays `Ready` for active+completed-oneshot.
- [ ] Restart pass holds the cell at non-sticky `Degraded` while a restart is owed; `Error` still settles when the restart budget exhausts.
- [ ] `Degraded` is non-sticky (`cellStateIsSticky` false), does not auto-delete (`cellStateAutoDeleteTriggers` false), latches `ReadyObserved`, and is treated as live by `VolumeMountedByLiveCell` / `IsCellRunning` / the `kuke run`/`restart`/`attach` routing.
- [ ] `docs/site/manifests/cell.md` documents the new state.

## Out of scope

- Persisted restart accounting (`restartCount`/`restartTime`) — that is #1234.
- User-tunable backoff/cap knobs — #1235.

## References

- Surfaced during PR #1308 (restart-on-exit) testing.
- `internal/controller/runner/refresh.go` (derivation + restart pass + `latchReadyObserved`), `cmd/kuke/get/cell/cell.go:204` (`renderContainers`).

eminwux:
## Summary

- Adds a non-sticky, non-terminal `CellStateDegraded` so a cell whose root/sandbox is up but whose non-root workload is **down or restarting** no longer reports a contradictory `Ready 1/2`. Surfaced while testing #1308 (restart-on-exit) with a crash-looping workload.
- `STATE` (derived from the root/union) and the `READY n/m` count (`renderContainers`) read different sources, so a crashed-but-live-sibling cell showed `Ready` while `1/2` said otherwise. `Degraded` is the k8s-style "phase vs readiness" rung the enum was missing.

## Why a new state (and why non-sticky)

Making a crashed-workload cell go `Error` is not viable: `Error` is **sticky** (`cellStateIsSticky`) and short-circuits `ReconcileCell`, which would strand #1308's restart loop. `Degraded` is deliberately **non-sticky / non-terminal**: it stays re-derivable, returns to `Ready` on recovery, and settles to the sticky `Error` only once an `on-failure` crash-looper exhausts its restart budget.

## What changed

- **Enum**: `CellStateDegraded` appended **last** in both `internal/modelhub/cell.go` and `pkg/api/model/v1beta1/cell.go` (ordinal lockstep for the `apischeme` direct int-cast), plus `String()`, `parseCellState`, and the `assignCellStateInt` upper bound + `StateDegradedStr`.
- **Derivation, mixed case** (`deriveCellStateFromNonRootContainerStatuses`): some workload active **and** another terminal-non-clean (crashed, or exited under a policy that should keep it running) → `Degraded`. A cleanly-completed oneshot (`never`/`on-failure` + exit 0) stays `Ready` (sidecar+job pattern). New shared predicate `nonRootWorkloadDegraded`.
- **Restart pass** (`ReconcileCell`): while a restart is owed, `restartFired`/`restartDeferred` now hold the cell at non-sticky `Degraded` (clearing the crash breadcrumb) instead of the old "restore pre-derivation `Ready`" hack; `restartFired` persists it explicitly. `restartNone` (cap exhausted / no restart owed) still lets the sticky `Error` settle.
- **Operator recovery routing** (`controller.StartCell`, fixed in review round 2): `Degraded` joins `Error`/`Failed` in the recovery `switch` so a `kuke start`/`kuke restart` of a Degraded cell routes through `RecreateCell` → `markCellReady` and lands `Ready` in **one** action (see review fix-up below).
- **Liveness/latch gates** (correctness, not cosmetic):
  - `latchReadyObserved` latches on `Degraded` — a cell whose **first** observed state is `Degraded` (sibling crashed before it ever derived `Ready`) has still come up; without this the restart pass's `ReadyObserved` gate would skip it and never restart the sibling.
  - `VolumeMountedByLiveCell` treats `Degraded` as live, so `kuke delete volume` won't yank a mount out from under a still-running container.
  - `IsCellRunning` counts `Degraded` as running.
- **CLI routing** (exhaustive-switch members): `kuke run` / `kuke restart` / `kuke attach` liveness (`cell_liveness.go`) groups `Degraded` with `Ready` (a live, attachable cell). The daemon-side `StartCell` recovery switch routes `Degraded` to recovery (above) — distinct decision from CLI-side liveness. `kuke restart`'s own state-machine (`restart.go`) routes a `Degraded` cell through the no-stop-first recovery path (`restartStopped`), not the in-place stop+start `Ready` takes (fixed in review round 3 — see below).
- **Docs**: `docs/site/manifests/cell.md` state table; `docs/site/manifests/container.md` restart-policy section (restart-on-exit truth table, restart timing, policy → cell-`STATE` table).

## Review fix-up (round 2) — a Degraded cell now recovers to Ready in one action

Addresses the blocking review on this PR (live-tested by the reviewer, deterministic 3/3; folds in #1320/#1321). Before this fix-up, a single `kuke restart` / `kuke stop`+`start` of a `Degraded` cell relaunched every container (3/3 Ready) yet left the cell stuck at sticky **`Error`** with `CONTAINERS 3/3` — the inverse of the `Ready 1/2` contradiction this PR set out to kill — and only a *second* restart recovered.

Root cause: the recovery `switch` at `internal/controller/start_cell.go` routed only `Error`/`Failed` through `RecreateCell` → `markCellReady` (which clears the sticky breadcrumb). `Degraded` fell through to the regular start path, where a live root + live sidecar trip the running-container guard and the daemon's re-derivation from the preserved non-zero workload exit re-stamps a sticky `Error` the plain start never clears.

Fix (the reviewer's option 1): add `CellStateDegraded` to that recovery `switch` so a Degraded cell recreates + `markCellReady` exactly like `Error`/`Failed` — the same path the reviewer's table confirms already recovers a cell cleanly (the 2nd restart). A single operator action now lands `Ready`.

Also adds the restart-policy / state-transition docs (`container.md`) the review asked for (folded in from #1320).

## Review fix-up (round 3) — `kuke restart` of a Degraded cell now recovers in one action too

The round-2 daemon-side fix recovered `kuke start` and `kuke stop`+`start`, but the reviewer's re-test of `4b70bac` found `kuke restart` still landed in a stuck `Error 3/3` (deterministic 3/3). The miss was CLI-side, not in `StartCell`: `kuke restart`'s state-machine pre-filter at `cmd/kuke/restart/restart.go` grouped `CellStateDegraded` with `Ready` → `restartInPlace` (stop-first then start). The stop-first flips the cell out of `Degraded` before the daemon's `StartCell` re-reads it, so the now-fixed Degraded recovery arm never fires and the relaunch leaves the cell stuck at sticky `Error N/N` — the same stop-first-defeats-recovery hazard already documented for `Error`/`Failed` at `restart.go:237-239`.

Fix (the reviewer's option 1): move `CellStateDegraded` out of the `Ready` case into the terminal/recovery case so it routes through `restartStopped` (calls `StartCell` directly, no stop-first), letting the daemon's recovery switch recreate it to `Ready` in a single action — exactly like `kuke start`, which the reviewer confirmed already recovers. Comments updated to record Degraded as a recovery member.

## Follow-up (round 4) — `kuke run` of a Degraded cell now recovers too (start+attach parity)

**This reverses the Degraded `kuke run` no-op that the prior reviews accepted, per maintainer direction.** Rounds 2–3 left `kuke run <degraded-cell>` as an intentional Ready-style no-op (`runExistingCell` grouped `Degraded` with `Ready`), on the rationale that `run` should not re-enter `StartCell` on a live cell (#630 CNI duplicate-allocation). The maintainer's call: `kuke run <existing-cell>` is documented as "**start + attach** an existing cell", and a Degraded cell is a startable state that `kuke start` and `kuke restart` already recover — so `run` leaving it un-recovered breaks the run/start/restart parity invariant. A user running `kuke run` on a degraded cell expects it brought back up, not silently no-op'd.

Fix: move `CellStateDegraded` out of the `Ready` case into the Stopped/Exited/Error/Failed StartCell group in `cmd/kuke/run/run.go`, so `run` calls `StartCell` directly (no stop-first) and the daemon recreates it to `Ready` — identical CLI shape to the other terminal states.

Why the #630 concern does **not** apply here (the original no-op rationale): the daemon's `StartCell` routes `Degraded` through the same `RecreateCell` path as `Error`/`Failed` (round 2), which tears the containers down — **releasing the CNI allocation** — before recreating. The duplicate-allocation re-entry only bites the *plain* `Ready` start path, which `Degraded` no longer takes. The Ready no-op (and its #630 guard) is preserved unchanged. `kuke attach` still treats Degraded as a live, attachable cell (`GuardCellTaskLiveness`) for the inspect-without-recreate path — that did not change.

Doc follow-up filed as #1322 (pm's spec-doc lane): `docs/cli-use-cases.md` enumerates the startable-state set as Stopped/Exited/Error/Failed and does not yet list `Degraded` now that all three verbs recover it — a doc-only change a code PR can't carry.

## Non-obvious calls (please push back)

- **Behavior flip, intentional**: the existing derivation test `active_workload_outranks_a_failed_sibling` asserted active+crashed-sibling → `Ready`. That was the masking behavior this PR fixes; the case is renamed and re-asserted to `Degraded`. Called out so it reads as a deliberate change, not an accidental break.
- **`Degraded` is now a recovery state in `StartCell`, not a fall-through** (reversed in review round 2). The daemon's `StartCell` recovery switch now groups `Degraded` with `Error`/`Failed` (recreate), while the CLI-side liveness switches (`cell_liveness.go`, attach) still group it with `Ready` (a live, attachable cell). These are different decisions: a Degraded cell is live enough to attach, but an explicit start/restart recovers it via recreate.
- **`ReadyAt` is not set on `Degraded`** (only on true `Ready`) — a never-fully-`Ready` cell keeps `ReadyAt` zero by design; `ReadyObserved` (the functional gate) does latch. Minor asymmetry, called out.

## Test plan

- [x] `make test` (full CI target: `test-root` + `test-kukebuild`) — exit 0, no failures.
- [x] `GOWORK=off go test ./internal/controller/ ./internal/controller/runner/ ./pkg/api/model/v1beta1/ ./cmd/kuke/...` — pass, incl. the new derivation rows (mixed→Degraded, completed-oneshot→Ready), `nonRootWorkloadDegraded` truth table, `cellStateIsSticky`/`cellStateAutoDeleteTriggers`/`latchReadyObserved` Degraded rows, marshal round-trip + int bound, `TestReconcileCell_RestartHoldsCellDegraded`, and the new recovery-routing tests below.
- [x] **Regression test** (review round 2): `TestStartCell_DegradedRecoversToReadyInOneAction` + a `TestStartCell_ReadyStateValidation` row asserting a Degraded cell (live root + sidecar, crashed job) routes through `RecreateCell` and lands `Ready`. Confirmed it catches the bug: reverting the one-line `switch` change makes both fail with `has running containers and must first be stopped` (the exact pre-fix path).
- [x] **Regression test** (review round 3): `TestRestartCmd/degraded cell: restarts via StartCell (no stop-first)` asserts a `Degraded` cell routes through `restartStopped` (StartCell only, `stop=0 start=1`), not the in-place stop+start. Confirmed it catches the bug: reverting only the `restart.go` switch change (keeping the test) makes it fail with `stop=1 start=1` — the in-place path that defeats recovery.
- [x] **Regression test** (follow-up round 4): `TestRun_ExistingCell_TerminalStates_StartsAndAttaches/degraded` asserts `kuke run` on a `Degraded` cell routes through a single `StartCell` RPC (no `CreateCell`), same as the other terminal states. Confirmed it catches the change: reverting only the `run.go` switch (keeping the test) makes the `degraded` subtest fail (it falls into the Ready no-op, `StartCell` count `0`).
- [x] **On-device** (follow-up round 4) `Degraded → run → Ready`: rebuilt `kuke` from this commit, drove `degraded-repro` (`root` + `sidecar(always)` + `job(never)`) to stable `Degraded` by SIGKILLing `job`, then `kuke run -d degraded-repro` → output flips from `containers: not started` (the old Ready no-op) to `containers: started` and the cell lands `Ready 3/3` in one action. 2/2 deterministic. A still-`Ready` cell's `run` no-op (`containers: not started`, #630-protected) verified unchanged.
- [x] `gofmt` / `go vet ./internal/controller/...` clean; `pre-commit run --files <changed>` — golangci-lint + prettier pass (prettier reformatted the new `container.md` tables).
- [x] `make dev-init` live smoke / on-device `Degraded → restart → Ready` transition — **now run on-device** (default-profile nested daemon, `/run/kukeon-dev/kukeond.sock`). Built the daemon from `8538895`, created a `root` + `sidecar(always)` + `job(never)` cell, SIGKILLed `job` → cell derived stable `Degraded` (root/sidecar Ready, job Error). Recovery matrix, all landing `Ready 3/3` in **one** action: `kuke restart` 3/3 deterministic (each prints `Started cell …` — the `restartStopped`/no-stop-first path, the round-3 fix), `kuke start` ✅, `kuke stop`+`start` ✅. No stuck `Error 3/3` on any verb. See the verification comment below for the full trace.

## Sequencing

Branched off #1308 (now merged to `main`); base auto-retargeted to `main`.

Closes #1317


eminwux:
**PR #1318 Review — `feat: add Degraded cell state for cells with a down or restarting workload`**

LGTM — clean, non-sticky `CellStateDegraded` that resolves the `Ready 1/2` STATE-vs-readiness contradiction. Verified the load-bearing invariants:

- Both enums append `Degraded` **last** in identical ordinal order (`internal/modelhub/cell.go:217`, `pkg/api/model/v1beta1/cell.go:270`) — lockstep for the apischeme int-cast; `assignCellStateInt` bound, `String()`, `parseCellState`, and the marshal round-trip/int-bound test all track it.
- `nonRootWorkloadDegraded` (`internal/controller/runner/refresh.go:1622`) is consulted only inside the terminal-state `case`, and `Degraded` is returned only in the `anyActive` partial-health branch — an all-down set still settles terminal `Error`.
- `latchReadyObserved` runs (`refresh.go:781`) before the restart gate reads `ReadyObserved`, so the `restartDeferred` reset to `originalStatus` is harmless (re-latched on `newState==Degraded` next tick); `restartFired` persists non-sticky `Degraded` and clears the crash breadcrumb.
- Every `CellStateReady` gating site is either grouped with `Degraded` (IsCellRunning, run/restart/attach, GuardCellTaskLiveness, VolumeMountedByLiveCell, latchReadyObserved, start_cell fall-through) or correctly excludes it (sticky / auto-delete / ReadyAt / rematerialize predicates). All commits signed; derivation rows, truth table, and the `RestartHoldsCellDegraded` integration test cover the change.

Non-blocking, for your pre-merge checklist:
- The `make dev-init` live smoke is unchecked (host-gated, documented unrunnable here) — the runtime `Ready 1/2 → Degraded` transition is covered by the `ReconcileCell` integration test, but worth a confirm on a clean host.
- Sequencing: based on `feat/reconciler-restart-on-exit` (#1308) — merge after #1308 lands and GitHub auto-retargets to `main`.

eminwux:
## Changes requested — Degraded cells can't be cleanly recovered by an operator

Live-tested this build (`feat/degraded-cell-state` @ `fce2a76`, dev profile, `reconcileInterval: 5s`). The new `Degraded` state derives correctly and the restart-on-exit matrix (`always` / `on-failure` / `never`) behaves as specified. But the recovery path out of `Degraded` is broken, and it should be fixed before merge since `Degraded` is what this PR introduces.

**Filed as #1321** (verified, deterministic 3/3 trials).

### Repro
Cell with `root` + `sidecar` (`restartPolicy: always`) + `job` (`restartPolicy: never`); kill `job` → cell goes to stable `Degraded` (correct). Then try to recover:

| Recovery verb on the `Degraded` cell | Containers after | Cell `STATE` after |
| --- | --- | --- |
| `kuke restart` (1st) | 3/3 Ready | **`Error`** (stuck) |
| `kuke restart` (2nd) | 3/3 Ready | `Ready` |
| `kuke stop` + `kuke start` | 3/3 Ready | **`Error`** (stuck) |
| `kuke run -f` | unchanged | `Degraded` (no-op) |

### The problem
A single `kuke restart` (or `stop`+`start`) relaunches the workload — all three containers go `Ready/0` — but the cell `STATE` lands on sticky **`Error`** with `CONTAINERS 3/3`. That's a STATE-vs-readiness contradiction (the inverse of the `Ready 1/2` this PR set out to kill), and because `Error` is sticky (`cellStateIsSticky`, `refresh.go:987`) it short-circuits `ReconcileCell` and never self-heals. It takes a **second** restart to reach `Ready`. `kuke run -f` can't recover it at all.

### Fixes needed
1. **Make a single operator recovery of a `Degraded` cell land in `Ready`.** Root cause candidate: the recovery `switch` at `internal/controller/start_cell.go:103-121` routes only `Failed`/`Error` through `RecreateCell` → `markCellReady` (which clears the sticky breadcrumb); `Degraded` isn't a member, so the stop phase derives+persists sticky `Error` from `job`'s preserved exit 137 and the start phase relaunches containers but doesn't clear it. Either:
   - add `Degraded` to that recovery `switch` so it recreates+markCellReady like `Error`/`Failed`, **or**
   - on the start/recreate path, clear a sticky `Error`/`Failed` once the post-recreate snapshot shows every non-root container `Ready` (so the cell can never persist `Error N/N`).
2. **Add a regression test** for the Degraded → `restart` → `Ready` transition (single cycle, no intermediate stuck `Error`), alongside the existing `ReconcileCell` restart-hold tests.

Everything else in the PR looks good; this is the one blocker. See #1321 for the full trace.

eminwux:
## ⛔ Blocking changes — fix all of the following in this PR before re-review

Consolidating the live-test findings here so this PR is the single source of truth. The standalone issues I filed (#1320, #1321) are being **closed in favor of fixing them in this PR**. Three required changes:

### 1. Bug: a `Degraded` cell can't be cleanly recovered by a single operator action (blocker)
Live-tested `feat/degraded-cell-state` @ `fce2a76` (dev profile, `reconcileInterval: 5s`), deterministic across 3/3 trials. Cell with `root` + `sidecar` (`restartPolicy: always`) + `job` (`restartPolicy: never`); kill `job` → stable `Degraded` (correct). Recovery then fails:

| Recovery verb on the `Degraded` cell | Containers after | Cell `STATE` after |
| --- | --- | --- |
| `kuke restart` (1st) | 3/3 Ready | **`Error`** (stuck) |
| `kuke restart` (2nd) | 3/3 Ready | `Ready` |
| `kuke stop` + `kuke start` | 3/3 Ready | **`Error`** (stuck) |
| `kuke run -f` | unchanged | `Degraded` (no-op) |

A single `restart`/`stop`+`start` relaunches the workload (all containers `Ready/0`) but the cell `STATE` lands on sticky **`Error`** with `CONTAINERS 3/3` — a STATE-vs-readiness contradiction (the inverse of the `Ready 1/2` this PR set out to kill). Because `Error` is sticky (`cellStateIsSticky`, `refresh.go:987`) it short-circuits `ReconcileCell` and never self-heals; only a **second** restart recovers.

**Root cause candidate:** the recovery `switch` at `internal/controller/start_cell.go:103-121` routes only `Failed`/`Error` through `RecreateCell` → `markCellReady` (which clears the sticky breadcrumb); `Degraded` isn't a member, so the stop phase derives+persists sticky `Error` from `job`'s preserved exit 137 and the start phase relaunches containers but doesn't clear it.

**Fix:** either add `Degraded` to that recovery `switch` so it recreates + `markCellReady` like `Error`/`Failed`, **or** clear a sticky `Error`/`Failed` on the start/recreate path once the post-recreate snapshot shows every non-root container `Ready` (so the cell can never persist `Error N/N`).

### 2. Regression test
Add a test for the single-cycle `Degraded → kuke restart → Ready` transition (no intermediate stuck `Error 3/3`), alongside the existing `ReconcileCell` restart-hold tests in `internal/controller/runner/`.

### 3. Docs: restartPolicy runtime behavior + state transitions (folded in from #1320)
`docs/site/manifests/container.md` § "Restart policy" (this PR's neighbor) frames `restartPolicy` only around the wind-down/reap gate. Since this PR already touches `docs/site/manifests/cell.md`, extend the docs to cover:
- the restart-on-exit truth table (`restartPolicyRequiresRestart`, `refresh.go:1224`): `always` → any exit; `on-failure` → non-zero only; `never`/empty → never;
- restart timing — `restartBackoff` 30s (`refresh.go:1179`), `onFailureMaxRestarts` 5 (`refresh.go:1189`), cap-exhaustion → sticky `Error`; reconcile-interval gating;
- a policy → cell-`STATE` table (`Ready` / `Degraded` / `Error`), cross-linked with the `Degraded` definition in `cell.md`.

The behavior matrix verified in testing:

| `restartPolicy` | killed workload | restarted? | cell `STATE` settled |
| --- | --- | --- | --- |
| `always` | exit 137 | yes | `Ready` (transits `Degraded`) |
| `on-failure` | exit 137 (non-zero) | yes | `Ready` (transits `Degraded`) |
| `never` | exit 137 | no | `Error` (sticky) |
| mixed (sidecar up + crashed never-job) | — | n/a | `Degraded` (stable) |

Everything else in the PR tested clean — the `Degraded` derivation and the restart-on-exit matrix work as specified. These three are the changes needed before merge.

eminwux:
Addressed in 4b70bac:

1. **Bug — Degraded cell can't be cleanly recovered by one operator action.** Added `CellStateDegraded` to the recovery `switch` in `internal/controller/start_cell.go` (your option 1) so a `kuke start`/`kuke restart` of a Degraded cell routes through `RecreateCell` → `markCellReady` exactly like `Error`/`Failed` — the path your table confirms already recovers a cell on the 2nd restart. A single action now lands `Ready` instead of a stuck `Error 3/3`. Updated both recovery-switch doc-comments (Degraded is now a recovery member, no longer fall-through).

2. **Regression test.** `TestStartCell_DegradedRecoversToReadyInOneAction` asserts RecreateCell is taken (not a plain StartCell) and the result is `Ready`; plus a `TestStartCell_ReadyStateValidation` row for a Degraded cell with a live root+sidecar and a crashed job. Verified they catch the bug: reverting the one-line switch change makes both fail with `has running containers and must first be stopped` — the exact pre-fix path (the live root/sidecar trips the running-container guard once Degraded falls through).

3. **Docs.** Extended `docs/site/manifests/container.md` § Restart policy with a "Restart on exit" subsection: the restart-on-exit truth table (`always`→any exit, `on-failure`→non-zero only, `never`/empty→never), restart timing (30s `restartBackoff`, 5-attempt `onFailureMaxRestarts` cap → sticky `Error`, reconcile-tick gating), and a policy → cell-`STATE` table cross-linked to the `Degraded` definition in `cell.md`. (Folds in #1320.)

**Deferred / not run here:** the `make dev-init` live smoke and the on-device `Degraded → restart → Ready` runtime transition — this host has no live daemon. Per the behavioral-fix verification bar I've left that checkbox unticked: the fix is verified at the routing/unit level (the regression test, confirmed against a revert), not by an on-device trace. You have the deterministic repro env (`reconcileInterval: 5s`, dev profile) and are best placed to confirm the end-to-end recovery.

Note: #1308 has merged, so the base auto-retargeted to `main`.

eminwux:
## Re-tested `4b70bac` — fix is partial; `kuke restart` still lands in stuck `Error 3/3`

Rebuilt the daemon from `4b70bac` and re-ran the recovery matrix (dev profile, `reconcileInterval: 5s`). The daemon-side `StartCell` fix works, and the new unit tests + `internal/controller`, `internal/controller/runner`, `pkg/api/model/v1beta1` packages all pass. But one recovery verb is still broken — deterministically, 3/3 trials.

| Recovery verb on a stable-`Degraded` cell | Cell `STATE` after (1 action) | Result |
| --- | --- | --- |
| `kuke start` | `Ready` 3/3 | ✅ fixed |
| `kuke stop` + `kuke start` | `Ready` 3/3 | ✅ fixed |
| **`kuke restart`** | **`Error` 3/3 (stuck)** | ❌ **still broken** |
| `kuke run -f` | `Degraded` (no-op) | n/a (not a recovery verb) |

### Root cause — the miss is CLI-side, not in `StartCell`
`4b70bac` correctly added `Degraded` to the daemon's `StartCell` recovery switch, which is why `kuke start` (and the second command of `stop`+`start`) now recover. But `kuke restart` never reaches that arm: the CLI pre-filter in `cmd/kuke/restart/restart.go:225` groups `CellStateDegraded` **with `Ready`** → `restartInPlace`, which does **stop-first then start**. The stop-first flips the cell out of `Degraded` (to a transient `Stopped`/`Error`) before `StartCell` re-reads it, so the Degraded recovery arm never fires and the relaunch leaves the cell stuck at `Error N/N` — the exact same stop-first-defeats-recovery hazard the sibling comment at `restart.go:237-239` already calls out for `Error`/`Failed`.

### Fix needed (one-line routing change + the symmetric Error/Failed handling already there)
In `cmd/kuke/restart/restart.go`, move `v1beta1.CellStateDegraded` out of the line-225 `Ready` case and into the line-230 terminal/recovery case so it routes through `restartStopped` (calls `StartCell` directly, **no stop-first**) — letting the daemon's now-fixed recovery switch recreate the Degraded cell to `Ready` in a single action. This mirrors the rationale already documented for `Error`/`Failed` at `restart.go:237-239` ("a stop-first would flip it to Stopped and skip the recovery"). A `kuke restart` regression test for the `Degraded → Ready` single-action transition should accompany it (the existing `start_cell_test.go` covers the daemon path but not the CLI routing).

Everything else from the prior round is resolved — docs and the `StartCell` recovery both look good. This last routing gap is the only remaining blocker.

eminwux:
Addressed in 8538895 (review round 3):

**`kuke restart` of a Degraded cell now recovers in one action.** You correctly traced the remaining miss to the CLI side, not `StartCell`. The state-machine pre-filter at `cmd/kuke/restart/restart.go` grouped `CellStateDegraded` with `Ready` → `restartInPlace` (stop-first then start); the stop-first flips the cell out of `Degraded` before the daemon's `StartCell` re-reads it, so the round-2 recovery arm never fired.

Fix (your option 1): moved `CellStateDegraded` out of the line-225 `Ready` case into the terminal/recovery case so it routes through `restartStopped` (calls `StartCell` directly, no stop-first). The daemon now sees the persisted `Degraded` state and takes the recreate → `markCellReady` path — same as `kuke start`, which your re-test confirmed already recovers. Updated the switch doc-comments to record Degraded as a recovery member (mirrors the existing `Error`/`Failed` rationale at `restart.go:237-239`).

**Regression test:** `TestRestartCmd/degraded cell: restarts via StartCell (no stop-first)` asserts the Degraded cell routes through `restartStopped` (`stop=0 start=1`). Confirmed it catches the bug — reverting only the `restart.go` switch change (keeping the test) makes it fail with `stop=1 start=1`, the in-place path that defeats recovery.

`make test` (full CI target) exit 0; gofmt / go vet / pre-commit (golangci-lint) clean.

**Deferred / not run here:** the `make dev-init` live smoke and the on-device `Degraded → kuke restart → Ready` runtime transition — this host has no live daemon. Verified at the routing/unit level (the regression test, confirmed against a revert), not by an on-device trace. You have the deterministic repro env (`reconcileInterval: 5s`, dev profile) and are best placed to confirm the end-to-end recovery.

eminwux:
## On-device verification — `Degraded → recovery → Ready` confirmed live (round-3 fix)

Ran the end-to-end smoke I'd previously deferred. Built the daemon from `8538895` (default-profile nested run, daemon socket `/run/kukeon-dev/kukeond.sock`), created the reviewer's repro cell — `root` + `sidecar` (`restartPolicy: always`) + `job` (`restartPolicy: never`) — then `SIGKILL`ed the `job` task.

**Derivation:** after the reconcile tick the cell settled to a **stable `Degraded`** (`root` Ready, `sidecar` Ready, `job` Error), confirmed stable across repeated polls. ✅

**Recovery matrix** (each verb on a freshly-degraded cell, first-settled state, no second action):

| Recovery verb | CLI output | First-settled state |
| --- | --- | --- |
| `kuke restart` (×3 trials) | `Started cell "…"` | **`Ready` 3/3** ✅ (3/3 deterministic) |
| `kuke start` | `Started cell "…"` | `Ready` 3/3 ✅ |
| `kuke stop` + `kuke start` | `Stopped …` / `Started …` | `Ready` 3/3 ✅ |

The decisive signal for the round-3 fix: `kuke restart` now prints **`Started cell`** (not `Restarted cell`), proving it routes through `restartStopped` (StartCell directly, no stop-first) so the daemon observes the persisted `Degraded` and takes the recreate→`markCellReady` recovery. No verb left the cell stuck at `Error 3/3` — the pre-fix failure mode is gone.

Flipped the `make dev-init` / on-device checkbox in the test plan to `[x]` accordingly. Test cell purged after the run.

> Note: nested-on-default-profile is the host I had (the existing `/opt/kukeon` instance is default-profile); `make dev-init` emitted its expected namespace/cgroup-collision warning but the recovery behavior under test is profile-independent.

eminwux:
Follow-up in 6a210ce (per maintainer direction) — `kuke run` of a `Degraded` cell now recovers it (start + attach) instead of no-op'ing.

**This reverses the Degraded `run` no-op the prior review rounds accepted.** Rounds 2–3 settled `kuke run <degraded>` as a Ready-style no-op; the maintainer's call is that `kuke run <existing-cell>` is documented as "start + attach an existing cell", and a Degraded cell is a startable state `kuke start`/`kuke restart` already recover — so leaving `run` as a no-op broke the run/start/restart parity invariant.

What changed (`cmd/kuke/run/run.go`): moved `CellStateDegraded` out of the `Ready` case into the Stopped/Exited/Error/Failed `StartCell` group, so `run` calls `StartCell` directly (no stop-first) and the daemon's `RecreateCell` path lands it `Ready` in one action — same CLI shape as the other terminal states.

Why the original #630 no-op rationale doesn't apply: the daemon routes `Degraded` through the same `RecreateCell` path as `Error`/`Failed` (round 2), which tears containers down (releasing the CNI allocation) before recreating. The duplicate-allocation re-entry only bites the *plain* `Ready` start, which `Degraded` no longer takes. The `Ready` no-op and its #630 guard are preserved unchanged; `kuke attach` still treats `Degraded` as a live, attachable cell.

Verification:
- `TestRun_ExistingCell_TerminalStates_StartsAndAttaches/degraded` (new row) — `run` on `Degraded` → single `StartCell`, no `CreateCell`. Confirmed it catches the change (revert `run.go` only → degraded subtest fails into the Ready no-op).
- `make test` (full CI: `test-root` + `test-kukebuild`) — exit 0, 0 failures.
- On-device: drove `degraded-repro` to stable `Degraded` (SIGKILL `job`), then `kuke run -d` → output flips `containers: not started` → `containers: started`, cell `Ready 3/3` in one action, 2/2 deterministic. Still-`Ready` `run` no-op verified unchanged.

eminwux:
**PR #1318 Review — `feat: add Degraded cell state for cells with a down or restarting workload`** (second reviewer)

LGTM — independently traced the enum lockstep (`Degraded` appended last in both `modelhub` and `v1beta1` in identical order, `assignCellStateInt` bound + marshal/parse updated), the `anyActive && anyDegraded` derivation gate (clean oneshots correctly stay `Ready`), and swept every `CellStateReady` comparison in the tree — all six sites are either updated for liveness (`IsCellRunning`, `VolumeMountedByLiveCell`, `GuardCellTaskLiveness`, `latchReadyObserved`) or intentionally Ready-only (`metadata.go:202` ReadyAt by design; `start_cell.go:156` unreachable since the recovery switch claims Degraded first). `cellStateIsSticky`/`cellStateAutoDeleteTriggers` correctly exclude it (non-sticky/non-terminal). CLI routing (`run`/`restart`/`start`) and the daemon recovery switch agree; tests cover the behavior flip and all three verbs; docs accurate; commits verified; CI green. Nothing to add over the prior rounds.

eminwux:
**PR #1318 Review — `feat: add Degraded cell state for cells with a down or restarting workload`**

LGTM — adds a non-sticky, non-terminal `CellStateDegraded` (appended last in both enums, ordinal lockstep preserved) so a partially-healthy cell stops reporting a contradictory `Ready 1/2`. The new `nonRootWorkloadDegraded` predicate is consulted only in the terminal branch per its contract, and the `restartFired`/`restartDeferred` hold-at-Degraded logic is sound (restart only fires when `ReadyObserved` is already latched, so persisting Degraded without resetting the latch is correct).

The main risk for a new enum member is a missed classification site — I swept every cell-state switch/predicate in the tree: `Degraded` is correctly grouped with `Ready` as live (`IsCellRunning`, `GuardCellTaskLiveness`, `VolumeMountedByLiveCell`), with `Error`/`Failed` as recoverable (`StartCell` recreate switch, `restart.go`, `run.go`), and correctly excluded from the sticky/auto-delete/`ReadyAt` sites (`cellStateIsSticky`, `cellStateAutoDeleteTriggers`, `stampCellLifecycle`). No gaps. Tests cover the derivation matrix, the predicate truth table, the three recovery verbs, and the marshal round-trip + int bound; all 4 commits verified, CI green, and the `make dev-init` smoke is documented as run on-device.

PR #1327, opened 2026-06-16
fix: resolve run-path conflist and sweep leaked masquerade chains on CNI purge

### LINKED ISSUE #1324: cni: per-cell masquerade nat rules leak on cell/realm purge — findCNIConfigPath resolves the wrong conflist path, CNI DEL is silently skipped
## Context

Purging a CNI-networked cell (directly, or via cascade realm purge) leaves the bridge plugin's **per-container masquerade nat rules orphaned** in the host `nat` table. The IPAM lease and CNI cache are cleaned, so purge reports success, but the `CNI-<hash>` masquerade chain and its `POSTROUTING` jump survive forever. They accumulate with every networked-cell purge.

Surfaced during a `KUKEON_PROFILE=dev` `make dev-init` run: the attach-smoke cell `dev-init-attach/ds/dks/cattach` (pod `10.89.2.2`) was created then purged, and afterwards:

```
# iptables -t nat -S   (cell long gone; 10.89.2.2 belongs to nothing)
-N CNI-ec12256cec4057b1122e3662
-A POSTROUTING -s 10.89.2.2/32 -m comment --comment "name: \"dev-init-attach-ds\" id: \"ds_dks_cattach_root\"" -j CNI-ec12256cec4057b1122e3662
-A CNI-ec12256cec4057b1122e3662 -d 10.89.2.0/24 ... -j ACCEPT
-A CNI-ec12256cec4057b1122e3662 ! -d 224.0.0.0/4 ... -j MASQUERADE
```

`/var/lib/cni/networks/` was empty (IPAM released) and the realm/space/stack/cell metadata was fully purged — only the nat masquerade rules leaked.

Host connectivity is **not** affected (the rules only match a source IP that no longer exists); this is a host-state resource leak, not an outage.

## Root cause

The daemon log proves the CNI DEL was never run during purge:

```
level=INFO msg="purged CNI resources for container" container=ds_dks_cattach_root purged=[ipam-file:10.89.2.2]
```

`purged=[ipam-file:...]` contains **no `cni-del`** entry. Per `internal/controller/runner/helpers.go:73`, `"cni-del"` is only appended when `DelContainerFromNetwork` succeeds — so the bridge-plugin DEL (the only thing that tears down the ipMasq chain) never ran. The container task was still RUNNING at DEL time (its stop timed out 10s later), so this is **not** the netns-gone case fixed in #1174 or the daemon-netns case in #1219.

The skip happens at the conflist lookup. `purgeCNIForContainer` calls `findCNIConfigPath(networkName)` and silently skips the whole DEL block when it errors:

`internal/controller/runner/helpers.go:53-81`
```go
if networkName != "" {
    cniConfigPath, err := r.findCNIConfigPath(networkName)
    if err == nil {            // <-- err != nil => entire DEL block skipped, no log above Debug
        ...
        if err = cniMgr.DelContainerFromNetwork(r.ctx, containerID, netnsPath); err == nil {
            purged = append(purged, "cni-del")
        } else {
            r.logger.DebugContext(...)   // <-- DEL failure swallowed at Debug, returns nil
        }
    }
}
```

`findCNIConfigPath` only ever checks `<CniConfigDir>/<networkName>.conflist`, and its documented run-path fallback is unimplemented:

`internal/controller/runner/helpers.go:378-388`
```go
func (r *Exec) findCNIConfigPath(networkName string) (string, error) {
    configPath := filepath.Join(r.cniConf.CniConfigDir, networkName+".conflist")
    if _, err := os.Stat(configPath); err == nil {
        return configPath, nil
    }
    // Try to find in run path (space network configs)
    // This requires listing spaces, which we'll do as a fallback
    return "", fmt.Errorf("CNI config not found for network %q", networkName)   // <-- fallback never implemented
}
```

But space conflists are written per-space as `network.conflist` under the run path, **not** as `<networkName>.conflist` under `CniConfigDir`. The cell was created against `/opt/kukeon/data/dev-init-attach/ds/network.conflist` (from the create log: `cniConfig=/opt/kukeon/data/dev-init-attach/ds/network.conflist`), and on disk:

```
/opt/kukeon/data/default/default/network.conflist
/opt/kukeon/data/kuke-system/kukeon/network.conflist
# no dev-init-attach-ds.conflist exists anywhere
```

So `findCNIConfigPath("dev-init-attach-ds")` looks for `dev-init-attach-ds.conflist`, finds nothing, returns an error → DEL block skipped → masquerade rules leak. The IPAM-file removal further down (`helpers.go:84-128`) is path-derived from `networkName` rather than the conflist, so it still succeeds — which is exactly why the log shows `ipam-file` cleaned but no `cni-del`.

`purge_cell.go:167` (and `purge_stack.go`) call `purgeCNIForContainer(containerID, netnsPath, networkName)` with only the network name, so they fully depend on this broken lookup. `teardownSpaceCNI`/`TeardownNetwork` (space-purge path) only removes the bridge link + conflist file; it does not remove the per-container `CNI-<hash>` masquerade chains, which are keyed by container ID and only removable by a per-container DEL.

This is a **separate** leak from #1174 (DEL gated on netnsPath) and #1219 (DEL against daemon netns) — both already fixed, both moot here because the conflist lookup fails before any netns logic runs.

## Proposal

1. Fix `findCNIConfigPath` to resolve the per-space run-path conflist (`/opt/kukeon/data/<realm>/<space>/network.conflist`), or thread the known conflist path into `purgeCNIForContainer` from the purge callers (they already know the realm/space and read that conflist at create time). Either resolves the lookup so CNI DEL actually runs.
2. Stop silently swallowing a *skipped* or *failed* CNI DEL during purge. A failed/skipped DEL that leaks host nat rules should surface above Debug (Warn) and be reflected in the purge result, not reported as clean. (`helpers.go:53-81`.)
3. Consider a belt-and-suspenders sweep: on purge, best-effort delete the deterministic `CNI-<FormatChainName(network, containerID)>` chain + its `POSTROUTING -s <podIP>` jump even when the conflist can't be found, so a config-lookup miss can't strand host nat state.

## Acceptance criteria

- [ ] After `kuke purge` (cell, and cascade realm) of a CNI-networked cell, `iptables -t nat -S` contains no residual `CNI-<hash>` chain or `POSTROUTING -s <podIP> -j CNI-<hash>` rule for the purged cell.
- [ ] The purge log/result reflects a real CNI DEL (`cni-del` in the purged list) for networked cells, and a failed/skipped DEL is surfaced above Debug.
- [ ] A regression test covers the create-networked-cell → purge → assert-no-leaked-nat-rule cycle (or, where iptables isn't available in CI, asserts `findCNIConfigPath` resolves the per-space run-path conflist and `purgeCNIForContainer` records `cni-del`).

## Out of scope

- Host connectivity / FORWARD admission chains (`KUKEON-EGRESS`/`KUKEON-FORWARD`) — verified healthy in the same run.
- The already-fixed netns-gating (#1174) and daemon-netns (#1219) DEL paths.

## References

- `internal/controller/runner/helpers.go:378-388` — `findCNIConfigPath` (wrong path/name, unimplemented fallback)
- `internal/controller/runner/helpers.go:53-81` — DEL block skip + swallowed-error path in `purgeCNIForContainer`
- `internal/controller/runner/purge_cell.go:167`, `internal/controller/runner/purge_stack.go` — callers passing only `networkName`
- `internal/cni/container.go:87-96` — `DelContainerFromNetwork` → libcni `DelNetworkList`
- Prior related fixes: #1174 (DEL-while-stopped netns gate), #1219 (daemon-own netns)
- Repro: `KUKEON_PROFILE=dev make dev-init` (attach-smoke cell `dev-init-attach/ds/dks/cattach`, pod `10.89.2.2`); daemon log at `/opt/kukeon/data/kuke-system/kukeon/kukeon/kukeond/kukeond/log`

eminwux:
## Summary

Purging a CNI-networked cell (directly or via cascade realm purge) left the bridge plugin's per-container `CNI-<hash>` masquerade chain and its `POSTROUTING -s <podIP>` jump stranded in the host `nat` table, accumulating on every networked-cell purge (issue #1324). This PR fixes it in three parts:

1. **`findCNIConfigPath` run-path fallback** (`helpers.go`). The documented-but-unimplemented fallback now resolves the per-space conflist at `<RunPath>/<metadata>/<realm>/<space>/network.conflist`. Space conflists are never written as `<networkName>.conflist` under `CniConfigDir`, so the old lookup always missed for space networks and the CNI DEL was silently skipped. Resolution enumerates realm/space dirs and matches by the recomputed network name (robust to `-` in realm/space names, unlike reverse-splitting `<realm>-<space>`). Keeps the signature unchanged so all ~15 `purgeCNIForContainer` callers benefit at once — lowest blast radius.

2. **Surface skipped/failed CNI DEL** (`runCNIDel` in `helpers.go`). Every non-success DEL path (conflist not found, manager build error, config-list load error, DEL failure) now logs at **Warn** (was Debug or fully swallowed) and records its outcome (`cni-del-skipped:*` / `cni-del-failed`) in the purge result instead of reporting a clean purge.

3. **Belt-and-suspenders masquerade-chain sweep** (`sweepCNIMasqChain` in `helpers.go`). After the DEL, directly remove the deterministic `CNI-<FormatChainName(network, containerID)>` chain and its `POSTROUTING` jumps. **See the premise-rot note below — this turned out to be required, not optional.**

### Premise-rot finding (documented per dev CLAUDE.md)

The issue diagnosed the leak as purely a conflist-lookup miss (fix #1) and listed the chain sweep (#3) as an optional "consider". Validating fix #1 on-device surfaced a **deeper root cause**: even when the DEL runs and returns success (`purged=[cni-del]`), the masquerade chain still leaks on the purge path. The bridge plugin's `cmdDel` (`containernetworking/plugins@v1.9.0` bridge.go:788-789) early-returns after the IPAM release when invoked with an **empty netns** — never reaching its `TeardownIPMasqForNetworks` call. On the purge path the container task is already gone, so the DEL always runs with an empty netns. I confirmed live: a live-netns DEL (`kuke stop`) cleans the chain; the empty-netns DEL (`kuke purge`) does not. So fix #1+#2 satisfy AC #2 but **cannot** satisfy AC #1 alone — the sweep (#3) is the only thing that reclaims the chain when there is no live netns. The underlying intent (no leaked nat rules after purge) is unchanged, so I implemented #3 as a core part of the fix rather than stalling.

### Design calls (for reviewer pushback)

- **Shell out to iptables rather than import `containernetworking/plugins`.** `ip.TeardownIPMasqForNetworks` would be the exact teardown, but `plugins` is not a current dependency and would pull a new closure (`plugins` + `coreos/go-iptables`) for three commands. The project's established convention (`internal/netpolicy.IptablesEnforcer`) shells out to `iptables` via `exec.Command`; the sweep matches that. Adopting a new dependency is a call I avoided making unilaterally.
- **Deterministic chain-name replica.** `cniMasqChainName` reproduces `utils.FormatChainName` (`"CNI-" + hex(sha512(network+containerID))[:24]`). Verified byte-for-byte against two live-created chains in `TestCNIMasqChainName` (golden values), including the issue's exact `CNI-ec12256cec4057b1122e3662`.
- **Jump removal by line number**, parsed from `iptables -L POSTROUTING --line-numbers -n` and deleted descending — robust to the rule's source IP and comment text, no IP capture or comment-string matching required.

## Test plan

- [x] `make test` (full CI suite: `test-root` + `test-kukebuild`) — pass
- [x] `pre-commit run --files <changed>` — pass (golangci-lint incl. `gochecknoglobals` nolint, gofmt, addlicense)
- [x] New unit tests (`internal/controller/runner/helpers_test.go`):
  - `TestFindCNIConfigPath_ResolvesRunPathConflist` — run-path fallback resolves the per-space conflist (realm name with `-` exercises no-reverse-split)
  - `TestPurgeCNIForContainer_RunPathConflistRunsDEL` — DEL fires via the run-path conflist and records `cni-del` (fake CNI plugin)
  - `TestPurgeCNIForContainer_SurfacesSkippedDEL` — skipped DEL surfaces at Warn + `cni-del-skipped:config-not-found`
  - `TestCNIMasqChainName` — chain-name replica matches live chains (golden)
  - `TestSweepCNIMasqChain` / `_NoChainNoMarker` — sweep deletes jumps (descending) + flush/delete chain; no false marker when chain absent
  - `TestMain` default-stubs the iptables shell-out so no unit test touches the host firewall
- [x] **On-device end-to-end** (`KUKEON_PROFILE=dev make dev-init`, exit 0):
  - Clean cell create→start→purge (`natleak3`): chain `CNI-fa5b2f0a...` + `POSTROUTING` jump fully reclaimed; daemon log: `purged="[cni-del cni-masq-swept:CNI-fa5b2f0a6a8eeb62a3411ecd]"`
  - **Cascade realm purge** (dev-init attach-smoke `dev-init-attach/ds/dks/cattach`): the issue's exact persistent leak `CNI-ec12256cec4057b1122e3662` swept; `iptables -t nat -S` shows no `dev-init-attach` residue
  - Daemon-parity tail + attach smoke pass

## Acceptance criteria

- [x] After `kuke purge` (cell, and cascade realm) of a CNI-networked cell, `iptables -t nat -S` contains no residual `CNI-<hash>` chain or `POSTROUTING -s <podIP> -j CNI-<hash>` rule for the purged cell — validated on-device for both paths.
- [x] The purge log/result reflects a real CNI DEL (`cni-del` in the purged list) for networked cells, and a failed/skipped DEL is surfaced above Debug (Warn).
- [x] A regression test covers the resolution + sweep (iptables-free unit tests asserting `findCNIConfigPath` resolves the run-path conflist, `purgeCNIForContainer` records `cni-del`, and the masquerade-chain sweep issues the expected delete/flush/delete sequence), plus on-device validation of the full create→purge→no-leaked-nat-rule cycle.

## Out of scope

- Retroactive GC of chains leaked by pre-fix purges (the fix prevents new leaks; it does not reclaim historical orphans the issue did not ask to GC).
- Host connectivity / `KUKEON-EGRESS`/`KUKEON-FORWARD` (verified healthy).
- The already-fixed netns-gating (#1174) and daemon-netns (#1219) DEL paths.

Closes #1324

### DIFF
--- a/internal/controller/runner/helpers.go
+++ b/internal/controller/runner/helpers.go
@@ -18,11 +18,14 @@ package runner
 
 import (
 	"context"
+	"crypto/sha512"
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -48,35 +51,23 @@ func (r *Exec) purgeCNIForContainer(containerID, netnsPath, networkName string) 
 	// lease.
 	netnsPath = r.sanitizeDELNetns(netnsPath)
 
-	// Always try to call CNI DEL if network name is available
-	// Some CNI plugins may handle empty/invalid netns gracefully
+	// Always try to call CNI DEL if network name is available; some CNI plugins
+	// handle empty/invalid netns gracefully. A skipped or failed DEL leaves the
+	// bridge plugin's per-container masquerade nat chains stranded on the host,
+	// so runCNIDel surfaces each failure mode above Debug and records its outcome
+	// in the purge result rather than reporting a clean purge (issue #1324).
 	if networkName != "" {
-		cniConfigPath, err := r.findCNIConfigPath(networkName)
-		if err == nil {
-			var cniMgr *cni.Manager
-			cniMgr, err = cni.NewManager(
-				r.cniConf.CniBinDir,
-				r.cniConf.CniConfigDir,
-				r.cniConf.CniCacheDir,
-			)
-			if err == nil {
-				if err = cniMgr.LoadNetworkConfigList(cniConfigPath); err == nil {
-					// Attempt CNI DEL even when netnsPath is empty. libcni's
-					// DelNetworkList is netns-optional: it reconstructs teardown
-					// (including the bridge plugin's ipMasq iptables chains/rules)
-					// from the cached CNI result, which the cache-removal block
-					// below only clears *after* this DEL has run. A stopped cell
-					// has no containerd task and thus no netns path, so gating DEL
-					// on netnsPath != "" permanently leaked the per-cell CNI-*
-					// masquerade chains on every delete-while-stopped (issue #1174).
-					if err = cniMgr.DelContainerFromNetwork(r.ctx, containerID, netnsPath); err == nil {
-						purged = append(purged, "cni-del")
-						r.logger.DebugContext(r.ctx, "called CNI DEL for container", "container", containerID)
-					} else {
-						r.logger.DebugContext(r.ctx, "CNI DEL failed (netns may be invalid)", "container", containerID, "error", err)
-					}
-				}
-			}
+		purged = append(purged, r.runCNIDel(containerID, netnsPath, networkName))
+		// Belt-and-suspenders: the plugin-chain DEL above only tears down the
+		// bridge plugin's per-container masquerade chain when it runs with a
+		// LIVE netns. On the purge path the container task is already gone, so
+		// the DEL runs with an empty netns and the bridge plugin early-returns
+		// after the IPAM release — never reaching its ipMasq teardown
+		// (containernetworking/plugins bridge cmdDel). This sweep reclaims the
+		// stranded chain + its POSTROUTING jump directly so a no-netns DEL (or a
+		// conflist-lookup miss) can't leak host nat state (issue #1324).
+		if marker := r.sweepCNIMasqChain(networkName, containerID); marker != "" {
+			purged = append(purged, marker)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #1327 Review — `fix: resolve run-path conflist and sweep leaked masquerade chains on CNI purge`**

Correct and well-tested. The three-part fix (run-path conflist fallback → surfaced DEL outcomes → masquerade-chain sweep) is sound: the per-container chain name hashes `network+containerID`, so a sweep can never touch a sibling container's chain or jump; descending line-number deletion avoids renumbering; the `Atoi` guard correctly skips header rows; and `cniMasqChainName` is byte-for-byte `utils.FormatChainName`, pinned with golden values. The premise-rot escalation to make the sweep mandatory is well-justified and validated on-device for both the cell-purge and cascade-realm-purge paths. dev-init smoke is in the test plan, all commits are signature-verified, no dead code, no leaks.

One non-blocking nit:

- **Stale comment** `internal/controller/runner/helpers.go:194-197`: the retained `runCNIDel` comment still claims the netns-optional DEL "reconstructs teardown (including the bridge plugin's ipMasq iptables chains/rules) from the cached CNI result" — but this PR's premise-rot finding establishes the bridge plugin early-returns before ipMasq teardown on an empty netns, which is exactly why the sweep exists. Worth a one-line correction so a future reader isn't re-misled. The code is correct as-is.

PR #1329, opened 2026-06-16
fix: resolve daemon namespace under non-default suffix in kuke status STATE/STORAGE

### LINKED ISSUE #1325: image: kuke image */kuke get image* ignore containerdNamespaceSuffix — target <realm>.kukeon.io under dev profile, can't see/load the daemon's images

### LINKED ISSUE #1326: status: kuke status STATE flags live dev namespaces as 'residual' (suggests deleting them) and STORAGE measures wrong namespaces under dev profile
## Context

Under `KUKEON_PROFILE=dev` (the canonical dev-init smoke per `kukeon/CLAUDE.md`), `kuke status` mis-resolves containerd namespaces in its in-process `STATE` and `STORAGE` sections — both ignore the configured `containerdNamespaceSuffix` and assume the default `<realm>.kukeon.io` suffix. The daemon-routed `PARITY` section is correct, which makes the inconsistency stand out within a single report.

Observed against the running dev daemon (`KUKE_CONFIGURATION=/root/kukeon/kuke-dev.yaml`, suffix `dev.kukeon.io`; `KUKEON_HOST=unix:///run/kukeon-dev/kukeond.sock`):

```
$ kuke status
...
STATE
  run-dir          OK    /run/kukeon (no orphan entries)
  containerd-ns    WARN  residual namespaces: default.dev.kukeon.io, kuke-system.dev.kukeon.io, kuke-system.dev.kukeon.io_history
                         ↳ `kuke uninstall` or hand-cleanup with `ctr -n <ns> namespace remove`

STORAGE
  default          OK    default.kukeon.io (0 snapshots, 0 leases, 0 blobs, 0 B)
  kuke-system      OK    kuke-system.kukeon.io (0 snapshots, 0 leases, 0 blobs, 0 B)

PARITY
  realms           OK    2 items match (default, kuke-system)
  ...              OK    (all kinds match — correctly sees the dev-profile data)

Status: OK
```

Two distinct defects:

1. **`STATE` false-flags the live namespaces as residual.** The `*.dev.kukeon.io` namespaces are the ones the running daemon actively uses (the kukeond cell lives in `kuke-system.dev.kukeon.io`). The check lists them as "residual" and the remediation suggests `ctr -n <ns> namespace remove` — **following it would delete the running daemon's namespaces and its data.** A dangerous false positive on every dev-profile host.

2. **`STORAGE` measures the wrong namespaces.** It reports `default.kukeon.io` / `kuke-system.kukeon.io` (default suffix, `0 snapshots, 0 leases, 0 blobs`) instead of the populated `*.dev.kukeon.io` namespaces, so the per-realm footprint is always zero under the dev profile.

## Diverged invariant

`docs/cli-use-cases.md` → "Status snapshot (`kuke status`)":

> Sections … `STATE` (no orphan run-dir sockets, no residual containerd namespaces), `STORAGE` (per-realm snapshot / lease / content-blob counts + summed bytes) …

Under the dev profile, `STATE` reports the *active* namespaces as residual and `STORAGE` counts the wrong (empty) namespaces. Both should resolve the configured suffix, as `PARITY` already does.

## Root cause

Both sections resolve the namespace via `consts.RealmNamespace(realm)` (= `realm + consts.RealmNamespaceSuffix`, default `.kukeon.io`):

- `cmd/kuke/status/checks.go:395` — `ns := consts.RealmNamespace(realms[i].Metadata.Name)` builds the *expected* set; any actual containerd namespace not in it is flagged residual (`checks.go:415`). With the default suffix, every `*.dev.kukeon.io` namespace looks unexpected.
- `cmd/kuke/status/storage.go:123` — `ns := consts.RealmNamespace(realm)` selects the namespace whose footprint is summed.

`RealmNamespaceSuffix` is overridden only by `consts.ConfigureRuntime` (`internal/consts/consts.go:303`/`:249`), which the client root drives from `KUKE_CONFIGURATION` (`cmd/kuke/kuke.go:315`, `:364-368`). As with #1325 (the `kuke image*` / `kuke get image*` family), the in-process `kuke status` STATE/STORAGE path does not see that override — so `consts.RealmNamespace` returns the default suffix. `PARITY` is correct because it routes through the daemon, which ran `ConfigureRuntime` with the dev suffix at its own startup.

**Shared root cause with #1325** — in-process containerd namespace resolution ignores the configured `containerdNamespaceSuffix`. A unified fix (apply `applyClientConfiguration` + `consts.ConfigureRuntime` on these in-process command paths, or thread the configured suffix into `consts.RealmNamespace` call sites) resolves both. Filed separately because the `kuke status` symptom is distinct and the residual-namespace remediation is actively dangerous.

## Proposal

1. Ensure the `kuke status` command applies the configured `containerdNamespaceSuffix` (run the same `ConfigureRuntime` the root client path does) before the STATE/STORAGE checks evaluate `consts.RealmNamespace`.
2. Alternatively / additionally, make the residual-namespace detector suffix-aware: compute the expected set from the resolved suffix so a dev-profile host's live namespaces are never flagged residual.
3. Regression test: with `containerdNamespaceSuffix: dev.kukeon.io`, `kuke status` STATE does not flag `*.dev.kukeon.io` as residual and STORAGE reports the footprint of `<realm>.dev.kukeon.io` (non-zero when populated).

## Acceptance criteria

- [ ] With a dev-suffix client config, `kuke status` STATE does **not** list the active `*.dev.kukeon.io` namespaces as residual (no dangerous `ctr namespace remove` remediation against live namespaces).
- [ ] `kuke status` STORAGE reports the footprint of `<realm>.<configured-suffix>` (matches the namespaces `PARITY` and the daemon use).
- [ ] Default profile behavior unchanged (`<realm>.kukeon.io` when no suffix configured).

## Out of scope

- The `kuke image*` / `kuke get image*` namespace-suffix bug (#1325) — same root cause, separate command surface.
- The `PARITY` section — already correct.

## References

- `docs/cli-use-cases.md` → "Status snapshot (`kuke status`)" (STATE / STORAGE invariants)
- `cmd/kuke/status/checks.go:395` (expected-namespace build), `:415` (residual detail)
- `cmd/kuke/status/storage.go:123` (per-realm footprint namespace)
- `internal/consts/consts.go:303` (`RealmNamespace`), `:249` (`ConfigureRuntime`)
- `cmd/kuke/kuke.go:315`, `:364-368` (client suffix → ConfigureRuntime)
- Related (shared root cause): #1325
- Repro: dev daemon from `KUKEON_PROFILE=dev make dev-init`; `kuke status` with `KUKE_CONFIGURATION=kuke-dev.yaml`

eminwux:
## Summary

`kuke status` mis-resolved containerd namespaces in its in-process `STATE` and `STORAGE` sections under a non-default `containerdNamespaceSuffix` (the dev profile's `dev.kukeon.io`). Both recomputed the namespace via `consts.RealmNamespace(realm)`, which reads the process-global `RealmNamespaceSuffix`. That global is only set to the dev suffix when `ConfigureRuntime` runs with the client config — which the in-process status path does not pick up — so it stayed at the default `.kukeon.io`:

- **STATE** built the *expected* namespace set with the default suffix, so the live `*.dev.kukeon.io` namespaces the running daemon uses were flagged **residual** with a `ctr -n <ns> namespace remove` remediation — following it would **delete the running daemon's namespaces and data**.
- **STORAGE** probed `<realm>.kukeon.io` (default suffix, empty) instead of the populated `<realm>.dev.kukeon.io`, so the per-realm footprint always read zero.

The `PARITY` section was already correct because it routes through the daemon. This change makes `STATE`/`STORAGE` use the **same daemon-authoritative source `PARITY` uses**: each realm's `Spec.Namespace` (the namespace the daemon resolved at its own startup with the configured suffix) rather than recomputing it in-process.

### Why `Spec.Namespace` rather than the issue's literal proposals
The issue proposed (1) running `ConfigureRuntime` on the status path or (2) making the residual detector "suffix-aware" from the resolved suffix. Investigation showed the deeper root cause is that **`KUKE_CONFIGURATION` is never bound to viper** — `scripts/dev-init.sh:332-333` asserts `KUKE_CONFIGURATION.BindEnv` exists in `cmd/kuke/kuke.go`'s `loadConfig`, but it does not (verified: passing the dev config via the `KUKE_CONFIGURATION` env var is silently ignored, while `--configuration` is honored). So both literal proposals would still read the unconfigured default suffix in-process and **not** fix the bug. Using the realm's daemon-resolved `Spec.Namespace` fixes `STATE`/`STORAGE` precisely, is the lowest-blast-radius change (two localized call sites, no PreRun wiring touched), and is robust regardless of the env-binding gap.

This deeper env-binding gap is the **shared root cause with #1325** (out of scope here per the issue). It is left for the unified fix #1325 contemplates; this PR does not touch the `kuke image*` surface. The unbound-env / false dev-init comment is surfaced here for the reviewer/PM rather than fixed in this scoped PR.

A realm whose `Spec.Namespace` is unset falls back to the recomputed `consts.RealmNamespace(name)`, preserving default-profile behavior.

## Test plan
- [x] `make test-root` (full root module suite) — pass
- [x] `make test-kukebuild` — pass
- [x] `GOWORK=off go test ./cmd/kuke/status/...` — pass, incl. new regression tests
- [x] `GOWORK=off go vet ./cmd/kuke/status/...` — clean
- [x] `pre-commit run --files <changed>` — all hooks pass (gofmt, golangci-lint, addlicense, go-mod-tidy)
- [ ] `KUKEON_PROFILE=dev make dev-init` end-to-end smoke — not run; requires bootstrapping a dev-profile daemon on the host. The change is CLI-only (does not touch the build path, daemon, or `/opt/kukeon`), and the dev-profile `STATE`/`STORAGE` behavior is covered by the two new unit regression tests below.

### New regression tests
- `TestCheckStateNamespaces_DevSuffixNotResidual` — dev-suffix realms + live `*.dev.kukeon.io` ctr namespaces ⇒ STATE is OK, no `residual` flag.
- `TestCheckStorage_DevSuffixProbesLiveNamespaces` — STORAGE probes the realms' `Spec.Namespace` (`*.dev.kukeon.io`) and reports their footprint, not the empty default-suffix namespaces.

## Acceptance criteria
- [x] Dev-suffix client config: STATE does **not** list active `*.dev.kukeon.io` as residual (no dangerous remediation against live namespaces).
- [x] STORAGE reports the footprint of `<realm>.<configured-suffix>` (matches PARITY / the daemon).
- [x] Default profile behavior unchanged (`<realm>.kukeon.io` when no suffix configured) — unset `Spec.Namespace` falls back to `consts.RealmNamespace`; existing tests still green.

Closes #1326

### DIFF
--- a/cmd/kuke/status/checks.go
+++ b/cmd/kuke/status/checks.go
@@ -26,7 +26,27 @@ import (
 	"time"
 
 	"github.com/eminwux/kukeon/internal/consts"
+	v1beta1 "github.com/eminwux/kukeon/pkg/api/model/v1beta1"
 )
+
+// realmNamespace returns the containerd namespace to attribute to a realm,
+// preferring the daemon-resolved Spec.Namespace over a recomputed
+// consts.RealmNamespace. The recomputation reads the process-global
+// RealmNamespaceSuffix, which the in-process status STATE/STORAGE path does
+// not pick up from a KUKE_CONFIGURATION client config — so under a
+// non-default containerdNamespaceSuffix (the dev profile's dev.kukeon.io) it
+// returns the default-suffix namespace, false-flagging the live
+// *.dev.kukeon.io namespaces as residual and measuring the wrong (empty)
+// namespaces in STORAGE (issue #1326). Spec.Namespace is the same
+// authoritative value the daemon and the PARITY section already use. Falls
+// back to the recomputed form for a realm whose Spec.Namespace is unset
+// (e.g. a hand-built test realm or a pre-Spec.Namespace persisted doc).
+func realmNamespace(r v1beta1.RealmDoc) string {
+	if r.Spec.Namespace != "" {
+		return r.Spec.Namespace
+	}
+	return consts.RealmNamespace(r.Metadata.Name)
+}
 
 // Section names — kept as constants so the renderer's section grouping and
 // the JSON `section` field are spelled the same way everywhere.
@@ -392,7 +412,7 @@ func checkStateNamespaces(ctx context.Context, rc *runCtx) Result {
 	// in this set and falls through to the residual list below (issue #1183).
 	expectedNS := make(map[string]bool, len(realms))
 	for i := range realms {
-		ns := consts.RealmNamespace(realms[i].Metadata.Name)
+		ns := realmNamespace(realms[i])
 		expectedNS[ns] = true
 		expectedNS[consts.BuildKitHistoryNamespace(ns)] = true
 	}
--- a/cmd/kuke/status/status_test.go
+++ b/cmd/kuke/status/status_test.go
@@ -186,6 +186,35 @@ func TestCheckStateNamespaces_LiveRealmHistoryIsClean(t *testing.T) {
 	}
 }
 
+// TestCheckStateNamespaces_DevSuffixNotResidual is the issue #1326 regression
+// guard: under a non-default containerdNamespaceSuffix (the dev profile's
+// dev.kukeon.io), the live *.dev.kukeon.io namespaces the daemon actively uses
+// must not be flagged residual. The check attributes each realm's namespace
+// from its daemon-resolved Spec.Namespace, so the expected set matches the
+// actual namespaces and the dangerous `ctr namespace remove` remediation
+// against live namespaces never fires.
+func TestCheckStateNamespaces_DevSuffixNotResidual(t *testing.T) {
+	rc := &runCtx{
+		daemonClient: newFakeClient().
+			withRealmNamespace("default", "default.dev.kukeon.io").
+			withRealmNamespace("kuke-system", "kuke-system.dev.kukeon.io"),
+		ctrClient: &fakeCtrClient{namespaces: []string{
+			"default.dev.kukeon.io",
+			"kuke-system.dev.kukeon.io",
+			"kuke-system.dev.kukeon.io_history",
+		}},
+		logger: testLogger(),
+	}
+
+	r := checkStateNamespaces(context.Background(), rc)
+	if r.Status != StatusOK {
+		t.Fatalf("dev-suffix live namespaces must not be residual; got %s (%q)", r.Status, r.Detail)
+	}
+	if strings.Contains(r.Detail, "residual") {
+		t.Errorf("Detail must not flag residual namespaces; got %q", r.Detail)
+	}
+}
+
 // TestCheckParity_NestedDescent confirms the walk recurses into the
 // realm-set intersection (spaces / stacks / cells / containers) — a
 // regression in the descent loop would silently stop reporting nested
@@ -609,6 +638,19 @@ func (f *fakeClient) withRealms(names ...string) *fakeClient {
 			Metadata: v1beta1.RealmMetadata{Name: name},
 		})
 	}
+	return f
+}
+
+// withRealmNamespace seeds one realm carrying an explicit daemon-resolved
+// Spec.Namespace — the shape a non-default containerdNamespaceSuffix (the dev
+// profile's dev.kukeon.io) produces. The STATE/STORAGE checks attribute the
+// namespace from Spec.Namespace rather than recomputing it, so this exercises
+// the suffix-aware path (issue #1326).
+func (f *fakeClient) withRealmNamespace(name, namespace string) *fakeClient {
+	f.realms = append(f.realms, v1beta1.RealmDoc{
+		Metadata: v1beta1.RealmMetadata{Name: name},
+		Spec:     v1beta1.RealmSpec{Namespace: namespace},
+	})
 	return f
 }
 
--- a/cmd/kuke/status/storage.go
+++ b/cmd/kuke/status/storage.go
@@ -60,8 +60,8 @@ func checkStorage(ctx context.Context, rc *runCtx) []Result {
 		}}
 	}
 
-	realms := enumerateRealmsForStorage(ctx, rc)
-	if len(realms) == 0 {
+	targets := enumerateRealmsForStorage(ctx, rc)
+	if len(targets) == 0 {
 		return []Result{{
 			Section: sectionStorage,
 			Name:    "realms",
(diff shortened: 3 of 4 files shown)


eminwux:
**PR #1329 Review — `fix: resolve daemon namespace under non-default suffix in kuke status STATE/STORAGE`**

LGTM — reads each realm's daemon-resolved `Spec.Namespace` instead of recomputing via the process-global suffix the in-process status path never picks up, so STATE no longer false-flags live `*.dev.kukeon.io` namespaces as residual (dangerous `ctr namespace remove`) and STORAGE probes the populated namespaces; unset `Spec.Namespace` falls back to `consts.RealmNamespace`, preserving default-profile behavior. Both regression tests exercise the real pre-fix failure, all symbols used, commits verified, change is CLI-only so the unchecked host-gated dev-init smoke is the documented carve-out.

eminwux:
**Heads-up (non-blocking, merge-time metadata):** GitHub's `closingIssuesReferences` for this PR lists **both #1326 and #1325**, so merging will auto-close #1325 too — even though the body explicitly scopes #1325 out ("left for the unified fix #1325 contemplates") and `Closes #1326` is the only closing keyword in the description. #1325 was likely linked manually via the Development sidebar. If #1325 should stay open for the future unified fix, unlink it from this PR before merging. Not a code issue — the diff is correct and LGTM stands.

PR #1334, opened 2026-06-17
fix: exclude restart-policy-pinned root from StartCell running-container guard on Exited cells

### LINKED ISSUE #1316: run/start: Exited cell with a restart-policy-pinned live root cannot be restarted — running-container guard trips on the root
## Repro

A cell with a non-root workload whose `restartPolicy` is `on-failure` (and the workload exits cleanly, code 0) or `never`:

1. Workload exits 0. The reconciler derives `CellStateExited`.
2. `restartPolicyPermitsCellReap(cell)` returns **false** (clean exit + `on-failure`, or `never`) — #1003 — so `shouldWindDownCell` is vetoed (`internal/controller/runner/refresh.go:803`) and the **root container is left running by design**.
3. `kuke get cell -o wide` shows the cell `Exited` with `READY 1/2` (root running, non-root terminal).
4. `kuke run <cell>` (or `kuke start` / `kuke restart`):

```
$ kuke get cell -o wide | grep agents-meta-dev
agents-meta-dev-0   default   default   default   Exited   Synced   5d   1/2   k-703e8b3a

$ kuke run agents-meta-dev-0
Error: cell "agents-meta-dev-0" has running containers and must first be stopped
```

The cell is wedged: it won't wind down (restart-policy veto, by design) and it won't restart (the guard refuses). Workaround is `kuke stop <cell>` (kills the leftover root) then `kuke run <cell>`.

## Observed

`kuke run` on an `Exited` cell calls `StartCell` directly (`cmd/kuke/run/run.go:777-793`). The daemon's `StartCell` routes recovery by state (`internal/controller/start_cell.go:103-152`): `Failed`/`Error` route to `RecreateCell` (which handles a leftover-running root), but `Exited` falls through to the running-container guard at `internal/controller/start_cell.go:129-143`:

```go
for _, containerStatus := range internalCell.Status.Containers {
    if containerStatus.State == intmodel.ContainerStateReady {
        hasRunningContainer = true   // counts the ROOT container too
        break
    }
}
if hasRunningContainer {
    return res, fmt.Errorf("cell %q has running containers and must first be stopped", ...)
}
```

The loop iterates **all** `Status.Containers` and does not exclude the root (confirmed by `start_cell_test.go:196-219`, which uses unmarked containers). On an `Exited` cell whose root is deliberately left alive, it trips on the root.

## Root cause

Interaction between two features that don't account for each other:

- **#1003** — a non-root workload with `restartPolicy: on-failure` (clean exit) or `never` makes `restartPolicyPermitsCellReap` return false (`refresh.go:1059-1066`), vetoing the wind-down so the **root container stays running** on an otherwise `Exited` cell.
- **#1268 / #1274** — the StartCell recovery routing special-cases a leftover-running root only for `Failed`/`Error` (→ `RecreateCell`). #1268 explicitly assumed an `Exited` cell has no live container ("exactly like a `Stopped` cell"), so `Exited` falls through to the running-container guard, which trips on the live root.

The guard's own comment at `internal/controller/start_cell.go:126-128` asserts "a leftover-running root on a recoverable terminal cell no longer trips this guard" — true only for `Failed`/`Error`; it misses the #1003 `Exited`-with-pinned-root case.

State derivation confirming the running container is the root: `deriveCellStateFromNonRootContainerStatuses` (`refresh.go:1190`) derives cell state from non-root containers only, so an `Exited` cell has all non-root containers terminal — the one `Ready` container in `1/2` is necessarily the root.

## Suggested fix

Make the running-container guard at `internal/controller/start_cell.go:129-143` exclude the **root** container — cross-reference `Status.Containers` IDs against the non-root spec (the same root/non-root split `deriveCellStateFromNonRootContainerStatuses` and `findRootContainerSpec` already use), so the guard only blocks on live *workloads*, not the cell shell. Alternatively, route `Exited`-with-a-still-running-root through the same `RecreateCell` recovery `Failed`/`Error` use (`start_cell.go:112-121`). The first option is lower blast radius and matches the guard's documented intent. Add a test case mirroring `start_cell_test.go:196-219` but with the running container marked as the root and cell state `Exited`, asserting the start is allowed (and the root is restarted, not refused).

Surfaced alongside #1268 / #1274 (which fixed the `Failed`/`Error` and no-live-root `Exited` paths); this is the remaining `Exited`-with-live-root path those left uncovered.

eminwux:
## Summary

- An `Exited` cell whose non-root workload cleanly exited under a `restartPolicy` that vetoes wind-down (`on-failure` + clean exit, or `never`) deliberately keeps its **root** container alive (#1003 — `restartPolicyPermitsCellReap` returns false, so `shouldWindDownCell` is vetoed and the root is left running by design). Such a cell shows `Exited` with `READY 1/2`.
- `kuke run`/`start`/`restart` on that cell calls `StartCell`, which routes `Failed`/`Error`/`Degraded` to `RecreateCell` (those handle a leftover-running root) but lets `Exited` fall through to the running-container guard at `internal/controller/start_cell.go`. That guard iterated **all** `Status.Containers` and tripped on the deliberately-alive root, returning `cell %q has running containers and must first be stopped` — wedging the cell (it won't wind down by design, and won't restart). The only workaround was `kuke stop` then `kuke run`.
- Fix narrows the guard to live **workloads**: on a cell with a non-root workload spec the **root** container is excluded from the running-container check (cross-referencing `Status.Containers` IDs against the root spec via the same `rootContainerID` / root-vs-non-root split `deriveCellStateFromNonRootContainerStatuses` already uses). This is the issue's suggested option 1 — lower blast radius than re-routing `Exited` through `RecreateCell`, and it matches the guard's documented intent.
- **Guardrail preserved for root-only cells.** The root-exclusion is gated on `hasNonRootContainerSpec`: kukeond-style / cgroup-only cells (no non-root spec) keep the root in the check, because there the root *is* the workload and a running root must still block a duplicate start. A new `hasNonRootContainerSpec` helper is added to the `controller` package (`get_cell.go`), mirroring the runner package's same-named helper — `runner`'s is unexported so it can't be reused across the package boundary, and duplicating the trivial 5-line check keeps the root/non-root split consistent without exporting internals.

## Implementation notes

- The suggested fix referenced `internal/controller/start_cell.go:129-143` and `:126-128`; those line numbers shifted slightly after #1318 added `Degraded` to the recovery switch. The fix lands at the same logical guard (now ~`:140-154`). The guard's doc-comment is updated: the prior "a leftover-running root … no longer trips this guard" claim was true only for the `Failed`/`Error`/`Degraded` recreate routes; it now states the `Exited`-with-pinned-root exclusion explicitly.

## Test plan

- [x] `make test` (full CI suite, `GOWORK=off`) — exit 0, no failures.
- [x] `GOWORK=off go vet ./internal/controller/...` — clean.
- [x] New table case `cell in Exited state with restart-policy-pinned live root allows start` (mirrors `start_cell_test.go`'s `Error`-live-root case but with cell state `Exited`, the running container marked root, and a non-root workload `Exited`): asserts the start is allowed and routes through a plain `StartCell` (not refused).
- [x] New guardrail case `root-only cell with running root still blocks start`: asserts a `Ready` root-only cell with a live root still trips the guard — pins that the root-exclusion does not leak into root-only cells.

Closes #1316

### DIFF
--- a/internal/controller/get_cell.go
+++ b/internal/controller/get_cell.go
@@ -149,6 +149,21 @@ func rootContainerID(cell intmodel.Cell) string {
 	return ""
 }
 
+// hasNonRootContainerSpec reports whether the cell carries at least one
+// non-root (workload) container in spec. It mirrors the runner package's
+// derivation split (deriveCellStateFromNonRootContainerStatuses): a cell with
+// a non-root workload follows its workloads' lifecycle and its root is just the
+// cell shell, whereas a root-only cell (kukeond-style / cgroup-only) has the
+// root standing in as the workload.
+func hasNonRootContainerSpec(cell intmodel.Cell) bool {
+	for i := range cell.Spec.Containers {
+		if !cell.Spec.Containers[i].Root {
+			return true
+		}
+	}
+	return false
+}
+
 // ListCells lists all cells, optionally filtered by realm, space, and/or stack.
 func (b *Exec) ListCells(realmName, spaceName, stackName string) ([]intmodel.Cell, error) {
 	return b.runner.ListCells(realmName, spaceName, stackName)
--- a/internal/controller/start_cell.go
+++ b/internal/controller/start_cell.go
@@ -133,14 +133,27 @@ func (b *Exec) StartCell(cell intmodel.Cell) (StartCellResult, error) {
 	// This prevents blocking starts when containers have crashed externally but metadata still shows Ready state.
 	// Only Ready / Stopped / Exited reach here — the recovery switch above
 	// already claimed Pending / Unknown (refused) and Failed / Error / Degraded
-	// (recreate), so a leftover-running root on a recoverable terminal cell no
-	// longer trips this guard. Degraded is no longer in the fall-through set: an
-	// operator start/restart of a partially-down cell routes through the recovery
-	// recreate above so it lands Ready in a single action (#1317 review).
+	// (recreate). The guard blocks on live *workloads*, not the cell shell: on a
+	// cell with a non-root workload spec the root container is excluded, because
+	// a #1003 restart-policy veto (a non-root workload that exits cleanly under
+	// on-failure, or exits under never) deliberately leaves the root alive on an
+	// otherwise-Exited cell, and that pinned root must not wedge the cell against
+	// `kuke run` / `start` / `restart` (#1316). Root-only cells (kukeond-style /
+	// cgroup-only, no non-root spec) keep the root in the check — there the root
+	// IS the workload, so a running root must still block a duplicate start. The
+	// root/non-root split is the same one deriveCellStateFromNonRootContainer-
+	// Statuses uses to derive the cell state in the first place.
 	if len(internalCell.Status.Containers) > 0 {
-		// Check if any container is actually running
+		rootID := ""
+		if hasNonRootContainerSpec(internalCell) {
+			rootID = rootContainerID(internalCell)
+		}
+		// Check if any non-root container is actually running
 		hasRunningContainer := false
 		for _, containerStatus := range internalCell.Status.Containers {
+			if rootID != "" && containerStatus.ID == rootID {
+				continue
+			}
 			if containerStatus.State == intmodel.ContainerStateReady {
 				hasRunningContainer = true
 				break
--- a/internal/controller/start_cell_test.go
+++ b/internal/controller/start_cell_test.go
@@ -546,6 +546,89 @@ func TestStartCell_ReadyStateValidation(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		{
+			// #1316: an Exited cell whose non-root workload cleanly exited under a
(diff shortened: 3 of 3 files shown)


eminwux:
**PR #1334 Review — `fix: exclude restart-policy-pinned root from StartCell running-container guard on Exited cells`**

LGTM — narrows the running-container guard to live non-root *workloads* so a #1003 restart-policy-pinned root no longer wedges an `Exited` cell against `kuke run`/`start`/`restart`. The exclusion cross-references `Status.Containers[].ID` against `rootContainerID` in the same ID space `deriveCellStateFromNonRootContainerStatuses` uses, the root-only guardrail is correctly gated on `hasNonRootContainerSpec`, and the `Exited`-pinned-root case is functionally complete downstream — `runner.StartCell` kills+restarts the live root via `task.Delete(WithProcessKill)` (`internal/controller/runner/start.go:721`). Both table cases pin the fix (the Exited case fails pre-fix), commits verified, CI green.

